PR #2146, opened 2021-01-07
Bug in clearArea function from costmap_layer

jginesclavero:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu 20.04 Foxy |
| Robotic platform tested on | tb3 Gazebo |

---

I am working in a nav2 layer. Trying to clear an area between iterations I used clearArea method from costmap_layer, and it was not work correctly. Here you have the default behavior.
clearArea_default
And here you have the behavior after the fix.

clearArea_fixed

None

---

### DIFF
--- a/nav2_costmap_2d/src/costmap_layer.cpp
+++ b/nav2_costmap_2d/src/costmap_layer.cpp
@@ -66,15 +66,14 @@ void CostmapLayer::clearArea(int start_x, int start_y, int end_x, int end_y, boo
   current_ = false;
   unsigned char * grid = getCharMap();
   for (int x = 0; x < static_cast<int>(getSizeInCellsX()); x++) {
-    bool xrange = x > start_x && x < end_x;
+    bool xrange = x >= start_x && x <= end_x;
 
     for (int y = 0; y < static_cast<int>(getSizeInCellsY()); y++) {
-      if ((xrange && y > start_y && y < end_y) != invert) {
-        continue;
-      }
-      int index = getIndex(x, y);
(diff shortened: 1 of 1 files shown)


nav2_costmap_2d/src/costmap_layer.cpp
```diff
@@ -66,15 +66,14 @@ void CostmapLayer::clearArea(int start_x, int start_y, int end_x, int end_y, boo
...
+    bool xrange = x >= start_x && x <= end_x;
 
     for (int y = 0; y < static_cast<int>(getSizeInCellsY()); y++) {
-      if ((xrange && y > start_y && y < end_y) != invert) {
-        continue;
-      }
-      int index = getIndex(x, y);
-      if (grid[index] != NO_INFORMATION) {
-        grid[index] = NO_INFORMATION;
+      if (xrange && y >= start_y && y <= end_y) {
```
SteveMacenski, line 72: Are you sure its not just because you didn't set `invert` to the correct value? That seems like the only major change here (beyond the `=`) and you removed it entirely

SteveMacenski: I don't think the `=`'s would also have that great of an impact, it should only really change the behavior at the edges I would think

jginesclavero: You're right. I am working with foxy-devel and the `invert` it is not implemented in there. Thank you for your reply! I close the PR

PR #2152, opened 2021-01-19
Regulated pure pursuit controller

SteveMacenski:
Introducing my new variation on the Pure Pursuit algorithm: Regulated Pure Pursuit. The tl;dr on it is
- Introduces new heuristics to adjust translational velocities proportional to dangerous situations like close proximity obstacles or high curvature path changes -- also drastically reduces overshoot and on average does a better job at path tracking
- Collision detection proportional to a forward projection time on the pure pursuit arc
- Configurable for Pure Pursuit, Adaptive Pure Pursuit, and Regulated Pure Pursuit
- Additional settings for slowing on approach to smooth out goal state transitions
- Visualizations of the carrot marker and collision arc
- Runs at > 1 kHz

An _extremely_ early partial draft of the paper can be seen here for general motivation and description, [here](https://github.com/ros-planning/navigation2/files/5839285/IROS__CASE__ICRA__ISRR_2021__Regulated_Pure_Pursuit.pdf).

Also added a new parser to the rewritten YAML functions so that you can remap full-path parameters when you need to change specific values that have multiple uses (like `plugin`)

TODO
- [x] rotate to goal if param on
- [x] unit and integration tests (>=85%) https://codecov.io/gh/ros-planning/navigation2/tree/fd3f93b61e774cc4ecfbdd26c414e6d1b3fce4c7/nav2_regulated_pure_pursuit_controller
- [x] ~interpolated distances (optional)~

Needs 
- [ ] testing on real robots (@fmrico) 
- [x] review (@shrijitsingh99)
- [x] Configuration guide, migration guide, video, plugin list (@SteveMacenski)  https://github.com/ros-planning/navigation.ros.org/pull/132
- [x] Update tutorial on PP (@SteveMacenski)  https://github.com/ros-planning/navigation.ros.org/pull/132
- [x] what to do with PP in the tutorials? Remove: tutorial updated with git hash of last simplified version   https://github.com/ros-planning/navigation2_tutorials/pull/22


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -0,0 +1,570 @@
...
+  }
+
+  // check current point and carrot point are OK, most likely ones to be in collision
+  if (inCollision(carrot_in_odom.pose.position.x, carrot_in_odom.pose.position.y) ||
+    inCollision(robot_pose.pose.position.x, robot_pose.pose.position.y))
+  {
+    return true;
+  }
+
+  // debug messages
```
AlexeyMerzlyakov, line 332: Looks like this is lookahead arc path made for debugging visualization. This might be a confusion about `RCLCPP_DEBUG()` which are also called "debug messages". I think, it is better to change this comment to something more descriptive.

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -0,0 +1,133 @@
...
+    const geometry_msgs::msg::PoseStamped &,
+    const geometry_msgs::msg::PoseStamped &,
+    const double &, const double &, const double &);
+
+  bool inCollision(const double & x, const double & y);
+
+  double costAtPose(const double & x, const double & y);
+
+  void applyConstraints(
+    double & linear_vel,
```
AlexeyMerzlyakov, line 90: Comparing to `transformPose()` where input function arguments comes first, and then output in the end, why here `linear_vel` is a fist, not last argument?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -0,0 +1,570 @@
...
+{
+  unsigned int mx, my;
+  costmap_->worldToMap(x, y, mx, my);
+
+  unsigned char cost = costmap_->getCost(mx, my);
+
+  if (costmap_ros_->getLayeredCostmap()->isTrackingUnknown()) {
+    return cost >= INSCRIBED_INFLATED_OBSTACLE && cost != NO_INFORMATION;
+  } else {
+    return cost >= INSCRIBED_INFLATED_OBSTACLE;
```
AlexeyMerzlyakov, line 395: Could here be a situation when local costmap's default value another than `NO_INFORMATION`, but after some costmap layers `NO_INFORMATION` still might appear on costmaps? E.g. there is no guarantee that some algorithms or costmap filters in the future might not emit such costmaps. In that case we will consider no-information as a collision. In my opinion, `cost >= INSCRIBED_INFLATED_OBSTACLE && cost != NO_INFORMATION` condition is sufficient for both cases. If-statement in this case is unnecessary.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -0,0 +1,570 @@
...
+  angle_to_path = atan2(carrot_pose.pose.position.y, carrot_pose.pose.position.x);
+  return use_rotate_to_heading_ && fabs(angle_to_path) > rotate_to_heading_min_angle_;
+}
+
+void RegulatedPurePursuitController::rotateToHeading(
+  double & linear_vel, double & angular_vel,
+  const double & angle_to_path, const geometry_msgs::msg::Twist & curr_speed)
+{
+  // Rotate in place using max angular velocity / acceleration possible
+  linear_vel = 0.0;
```
AlexeyMerzlyakov, line 279: Just a question. Linear velocity in one moment changes from some value (may be big one) to zero. Robot moment-ally stops. This might cause a robot drift. Do we consider in the model of Controller these possible linear drifts that will cause undesirable frame shifts?

nav2_regulated_pure_pursuit_controller/README.md
```diff
@@ -0,0 +1,139 @@
...
+Code based on a simplified version of this controller is referenced in the [Writing a New Nav2 Controller](https://navigation.ros.org/plugin_tutorials/docs/writing_new_nav2controller_plugin.html) tutorial.
+
+This plugin implements the `nav2_core::Controller` interface allowing it to be used acros the navigation stack as a local trajectory planner in the controller server's action server (`controller_server`).
+
+It builds on top of the ordinary pure pursuit algorithm in a number of ways. It also implements all the common variants of the pure pursuit algorithm such as adaptive pure pursuit. This controller is suitable for use on all types of robots, including differential, legged, and ackermann steering vehicles. It may also be used on omni-directional platforms, but won't be able to fully leverage the lateral movements of the base (you may consider DWB instead).
+
+This controller has been measured to run at well over 1 kHz on a modern intel processor.
+
+<p align="center">
+  <img src="https://user-images.githubusercontent.com/14944147/102563918-3cd49d80-408f-11eb-8e03-b472815a05d8.gif">
```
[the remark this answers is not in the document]
SteveMacenski: its a _big_ image so by having it hosted elsewhere, we dont own that every time someone clones the repo

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: I don't understand what you mean by robot drift

nav2_regulated_pure_pursuit_controller/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
+cmake_minimum_required(VERSION 3.5)
+project(nav2_regulated_pure_pursuit_controller)
```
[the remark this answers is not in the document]
SteveMacenski: I dont really want to make it unclear to people what it is `rpp` isn't a trade term so no one's really going to know what that is. but Pure Pursuit people know. It definitely borders on too long, but I think it stays right on that edge. I suppose I could drop `_controller` but given DWB's naming that makes it less consistent (and with the `_planner` suffix for plugin class types)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Sure, but I'm not sure what that has to do with this - unknown is a special class type that needs to be handled uniquely from any other cost value because of the fact that sometimes you track it and other times you don't. Its the general understanding in all of the plugins that if you track it then its no-go space because you're not confident about anything about it (like if there's a drop off for stairs). The default value of the costmap isn't necessarily related to this check. The default could be 100 for some reason, but we still need to handle no information as a special case.

SteveMacenski: Sorry, also rereading this - was there a specific action I can take here?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: This is also for rotating to heading if the heading is off by a ton, this isn't called in normal operations. Only on starting of a new path when there could be a huge diff needing correction

SteveMacenski:
At 56% in unit coverage, another 30% or so will be had from an actual integration test inclusion - so should be able to hit 85% by tomorrow afternoon

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
AlexeyMerzlyakov: Maybe the "drifting" term isn't correct. I mean robot slippage when it stops with braking distances trying to decrease speed value to 0 in one time. Something like here on this YouTube video: https://youtu.be/TdwuftIUwYY?t=131. Indeed, this is not a car robot (although the mobile robot can move quickly too), but the drifting/slipping principle I want to point is good shown here.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
AlexeyMerzlyakov: OK, thanks for clarification. That sounds reasonable. However, one point I still not confident about. You've mentioned that:
 > Its the general understanding in all of the plugins that if you track it then its no-go space

In other words, if we are tracking unknowns, the areas with unknown values will be treated as no-go, right? In this case we should return `true` if `cost == NO_INFROMATION` for `isTrackingUnknown()`. In the code, the situation is seems to be opposite: currently we are consider unknown places as no-go if `isTrackingUnknown()` is false.

> I could see where you think that.

I've just checked the code implementation of `isTrackingUnknown()`: it return `true` if costmap's default value is `NO_INFORMATION`, otherwise - `false`. Therefore, I've pointed my interest on this statement. How costmap's default value could affect a moving through unknown area decisions?

> Sorry, also rereading this - was there a specific action I can take here?

After reading a clarification that you've gave, if-statement here looks to have a reason. However, let's make it clear for the first question in this comment. When should we avoid unknowns: if it is tracked or if it is not tracked?

SteveMacenski:
93% test coverage!

SteveMacenski:
@AlexeyMerzlyakov I'm seeing a bunch of costmap filter test failures in debug

```
test_failure_navigator.missing_result
test_keepout_filter.missing_result
test_speed_filter_global.missing_result
test_speed_filter_local.missing_result

```

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: >  the areas with unknown values will be treated as no-go, right?

Not necessarily. Again, that's to do with if we track it or not. Different algorithms will do different things, that's why I check it

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Ah, well robots have kinematically feasible linear velocity transitions in this controller so that isn't an issue. Also its expected that robots make sure commands are valid before executing, its on all robots that I'm aware of.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: We could add a param whether unknown space is traversible or not, but it seems redundant since the costmap provides us that implicitly

AlexeyMerzlyakov: Agree with last argument. This does not make sense.

PR #2153, opened 2021-01-22
Feat/Dashing add map and frame topic

potentialdiffer:
Signed-off-by: Spinnler Christian <christian.spinnler@siemens.com>

---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | (Debian) |

---

In the dashing version of map_server it is not possible to use custom map and frame_id topics. I added support for this. 

New parameters are `topic_name` and `frame_id` which both default to "map".

---

### DIFF
--- a/nav2_map_server/src/map_server.cpp
+++ b/nav2_map_server/src/map_server.cpp
@@ -35,6 +35,8 @@ MapServer::MapServer()
 
   // Declare the node parameters
   declare_parameter("yaml_filename", rclcpp::ParameterValue(std::string("map.yaml")));
+  declare_parameter("topic_name", rclcpp::ParameterValue(std::string("map")));
+  declare_parameter("frame_id", rclcpp::ParameterValue(std::string("map")));
 }
 
 MapServer::~MapServer()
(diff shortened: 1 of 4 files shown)


nav2_map_server/src/map_server.cpp
```diff
@@ -35,6 +35,8 @@ MapServer::MapServer()
 
   // Declare the node parameters
   declare_parameter("yaml_filename", rclcpp::ParameterValue(std::string("map.yaml")));
+  declare_parameter("topic_name", rclcpp::ParameterValue(std::string("map")));
```
SteveMacenski, line 38: So I don't have an issue with this, but I'm not sure if this is ABI compatible adding a new param API even if the defaults are the same. Do you know for sure one way or another?

At least on the topic side, you can remap pretty easily. The frame part is indeed harder. Is this in `main` branch?

SteveMacenski (changes requested):

nav2_map_server/src/map_server.cpp
potentialdiffer: You have point. I cannot say for sure, I have changed a class constructor which could lead to issues though. 
The parameters are implemented almost the same way in `main`. https://github.com/ros-planning/navigation2/blob/61710875ed1d73e807b7cd1a9e577d995f2e4ead/nav2_map_server/src/map_server/map_server.cpp#L72

SteveMacenski: That is a good point - that class constructor shouldn't change. Normally I wouldn't really care because its an internal class, but I do know some folks rely on that object as is. 

I hate to reject a PR that's a super reasonable request, but I'm not sure I see a path forward for this.

potentialdiffer: That is fine. I did not consider the ABI  problem.

SteveMacenski:
Closing due to no way to do this in an ABI compatible way. Sorry :(

Issue #2040, opened 2020-10-13
Hybrid-A* / Smac planner: backtrace function runs into infinite loops


dkuenster:
## Bug report

**Required Info:**

- Operating System:
  - Debian 10
- ROS2 Version:
  - rolling
- Version or commit hash:
  - a13623922bc93ec8f12cdd40d2a817ea54998aaf (current nav2 main)
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue

Start tb3 simulation with ros2 launch nav2_bringup tb3_simulation_launch.py. Replace planner_server parameters in nav2_params with the ones suggested in 
https://github.com/ros-planning/navigation2/blob/d3fc28af28df764a153c85434f8b30270e963622/smac_planner/README.md#parameters
Set multiple waypoints and look at the memory usage of your system. If your used memory suddenly goes to 100% in a matter of about 15 seconds when reaching a waypoint look for the error message described below.

#### Expected behavior

Planner reaches one waypoint after the other and then plans to the next one.

#### Actual behavior

Sometimes when switching from one waypoint to the next, the planner gets stuck in an infinite loop during the backtrace function.
https://github.com/ros-planning/navigation2/blob/a13623922bc93ec8f12cdd40d2a817ea54998aaf/smac_planner/src/a_star.cpp#L290-L291. 
https://github.com/ros-planning/navigation2/blob/a13623922bc93ec8f12cdd40d2a817ea54998aaf/smac_planner/src/a_star.cpp#L448-L463
It keeps backtracing until it runs out of memory and then crashes with a std::bad_alloc.

#### Error output

[controller_server-8] [INFO] [1602599767.854334277] [controller_server]: Passing new path to controller.
[controller_server-8] [INFO] [1602599769.362043128] [controller_server]: Reached the goal!
[bt_navigator-11] [INFO] [1602599769.387884280] [bt_navigator]: Navigation succeeded
[waypoint_follower-12] [INFO] [1602599770.074093025] [waypoint_follower]: Succeeded processing waypoint 0, processing waypoint task execution
[waypoint_follower-12] [INFO] [1602599770.074309872] [waypoint_follower]: Task execution at waypoint 0 succeeded
[waypoint_follower-12] [INFO] [1602599770.074368300] [waypoint_follower]: Handled task execution on waypoint 0, moving to next.
[bt_navigator-11] [INFO] [1602599770.125854408] [bt_navigator]: Begin navigating from current location to (-1.74, -0.06)
[...]

PR #2159, opened 2021-01-28
smac_planner: fix goal checking in getAnalyticPath()

mkolodziejczyk-piap:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2040 |
| Robotic platform tested on | my own simulation stack |

---

* Added condition to exclude goal from poses added through analytic path expansion, because the goal is added always at the end and adding it twice causes infinite loops in the path

### DIFF
--- a/smac_planner/src/a_star.cpp
+++ b/smac_planner/src/a_star.cpp
@@ -411,7 +411,7 @@ AStarAlgorithm<NodeSE2>::NodePtr AStarAlgorithm<NodeSE2>::getAnalyticPath(
   prev = node;
   for (const auto & node_pose : possible_nodes) {
     const auto & n = node_pose.first;
-    if (!n->wasVisited()) {
+    if (!n->wasVisited() and n->getIndex()!=_goal->getIndex()) {
       // Make sure this node has not been visited by the regular algorithm.
       // If it has been, there is the (slight) chance that it is in the path we are expanding
       // from, so we should skip it.


smac_planner/src/a_star.cpp
```diff
@@ -411,7 +411,7 @@ AStarAlgorithm<NodeSE2>::NodePtr AStarAlgorithm<NodeSE2>::getAnalyticPath(
   prev = node;
   for (const auto & node_pose : possible_nodes) {
     const auto & n = node_pose.first;
-    if (!n->wasVisited()) {
+    if (!n->wasVisited() and n->getIndex()!=_goal->getIndex()) {
```
SteveMacenski, line 414:
```suggestion
    if (!n->wasVisited() && n->getIndex() != _goal->getIndex()) {
```

Just to be consistent in styling

SteveMacenski (changes requested):
Just a styling change but LGTM. Can you explain in a few sentences what the specific issue you hit? I'm curious how we didn't also run into this in our testing.

mkolodziejczyk-piap:
IMO the issue arises, when the vehicle is crossing the goal but the orientation is not the goal orientation and after crossing this goal ompl plans to correct the yaw with some back and forth manouvers. In my case, this loop happend almost all the time near the goal, probably because (1) my controller was not tuned enought to follow the path exactly (so the initial "smooth" global plan turned into some wild manouvers to correct the yaw) (2) my vehicle has large turning radius

SteveMacenski:
> vehicle is crossing the goal but the orientation is not the goal orientation

Ooooooh, yeah I did _not_ test for that. Thanks!

Please make the update / `main` PR and we can merge the set.

SteveMacenski:
@mkolodziejczyk-piap just following up here. Just need the small styling update & `main` PR to merge both

SteveMacenski:
@mkolodziejczyk-piap just following up

This is essentially done, just needs superficial update

SteveMacenski:
Closing - superseded by linked PRs

PR #2161, opened 2021-01-31
fix boundary point exclusion in convexFillCells

ChristofDubs:

equivalent of [this ros-navigation pull request](https://github.com/ros-planning/navigation/pull/1095):

> A bug in the function that returns all the cells contained in a (convex) polygon:
> 
> - sorts all the points along the polygon outline by x coordinate
> - for all points with same x coordinate, find min and max y coordinate
> - for that x coordinate, considers all points between min and max y coordinate to be within the polygon
> 
> The bug is that the maximum y coordinate also needs to be included.

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d.cpp
+++ b/nav2_costmap_2d/src/costmap_2d.cpp
@@ -467,7 +467,7 @@ void Costmap2D::convexFillCells(
 
     MapLocation pt;
     // loop though cells in the column
-    for (unsigned int y = min_pt.y; y < max_pt.y; ++y) {
+    for (unsigned int y = min_pt.y; y <= max_pt.y; ++y) {
       pt.x = x;
       pt.y = y;
       polygon_cells.push_back(pt);


SteveMacenski (commented):
Please explain the situation that this fixes for you. I read this and don't see where this would necessarily change the result. https://github.com/ros-planning/navigation2/blob/f44cc8ba45a419e0d3b0cfe75dce3806442a2614/nav2_costmap_2d/src/costmap_2d.cpp#L418 this gets the entire outline before we start filling so by the time you enter here https://github.com/ros-planning/navigation2/blob/f44cc8ba45a419e0d3b0cfe75dce3806442a2614/nav2_costmap_2d/src/costmap_2d.cpp#L469 you would already have the max elements.

So I think to merge this we need to know the specific situations that you ran into that this failed in which require this solution to resolve.

ChristofDubs:
>  I read this and don't see where this would necessarily change the result.

When this line
https://github.com/ros-planning/navigation2/blob/f44cc8ba45a419e0d3b0cfe75dce3806442a2614/nav2_costmap_2d/src/costmap_2d.cpp#L470
is reached, all points on the polygon boundary with x-coordinate `x` were evaluated, with `min_pt` and `max_pt` being set to the minimum and maximum y-coordinate of all boundary points with x-coordinate `x`.

What this pull request changes is that both boundary points `[x, min_pt]` and `[x, max_pt]` (and all cells in between) are considered being inside the polygon (whereas in the current main branch, `[x, min_pt]` is considered being inside the polygon, **but not** `[x, max_pt]`).

@SteveMacenski Does that clarify what this changes? And do you agree that the change is correct?

> Please explain the situation that this fixes for you.

As indicated in the initial description, I came across this bug while using the ROS1 navigation stack. The issues that it fixed there was that the robot would be stuck occasionally in case an obstacle point was marked on the footprint polygon boundary, because footprint clearing uses the `setConvexPolygonCost` method which relies on the `convexFillCells` method.

SteveMacenski (approved):
Got it - seems reasonable! Just making sure it fixes an actually observed issue and not a perceived issue

PR #2164, opened 2021-02-02
Update cost scaling heuristic to vary speed linearly with distance

shrijitsingh99:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2152  |
| Primary OS tested on | Ubuntu 18.04
| Robotic platform tested on | Turtlebot simulation

---

The new cost scaling heuristic will scale the speed linearly based on distance from a lethal obstacle.
Updated heuristic: 
CodeCogsEqn

Earlier velocity decreased exponentially since it was based on inflation cost which is computed using an exponential function. This called the velocity to drop significantly in highly inflated areas even though the robot could safely traverse at a higher speed.

This updated heuristic will allow the robot to move at a higher speed in inflated areas will ensuring it slows down if it comes near an obstacle.

**TODO:**

- [ ] Documentation needs to be updated to reflect the new parameters

Signed-off-by: Shrijit Singh <shrijitsingh99@gmail.com>

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/README.md
+++ b/nav2_regulated_pure_pursuit_controller/README.md
@@ -63,6 +63,9 @@ Mixing the proximity and curvature regulated linear velocities with the time-sca
 | `max_allowed_time_to_collision` | The time to project a velocity command to check for collisions | 
 | `use_regulated_linear_velocity_scaling` | Whether to use the regulated features for curvature | 
 | `use_cost_regulated_linear_velocity_scaling` | Whether to use the regulated features for proximity to obstacles | 
+| `cost_scaling_dist` | The minimum distance from an obstacle to trigger the scaling of linear velocity, if `use_cost_regulated_linear_velocity_scaling` is enabled. The value set should be smaller than the `inflation_radius` set in the inflation layer of costmap, since inflation is used to compute the distance from obstacles | 
+| `cost_scaling_gain` | A multiplier used to further scale the speed when within the specified `cost_scaling_dist` | 
+| `inflation_cost_scaling_factor` | The value of `cost_scaling_factor` set for the inflation layer in the local costmap. The value should be exactly the same for accurately computing distance from obstacles using the inflated cell values | 
 | `regulated_linear_scaling_min_radius` | The turning radius for which the regulation features are triggered. Remember, sharper turns have smaller radii | 
 | `regulated_linear_scaling_min_speed` | The minimum speed for which the regulated features can send, to ensure process is still achievable even in high cost spaces with high curvature. | 
 | `use_rotate_to_heading` | Whether to enable rotating to rough heading and goal orientation when using holonomic planners. Recommended on for all robot types except ackermann, which cannot rotate in place. | 
--- a/nav2_regulated_pure_pursuit_controller/README.md
+++ b/nav2_regulated_pure_pursuit_controller/README.md
@@ -63,8 +63,8 @@ Mixing the proximity and curvature regulated linear velocities with the time-sca
 | `max_allowed_time_to_collision` | The time to project a velocity command to check for collisions | 
 | `use_regulated_linear_velocity_scaling` | Whether to use the regulated features for curvature | 
 | `use_cost_regulated_linear_velocity_scaling` | Whether to use the regulated features for proximity to obstacles | 
-| `cost_scaling_dist` | The minimum distance from an obstacle to trigger the scaling of linear velocity, if `use_cost_regulated_linear_velocity_scaling` is enabled. The value set should be smaller than the `inflation_radius` set in the inflation layer of costmap, since inflation is used to compute the distance from obstacles | 
-| `cost_scaling_gain` | A multiplier used to further scale the speed when within the specified `cost_scaling_dist` | 
+| `cost_scaling_dist` | The minimum distance from an obstacle to trigger the scaling of linear velocity, if `use_cost_regulated_linear_velocity_scaling` is enabled. The value set should be smaller or equal to the `inflation_radius` set in the inflation layer of costmap, since inflation is used to compute the distance from obstacles | 
+| `cost_scaling_gain` | A multiplier gain, which should be <= 1.0, used to further scale the speed when an obstacle is within `cost_scaling_dist`. Lower value reduces speed more quickly. | 
 | `inflation_cost_scaling_factor` | The value of `cost_scaling_factor` set for the inflation layer in the local costmap. The value should be exactly the same for accurately computing distance from obstacles using the inflated cell values | 
 | `regulated_linear_scaling_min_radius` | The turning radius for which the regulation features are triggered. Remember, sharper turns have smaller radii | 
 | `regulated_linear_scaling_min_speed` | The minimum speed for which the regulated features can send, to ensure process is still achievable even in high cost spaces with high curvature. | 
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -154,8 +154,8 @@ void RegulatedPurePursuitController::configure(
   control_duration_ = 1.0 / control_frequency;
 
   if (inflation_cost_scaling_factor_ <= 0.0) {
-    RCLCPP_WARN(logger_, "The value inflation_cost_scaling_factor is incorrectly set, \
-    it should be >0. Disabling cost regulated linear velocity scaling.");
+    RCLCPP_WARN(logger_, "The value inflation_cost_scaling_factor is incorrectly set, "
+    "it should be >0. Disabling cost regulated linear velocity scaling.");
     use_cost_regulated_linear_velocity_scaling_ = false;
   }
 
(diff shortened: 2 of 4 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -133,6 +139,9 @@ void RegulatedPurePursuitController::configure(
   node->get_parameter(plugin_name_ + ".max_allowed_time_to_collision", max_allowed_time_to_collision_);
   node->get_parameter(plugin_name_ + ".use_regulated_linear_velocity_scaling", use_regulated_linear_velocity_scaling_);
   node->get_parameter(plugin_name_ + ".use_cost_regulated_linear_velocity_scaling", use_cost_regulated_linear_velocity_scaling_);
+  node->get_parameter(plugin_name_ + ".cost_scaling_dist", cost_scaling_dist_);
```
SteveMacenski, line 142: The usual things needed - add params to readme file of this package and on the docs website. These aren't immediately clear what they represent to someone unfamiliar (dist min to trigger, max to start, etc)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -144,6 +153,12 @@ void RegulatedPurePursuitController::configure(
   transform_tolerance_ = tf2::durationFromSec(transform_tolerance);
   control_duration_ = 1.0 / control_frequency;
 
+  if (inflation_cost_scaling_factor_ <= 0.0) {
+    RCLCPP_WARN(logger_, "The value inflation_cost_scaling_factor is incorrectly set, \
```
SteveMacenski, line 157: Remove the \ and just split into 2 quotes, e.g. 

```
RCLCPP_WARN(logger_, "The value inflation_cost_scaling_factor is incorrectly set" 
  "it should be >0. Disabling cost regulated linear velocity scaling.");
```

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -436,18 +451,15 @@ void RegulatedPurePursuitController::applyConstraints(
...
-    double max_non_collision = 128.0;
-    if (pose_cost > max_non_collision) {
-      cost_vel = regulated_linear_scaling_min_speed_;
-    } else {
-      const double slope = (regulated_linear_scaling_min_speed_ - cost_vel) / max_non_collision;
-      cost_vel = slope * pose_cost + cost_vel;
-    }
+    auto inscribed_radius = costmap_ros_->getLayeredCostmap()->getInscribedRadius();
+
+    if (pose_cost != FREE_SPACE) {
```
SteveMacenski, line 456: This `if` can be placed on L452 so that we check if pose_cst is not NO_INFORMATION and also not FREE_SPACE (and freespace needs to be casted to double)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -436,18 +451,15 @@ void RegulatedPurePursuitController::applyConstraints(
...
-    // in when in questionable states. This creates a linear mapping of
-    // cost [0, 128] to speed [min regulated speed, linear vel]
-    double max_non_collision = 128.0;
-    if (pose_cost > max_non_collision) {
-      cost_vel = regulated_linear_scaling_min_speed_;
-    } else {
-      const double slope = (regulated_linear_scaling_min_speed_ - cost_vel) / max_non_collision;
-      cost_vel = slope * pose_cost + cost_vel;
-    }
+    auto inscribed_radius = costmap_ros_->getLayeredCostmap()->getInscribedRadius();
```
SteveMacenski, line 454:
```suggestion
    const double inscribed_radius = costmap_ros_->getLayeredCostmap()->getInscribedRadius();
```

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -436,18 +451,15 @@ void RegulatedPurePursuitController::applyConstraints(
...
+
+    if (pose_cost != FREE_SPACE) {
+        const double lethal_dist = (-1.0 / inflation_cost_scaling_factor_) 
+        * std::log(pose_cost / (INSCRIBED_INFLATED_OBSTACLE - 1)) + inscribed_radius;
+        if (lethal_dist < cost_scaling_dist_){
```
SteveMacenski, line 459:
```suggestion
        if (lethal_dist < cost_scaling_dist_) {
```

just a linting thing

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -436,18 +451,15 @@ void RegulatedPurePursuitController::applyConstraints(
...
-    if (pose_cost > max_non_collision) {
-      cost_vel = regulated_linear_scaling_min_speed_;
-    } else {
-      const double slope = (regulated_linear_scaling_min_speed_ - cost_vel) / max_non_collision;
-      cost_vel = slope * pose_cost + cost_vel;
-    }
+    auto inscribed_radius = costmap_ros_->getLayeredCostmap()->getInscribedRadius();
+
+    if (pose_cost != FREE_SPACE) {
+        const double lethal_dist = (-1.0 / inflation_cost_scaling_factor_)
```
SteveMacenski, line 457: This tripped me up for a minute, lets rename this to `min_distance_to_obstacle`, since this isn't necessarily `lethal`. This is the distance backed out from the inflation layer based on the cell cost of the robot.

SteveMacenski (changes requested):

SteveMacenski:
This seems to have crashed the constraints tests:

```
    auto inscribed_radius = costmap_ros_->getLayeredCostmap()->getInscribedRadius();
```

In these tests `getLayeredCostmap` is a nullptr, so I think the best short term answer is to comment out the tests in L300-L328 for now

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
shrijitsingh99: Added the parameters to README. If no changes are needed in the description then will create a PR for docs.

nav2_regulated_pure_pursuit_controller/README.md
```diff
@@ -63,6 +63,9 @@ Mixing the proximity and curvature regulated linear velocities with the time-sca
 | `max_allowed_time_to_collision` | The time to project a velocity command to check for collisions | 
 | `use_regulated_linear_velocity_scaling` | Whether to use the regulated features for curvature | 
 | `use_cost_regulated_linear_velocity_scaling` | Whether to use the regulated features for proximity to obstacles | 
+| `cost_scaling_dist` | The minimum distance from an obstacle to trigger the scaling of linear velocity, if `use_cost_regulated_linear_velocity_scaling` is enabled. The value set should be smaller than the `inflation_radius` set in the inflation layer of costmap, since inflation is used to compute the distance from obstacles | 
+| `cost_scaling_gain` | A multiplier used to further scale the speed when within the specified `cost_scaling_dist` |
```
SteveMacenski, line 67:
```suggestion
| `cost_scaling_gain` | A multiplier gain, which should be > 1.0, used to further scale the speed when an obstacle is within `cost_scaling_dist`. Higher value reduces speed more quickly. | 
```

nav2_regulated_pure_pursuit_controller/README.md
```diff
@@ -63,6 +63,9 @@ Mixing the proximity and curvature regulated linear velocities with the time-sca
 | `max_allowed_time_to_collision` | The time to project a velocity command to check for collisions | 
 | `use_regulated_linear_velocity_scaling` | Whether to use the regulated features for curvature | 
 | `use_cost_regulated_linear_velocity_scaling` | Whether to use the regulated features for proximity to obstacles | 
+| `cost_scaling_dist` | The minimum distance from an obstacle to trigger the scaling of linear velocity, if `use_cost_regulated_linear_velocity_scaling` is enabled. The value set should be smaller than the `inflation_radius` set in the inflation layer of costmap, since inflation is used to compute the distance from obstacles |
```
SteveMacenski, line 66:
```suggestion
| `cost_scaling_dist` | The distance from an obstacle to trigger the scaling of linear velocity, if `use_cost_regulated_linear_velocity_scaling` is enabled. The value set should be smaller or equal to the `inflation_radius` set in the inflation layer of costmap, since inflation is used to compute the distance from obstacles | 
```

SteveMacenski (commented):
Just a couple of quick wording updates to make more clear - please reflect these in a docs PR but this looks good to me

SteveMacenski:
@shrijitsingh99 you forgot to comment out the second half of the `applyConstraints` test so its still segfaulting. I think the best short term answer is to comment out the tests in L300-L328 for now. Once CI is failing on the usual and not on this, I'll merge

nav2_regulated_pure_pursuit_controller/README.md
shrijitsingh99: Currently gain is being multiplied with the velocity so it should be <1.
I realise gain means amplification so should I just multiply the velocity with the inverse of gain or we could use some other term in place of gainz

SteveMacenski: Oh, sorry - ok, update to say that then :-)

SteveMacenski: dont worry about changing the formulation, either way you need to tell people > 1 or < 1, so lets just go with what we got

SteveMacenski (approved):
Will merge when CI passes (or only fails on things not related)

A docs PR would be good now too on these updates

PR #2170, opened 2021-02-04
attempt at bringing up CI (2158)

SteveMacenski:
Performing context updates on parameters  probably (?) causing `Original error: parameter 'GridBased.use_astar' has invalid type: expected [bool] got [string]`. 

Updating warning that occasionally crashes test on rclpy QoS updates

Adding some context in service publisher timeouts

Future:
- Switch to cyclone in CI?
- debug gazebo crashes
- rcl lifecycle / rcl rmw crash debugging with eprosima
- gazebo gzserver rather than the .so file call

### DIFF
--- a/nav2_bringup/bringup/worlds/waffle.model
+++ b/nav2_bringup/bringup/worlds/waffle.model
@@ -137,7 +137,7 @@
             </linear_acceleration>
           </imu>
           <plugin name="turtlebot3_imu" filename="libgazebo_ros_imu_sensor.so">
-            <initial_orientation_as_reference>true</initial_orientation_as_reference>
+            <initial_orientation_as_reference>false</initial_orientation_as_reference>
             <ros>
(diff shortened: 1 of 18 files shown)


SteveMacenski:
In the last run only `test_bt_navigator_with_groot_monitoring.missing_result - nav2_system_tests` was the issue in both runs after setting initial_orientation_as_reference

SteveMacenski:
Interestingly initial_orientation_as_reference makes it work better-ish, but extremely flaky. Some jobs might only fail on 1 test now and others all 6 of the integration tests. 

Interesting, localization & recovery tests never fail, also making use of gazebo, but in markedly simpler situations.  Maybe the jobs are overtaxing CPU / networking?

SteveMacenski:
Makes incremental improvement - more to do but merging in to get marginal benefits

PR #2172, opened 2021-02-06
CI resolution Part 3

SteveMacenski:
#2158 

Trying to see if changing the default RMW will fix the issue. I'm seeing failures in the action server now with goals failing which makes me think now this is just a performance regression that is overloading our CI container limits


SteveMacenski:
Runs:
- release pass, debug 1 recovery job failed

SteveMacenski:
Still not perfect, but now back up to a "flaky but functionalish" status

Issue #2178, opened 2021-02-11
Ability to get a path from the planner from any 2 poses (start_pose and goal) 


doisyg:
## Feature request
Add `use_start_pose` and `start_pose` to `ComputePathToPose`
#### Feature description
The current `ComputePathToPose` action exposed by the planner interface uses the current robot pose as a start pose to compute a path. It would be useful to be able to override the current robot pose with a provided `start_pose` so one could get a path from any 2 poses of the costmap.

#### Implementation considerations
Current behavior should be maintained as default one. Quite straightforward since the `global_planer` interface already implements:
```
nav_msgs::msg::Path createPlan(
    const geometry_msgs::msg::PoseStamped & start,
    const geometry_msgs::msg::PoseStamped & goal)
```
 Simple implementation in this draft PR : https://github.com/ros-planning/navigation2/pull/2179

SteveMacenski:
merged

PR #2179, opened 2021-02-11
Ability to get a path from the planner from any 2 poses (start_pose and goal): add use_start_pose and start_pose to ComputePathToPose action

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2178 |

---

* Proposal implementation of the following feature request https://github.com/ros-planning/navigation2/issues/2178. This is a draft. Will test properly if we agree on the relevance and the implementation.

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
@@ -42,7 +42,8 @@ public:
       {
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathToPose node"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
-        BT::InputPort<geometry_msgs::msg::PoseStamped>("start_pose", "Start pose of the path if overriding current robot pose"),
+        BT::InputPort<geometry_msgs::msg::PoseStamped>(
+            "start", "Start pose of the path if overriding current robot pose"),
         BT::InputPort<std::string>("planner_id", ""),
       });
   }
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
@@ -43,7 +43,7 @@ public:
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathToPose node"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
-            "start", "Start pose of the path if overriding current robot pose"),
+          "start", "Start pose of the path if overriding current robot pose"),
         BT::InputPort<std::string>("planner_id", ""),
       });
   }
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
@@ -42,6 +42,7 @@ public:
       {
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathToPose node"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
+        BT::InputPort<geometry_msgs::msg::PoseStamped>("start_pose", "Start pose of the path if overriding current robot pose"),
         BT::InputPort<std::string>("planner_id", ""),
       });
   }
--- a/nav2_behavior_tree/nav2_tree_nodes.xml
+++ b/nav2_behavior_tree/nav2_tree_nodes.xml
@@ -18,7 +18,7 @@
 
     <Action ID="ComputePathToPose">
       <input_port name="goal">Destination to plan to</input_port>
-      <input_port name="start_pose">Start pose of the path if overriding current robot pose</input_port>
+      <input_port name="start">Start pose of the path if overriding current robot pose</input_port>
       <output_port name="path">Path created by ComputePathToPose node</output_port>
       <input_port name="planner_id"/>
     </Action>
--- a/nav2_behavior_tree/nav2_tree_nodes.xml
+++ b/nav2_behavior_tree/nav2_tree_nodes.xml
@@ -18,6 +18,7 @@
 
     <Action ID="ComputePathToPose">
       <input_port name="goal">Destination to plan to</input_port>
+      <input_port name="start_pose">Start pose of the path if overriding current robot pose</input_port>
       <output_port name="path">Path created by ComputePathToPose node</output_port>
       <input_port name="planner_id"/>
     </Action>
--- a/nav2_behavior_tree/test/plugins/action/test_compute_path_to_pose_action.cpp
+++ b/nav2_behavior_tree/test/plugins/action/test_compute_path_to_pose_action.cpp
@@ -42,7 +42,7 @@ protected:
     const auto goal = goal_handle->get_goal();
     auto result = std::make_shared<nav2_msgs::action::ComputePathToPose::Result>();
     result->path.poses.resize(1);
-    result->path.poses[0].pose.position.x = goal->pose.pose.position.x;
+    result->path.poses[0].pose.position.x = goal->goal.pose.position.x;
     goal_handle->succeed(result);
   }
 };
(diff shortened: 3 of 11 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
```diff
@@ -42,6 +42,7 @@ class ComputePathToPoseAction : public BtActionNode<nav2_msgs::action::ComputePa
       {
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathToPose node"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
+        BT::InputPort<geometry_msgs::msg::PoseStamped>("start_pose", "Start pose of the path if overriding current robot pose"),
```
SteveMacenski, line 45: Lets just do `start` since `goal` doesn't have `_pose`

nav2_msgs/action/ComputePathToPose.action
```diff
@@ -1,6 +1,8 @@
 #goal definition
 geometry_msgs/PoseStamped pose
```
SteveMacenski, line 2: I think `pose` needs to change too now that there's a start. This should indicate goal

nav2_planner/src/planner_server.cpp
```diff
@@ -232,10 +232,15 @@ PlannerServer::computePlan()
...
     }
 
+    // Use start_pose if provided otherwise use current robot pose
     geometry_msgs::msg::PoseStamped start;
-    if (!costmap_ros_->getRobotPose(start)) {
-      action_server_->terminate_current();
-      return;
+    if (goal->use_start_pose) {
+      start = goal->start_pose;
+    } else {
```
SteveMacenski, line 239: This can be an `else if`

SteveMacenski (changes requested):
LGTM conceptually, just some cleanup to make it consistent

SteveMacenski:
Also some testing to show that it doesn't break current behaviors.

nav2_msgs/action/ComputePathToPose.action
doisyg: I agree with the need for consistency, but this change will break application using the action in its current state, are you sure ?

doisyg:
> Also some testing to show that it doesn't break current behaviors. A unit test for changing the start and showing that it works would be good too.

Sure, any tests already targeting the planner that I should take inspiration from ? Should I create a test/unit folder under the nav2_planner package ?

nav2_msgs/action/ComputePathToPose.action
SteveMacenski: Yes, this is `main`, we don't have to maintain any API compatibility here. We need to just document this in our migration guide that this was changed and the start / `use_start` was added.

Also in this action file, please add a comment about `use_start`. e.g. if false, will use current robot pose. If true, will use this start pose instead. 

(and rename `use_start_pose` as `use_start`).

nav2_planner/src/planner_server.cpp
```diff
@@ -232,8 +232,11 @@ PlannerServer::computePlan()
       r.sleep();
     }
 
+    // Use start_pose if provided otherwise use current robot pose
     geometry_msgs::msg::PoseStamped start;
-    if (!costmap_ros_->getRobotPose(start)) {
+    if (goal->use_start_pose) {
+      start = goal->start;
```
SteveMacenski, line 238: https://github.com/ros-planning/navigation2/issues/2186, it looks like we overlooked the stamped portion of the message for the goal (and now start). We should also check the `frame_id` of `start` and transform / print error if different from the global frame of the server's costmap used to plan in.

SteveMacenski: The `costmap_ros_->getRobotPose(start)` does this implicitly

SteveMacenski (changes requested):
Just working on details. A small unit test making use of this would be good.

SteveMacenski:
@doisyg any update? I think this just needs a superficial facelift and ready to merge

doisyg:
wip, missing:
- `nav2_system_tests/planner` test
- transform / print error if different from the global frame of the server's costmap (but do we want that in this PR or a separated one ?)
- migration doc update PR

Will complete in the next few days when I ll have a minute

nav2_behavior_tree/nav2_tree_nodes.xml
```diff
@@ -18,6 +18,7 @@
 
     <Action ID="ComputePathToPose">
       <input_port name="goal">Destination to plan to</input_port>
+      <input_port name="start_pose">Start pose of the path if overriding current robot pose</input_port>
```
SteveMacenski, line 21:
```suggestion
      <input_port name="start">Start pose of the path if overriding current robot pose</input_port>
```

SteveMacenski (commented):
Just missed one :-)

SteveMacenski:
> do we want that in this PR or a separated one 

A separate one would be OK with me, but I just don't want to lose track of it / have it be forgotten

doisyg:
> > do we want that in this PR or a separated one
> 
> A separate one would be OK with me, but I just don't want to lose track of it / have it be forgotten

Ok, I did it in the last commit (I can always split it in a different PR if we need dedicated implementation discussion).
I added a `getPoseInGlobalFrame` fonction in costmap_2d_ros since we already have a tf_buffer there, and this fonction calls an added `getPoseInTargetFrame` in nav2_util. I tried to reproduce the logic of `Costmap2DROS::getRobotPose` / `nav2_util::getCurrentPose`

doisyg:
> Then on the planner server side, you could make a test to call it on the `nav2_system_tests/planner` tests to change start and check that the first pose is within tol of requested start vs the robot's set pose.

On second thought, it would require creating a test that calls the ComputePathToPose action since the planner_tester calls directly the `createPlan(start, goal)` fonction
https://github.com/ros-planning/navigation2/blob/740afd5a7ab004f5922e7822db41fa1c61e155ed/nav2_system_tests/src/planning/planner_tester.hpp#L81-L86

nav2_behavior_tree/test/plugins/action/test_compute_path_to_pose_action.cpp
```diff
@@ -41,8 +41,13 @@ class ComputePathToPoseActionServer : public TestActionServer<nav2_msgs::action:
   {
     const auto goal = goal_handle->get_goal();
     auto result = std::make_shared<nav2_msgs::action::ComputePathToPose::Result>();
-    result->path.poses.resize(1);
-    result->path.poses[0].pose.position.x = goal->pose.pose.position.x;
+    result->path.poses.resize(2);
```
SteveMacenski, line 44: This is not a test that this works in the planner, just that the behavior tree responds properly. You must test using the actual planner server, please update the system tests for planning. This can also be helpful, but this isn't a test that these changes are actually functional since this just mocks up the planner server interface

doisyg: > You can do this using the `system_tests/planner` tests. There's a random and fixed map tests you could add an additional test case to where you set `use_start` and a `start` to show that it respects that and not the current robot TF starting pose.

It is more complex than that, as I tried to express above, the current `system_tests/planner` tests are never testing the `ComputePathToPose` `action_server` of the `planner_server` defined here:
https://github.com/ros-planning/navigation2/blob/2d88d14b1dd9f678e9b95dd50c5db774667cd62e/nav2_planner/src/planner_server.cpp#L132-L136

And hence, the logic of getting the pose of the robot as a start of the path (or the `use_start` potential override) as implemented in the action server of planner_server is never tested.
Instead, it is the only the `createPlan(start, goal)` fonction of the `planner_server` which is tested, not the full action. And the `planner_tester` implements its own logic for retrieving the start pose of the path, see here:
https://github.com/ros-planning/navigation2/blob/2d88d14b1dd9f678e9b95dd50c5db774667cd62e/nav2_system_tests/src/planning/planner_tester.hpp#L77-L86

What we need, I believe, is to create tests calling the `ComputePathToPose` `action_server` of the `planner_server`, for both `use_start` `true` and `false`. I might consider doing it if align and if you give me pointers on the architecture of it. Should `planner_tester` be extended to call the action ? Or a separate class ?

SteveMacenski: Got it, thanks for the explanation. I agree those tests are not probably in the structure for this testing.

The system tests [have this structure](https://github.com/ros-planning/navigation2/blob/main/nav2_system_tests/src/system/test_system_launch.py#L108-L118) of launching a launch description (which could just contain a lifecycle manager to transition up the planner server and the planner server) and then a test python node to trigger inputs and outputs. That could work here but the issue I see is that no info would actually come to the planner server, so you'd need to also have a `map_server` to publish some map for the planner server's costmap to be aware of. Then when you called `ComputePathToPose` you should be good to go (but probably have to fake out TF).

It turns into a bit of a challenge. I could wave the testing of this one in CI if you verified locally that this worked fully. Though it might be a good exercise if you haven't done something like that before and I would never argue with more :smile:

doisyg: Hum, it will be indeed a good exercise. No promise as I am very busy atm, but don't hesitate to ping me in a week or so if you see no progress (and yes, it was tested locally).

SteveMacenski (approved):
@doisyg just need some updates to the documentation 
- https://navigation.ros.org/migration/Foxy.html migration guide that this is now available

SteveMacenski:
```
Diff with 5 lines
--- include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
+++ include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp.uncrustify
@@ -46 +46 @@
-            "start", "Start pose of the path if overriding current robot pose"),
+          "start", "Start pose of the path if overriding current robot pose"),

```

And fix the linting error

SteveMacenski:
Awesome, thanks for iterating!

SteveMacenski:
@doisyg just wanted to give you a shout out, this addition has made my life so much easier while I'm benchmarking path planners! Great addition!

PR #2181, opened 2021-02-11
[NavFn] Make the 3 parameters changeable at runtime

doisyg:
## Description of contribution

* Get parameters  `tolerance`, `use_astar` and `allow_unknown` before calling `makePlan`
* Had to add `rclcpp_lifecycle::LifecycleNode::WeakPtr node_` as a class member and use `auto node = node_.lock()` in order to access `get_parameter` outside the `configure` function

---

## Future work that may be required

* Conforming to a dynamic parameter refresh pattern once it will be decided what the best practice. It seems parameters feature are still a wip, for instance https://github.com/ros2/rclcpp/pull/829


nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -123,6 +124,12 @@ nav_msgs::msg::Path NavfnPlanner::createPlan(
   steady_clock::time_point a = steady_clock::now();
 #endif
 
+  // Refresh parameter
```
SteveMacenski, line 127: This isn't a great way to handle this - instead we can use the parameter events callbacks to change variables so its event based and not just polling (see DWB which does this already)

SteveMacenski (changes requested):

nav2_navfn_planner/include/nav2_navfn_planner/navfn_planner.hpp
```diff
@@ -140,6 +140,11 @@ class NavfnPlanner : public nav2_core::GlobalPlanner
 
   // Whether to use the astar planner or default dijkstras
   bool use_astar_;
+
+  // Subscription for parameter change
+  rclcpp::AsyncParametersClient::SharedPtr parameters_client_;
+  rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr parameter_event_sub_;
+  void on_parameter_event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
```
SteveMacenski, line 147: Please add doxygen for this, if you pull in `main`, `navfn` has now complete doxgyen coverage

SteveMacenski (changes requested):
LGTM, just documentation. Also please add a really short (1-2 sentences) to the migration guide that navfn now has runtime reconfigurable parameters https://navigation.ros.org/migration/Foxy.html

SteveMacenski:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/4663/workflows/d74eaa7f-1620-4fc4-b9e5-b3d2033f41d0/jobs/17371

Also linting issue in CI

SteveMacenski (approved):

Issue #2189, opened 2021-02-17
Backup recovery action collision check looks too far ahead


acyen:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04.2
- ROS2 Version:
  - Foxy binaries,
- Version or commit hash:
  - navigation2 foxy-devel 77c1b742e959dda02acf6830c2a814f2278bb4a4
- DDS implementation:
  - Fast-RTPS
 
#### Steps to reproduce issue
In order for the backup recovery action to be used, make the fix in  `nav2_behavior_tree/plugins/action/back_up_action.cpp`:
```
@@ -75,7 +75,7 @@ BT_REGISTER_NODES(factory) 
     [](const std::string & name, const BT::NodeConfiguration & config)
     {
       return std::make_unique<nav2_behavior_tree::BackUpAction>(
-        name, "back_up", config);
+        name, "backup", config);
     };
```
Run the navigation2 tb3 demo using the attached parameter and behavior tree files. This replaces the Spin recovery action with BackUp and changes the local planner to the TEB planner.  Other param changes make it more likely that the local controller path will hit an obstacle.

[nav2_params.yaml.txt](https://github.com/ros-planning/navigation2/files/5994382/nav2_params.yaml.txt)
[navigate_w_replanning_and_recovery.xml.txt](https://github.com/ros-planning/navigation2/files/5994383/navigate_w_replanning_and_recovery.xml.txt)

Run
`ros2 launch nav2_bringup tb3_simulation_launch.py`
Using Rviz, place the starting position and the goal position on opposite sides of an obstacle.

#### Expected behavior
The robot should stop when it detects a collision and back up.

#### Actual behavior
No backup even if there is sufficient space behind the robot, message:
```
[recoveries_server-9] [INFO] [1613530319.113067431] [recoveries_server]: Attempting backup
[recoveries_server-9] [WARN] [1613530319.113508218] [recoveries_server]: Collision Ahead - Exiting BackUp
```
is printed in the log.
#### Additional information
The code for `nav2_recoveries/plugins/back_up.cpp` contains the following fragment in the function `BackUp::isCollisionFree`:
```
  while (cycle_count < max_cycle_count) {
    sim_position_change = cmd_vel->linear.x * (cycle_count / cycle_frequency_);
    pose2d.x += sim_position_change * cos(pose2d.theta);
    pose2d.y += sim_position_change * sin(pose2d.theta);
    cycle_count++;

[...]

PR #2193, opened 2021-02-18
Fix recovery action collision check.

acyen:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2189 |
| Robotic platform tested on | Turtlebot simulation |

---

Fix collision check in `back_up.cpp` and also `spin.cpp`.

None

### DIFF
--- a/nav2_recoveries/plugins/back_up.cpp
+++ b/nav2_recoveries/plugins/back_up.cpp
@@ -120,11 +120,12 @@ bool BackUp::isCollisionFree(
   double sim_position_change;
   const double diff_dist = abs(command_x_) - distance;
   const int max_cycle_count = static_cast<int>(cycle_frequency_ * simulate_ahead_time_);
-
+  geometry_msgs::msg::Pose2D init_pose = pose2d;
+  
   while (cycle_count < max_cycle_count) {
     sim_position_change = cmd_vel->linear.x * (cycle_count / cycle_frequency_);
-    pose2d.x += sim_position_change * cos(pose2d.theta);
-    pose2d.y += sim_position_change * sin(pose2d.theta);
+    pose2d.x = init_pose.x + sim_position_change * cos(init_pose.theta);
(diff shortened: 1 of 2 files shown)


nav2_recoveries/plugins/back_up.cpp
```diff
@@ -120,11 +120,12 @@ bool BackUp::isCollisionFree(
   double sim_position_change;
   const double diff_dist = abs(command_x_) - distance;
   const int max_cycle_count = static_cast<int>(cycle_frequency_ * simulate_ahead_time_);
-
+  geometry_msgs::msg::Pose2D init_pose = pose2d;
```
SteveMacenski, line 123: Did you test this to fix your issue with too long looking back?

SteveMacenski:
@acyen I'm happy to also merge this into `Foxy`, but please also direct a PR to `main` so that this change is reflected into all newer distributions for now on and not just in `Foxy`.

nav2_recoveries/plugins/back_up.cpp
```diff
@@ -120,11 +120,12 @@ bool BackUp::isCollisionFree(
...
+
```
SteveMacenski, line 124: 12:48:51 3: /tmp/ws/src/navigation2/nav2_recoveries/plugins/back_up.cpp:124:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]

Linting issue as well

SteveMacenski (changes requested):

nav2_recoveries/plugins/back_up.cpp
acyen: Yes, manual test using tb3 simulation.

SteveMacenski (approved):
@acyen please submit a PR with this change to `main` as well so that its reflected in all new distributions as well, not just Foxy

Issue #2154, opened 2021-01-22
Improve NavigateToPose Action Feedback


SteveMacenski:
CC @timple 

Distance to goal integrating path points rather than just L2 distance. 

ETA based on some conservative metrics (max speed or min speed, current speed?)

deepaktalwardt:
I will be taking a crack at this issue over the next few days :)

deepaktalwardt:
PR for this feature is up here https://github.com/ros-planning/navigation2/pull/2194

SteveMacenski:
Merging imminent after CI passes

PR #2194, opened 2021-02-19
Improve NavigationToPose action feedback

deepaktalwardt:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |#2154 |
| Primary OS tested on |Ubuntu 20.04 |

---

* Improves feedback of the `NavigateToPose` action server by adding:
    - Better distance to goal estimate: by integrating distance between planned path poses.
    - Estimated time remaining: by dividing the distance remaining by current robot linear speed (using smoothed odometry).
* Updates the feedback in `.action` file for `NavigateToPose` action server. 
* Adds a utility to the `nav2_util::geometry_utils` to calculate path length.
* Adds simple unit testing for this function.

* Documentation for `.action` needs to be updated.

---

* Currently the time remaining largely ignores the target yaw of the robot. It may be the case that the robot needs more time to rotate to the correct heading.

### DIFF
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -34,7 +34,7 @@ BtNavigator::BtNavigator()
 : nav2_util::LifecycleNode("bt_navigator", "", false),
   start_time_(0),
   current_linear_speed_(0.0),
-  estimated_navigation_time_remaining_(rclcpp::Duration::from_seconds(0.0))
+  estimated_time_remaining_(rclcpp::Duration::from_seconds(0.0))
 {
   RCLCPP_INFO(get_logger(), "Creating");
 
@@ -289,14 +289,14 @@ BtNavigator::navigateToPose()
         nav2_util::geometry_utils::calculate_path_length(current_path);
 
       if (distance_remaining <= transform_tolerance_) {
-        estimated_navigation_time_remaining_ = rclcpp::Duration::from_seconds(0.0);
+        estimated_time_remaining_ = rclcpp::Duration::from_seconds(0.0);
       } else {
         // Get current speed
         current_linear_speed_ = odom_smoother_->getTwist().linear.x;
 
-        // Calculate estimated time taken to goal if speed is higher than 1mm/s
-        if (std::abs(current_linear_speed_) > 0.05) {
-          estimated_navigation_time_remaining_ =
+        // Calculate estimated time taken to goal if speed is higher than 5mm/s
+        if (std::abs(current_linear_speed_) > 0.005) {
+          estimated_time_remaining_ =
             rclcpp::Duration::from_seconds(distance_remaining / std::abs(current_linear_speed_));
         }
       }
@@ -304,7 +304,7 @@ BtNavigator::navigateToPose()
       int recovery_count = 0;
       blackboard_->get<int>("number_recoveries", recovery_count);
       feedback_msg->distance_remaining = distance_remaining;
-      feedback_msg->estimated_navigation_time_remaining = estimated_navigation_time_remaining_;
+      feedback_msg->estimated_time_remaining = estimated_time_remaining_;
       feedback_msg->number_of_recoveries = recovery_count;
       feedback_msg->navigation_time = now() - start_time_;
       action_server_->publish_feedback(feedback_msg);
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -21,6 +21,7 @@
 #include <utility>
 #include <vector>
 #include <set>
+#include <cmath>
 
 #include "nav2_util/geometry_utils.hpp"
 #include "nav2_util/robot_utils.hpp"
@@ -32,7 +33,9 @@ namespace nav2_bt_navigator
 
 BtNavigator::BtNavigator()
 : nav2_util::LifecycleNode("bt_navigator", "", false),
-  start_time_(0)
+  start_time_(0),
+  current_linear_speed_(0.0),
+  estimated_navigation_time_remaining_(rclcpp::Duration::from_seconds(0.0))
 {
   RCLCPP_INFO(get_logger(), "Creating");
 
@@ -102,6 +105,9 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
     "goal_pose",
     rclcpp::SystemDefaultsQoS(),
     std::bind(&BtNavigator::onGoalPoseReceived, this, std::placeholders::_1));
+  
+  // Odometry smoother object for getting current speed
+  odom_smoother_ = std::make_unique<nav2_util::OdomSmoother>(client_node_, 0.3);
 
   action_server_ = std::make_unique<ActionServer>(
     get_node_base_interface(),
@@ -280,20 +286,35 @@ BtNavigator::navigateToPose()
       blackboard_->get("path", current_path);
 
       // Calculate distance on the path
+      double distance_remaining = nav2_util::geometry_utils::calculate_distance_to_goal(current_path);
 
-      // Get current speed
+      if (distance_remaining <= transform_tolerance_)
+      {
+        estimated_navigation_time_remaining_ = rclcpp::Duration::from_seconds(0.0);
+      } 
+      else
+      {
+        // Get current speed
+        current_linear_speed_ = odom_smoother_->getTwist().linear.x;
 
-      // Calculate estimated time taken to goal
+        // Calculate estimated time taken to goal if speed is higher than 1mm/s
+        if (std::abs(current_linear_speed_) > 0.05)
+        {
+          estimated_navigation_time_remaining_ =
+            rclcpp::Duration::from_seconds(distance_remaining / std::abs(current_linear_speed_));
+        }
+      }
 
-      // TODO (Deepak): Update this to add path distance computation
-      feedback_msg->distance_remaining = nav2_util::geometry_utils::euclidean_distance(
-        feedback_msg->current_pose.pose, goal_pose.pose);
-
-      // feedback_msg->estimated_navigation_time_remaining = nav2_util::geometry_utils::
-      //   estimate_navigation_time_remaining(distance, current_velocity);
+      std::cout << "=============== Current stats ================" << std::endl
+      << "Distance remaining: " << feedback_msg->distance_remaining << std::endl
+      << "Speed: " << current_linear_speed_ << std::endl
+      << "Time remaining: " << estimated_navigation_time_remaining_.seconds() << "." <<
+      estimated_navigation_time_remaining_.nanoseconds() << std::endl;
 
       int recovery_count = 0;
       blackboard_->get<int>("number_recoveries", recovery_count);
+      feedback_msg->distance_remaining = distance_remaining;
+      feedback_msg->estimated_navigation_time_remaining = estimated_navigation_time_remaining_;
       feedback_msg->number_of_recoveries = recovery_count;
       feedback_msg->navigation_time = now() - start_time_;
       action_server_->publish_feedback(feedback_msg);
@@ -347,5 +368,6 @@ BtNavigator::onGoalPoseReceived(const geometry_msgs::msg::PoseStamped::SharedPtr
   goal.pose = *pose;
   self_client_->async_send_goal(goal);
 }
+
 
 }  // namespace nav2_bt_navigator
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -22,7 +22,7 @@
 #include <vector>
 #include <set>
 
-#include "nav2_util/geometry_utils.hpp"
+// #include "nav2_util/geometry_utils.hpp"
 #include "nav2_util/robot_utils.hpp"
 #include "nav2_behavior_tree/bt_conversions.hpp"
 #include "nav2_bt_navigator/ros_topic_logger.hpp"
@@ -104,7 +104,7 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
     "goal_pose",
     rclcpp::SystemDefaultsQoS(),
     std::bind(&BtNavigator::onGoalPoseReceived, this, std::placeholders::_1));
-  
+
   // Odometry smoother object for getting current speed
   odom_smoother_ = std::make_unique<nav2_util::OdomSmoother>(client_node_, 0.3);
 
@@ -285,20 +285,17 @@ BtNavigator::navigateToPose()
       blackboard_->get("path", current_path);
 
       // Calculate distance on the path
-      double distance_remaining = nav2_util::geometry_utils::calculate_distance_to_goal(current_path);
+      double distance_remaining =
+        nav2_util::geometry_utils::calculate_distance_to_goal(current_path);
 
-      if (distance_remaining <= transform_tolerance_)
-      {
+      if (distance_remaining <= transform_tolerance_) {
         estimated_navigation_time_remaining_ = rclcpp::Duration::from_seconds(0.0);
-      } 
-      else
-      {
+      } else {
         // Get current speed
         current_linear_speed_ = odom_smoother_->getTwist().linear.x;
 
         // Calculate estimated time taken to goal if speed is higher than 1mm/s
-        if (std::abs(current_linear_speed_) > 0.05)
-        {
+        if (std::abs(current_linear_speed_) > 0.05) {
           estimated_navigation_time_remaining_ =
             rclcpp::Duration::from_seconds(distance_remaining / std::abs(current_linear_speed_));
         }
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -275,8 +275,22 @@ BtNavigator::navigateToPose()
       geometry_msgs::msg::PoseStamped goal_pose;
       blackboard_->get("goal", goal_pose);
 
+      // Get current path points
+      nav_msgs::msg::Path current_path;
+      blackboard_->get("path", current_path);
+
+      // Calculate distance on the path
+
+      // Get current speed
+
+      // Calculate estimated time taken to goal
+
+      // TODO (Deepak): Update this to add path distance computation
       feedback_msg->distance_remaining = nav2_util::geometry_utils::euclidean_distance(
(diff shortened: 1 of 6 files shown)


nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -275,11 +280,31 @@ BtNavigator::navigateToPose()
...
+        nav2_util::geometry_utils::calculate_path_length(current_path);
+
+      if (distance_remaining <= transform_tolerance_) {
+        estimated_navigation_time_remaining_ = rclcpp::Duration::from_seconds(0.0);
+      } else {
+        // Get current speed
+        current_linear_speed_ = odom_smoother_->getTwist().linear.x;
+
+        // Calculate estimated time taken to goal if speed is higher than 1mm/s
+        if (std::abs(current_linear_speed_) > 0.05) {
```
Timple, line 298: This seems to be 5cm/s?

Timple, line 298: What happens if the speed is lower?

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -275,11 +280,31 @@ BtNavigator::navigateToPose()
...
 
-      feedback_msg->distance_remaining = nav2_util::geometry_utils::euclidean_distance(
-        feedback_msg->current_pose.pose, goal_pose.pose);
+      // Get current path points
+      nav_msgs::msg::Path current_path;
+      blackboard_->get("path", current_path);
+
+      // Calculate distance on the path
+      double distance_remaining =
+        nav2_util::geometry_utils::calculate_path_length(current_path);
```
Timple, line 289: This returns the total path length, so how is this `distance_remaining`?

Not too familiar with the codebase yet, might be that `current_path` is a bit misleading name and actually means `path_left` :slightly_smiling_face:

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: Oops, I forgot to change that back! I meant for it to be 5mm/s :)

deepaktalwardt: As the speed tends towards 0, the time remaining division tends to infinity, which is not what we want. That's why I added a threshold below which the time remaining won't update.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: The path on the blackboard updates over time. So the `current_path` is essentially path remaining.

deepaktalwardt: But I could potentially find the point closest to the current pose of the robot and then calculate the distance based off of that pose, not sure if that's actually better though.

nav2_bt_navigator/src/bt_navigator.cpp
Timple: Ah, it simply doesn't update of course.

But why don't we want infinity? I think it's the most correct value. Limiting to 5mm/s seems a bit arbitrary.

nav2_bt_navigator/src/bt_navigator.cpp
Timple: > The path on the blackboard updates over time. So the current_path is essentially path remaining.

> But I could potentially find the point closest to the current pose of the robot and then calculate the distance based off of that pose

What is the difference between these two implementations?

nav2_msgs/action/NavigateToPose.action
```diff
@@ -7,5 +7,6 @@ std_msgs/Empty result
 ---
 geometry_msgs/PoseStamped current_pose
 builtin_interfaces/Duration navigation_time
+builtin_interfaces/Duration estimated_navigation_time_remaining
```
SteveMacenski, line 10:
```suggestion
builtin_interfaces/Duration estimated_time_remaining
```

Lets keep it a little more concise :-)

nav2_util/test/test_geometry_utils.cpp
```diff
@@ -49,3 +51,24 @@ TEST(GeometryUtils, euclidean_distance_pose)
 
   ASSERT_NEAR(euclidean_distance(pose1, pose2), 10.24695, 1e-5);
 }
+
+TEST(GeometryUtils, calculate_path_length)
```
SteveMacenski, line 55: Can you add a test where you generate a circle and test that the circle perimeter is within reasonable tolerances to the actual radius? That way its not just a strightline test

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -62,6 +63,15 @@ inline double euclidean_distance(
   return euclidean_distance(pos1.pose, pos2.pose);
 }
 
+inline double calculate_path_length(const nav_msgs::msg::Path & path)
```
SteveMacenski, line 66: Good thinking adding this to the utils :+1:

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: Because divide by zeros are not reliable. I'd rather get a "I don't know" than a "NaN" or "Inf".

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: I suppose it depends - if the BT has absolutely no replanning, then yeah this would be never changing. However if there is replanning, then the `path` gets updated at the rate / time / distance in which the plan is being updated at.

We could do something like what we do in the regulated pure pursuit controller where we prune the path we already passed, but what do you think @Timple ? Is assuming replanning and updating this data at that rate sufficient?

SteveMacenski (changes requested):

SteveMacenski:
Thanks for this - great intro to the stack and I appreciate that you added tests off the bat without me having to ask.

We typically target PRs to `main` though, not foxy. In ROS2, we maintain usually master/main branches for development that are released into distributions in syncs (unlike in ROS1). This is so that we can make sure we maintain ABI compatibility. This will by synced to Foxy in the next sync, but we should merge this into `main` so that every distribution moving forward has this work as well!

nav2_bt_navigator/src/bt_navigator.cpp
Timple: Isn't NaN the definition of "I don't know" / "Cannot compute"

SteveMacenski: I never want NaNs / Infs intentionally injected into my systems, as a personal thing. Error handling for them at all levels for all inputs is error prone and annoying

nav2_bt_navigator/src/bt_navigator.cpp
Timple: > Is assuming replanning and updating this data at that rate sufficient?

Not really, **exactly** in the case of using the pure pursuit controller. The main reason to use the pure pursuit controller would be to stick to a global plan as good as possible. So then I don't expect any replans...

Or am I messing up the terminology here?

deepaktalwardt: From what I observed the `path` on the blackboard got updated at roughly 2 Hz, which resulted in pretty accurate action feedback.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: I agree with Steve on this. Without this logic, the ETA, for example, might look like this 3, 2, 1, 0.5, 12, 345, 646746, 1e9 seconds... when you're almost at your goal, which doesn't really make sense.

@SteveMacenski does 5mm/s make sense? I was trying to tune this threshold but I wasn't sure what would be reasonable here.

nav2_msgs/action/NavigateToPose.action
deepaktalwardt: Sounds good :)

nav2_util/include/nav2_util/geometry_utils.hpp
deepaktalwardt: thanks :)

deepaktalwardt:
>We typically target PRs to main though, not foxy. In ROS2, we maintain usually master/main branches for development that are released into distributions in syncs (unlike in ROS1). This is so that we can make sure we maintain ABI compatibility. This will by synced to Foxy in the next sync, but we should merge this into main so that every distribution moving forward has this work as well!
Ah I wasn't aware! Let's continue the discussion on this PR, once we agreed upon the code I'll merge `main` in and create a new PR from this branch. Thanks!

nav2_util/test/test_geometry_utils.cpp
deepaktalwardt: Great idea! Just added it and the results are within 10 cm of the circumference given a 1 deg discretization of the perimeter.

nav2_bt_navigator/src/bt_navigator.cpp
Timple: Ah, of course. You typically slow down on the end of the path. Having 3, 2, 1, 0.5, 12, 345, 646746 is terrible indeed!

I thought only of situations where you slow down the robot to zero halfway a path. For example on obstacles. In that case I thought `inf` to be a better estimation than 2 minutes.

nav2_bt_navigator/src/bt_navigator.cpp
Timple: > at roughly 2 Hz

That would be okay I guess. Are there any parameters that influence this 2hz? And can we have situations where it is actually 0hz?

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -275,11 +280,31 @@ BtNavigator::navigateToPose()
...
+
+      if (distance_remaining <= transform_tolerance_) {
+        estimated_time_remaining_ = rclcpp::Duration::from_seconds(0.0);
+      } else {
+        // Get current speed
+        current_linear_speed_ = odom_smoother_->getTwist().linear.x;
```
SteveMacenski, line 295: I don't think current_linear_speed_ or estimated_time_remaining_ need to be class variables - why not local to this function?

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: It all depends on the behavior tree used and what parameters you throw into it. All the BTs in nav2 by default have some type of consistent replanning based on time, distance traveled, or proportional rates to speed. It's certainly possible to have a situation where the path isn't updated, but none of the default trees do so. I think the default tree replans at 1hz.

I think it might be a good idea to prune the path in a local variable (read: don't update the blackboard path) and use that pruned path for this computation so you can get relatively frequent updated ETAs even with inconsistent replanning. 

What do you think @Timple ? Would it be reasonable to assume that the ETA/distances update at planning rate?

Timple: You're referring to the planning rate of the local planner right?
In that case that would be fine yes.

SteveMacenski: No, global planner

Timple: But as I mentioned. The global planner simply might not have an update rate. Sometimes we just want to follow 'exactly this plan'. Without any replanning.
In that case the ETA/distances would never be updated.

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: but it would be `3, 2, 1, 0.5, 0, 0, 0` if you set it to something sufficiently low that you really shouldn't see it except when actually stopped or very near to it (like 5 mm/s).

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: So then maybe pruning on this update is valuable. You already have the current pose, so with that and the path, you can find the nearest point to the robot, and throw that to the end of the path to the util function instead. You can see an example of a lambda that does this in robot base frame in the regulated pure pursuit controller.

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski, line 295: In case holonomic is used, you could to a `hypot(linear.x, linear.y)` so that you can capture the actual speed in all directions. This only really works for differential robots (non-ackermann, non-omnidirectional)

nav2_bt_navigator/src/bt_navigator.cpp
doisyg: I second this pruning approach. I have a couple of use cases (bt setted) where the global plan is rarely if not never updated

SteveMacenski:
@deepaktalwardt sounds like the pruning is probably required to cover our bases

deepaktalwardt:
> @deepaktalwardt sounds like the pruning is probably required to cover our bases

@SteveMacenski sounds good. I'll attempt adding that this weekend.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: @SteveMacenski how can I tell what type of robot is being used? It is true I assumed differential drive here.

deepaktalwardt: The reason why I made these class variables is that I wanted to preserve previous values in case we don't need to update them (like when speed is less than 5mm/s). Does that make sense?

deepaktalwardt: Actually, you can ignore that question, `hypot(linear.x, linear.y)` should work in all cases where the robot motion is in the xy plane.

SteveMacenski: It is computed every iteration though except when `distance_remaining > transform_tolerance_`. What you could do instead is just create those 2 variables locally and set them to `0` on instantiation. Performance wise it doesn't really matters either way, but its more of a code-cleanliness thing so that we only have class variables for things that really are used in multiple places or through various iterations.

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -269,17 +274,58 @@ BtNavigator::navigateToPose()
...
+          return closest_pose_idx;
+        };
+
+      size_t closest_pose_idx = find_closest_pose_idx();
+
+      // Calculate distance on the path
+      double distance_remaining =
+        nav2_util::geometry_utils::calculate_path_length(current_path, closest_pose_idx);
+
+      if (distance_remaining <= transform_tolerance_) {
```
SteveMacenski, line 309: This check is not great, `transform_tolerance` is a time (e.g. 0.1 seconds), using a time to check distance is unit mismatch and for an unrelated purpose.

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: Yeah the more I look at this with values stored between iterations, the more I see opportunities for edge cases, please remove those class variables and initialize them as `0`s so they are `0` if whatever `if` statements don't pass to populate them.

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -62,6 +63,18 @@ inline double euclidean_distance(
   return euclidean_distance(pos1.pose, pos2.pose);
 }
 
+inline double calculate_path_length(const nav_msgs::msg::Path & path, size_t start_index)
```
SteveMacenski, line 66: Default start index to `0` so that it can be used as is without inputting that field

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -269,17 +274,58 @@ BtNavigator::navigateToPose()
...
+            }
+          }
+          return closest_pose_idx;
+        };
+
+      size_t closest_pose_idx = find_closest_pose_idx();
+
+      // Calculate distance on the path
+      double distance_remaining =
+        nav2_util::geometry_utils::calculate_path_length(current_path, closest_pose_idx);
```
SteveMacenski, line 307:
```suggestion
        nav2_util::geometry_utils::calculate_path_length(current_path, find_closest_pose_idx());
```

This lets you remove L303 `size_t closest_pose_idx = find_closest_pose_idx();` and the white space around it

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -269,17 +274,58 @@ BtNavigator::navigateToPose()
...
+      // Get current path points
+      nav_msgs::msg::Path current_path;
+      blackboard_->get("path", current_path);
+
+      // Find the closest pose to current pose on global path
+      auto find_closest_pose_idx = 
+        [&current_pose, &current_path]() {
+          size_t closest_pose_idx = 0;
+          double curr_min_dist = std::numeric_limits<double>::max();
+          for (auto curr_it = current_path.poses.begin(); curr_it != current_path.poses.end(); ++curr_it) {
```
SteveMacenski, line 293: Why not iterate with indices if you're trying to extract indices later? Using iterators in this context doesn't save you anything.

SteveMacenski (changes requested):

nav2_bt_navigator/src/bt_navigator.cpp
SteveMacenski: I don't think this `if` is even required if you remove the class variables to be local scope variables. You can just use the `else` statement since L318 below will make sure these values are nominally large enough. Else, it would just remain 0 from initialization as a local scope variable.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: I totally agree, I'll remove iterators here.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: Oh wow, I completely misunderstood the `transform_tolerance` threshold. My bad! And I agree with both of your suggestions.

nav2_util/include/nav2_util/geometry_utils.hpp
deepaktalwardt: Sounds good!

SteveMacenski (approved):
LGTM! Just needs to have the `main` version as well passing CI and we can merge both!

deepaktalwardt:
A PR into `main` was created here https://github.com/ros-planning/navigation2/pull/2246. Closing this PR.

Issue #2196, opened 2021-02-19
Provide slam_toolbox parameter file argument through nav2_bringup


v-lopez:
#### Feature description
Currently nav2_bringup/bringup_launch.py is perfect for launching the navigation of a robot. With minimal modifications to the parameter file I can run [on our PMB2 robot](https://github.com/pal-robotics/pmb2_navigation/blob/foxy-devel/pmb2_2dnav/params/pmb2_params.yaml). 

But when setting `slam:=True`, `bringup_launch.py` includes `slam_launch.py` which includes `slam_toolbox/online_sync_launch.py` without passing a parameter file. It always loads the one included on the package.

To change such params, I have to create a separate online_sync_launch.py file, and therefore modify bringup_launch.py and slam_launch.py.

SteveMacenski:
I'd be happy to consider a specific way of supporting your requested behavior in `nav2_bringup` directly, if it works for all else for your robots. Do you have a specific suggestion of what to change in Nav2/SLAM Toolbox? 

Keep in mind that is an example bringup package, its expected that a specific robot or specific company creates their own `<my-robot>_bringup` or `<my-robot>_nav` package containing your launch / configuration files no differently from ROS1. I'm happy that our provided one so closely mirrors your needs so I'd like to see if we could make it work general-purpose - but that's kind of what I was thinking in general.

v-lopez:
Short answer, I believe that it might be interesting to be able to specify the slam_toolbox parameters in the params file where the rest of parameters are.  I have not had the time to look at what all the nodes do, and whether it makes sense that they are all in the same file or not.

Regarding the `<my-robot>_bringup`, one of my issues with ROS2 is how complex launch files have become. 
For years we've been eager to have python launch files because the ROS1 XML where limiting us in many ways, but the current implementation can be difficult to work with. This is not navigation2's fault, it's just the way it is now.

Right now the basic launch files in nav2_bringup amount to 500 lines of python code. We have to mantain navigation for at least 5 different robots, hence why I tried to avoid copying them and using directly nav2_bringup_launch.py with just a different param file. 

Eventually I'll guess I'll have to split and maybe only reuse `slam_launch.py` and rewrite some of the others. But every reusable bit reduces entropy.

SteveMacenski:
[...]

PR #2214, opened 2021-03-03
If provided param file has no slam_toolbox params, don't forward it

v-lopez:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2196|

---

* After https://github.com/SteveMacenski/slam_toolbox/pull/347 params-file is automatically passed to slam_toolbox, these changes make sure that only happens if the params_file has `slam_toolbox` parameters.

### DIFF
--- a/nav2_bringup/bringup/launch/slam_launch.py
+++ b/nav2_bringup/bringup/launch/slam_launch.py
@@ -18,13 +18,13 @@ from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
 from launch.actions import (DeclareLaunchArgument, IncludeLaunchDescription,
-                            PushLaunchConfigurations, PopLaunchConfigurations,
+                            PopLaunchConfigurations, PushLaunchConfigurations,
                             UnsetLaunchConfiguration)
 from launch.conditions import UnlessCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
-from nav2_common.launch import RewrittenYaml, HasNodeParams
+from nav2_common.launch import HasNodeParams, RewrittenYaml
 
 
 def generate_launch_description():
@@ -75,7 +75,7 @@ def generate_launch_description():
     # Nodes launching commands
 
     # If the provided param file doesn't have slam_toolbox params, we must remove the 'params_file'
-    # LaunchConfiguration, or it will be passed automatically to slam_toolbox and will not load 
+    # LaunchConfiguration, or it will be passed automatically to slam_toolbox and will not load
     # the default file
     has_slam_toolbox_params = HasNodeParams(source_file=params_file,
                                             node_name='slam_toolbox')
@@ -84,7 +84,7 @@ def generate_launch_description():
             condition=UnlessCondition(has_slam_toolbox_params))
 
     remove_params_file = UnsetLaunchConfiguration(
-            "params_file", condition=UnlessCondition(has_slam_toolbox_params))
+            'params_file', condition=UnlessCondition(has_slam_toolbox_params))
 
     start_slam_toolbox_cmd = IncludeLaunchDescription(
         PythonLaunchDescriptionSource(slam_launch_file),
--- a/nav2_bringup/bringup/launch/slam_launch.py
+++ b/nav2_bringup/bringup/launch/slam_launch.py
@@ -74,25 +74,6 @@ def generate_launch_description():
 
     # Nodes launching commands
 
-    # If the provided param file doesn't have slam_toolbox params, we must remove the 'params_file'
-    # LaunchConfiguration, or it will be passed automatically to slam_toolbox and will not load
-    # the default file
-    has_slam_toolbox_params = HasNodeParams(source_file=params_file,
-                                            node_name='slam_toolbox')
-    # slam_toolbox with provided param_file
-    push_launch_config = PushLaunchConfigurations(
-            condition=UnlessCondition(has_slam_toolbox_params))
-
-    remove_params_file = UnsetLaunchConfiguration(
-            'params_file', condition=UnlessCondition(has_slam_toolbox_params))
-
-    start_slam_toolbox_cmd = IncludeLaunchDescription(
-        PythonLaunchDescriptionSource(slam_launch_file),
-        launch_arguments={'use_sim_time': use_sim_time}.items())
-
-    pop_launch_config = PopLaunchConfigurations(
-            condition=UnlessCondition(has_slam_toolbox_params))
-
     start_map_saver_server_cmd = Node(
             package='nav2_map_server',
             node_executable='map_saver_server',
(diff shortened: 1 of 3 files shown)


nav2_bringup/bringup/launch/slam_launch.py
```diff
@@ -17,11 +17,14 @@
 from ament_index_python.packages import get_package_share_directory
```
SteveMacenski, line 17: https://app.circleci.com/pipelines/github/ros-planning/navigation2/4753/workflows/14e3b0f6-9171-4b98-88ca-0aef32421b05/jobs/17651 bunches of linter errors

nav2_common/nav2_common/launch/has_node_params.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2019 Intel Corporation
```
SteveMacenski, line 1: I didn't realize you worked for intel?

nav2_bringup/bringup/launch/slam_launch.py
```diff
@@ -70,10 +73,26 @@ def generate_launch_description():
...
+    has_slam_toolbox_params = HasNodeParams(source_file=params_file,
+                                            node_name='slam_toolbox')
+    # slam_toolbox with provided param_file
+    push_launch_config = PushLaunchConfigurations(
+            condition=UnlessCondition(has_slam_toolbox_params))
+
+    remove_params_file = UnsetLaunchConfiguration(
+            "params_file", condition=UnlessCondition(has_slam_toolbox_params))
+
     start_slam_toolbox_cmd = IncludeLaunchDescription(
```
SteveMacenski, line 89: I don't see where slam toolbox would get the parameters, the launch arguments of this don't have the param file given.

nav2_bringup/bringup/launch/slam_launch.py
```diff
@@ -70,10 +73,26 @@ def generate_launch_description():
...
 
     # Nodes launching commands
+
+    # If the provided param file doesn't have slam_toolbox params, we must remove the 'params_file'
+    # LaunchConfiguration, or it will be passed automatically to slam_toolbox and will not load 
+    # the default file
+    has_slam_toolbox_params = HasNodeParams(source_file=params_file,
+                                            node_name='slam_toolbox')
+    # slam_toolbox with provided param_file
+    push_launch_config = PushLaunchConfigurations(
```
SteveMacenski, line 83: If you have `has_slam_toolbox_params`, why not just use normal python `if else` to call `start_slam_toolbox_cmd` with and without the param file if it has it?

SteveMacenski (changes requested):

nav2_common/nav2_common/launch/has_node_params.py
v-lopez: Fixing, sorry since it started as a blatant copy of `rewritten_yaml.py` I never know at which point it's "our" copyright.

nav2_bringup/bringup/launch/slam_launch.py
v-lopez: Because I can't evaluate a LaunchConfiguration until it is executed.

If I understood correctly how `launch` works, there's two ways of processing a launch.py file, "describe" and "execute".
In the describe phase has_slam_toolbox_params does not evaluate to "True" or "False", but to a description of the substitution. Only on execution, it is able to resolve `params_file` into a file and see if it contains the node parameters.

nav2_bringup/bringup/launch/slam_launch.py
v-lopez: All LaunchConfigurations are passed automatically to included LaunchDescriptions.

It is discussed here: https://github.com/ros2/launch/issues/313

A comment by wjwwood
>  We would like to get back to this with launch. Currently an included launch description has access to all configurations of the launch file that included it.

nav2_bringup/bringup/launch/slam_launch.py
v-lopez: Sorry, I was not too eager on giving access to my github account to a third party application just to see CI results. I now saw that there's an option to give access only to Public repos. I'll check now.

nav2_bringup/bringup/launch/slam_launch.py
SteveMacenski: Maybe I'm missing something, but wouldn't 

```
    has_slam_toolbox_params = HasNodeParams(source_file=params_file,
                                            node_name='slam_toolbox')

    start_slam_toolbox_cmd = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(slam_launch_file),
        condition=IfCondition(has_slam_toolbox_params),
        launch_arguments={'use_sim_time': use_sim_time, 'params_file', params_file}.items())
    start_slam_toolbox_cmd = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(slam_launch_file),
        condition=UnlessCondition(has_slam_toolbox_params),
        launch_arguments={'use_sim_time': use_sim_time}.items())
```

work? `Only on execution, it is able to resolve params_file into a file and see if it contains the node parameters.` I'm not sure I see how that would make this not work? You would be executing this at runtime in all cases of python. The if/unless here would be analog to ROS1 XML as well and I think this should work, its how we do [`if/else` on slam vs localization](https://github.com/ros-planning/navigation2/blob/dcc1d4dafba6f6c7588cb5f7f6ca918b2a6c48ef/nav2_bringup/bringup/launch/bringup_launch.py#L92-L109)

nav2_bringup/bringup/launch/slam_launch.py
SteveMacenski: But I thought we talked about adding a params file field that will be given to the nodes? https://github.com/SteveMacenski/slam_toolbox/blob/foxy-devel/launch/online_sync_launch.py#L12 so if you don't give a `launch_arguments` of `params_file` then it would default to the given online sync slam?

SteveMacenski: This solution might very well work, but it makes launch even more confusing so looking for ways to keep the patterns at least consistent with other launch files rather than introducing new patterns. If humanly possible.

nav2_bringup/bringup/launch/slam_launch.py
v-lopez: My first implementation was exactly like this.

But I faced the problem of the discussion above, that all this file LaunchConfigurations are passed automatically to included configurations, regardless of specifiying `'params_file', params_file`. So I had to do the UnsetLaunchConfiguration thing.

So the file was always being passed, regardless of the value of has_slam_toolbox_params.

nav2_bringup/bringup/launch/slam_launch.py
v-lopez: Right now, `online_sync_launch.py` will use the `params_file` of the `slam_launch.py` automatically, unless you unset it, due to the propagation of the LaunchConfiguration by default that I linked above.

If we changed `online_sync_launch.py` LaunchConfiguration name to `slam_toolbox_params_file`, then it would not be automatically passed. To pass it, it would need to be set explicitely via `launch_arguments={'use_sim_time': use_sim_time, 'slam_toolbox_params_file', params_file}.items())`

I agree, it is quite confusing. How would you like to proceed?
I can change the LaunchConfiguration name in `slam_launch.py`, but it might be a hacky solution that hides the problem.

I can explicitely pass it here, but we also need to unset it, so it becomes something like this monster:
```
    has_slam_toolbox_params = HasNodeParams(source_file=params_file,
                                            node_name='slam_toolbox')

    start_slam_toolbox_cmd = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(slam_launch_file),
        condition=IfCondition(has_slam_toolbox_params),
        launch_arguments={'use_sim_time': use_sim_time, 'params_file', params_file}.items())


   # All this below is required to NOT pass the params_file automatically to the included launch description
    push_launch_config = PushLaunchConfigurations(
            condition=UnlessCondition(has_slam_toolbox_params))

    remove_params_file = UnsetLaunchConfiguration(
            "params_file", condition=UnlessCondition(has_slam_toolbox_params))

    start_slam_toolbox_cmd = IncludeLaunchDescription(
        PythonLaunchDescriptionSource(slam_launch_file),
        condition=UnlessCondition(has_slam_toolbox_params),
        launch_arguments={'use_sim_time': use_sim_time}.items())

    pop_launch_config = PopLaunchConfigurations(
            condition=UnlessCondition(has_slam_toolbox_params))







```

nav2_bringup/bringup/launch/slam_launch.py
SteveMacenski: OK. With that in mind, can we move `start_map_saver_server_cmd` and `start_lifecycle_manager_cmd` before all this stuff so its at the end? 

I see you `UnsetLaunchConfiguration`, do you need to re-`LaunchConfiguration` it? I see how push/pop works. In your comment on L77, adding a sentence or two about what specifically is happening here would be really good for both my sanity in the future and anyone else reading this without context. It explains well why this is happening but doesn't talk about the push/pop.

v-lopez: I've reordered and extended comments, please take a look when you have time.

And sorry about the mess, I also thought such a mess would not be required to pass a parameter optionally.

SteveMacenski (approved):
If you've tested this working to your satisfaction, I'll merge

SteveMacenski:
Oh @v-lopez we also need an analog of this PR for `ros2` on SLAM Toolbox (not just foxy) so that all new distros will get this work too https://github.com/SteveMacenski/slam_toolbox/pull/347

v-lopez:
Wasn't that done here? https://github.com/SteveMacenski/slam_toolbox/pull/348

SteveMacenski:
Ah sorry, I missed / forgot about that :-)

Issue #816, opened 2019-06-10
Get rid of service client nodes


mjeronimo:
There are several nodes in the system that are created and used solely for interfacing to services. These "client nodes" are used because spin_until_future_complete cannot accept a node that is already on an executor. Instead, we can use future.wait_until instead of spin_until_future_complete. This will allow us to remove several of these client nodes in the system, which will hopefully ease discovery congestion.

mjeronimo:
@orduno See #808 for some background about this issue and a quick code sample.

SteveMacenski:
@mjeronimo is this still an issue?

crdelsey:
It is still an issue. Having too many nodes impacts startup times and seems to overwhelm the current DDS implementations.

Possible fixes to this issue are:
 1. Fix the DDS implementations to better handle many nodes.
 2. Fix rcl to allow re-entrant spinning on the current node.
 2. Fix rcl/rclcpp to allow continuations to be returned from service callbacks.
 3. Re-architect our service calls to cleverly use callback groups and multi-threaded executors.
 4. Re-architect nav2 to not wait for network activity in the context of a callback.

SteveMacenski:
@crdelsey can you provide a few examples in the stack where we use "client nodes", other than costmap spinning, I had a hard time finding any

crdelsey:
I think the big one is here in ServiceClient. https://github.com/ros-planning/navigation2/blob/7652518af52f257d097269309986943771d98683/nav2_util/include/nav2_util/service_client.hpp#L38

Then, there's another use in LifecycleManagerClient (https://github.com/ros-planning/navigation2/blob/a095038da9f4e3729679e5d20b39e1d95582fb16/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp#L30).

There may be other places.

SteveMacenski:
Got it, I didn't see that hidden in there. "Fix DDS" I feel like is a never ending battle, I'd say 2 or 3 seem good general long term fixes but 4 seems more practically in our control.

SteveMacenski:
Any objection to using multithreaded executors? It seems that way we can have these all going at once with one node. Do you think there are applications here sensitive enough (or otherwise not well formatted) for multithreaded processing? Most places have mutex where we might want to use a multithreaded spinner. I wouldn't think it would be too much of a hassle to update the few places required.

shivaang12:
@mjeronimo Do you have any updates on this? Let me know even if you have any unfinished work on this.

[...]

PR #2216, opened 2021-03-04
ServiceClient use callback group

BriceRenaudeau:
---

## Basic Info

| ------ | ----------- |

---

* I change the timeout default value to fix the runtime error that appears when using `service_client->invoke(req)`.
* I create a `callback_group` and a `SingleThreadedExecutor` to spin_until_future_complete.

This allows the use of the `ServiceClient` the `invoke` method in a callback (topic, timer, service) handle by the node.
```
static std::shared_ptr<nav2_util::ServiceClient<std_srvs::srv::Trigger>> service_client;

void sendRequest(){
  // Send request
  res_received = false;
  auto service_req = std::make_shared<std_srvs::srv::Trigger::Request>();
  auto service_res = service_client->invoke(service_req); // std::chrono::seconds(1) default value 

  if (!service_res->success){
    RCLCPP_WARN_STREAM(node->get_logger(), "Someting went wrong");
  } else {
    RCLCPP_INFO_STREAM(node->get_logger(), "Everything is fine");
  }
  timer->reset();
  return;
}

int main(int argc, char ** argv){
  rclcpp::init(argc, argv);

  node = std::make_shared<rclcpp::Node>("main_node");

  // Creates service client
  service_client = std::make_shared<nav2_util::ServiceClient<std_srvs::srv::Trigger>>("my_service", node);

  // Wait for service server
  service_client->wait_for_service();

  // Creates timer to call service server every second
  timer = node->create_wall_timer(std::chrono::seconds(1), &sendRequest);

  rclcpp::spin(node);
  rclcpp::shutdown();
  return 0;
}
```

## Discution

This will create a callback group for every instance of `ServiceClient`.
This allows you to not create a "subnode" like `_rclccp_node` just for a service call, it doesn't break the API if you really want to create one.


SteveMacenski:
Can you add a unit test showing that you can process the service call in a callback function of a subscriber (very similar to the code block you provided)? That would be very helpful to demonstrate that this fixes that issue (and also lets us have a test in case there's a regression in rclcpp on this). Does this also currently work with action servers in `rclcpp`? If so, this is a **really** nice thing to know, as then we might be able to reduce the number of DDS participants by 30-40% in Nav2.  

2 such places for a demonstration that this fixes the issue are:
- https://github.com/ros-planning/navigation2/blob/992deb0ff740e60189234455ac8c372fcaf2759c/nav2_controller/src/nav2_controller.cpp#L174-L181 (service + action)
- https://github.com/ros-planning/navigation2/blob/567f8c85b77e4b569c1b76b36c4275a57847a86c/nav2_waypoint_follower/src/waypoint_follower.cpp#L67-L75 (action + action)

If you can provide a reference implementation, I can (or someone I can find) work on getting this done across all of the nodes.

SteveMacenski:
CC https://github.com/ros-planning/navigation2/issues/816

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -37,22 +37,36 @@ class ServiceClient
     } else {
       node_ = generate_internal_node(service_name + "_Node");
```
SteveMacenski, line 38: I know this is not part of your PR, but if this really allows us to fix what you propose, you can remove the default node parameter requiring `generate_internal_node`, and require that a user provide a node to use. I looked through the codebase and there's 1 test you'd need to update/remove but otherwise that should work.

BriceRenaudeau:
Thanks for your quick review.
I will make the unit test for the service call in a topic callback.

With @doisyg we are currently using detached threads to perform action calls in an action server.
But we will test a way to use callback groups in action server.

BriceRenaudeau:
> Does this also currently work with action servers calling action servers in rclcpp?

The root problème is that the standard implementation of a node is like this :
```
int main(int argc, char **argv)
{
  rclcpp::init(argc, argv);
  auto node = std::make_shared<MyActionNode>();
  rclcpp::spin(node);
  rclcpp::shutdown();
  return 0;
}
```
The `spin(node)` function is here to make the node do something and not quit the main.
But we cannot use `spin_some()` or `spin_until_future_complete()` in a callback handle by the node thread.
We need to create a separate thread to be spin -> `Callbackgroup`.

So there are 2 options.

1. Attach the server(action/subscriber/timer) to a callbackgroup and make it spin in a separate thread. So we can make and wait the service/action call using `.wait()` method on the future. 
2. Attach the server(action/subscriber/timer) to the node and make it spin the standard way. Then we can attach the service/action client to a callbackgroup, and wait for the result using `callbackgroup_executor.spin_until_future_complete()`.

The second method looks more appropriate to me and is what I made for the service client.

To answer the question, an action client can be attached to a callbackgroup and we can use `callbackgroup_executor.spin_until_future_complete()` to wait the result. So there is all the chance for it to work.

I didn't try it yet, we are currently using another pattern based on std::thread:

_In the constructor_
```
// create sync action clients
sync_action_client_cb_group = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive, false);
action_client = rclcpp_action::create_client<Action>(
          this->get_node_base_interface(),
          this->get_node_graph_interface(),
          this->get_node_logging_interface(),
          this->get_node_waitables_interface(),
          "action_name",
          sync_action_client_cb_group);
// spin callback groups
callback_group_executor_thread = std::thread([this]() {
      callback_group_executor.add_callback_group(sync_srv_client_cb_group, this->get_node_base_interface());
      callback_group_executor.add_callback_group(sync_action_client_cb_group, this->get_node_base_interface());
      callback_group_executor.spin();
    });
```
_in the action server callback_
```
auto action_client_future = action_client->async_send_goal(*goal);
auto action_result = action_client->async_get_result(action_client_future.get());
while(action_result.wait_for(std::chrono::milliseconds(10) != std::future_status::ready)
```

BriceRenaudeau:
I made a class to handle any action type.

```c++
#include <rclcpp/rclcpp.hpp>
#include <rclcpp_action/rclcpp_action.hpp>

template<class ActionT>
class SimpleActionClient
{
public:
  SimpleActionClient(
      const std::string & action_name,
      const rclcpp::Node::SharedPtr & provided_node)
  : action_name_(action_name), node_(provided_node)
  {
    callback_group_ = node_->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive, false);
    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
    client_ = rclcpp_action::create_client<ActionT>(
          node_->get_node_base_interface(),
          node_->get_node_graph_interface(),
          node_->get_node_logging_interface(),
          node_->get_node_waitables_interface(),
          action_name_,
          callback_group_);
  }

  using GoalHandle = typename rclcpp_action::ClientGoalHandle<ActionT>;
  using Feedback = typename ActionT::Feedback;
  using WrappedResult = typename GoalHandle::WrappedResult;

public:
  WrappedResult
  send_goal(
    typename ActionT::Goal::SharedPtr & goal,
    const std::chrono::milliseconds timeout = std::chrono::milliseconds::max())
  {
    if(!client_){
      throw std::runtime_error(
              action_name_ + " action client not initialized");
    }

    while (!client_->action_server_is_ready()) {
      if (!rclcpp::ok()) {
        throw std::runtime_error(
                action_name_ + " action client: interrupted while waiting for service");
      }
      RCLCPP_INFO(
        node_->get_logger(), "%s action client: waiting for service to appear...",
        action_name_.c_str());
      rclcpp::Rate(100).sleep();
    }

    RCLCPP_DEBUG(
      node_->get_logger(), "%s action client: send async goal",
      action_name_.c_str());
    auto future = client_->async_send_goal(*goal);
    callback_group_executor_.spin_until_future_complete(future);

    auto result = client_->async_get_result(future.get());
    auto result_status = rclcpp::FutureReturnCode::TIMEOUT;
    // TODO : make clear timeout implementation
    while(rclcpp::ok() && result_status != rclcpp::FutureReturnCode::SUCCESS)
    {
      result_status = callback_group_executor_.spin_until_future_complete(
            result,
            std::chrono::milliseconds(10));
      RCLCPP_DEBUG(
        node_->get_logger(), "%s action client: waiting for result",
            action_name_.c_str());
    }

    if (future.get()->get_status() != rclcpp_action::GoalStatus::STATUS_SUCCEEDED){
      throw std::runtime_error(
            action_name_ + " action client: async_send_request failed");
    }

    return result.get();
  }

  void
  wait_for_action_server(
      const std::chrono::milliseconds timeout = std::chrono::milliseconds::max())
  {
    client_->wait_for_action_server(timeout);
  }

private:
  std::string action_name_;
  typename rclcpp_action::Client<ActionT>::SharedPtr client_;
};
```
And made a quick test and it works perfectly for the action client.
```c++
void timerCallback()
{
  auto goal = std::make_shared<Fibonacci::Goal>();
  goal->order = goal_order;

  auto action_result = action_client->send_goal(goal);

  if (action_result.code != rclcpp_action::ResultCode::SUCCEEDED){
    RCLCPP_INFO_STREAM(node->get_logger(), "Goal not achieved. Will try again to send goal order " << goal_order);
  } else {
    goal_order++;
  }

  return;
}

int main(int argc, char ** argv){
  rclcpp::init(argc, argv);
  node = std::make_shared<rclcpp::Node>("fibonacci_client_node");

  // Creates action client
  action_client = std::make_shared<SimpleActionClient<Fibonacci>>("fibonacci", node);

  // Wait for action server
  action_client->wait_for_action_server();

  // Creates timer to call action server every second
  auto timer = node->create_wall_timer(std::chrono::seconds(1), &timerCallback);

  rclcpp::spin(node);
  rclcpp::shutdown();
  return 0;
}
```

But I think this is not the topic of this PR.
And there are several points to discuss about the behavior (timeout) and the implementation.

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -29,30 +28,25 @@ class ServiceClient
...
-  }
-
-  ServiceClient(const std::string & service_name, const std::string & parent_name)
-  : service_name_(service_name)
-  {
-    node_ = generate_internal_node(parent_name + std::string("_") + service_name + "_client");
-    client_ = node_->create_client<ServiceT>(service_name);
+    callback_group_ = node_->create_callback_group(
+      rclcpp::CallbackGroupType::MutuallyExclusive,
+      false);
```
SteveMacenski, line 36: This is me asking a question because I haven't used callback groups like this before (previously in the context of priorities only):

I see the second argument means whether to add it to the node's executor or not. Why do you need to run it against a separate executor if its in another callback group? Wouldn't the `MutuallyExclusive` callback group type make it so that it could be processed at the same time as other callbacks in the same executor?

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -29,30 +28,25 @@ class ServiceClient
...
+      callback_group_);
   }
 
   using RequestType = typename ServiceT::Request;
   using ResponseType = typename ServiceT::Response;
 
   typename ResponseType::SharedPtr invoke(
     typename RequestType::SharedPtr & request,
-    const std::chrono::nanoseconds timeout = std::chrono::nanoseconds::max())
+    const std::chrono::seconds timeout = std::chrono::seconds(1))
```
SteveMacenski, line 49: Don't change the defaults, please change back to max as default argument. 1 second is pretty arbitrary for a random system and no-timeout is a pretty reasonable default.

SteveMacenski:
```
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from ServiceClient
[ RUN      ] ServiceClient.can_ServiceClient_use_passed_in_node
[       OK ] ServiceClient.can_ServiceClient_use_passed_in_node (8 ms)
[ RUN      ] ServiceClient.can_ServiceClient_invoke_in_callback
/opt/overlay_ws/src/navigation2/nav2_util/test/test_service_client.cpp:83: Failure
  a
    Which is: 0
  1
terminate called without an active exception
```

your new test actually fails

SteveMacenski:
Overall, this topic brings up a good point that callback groups / multithreaded executors have now become profoundly more complete than when we originally designed the stack in Beta 1 days. We can probably get rid of alot of these `rclcpp_node_` and `client_node_`s now using this type of design pattern (need to look into each use-case individually first).

If you had some thoughts on #816, I'd appreciate if you dropped a comment in there with some thoughts. My knowledge on this topics is woefully outdated it appears from reviewing this PR (and still no good ROS2 docs to explain it either *sigh* back to the source code).

This PR looks very reasonable, I'd just like to see it in use somewhere in the stack where this feature is highlighted (and I don't think there are any, looks like `ServiceClient` is only used in the lifecycle manager state transition service calls) or potentially a follow-on PR removing some of these `client_node_` or `rclcpp_node_`'s in favor for this as a design reference. Not necessary if we have tests that proves it doesn't impact default behavior, but you've started a good discussion that could really impact performance here in a very positive way. If you had interest in continuing down this line of thinking in the project, I'd be **very** happy to get rid of some of these nodes and grateful for that contribution! Even doing it in 1-2 places I can use as a reference would go a long way.

nav2_util/include/nav2_util/service_client.hpp
SteveMacenski: I've done some digging, is the reason that doesn't work because the executor being run right now is single threaded? So you're using a second single threaded executor to execute this callback group separately due to the lack of multithreaded executor? 

Edit: I've found that the default callback group when not set is MutuallyExclusive single threaded executor, so if you added this to the existing one, then it wouldn't work because it was already blocked. It seems odd to me to have an internal executor - but I think this is OK. Any significan toverhead associated with just an orphaned executor over here for just this object?

nav2_util/include/nav2_util/service_client.hpp
BriceRenaudeau: > Don't change the defaults, please change back to max as default argument. 1 second is pretty arbitrary for a random system and no-timeout is a pretty reasonable default. If you `invoke` and you don't want an infinite timeout, you should include one

I agree with you, but the problem is that using the default value the service client doesn't work. Using default timeout makes it fail "service client: async_send_request failed".

BriceRenaudeau:
> your new test actually fails

Strange, it works fine on my computer. I will investigate.

BriceRenaudeau:
> It sounds to me like if we have a multithreaded executor in all of our nodes, with callback groups for each individual `service` or `action` we should be good to go. Does that sound more or less right to you?

The issues with one multithreaded executor containing all the callback groups are when do you spin it? and how do you add new callback groups?

- You can spin it using node->executor->spin() in the main instead of rclcpp::spin(node)
But you will have to make the executor public and all the threads will always spin.

- The code will be less simple and require that every node using ServiceClient use a multithreaded executor.

If it's ok for you to have one callback group for every action/service client, I think it will be better to use a single-threaded executor in the Client class and only spin it when you need it.

BriceRenaudeau:
My point of view is that to make a node multithreaded it is better to have a single-threaded node with multiple single-threaded "autonomous" agents rather than a big multithreaded executor spinning all the time.

nav2_util/include/nav2_util/service_client.hpp
BriceRenaudeau: The second argument will automatically add the callback group to the node executor.
Hence, first, the node needs an executor, and second, you need to change the way you spin your node.
```c++
executor.add_node(node);
executor.spin();
```

BriceRenaudeau: Using `MutuallyExclusive`, only the callbacks in the same callback group are mutually exclusive. If you create a callback group per client, there are not blocking each other.

nav2_util/include/nav2_util/service_client.hpp
BriceRenaudeau: I just saw that the `ClearEntirelyCostmapServiceClient` inherit from `ServiceClient` without provided node.
But this Class is never used. Should I remove it?

BriceRenaudeau:
Using one callback group with `Reentrant` param and a multithreaded executor also works, it allows doing a service call in a timer callback.

```c++
static rclcpp::Client<std_srvs::srv::Trigger>::SharedPtr service_client;

void sendRequest(){
  // Send request
  res_received = false;
  auto service_req = std::make_shared<std_srvs::srv::Trigger::Request>();
  auto future = service_client->async_send_request(service_req);

  future.wait();

  if (!future.get()->success){
    RCLCPP_WARN_STREAM(node->get_logger(), "Someting went wrong");
  } else {
    RCLCPP_INFO_STREAM(node->get_logger(), "Everything is fine");
  }
}

int main(int argc, char ** argv){
  rclcpp::init(argc, argv);

  node = std::make_shared<rclcpp::Node>("main_node");
  callback_group = node->create_callback_group(rclcpp::CallbackGroupType::Reentrant, true);

  // Creates service client using the callback group
  service_client = node->create_client<nav2_util::ServiceClient<std_srvs::srv::Trigger>>("my_service", 
rmw_qos_profile_services_default, callback_group);

  // Creates timer using the same callback group
  timer = node->create_wall_timer(std::chrono::seconds(1), &sendRequest);

  rclcpp::spin(node);
  rclcpp::shutdown();
  return 0;
}
```

If you want to use this pattern in nav2_util::ServiceClient, we will need to create a node_callback_group and pass it to the ServiceClient to attach the service_client to it.

nav2_util/include/nav2_util/service_client.hpp
SteveMacenski: 1 second is arbitrary, we don't accept arbitrary magic numbers in ROS2. If there's a default at all, it must be reasonable or provide no default at all

SteveMacenski:
I don't think reentrant callbacks are reasonable for Nav2, we don't want multiple actions / services called at the same time. 

> My point of view is that to make a node multithreaded it is better to have a single-threaded node with multiple single-threaded "autonomous" agents rather than a big multithreaded executor spinning all the time.

I wouldn't normally agree with that, but due to the need for `spin_until_XYZ` for services / actions, I don't think a multi-threaded executor could 100% solve what we need if we need to both `spin` some for subscriber callbacks and also `spin_until_XYZ` to process actions / services so that group couldn't be just `spin`-ed (since ROS2 currently doesn't allow you to spin an already spinning node). 

So I think you're right, we'd need to work with a few single threaded executors instead, one for each callback group on the single node. The "main" node one can `spin` but we need the service / action one to not be spinning so we can `spin_until_XYZ` when calling. But this also depends on if we're talking about a client or the server. I think the server can be multi-threaded executed A-OK. I think its just the clients that couldn't be.

SteveMacenski:
Please pull in master / rebase after https://github.com/ros-planning/navigation2/pull/2224 is merged, CI was temporarily busted

nav2_util/include/nav2_util/service_client.hpp
BriceRenaudeau: Ok, using -1 as default value like `spin_until_future_complete` does works fine.

SteveMacenski (approved):

SteveMacenski:
Thanks for the contribution and good discussion!

doisyg:
Congrats @BriceRenaudeau on your first nav2 PR!

PR #2222, opened 2021-03-08
Fix #2186

### LINKED ISSUE #2186: Start/Goal frame is ignored in planner server
## Bug report

I recently found that while `nav2_planner`  `PlannerServer` accepts goals in frames other than its global frame, this information is subsequently ignored in planners such as `navfn_planner`or `smac_planner`. This leads to strange errors as goal and start frame are silently treated as if they were in the same frame. 

I personally would not expect such a behavior for a service taking a `PoseStamped` as an argument. Is there an obvious reason for this limitation that I am missing?

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Foxy binaries
- Version or commit hash:
  - navigation2 main sources (d3fc28af28df764a153c85434f8b30270e963622)
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
- ensure a valid tf tree exists e.g. `world`->`map`->`base_link`
- initialize navigation2 with `global_frame` set to `map`
- call  `compute_path_to_pose` action with a pose in `world` frame and `navfn_planner`

#### Expected behavior
Generally, if a `PoseStamped` is used in ROS, one silently assumes that the frame_id is handled by the underlying function and transformed to the required frame implicitly as long as the frame is part of a valid tf tree. At least I would expect an error message saying that the frame does not match the expected one as for example it is done in the [planner plugin tutorial](https://navigation.ros.org/plugin_tutorials/docs/writing_new_nav2planner_plugin.html) from nav2 docs.

#### Actual behavior
Both `navfn_planner` and `smac_planner` assume that the goal frame matches the `global_frame` and blindly ignore the `frame_id`.

doisyg:
Fix https://github.com/ros-planning/navigation2/issues/2186
Proper PR for a commit that was proposed as part of https://github.com/ros-planning/navigation2/pull/2179
Implementation can be discussed, but I PRed it as the work as been done and for it not to be forgotten and lost.

There is already a `getCurrentPose` fonction in `nav2_utils` that is using tf to transform the `robot_frame` to a target frame (`global_frame`) and convert this transform into a PoseStamped (`global_pose`) with the proper try/catch of `tf2::*Exception`. This function is used for the `getRobotPose` of `costamp_2d_ros` (which provides the tf_buffer). And this `getRobotPose` of `costamp_2d_ros` is already called in the `planner_server` (which owns a `nav2_costmap_2d::Costmap2DROS`).

All I did was to mimick this architecture. I added a `getPoseInTargetFrame` fonction in `nav2_utils` that is using tf to transform the passed PoseStamped `pose` to the `target_frame` in the same way as the `getCurrentPose` fonction. Then I added `getPoseInGlobalFrame` in `costamp_2d_ros` which is using it (and provides the tf_buffer and the target_frame). And finally `getPoseInGlobalFrame` can be called in the `planner_server` to ensure that the pose received from the goal is transformed in the global_frame (aka, most of the time, `map`). 

No unnecessary transform is done if the pose is already in the correct frame.
Tested on a real robot. TBD: proper unit tests (I believe the ones for robot_utils needs to be augmented, none are fully testing `nav2_util::getCurrentPose`)

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
@@ -146,7 +146,8 @@ public:
    * @param transformed_pose pose transformed
    * @return True if the pose was transformed successfully, false otherwise
    */
-  bool transformPoseToGlobalFrame(const geometry_msgs::msg::PoseStamped & input_pose,
+  bool transformPoseToGlobalFrame(
+    const geometry_msgs::msg::PoseStamped & input_pose,
     geometry_msgs::msg::PoseStamped & transformed_pose);
 
   /** @brief Returns costmap name */
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
@@ -140,6 +140,14 @@ public:
    */
   bool getRobotPose(geometry_msgs::msg::PoseStamped & global_pose);
 
+  /**
+   * @brief Transform the pose in the global frame of the costmap
+   * @param pose will be set to the pose of the robot in the global frame of the costmap
+   * @return True if the pose was transformed successfully, false otherwise
+   */
+
+  bool getPoseInGlobalFrame(geometry_msgs::msg::PoseStamped & pose);
+
   /** @brief Returns costmap name */
   std::string getName() const
   {
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
@@ -141,12 +141,13 @@ public:
   bool getRobotPose(geometry_msgs::msg::PoseStamped & global_pose);
 
   /**
-   * @brief Transform the pose in the global frame of the costmap
-   * @param pose will be set to the pose of the robot in the global frame of the costmap
+   * @brief Transform the input_pose in the global frame of the costmap
+   * @param input_pose pose to be transformed
+   * @param transformed_pose pose transformed
    * @return True if the pose was transformed successfully, false otherwise
    */
-
-  bool getPoseInGlobalFrame(geometry_msgs::msg::PoseStamped & pose);
+  bool transformPoseToGlobalFrame(const geometry_msgs::msg::PoseStamped & input_pose,
+    geometry_msgs::msg::PoseStamped & transformed_pose);
 
   /** @brief Returns costmap name */
   std::string getName() const
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -244,7 +244,8 @@ PlannerServer::computePlan()
     // Changing the start and goal pose frame to the global_frame_ of costmap_ros_ if needed
     geometry_msgs::msg::PoseStamped goal_pose = goal->goal;
     if (!costmap_ros_->transformPoseToGlobalFrame(start, start) ||
-        !costmap_ros_->transformPoseToGlobalFrame(goal->goal, goal_pose)) {
+      !costmap_ros_->transformPoseToGlobalFrame(goal->goal, goal_pose))
+    {
       RCLCPP_WARN(
         get_logger(), "Could not transform the start or goal pose in the costmap frame");
       action_server_->terminate_current();
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -241,11 +241,21 @@ PlannerServer::computePlan()
       return;
     }
 
+    // Copy needed as goal is const
+    geometry_msgs::msg::PoseStamped goal_pose = goal->goal;
+    if (!costmap_ros_->getPoseInGlobalFrame(start) ||
+        !costmap_ros_->getPoseInGlobalFrame(goal_pose)) {
+      RCLCPP_WARN(
+        get_logger(), "Could not transform the start or goal pose in the costmap frame");
+      action_server_->terminate_current();
+      return;
+    }
+
     if (action_server_->is_preempt_requested()) {
       goal = action_server_->accept_pending_goal();
     }
 
-    result->path = getPlan(start, goal->goal, goal->planner_id);
+    result->path = getPlan(start, goal_pose, goal->planner_id);
 
     if (result->path.poses.size() == 0) {
       RCLCPP_WARN(
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -241,7 +241,7 @@ PlannerServer::computePlan()
       return;
     }
 
-    // Copy needed as goal is const
+    // Changing the start and goal pose frame to the global_frame_ of costmap_ros_ if needed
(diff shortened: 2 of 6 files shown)


nav2_util/src/robot_utils.cpp
```diff
@@ -66,4 +66,41 @@ bool getCurrentPose(
...
 }
 
+bool getPoseInTargetFrame(
+  geometry_msgs::msg::PoseStamped & pose,
+  tf2_ros::Buffer & tf_buffer, const std::string target_frame,
+    const double transform_timeout)
+{
+  static rclcpp::Logger logger = rclcpp::get_logger("getPoseInTargetFrame");
+
+  try {
```
SteveMacenski, line 76: This is super redundant to the other function, please refactor them or remove one such that we don't have identical calls to TF

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -140,6 +140,14 @@ class Costmap2DROS : public nav2_util::LifecycleNode
    */
   bool getRobotPose(geometry_msgs::msg::PoseStamped & global_pose);
 
+  /**
+   * @brief Transform the pose in the global frame of the costmap
+   * @param pose will be set to the pose of the robot in the global frame of the costmap
+   * @return True if the pose was transformed successfully, false otherwise
+   */
+
+  bool getPoseInGlobalFrame(geometry_msgs::msg::PoseStamped & pose);
```
SteveMacenski, line 149: This feels weird to me to be in the costmap, why not put this in the planner server? This name also confuses me a bit, you're really just transforming a given pose into the global frame, but the `get` is just odd as if the input is an empty pose, but really its a meaningful pose. Something about this just doesn't quite register right in my mind, do you have that feeling too?

nav2_planner/src/planner_server.cpp
```diff
@@ -241,11 +241,21 @@ PlannerServer::computePlan()
...
+      action_server_->terminate_current();
+      return;
+    }
+
     if (action_server_->is_preempt_requested()) {
       goal = action_server_->accept_pending_goal();
     }
 
-    result->path = getPlan(start, goal->goal, goal->planner_id);
+    result->path = getPlan(start, goal_pose, goal->planner_id);
```
SteveMacenski, line 258: There are tons of placed below still using `goal->goal`

nav2_planner/src/planner_server.cpp
```diff
@@ -241,11 +241,21 @@ PlannerServer::computePlan()
       return;
     }
 
+    // Copy needed as goal is const
```
SteveMacenski, line 244: Remove comment, I think that's sufficiently clear. I think instead a comment above this block explaining its checking / changing the frames to `global_frame_` in case requests came in in another

nav2_util/src/robot_utils.cpp
doisyg: Ok, I will come up with something but the catch / RCLCPP_ERRORs will be common to both (and the logger too)

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
doisyg: > This feels weird to me to be in the costmap, why not put this in the planner server?

In order to avoid to create a tf_buffer and a tf_listener in the planner server just for this. And since the frame in which we want the pose to be relative to is the global_frame of the costmap.
Also because I followed the logic of `getRobotPose`



> Something about this just doesn't quite register right in my mind, do you have that feeling too?

I do too. Let's find a better name, few propositions but I cannot find a satisfactory one:  `transformPoseInGlobalFrame`, `referencePoseRelativeToGlobalFrame`, `putPoseInCostmapFrame` ?

nav2_planner/src/planner_server.cpp
```diff
@@ -241,26 +241,36 @@ PlannerServer::computePlan()
...
       action_server_->terminate_current();
       return;
     }
 
     RCLCPP_DEBUG(
       get_logger(),
       "Found valid path of size %lu to (%.2f, %.2f)",
-      result->path.poses.size(), goal->goal.pose.position.x,
-      goal->goal.pose.position.y);
+      result->path.poses.size(), goal_pose.pose.position.x,
```
SteveMacenski, line 272: Still missing one https://github.com/ros-planning/navigation2/blob/456b6fd4af5c8697d7b37faf47f8535537149a8c/nav2_planner/src/planner_server.cpp#L292-L293

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
SteveMacenski: https://github.com/ros-planning/navigation2/blob/main/nav2_planner/include/nav2_planner/planner_server.hpp#L135 the planner server already has the buffer/listener from the costmap itself. I'd probably be OK with this if the name / API of this function felt "more right" to us, and I know that's a really wishy-washy thing to say :laughing: 

The things that are really throwing me for a loop
- the `get` word when you're actually inputting a valid non-identity pose, so its not just "get" but "transform"ing
- The fact that the input and output are the same variable and remapped over each other

I think `transformPoseToGlobalFrame` would do.

nav2_planner/src/planner_server.cpp
doisyg: No, these lines are outside the block that defines goal_pose

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
doisyg: Ok, I changed the names for `transformPoseToGlobalFrame` and `transformPoseInTargetFrame` + ` const input_pose` and `transformed_pose` parameter pattern.
If you are ok with that, let's leave the discussion about the best place to call the nav2_utils fonctions (ie. planner_server or costmap) for another day.

nav2_util/src/robot_utils.cpp
```diff
@@ -28,39 +28,47 @@ bool getCurrentPose(
...
-  robot_pose.header.frame_id = robot_frame;
-  robot_pose.header.stamp = rclcpp::Time();
+  global_pose.header.frame_id = robot_frame;
+  global_pose.header.stamp = rclcpp::Time();
+
+  return transformPoseInTargetFrame(
+    global_pose, global_pose, tf_buffer, global_frame, transform_timeout);
+}
+
+bool transformPoseInTargetFrame(
```
SteveMacenski, line 39: With the new API for `transformPoseToGlobalFrame`, wouldn't you only need the original function that was here?

SteveMacenski (commented):
We can leave in costmap_2d with that change in name / API. 

One last question, not sure we need any of the robot_utils changes anymore with that change in the API, no? 

Just need a quick validation that this works on both the start / goal side and I'm happy to merge

nav2_util/src/robot_utils.cpp
doisyg: First, yes, tested and working on a real robot.

Second, I am not sure I am following. I see many different ways of doing this, but I don't see how to do it without keeping both `nav2_util::transformPoseInTargetFrame` and `nav2_util::getCurrentPose`.

SteveMacenski (approved):
Merge when CI passes

Issue #2197, opened 2021-02-19
[BT plugin node] Only trigger underlying subtree / node once time 


SteveMacenski:
In case you want to do something like "no replanning" so you place the planner BT node under this "only once" so that its skipped for all over traversals of the tree during a particular run.

Opletts:
I'd like to give this a shot, if that's alright?

SteveMacenski:
That would be more than alright! :smile:

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/2236 completes, @Opletts apologies, I needed this done for a demo :(

PR #2236, opened 2021-03-11
new behavior tree node for triggering just once

SteveMacenski:
Per https://github.com/ros-planning/navigation2/issues/2197

Requires
- testing DONE
- migration guide, nav2 plugins list, and param guide updates DONE


Timple:
Out of curiosity, why is this implemented as a separate node? Not the standard bt node with a zero frequency?

SteveMacenski:
I don't understand what you're suggesting as the alternative. You mean like the RateController with rate set to `0`? That would be semantically weird and not really using a "rate controller" because you're just triggering once. There's lots of reasons it would make it more readable and designer-friendly like this.

Timple:
> You mean like the RateController with rate set to 0

That was what I meant. 

> semantically weird

That is true, and that is important. I thought it would save you quite some boilerplate code for a new Controller. But again, curious, not opposed to the implementation.

Issue #2240, opened 2021-03-12
Revert / fix SLAM launch file configuration pop/push failures


SteveMacenski:
CC @v-lopez

v-lopez:
Could you please share some details on what's the failure?

SteveMacenski:
I don't recall off hand, but it should be pretty obvious to you if using `slam:=True` option on the TB3 simulation launch file. The stack doesn't come up.

v-lopez:
Yes, I've seen now.

It seems that Pop doesn't pop the configuration, or some tasks are run concurrently and nodes such as the controller_server are started with the param file of `slam_toolbox`.

Looking into it, I apologise for the incovenience.

v-lopez:
I believe it's a bug in `launch`. Not sure if we can fix it right now.

The workaround I'd propose is to change the `params-file` argument name to something that doesn't clash, like `slam-toolbox-params-file` or `slam-params-file`.

But it's going to get hard to keep all LaunchConfiguration names unique with the biggest launch files.

What do you think about this?

SteveMacenski:
so a different param input file for slam vs everything else? Or just the variable name in a single launch file?

v-lopez:
Just the variable name in `online_sync_launch.py` and in `online_async_launch.py`.

So if the nav2_bringup param_file has  parameters for slam_toolbox, I will pass `params-file` from nav2_bringup as `slam-params-file` for slam_toolbox. If it doesn't have such parameters, I will not pass it, and it will not be picked up automatically (which is what happens now, and the reason why the Push, Unset, Pop) was needed.

SteveMacenski:
Oh, that would be totally fine

v-lopez:
Created PR on ros2/main and foxy branches.

Going to sleep now, I'll address any issues tomorrow if there are.

SteveMacenski:
@v-lopez did you by chance file a ticket about these launch issues we ran into (or was that already reported)?

v-lopez:
> @v-lopez did you by chance file a ticket about these launch issues we ran into (or was that already reported)?
https://github.com/ros2/launch/issues/313

Here it is discussed. The current state as described by wjwood:
> We would like to get back to this with launch. Currently an included launch description has access to all configurations of the launch file that included it.

So I believe their intended approach is to:
1. Add a way to explicitely promote arguments from an included launch files to expose them at the current launch file.
2. Disable automatic promotion (current behavior, which is what is causing these issues)

PR #2242, opened 2021-03-13
Use different param name for slam_toolbox param_file

v-lopez:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2240|

---

* Use renamed slam_toolbox param-file

### DIFF
--- a/nav2_bringup/bringup/launch/slam_launch.py
+++ b/nav2_bringup/bringup/launch/slam_launch.py
@@ -17,10 +17,8 @@ import os
 from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import (DeclareLaunchArgument, IncludeLaunchDescription,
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
Did you test that this works for both cases?

v-lopez:
Yes, with default param file it will load the slam_toolbox parameters.

If I copy those contents into the nav2_bringup params, it loads the nav2_bringup params file.

SteveMacenski:
@v-lopez main is A-OK for this and SLAM Toolbox. Still the other one I need to think more about with foxy-devel given the API change

SteveMacenski:
@v-lopez I've managed to get my wires super crossed, were we talking about multiple goal checker plugins? I believe so, but for the life of me, I can't seem to find the PR / issue / discussion we were having on it.

PR #2243, opened 2021-03-13
Use different param name for slam_toolbox param_file

v-lopez:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2240|

---

* Use renamed slam_toolbox param-file

### DIFF
--- a/nav2_bringup/bringup/launch/slam_launch.py
+++ b/nav2_bringup/bringup/launch/slam_launch.py
@@ -17,10 +17,8 @@ import os
 from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import (DeclareLaunchArgument, IncludeLaunchDescription,
-                            PopLaunchConfigurations, PushLaunchConfigurations,
-                            UnsetLaunchConfiguration)
-from launch.conditions import UnlessCondition
+from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription
+from launch.conditions import IfCondition, UnlessCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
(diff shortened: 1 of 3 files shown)


SteveMacenski (approved):

SteveMacenski:
I'm not sure if we should have merged in the foxy one on slam toolbox, that changes the API for params file for anyone with scripts. What do you think? ros2/main is fine, but I'm concerned about foxy actually that I should revert that merge.

v-lopez:
Well before there was no param, and now there is one that should not conflict with any.

I believe the only people that might be affected are those that had slam_toolbox parameters in their nav2 bringup parameter file. Which did not make sense since they were not being used.

SteveMacenski:
https://github.com/SteveMacenski/slam_toolbox/pull/364/files what do you mean? there was a parameter that was renamed

v-lopez:
Yes sorry, the param that we added last week was just renamed, but I think it was after the last sync to foxy.

So the rename would only affect people who use slam_toolbox from source and started using the parameter in the last 8 days.

SteveMacenski:
OK, reverting that foxy-devel PR https://github.com/SteveMacenski/slam_toolbox/pull/365 but we're leaving / merging the `ros2` ones that will roll into Galactic and anything newer. We don't want to break behavior in the middle of a distribution

SteveMacenski:
I think that means we can close this PR too since this is `foxy` specific and the slam toolbox params won't support changing that name in the middle of a released distro? Or is there a different way around this for just foxy to enable this behavior?

v-lopez:
I don't think I follow.

We added a param 10 days ago and we have to remove it or rename it to fix the issue you reported.

Removing it breaks API as much as renaming it. But renaming with the changes I made will allow us to keep the functionality.

SteveMacenski:
https://github.com/SteveMacenski/slam_toolbox/pull/365 I did remove it

SteveMacenski:
#2240 was never merged into foxy, only main. None of those push/pop things were ever put into foxy to break anything. 

I don't understand your point. We added something to `main` that broke default behavior for Slam toolbox. You fixed it and it was merged. But you also fixed something that wasn't broken in foxy because there was never a foxy Nav2 patch for the issue. But we merged it into slam toolbox for some reason so I reverted it.

So foxy, if we close this PR, will be as if nothing happened at all. Main in Nav2/ROS2 in SLAM Toolbox will contain the behavior you were looking to add with setting parameters via the main nav2 file. Foxy will not.

v-lopez:
You are right, I was mixing the push/pop issue that affects the nav2 nodes, with the issue that using `params_file` as a LaunchConfiguration in both nav2_bringup and slam_toolbox causes.

The whole purpose of the Push/Pop was to prevent the nav2 `params_file` from being automatically passed to slam_toolbox. In turn, the push/pop caused #2240.

But in foxy the `params_file` provided to nav2_bringup is being forwarded automatically to slam_toolbox, preventing the loading of the default parameter files.

That's why I was saying that we should rename `params_file` in slam_toolbox via https://github.com/SteveMacenski/slam_toolbox/pull/364 or remove the param altogether, but I was wrongly referencing #2240.

SteveMacenski:
So can we close this since this is for `foxy-devel`?

v-lopez:
> So can we close this since this is for `foxy-devel`? I'm not disagreeing with the issue, but I'm saying that we **cannot** rename a param like that mid-release. We would need to find another way around it for Foxy. Hence, closing this PR is what I'm asking if you're OK with.

Ok, let's close this then.

For fixing foxy `slam_toolbox` bug we could search if the provided param file has `slam_toolbox`parameters, and if it doesn't, ignore the parameter and load the default one. It's hacky but it'd be backwards compatible.

SteveMacenski:
I don't think there's a bug in Slam toolbox, there's a bug in launch that impacts Nav2's usage of slam toolbox

SteveMacenski:
Sure, we could do that.

PR #2245, opened 2021-03-13
Fix SingleTrigger BT node

naiveHobo:

---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Gazebo TurtleBot3 simulation |

---

* Fixed SingleTrigger BT node tick function, earlier is failed to execute its child node. 

#### For Maintainers: 


nav2_behavior_tree/plugins/decorator/single_trigger_node.cpp
```diff
@@ -37,11 +37,24 @@ BT::NodeStatus SingleTrigger::tick()
...
+
+    switch (child_state) {
+      case BT::NodeStatus::RUNNING:
+        return BT::NodeStatus::RUNNING;
+
+      case BT::NodeStatus::SUCCESS:
+        first_time_ = false;
+        return BT::NodeStatus::SUCCESS;
+
+      case BT::NodeStatus::FAILURE:
```
SteveMacenski, line 50: I think even failure `first_time_` should be reset to `false`, since this node implies we only trigger this child once, regardless of the return code

SteveMacenski: e.g. mimic the `default` below

SteveMacenski (commented):
Thanks for the fix!

SteveMacenski:
dont worry about CI, its being fixed by a new docker sync tomorrow

nav2_behavior_tree/plugins/decorator/single_trigger_node.cpp
naiveHobo: The switch case moves to the default case when case is `BT::NodeStatus::FAILURE`, no need to mimic the default again.

SteveMacenski: https://stackoverflow.com/questions/38218812/c-empty-case-before-default-useless

I tend to agree with the first commenter, this looks incomplete and has some implicit flow requiring the reader to me more than average knowledgeable about how `case/switch` works in C++ and easily scanned over incorrectly. There's no functional difference between these 2 sets of code, but sets explicit boundaries about what's happening.

naiveHobo: Yup, makes sense

SteveMacenski (approved):

SteveMacenski:
@naiveHobo pull in master to fix CI

SteveMacenski:
Thanks!

PR #2246, opened 2021-03-14
Improve NavigationToPose action feedback

deepaktalwardt:
For complete discussion please see PR https://github.com/ros-planning/navigation2/pull/2194

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |#2154 |
| Primary OS tested on |Ubuntu 20.04 |

---

* Improves feedback of the `NavigateToPose` action server by adding:
    - Better distance to goal estimate: by integrating distance between planned path poses.
    - Estimated time remaining: by dividing the distance remaining by current robot linear speed (using smoothed odometry).
* Updates the feedback in `.action` file for `NavigateToPose` action server. 
* Adds a utility to the `nav2_util::geometry_utils` to calculate path length.
* Adds simple unit testing for this function.

* Documentation for `.action` needs to be updated.

---

* Currently the time remaining largely ignores the target yaw of the robot. It may be the case that the robot needs more time to rotate to the correct heading.


SteveMacenski (approved):

SteveMacenski:
https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1251 build issues are from a bug fixed already, waiting for a new docker image to be released. I'll retrigger this tomorrow and it should go

SteveMacenski:
@deepaktalwardt pull in master to fix CI

deepaktalwardt:
@SteveMacenski Sorry for the delay! Hopefully all works well now :)

SteveMacenski:
@deepaktalwardt it looks like your branch has some legit compiling issues (not CI anymore)

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -116,6 +117,10 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
   blackboard->set<bool>("initial_pose_received", false);  // NOLINT
   blackboard->set<int>("number_recoveries", 0);  // NOLINT
 
+  // Odometry smoother object for getting current speed
+  auto node_handle = blackboard->get<rclcpp::Node::SharedPtr>("node");
```
SteveMacenski, line 121: Please use the BT Navigator lifecycle node, this node is reserved for the behavior tree nodes

SteveMacenski (changes requested):

SteveMacenski:
@deepaktalwardt, all of the system level tests are failing. This indicates to me something about this breaks the BT navigator if none of the system can operate with these changes. Please run this locally and debug the issue or ask any questions you might have

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -116,6 +117,9 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
   blackboard->set<bool>("initial_pose_received", false);  // NOLINT
   blackboard->set<int>("number_recoveries", 0);  // NOLINT
 
+  // Odometry smoother object for getting current speed
+  odom_smoother_ = std::make_unique<nav2_util::OdomSmoother>(rclcpp_node_, 0.3);
```
naiveHobo, line 121: Probably a better idea to pass the `bt_navigator` node shared_ptr using `shared_from_this()`. I can't see an `rclcpp_node_` object defined anywhere.

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -28,7 +29,7 @@ namespace nav2_bt_navigator
 {
 
 BtNavigator::BtNavigator()
-: nav2_util::LifecycleNode("bt_navigator", "", false),
+: nav2_util::LifecycleNode("bt_navigator", "", true),
```
naiveHobo, line 32: Not sure why this was changed, can't think of a reason why `use_rclcpp_node` would be set to true while initializing `nav2_util::LifecycleNode`. Can you explain?

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -206,22 +210,59 @@ void
...
+  // Calculate distance on the path
+  double distance_remaining =
+    nav2_util::geometry_utils::calculate_path_length(current_path, find_closest_pose_idx());
+
+  // Default value for time remaining
+  rclcpp::Duration estimated_time_remaining = rclcpp::Duration::from_seconds(0.0);
+
+  // Get current speed
+  geometry_msgs::msg::Twist current_odom = odom_smoother_->getTwist();
+  double current_linear_speed = hypot(current_odom.linear.x, current_odom.linear.y);
```
naiveHobo, line 252: Use `std::hypot` with the namespace for clarity

naiveHobo (changes requested):

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: I have tried that and it doesn't work as the `OdomSmoother` class only accepts a reference to  `rclcpp::Node::WeakPtr` and not a `LifecycleNode`. The `rclcpp_node_` object is defined and initialized in the parent class of `BtNavigator` (`nav2_util::LifecycleNode`) here: https://github.com/ros-planning/navigation2/blob/3bf1f7a93d740c54b4c7ef7fc6dd914175bf46be/nav2_util/src/lifecycle_node.cpp#L62. And since it is a protected member, it should be available for this object to access. And as stated in the header file here https://github.com/ros-planning/navigation2/blob/3bf1f7a93d740c54b4c7ef7fc6dd914175bf46be/nav2_util/include/nav2_util/lifecycle_node.hpp#L194 that seems to be the intent behind this member variable.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: As I mentioned in the previous comment, changing this to true makes available the `rclcpp_node_` protected member variable that I was then passing to the `OdomSmoother`.

nav2_bt_navigator/src/bt_navigator.cpp
deepaktalwardt: Agreed! thanks for the suggestion.

SteveMacenski:
As we talked in slack, just add a new constructor for the lifecycle node to the odom smoother or template it so you can use the lifecycle node. It would be very inefficient to spin up a new node just for this.

SteveMacenski (commented):
https://app.circleci.com/pipelines/github/ros-planning/navigation2/4944/workflows/cd88ba0a-d611-4eb2-b790-e51fd1acfdde/jobs/18276 all of the major tests are still failing, please, test locally and debug the issue before pushing changes

> Action server failed while executing action callback: "Any::cast failed because it is empty"

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -206,22 +210,59 @@ void
...
   auto blackboard = bt_action_server_->getBlackboard();
 
-  geometry_msgs::msg::PoseStamped goal_pose;
-  blackboard->get(goal_blackboard_id_, goal_pose);
-
-  feedback_msg->distance_remaining = nav2_util::geometry_utils::euclidean_distance(
-    feedback_msg->current_pose.pose, goal_pose.pose);
+  // Get current path points
+  nav_msgs::msg::Path current_path;
+  blackboard->get("path", current_path);
```
SteveMacenski, line 225: I think you didn't set the template get for path, so maybe couldn't imply the cast. Also I don't see path in the https://github.com/ros-planning/navigation2/blob/3b36c29989b429a60c5e8161aa4cd57637b81707/nav2_behavior_tree/include/nav2_behavior_tree/bt_conversions.hpp string conversion list

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -206,19 +210,61 @@ void
...
+    double current_linear_speed = std::hypot(current_odom.linear.x, current_odom.linear.y);
+
+    // Calculate estimated time taken to goal if speed is higher than 1cm/s
+    // and at least 10cm to go
+    if ((std::abs(current_linear_speed) > 0.01) && (distance_remaining > 0.1)) {
+      estimated_time_remaining =
+        rclcpp::Duration::from_seconds(distance_remaining / std::abs(current_linear_speed));
+    }
+
+    feedback_msg->current_pose = current_pose;
```
SteveMacenski, line 262: I think setting the pose should be outside of the try / catch. It should be valid regardless if you get `path`

SteveMacenski:
Note: This PR will be in conflict with #2271, so I'd like to get this PR in before mine so I have to rectify it rather than you.

SteveMacenski (approved):

SteveMacenski:
I'm rerunning CI, need at least one of the test jobs to successfully run

SteveMacenski:
Thanks for this contribution! Makes the feedback WAY more useful!

deepaktalwardt:
> Thanks for this contribution! Makes the feedback WAY more useful!

Thanks for all your help @SteveMacenski! I'm glad to be able to finally get this in :)

Issue #2223, opened 2021-03-08
Give controllers pointer to goal checker


SteveMacenski:
Thus it can use that information locally as well if that state information is useful.

pabloinigoblasco:
I experienced cases where that is convenient. 
I have first to say that isolating the concept of goal checker looks correct to me (for the typical cases).

However, I think it is not always possible to decouple the controller and the goal checker. 

For example, the controller may contain semantic about how well it is important to follow the path. Maybe you have a path with a loop and pass twice on the same state. 

It looks to me like the simple_goal_checker is not able to identify that.

For example. In some of our examples we have a "backward-undo planner" that must exactly the same motion it did forward at the inverse direction. You must totally complete the path inversely. Sometimes the path contains loops with some state (x,y,theta) repeated.

We should not stop in that case when we reach the "goal state" first.

SteveMacenski:
> For some ways of motion, It looks to me like the simple_goal_checker is not able to identify that and handle this situation.

It is afterall the _simple_ goal checker, but agreed potentially that could be an addition. Though if in a loop, I'd hope that you didn't end up in the same XY at the acceptable theta range to trigger the `true` case - else hey - you didn't even need to do the loop at all, the initial one was good enough by your stated parameters :wink: 

>  Sometimes the path contains loops with some repeated state (x,y,theta). 

That... would be fine? Its only a problem if you **goal** state you go into with the same-ish (x,y,theta) as an intermediate point on the path. If you intersect the same (x,y,theta) as other points on the path, that doesn't matter.

pabloinigoblasco:
> In this ticket, I'm suggesting the controllers get a pointer to the current goal checker, not that the goal checker gets access to the current controller. 

I see. 

This case could be even more clear. Lets say, for example that the controller must predict trajectories to evaluate the quality of the trajectories, considering the final state and making sure that the goal is reached but not left (because of the dynamics).

> Though the goal checker doesn't have access to the path right now, that would be a PR I'd happily merge. With the path, current pose, and goal pose, you should trivially be able to detect and not return true on a loop without needing any information about the state of the controller at all in a new GoalChecker plugin.

I agree, passing the path is more convenient, it should not be hard to do. Let my have a try.

SteveMacenski:
> This case could be even more clear. Lets say, for example that the controller must predict trajectories to evaluate the quality of the trajectories, considering the final state and making sure that the goal is reached but not left (because of the dynamics).

[...]

PR #2247, opened 2021-03-16
Controller plugins include goal checker when computing commands

SteveMacenski:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2223 |
| Robotic platform tested on | Sim TB3 |

---

* Added a goal checker pointer to the controllers so that they may access its information while computing commands
* If it returns true, the controller could use that information to halt 
* Add function to goal checkers so that a plugin can access more particular information, like the XY vs yaw tolerance met to rotate in place, or robot "done" speed requirements so that it can dissect more information. 
* Use these getTolerance parameters in RPP rather than any parameters duplicated across it

* Add to documentation migration guide this API changed
* Remove parameters duplicated now removed from RPP in config page

---

? N/A ?

#### For Maintainers: 


pabloinigoblasco:
Changes look good to me.

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/2252 superseded

PR #2248, opened 2021-03-16
[SimpleGoalChecker] Make parameters changeable at runtime

doisyg:
I need to change the **SimpleGoalChecker** params at runtime, so here is a PR activating this feature in a similar fashion as https://github.com/ros-planning/navigation2/pull/2181
Working and tested on my simulation.
If there is an interest I can add to this PR changeable parameters for the 2 others `nav2_controller` plugins: **SimpleProgressChecker** and **StoppedGoalChecker**

### DIFF
--- a/nav2_controller/plugins/simple_goal_checker.cpp
+++ b/nav2_controller/plugins/simple_goal_checker.cpp
@@ -43,6 +43,9 @@
 #include "tf2/utils.h"
 #pragma GCC diagnostic pop
 
+using rcl_interfaces::msg::ParameterType;
+using std::placeholders::_1;
+
 namespace nav2_controller
 {
 
@@ -59,6 +62,7 @@ void SimpleGoalChecker::initialize(
   const rclcpp_lifecycle::LifecycleNode::WeakPtr & parent,
   const std::string & plugin_name)
 {
+  plugin_name_ = plugin_name;
   auto node = parent.lock();
 
   nav2_util::declare_parameter_if_not_declared(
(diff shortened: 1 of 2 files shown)


nav2_controller/plugins/simple_goal_checker.cpp
```diff
@@ -105,6 +119,30 @@ bool SimpleGoalChecker::isGoalReached(
...
+SimpleGoalChecker::on_parameter_event_callback(
+  const rcl_interfaces::msg::ParameterEvent::SharedPtr event)
+{
+  for (auto & changed_parameter : event->changed_parameters) {
+    const auto & type = changed_parameter.value.type;
+    const auto & name = changed_parameter.name;
+    const auto & value = changed_parameter.value;
+
+    if (type == ParameterType::PARAMETER_DOUBLE) {
+      if (name == plugin_name_ + ".xy_goal_tolerance") {
```
SteveMacenski, line 132: Wouldn't this be more than just name + parameter? Wouldn't it also be in the controller_server namespace at minimum? This doesn't have its own node. Did you print out the `name` to make sure this full path is valid?

SteveMacenski:
@doisyg pull in master to fix CI

nav2_controller/plugins/simple_goal_checker.cpp
doisyg: Yes, same logic as for NavFn I believe.
If I do a print in the code:
```
[controller_server-7] [WARN] [SimpleGoalChecker]: name: goal_checker.xy_goal_tolerance
[controller_server-7] [WARN] [SimpleGoalChecker]: plugin_name_: goal_checker
```

doisyg:
Should we investigate if it would be useful to use the newly merged ParameterEventsSubscriber class (https://github.com/ros2/rclcpp/pull/829) for these parameters change use cases ?

nav2_controller/plugins/simple_goal_checker.cpp
SteveMacenski: Ok, works for me then, just verifying

SteveMacenski:
Sure! That would make things much more readable than these single massive callback functions of every parameter

doisyg:
@anaelle-sw, once this is merged, can you try to modify it to use ParameterEventsSubscriber and do a new PR on top of it ?

SteveMacenski:
@doisyg I cancelled the debug workflow because I needed the CI time elsewhere since you already had the release build / test pass.

SteveMacenski (approved):

PR #2249, opened 2021-03-16
Planner selector

pabloinigoblasco:
---

## Basic Info

PR of the planner selector to be integrated in navigation2 main branch. With this BTNode we use a subscriber to enable the ability of selecting the current planner for the navigation.

| Info |  |
| ------ | ----------- |
| Robotic platform tested on | (ROS Rolling and Gazebo simulation turtlebot waffle) |

---

* reading from topics to select the planner
* support for default planner (as static input port)
* support to setting the name of the topic to support multiple planner selectors.

---

#### For Maintainers: 


SteveMacenski:
@pabloinigoblasco please copy-paste an un-modified PR template and fill it in. There's checklists and items there important for my own sanity when reviewing PRs to make sure things don't drop through the cracks and that you're prompted about things you may have forgotten to do.

Though I do understand this is just a WIP first go

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -111,6 +111,9 @@ list(APPEND plugin_libs nav2_truncate_path_action_bt_node)
 add_library(nav2_goal_updater_node_bt_node SHARED plugins/decorator/goal_updater_node.cpp)
```
SteveMacenski, line 111: See this PR, this needs to be added to all of the param files / default BT plugin lists https://github.com/ros-planning/navigation2/pull/2236/files

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "std_msgs/msg/string.hpp"
+
+#include "behaviortree_cpp_v3/decorator_node.h"
+
+#include "rclcpp/rclcpp.hpp"
+
+namespace nav2_behavior_tree
+{
+
+class PlannerSelector : public BT::DecoratorNode
```
SteveMacenski, line 31: @naiveHobo does this make most sense as a decorator or an action node?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+private:
+  BT::NodeStatus tick() override;
+
+  void callback_planner_select(const std_msgs::msg::String::SharedPtr msg);
+
+  rclcpp::Subscription<std_msgs::msg::String>::SharedPtr planner_selector_sub_;
+
+  std_msgs::msg::String last_selected_planner_;
```
SteveMacenski, line 58: Why is this not just a `std::string`?

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
```diff
@@ -0,0 +1,73 @@
...
+  const BT::NodeConfiguration & conf)
+: BT::DecoratorNode(name, conf)
+{
+  auto node = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+
+  last_selected_planner_.data = "";
+
+  std::string planner_selector_topic;
+
+  node->get_parameter_or<std::string>(
```
SteveMacenski, line 42: If this wasn't declared, then it will never be able to `get` it. Also, I think this should be a port so that if you had multiple planner selectors, they could each have set their own individual topic (vs this would make a single param impact all of them)

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
```diff
@@ -0,0 +1,73 @@
...
+    planner_selector_topic, 10, std::bind(&PlannerSelector::callback_planner_select, this, _1));
+}
+
+inline BT::NodeStatus PlannerSelector::tick()
+{
+  if (last_selected_planner_.data.empty()) {
+    getInput("default_planner", last_selected_planner_.data);
+  }
+
+  setOutput("selected_planner", last_selected_planner_);
```
SteveMacenski, line 56: I think this should just be a `std::string`, I don't think this needs to at all relate to a ROS mesage

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  return child_node_->executeTick();
+}
+
+void
+PlannerSelector::callback_planner_select(const std_msgs::msg::String::SharedPtr msg)
+{
+  last_selected_planner_ = *msg;
```
SteveMacenski, line 64: convert to just a normal string here

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+
+using std::placeholders::_1;
+
+PlannerSelector::PlannerSelector(
+  const std::string & name,
+  const BT::NodeConfiguration & conf)
+: BT::DecoratorNode(name, conf)
+{
+  auto node = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
```
SteveMacenski, line 36: Should you spin_some in the main processing function so the node can process callbacks?

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -111,6 +111,9 @@ list(APPEND plugin_libs nav2_truncate_path_action_bt_node)
...
 list(APPEND plugin_libs nav2_goal_updater_node_bt_node)
 
+add_library(nav2_planner_selector_bt_node SHARED plugins/decorator/planner_selector_node.cpp)
```
SteveMacenski, line 114: Update readme table to include this BT node

SteveMacenski (changes requested):
LGTM from a high level, just some details to work out

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
naiveHobo: I think making it an action node makes more sense. As a decorator, it's not really affecting *how* its child node is ticked. The main task this node is supposed to perform (only setting something on the blackboard) is similar to the [SetBlackboard node](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/include/behaviortree_cpp_v3/actions/set_blackboard_node.h) which is also an action node.

SteveMacenski: Yeah, I know he was following the template of the goal updater node, which I also think is a bit odd to be a decorator

pabloinigoblasco: I asked about the question of the decorator or the action node in [the discussion issue 147](https://github.com/ros-planning/navigation.ros.org/issues/147#issuecomment-799728405). It is better think before code :-).

I will change it to an bt-action-node.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
pabloinigoblasco: I guess you refer to the internal last_selected_planner_ variable.
It is okay. I will change it. 
It was just following the pattern "lastMessage_" from a topic.

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
pabloinigoblasco: I do not know exactly the origin of node, I think we do not need to spin it in this class.

My impression was that was the main node of the bt_navigator_node process.
I think it is being "spinned" somewhere, but I can check.

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
pabloinigoblasco: One of the possible solution could be adding a constant string-input port to declare explicitly the topic name of the planner selector in the bt-xml file.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
SteveMacenski: Sorry, that got lost in the details I was thinking about in that comment! I also don't think there's an objectively right answer to this question

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
SteveMacenski: Yeah, so if we make an input port a string for topic name, with a logical default, then that would work and replace use of node parameters.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
pabloinigoblasco: No problem, please, just confirm me if this must be done. y/N?

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
SteveMacenski: There's no spinning of this node anywhere outside of the behavior tree, its only spun for `spin_until_XYZ_complete` and `spin_some` for actions in the FollowPath / ComputePathToPose nodes. So without spinning the node a little here, you're relying on those 2 nodes to be in the BT and processing to use this feature. 

That might be OK? I don't think it breaks anything to not include it here, assuming you have those other 2 nodes processed regularly in your BT.

pabloinigoblasco: Isnt that node spinned here?

https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L157

pabloinigoblasco: Then I think we should some call spin_some into the tick function before setting the selected_planner output port value.

I think would guarantee that the node callbacks are handled at least once each time the BT passes through the planner_selector and at least the last ros message is handled.

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
pabloinigoblasco: I think this also means that the topic subscription must be deferred to the first tick call where we have access to the values input ports values. (not a big deal)

This opens also the question of a dynamic topic name (I would keep it simple and would just use the first tick update)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
SteveMacenski: Yes, make it an action node, but you don't need the https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L30 BTActionNode object, since this isn't also a ROS action. I think you just need to use the BT.CPP SyncActionNode (or even `ActionNodeBase`). A simple example is [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/include/behaviortree_cpp_v3/actions/set_blackboard_node.h)

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
SteveMacenski: https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/action/back_up_action.cpp#L30 we rely on getting blackboard variables in the constructor before a `tick` when the tree is constructed. So its not waiting until the first tick to access them, yeah? 

I think we can assume static topic names safely, every other BT node doesn't assume that changes.

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
SteveMacenski: Agreed

SteveMacenski: https://github.com/ros-planning/navigation2/issues/2251

pabloinigoblasco: Having callback_group_executor_ sounds good.

This issue is related with my original question about how the nav-bt were supposed to interact with the external ros network receiving custom information/commands. So it looks these are the first times this sort of issues are being faced. It is an interesting issue, I will keep an eye, but no doubt there must be an optimal solution (maybe that one you say).

In this case, do you think it would be a problem if one message from other btnode is handled?
For example? would be a problem if a new goal message callback is handled? I would say no.

I think we can have an initial implementation based on rclcpp::spin_some and later move to a more advanced approach. What do you think?

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
pabloinigoblasco: Sounds good.
I will do that in the way you say.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/planner_selector_node.hpp
pabloinigoblasco: Correct, I see BTActionNodes are for action clients, so that does not apply our case.
I will check those base nodes you say.

nav2_behavior_tree/plugins/decorator/planner_selector_node.cpp
SteveMacenski: >  For example, would be a problem if a new goal message callback is handled? I would say no.

No, I don't think there's likely to be any "real" problem with them all processed together lump sum. If a message from the goal updater or another subscriber of "reasonable" update rate is processed when `spin_some`, I don't think there would be any noticable difference in the system.

I think the separation using the different executors though is a way to verifiably guarantee there can't and won't be an issue, even in some odd edge case (like if someone implements a stupidly long service server in the BT node for some reason or decides to subscribe to a 10kHz topic that is washing out everything else).

pabloinigoblasco: I agree, lets use the power of ROS2.

SteveMacenski: @pabloinigoblasco lets not do it in this PR. It's come to my attention that the executors might be creating a new context remapping to a new DDS participant. This would not be good if we had each BT node creating a new participant to process separately. Some more research has to be done to figure out if this is workable before we jump into it.

pabloinigoblasco:
I updated around the PR description.
Note: I am about to do a new commit that implements the review points.

pabloinigoblasco:
What about this new version?

nav2_behavior_tree/test/plugins/action/test_planner_selector_node.cpp
```diff
@@ -0,0 +1,135 @@
...
+std::shared_ptr<BT::Tree> PlannerSelectorTestFixture::tree_ = nullptr;
+
+TEST_F(PlannerSelectorTestFixture, test_tick)
+{
+  // create tree
+  std::string xml_txt =
+    R"(
+      <root main_tree_to_execute = "MainTree" >
+        <BehaviorTree ID="MainTree">
+          <PlannerSelector selected_planner="{selected_planner}" default_planner="GridBased" topic_name="planner_selector_custom_topic_name"/>
```
SteveMacenski, line 83: Another test to quickly show that it publishes to the right default when not specified would be nice to have

(basically just copy paste a simplified version of this test with no `topic_name` in the XML string and check that it still changes)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<std::string>(
+        "default_planner",
+        "the default planner to use if there is not any external topic message received."),
+
+      BT::InputPort<std::string>(
+        "topic_name",
```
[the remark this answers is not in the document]
SteveMacenski: I see you have the default set in the BT node itself. @naiveHobo any opinion on good or bad either way? I suppose I don't really care myself  too much

SteveMacenski (commented):
Overall LGTM, just a question for Sarthak and a test. Also the readme table update to include this new Bt node + the website update to include it I mention in above comments with an example.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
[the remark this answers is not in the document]
naiveHobo: Would be better to provide a default value in the port itself following behaviortree.cpp standards

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include "std_msgs/msg/string.hpp"
+
+#include "behaviortree_cpp_v3/action_node.h"
+
+#include "rclcpp/rclcpp.hpp"
+
+namespace nav2_behavior_tree
+{
+
+class PlannerSelector : public BT::SyncActionNode
```
naiveHobo, line 31: Need to add doxygen to this header like in #2254

nav2_behavior_tree/CMakeLists.txt
pabloinigoblasco: Should I add a new section to explain the selection of the planner (and later of the controller)? something like "Selecting your planner and controller from ros topics"

I do not see clearly this behavior should got into the Action section (even if it is in the action folder).

How do I proceed here?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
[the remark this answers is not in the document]
pabloinigoblasco: Not sure if I understand. Are you talking about the default value of the topic name? I think that is already solved in this second commit.

Have a look to line 38 of the planner selector, it sets "planner_selector" as default if the getInput("topic_name") returns false:
https://github.com/ros-planning/navigation2/pull/2249/commits/ff28145581aff2063b50e8520ca7f1386afc7f1d#diff-54951a5d7aae124ce5bb3d0444176fc0043083f30e3ae2b78ab428450de9f01c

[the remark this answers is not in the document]
SteveMacenski: https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/speed_controller.hpp#L54-L59

These are examples, the `InputPort` method itself has a default-able signature

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
pabloinigoblasco: I will do that for the class brief.

Or you mean to do that for all the methods and fields like you did in that PR?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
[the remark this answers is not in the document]
pabloinigoblasco: Agree. That is a better syntax. I did not know that overload of the getInput function.

nav2_behavior_tree/CMakeLists.txt
pabloinigoblasco: Isnt this requirement completed in this second commit?

nav2_behavior_tree/test/plugins/action/test_planner_selector_node.cpp
pabloinigoblasco: Already added for this third iteration (4th commit)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
pabloinigoblasco: Already added for this third iteration (4th commit)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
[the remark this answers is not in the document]
pabloinigoblasco: Already added for this third iteration (4th commit)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
```diff
@@ -0,0 +1,85 @@
...
+        "the input topic name to select the planner"),
+
+      BT::OutputPort<std::string>(
+        "selected_planner",
+        "Selected planner by subscription")
+    };
+  }
+
+private:
+  BT::NodeStatus tick();
```
SteveMacenski, line 70: https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/goal_updated_condition.hpp#L44-L57 missing doxygen, all functions should be documented.

SteveMacenski (changes requested):
LGTM - just missing complete docs. Following up in comment about readme, but also still missing the navigation.ros.org updates I also described. But no more changes to code required. For the website docs:
- Migration guide add this new BT node to the new BT node section https://navigation.ros.org/migration/Foxy.html#new-behavior-tree-nodes
- Navigation plugins page add this new BT node https://navigation.ros.org/plugins/index.html
- Configuration guide add page for the ports (ex https://navigation.ros.org/configuration/packages/bt-plugins/actions/ComputePathToPose.html)

nav2_behavior_tree/CMakeLists.txt
SteveMacenski: https://github.com/ros-planning/navigation2/tree/main/nav2_behavior_tree#navigation-specific-behavior-tree-nodes simply add a new entry into the table.

pabloinigoblasco:
> LGTM - just missing complete docs. Following up in comment about readme, but also still missing the navigation.ros.org updates I also described. But no more changes to code required. For the website docs:
> 
> * Migration guide add this new BT node to the new BT node section https://navigation.ros.org/migration/Foxy.html#new-behavior-tree-nodes
> * Navigation plugins page add this new BT node https://navigation.ros.org/plugins/index.html
> * Configuration guide add page for the ports (ex https://navigation.ros.org/configuration/packages/bt-plugins/actions/ComputePathToPose.html)

I am with these last tasks. However, I have a weird feeling with this PlanerSelector node about its "category", is it an Nav2 action plugin? Means "ActionNode" for the nav2 stack to inherit from BTActionNode<ActionT>? or at least to be an action client? If so, it looks like this PlannerSelector node is not an "Action".

This PlannerSelector is located inside plugins/action, is that correct? 

For all these documentation tasks I would need to know its category.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
pabloinigoblasco: These doxygen comments on functions is already done in the last commit

pabloinigoblasco:
> LGTM - just missing complete docs. Following up in comment about readme, but also still missing the navigation.ros.org updates I also described. But no more changes to code required. For the website docs:
> 
> * Migration guide add this new BT node to the new BT node section https://navigation.ros.org/migration/Foxy.html#new-behavior-tree-nodes
> * Navigation plugins page add this new BT node https://navigation.ros.org/plugins/index.html
> * Configuration guide add page for the ports (ex https://navigation.ros.org/configuration/packages/bt-plugins/actions/ComputePathToPose.html)

I did a pull request to the documentation respository:
https://github.com/ros-planning/navigation.ros.org/pull/152

SteveMacenski:
> Is it an Nav2 action plugin?

Yes, it belongs in the BT - Action subsection in the navigation plugins list, since its an action plugin (deriving from the action plugin type). It is a Behavior Tree Action node, it has nothing to do with ROS actions.

SteveMacenski (approved):
@naiveHobo you happy with this?

@pabloinigoblasco One more place to add it: https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/nav2_tree_nodes.xml for groot monitoring

SteveMacenski (approved):

SteveMacenski:
@pabloinigoblasco pull in `main` or rebase in order to get CI building, its unhappy about some `rclcpp` changes a couple of weeks ago

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
```diff
@@ -0,0 +1,97 @@
...
+   * @brief Function to perform some user-defined operation on tick
+   */
+  BT::NodeStatus tick();
+
+  /**
+   * @brief callback function for the planner_selector topic
+   *
+   * @param msg the message with the id of the planner_selector
+   */
+  void callback_planner_select(const std_msgs::msg::String::SharedPtr msg);
```
naiveHobo, line 84: Function naming styling should be `callbackPlannerSelect`

nav2_behavior_tree/plugins/action/planner_selector_node.cpp
```diff
@@ -0,0 +1,68 @@
...
+    getInput("default_planner", last_selected_planner_);
+  }
+
+  setOutput("selected_planner", last_selected_planner_);
+
+  return BT::NodeStatus::SUCCESS;
+}
+
+void
+PlannerSelector::callback_planner_select(const std_msgs::msg::String::SharedPtr msg)
```
naiveHobo, line 57: Function naming styling should be `callbackPlannerSelect`

naiveHobo (approved):
Just a small naming issue, otherwise LGTM!

pabloinigoblasco:
> @pabloinigoblasco pull in `main` or rebase in order to get CI building, its unhappy about some `rclcpp` changes a couple of weeks ago

I have to analyze this thoroughly. I just merged the "main branch" and I am getting erros on nav2_utils and some rclcpp Exceptions and not sure why. 

I am working on that.

naiveHobo:
You're probably on an older version of ros2 master build, you'll have to update your local source ros2 repositories and build ros2 master again to get the latest changes. https://docs.ros.org/en/foxy/Installation/Maintaining-a-Source-Checkout.html#update-your-repositories

pabloinigoblasco:
> You're probably on an older version of ros2 master build, you'll have to update your local source ros2 repositories and build ros2 master again to get the latest changes. https://docs.ros.org/en/foxy/Installation/Maintaining-a-Source-Checkout.html#update-your-repositories

Thanks I will have a look to the link you say.

I am working on rolling not directly on the ros2 main branch. So, according to what you say, that should be the problem.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
pabloinigoblasco: This should be already fixed

nav2_behavior_tree/plugins/action/planner_selector_node.cpp
pabloinigoblasco: This should be already fixed

SteveMacenski:
Sweet! This is good to go once CI passes + your documentation PR is completed (just 1 more missing thing so that the page is reachable)

SteveMacenski:
@pabloinigoblasco thanks for this! The controller / goal checker should be just exactly equal to this with different names / topics!

Issue #2051, opened 2020-10-20
[InflationLayer] inflation artifacts behind update bounds


AlexeyMerzlyakov:
The problem appears when apply some layer which is going after `inflation_layer`.
In my example, it is `keepout_filter` layer from [this tutorial](https://navigation.ros.org/tutorials/docs/navigation2_with_keepout_filter.html) applied after `inflation_layer` in a global costmap as follows:
```
@@ -206,7 +210,7 @@ global_costmap:
-      plugins: ["static_layer", "obstacle_layer", "inflation_layer"]
+      plugins: ["static_layer", "obstacle_layer", "inflation_layer", "keepout_filter"]
```
The costmap artifacts look like follows:
![Screenshot_2020-10-20_18-51-39](https://user-images.githubusercontent.com/60094858/96616331-1ef6f080-130b-11eb-8477-d4c3984a17d7.png)

This will look even worse when using two near keepout zones e.g. in a following warehouse example:
![Screenshot_2020-10-20_19-02-41](https://user-images.githubusercontent.com/60094858/96613294-74c99980-1307-11eb-8caf-58102b1ed09f.png)

In last case robot can not plan between two keepouts because of these artifacts.

The problem appears because of inflation layer in `updateCosts()` routine employs points out of costmap window (set in `updateBounds()`) in the following part of code:
```
void InflationLayer::updateCosts()
...
  // We need to include in the inflation cells outside the bounding
  // box min_i...max_j, by the amount cell_inflation_radius_.  Cells
  // up to that distance outside the box can still influence the costs
  // stored in cells inside the box.
  min_i -= static_cast<int>(cell_inflation_radius_);
  min_j -= static_cast<int>(cell_inflation_radius_);
  max_i += static_cast<int>(cell_inflation_radius_);
  max_j += static_cast<int>(cell_inflation_radius_);
```
Since `LayeredCostmap` does not know anything about it, it resets only bounds window between each iteration set by `updateBounds()` leaving untouched out-bounds-window. And, inflation layer re-inflates unnecessary points remaining from previous iteration which causes the artifacts from above.

I've started to check the history of the above part of code and found that corresponding borders change was made in:
https://github.com/ros-planning/navigation/commit/7ef4d926d97e001e5697e4de9c64433c3c5684b6
In this commit bounds expansion was removed from `updateBounds()` and added to `updateCosts()`.
However, it was moved back later in:
[...]

PR #2259, opened 2021-03-19
Fix for inflation artifacts behind update bounds

AlexeyMerzlyakov:
This appearing when some layer is going after `inflation_layer`. More details are in #2051 ticket.

### DIFF
--- a/nav2_costmap_2d/test/integration/inflation_tests.cpp
+++ b/nav2_costmap_2d/test/integration/inflation_tests.cpp
@@ -328,7 +328,7 @@ TEST_F(TestNode, testInflationAroundUnkown)
   layers.updateMap(0, 0, 0);
 
   layers.getCostmap()->setCost(4, 4, nav2_costmap_2d::NO_INFORMATION);
-  ilayer->updateCosts(*layers.getCostmap(), 0, 0, 8, 8);
+  ilayer->updateCosts(*layers.getCostmap(), 0, 0, 10, 10);
 
   validatePointInflation(4, 4, layers.getCostmap(), ilayer, inflation_radius);
 }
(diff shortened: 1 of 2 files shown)


AlexeyMerzlyakov:
ToDo: fix failing tests

SteveMacenski (commented):
This makes sense to me - though its unclear to me why CI is failing (though 1 is linting)

AlexeyMerzlyakov:
> This makes sense to me - though its unclear to me why CI is failing (though 1 is linting)

Great. Regarding failed test cases: in inflation_tests `updateCosts()` call boundaries did not cover full inflated area. Previously did worked as inflation spreaded out of bounds. Currently became to fail in test. Fixed in https://github.com/ros-planning/navigation2/pull/2259/commits/8ed97de8316e75128f2a6414e0c222cf8af4d11f.
Linting test was failed due to incorrect indent style I've missed. Also was fixed in the same commit.

nav2_costmap_2d/test/integration/inflation_tests.cpp
```diff
@@ -328,7 +328,7 @@ TEST_F(TestNode, testInflationAroundUnkown)
   layers.updateMap(0, 0, 0);
 
   layers.getCostmap()->setCost(4, 4, nav2_costmap_2d::NO_INFORMATION);
-  ilayer->updateCosts(*layers.getCostmap(), 0, 0, 8, 8);
+  ilayer->updateCosts(*layers.getCostmap(), 0, 0, 10, 10);
```
SteveMacenski, line 331: Why was this change required? is this working around an actual issue?

SteveMacenski (approved):
LGTM, anything more before its merged?

AlexeyMerzlyakov:
Would like to fix "wings" artifacts too: they can cause noticeable troubles for keepout filter on some cases. However, there is no trivial solution I see for it yet, as we cannot make seamless stitching of window boundaries and exclude keepout zones from obstacles in saved costmaps at the same time. Still thinking about possible workaround for that. If I will find any seems to be suitable clue, I'll open a new PR for that.

PR #2261, opened 2021-03-21
Handle navigate_to_pose preemptions with new BT

### LINKED ISSUE #1971: nav2_bt_navigator: preempting navigation requests with a new BT uses old BT
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- ROS2 Version:
  - Foxy source
- Version or commit hash:
  - master

#### Steps to reproduce issue
Use the following behavior tree which implements the most basic navigation logic; compute path to pose and follow path:

```xml
  <BehaviorTree ID="MainTree">
    <Sequence name="NavigateWithReplanning">
      <FollowPath path="{path}"  controller_id="FollowPath"/>
    </Sequence>
  </BehaviorTree>
</root>
```

Send a navigation goal through rviz and the bot should start executing a path to this goal. Send another goal before the previous goal is reached.

#### Expected behavior
The previous goal should be preempted and the bot should start following a path to the new goal.

#### Actual behavior
The bot continues to move towards the previous goal and the new navigation goal is lost.

naiveHobo:

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | closes #1971 |
| Robotic platform tested on | TurtleBot3 Gazebo simulation |

---

* `bt_navigator` is configured to handle navigation preemption requests with a different BT than the one that the current running action is executing
* If a preemption request uses the same BT as the current action, current BT is not halted. The new goal is accepted without explicitly cancelling the old goal and the new navigation goal pose is directly updated on the blackboard.
* If a preemption request asks for a BT change, the previous goal is cancelled and the new goal is processed by creating a new BT from the requested XML file.
* `BtActionServer` accepts an additional function `ShouldCancelGoalCallback` that is called during the `is_cancelling` check. This function can define cases when the current running action executing a BT should be cancelled other than when an explicit cancel is requested on the current action. `is_cancelling` first checks if cancel is requested for the current action, and only when there's no explicit cancel requested, the user defined `ShouldCancelGoalCallback` is executed to check if the current goal should be cancelled. Using this callback, `bt_navigator` checks if a preempt was requested with a different BT and cancels the current goal when that's the case.

* Explicitly explain how preemption is handled in the documentation?

---

* Still need to decouple the preemption logic from `ComputePathToPose` BT node when preemption request uses the same BT as the previous goal (refer https://github.com/ros-planning/navigation2/issues/1971#issuecomment-705826289)

#### For Maintainers: 


naiveHobo:
@SteveMacenski want to get your thoughts on this, does it seem like this is over-complicating the logic a bit?

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -229,10 +230,49 @@ BtNavigator::onLoop()
...
+      bt_action_server_->getCurrentBTFilename() == bt_action_server_->getDefaultBTFilename()))
+    {
+      return false;
+    }
+
+    // if pending goal requested a different BT we'll cancel the current goal
+    RCLCPP_INFO(
+      get_logger(),
+      "Goal was preempted with a different BT XML file, cancelling the previous goal.");
+    return true;
```
SteveMacenski, line 271: This all logically tracks for me, but really makes me wonder if this is OK. What quite literally we're doing is saying:
- Do you want to preempt with a new BT?
- OK, we're going to cancel and reissue 

Wouldn't we, in this situation, rather the user cancel themselves and issue a new goal when changing BTs? This isn't really "preemption" if we're actually cancelling, its not a continuous, smooth transition from BT 1 to BT 2. 

It makes me think that if this is the only clear way to enable preemption (by canceling and accepting the pending goal), we shouldn't support preemption of new BTs at all (e.g. the `onPreempt` if not the same BT should reject the goal.

SteveMacenski (commented):
So to make sure I understand, if preempted with same BT, the usual applies. Then if it doesn't pass, in the next `isCancelled()` call in the BT engine, it calls a callback to see if pending and a different BT to issue a cancel. The cancel then would put the pending goal on the now processing list.

The mechanics of "cancel" the tree to enact a "preemption" seems like a not good situation, since it doesn't meet the meaning of "preemption" if we're canceling or stopping a goals - resulting probably in the robot stuttering. If this is the only way to accomplish this, I wonder if the right reaction should be rejecting new BT preemption requests (e.g. log the failure and reject to add the goal to pending). While it wouldn't enable preemption of the BT, it would at least not incorrectly hide it

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -229,10 +230,49 @@ BtNavigator::onLoop()
...
+  {
+    // if pending goal requests the same BT as the current goal, accept the pending goal
+    // if pending goal has an empty behavior_tree field, it requests the default BT file
+    // accept the pending goal if the current goal is running the default BT file
+    initializeGoalPose(bt_action_server_->acceptPendingGoal());
+  }
+}
+
+bool
+BtNavigator::shouldCancelCurrentGoal()
```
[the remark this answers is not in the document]
naiveHobo: My idea here was that a BT application server developer could decide to add different cases where a particular goal should be cancelled without the action client having to explicitly cancel the goal, so providing this function as something you could pass to `BtActionServer` would make sense (like `onPreempt`, `onGoalReceived`, etc.)

naiveHobo:
> So to make sure I understand, if preempted with same BT, the usual applies. Then if it doesn't pass, in the next `isCancelled()` call in the BT engine, it calls a callback to see if pending and a different BT to issue a cancel. The cancel then would put the pending goal on the now processing list.

Exactly

> Do you think there's a way we can do this without literal cancellation? I'm not sure I see one. Because canceling or halting the tree results in a robot stoppage or stutter, these can't be considered preemption.

I don't think there's a way to do that, a new BT would always require the old BT to be halted which means there has to be a cancellation to achieve that

> Wouldn't we, in this situation, rather the user cancel themselves and issue a new goal when changing BTs? This isn't really "preemption" if we're actually cancelling, its not a continuous, smooth transition from BT 1 to BT 2.

I completely agree with everything you said above, even I'm not too happy about cancelling and accepting the pending goal to _mimic_ preemption with a new BT. The only argument I can see (though a weak one) is that if cancelling is **the only way** to shift to a new BT do we make a user cancel the action or do we provide the functionality in nav2 itself. I can see an argument to make this a default nav2 behavior since that's what a user would have to do anyway.

Rejecting a preemption request with a new BT makes sense to me though there's a potential issue I can see there. The new action with a different BT would be held as the pending goal handle by the `SimpleActionServer`. We would reject the preemption request through `bt_navigator` which would mean that the robot would continue following the current goal until it fails or succeeds. As soon as the last goal is terminated, the pending goal would be processed and `bt_navigator` would create a new BT and start moving towards the new goal. So although `bt_navigator` rejects the preemption with a new BT, **this goal would be processed later**.

nav2_bt_navigator/src/bt_navigator.cpp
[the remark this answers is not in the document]
SteveMacenski: OK, lets talk in the major points and circle back to this one since its impacted

SteveMacenski:
> I don't think there's a way to do that, a new BT would always require the old BT to be halted which means there has to be a cancellation to achieve that

I totally agree with you that being able to cancel -> reissue is a useful mechanism for people that don't want to track their goals and tasks carefully. But "preemption" strongly implies that we're going to be continuously moving from one task to the other without stoppage which we can't promise if we have to cancel completely the previous task before processing the new one. The BT halt would cause a stutter in the system where the robot tries to stop for a very short period from cancellation until a new plan was computed to follow. In many cases, planning with NavFn is super quick so its this jerk. In other cases where you want to do verifiable kinematic planning with collision checking could take 100ms - 1s so that wouldn't "preempt".

> do we make a user cancel the action and reissue or do we provide the functionality in nav2 itself

Agreed, that's a big question. But I think there's another question if preemption is the right mechanism by which we try to do this. A professional or careful application will be storing the goal handle of the task at hand. I don't feel like its too much to ask for an application to check if the current goal handle and the pending goal have the same task BT. If so, then preemption is OK. Else, cancel and reissue. 

If there was a new idea, other than preemption, to "cancel and reissue" within the action server itself, I think that would make sense. But that is not a preemption. My fear is that people will do this and then get made that they're preempting goals but the action isn't smooth. If we reject new BT preemption requests, then it would force users to track their goal handles better and check if changes and they can handle the cancel -> issuing new goal logic themselves in < 50 lines. I think that is generally good practice. 

I think in light of no way to enable "smooth" preemption between BT tasks, `OnPreempt()` should reject any preemption request with a new BT, log that it can't preempt with a new BT due to non-smooth transitions, and return the pending goal as rejected. If we make it clear in logging, action type returns, etc, then developers will know how to use the Nav2 action interface. I think their local application work arounds for this would be trivially easy to implement and its worth maintaining the "trueness" of preemption

naiveHobo:
Completely agree with everything you said above.

> However, I don't think preempt has access to the goal issuing on the nav2_util simple action server. I think goals there are automatically accepted, so we can't reject the goal outright there unless we restructure this a bit. Instead, we could only abort the existing pending goal handle?

That's right, `SimpleActionServer` automatically accepts a new goal and starts processing it once the previous goal is done. A goal is always accepted and put into the pending queue with the current implementation of `SimpleActionServer`. Also, right now there's no way to terminate the pending goal through an application that uses `SimpleActionServer`.

We could add `SimpleActionServer::abort_pending_goal()` as an analog to `SimpleActionServer::accept_pending_goal()` to allow rejection of pending goals so they are never processed even after the current goal finishes execution. That seems like a useful functionality to add to `SimpleActionServer` anyway. What do you think?

Once there's a way to make sure a preemption request with a different BT is aborted as soon as it is received, I completely agree with rejecting the preempt request through `OnPreempt` and logging the reason.

SteveMacenski:
> We could add SimpleActionServer::abort_pending_goal() ...

I agree the function would be useful! I'm looking through the waterfall of data and I believe that would solve the problem completely. We already have the mechanics in place in the BT Action Server node from this PR to get the pending handle in `onPreempt` (but not accepting it yet) to check for validity. If valid, accept. If not, abort. 

I think this is an error we'll get alot of attention around if its unclear, so this one should be _very_ verbose as to what's happening, why, and how to fix it.

SteveMacenski:
Any progress on this? I think we've come to the conclusion we should reject these goals

naiveHobo:
One more review and it should be ready for merge! @SteveMacenski

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
```diff
@@ -133,6 +150,24 @@ class BtActionServer
...
+  const std::shared_ptr<const typename ActionT::Goal> getPendingGoal() const
+  {
+    return action_server_->get_pending_goal();
+  }
+
+  /**
+   * @brief Wrapper function to check if preempt is requested
+   * @return bool true if preempt is requested, false otherwise
+   */
+  bool isPreemptRequested()
```
SteveMacenski, line 166: This function isn't used anymore. I totally agree its valuable, but dead code is exactly why costmap_2d is such a mess :laughing: We can readd it if its needed at any time

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -309,6 +309,24 @@ class SimpleActionServer
...
+  void terminate_pending_goal()
+  {
+    std::lock_guard<std::recursive_mutex> lock(update_mutex_);
+
+    if (!pending_handle_ || !pending_handle_->is_active()) {
+      error_msg("Attempting to terminate pending goal when not available");
+      return;
+    }
+
+    terminate(pending_handle_);
```
SteveMacenski, line 324: https://github.com/ros-planning/navigation2/blob/012fc19b059bf54a351809deb6709d523d3bfedd/nav2_util/include/nav2_util/simple_action_server.hpp#L304 should this also be reset after terminating?

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -231,10 +231,27 @@ BtNavigator::onLoop()
...
+  {
+    // if pending goal requests the same BT as the current goal, accept the pending goal
+    // if pending goal has an empty behavior_tree field, it requests the default BT file
+    // accept the pending goal if the current goal is running the default BT file
+    initializeGoalPose(bt_action_server_->acceptPendingGoal());
+  } else {
+    RCLCPP_WARN(
+      get_logger(),
+      "Preemption request was rejected since the requested BT XML file is not the same "
+      "as the one that the current goal is executing. Cancel the current goal and send a new "
```
SteveMacenski, line 250: A short sentence stating "why" would be nice, something like: Preemption with a new BT is invalid since it would require pausing task execution, not "preempting".

SteveMacenski (commented):
Very good - just a few cleanup items. I also retriggered CI to get the tests to pass, just to sanity check outselves

nav2_util/include/nav2_util/simple_action_server.hpp
naiveHobo: `terminate` resets the handle
https://github.com/ros-planning/navigation2/blob/012fc19b059bf54a351809deb6709d523d3bfedd/nav2_util/include/nav2_util/simple_action_server.hpp#L502

SteveMacenski (approved):

SteveMacenski:
Just waiting on CI! Thanks for this!

SteveMacenski:
Note: This PR will be in conflict with https://github.com/ros-planning/navigation2/pull/2271, so I'd like to get this PR in before mine so I have to rectify it rather than you.

Issue #2262, opened 2021-03-21
Nav2 equivalent to move_base controller_patience


doisyg:
#### Feature description
More than a feature request this is an interogation. In ROS1, `move_base` has a parameter called `controller_patience` described as:

>  How long the controller will wait in seconds without receiving a valid control before space-clearing operations are performed

(i.e. before giving up)

This was further extended in MBF with `planner_max_retries`:
> How many times we will recall the planner in an attempt to find a valid plan before giving up

Is there an equivalent in nav2 ? And if not would a PR be of interest (I need a similar functionality so I will implement it regardless) ?

#### Implementation considerations
I see different way of implementing it

1. (personally preferred) Exposing this parameter in `nav2_controller` and try/catching the call to `computeVelocityCommands` https://github.com/ros-planning/navigation2/blob/3bf1f7a93d740c54b4c7ef7fc6dd914175bf46be/nav2_controller/src/nav2_controller.cpp#L388-L392
And publishing 0 velocities in case of an exception as long we are under `controller_patience`
2. Exposing it at the controller plugin level

SteveMacenski:
See the progress checker, if progress isn't made, then the local BT recoveries are triggered. The BT recovery nodes also have `num_of_retries` built into them for the number of attempts at failed planning/control/navigation before failing or triggering a secondary recovery wave.

doisyg:
Thanks for you answer,  using BT we can achieve a similar result (i.e. retrying the control action a specific number of times in case of failure).

However I feel that something is missing. I will try to be clear.
In case of the `nav2_core::Controller` plugin throwing a `nav2_core::PlannerException` when calling `computeVelocityCommands`, the `nav2_controller` "FollowPath" action terminates.
And in case the progress checker detects that there is no progress, the `nav2_controller` "FollowPath" action also terminates in a similar way.
Hence there is no way of distinguishing between a "FollowPath" failure due to no progress, and a failure due to the `nav2_core::Controller `plugin outputting invalid commands.

This would be an issue if we wanted to distinguish these 2 cases. This is where I am biased as I am trying to implement the following behavior:

> Make the `nav2_core::Controller` plugin follow a path and retry if needed for a period of time (or a number of time), BUT fail if no progress is detected.

SteveMacenski:
>  Hence there is no way of distinguishing between a "FollowPath" failure due to no progress, and a failure due to the nav2_core::Controller plugin outputting invalid commands.

[...]

PR #2264, opened 2021-03-23
[Draft] [nav2_controller] Add controller_patience param

doisyg:
| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2262 |

---

This is a draft PR to support the discussion from here https://github.com/ros-planning/navigation2/issues/2262
Adds a `controller_patience` which makes the controller retry when for `computeVelocityCommands` throws an exception for the desired amount of second before failing. When equals to 0.0 the functionality is disabled. When equals to -1.0 the patience is infinite (the nav2_controller "FollowPath" action will likely fails thanks to the ProgressChecker though).

#### For Maintainers: 

Edit: I meant ProgressChecker instead of GoalChecker

### DIFF
--- a/nav2_controller/include/nav2_controller/nav2_controller.hpp
+++ b/nav2_controller/include/nav2_controller/nav2_controller.hpp
@@ -229,8 +229,13 @@ protected:
   double min_y_velocity_threshold_;
   double min_theta_velocity_threshold_;
 
+  double controller_patience_;
+
   // Whether we've published the single controller warning yet
   geometry_msgs::msg::Pose end_pose_;
+
+  // Last time the controller generated a valid command
+  rclcpp::Time last_valid_cmd_time_;
 
 private:
   /**
--- a/nav2_controller/include/nav2_controller/nav2_controller.hpp
+++ b/nav2_controller/include/nav2_controller/nav2_controller.hpp
@@ -229,7 +229,7 @@ protected:
   double min_y_velocity_threshold_;
   double min_theta_velocity_threshold_;
 
-  double controller_patience_;
+  double failure_tolerance_;
 
   // Whether we've published the single controller warning yet
   geometry_msgs::msg::Pose end_pose_;
--- a/nav2_controller/src/nav2_controller.cpp
+++ b/nav2_controller/src/nav2_controller.cpp
@@ -55,6 +55,8 @@ ControllerServer::ControllerServer()
 
   declare_parameter("speed_limit_topic", rclcpp::ParameterValue("speed_limit"));
 
+  declare_parameter("controller_patience", rclcpp::ParameterValue(0.0));
+
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
     "local_costmap", std::string{get_namespace()}, "local_costmap");
@@ -297,6 +299,7 @@ void ControllerServer::computeControl()
     setPlannerPath(action_server_->get_current_goal()->path);
     progress_checker_->reset();
 
+    last_valid_cmd_time_ = now();
     rclcpp::WallRate loop_rate(controller_frequency_);
     while (rclcpp::ok()) {
       if (action_server_ == nullptr || !action_server_->is_server_active()) {
@@ -385,11 +388,34 @@ void ControllerServer::computeAndPublishVelocity()
 
   nav_2d_msgs::msg::Twist2D twist = getThresholdedTwist(odom_sub_->getTwist());
 
-  auto cmd_vel_2d =
-    controllers_[current_controller_]->computeVelocityCommands(
-    pose,
-    nav_2d_utils::twist2Dto3D(twist),
-    goal_checker_.get());
+  geometry_msgs::msg::TwistStamped cmd_vel_2d;
+
+  if (controller_patience_ > 0 || controller_patience_ == -1.0) {
+    try {
+      cmd_vel_2d =
+        controllers_[current_controller_]->computeVelocityCommands(
+        pose,
+        nav_2d_utils::twist2Dto3D(twist),
+        goal_checker_.get());
+      last_valid_cmd_time_ = now();
+    } catch (nav2_core::PlannerException & e) {
+      RCLCPP_WARN(this->get_logger(), e.what());
+      cmd_vel_2d.twist.angular.x = 0;
+      cmd_vel_2d.twist.angular.y = 0;
+      cmd_vel_2d.twist.angular.z = 0;
+      cmd_vel_2d.twist.linear.x = 0;
+      cmd_vel_2d.twist.linear.y = 0;
+      cmd_vel_2d.twist.linear.z = 0;
+      cmd_vel_2d.header.frame_id = costmap_ros_->getBaseFrameID();
+      cmd_vel_2d.header.stamp = now();
+    }
+  } else {
+    cmd_vel_2d =
+      controllers_[current_controller_]->computeVelocityCommands(
+      pose,
+      nav_2d_utils::twist2Dto3D(twist),
+      goal_checker_.get());
+  }
 
   std::shared_ptr<Action::Feedback> feedback = std::make_shared<Action::Feedback>();
   feedback->speed = std::hypot(cmd_vel_2d.twist.linear.x, cmd_vel_2d.twist.linear.y);
@@ -398,6 +424,9 @@ void ControllerServer::computeAndPublishVelocity()
 
   RCLCPP_DEBUG(get_logger(), "Publishing velocity at time %.2f", now().seconds());
   publishVelocity(cmd_vel_2d);
+  if (controller_patience_ > 0 && (now() - last_valid_cmd_time_).seconds() > 5) {
+    throw nav2_core::PlannerException("Controller patience exceeded");
+  }
 }
 
 void ControllerServer::updateGlobalPath()
--- a/nav2_controller/src/nav2_controller.cpp
+++ b/nav2_controller/src/nav2_controller.cpp
@@ -409,8 +409,9 @@ void ControllerServer::computeAndPublishVelocity()
       cmd_vel_2d.twist.linear.z = 0;
       cmd_vel_2d.header.frame_id = costmap_ros_->getBaseFrameID();
       cmd_vel_2d.header.stamp = now();
-      publishVelocity(cmd_vel_2d);
-      if ((now() - last_valid_cmd_time_).seconds() > controller_patience_) {
+      if ((now() - last_valid_cmd_time_).seconds() > controller_patience_ &&
+        controller_patience_ != -1.0)
+      {
         throw nav2_core::PlannerException("Controller patience exceeded");
       }
     } else {
--- a/nav2_controller/src/nav2_controller.cpp
+++ b/nav2_controller/src/nav2_controller.cpp
@@ -55,7 +55,7 @@ ControllerServer::ControllerServer()
 
   declare_parameter("speed_limit_topic", rclcpp::ParameterValue("speed_limit"));
 
-  declare_parameter("controller_patience", rclcpp::ParameterValue(0.0));
+  declare_parameter("failure_tolerance", rclcpp::ParameterValue(0.0));
 
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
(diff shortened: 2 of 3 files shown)


SteveMacenski:
Couldn't this be embedded into the progress checker much more concisely? The progress checker API / use could be tweaked if necessary to support.

A concern I have here is if you set this to something like 5-10 seconds and you have the retries in the BT to more than 1, then you'll end up in this `5s * N` loop of trying to control and fail before actually entering meaningful recovery behaviors. Though I think the `RecoveryNode` will trigger the second child between retries if memory serves correctly, so you'd get costmap clearing actions but not the recovery sub-tree. If that's accurate (which I think it is from [looking over the code](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/control/recovery_node.cpp), but I didn't write this), then I could see a patience parameter being useful so you don't clear after a single failure. With that said, its pretty _bad_ if your controller is actually that non-deterministic where 2 consecutive runs would have that different of an outcome. 

With that said, I just want to clean up the code a bit more, more than anything else to be concise. I'd really hope we could deal with this kind of thing at the BT level or inside existing structures like the progress checkers. I think a controller failing to compute is a legitimate issue and maybe its just a different philosophical way of thinking about it that a controller should not return `exception failure` unless its actually failed to a point worthy of failing the action. If its recoverable in future iterations, maybe it shouldn't be throwing.

nav2_controller/src/nav2_controller.cpp
```diff
@@ -385,11 +388,34 @@ void ControllerServer::computeAndPublishVelocity()
...
   nav_2d_msgs::msg::Twist2D twist = getThresholdedTwist(odom_sub_->getTwist());
 
-  auto cmd_vel_2d =
-    controllers_[current_controller_]->computeVelocityCommands(
-    pose,
-    nav_2d_utils::twist2Dto3D(twist),
-    goal_checker_.get());
+  geometry_msgs::msg::TwistStamped cmd_vel_2d;
+
+  if (controller_patience_ > 0 || controller_patience_ == -1.0) {
```
SteveMacenski, line 393: You can remove the `else`, and move the `if` condition inside of the `catch()`. That way you only have 1 compute velocity call and if it catches, that's when you care about the patience to `throw()` if exceeding. I think all of the patience logic can be put into `catch() {...}`, you don't want to even reach the publish stage if you're going to be throwing a failure

doisyg: I moved all the patience logic inside the `catch() {...}` and piped the controller exception in case we don't use patience. Plus added the `publishVelocity` 0 inside (I assume we want to have the last cmd published to be 0 in case of exception). However, we now loose feedback when the controller plugin throws exceptions that we catch with the patience.
Plus properly getting the parameter and corrected the hardcoded 5 seconds patience (in a word, I properly tested and debuged it)

doisyg:
> Couldn't this be embedded into the progress checker much more concisely? The progress checker API / use could be tweaked if necessary to support.

That would be ideal, but I am not clearly seeing how ?


> A concern I have here is if you set this to something like 5-10 seconds and you have the retries in the BT to more than 1, then you'll end up in this `5s * N` loop of trying to control and fail before actually entering meaningful recovery behaviors. Though I think the `RecoveryNode` will trigger the second child between retries if memory serves correctly, so you'd get costmap clearing actions but not the recovery sub-tree. If that's accurate (which I think it is from [looking over the code](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/control/recovery_node.cpp), but I didn't write this), then I could see a patience parameter being useful so you don't clear after a single failure.

Independently from how the surrounding BT is configured (I use my own BT nodes), the ProgressChecker will cap the maximum time each loop will control and fail.


> With that said, its pretty _bad_ if your controller is actually that non-deterministic where 2 consecutive runs would have that different of an outcome.
> 
> With that said, I just want to clean up the code a bit more, more than anything else to be concise. I'd really hope we could deal with this kind of thing at the BT level or inside existing structures like the progress checkers. I think a controller failing to compute is a legitimate issue and maybe its just a different philosophical way of thinking about it that a controller should not return `exception failure` unless its actually failed to a point worthy of failing the action. If its recoverable in future iterations, maybe it shouldn't be throwing. Looking over DWB/RPP, they only throw when in collision or no valid paths what so ever, which seem like genuine, unrecoverable states worthy of failing the action & doing a costmap clear.
> 
> Can you explain what you're running into where you need to throw a genuine failure exception but shouldn't be considered action-terminal to clear a costmap or something?


Well, maybe this is not the best practice, but I can think of use cases where 2 consecutive runs would have different outcomes AND where the controller can throw and exception that you could want not to be action-terminal:

1. A dynamic obstacle (aka people) jumps on the trajectory OR an obstacle pass on the global plan in case of a controller configured with very little freedom to deviate from the global plan (and no global plan re-planning); in these cases the controller naturally consider the cmd as invalid, however you want to leave a chance to the obstacle to move before triggering the recovery behaviors.
2. A controller having a recovery logic depending on the number of consecutive failure, for instance TEB with `no_infeasible_plans_` and horizon shrinking: https://github.com/rst-tu-dortmund/teb_local_planner/blob/8e4be1bc2fa9789db564a10c1312382a574089fb/teb_local_planner/src/teb_local_planner_ros.cpp#L965-L991


> Just for completion, another option we have is to have different controller exceptions we can catch for different reasons, if that helps in the progress checker / controller / patience logic (from the issue thread).

I like that option too. I just have to admit that I am a bit confused by the whole feedback through exceptions design.

>     * Maybe put into progress checker, or make it at least more concise (see comment below). We also talked about in the BT as well maybe.

Can you clarify a bit how you see that going in the progress checker ?  I think that would be my favorite short term solution. And then, in addition mid term rethinking the action feedback/result so one could have the freedom of dealing with these cases at the BT level.


>     * Not envisioning the issues you've run into where this is necessary - is your controller flaky in computing values or something? I could see this being an issue in dynamic scenes where the robot gets cut off, but then I'd think the clear costmap and use of `retries` in the behavior tree would actually be the best solution!

dynamic scenes and robot getting cutoff is precisely my use case (explained above). I don't want to start a recovery behavior and destroy my robot situational awareness (i.e. clearing costmaps, that in my case is followed by a complete robot rotation to rebuild sense of the environment), just because of a passing obstacle.

nav2_controller/src/nav2_controller.cpp
SteveMacenski: Woah - if you want `publishVelocity`, when what's wrong with the current fail -> retry mechanism in the BT via the recovery nodes? I thought the goal here was to remove any stops. I believe this is what Nav1 did, but I can't recall off the top of my head

SteveMacenski: We'll also have to deal with that feedback issue - but lets deal with the big issues first

SteveMacenski:
OK, while I think I disagree with you about clearing costmaps being bad in dynamic situations, my robots also typically have really great sensor coverage so regenerating a half way decent costmap only takes a few milliseconds (so clear fast, clear often was my go to pre-STVL).

Honestly, I'm struggling a little here on what the best strategy is. It seems like this kind of logic could be handled in the behavior tree - but I'm also not sure that its the right place for it either. If we do it in the controller server, I wonder if this should be a parameterization of the action request or of the server itself (e.g. should it be a ROS parameter, or an action request field) to get different behaviors as necessary. 

Another question: Is the duration of time, `dt`, the way we want to measure this? Why not the number of failures to compute a trajectory in a row? I don't think either is "objectively" right, but semantically, do you think you as a robotics developer cares about the duration or the number of failures more? I think the absolute number, but I'm curious what you think.

nav2_controller/src/nav2_controller.cpp
doisyg: > Though I'm reading it, and I think it will still do feedback and publish velocities. You `catch` but if you don't rethrow due to patience failure, there's no `return`, so you will continue in to that last code block it looks like to me! (which also brings up the issue of not zeroing out the cmd_vel command outputted in case it's non-zero, you also wouldn't need to `publishZeroVelocity()` in the catch either if we go down to a `publishVelocity`. If we rethrow it due to being over time, then the main `catch` in the server will terminate the action and send the zero velocity command itself)

You are right, no need for the added `publishVelocity` and there is no feedback issue (try / catch manages to confuse me even when I write the code). Fixed in last commit (+ proper dealing with `controller_patience` set to -1.0 for infinite patience)

doisyg:
> OK, while I think I disagree with you about clearing costmaps being bad in dynamic situations, my robots also typically have really great sensor coverage so regenerating a half way decent costmap only takes a few milliseconds (1,000 cameras, so clear fast, clear often was my go to pre-STVL).

I wish I could have a great/perfect sensor coverage too ;) But nav2 is flexible enough to allow each developer to choose its own strategy.

> Honestly, I'm struggling a little here on what the best strategy is. It seems like this kind of logic could be handled in the behavior tree - but I'm also not sure that its the right place for it either. If we do it in the controller server, I wonder if this should be a parameterization of the action request or of the server itself (e.g. should it be a ROS parameter, or an action request field) to get different behaviors as necessary. I see arguments for both and its not clear to me what's best.

I am aligned with you. I don't have a strong opinion on what's the best strategy here. An action request field is a good idea too and that would be one parameter less to deal with, and making it dynamic by nature (let me know if you want me to switch to this option in this draft)



> Another question: Is the duration of time, `dt`, the way we want to measure this? Why not the number of failures to compute a trajectory in a row? I don't think either is "objectively" right, but semantically, do you think you as a robotics developer cares about the duration or the number of failures more? I think the absolute number, but I'm curious what you think.

Yes, it is easier for me to reason in time than in attempts as I don't necessarily control how fast or how slow the controller is. But this is very personal and I agree that neither is objectively right. I guess that's why MBF implemented both (`planner_max_retries`)

doisyg:
I lean just slightly toward leaving it as a parameter (I see it more as a configuration setting than something that would need to change between two "FollowPath" requests).
As for the name, let's brainstorm, maybe leaving controller implicit as we are in the controller server, `compute_cmd_tolerance` ?

SteveMacenski:
OK lets leave as a param for now then. 

Honestly naming for this is quite difficult for me as well :laughing: I don't really have any good suggestions. `invalid_cmd_dur`? I think something-duration would be better than "patience" which is non-descript

doisyg:
Agreed that duration should appear. I want also to convey the idea that it is a threshold / max value, `max_invalid_cmd_dur` ? But it starts to be a bit long.

SteveMacenski:
`failure_tolerance_dur` doesn't really feel right - `max_failure_dur`, `failure_dur`? `max_invalid_cmd_dur` I don't think is bad, its descriptive. 

`cmd_failure_tolerance` or even **`failure_tolerance`**?

doisyg:
> I suppose `transform_tolerance` commonly used has duration implied as well.

Agreed, taking that into account, I quite like `failure_tolerance`. Should we settle for this ?

SteveMacenski:
Sure, lets do it

nav2_controller/include/nav2_controller/nav2_controller.hpp
```diff
@@ -229,7 +229,7 @@ class ControllerServer : public nav2_util::LifecycleNode
   double min_y_velocity_threshold_;
   double min_theta_velocity_threshold_;
 
-  double controller_patience_;
+  double failure_tolerance_;
```
SteveMacenski, line 232: Please add this to the nav2_params.yaml file with some reasonable default (idk, 0.50 seconds?)

Also, add the new param to the configuration guide for the controller server (including what `-1` means)

SteveMacenski (approved):

nav2_controller/include/nav2_controller/nav2_controller.hpp
SteveMacenski: Also, the migration guide that this feature is now available

SteveMacenski:
All we're waiting for here is the documentation PR to navigation.ros.org

A test or two covering this feature would be nice, but not required

doisyg:
Will do by the end of the week!

nav2_controller/include/nav2_controller/nav2_controller.hpp
doisyg: > Please add this to the nav2_params.yaml file with some reasonable default (idk, 0.50 seconds?)

Just added 0.3, but I could also set it to 0.0 if you want to have the exact same behavior as before this PR (0.0 disables it), what do you think?

SteveMacenski: 0.3 works

PR #2266, opened 2021-03-23
Controller selector

pabloinigoblasco:
---

## Basic Info

PR of the controller selector to be integrated in navigation2 main branch. With this BTNode we use a subscriber to enable the ability of selecting the current controller for the navigation.

| Info |  |
| ------ | ----------- |
| Robotic platform tested on | (ROS Rolling and Gazebo simulation turtlebot waffle) |

---

* reading from topics to select the controller
* support for default controller (as static input port)
* support to setting the name of the topic to support multiple controller selectors.

---

#### For Maintainers: 


SteveMacenski (approved):
LGTM, just need the docs PR to match.

If you wanted at this point, you could change the default BT XML file to include these 2 nodes in it with setting the controller/planner IDs.

PR #2267, opened 2021-03-24
Remove service nodes in lifecycle_manager

BriceRenaudeau:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #816 based on #2216 |
| Robotic platform tested on | Wyca (simulation) |

---

* As LifecycleServiceClient use the `nav2_util::ServiceClient` I replaced the service node in the lifecycle manager with a `shared_from_this()`.
* I also made ServiceClient.wait_for_service() return if the service is ready.

This change should reduce by 1 the number of participants in the DDS (per LifecycleManager) as discussed in this [comment](https://github.com/ros-planning/navigation2/issues/816#issuecomment-800670674).

#### For Maintainers: 


nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -125,8 +125,9 @@ class ServiceClient
   /**
   * @brief Block until a service is available
   * @param timeout Maximum timeout to wait for, default infinite
+  * @return true if the client is ready
   */
-  void wait_for_service(const std::chrono::nanoseconds timeout = std::chrono::nanoseconds::max())
+  bool wait_for_service(const std::chrono::nanoseconds timeout = std::chrono::nanoseconds::max())
```
SteveMacenski, line 130: Why a bool? It throws an exception if its not ready, there's no need to return the bool if success/failed because you'll know if an exception wasn't thrown

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -78,19 +75,23 @@ LifecycleManager::LifecycleManager()
   transition_label_map_[Transition::TRANSITION_DEACTIVATE] = std::string("Deactivating ");
   transition_label_map_[Transition::TRANSITION_UNCONFIGURED_SHUTDOWN] =
     std::string("Shutting down ");
+}
 
+LifecycleManager::~LifecycleManager()
+{
+  RCLCPP_INFO(get_logger(), "Destroying %s", get_name());
```
SteveMacenski, line 82:
```suggestion
  RCLCPP_INFO(get_logger(), "Destroying %s", get_name().c_str());
```

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -78,19 +75,23 @@ LifecycleManager::LifecycleManager()
...
+}
+
+void
+LifecycleManager::init()
```
SteveMacenski, line 86: So I understand why you did this, you wanted to use `shared_from_this()` which can't be called from the constructor, but we need to find another way. Creating the `init` function that the `main` must call is not a sufficient solution because then it breaks the ability of this server to ever become a ROS2 component node. 

Another options might be to create a timer on a 100ms timeout created at the end of the constructor that triggers this initialization behavior in a lambda.

SteveMacenski (changes requested):
Always happy to have less nodes!

nav2_util/include/nav2_util/service_client.hpp
BriceRenaudeau: The exception is thrown only if rclcpp is not ok.
And it might be better to remove the while loop and only keep wait_for_service(timeout)?
I don't really get the behavior with it.
https://github.com/ros-planning/navigation2/blob/ed43bed86253961f2723bb38389e51cabf6e7ebf/nav2_util/include/nav2_util/service_client.hpp#L132-L138
And using a return to get the result seems cleaner than using a try/catch.

nav2_lifecycle_manager/src/lifecycle_manager.cpp
BriceRenaudeau: get_name() is already a char* not a std::string. There is no .c_str() method.

nav2_util/include/nav2_util/service_client.hpp
BriceRenaudeau:
```
return client_->wait_for_service(timeout);
```
Should be sufficient?

SteveMacenski: This is an infinite while loop, this will never return a non-successful state because it will block forever or throw an exception in a failure state.

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -51,9 +51,13 @@ class LifecycleManager : public rclcpp::Node
    */
   ~LifecycleManager();
 
+  /**
+   * @brief init create service cleant and start if autostart is true
```
SteveMacenski, line 55:
```suggestion
   * @brief init create service clients and start if autostart is true
```

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -79,18 +76,27 @@ LifecycleManager::LifecycleManager()
...
 
 LifecycleManager::~LifecycleManager()
 {
   RCLCPP_INFO(get_logger(), "Destroying %s", get_name());
 }
 
+void
+LifecycleManager::init()
+{
+  init_timer_->cancel();
```
SteveMacenski, line 92: This should be a lambda instead since its very small, here's an example https://github.com/ros2/examples/blob/master/rclcpp/topics/minimal_publisher/lambda.cpp#L35

BriceRenaudeau:
> This is an infinite while loop, this will never return a non-successful state because it will block forever or throw an exception in a failure state.

Is it good?
Don't you think the `wait_for_service(timeout)` method should have the same behavior in ServiceClient and in rclcpp::Client?

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -79,11 +76,15 @@ LifecycleManager::LifecycleManager()
   transition_label_map_[Transition::TRANSITION_UNCONFIGURED_SHUTDOWN] =
     std::string("Shutting down ");
 
-  createLifecycleServiceClients();
-
-  if (autostart_) {
-    startup();
-  }
+  init_timer_ = this->create_wall_timer(
+    std::chrono::nanoseconds(10),
```
SteveMacenski, line 80: Is 10 nanoseconds enough to make sure you got out of the constructor reliably? (It might be, I really don't know)

SteveMacenski (approved):
LGTM, just one last question

SteveMacenski:
I agree the timeout there is a bit strange and not really meeting the intent of the API I think, but that should be a part of another PR to fix that

nav2_lifecycle_manager/src/lifecycle_manager.cpp
BriceRenaudeau: The node is mono-threaded (no callback group for the timer), so the timer callback will not be executed before the end of the constructor.

SteveMacenski:
Thanks!

I agree the wait for service throwing is weird and that timeout isn't being used as expected. Another PR to fix that would be merged.

PR #2108, opened 2020-11-27
supporting multiple goal checkers

pabloinigoblasco:
---

## Basic Info

Supporting multiple goal checkers in controller_server and switch between them dynamically.
Motivation; at the current moment there isnt (AFAIK) any mechanism to update dynamically the goal checker or the goal checker parameters at runtime.

This pull request for main is a correction of the same pull request made for foxy previously (https://github.com/ros-planning/navigation2/pull/2098)

| Info |  |
| ------ | ----------- |
| Ticket(s) this addresses   | no associated ticket |
| Primary OS tested on | (Ubuntu) |
| Robotic platform tested on | (gazebo simulation turtlebot waffle) |

---

## Description of contribution in a few bullet points

* This is a basic functionality to support multiple goal checkers in the controller_server (in the same way it is supported multiple controllers )

* This initial proposal uses a controller_server parameter to store the "current_goal_checker" (alternatively it could be in the FollowPath action request like it is the controller_id).

* It is provided with a "nav_params.yaml configuration" with a second "goal checker" to show how it is used. But that could be removed final pull request.

## Description of documentation updates required from your changes

No documentation change should be required at (least in the c++ code)
- "documentation chunks" were copied from "multiple controller" support, for example "findGoalCheckerId"
---

## Future work that may be required in bullet points

* if the "parameter approach" is not convenient it could be replaced by an additional field in the FollowAction request

Issue #2251, opened 2021-03-16
Update BT nodes to use callback groups / executors to process


SteveMacenski:
For example: https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/service_client.hpp#L43

Subs:
- [x] Battery low
- [x] Goal updater
- [x] planner / controller / goal checker selectors 

Actions:
- [x] recoveries (spin, wait, backup)
- [x] follow path
- [x] compute path
- [x] NavigateToPose

Services:
- [x] clear all
- [x] clear region
- [x] clear except
- [x] relocalize global (does this even work anymore?)

SteveMacenski:
It's come to my attention that the executors might be creating a new context remapping to a new DDS participant. This would not be good if we had each BT node creating a new participant to process separately. Some more research has to be done to figure out if this is workable before we jump into it.

Figure out if this does indeed add more participants to the network, if so, if there's a workable way around it so we can process callbacks uniquely in our BT nodes with little overhead.

If not, leave as is.

BriceRenaudeau:
Right now, BT nodes are using there parent node("_rclcpp_node") to create subscriber/service/action so there are not creating new context.
May be we can also pass a multi Threaded executor create and spin in the parent node through the blackboard to attach callback groups. 
It will still create a new context but only one for all the BT node and remove the client_node.

SteveMacenski:
> BT nodes are using there parent node("_rclcpp_node") to create subscriber/service/action so there are not creating new context

Correct @BriceRenaudeau , but this ticket was considering adding the same logic as you introduced in the simple service client. From the design document I linked to above and a little digging, the impression I'm under is that a new DDS participant is made in each new context. The constructor of the `executor` creates a new context object by default when none is given. 

Therefore, I think that we're actually creating a new network participant in the simple service client, which is probably worse for performance than before. It may be fixable, but I haven't had cycles to look into it.

This ticket is wanting to use that same design with a sub/pub/etc made from `node` but added to a local BT-node specific executor. I don't think we can do this due to the fact that it would create a new context, ergo a new DDS participant.

SteveMacenski:
[...]

PR #2269, opened 2021-03-24
Goal Checker selector

pabloinigoblasco:
---

## Basic Info

Support for multiple Goal Checkers in the nav2_controller node. The goal checkers can be selected from the nav-BT based on a topic subscription. This PR follow the same approach that two previous PRs (planner_selector and controller_selector). 
However, the changes in this PR are deeper because several core components are affected: 
- nav2_controller::Controller, 
- FollowPathAction
- FollowPath.action. 

The most of this code was already tested in some old (incomplete) PR last November: https://github.com/ros-planning/navigation2/pull/2108. The main difference with that attempt lie in "the input" for the controller server was the parameter server (as in ros1 - dynamic_reconfigure). The presented solution, instead, uses the FollowPath.action message (goal_checker_id field) to select the current goal_checker and there is a GoalSelector BTAction node that interacts with the "external ROS network", keeping the "communication" with the controller_server  action server "private" in the context of navigation.

| Info |  |
| ------ | ----------- |
| Robotic platform tested on | (ROS2 nightly and Gazebo simulation turtlebot waffle) |

---

* reading from topics to select the goal checker
* support for default controller (as static input port)
* support to setting the name of the topic to support multiple controller selectors.

**Requirements**

- goal checker selector node
- the follow path should have an input port for goal checker to use
- follow path action server should have a goal checker ID field
- controller server should use that field to use a specific goal checker on a specific request, the same way we do for controller plugins
- controller server should have a map of goal checker plugins

---

#### For Maintainers: 


SteveMacenski:
I'll get to this later today / tomorrow, but should then the old PR be closed? Looks like its a full rewrite from that

pabloinigoblasco:
> I'll get to this later today / tomorrow, but should then the old PR be closed? Looks like its a full rewrite from that

Yes, the old one is now obsolete and can be discarded. I already reused some part of those changes (so that effort was good).

pabloinigoblasco:
Today I will use this PR code to check the implemented code in an example application using gazebo and turtlebot.

I will start from the default nav-bt xml to add the new core elements and test them working together.
I think this could be considered some kind of "integration test". Once my application is working (that makes some intensive usage of these new features) I will add the changes to the default nav-bt xml (or alternatively we can add another different xml file example)

I guess that once this application demo is working it should be safe to totally accept these changes that affect to some core elements such as the controller_server or the FollowPathAction. If I find something weird I will commit some new changes on this PR.

I also will have a look to that small test/linting error.

SteveMacenski:
Sounds good - it would be good to also have a BT demo that is utilizing this / the other selectors. Consider adding the selector nodes to the main BT XML and 2 goal checkers to the configuration file (e.g. `precise` and `general` or what have you) 

Check CI, there are some errors to fix.

nav2_behavior_tree/plugins/action/follow_path_action.cpp
```diff
@@ -46,6 +47,22 @@ void FollowPathAction::on_wait_for_result()
     goal_.path = new_path;
     goal_updated_ = true;
   }
+
+  std::string new_controller_id;
+  getInput("controller_id", new_controller_id);
```
SteveMacenski, line 52: I don't know how I feel about this one - I don't  think the controller / goal checkers should be able to be changed mid-execution. The goal checker I could almost make an argument for - but the planner ID is a bit more suspect. Can you explain the thinking here?

Also, why the `|=`? Those are the same as `=` for this situation (e.g. `goal_updated_` will always be `true` b/c the hardcoded-`true`)

nav2_bringup/bringup/params/nav2_params_selectable_components_test.yaml
```diff
@@ -0,0 +1,317 @@
+amcl:
```
SteveMacenski, line 1: I don't think you meant to commit this file? I think you could add the 2 to the main config file (though named better!) 

Either way, the config files in this directory will need an update for the change in type of the goal checker parameter to a vector

SteveMacenski (changes requested):
Otherwise, LGTM in the broad strokes!

nav2_behavior_tree/plugins/action/follow_path_action.cpp
pabloinigoblasco: Hello. I have been out for some days.

Concerning the second question:

- Maybe I am missing something. From my understanding when goal_updated_ == true  the async_send_goal is executed. This change means that a new async_send_goal request is done if any of the inputs changes (goal or goal_checker or controller). It might be the case where the goal keeps the same but the goal checker changed, I think that information must be notified to the controller server. What I am missing?

First question: I more or less understand your concern. But IMHO, if we are working with topics we need this ability. Let me explain that:

**"The transactional configuration issue"**
Since controller_id, planner_id and goal_checker id inputs are topics if your application wants to use some "specific combination" you will face the issue of "the order". Input configuration topic messages may arrive unsorted and with some gap in time. 

The ideal behavior I think should be able to support the case of  "transactional one-shot configuration". Our solution is based on topics and if we do not allow to dynamically update planer, controller and/or goal checker we can face the situation where only one of them is updated and the FollowPathAction already started (what happens with the rest?)

From my understanding this issue may be "fair enough solved" using topics If we accept that controller, planners or goal checkers can be updated at any moment, even preempting the current execution.
The way I imagine the topics is used is keeping some kind of "continuous flow" of input messages showing the selected planner, controller and goal checker.

When I suggested the usage of services or actions in https://github.com/ros-planning/navigation.ros.org/issues/147 to mimic dynamic reconfigure.) I was thinking on this issue. I should have insisted about this.

What is your point of view this challenge? Is it something I am not seeing?

nav2_bringup/bringup/params/nav2_params_selectable_components_test.yaml
pabloinigoblasco: Indeed, I did not mean to add anything not related with our changes. I will fix that.
Just tell me a better name for this file (I am not very good on naming).

Correct, main files also will need changes to support a vector for goal checkers. That is pending.

nav2_behavior_tree/plugins/action/follow_path_action.cpp
doisyg: My 2 cents, I generally think that it makes senses to expose the goal fields of an action in the BT action node calling it.
But to allow an update of these fields just feels weird to me (even for the `path`, though I understand why it is useful). After all, what a BtActionNode is doing underneath is to preempt the previous goal and executing a new one when `goal_updated_ == true`.
I feel that the same result could be achieved at the tree level with condition nodes and reactive fallbacks/sequences. 
(though to be logical, I should argue the same for `path`, propose to get rid completely of the `goal_updated_` logic and demonstrate how to do without it)

pabloinigoblasco: The concept of preemption is usually annoying, I agree. But please, take into account the example I described. I think the concept Is very simple "making the navigation configurable from the ros network" (planner, controller and goal_update - specially the last two together)

Could you please provide a mock example of a xml bt-tree that did that? The only way I see if we do not accept preemption is introducing some BTAction that stores the updates of all the and update everything in one shot.

nav2_bringup/bringup/params/nav2_params_selectable_components_test.yaml
pabloinigoblasco: I updated the yaml files taking into account your notes.

Besides, the debugging of the solution is still pending but it is in progress.

nav2_bringup/bringup/params/nav2_params.yaml
```diff
@@ -96,7 +97,8 @@ controller_server:
     min_y_velocity_threshold: 0.5
     min_theta_velocity_threshold: 0.001
     progress_checker_plugin: "progress_checker"
-    goal_checker_plugin: "goal_checker"
+    goal_checker_plugins: ["precise_goal_checker", "general_goal_checker"]
+    default_goal_checker: "precise_goal_checker"
```
SteveMacenski, line 101: As far as I can tell, this is not a valid parameter read in anywhere or utilized.

SteveMacenski: There's also a question if this should even be a parameterization. The controller plugins don't have a default when multiple exist; it simply rejects all requests lacking specification if > 1 exists on the server.

pabloinigoblasco: It is fine to me to follow that approach you say. No problem with that on my side.

The only minor issue I see is that the goal_checker takes maybe a bit more importance that It should have for the general case. (I think)

What abut the cases where the goal_checker selector is not being used?

nav2_behavior_tree/plugins/action/follow_path_action.cpp
SteveMacenski: I'm specifically asking 2 questions:
- Why are you using `|=` and not just `=`, it doesn't make any semantic difference to this application?
- If we should allow for changing controller ID / goal checkers mid-execution of the action.

---

Changing / selecting the planner / controller / algorithm makes alot of sense, so that when it gets updated **the next task** gets to use that algorithm. I did not intend this to be dynamically remapping in the middle of a running task. None of the logic of the controller server / action pipeline from the BT has been evaluated under those conditions. 

The planner doesn't really matter, you can remap every tick of the BT and that's OK since its a short running process that should be returning near immediately. The controller is unique that it is long-running. I'm not sure design-wise it would be smart to allow something like these to be changed in the middle of the navigation task run. Between preemption requests or goal cancellations, sure, because you're changing the nature of the request itself. You handle this in `on_tick()` that the new selected algorithms will be utilized for a new run. But changing the algorithm in the middle of its task run for the controller feels like such a fundamental change to the navigation request that doing so without a preemption/cancel feels off to me.

To be clear, if your argument is only that you want to enable this because you want to _guarantee_ that `new_controller` is used for the next task rather than `original_controller`, then maybe using services to get a response that the update was made is the most logical choice. If your argument though is that in the middle of a task from `A->B` you'd like to change the controller from `original_controller` to `new_controller` at any time, for any reason, anywhere, that's the "this feels wrong to me" starts to creep in. I don't think I'd ultimately block it if it was something you wanted to do, but I really want to have a discussion about this and make sure we all make smart choices. 

Not here nor there, but using dynamic reconfigure at run-time like that is a big no-no. Having done that in ROS1 wasn't a "solution" it was a hack. That was meant for local parameter tuning, not for use in steady-state and that was made pretty clear through its stability / documentation. 

---

@doisyg:  `goal` is special because that's how we communicate the action information to the behavior tree. `path` isn't anything special, its just a blackboard variable set as the output port of the planner and the input port of the controller BT nodes. There's no remapping / "changing" of these, that's just how the BT nodes talk to each other and a proper design pattern for blackboard/behavior tree communication. Even then, goal isn't actually hardcoded anymore, its all parameterizable and dynamic variably set in the XML. 

Also, @pabloinigoblasco wants to change the algorithm based on an external application. I 100% agree most people should use the BT contexts to set algorithms but he doesn't want to encode this information in the BT itself but from an external application (that's why we made the selector nodes).

pabloinigoblasco: > - Why are you using |= and not just =, it doesn't make any semantic difference to this application?

You are right. My brain wires crossed here, I was thinking on the &=true case.

> - If we should allow for changing controller ID / goal checkers mid-execution of the action.

IMO, If the controller or the goal_checker change the nav_controller action server should be preempted. Are we aligned on this? 
Otherwise, if we have to wait for the FollowPath action to complete, even if these inputs change, there will be a race condition between the selected goal_checker and the selected controller. (you usually want to configure both together) 

> Changing / selecting the planner / controller / algorithm makes alot of sense, so that when it gets updated the next task gets to use that algorithm. 

Yes.

> I did not intend this to be dynamically remapping in the middle of a running task. None of the logic of the controller server / action pipeline from the BT has been evaluated under those conditions.

> The planner doesn't really matter, you can remap every tick of the BT and that's OK since its a short running process that should be returning near immediately. The controller is unique that it is long-running. I'm not sure design-wise it would be smart to change things like these in the middle of the navigation task run. Between preemption requests or goal cancellations, sure, because you're changing the nature of the request itself. 

> You handle this in on_tick() that the new selected algorithms will be utilized for a new run. But changing the algorithm in the middle of its task run for the controller feels like such a fundamental change to the navigation request that doing so without a preemption/cancel feels off to me.

Maybe I am not grasping something. From my perspective we are just preempting the nav_controller action server.

Do you mean here that we should cancel the action request and finish the FollowPath task if some of the inputs change? If so, that also sounds good to me.

> To be clear, if your argument is only that you want to enable this because you want to guarantee that new_controller is used for the next task rather than original_controller, then maybe using services to get a response that the update was made is the most logical choice. 

I think that topics can be enough to solve the most general case. It is true that we do not get any feedback if something failed on the selection though.

> If ​your argument though is that in the middle of a task from A->B you'd like to change the controller from original_controller to new_controller at any time, for any reason, anywhere, that's the "this feels wrong to me" starts to creep in. I don't think I'd ultimately block it if it was something you wanted to do, but I really want to have a discussion about this and make sure we all make smart choices.

From the point of view of the nav_controller I think it is equivalent to keep alive the same FollowPath action or finish that task and create a new one. The result is that the current action is preempted.
Aren't we just sending a new request to the action server if goal_updated_ == true? (preempting)

Nonetheless, if your concern is that the FollowPath action should finish when the goal_checker or the controller change, that is also good to me. Is that what you mean?

> Not here nor there, but using dynamic reconfigure at run-time like that is a big no-no. Having done that in ROS1 wasn't a "solution" it was a hack. That was meant for local parameter tuning, not for use in steady-state and that was made pretty clear through its stability / documentation.

I did not mean at all to use a dynamic_reconfigure solution, I just meant that ROS1 provided a solution that solved the issue of that potential "race condition" and we should find a solution that takes that into account.

nav2_bringup/bringup/params/nav2_params.yaml
SteveMacenski: Its no different than the controller / planners before the selectors were created - its hardcoded per request or set via a blackboard variable

nav2_behavior_tree/plugins/action/follow_path_action.cpp
pabloinigoblasco: I think you know better how to define some kind of interruption with Behavior Trees. I have a few ideas but all of them are quite dirty.

The other alternative is to make a "buffer" BTNode with all the information to update and commit them in one shot. But that would make the BT more dirty and would show that the current selector approach is limited.

SteveMacenski: > IMO, If the controller or the goal_checker change the nav_controller action server should be preempted. Are we aligned on this?

I don't understand the statement.

> From my perspective we are just preempting the nav_controller action server

Yes, which is what @doisyg and I are concerned about. We don't think that it necessarily makes sense to make it allowable to change that information mid-execution by dynamically set blackboard variables from the selector. The impression I was under was that the selector would be used to change the algorithms between preemptions, requests, cancellations, etc. I did not think about using the BT nodes to change it mid-execution without notifying anyone mid-request.

I'm more in favor of allowing for it than disallowing it, but it seems like a real potential for danger if the selector nodes are midused

pabloinigoblasco: I understand your concerns but I think that safety issue you say already may happen now.

For example, If your goal checker only checks position and not speed, you may enter in another FollowPath task with a different controller that starts the navigation in some "unsafe" initial state.

SteveMacenski: > I think that if you are really concerned about that safety issue it would be needed to introduce some safety checks (that you could optionally disabled) in the FollowPath action or in the nav_controller action server.

Or just not allow changing the algorithms mid-execution :wink:, but I'm leaning towards your side of things. @doisyg what do you think? Anything here convince you that this should be allowed? I'm leaning towards "yes" to allow it, even though I think its a bad idea to use it. I suppose its not really my place to tell people what they should/shouldn't do as long as it doesn't break general behavior.

pabloinigoblasco: > Or just not allow changing the algorithms mid-execution wink

haha right! 

I am having a few more ideas (I do not want to spam though):

I see that one of the ways to finish the FollowPath is executing the cancel of the action request. What if the controllerSelector or the goalChecker or any other node sends cancel() to the nav_controller?

Some BT Pseudocode:
```
<ControllerSelector selected_controller="{selected_controller}>
<GoalCheckerSelector selected_goal_checker="{selected_goal_checker}">

<Parallel finish on any>
    <FollowPath controller_id="{controller}" goal_checker="selected_goal_checker"/>
    <InterruptFollowPath onchange="{selected_controller}"/>
    <InterruptFollowPath onchange="{selected_goal_checker}"/>
</Parallel>
```

SteveMacenski: I think that could work for a custom BT of a user's design (probably not good for the default user BT) mostly just because its overkill for the average person -- but I suppose at that the selector nodes aren't even in the default BT (yet?). 

What is the goal you're trying to accomplish, specifically? Right now, `on_tick()` in only called on a new nav request (after a cancel) or a new Bt loaded. So that by itself would change the controller plugin algorihtms only when "new" without any additional work. It just wouldn't allow for preemption changes. I'm saying above that I'm OK with how it is now, as long as you are aware that your controller plugins need to be able to handle instantaneous changes if you use it in that way.

pabloinigoblasco: > What is the goal you're trying to accomplish, specifically? 

The race condition: "changing in one controller and goal_checker". I need the FollowPath is restarted if the controller or the goal checker is updated.

> I'm saying above that I'm OK with how it is now, as long as you are aware that your controller plugins need to be able to handle instantaneous changes if you use it in that way.

Nonetheless, if the original proposal is fine, I am fine.

SteveMacenski: I'm not sure I see a big issue with a race condition, what's the issue you're trying to avoid? I don't think cancel/request on a change of controller ID / goal checker is necessary. I think if I were designing an industrial / consumer robot navigation system, I'd want the opposite: for a cancel/request, I'd want the newest controller ID / goal checker. See what I mean?

pabloinigoblasco: > I'm not sure I see a big issue with a race condition, what's the issue you're trying to avoid?

This issue:
image


The previous case would be an incorrect execution, because the decision making about "the mode of navigation" should be executed in one shoot.

> I'd want the opposite: for a cancel/request, I'd want the newest controller ID / goal checker. See what I mean?

I agree with this sentence. But I think that the issue described in the previous drawing is still there.
 
> This feature as is works when it works, the problem is that many people might not understand the stability of swapping controllers on the fly at full speed while in an existing navigation request.

Indeed, any controller switching may be dangerous. With or without the controller selector.

SteveMacenski: Well if you publish the selector topics before the action request, you're probably in good hands that things will work out but this is the nature of the beast of doing things this way vs in subtree contexts like we suggest. I can't really offer much help as to what to do here at that point because it starts to dig deeply into your specific systems.

pabloinigoblasco: >  This is also why we allow folks to specify their own BT XMLs versus only the default ones we can design for the typical user

This was the discussion for months, I see your point of view totally legit. But I think it is also legit to use "the navigation stack as a remote service to request motions tasks". This approach also totally fits the philosophy of work of a distributed framework as ROS2.

Discussions aside, lets try to go forward with a solution that make everybody happy.  

I would like to suggest a couple more of ideas that can make the selectors better and may remove the need of dynamic blackboard variables and solve some of the problems we are discussing. I am not just thinking in my case, but in a general and flexible usage of the selectors.

If they are not accepted it is okay, we can keep the current basic implementation (I can even remove that preemption of the nav_controller if feel better with that). Nonetheless I would like to know your opinion about them:

**Service inputs**
After the discussion I now think the input of the selectors should move from topic to service:

 - That may be good for a better the synchronization in the client that needs to configure different parameters. The client can verify the configuration was properly received before sending the final goal request.
 - Service result could be also be used to provide some feedback about the success/failure of the change (maybe checking if the controller/planner/goal_checker exists? maybe rejecting the change if the nav_controller is busy?) 
   
**Blocking selectors**
I would also introduce in the selectors a blocking attribute. With this approach you could define optional or mandatory configurations that must be specified mandatory for each NavigateToGoal request.  
That would be also good for the case you specify a new bt file in the NavigateToGoal message avoiding race conditions.
 
Something like this:
  
```xml
<btree>
<sequence>
<!-- configuration stage -->
<parallel> 
    <PlannerSelector blocking="true" .../>
    <ControllerSelector blocking="true" .../>
    <GoalChecker blocking="true" .../>
</parallel>
<-- navigation stage-->
<ComputePlan ...>
<FollowPath .../>
</sequence>
</btree>
```

SteveMacenski: >  lets try to go forward with a solution that make everybody happy

I thought I already seeded this point to you that we could keep the dynamically set controller ID / goal checkers? My only comment was that its on the user to make sure that your controllers have some stability to operate in that fashion. As long as the controller makes use of https://github.com/ros-planning/navigation2/blob/d8bd3c5ea8d196ad7f0bace1edaeaa8e55f06944/nav2_controller/src/nav2_controller.cpp#L386 the inputted current velocity, I think things should be OK for the most part. I know for instance RPP will, not I can't speak to the others.

> Service inputs

Services make a ton of sense on a first glance. This is exactly what they do well. I did not explicitly suggest that because of the spinning issue to process requests. We'd either have to create a new node on the network to handle those which would be alot of overhead or a separate single threaded executor spinning a callback group (like the service client in utils was changed to recently). The second is the subject of https://github.com/ros-planning/navigation2/issues/2251 which I'm still not 100% certain if creating the new executor adds a new DDS participant to the network (which is ~75% the overhead of just a new node). If I knew for certain one way or another, I could discount or immediately embrace the service idea. 

> Preload behavior tree service

That's a good point on race conditions if its even alive, I didn't think about using selectors in the case of also changing the BT in the `NavigateToPose.action`. One way you could actually help resolve that is to have your publisher of the selector **and** the subscriber be transient-local. that way you get the "latched" behavior of ROS1 so that if a message is published **after** the subscriber is created, the message still makes it there. I think that would largely resolve the concern that this suggestion is fixing. I'd actually suggest adding transient local QoS to the subscribers in all 3 of these right now from that motivating example. 

> Blocking selectors

How different would that be from specifying non-sensical default values in the selector node? If they didn't make sense, then when one was requested, then it would fail, and fail the request the same way something blocking would also fail the request. Though semantically that seems dubious. A boolean for `blocking` that is default to `false` would be acceptable. Alternatively, we could also make the default algorithm be an empty string. And when its an empty string, we enable `blocking` behavior because clearly that is not a valid algorithm to select. That would implicitly do blocking when no default is acceptable without adding a new parameter. It would only require a minor update to the code and an update to documentation mentioning this is the fact for the default plugins ports.

Tl;dr
- I like services too, but I'm not sure about #2251, if you knew for a fact (or could confidently determine it) that it didn't add a new DDS participant due to the new context, then that's the best solution
- QoS for all 3 to `rclcpp::KeepLast(1)).transient_local().reliable()`, and your publisher being transient local would be good to make sure even in race conditions of a new BT that it was received 
- Blocking is a good feature idea! Though I think we can do it if the `default` port is `""`, then block / fail. Does blocking implicitly when there is no valid default value which I think is your intent without adding a new port.

nav2_bringup/bringup/params/nav2_params.yaml
pabloinigoblasco: In fact, that parameter was not being used in nav_controller.cpp.
Now it is removed in the current version.

nav2_bringup/bringup/params/nav2_params.yaml
```diff
@@ -96,7 +97,7 @@ controller_server:
     min_y_velocity_threshold: 0.5
     min_theta_velocity_threshold: 0.001
     progress_checker_plugin: "progress_checker"
-    goal_checker_plugin: "goal_checker"
+    goal_checker_plugins: ["precise_goal_checker", "general_goal_checker"]
```
SteveMacenski, line 100: I really like the idea of having multiple here as an example, however since this is `> 1` and the BT nodes don't specify any, all nav requests should fail (and if CI isn't failing, we have a bug) since it wasn't specified. 

We should change this to `goal_checker_plugins: ["general_goal_checker"] #precise_goal_checker"` and comment out the `precise_goal_checker` below. I'd like to keep them in the yaml to make the point but shouldn't be commented in just yet

SteveMacenski:
2 linting:

```
whitespace/forcolon [2] (/opt/overlay_ws/src/navigation2/nav2_controller/src/nav2_controller.cpp:70) - nav2_controller.cpplint

Missing space around colon in range-based for loop


whitespace/line_length [2] (/opt/overlay_ws/src/navigation2/nav2_controller/src/nav2_controller.cpp:320) - nav2_controller.cpplint

Lines should be <= 100 characters long
```

What are the next steps here?

SteveMacenski:
Note: This PR will be in conflict with #2271, so I'd like to get this PR in before mine so I have to rectify it rather than you.

nav2_behavior_tree/plugins/action/follow_path_action.cpp
pabloinigoblasco: I read carefully everything you described. Thanks for the detailed answer.

> - I like services too, but I'm not sure about #2251, if you knew for a fact (or could confidently determine it) that it didn't add a new DDS participant due to the new context, then that's the best solution

Okay. I understand that this must wait, but lets keep an eye to this since it looks to be the best solution.

> - QoS for all 3 to rclcpp::KeepLast(1)).transient_local().reliable(), and your publisher being transient local would be good to make sure even in race conditions of a new BT that it was received

Okay, this is something I am about to commit and push. However, I think this is not a solution for the race condition, maybe it is for the first navigation request, but I think it is not a solution second motion request. (see pseudocode below)

> - Blocking is a good feature idea! Though I think we can do it if the default port is "", then block / fail. Does blocking implicitly when there is no valid default value which I think is your intent without adding a new port is "", then block / fail. Does blocking implicitly when there is no valid default value which I think is your intent without adding a new port.

I think this need a bit more discussion because I am not sure if we are thinking in the same solution. 
What is blocking for you? 
- Rejecting the navigateToGoal request? 
- With "blocking" I mean "wait until you receive a new input". That is: replacing the navigation request from "one shot action" to some "interactive protocol between the user and the action server". I think this "weird" concept make the point of view different.

My main concern is still the race condition. I think that "reject the request" only would solve the issue for the first time.
Let me show the example I have in mind "two consecutive navigation requests":

```
- select planner_A
- select controller_A
- select goal_checker_A
[Here is the race condition, the request returns fail if elements were not properly selected,
Lets assume this first attempt worked fine]
- navigate to goal X1
- wait to finished navigation
[Navigation returns success]

[Now second motion request]
- select planner_B
- select controller_B
- select goal_checker_B
[Race condition, select_controller_B was late and controller_A was still selected]
- navigate to goal X2
- wait to finished navigation
[Navigation returns success]
```

The second NavigateToGoalPose request would work if select_controller was blocking. On the other hand, if we just return fail if the "controller is not known" would end up in some incorrect navigation execution of the second motion.

The "workaround" I am having in mind is just using `Wait(seconds)` between the selections and each navigate request. But that is ugly and blind (the client never knows if things are all right). Services would have been enough (I know they are frozen)

I think this race condition is not only something of my case. I think it is the nature of the solution we are providing for selectors and may affect to anyone using them.

pabloinigoblasco:
Just to confirm.

Do we then agree that with the "dynamic blackboard variables" we can get two FollowPath requests to the nav_controller from one single FollowPath BTActionNode in one single NavigateToGoal request?

pabloinigoblasco:
last pushes include the following changes:
- updates on the navigation yaml to leave only the general goal checker and comment the precise one
- removal of the |=
- conflict fix on nav_controller.cpp
- fixing some linting errors
- latching behavior qos for the three subscribers

nav2_bringup/bringup/params/nav2_params.yaml
```diff
@@ -106,11 +107,16 @@ controller_server:
...
+    #  yaw_goal_tolerance: 0.25
+    #  stateful: True
+    general_goal_checker:
       plugin: "nav2_controller::SimpleGoalChecker"
-      xy_goal_tolerance: 0.25
-      yaw_goal_tolerance: 0.25
-      stateful: True
+      xy_goal_tolerance: 0.35
+      yaw_goal_tolerance: 0.35
+      stateful: False
```
SteveMacenski, line 119: Looking at the diff, the "general" one was stateful true and the tolerances were both `0.25`, can we change back to that?

nav2_controller/src/nav2_controller.cpp
```diff
@@ -68,7 +68,13 @@ ControllerServer::ControllerServer()
 ControllerServer::~ControllerServer()
 {
   progress_checker_.reset();
-  goal_checker_.reset();
+
+  for (auto & goal_checker: goal_checkers_) {
+    goal_checker.second.reset();
```
SteveMacenski, line 73: This will be done automatically in L76 when you reset the entire vector container, the destructors for each unique pointer will be called

SteveMacenski (changes requested):
2 small changes and LGTM

SteveMacenski:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/5083/workflows/312e3591-c908-4e8f-8604-e12f62cdc119/jobs/18766 you have a number of test failures / linting issues

nav2_behavior_tree/plugins/action/follow_path_action.cpp
SteveMacenski: On blocking: what I suggest is:
- Make the default planner / controller / goal checker IDs an empty string
- If the current is empty string in the `tick` because non set and default was an empty string, then nothing has been set, fail the bt status -- blocking motion
- If the current is non-empty string in `tick`, then something was set and no need to block
- If the default was set to a non-empty string in the XML, that disables blocking because there is a logical default

So without a specific `blocking` parameter, we can get that behavior by just checking the current algorithm ID and the default (if set to empty to enable -- meaning semantically that there IS no default and there MUST be a value set to pass). 

---

If you're concerned about bleeding between runs, you could specify a BT to use in the action request so that a new tree is loaded - though that only works if you're constantly changing trees between requests. I suppose we could have a `force_reload` parameter / action request field so that the BT is reloaded every time a navigation request comes in to prevent that.

Another options with the selector nodes is that you can override the `halt` function to reset the current back to empty string (if the default is empty string to enable blocking) when the BT exists so that each run will be a clean state execution. That's probably the cleanest solution to this particular problem, but the `force_reload` is the most general to solve all such problems in the future (though very inefficient and adds latency)

pabloinigoblasco:
Some of the test failures are because of the configuration of that "latch behavior" of the subscriber. 
I think something is incorrect.  Is this the way this should be configured?

```
rclcpp::QoS qos(rclcpp::KeepLast(1));
qos.transient_local().reliable();

controller_selector_sub_ = node_->create_subscription<std_msgs::msg::String>(topic_name_, 1, std::bind(&ControllerSelector::callbackControllerSelect, this, _1));
```
One question around, I still do not control the qos details in ros2.
Shouldn't that latch behavior be configured on the publisher?

By the moment I disabled that latch behavior, replacing again back qos to 1.

nav2_bringup/bringup/params/nav2_params.yaml
pabloinigoblasco: This should be already done.

nav2_controller/src/nav2_controller.cpp
pabloinigoblasco: Okay, I removed that part.
I also removed the part of the goal_checker_ids.clear() to totally mimic the usage of the controllers vector (we did not have controller_ids.clear())

pabloinigoblasco:
I keep doing tests on these features on actual demos.

SteveMacenski:
>  I think something is incorrect. Is this the way this should be configured?

No, you're not using the qos object, you have to give it to the subscriber. Also the test publishers should also be set to use that same QoS (probably why the test is failing, incorrect QoS) ex:

```
rclcpp::QoS qos(rclcpp::KeepLast(1));
qos.transient_local().reliable();

controller_selector_sub_ = node_->create_subscription<std_msgs::msg::String>(topic_name_, qos, std::bind(&ControllerSelector::callbackControllerSelect, this, _1));

```

Yes, the publisher needs to also be latched too.

pabloinigoblasco:
Sorry, the code I put is missleading.
Originally the qos was assigned to the subscriber, but I disabled that (and restore the value 1) because that was making the test fail.

I will try to check passing the same parameters to the publisher to see if test passes.

SteveMacenski:
OK - let me know when you want me to take another look @pabloinigoblasco 

The tests seem to be passing now.

pabloinigoblasco:
Yes, I've been with some other issues these last days, but I am validating this initial iteration we did (that should work) and once it works we can do this last step you say.

The last merge we did broke my environment and delayed me a bit, but they are about to be working again.

SteveMacenski:
Got it, thanks for the update!

pabloinigoblasco:
I just introduced in the last commit the changes about returning FAILURE if default planner was not set and selection message was not received.

I also did quite intensive testing on a simulation/application that uses the navigation stack to request different combinations of planner, controller and goal_checker. No race condition happened.

Something I feel like it is missing is doing some cleanup after the action execution to reset PlannerSelector::last_selected_planner_, ControllerSelector::last_selected_controller_ and GoalCheckerSelector::last_selected_goal_checker_

pabloinigoblasco:
Hello
How is the acceptance of this last version going?

SteveMacenski:
Sorry, I was very busy last week and didn't get to this.

> Something I feel like it is missing is doing some cleanup after the action execution to reset PlannerSelector::last_selected_planner_, ControllerSelector::last_selected_controller_ and GoalCheckerSelector::last_selected_goal_checker_.

You can override the `halt` function to reset these values if you want, but I think its fine without this - I could see some people wanting both behaviors

nav2_behavior_tree/plugins/action/controller_selector_node.cpp
```diff
@@ -36,16 +36,25 @@ ControllerSelector::ControllerSelector(
 
   getInput("topic_name", topic_name_);
 
+  rclcpp::QoS qos(rclcpp::KeepLast(1));
+  qos.transient_local().reliable();
```
SteveMacenski, line 40: When you have the PR for the goal checker docs, make sure to see that there expecting the QoS to be reliable and transient local. Add that to the other selector configuration pages as well so its not a surprise to folks when things don't connect as expected without changing QoS

nav2_behavior_tree/plugins/action/controller_selector_node.cpp
```diff
@@ -36,16 +36,25 @@ ControllerSelector::ControllerSelector(
...
 
 BT::NodeStatus ControllerSelector::tick()
 {
   rclcpp::spin_some(node_);
 
   if (last_selected_controller_.empty()) {
-    getInput("default_controller", last_selected_controller_);
+    std::string default_controller;
+    getInput("default_controller", default_controller);
+    if (default_controller.empty()) {
```
SteveMacenski, line 53: A comment here about what this is implementing would be great. Something just letting people know its so that if no default is set & none received failing. If you want to proceed without a set, must set a default. Would also be good context to add to the configuration page for this variable.

nav2_behavior_tree/plugins/action/goal_checker_selector_node.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+BT::NodeStatus GoalCheckerSelector::tick()
+{
+  rclcpp::spin_some(node_);
+
+  if (last_selected_goal_checker_.empty()) {
+    std::string default_goal_checker;
+    getInput("default_goal_checker", default_goal_checker);
+    if (default_goal_checker.empty()) {
```
SteveMacenski, line 53: A comment here about what this is implementing would be great. Something just letting people know its so that if no default is set & none received failing. If you want to proceed without a set, must set a default. Would also be good context to add to the configuration page for this variable.

nav2_behavior_tree/plugins/action/planner_selector_node.cpp
```diff
@@ -36,16 +36,25 @@ PlannerSelector::PlannerSelector(
...
+    topic_name_, qos, std::bind(&PlannerSelector::callbackPlannerSelect, this, _1));
 }
 
 BT::NodeStatus PlannerSelector::tick()
 {
   rclcpp::spin_some(node_);
 
   if (last_selected_planner_.empty()) {
-    getInput("default_planner", last_selected_planner_);
+    std::string default_planner;
```
SteveMacenski, line 51: A comment here about what this is implementing would be great. Something just letting people know its so that if no default is set & none received failing. If you want to proceed without a set, must set a default. Would also be good context to add to the configuration page for this variable.

SteveMacenski (commented):
LGTM, I think we're just missing some context in the configuration page around QoS & the empty set default algorithm IDs, but otherwise good to go

pabloinigoblasco:
> Sorry, I was very busy last week and didn't get to this.
> 
> > Something I feel like it is missing is doing some cleanup after the action execution to reset PlannerSelector::last_selected_planner_, ControllerSelector::last_selected_controller_ and GoalCheckerSelector::last_selected_goal_checker_.
> 
> You can override the `halt` function to reset these values if you want, but I think its fine without this - I could see some people wanting both behaviors

I am afraid that SynchronousActionNode::halt() is final and cannot be overridden.

Maybe we can leave it in the way it is now if you feel it fine. The selected element can be "deselected" manually from the decision level. 

It is a bit weird though, the current implementation looks more like a persistent "node parameter or a dynamic property" instead of part of the behavior tree execution (since previous execution play some role).

What do you think about this?

SteveMacenski (approved):
Perfect, I know this is a long time coming so I appreciate your diligence and helping push this to the end. Assuming the documentation is good, this will be merged definitely this week.

SteveMacenski:
Docs PR merged, just need tests to pass here from linting

SteveMacenski:
@pabloinigoblasco still failing on linting :-)

pabloinigoblasco:
> @pabloinigoblasco still failing on linting :-)

Oups! okay! I am on that!
:-)

Issue #2270, opened 2021-03-26
[Input Requested] Removal of launch configuration of default BT XML


SteveMacenski:
Currently, there are 2 ways of setting the default Behavior Tree XML file:
- Via the commandline-remapping in https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/bringup/launch/tb3_simulation_launch.py#L41 launch
- Via configuration files https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/bringup/params/nav2_params.yaml#L59

Up until now, we haven't had any default set, so the commandline / launch file method gave us a clear way to set a default using the shared file paths to the BT navigator package (since these must be complete filepaths). However, C++ now has this ability and I have a PR coming soon which will use it to set the default behavior tree XML files in BT navigator itself.

As such, I'm looking to simplify our launch files for readability and I'm not sure remapping in this way is even used by any of our users. I'm looking for information from folks if they're OK with me removing the launch configuration in the launch file and have the BT XML file set via the param file parameterization only. This parameterization would need to be the full path to the XML file (as would the commandline setting). However, the BT navigator would also have a reasonable default behavior tree XML set using the `ament_cpp_index` package to find the `nav2_bt_navigator` package and its behavior tree files, which it currently does not. 

Therefore, if you are not using a custom behavior tree, nothing would change for you as an end user. If you did change the BT, you'd have to do so through the param file.

joe28965:
> This parameterization would need to be the full path to the XML file

In the launch file, this is done using `get_package_share_directory` to get the completed path. I am not aware of a way to do this in the param file (please excuse me if I'm mistaken). Would this not cause issues for multiple people running the same code on their own computers?

SteveMacenski:
We do this in C++ in the BT node now (as of #2271) for the default trees to find the location. The parameter file requires a `full/path/to/bt.xml` as it did before as well.

But that has nothing to do with this request. That was non-remappable. That only worked for the default tree (and that logic was just shifted into the BT navigator node itself). The `get_package_share_directory` was never used for user remapped values of the default BT XML, there you also have to specify your own `full/path/to/bt.xml`.

joe28965:
I'm sorry, I am afraid I phrased it rather poorly.

[...]

PR #2271, opened 2021-03-26
Navigate through poses action ability

SteveMacenski:
Creates:
- new navigate through poses navigator
- abstract out this logic to allow for expansion (but not plugins)
- create a new onComplete callback in the bt action server to populate results meaningfully or indicate completion
- make sure only 1 navigator at a time can be processed with a muxer handled for plugins automatically
- nav through poses cull pts passed so don't replan through old areas via the `RemovePassedGoals` BT node when met a tolerance metric
- fixed an issue allowing a failure in processing the simple action server from making the BT navigator lock up and think its still processing indefinitely.
- added completion callback option to the simple action server in case it terminates on its own from un-handled exceptions thrown
- Inserted all BT nodes into the plugins libs list and in the groot xml
- Set defaults individually for both navigators + updated parameter / launch files accordingly 
- Rename behavior trees to be more descriptive

Questions:
- Seperate into 2 servers or keep in 1 server?

TODO
- [x] Testing: system tests
- [x] config page for new bt navigator params (or moved ones for specific navigators: path_blackboard_id, goals_blackboard_id, goal_blackboard_id, default_nav_through_poses_bt_xml, default_nav_to_pose_bt_xml)
- [x] Testing: unit tests for new BT nodes: compute path through poses, navigate through poses, remove passed goals
- [x] Website: plugins list, migration guide, config guide https://github.com/ros-planning/navigation.ros.org/pull/155
- [x] Website: Create nav BT XML list on website with explanations of the basics of what they are https://github.com/ros-planning/navigation.ros.org/pull/155

#### For Maintainers: 

Related PRs to resolve against as merging
- [x] https://github.com/ros-planning/navigation2/pull/2261
- [x] https://github.com/ros-planning/navigation2/pull/2246
- [x] https://github.com/ros-planning/navigation2/pull/2269 ComputePathThroughPoses different update port / goal checkers


naiveHobo:
I think the implementation of the navigate through poses functionality is absolutely on point. The only thing I find a bit unsettling is the need to keep track of two variables in several places, one for the single goal scenario and one for the vector of goals scenario:

- There are BT nodes which are unrelated to `NavigateToPose` and `NavigateThroughPoses` actions like `SpeedController` and `GoalUpdated` that now have to keep track of two variables for the single and multi goal scenarios.
- Keeping track of two default BT XML files in the nodes, launch files, and params list.

I definitely see why the `NavigatorMuxer` and the `Navigator` interface are useful and I think they're really well implemented. What I'm worried about is that having a `Navigator` interface would suggest to users that that is how they would have to implement a navigation ability for their use-case. To give an example, waypoint following could also be considered as a navigation task with slightly different metrics. Like `NavigateThroughPoses` it accepts a list of goals the only difference being that in case of waypoint following, you need to get to a waypoint first, perform some task, and then move on to the next waypoint. In that case, why not implement waypoint following as another `Navigator` plugin and have it managed by `BtNavigator`?

I feel that the introduction of `NavigatorMuxer` and `Navigator` classes might be unnecessary if the `NavigateToPose` action ability is deprecated in the future (if that's the plan).

In general though, absolutely love this PR. Can't seem to find any faults implementation wise.

SteveMacenski:
What do you think about both being in the same server? I'm kind of on the fence about it. The boilerplate in the bt_navigator.cpp is basically just lifecycle exposure now with very little functional code from the abstraction. Also the `mux` is required if in the same servers so that you don't accidentally attempt to navigate via both interfaces at the same time. 

> SpeedController and GoalUpdated that now have to keep track of two variables for the single and multi goal scenarios.

True, keep in mind `goal` was a special / hardcoded variable before. So to make sure these nodes work with NavThroughPoses, we need to make it look at both sets. It shouldn't be a concern to anyone if we look it up in BTs not containing the other because it won't ever change. Both nodes look to see if the `goal` or `goals` changed to compute some action. For the case that the other isn't used, it'll have no effect. 

> Keeping track of two default BT XML files in the nodes, launch files, and params list.

Agreed, this is annoying. Different servers would fix the params list one at the very least. In the launch files it would likely need to remain unless I could get feedback on https://github.com/ros-planning/navigation2/issues/2270 which asks "hey, can we just get rid of these from the launch file?" and make it only specified in the yaml file. 

If/when we depreciate NavToPose for NavThroughPoses, we can always remove the muxer then and re-simplify this code

SteveMacenski:
https://github.com/ros-planning/navigation.ros.org/pull/158 documentation PR with params

SteveMacenski:
Weeeeee fixing 3 PR's worth of merge conflicts :laughing:

PR #2274, opened 2021-03-29
Enable rmw_connextdds for nightly CI

ruffsl:
To again have a third RMW to triage CI failures

https://github.com/ros2/rmw_connextdds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -452,7 +452,7 @@ workflows:
           matrix:
             parameters:
               rmw:
-                # - rmw_connext_cpp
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I don't object if you need it - but why would adding connext's builds fix anything?

ruffsl:
> but why would adding connext's builds fix anything?

Not directly. We have a few regular flaky tests that fail both for Cyclone and FastRTPS, so I'd l'd like to observe tests with a third RMW to triage if the issues are with the RMWs or something with rcl and nav2.

PR #2276, opened 2021-03-29
[nav2_behavior_tree] Add OnBtExitedCallback 

doisyg:
---

## Basic Info

| ------ | ----------- |

---

Added an optional `OnBtExitedCallback` to `nav2_behavior_tree::BtActionServer`. This allows getting data from the blackboard and filling the result of the action before it returns (and just after the BT is done).
This would be a prerequisite to use and extent the result fields of the `NavigateToPose` action for instance.

---

#### For Maintainers: 


SteveMacenski:
@doisyg I think this can be closed, this was added in my PR https://github.com/ros-planning/navigation2/pull/2271 for navigating through poses capabilities

doisyg:
:facepalm: I should always check the current open PR.
Glad we ended up doing the exact same implementation...

SteveMacenski:
Just needed it for different reasons I'm sure :laughing:

PR #2278, opened 2021-03-30
ServiceClient wait_for_service not infinite

BriceRenaudeau:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [comment](https://github.com/ros-planning/navigation2/pull/2267#issuecomment-809671090) |
| Robotic platform tested on | Wyca (simulation)|

---

* I changed the behavior of wait_for_service to be not blocked forever if the service doesn't exist
* The method returns the result of `client.wait_for_service()`
* I added a unit test to check this behavior

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/service_client.hpp
+++ b/nav2_util/include/nav2_util/service_client.hpp
@@ -123,19 +123,13 @@ public:
   }
 
   /**
-  * @brief Block until a service is available
+  * @brief Block until a service is available or timeout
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

SteveMacenski:
There are some action server test failures from this... unrelated probably, but re-running CI to make sure

SteveMacenski:
Thanks for the contribution!

PR #2279, opened 2021-03-30
[nav2_behavior_tree] Reduce bt_action_node verbose

doisyg:
For my console sanity, switched RCLCPP_INFO to RCLCPP_DEBUG


SteveMacenski:
You should also do the service node https://github.com/ros-planning/navigation2/blob/2c1587e4689fe31b4680f3d685512bfdf00135a8/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp to be analogous

SteveMacenski (approved):
merge when CI passes

PR #1958, opened 2020-08-19
Add groot monitoring behavior tree visualization

gramss:
Gif is loading..
[click for bigger gif](https://gifyu.com/image/cFdR)
---

## Basic Info

| Info |   |
| ------ | ----------- |
| Ticket(s) this addresses   | #1700 #1725 ...? |
| Primary OS tested on | Ubuntu 20.04 |
| Robotic platform tested on | gazebo simulation of turtlebot |

---

## Description of contribution in a few bullet points

* Groot monitor support for behavior trees with custom nodes
* Reorganizing behavior_tree_engine

## Detailed Steps

- [x] Added ZMQ Publisher for BT::Tree in run (moved later), but worked (video in Slack)
- [x] Found bug (in nav2) why Groot crashes with current BT::Tree (manifest from BT::factory was not attached - see first commit)
-- This was already "fixed" / available in BT.CPP V3, but not used in nav2/behavior_tree_engine
- [x] switching / adding / using BT.CPP V3 implementation directly of creatingTreefrom[Text|File]
- [x] Added new parameter `bt_activate_groot_monitoring`
- [x] Added ZMQ_Publisher as unique_ptr protected member variable of `behavior_tree_engine`
- [x] Using parameter and resetting ZMQ_publisher
- [x] Moving ZMQ Publisher: Tree is already monitorable before it is running - after Pose estimation in RVIZ
-- Fixes only partial displaying Status of current Tree
- [x] ZMQ Publisher gets deconstructed when new tree is loaded from XML or lifecycle: on_cleanup()
- [x] Add CI Testing

## Description of documentation updates required from your changes

- [x] Add new parameter to documentation / YAML
- [ ] Add Groot monitoring tutorial (incl. gifs)
- [x] Add Gif to PR... Way too much time wasted here compressing..
- [x] Add all corresponding tickets

---

## Future work that may be required in bullet points

- [x] Test Groot working with switching XMLs
-- searching an easy way to test this (is their a small tool for crafting action calls like in ROS1 availible?)
- [x] Fixing Groot crashing when lifecycle is resetted
-- currently my way to test behavior of a new ZMQ Publisher
-- Pose reached -> Cancel -> Reset -> Groot disconnecting -> StartUp -> NavigateToPose -> Groot Connecting = works
-- Pose reached -> Cancel -> Reset -> Groot disconnecting -> StartUp -> Groot Connecting = crashes
- [x] Adding Parameter ZMQ Publisher Ports
[...]

PR #2291, opened 2021-04-05
foxy Sync 5.1

SteveMacenski:
New sync to foxy:

    Groot monitoring
    New Regulated pure pursuit controller
    various bug fixes and small features like parameterizations


matthews-jca:
For anyone updating to 0.4.7, this merge does have a breaking change in the api. We run a forked BT Navigator to simplify running of our trees. 
We also appreciate that groot is able to monitor too.
The breaking api change is documented here. 
https://github.com/ros-planning/navigation2/pull/1958/commits/8ac97cbf2291665dbaff3baa09c513768a758bac

PR #2296, opened 2021-04-06
[nav2_bringup] Update waffle.model

Jovian-Dsouza:
Changed to the latest tire mesh file names for waffle as per the latest `turtlebot3_gazebo` package.

This results in faster loading and resolves the errors that come in `gazebo --verbose`


SteveMacenski (approved):
Awesome, thanks!

PR #2298, opened 2021-04-06
adjusting new files for tb3 model (ros2 of 2296)

SteveMacenski:
See https://github.com/ros-planning/navigation2/pull/2296 for details

TB3 model has changed the file paths for the tires, requiring people to pull in the new changes `turtlebot3_simulation` repository if building locally. 

@ruffsl for CI, I don't see where it pulls `turtlebot3_simulation` to get these files, can you clarify for me? The underlay workspace doesn't include it and nor does `ros2 master`. We don't base on foxy so it can't be getting those binaries. 

Foxy users we tell to get the binaries in the build instructions. ROS2 main developers have a local copy. Its not clear to me where its coming from for CI. **Should we be including turtlebot3_simulation in the underlay?**

### DIFF
--- a/nav2_bringup/bringup/worlds/waffle.model
+++ b/nav2_bringup/bringup/worlds/waffle.model
@@ -267,7 +267,7 @@
           <pose>0.0 0.144 0.023 0 0 0</pose>
           <geometry>
             <mesh>
-              <uri>model://turtlebot3_waffle/meshes/left_tire.dae</uri>
+              <uri>model://turtlebot3_waffle/meshes/tire.dae</uri>
               <scale>0.001 0.001 0.001</scale>
             </mesh>
           </geometry>
@@ -325,7 +325,7 @@
           <pose>0.0 -0.144 0.023 0 0 0</pose>
           <geometry>
             <mesh>
-              <uri>model://turtlebot3_waffle/meshes/right_tire.dae</uri>
+              <uri>model://turtlebot3_waffle/meshes/tire.dae</uri>
               <scale>0.001 0.001 0.001</scale>
             </mesh>
           </geometry>


ruffsl:
> @ruffsl for CI, I don't see where it pulls turtlebot3_simulation to get these files, can you clarify for me?

I don't think we ever have been? Those appear to be `visual` elements, not a collision geometry, so perhaps gazebo doesn't necessary error when starting the simulation with missing `visual` elements, as long as the `collision` elements have valid `geometry` elements or found mesh files.

> Should we be including turtlebot3_simulation in the underlay?

We could... but because of how ROBOTIS-GIT structures their git repos, there would be some circular dependencies we'd have to skip. I've reduced the dependencies from the `turtlebot3_simulations` repo down to just `turtlebot3_msgs` and `turtlebot3_description` packages:

https://github.com/ROBOTIS-GIT/turtlebot3_simulations/pull/148

But `turtlebot3_description` is located in the `turtlebot3` repo with many other packages we'd ignore: 

https://github.com/ROBOTIS-GIT/turtlebot3/tree/ros2-devel

The minimal set of nodes we'd need is something like this:

```
turtlebot3_description
turtlebot3_fake_node
turtlebot3_gazebo
turtlebot3_msgs
turtlebot3_simulations
```

I've had to do something like this already for the ros-swg/turtlebot3_demo:

https://github.com/ros-swg/turtlebot3_demo/commit/2271b823f0cbf1f90c00b2b4e139e161104bf789

If we don't need the visual elements, I think it would be simplest to avoid adding them into the underlay.

SteveMacenski:
Given we have our own URDFs in the `nav2_bringup` package, I think we mostly just need the `_simulation`'s world files more than anything else, we don't actually require the description. We also don't even need to build it, just as the `GAZEBO_MODEL_PATH` to include the models directory (I set this in my bashrc file locally and have the package `CATKIN/COLCON IGNORE`)

PR #2302, opened 2021-04-07
major smac planner collision checking speedups and improvements

SteveMacenski:
@maxlein in my testing, this makes the full-footprint collision checking nearly immeasurably slower than the radius checking. I tested using the map and footprint you provided and found only a **2% slower** total planning this way. The radius-only planning does so at ~228 search iterations per milliscond and the full footprint checking is 222 search iterations per millisecond (up from ~140) improving performance by **around 40%**. 

Improvements include:
- Precomputing footprint orientations for collision checking
- Stopping collision checking early if already known to be in collision
- Optimizing collision checking code to minimize computations
- Inlining a simplified function from prior optimizations
- Only doing full collision checking if the robot is possibly inscribed (if we know its in free space for instance, there' no point in going through that trouble)  

@maxlein @naiveHobo  can you review?

### DIFF
--- a/nav2_costmap_2d/src/footprint_collision_checker.cpp
+++ b/nav2_costmap_2d/src/footprint_collision_checker.cpp
@@ -84,7 +84,7 @@ double FootprintCollisionChecker<CostmapT>::footprintCost(const Footprint footpr
 
   // we also need to connect the first point in the footprint to the last point
   // the last iteration's x1, y1 are the last footprint point's coordinates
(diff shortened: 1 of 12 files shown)


maxlein:
Sounds awesome, but I don't have the time to try this out right now.  
Will take a few days/weeks.

SteveMacenski:
Pulling PR to merge in with the larger work happening with state lattice

Issue #2273, opened 2021-03-27
Simple action server deadlock


KavenYau:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Foxy binaries
- Version or commit hash:
  - 41996f4
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue

Actually It's similar to this issue #1961

#### Expected behavior
 always accept goals 

#### Actual behavior
the server will not accept goals when the deadlock happens

#### Additional information
- here is the backtrace, https://gist.github.com/KavenYau/562d2180c69d22ee87bfbd5306ced44d

#### backtrace
- Thread 9: main thread
Frame# 6 [nav2_util::SimpleActionServer<nav2_msgs::action::ComputePathToPose, rclcpp::Node>::handle_cancel](https://github.com/ros-planning/navigation2/blob/41996f409fcd7b005cf779e0376273cee3de6d76/nav2_util/include/nav2_util/simple_action_server.hpp#L102)
Frame# 13 [rclcpp_action::Server<nav2_msgs::action::ComputePathToPose>::call_handle_cancel_callback](https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server.hpp#L361)

- Thread 3612: working thread
Frame# 5 [rclcpp_action::Server<nav2_msgs::action::ComputePathToPose>::call_goal_accepted_callback](https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server.hpp#L393)
Frame# 9 [rclcpp_action::ServerGoalHandle<nav2_msgs::action::ComputePathToPose>::succeed](https://github.com/ros-planning/navigation2/blob/41996f409fcd7b005cf779e0376273cee3de6d76/nav2_util/include/nav2_util/simple_action_server.hpp#L316)

#### deadlock
-  Thread 9: main thread
  1. [std::lock_guard\<std::mutex\> lock(goal_handles_mutex_);](https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server.hpp#L355)
[...]

PR #2304, opened 2021-04-08
checking if handle is active in handle_cancel

KavenYau:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |#2273 |
| Robotic platform tested on |Turtlebot3 Simulation |

---

I added `is_publishing_succeeded_state_` and `is_handling_cancel_request_` to mark down current handling state. Make sure that `succeeded_current` and` cancel_handle` can not be handled simultaneously. And I also added `handle_state_mutex_` to protect the two variables.

1. I think all invailed state transformations should be avoided, such as `SUCCEEDED` to `CANCELING`, `ABORTED` to `CANCELING`, and so on. But It seems to be a little complicated to SimpleActionServer. 

2. I'm not sure if action server only run in one thread could be better. It seems that these problems are all caused by multithreading.

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -133,9 +133,17 @@ public:
    * @return CancelResponse response of the goal cancelled
    */
   rclcpp_action::CancelResponse handle_cancel(
-    const std::shared_ptr<rclcpp_action::ServerGoalHandle<ActionT>>/*handle*/)
+    const std::shared_ptr<rclcpp_action::ServerGoalHandle<ActionT>> handle)
   {
     std::lock_guard<std::recursive_mutex> lock(update_mutex_);
+
+    if (!handle->is_active()) {
+      warn_msg(
+        "Received request for goal cancellation,"
+        "but the handle is inactive, so reject the request");
+      return rclcpp_action::CancelResponse::REJECT;
+    }
+
     debug_msg("Received request for goal cancellation");
     return rclcpp_action::CancelResponse::ACCEPT;
   }


SteveMacenski:
@KavenYau all PRs must be targeted to the `ros2` branch initially so that any changes are reflected on bleeding edge and in the future distributions. I'd be happy to merge a patch to `foxy` once that is complete.

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -97,8 +97,29 @@ class SimpleActionServer
   }
 
   rclcpp_action::CancelResponse handle_cancel(
-    const std::shared_ptr<rclcpp_action::ServerGoalHandle<ActionT>>/*handle*/)
+    const std::shared_ptr<rclcpp_action::ServerGoalHandle<ActionT>> handle)
   {
+    if (!handle->is_active()) {
```
SteveMacenski, line 102: Here's a question: What if we kept this `if (!handle->is_active()) { ... }` condition, but placed it after the `std::lock_guard<std::recursive_mutex> lock(update_mutex_);`? 

If `succeeded_current` is called first, then the `update_mutex_` would be locked, and once it was done, then `handle` would no longer be active (or valid) and then it could use this condition to `REJECT` without the need of an additional mutex / boolean?

If `cancel` is called first, then I think that might still be an issue (?). That could be solved maybe by adding a check if cancelling in `succeeded_current` to process that instead of `current_handle_->succeed(result);`?

I _believe_ that would work and involve no new variables / locks. What do you think?

KavenYau: Actually, at first, I used this way too. But I found an issue, if I was not mistaken.

> If succeeded_current is called first, then the update_mutex_ would be locked, and once it was done, then handle would no longer be active (or valid) and then it could use this condition to REJECT without the need of an additional mutex / boolean?

In this case, I think it should work. 

> If cancel is called first, then I think that might still be an issue (?). That could be solved maybe by adding a check if cancelling in succeeded_current to process that instead of current_handle_->succeed(result);?

But in this case. I think it was still an issue. Actually the goal state would be transformed to `CANCELING` only after this line executed(https://github.com/ros2/rclcpp/blob/543a3c39c16fdeb3f0109d0e906105870f547fe1/rclcpp_action/include/rclcpp_action/server.hpp#L367). 
So once `cancel_handle` returned `ACCEPT`, then checked if goal state was cancelling, it would be `false`, then `current_handle_->succeed(result);` would still be executed.

It seems that it's missing some mutexes (or other synchronization mechanisms) between goal handle methods and goal handle callbacks.  If so, it might be an rclcpp issue. What's your opinion?

KavenYau:
So what should I do next? 
Resubmit a PR target to branch `main`? Or We just discuss here, and try to find out the best way to handle the issue.

nav2_util/include/nav2_util/simple_action_server.hpp
SteveMacenski: You're definitely welcome to bring up those synchronization issues to rclcpp if you like, definitely would rather have them fixed upstream. 

I think that would work, what we'd do is in `succeeded_current`, after we lock the mutex (so that if cancel was called first, this function will be blocked until completed), **then** check if cancellation if requested to process that. If not, then do `succeeded(result)`. So it'd be an if/else to do `current_handle_->succeeded(result);` or `current_handle_->canceled(result);`.

KavenYau: yes, I had already opened an issue in rclcpp project. If I have a better solution, I will submit a PR to rclcpp.

OK, this would be a solution about the second case if we treat this as a synchronization issue of rclcpp.

I think I had made a mistake for the first case cause these cases might be a little complicated.
If `succeeded_current` is called first, then `cancel_handle ` would be blocked. Then `succeeded_current` would never be completed. Because in rclcpp, once  `cancel_handle ` was called, rclcpp 's mutex (`goal_handles_mutex_`) would be locked. 
Both `current_handle_->succeeded(result);` and `current_handle_->canceled(result);` would try to lock the mutex too. So that it became a deadlock issue.
Should we treat this as a synchronization issue of rclcpp too?

Some mentioned code here:
https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server.hpp#L355
https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server.hpp#L393
https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server_goal_handle.hpp#L194
https://github.com/ros2/rclcpp/blob/47f21dab3d2c01710c8ea04784e970eec4a92e15/rclcpp_action/include/rclcpp_action/server_goal_handle.hpp#L214

KavenYau:
I think I should test on ros2 master + navigation2 master to confirm that whether this is still an issue.

nav2_util/include/nav2_util/simple_action_server.hpp
SteveMacenski: > If succeeded_current is called first, then cancel_handle would be blocked. Then succeeded_current would never be completed.

Why? Cancel would be blocked until succeeded current was completed, that's true, but there's no reason it wouldn't be completed that I can see. Handling of the cancel request would entirely be blocked until succeeded_current is done and once its done then the cancel callback would see its not active anymore (from `!handle->is_active()` you added) and return.

KavenYau: > Cancel would be blocked until succeeded current was completed, that's true.

Yes. To navigation2, I think it's a correct logic.

> Handling of the cancel request would entirely be blocked until succeeded_current is done.

It would be blocked when doing `current_handle_->succeed(result);`, because it was trying to lock the rclcpp 's mutex (`goal_handles_mutex_`) which had been locked before `cancel_handle` called.

I think the problem might be that rclcpp has not considered about the user mutex locked in the callback.

SteveMacenski: OK, I'd explain that more concisely in the rclcpp issue: cancel/succeed are problematic with a race condition when the user **has** to lock the goal handle resource due to the multithreaded implications of the action server.

SteveMacenski: Do you see any way around this other than throwing in more mutexes in this? Is there a solution in actions in rclcpp that would resolve this issue?

KavenYau: https://github.com/KavenYau/rclcpp/commit/d1485ce6cfe46834b01585a73495b8c14bc4615e

I hope this patch could fix the deadlock issue. I'll test it in the next few days.

KavenYau:
I reproduced the issue on ros2 master + navigation2 master.

Here is the backtrace:
https://gist.github.com/KavenYau/cfbe23dbf4c406b3908c41a81058d22f

nav2_util/include/nav2_util/simple_action_server.hpp
KavenYau: I have tested by adding a unit test (https://github.com/KavenYau/rclcpp/commit/9fe631a20e6cb22a1db5812c0ae662b6a45b9d3c).

Without the this patch KavenYau/rclcpp@b8f567b, the test was failed.
After applying the patch, the test was passed.

SteveMacenski: Is that an open PR on rclcpp? I'd support that change!

KavenYau: Yes, this is the PR. https://github.com/ros2/rclcpp/pull/1635

SteveMacenski: Is our current status for this now to wait until that's put in or anything we can do here in the meantime?

KavenYau: The initial patch which I submited might be a temporary solution.

But I think the final solution should be based on rclcpp's changes. Maybe we can do some changes here first?

KavenYau:
Maybe this PR is not necessary because it seems that all issues would be resolved in rclcpp.

This PR and #2273 could be closed If this issue https://github.com/ros-planning/navigation2/pull/2304 were fixed. 😄

SteveMacenski:
OK, with that merged now, can we / should we close this PR?

KavenYau:
Sorry, I made a mistake.

> This PR and #2273 could be closed If this issue #2304 were fixed.

 Edit: This PR and #2273 could be closed If this issue https://github.com/ros2/rclcpp/issues/1599 were fixed. 

Should we just wait until https://github.com/ros2/rclcpp/issues/1599 finished?

KavenYau:
@SteveMacenski  No, https://github.com/ros2/rclcpp/pull/1635 only fix the deadlock issue, but https://github.com/ros2/rclcpp/pull/1641 could totally fix it. Actually there is already one reviewer approve it. I guess it'd not cost much time.

And the temporary solution is to add `if (!handle->is_active()) { ... } condition` to `cancel_handle` as we discussed.

SteveMacenski:
Given the superficial nature of Jacob and Ivan's review comments, I'm almost entirely sure that this will be merged if you make those minor updates so I'm not sure its worth the short term patch here :small_airplane:

KavenYau:
I have already modified the PR. I think it works in most of the time ( 99.9999% 😄 ) except for very few of context switching cases like https://github.com/ros2/rclcpp/pull/1635 mentioned.
- user callback checks if (!handle->is_active()) which fails
- user returns CancelResponse::ACCEPT
- (⚡) context switch, goal finishes (terminates? enters terminal state? moves to aborted/succeeded state? not sure on the terminology)
- switch back to original thread, call goal_handle->_cancel_goal() because CancelResponse::ACCEPT == resp
- now we're canceling a finished goal


> That's a safe non-hacky solution and would give us error messages in case this became an issue again down the line.

For this propose, I think it should be a warning message?

SteveMacenski (approved):

SteveMacenski:
```
Diff with 11 lines
--- include/nav2_util/simple_action_server.hpp
+++ include/nav2_util/simple_action_server.hpp.uncrustify
@@ -141,4 +141,4 @@
-       warn_msg(
-          "Received request for goal cancellation,"
-          "but the handle is inactive, so reject the request");
-        return rclcpp_action::CancelResponse::REJECT;
+      warn_msg(
+        "Received request for goal cancellation,"
+        "but the handle is inactive, so reject the request");
+      return rclcpp_action::CancelResponse::REJECT;

```

Just have a linting issue

KavenYau:
`uncrustify` check passed :smile: .

SteveMacenski:
Thanks! Awesome 😎

PR #2308, opened 2021-04-13
[WIP] Fix for wings artifacts: adding new filters entity to separately update costmaps

AlexeyMerzlyakov:
This fixes remaining "wings" artifacts from #2051 issue.
Wings artifacts appeared on the boundary window edges when costmap filters (e.g. `keepout_filter`) are being applied _after_ `inflation_layer`. The result of filter work at the `LayeredCostmap::updateMap()` stage is being merged into resulting costmap and on next `updateMap()` iterations treated as normal obstacles and finally being inflated inside boundary window which is causing "wings".
This can lead to such effects as shown in the picture below:
no_fix_2d

The proposed fix separates plugins and filters, applying plugins on `plugins_costmap_` and filters over the plugins, to usual `costmap_`. After fix applied, the resulting costmap looks like expected:
fixd

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2051 |
| Robotic platform tested on | Gazebo simulation of TB3 in a [AWS](https://github.com/aws-robotics/aws-robomaker-small-warehouse-world) world |

---

* Added new `filters` entity, the same as `plugins` but separated from them.
* Separeted costmaps updating mechanism for plugins and filters in order to allow filters to be applied over the plugins.
* Remain costmaps updating mechanism to be untouched when there are no filters enabled.
* Added auxilary `Costmap2D::copyWindow()` method.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp
@@ -115,6 +115,22 @@ public:
     double win_size_y);
 
   /**
+   * @brief Copies the (x0,y0)..(xn,yn) window from source costmap into a current costmap
+     @param source Source costmap where the window will be copied from
+     @param sx0 Lower x-boundary of the source window to copy, in cells
+     @param sy0 Lower y-boundary of the source window to copy, in cells
+     @param sxn Upper x-boundary of the source window to copy, in cells
+     @param syn Upper y-boundary of the source window to copy, in cells
+     @param dx0 Lower x-boundary of the destination window to copy, in cells
+     @param dx0 Lower y-boundary of the destination window to copy, in cells
+     @returns true if copy was succeeded or false in negative case
+   */
+  bool copyWindow(
+    const Costmap2D & source,
+    unsigned int sx0, unsigned int sy0, unsigned int sxn, unsigned int syn,
+    unsigned int dx0, unsigned int dy0);
+
+  /**
    * @brief  Default constructor
    */
   Costmap2D();
--- a/nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
@@ -136,11 +136,27 @@ public:
   }
 
   /**
+   * @brief Get the vector of pointers to the costmap filters
+   */
+  std::vector<std::shared_ptr<Layer>> * getFilters()
+  {
+    return &filters_;
+  }
+
+  /**
    * @brief Add a new plugin to the plugins vector to process
    */
   void addPlugin(std::shared_ptr<Layer> plugin)
   {
     plugins_.push_back(plugin);
+  }
+
+  /**
+   * @brief Add a new costmap filter plugin to the filters vector to process
+   */
+  void addFilter(std::shared_ptr<Layer> filter)
+  {
+    filters_.push_back(filter);
   }
 
   /**
@@ -197,7 +213,7 @@ public:
   bool isOutofBounds(double robot_x, double robot_y);
 
 private:
-  Costmap2D costmap_;
+  Costmap2D plugins_costmap_, costmap_;
   std::string global_frame_;
 
   bool rolling_window_;  /// < @brief Whether or not the costmap should roll with the robot
@@ -207,6 +223,7 @@ private:
   unsigned int bx0_, bxn_, by0_, byn_;
 
   std::vector<std::shared_ptr<Layer>> plugins_;
+  std::vector<std::shared_ptr<Layer>> filters_;
 
   bool initialized_;
   bool size_locked_;
--- a/nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
@@ -108,7 +108,7 @@ public:
    */
   Costmap2D * getCostmap()
   {
-    return &costmap_;
+    return &combined_costmap_;
   }
 
   /**
@@ -124,7 +124,7 @@ public:
    */
   bool isTrackingUnknown()
   {
-    return costmap_.getDefaultValue() == nav2_costmap_2d::NO_INFORMATION;
+    return combined_costmap_.getDefaultValue() == nav2_costmap_2d::NO_INFORMATION;
   }
 
   /**
@@ -213,7 +213,12 @@ public:
   bool isOutofBounds(double robot_x, double robot_y);
 
 private:
-  Costmap2D plugins_costmap_, costmap_;
+  // primary_costmap_ is a bottom costmap used by plugins when costmap filters were enabled.
+  // combined_costmap_ is a final costmap where all results produced by plugins and filters (if any)
+  // to be merged.
+  // The separation is aimed to avoid interferences of work between plugins and filters.
+  // primay_costmap_ and combined_costmap_ have the same sizes, origins and default values.
+  Costmap2D primary_costmap_, combined_costmap_;
   std::string global_frame_;
 
   bool rolling_window_;  /// < @brief Whether or not the costmap should roll with the robot
--- a/nav2_costmap_2d/src/clear_costmap_service.cpp
+++ b/nav2_costmap_2d/src/clear_costmap_service.cpp
@@ -112,6 +112,15 @@ void ClearCostmapService::clearRegion(const double reset_distance, bool invert)
       clearLayerRegion(costmap_layer, x, y, reset_distance, invert);
     }
   }
+
+  layers = costmap_.getLayeredCostmap()->getFilters();
+
+  for (auto & layer : *layers) {
+    if (layer->isClearable()) {
+      auto costmap_layer = std::static_pointer_cast<CostmapLayer>(layer);
+      clearLayerRegion(costmap_layer, x, y, reset_distance, invert);
+    }
+  }
 }
 
 void ClearCostmapService::clearLayerRegion(
--- a/nav2_costmap_2d/src/clear_costmap_service.cpp
+++ b/nav2_costmap_2d/src/clear_costmap_service.cpp
@@ -113,14 +113,8 @@ void ClearCostmapService::clearRegion(const double reset_distance, bool invert)
     }
   }
 
-  layers = costmap_.getLayeredCostmap()->getFilters();
-
-  for (auto & layer : *layers) {
-    if (layer->isClearable()) {
-      auto costmap_layer = std::static_pointer_cast<CostmapLayer>(layer);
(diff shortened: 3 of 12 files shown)


AlexeyMerzlyakov:
This is first version of proposed fix. After the main code will be agreed, testcases and documentation updates will be added.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp
```diff
@@ -114,6 +114,22 @@ class Costmap2D
     double win_size_x,
```
SteveMacenski, line 114: We also need to add this to a migration guide and update the tutorials for the keepout / speed zones (and parameter guides)

nav2_costmap_2d/src/clear_costmap_service.cpp
```diff
@@ -112,6 +112,15 @@ void ClearCostmapService::clearRegion(const double reset_distance, bool invert)
       clearLayerRegion(costmap_layer, x, y, reset_distance, invert);
     }
   }
+
+  layers = costmap_.getLayeredCostmap()->getFilters();
+
+  for (auto & layer : *layers) {
+    if (layer->isClearable()) {
```
SteveMacenski, line 119: https://github.com/ros-planning/navigation2/blob/2115e26849105fde9322879bb4a25790234f95d1/nav2_costmap_2d/include/nav2_costmap_2d/costmap_filters/costmap_filter.hpp#L125 this is always false for all costmap filters

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -157,6 +158,20 @@ Costmap2DROS::on_configure(const rclcpp_lifecycle::State & /*state*/)
...
     RCLCPP_INFO(get_logger(), "Initialized plugin \"%s\"", plugin_names_[i].c_str());
   }
+  // and costmap filters as well
+  for (unsigned int i = 0; i < filter_names_.size(); ++i) {
+    RCLCPP_INFO(get_logger(), "Using costmap filter \"%s\"", filter_names_[i].c_str());
+
+    std::shared_ptr<Layer> filter = plugin_loader_.createSharedInstance(filter_types_[i]);
+    layered_costmap_->addFilter(filter);
+
+    // TODO(mjeronimo): instead of get(), use a shared ptr
```
SteveMacenski, line 168: remove this TODO

nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -151,6 +163,24 @@ void LayeredCostmap::updateMap(double robot_x, double robot_y, double robot_yaw)
...
+    double prev_minx = minx_;
+    double prev_miny = miny_;
+    double prev_maxx = maxx_;
+    double prev_maxy = maxy_;
+    (*filter)->updateBounds(robot_x, robot_y, robot_yaw, &minx_, &miny_, &maxx_, &maxy_);
+    if (minx_ > prev_minx || miny_ > prev_miny || maxx_ < prev_maxx || maxy_ < prev_maxy) {
+      RCLCPP_WARN(
+        rclcpp::get_logger(
+          "nav2_costmap_2d"), "Illegal bounds change, was [tl: (%f, %f), br: (%f, %f)], but "
+        "is now [tl: (%f, %f), br: (%f, %f)]. The offending layer is %s",
```
SteveMacenski, line 178:
```suggestion
        "is now [tl: (%f, %f), br: (%f, %f)]. The offending filter is %s",
```

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
```diff
@@ -135,6 +135,14 @@ class LayeredCostmap
     return &plugins_;
   }
 
+  /**
+   * @brief Get the vector of pointers to the costmap filters
+   */
+  std::vector<std::shared_ptr<Layer>> * getFilters()
```
SteveMacenski, line 141: There is nothing wrong with this, since this is how the plugins are as well, but its a little odd to me that this is a pointer and not just a reference. It would make the semantics later in the code much easier if it was returned by reference so that it was just `filter->do_thing()` vs `(*filter)->do_thing()`.

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
```diff
@@ -197,7 +213,7 @@ class LayeredCostmap
   bool isOutofBounds(double robot_x, double robot_y);
 
 private:
-  Costmap2D costmap_;
+  Costmap2D plugins_costmap_, costmap_;
```
SteveMacenski, line 216: This is not a good name - we should figure out another name for these costmaps to separate them. Maybe `master_costmap_` for the main one? I'm not sure I have a better name suggestion for the plugins costmap, though it could potentially use one to be more clear what it is on a casual read-through

SteveMacenski (changes requested):

SteveMacenski:
This looks good to me in concept - I left a comment on the ticket, the additional CPU times you listed and percentage of increase on processing time doesn't seem quite right

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
AlexeyMerzlyakov: OK. In my opinion that is also suitable to go structure. The `plugins_` or `filters_` is a vector of shared pointers, so its iterator will have a double-pointer dimension (because of a `std::iterator` is a pointer to a vector' item which is a shared_ptr itself). Changing the `getFilters()` return value to a reference to std::vector will simplify the iterator boundaries in for-cycle (e.g. `filters->begin()` will be simplified to `filters.begin()`), but leave the `(*filter)->do_thing` odd construction to be the same. So I see no point in replacing pointer to a reference return value.

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
AlexeyMerzlyakov: Not finally sure about `master_costmap_`, as both costmaps by their meaning are master costmap. One of them is the basis/basement/bottom layer for the second final costmap. So, how about to rename `plugins_costmap_` -> to `basic_costmap_`, `preliminary_costmap_` or `bottom_costmap_`?

nav2_costmap_2d/src/clear_costmap_service.cpp
AlexeyMerzlyakov: Changed in https://github.com/ros-planning/navigation2/pull/2308/commits/a9f92d8711e0bd50fe1bd3b54f2c9304c3f87b01#diff-f2ea1827ae80319b8a5efca999c9f81e1a2cf95578e516ea48aa4e606c32cd1eL116-R117

nav2_costmap_2d/src/costmap_2d_ros.cpp
AlexeyMerzlyakov: Fixed in https://github.com/ros-planning/navigation2/pull/2308/commits/a9f92d8711e0bd50fe1bd3b54f2c9304c3f87b01#diff-d3369b82718255857b744f9289b4e31a34924f72fb40212604b4aa8276b857f2L168

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
SteveMacenski: `combined_costmap_` for the combined filter/plugins and `plugin_costmap_` for just the plugins?

Add a comment in the header file about why there are two and what each are for. That should help stem some of the confusion later.

AlexeyMerzlyakov: In this case, I'd still consider to remain `bottom/preliminary/basic_costmap_` and `costmap_` (or `combined_costmap_`) variables. Having `plugin_costmap_` and `combined_costmap_` indeed can cause some confusion: if there is no filters enabled we will use `combined_costmap_` directly only for the plugins, but if one plugin was enabled, we will write all plugins' work into `plugin_costmap_`. In other words, for the plugins `plugin_costmap_` won't always be used.
So, now I see most reasonable to have to variables:  `bottom/preliminary/basic_costmap_` for bottom layer and `costmap_` or  `combined_costmap_` for final result. What do you think about it?

nav2_costmap_2d/src/costmap_2d.cpp
```diff
@@ -184,6 +184,29 @@ bool Costmap2D::copyCostmapWindow(
...
 
+bool Costmap2D::copyWindow(
+  const Costmap2D & source,
+  unsigned int sx0, unsigned int sy0, unsigned int sxn, unsigned int syn,
+  unsigned int dx0, unsigned int dy0)
+{
+  const unsigned int sz_x = sxn - sx0;
+  const unsigned int sz_y = syn - sy0;
+
+  if (sxn > source.getSizeInCellsX() || syn > source.getSizeInCellsY()) {
```
SteveMacenski, line 195: Was this a condition you ran into?

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
SteveMacenski: How about `primary_costmap_` for the plugins and  `combined_costmap_` for combined / final?

nav2_costmap_2d/src/costmap_2d.cpp
AlexeyMerzlyakov: I've met it in tests. However, there is no one is safe from incorrect input data. This prevents the program to go out of costmap sizes which could lead to a segfault crashes. Therefore I've added the verification of source boundaries the same as for destination boundaries as well.

SteveMacenski: That's fine, just making sure, if you ran into this in actual situations, then it would make me question some other part of this PR that wasn't quite right. If it's just defensive, that's OK

SteveMacenski:
I think the last thing to change here is just those variable names (and then the documentation updates), no?

nav2_costmap_2d/include/nav2_costmap_2d/layered_costmap.hpp
AlexeyMerzlyakov: Agree with this naming. Changes were applied in https://github.com/ros-planning/navigation2/pull/2308/commits/9c1c7d319f8940e1953990e7606cbf66463651d6.

AlexeyMerzlyakov:
Yep, that are remained items before to-go. Renamed the costmap variables and updated the docs web-pages in a PR referenced above.

SteveMacenski (approved):
Approved, just need a successful CI run, retriggering

PR #2309, opened 2021-04-13
[nav2_lifecycle_manager] Support namespaced LifecycleManagers

doisyg:
---

## Basic Info

| ------ | ----------- |

---

When using a **LifecycleManagerClient** it is currently impossible to connect to a **LifecycleManager** that is on a namespace due to the use of the **LifecycleManager** node address for the **LifecycleManagerClient** service client node naming. For instance `my_namespace/my_lifecycle_manager` will result in the incorrect creation of an **LifecycleManagerClient** service client node with the name `my_namespace/my_lifecycle_manager_service_client` (a node name cannot have '/' characters).
This PR replaces `/` characters with `_`

This could have been solved also by putting the **LifecycleManagerClient** service client name under the namespace of the **LifecycleManager** it connects to by using the Node constructor namespace argument. But I am not so sure about changing the namespace of the **LifecycleManagerClient** service client node without the user having control on it.

---

#### For Maintainers: 

### DIFF
--- a/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
+++ b/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
@@ -45,7 +45,7 @@ public:
   /**
    * @brief A constructor for LifeCycleMangerClient
    */
-  explicit LifecycleManagerClient(const std::string & name);
+  explicit LifecycleManagerClient(const std::string & name, std::string namespace_ = "");
 
   // Client-side interface to the Nav2 lifecycle manager
   /**
--- a/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
+++ b/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
@@ -45,7 +45,7 @@ public:
   /**
    * @brief A constructor for LifeCycleMangerClient
    */
-  explicit LifecycleManagerClient(const std::string & name, std::string namespace_ = "");
+  explicit LifecycleManagerClient(const std::string & name, std::string & ns = "");
 
   // Client-side interface to the Nav2 lifecycle manager
   /**
--- a/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
+++ b/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
@@ -45,7 +45,7 @@ public:
   /**
    * @brief A constructor for LifeCycleMangerClient
    */
-  explicit LifecycleManagerClient(const std::string & name, std::string & ns = "");
+  explicit LifecycleManagerClient(const std::string & name, const std::string & ns = "");
 
   // Client-side interface to the Nav2 lifecycle manager
   /**
--- a/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
+++ b/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
@@ -16,7 +16,6 @@
 
 #include <cmath>
 #include <memory>
-#include <algorithm>
 #include <string>
 #include <utility>
 
@@ -27,15 +26,19 @@ namespace nav2_lifecycle_manager
 {
 using nav2_util::geometry_utils::orientationAroundZAxis;
 
-LifecycleManagerClient::LifecycleManagerClient(const std::string & name)
+LifecycleManagerClient::LifecycleManagerClient(
+  const std::string & name,
+  std::string namespace_)
 {
-  manage_service_name_ = name + std::string("/manage_nodes");
-  active_service_name_ = name + std::string("/is_active");
+  if (namespace_ != "") {
+    namespace_ += "/";
+  }
+
+  manage_service_name_ = namespace_ + name + std::string("/manage_nodes");
+  active_service_name_ = namespace_ + name + std::string("/is_active");
 
   // Create the node to use for all of the service clients
-  std::string node_name = name + "_service_client";
-  std::replace(node_name.begin(), node_name.end(), '/', '_');
-  node_ = std::make_shared<rclcpp::Node>(node_name);
+  node_ = std::make_shared<rclcpp::Node>(name + "_service_client");
 
   // Create the service clients
   manager_client_ = node_->create_client<ManageLifecycleNodes>(manage_service_name_);
--- a/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
+++ b/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
@@ -16,6 +16,7 @@
 
 #include <cmath>
 #include <memory>
+#include <algorithm>
 #include <string>
 #include <utility>
 
@@ -32,7 +33,9 @@ LifecycleManagerClient::LifecycleManagerClient(const std::string & name)
   active_service_name_ = name + std::string("/is_active");
 
   // Create the node to use for all of the service clients
-  node_ = std::make_shared<rclcpp::Node>(name + "_service_client");
+  std::string node_name = name + "_service_client";
+  std::replace(node_name.begin(), node_name.end(), '/', '_');
+  node_ = std::make_shared<rclcpp::Node>(node_name);
 
   // Create the service clients
   manager_client_ = node_->create_client<ManageLifecycleNodes>(manage_service_name_);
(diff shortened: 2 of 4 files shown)


nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -32,7 +33,9 @@ LifecycleManagerClient::LifecycleManagerClient(const std::string & name)
   active_service_name_ = name + std::string("/is_active");
 
   // Create the node to use for all of the service clients
-  node_ = std::make_shared<rclcpp::Node>(name + "_service_client");
+  std::string node_name = name + "_service_client";
+  std::replace(node_name.begin(), node_name.end(), '/', '_');
```
SteveMacenski, line 37: Where is there a `/` in the node name used? I can't find any examples and leading with an `_` would make that private context potentially in the ROS2 naming style

doisyg: Nowhere in nav2, but I am using `nav2_lifecycle_manager::LifecycleManagerClient`s to start/pause/resume `nav2_util::LifecycleNodes`s which are under a namespace.
Like here:
https://github.com/ros-planning/navigation2/blob/d537cb186dd61a9491c0b1c7865f1e4e6891ca45/nav2_system_tests/src/updown/test_updown.cpp#L25-L31
but with the difference that `lifecycle_manager_navigation` and `lifecycle_manager_localization` would be found at `my_ns/lifecycle_manager_navigation` and `my_ns/lifecycle_manager_localization`.
And if currently, doing this: 
```
 int main(int argc, char ** argv) 
 { 
   rclcpp::init(argc, argv); 
   RCLCPP_INFO(rclcpp::get_logger("test_updown"), "Initializing test"); 
   nav2_lifecycle_manager::LifecycleManagerClient client_nav("my_ns/lifecycle_manager_navigation"); 
   nav2_lifecycle_manager::LifecycleManagerClient client_loc("my_ns/lifecycle_manager_localization");
```
would throw an error:

>  terminate called after throwing an instance of 'rclcpp::exceptions::InvalidNodeNameError'
>   what():  Invalid node name: node name must not contain characters other than alphanumerics or '_':
>   'my_ns/lifecycle_manager_navigation_service_client'

SteveMacenski: So you want it to change to `my_ns_lifecycle_manager_navigation`? Why not just update the input string you're hardcoding into your main function? This doesn't seem like it would be good to mask from users if they  were genuinely expecting a namespaced output. 

https://docs.ros2.org/dashing/api/rclcpp/classrclcpp_1_1Node.html you can add a namespace to a field -- we could just have the lifecycle manager take in a potential namespace as an argument in addition to the name if you like (and just default to empty string)

doisyg: > So you want it to change to `my_ns_lifecycle_manager_navigation`? Why not just update the input string you're hardcoding into your main function? This doesn't seem like it would be good to mask from users if they were genuinely expecting a namespaced output.

Cause the constructor needs the full address of the lifecycle manager in order to initialize `manager_client_` and `is_active_client_` : 
https://github.com/ros-planning/navigation2/blob/d537cb186dd61a9491c0b1c7865f1e4e6891ca45/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp#L29-L40

> https://docs.ros2.org/dashing/api/rclcpp/classrclcpp_1_1Node.html you can add a namespace to a field -- we could just have the lifecycle manager take in a potential namespace as an argument in addition to the name if you like (and just default to empty string)

That would indeed be a proper solution if you agree on changing the constructor. So it would become: `LifecycleManagerClient(const std::string & name, const std::string & namespace = "")` ?

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -26,10 +26,16 @@ namespace nav2_lifecycle_manager
...
 
-LifecycleManagerClient::LifecycleManagerClient(const std::string & name)
+LifecycleManagerClient::LifecycleManagerClient(
+  const std::string & name,
+  std::string namespace_)
 {
-  manage_service_name_ = name + std::string("/manage_nodes");
-  active_service_name_ = name + std::string("/is_active");
+  if (namespace_ != "") {
+    namespace_ += "/";
```
SteveMacenski, line 34: The rclcpp node below has a constructor to take in the namespace, I'm not sure why you're adding it here. you could just do `node_ = std::make_shared<rclcpp::Node>(name + "_service_client", namespace);`

doisyg: Sure, I could, though I am not sure what's the best solution.
If we use the namespace to create the _service_client node, it will be under the same namespace as the manager it is controlling, and a different than the instance that created it.

For example, if in a process `process_containing_client_nav` started without a namespace, we have a manager client `client_nav` controlling a manager `lifecycle_manager_navigation` under the `lm_ns` namespace:
```
  nav2_lifecycle_manager::LifecycleManagerClient client_nav("lifecycle_manager_navigation", "lm_ns");
```
The _service_client node will be created under the manager namespace:

> $ ros2 node list
> /process_containing_client_nav
> /lm_ns/lifecycle_manager_localization
> /lm_ns/lifecycle_manager_localization_bond_client
> /lm_ns/lifecycle_manager_localization_service_client

Whereas if we don't use the manager namespace to create the _service_client node, it will be in the same namespace as the process that started it:

> $ ros2 node list
> /process_containing_client_nav
> /lifecycle_manager_localization_service_client
> /lm_ns/lifecycle_manager_localization
> /lm_ns/lifecycle_manager_localization_bond_client

I honestly don't know what's best

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
SteveMacenski: yeah, but if you're hiding `my_ms/topic` to `my_ns_topic`, then you **wouldn't** know the full address if you didn't know this masking was happening in the backend. Yes, we can change the interface.

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
SteveMacenski: I'd argue the first is more correct, because you need to namespace the `lifecycle_manager_localization` in order to run multiple instances of it, which I assume is the point of this PR.

doisyg: Ok, convinced

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
```diff
@@ -45,7 +45,7 @@ class LifecycleManagerClient
   /**
    * @brief A constructor for LifeCycleMangerClient
    */
-  explicit LifecycleManagerClient(const std::string & name);
+  explicit LifecycleManagerClient(const std::string & name, std::string namespace_ = "");
```
SteveMacenski, line 48: Both files need `namespace` not `namespace_` but then this is good to merge

SteveMacenski (changes requested):

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
doisyg: `namespace` is reserved in cpp I believe

SteveMacenski: Then use `ns` or something, the `_` is for class members only

SteveMacenski (approved):

SteveMacenski:
Much simpler :smile: just waiting on CI

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager_client.hpp
SteveMacenski: Oh, yeah please if you you're willing

SteveMacenski (approved):

SteveMacenski:
@doisyg you have a linting issue (maybe other tests too, but definitely linting)
```
--- src/lifecycle_node.cpp
+++ src/lifecycle_node.cpp.uncrustify
@@ -43 +43,2 @@
-LifecycleNode::LifecycleNode(const std::string & node_name,
+LifecycleNode::LifecycleNode(
+  const std::string & node_name,

```

PR #2314, opened 2021-04-16
Update ompl in underlay

ruffsl:
And revert old workaround in CI now that ompl/ompl#753 is fixed.

### DIFF
--- a/tools/underlay.repos
+++ b/tools/underlay.repos
@@ -22,4 +22,4 @@ repositories:
   ompl/ompl:
     type: git
     url: https://github.com/ompl/ompl.git
-    version: 1.5.0
+    version: main
--- a/tools/underlay.repos
+++ b/tools/underlay.repos
@@ -22,4 +22,4 @@ repositories:
   ompl/ompl:
     type: git
     url: https://github.com/ompl/ompl.git
-    version: main
+    version: 1.5.2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,22 +60,9 @@ _commands:
               sha256sum $PWD/checksum.txt >> checksum.txt
               apt-get update
               rosdep update
-              
-              # workarround for OMPL and rosdep
-              # https://github.com/ompl/ompl/issues/753
-              # Prevent searching $ROS_WS/install given it's too big for rosdep
-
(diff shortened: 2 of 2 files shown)


ruffsl:
```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
nav2_smac_planner: Cannot locate rosdep definition for [ompl]
```

> https://app.circleci.com/pipelines/github/ros-planning/navigation2/5200/workflows/50a39b60-3eaf-4c98-8e84-a3d1ec63f578/jobs/19164/parallel-runs/0/steps/0-114

Hmm, guess ompl still has some ament index issues.

tools/underlay.repos
```diff
@@ -22,4 +22,4 @@ repositories:
   ompl/ompl:
     type: git
     url: https://github.com/ompl/ompl.git
-    version: 1.5.0
+    version: main
```
SteveMacenski, line 25: Why not use a specific release? I'd really like to not rebuild ompl breaking the cache every commit to its master branch. The ompl builds take 30-40 minutes and make our CI not that fast when we have to build it

ruffsl: Ok, v1.5.2 still includes the recent ament changes

```suggestion
    version: 1.5.2
```

ruffsl: Should note that normally, only the dockerhub image build would be building underlay, not sure how frequently other PRs have been changing the `underlay.repos` file.

SteveMacenski: Good point, I’m ok with this. I see it occasionally but its not very often.

SteveMacenski:
@ruffsl I'm OK merging this once CI passes, it looks like the overlay workspace can't find OMPL for some reason now in main?

SteveMacenski:
@ruffsl happy to merge this, I retriggered CI but it wasn't building before. If this is important, I can help, just let me know

ruffsl:
Think we need to find what ompl is still missing in registering with the ament index, such that rosdep still can't find it in the sourced environment, without resorting to manually pointing to the underly source folder. I thought I fixed it before, but I guess not, but I'm bit too tied up atm to reinvestigate.

SteveMacenski:
OK - I'd prefer to close this until there's some more cycles

Issue #2280, opened 2021-03-30
Look into ramifications of increasing server_timeout in BT 


SteveMacenski:
CC https://github.com/ros2/ros2/issues/1074

See if we can increase this using async BT nodes without slowing the tick rate, or if we need to lower the tick rate proportionally to the time we can allot to this. 

18ms is the high number we've seen

SteveMacenski:
@KavenYau this is somewhat related to your other ticket. If you wanted to look into this, increasing this timeout will fix the `send_goal` failure mode. We just don't yet understand the ramifications of increasing this. The BT ticks at 100hz (so the 10ms number we have set) so if this takes longer, we need to profile that this is OK, make it OK, or slow down the BT ticking rate so that it is OK. 

I'd be more than happy to seed this to you if you had immediate need to fix it. It might take me a couple of weeks to get to this.

SteveMacenski:
This timeout should also be a parameter, right now its hardcoded at 10ms. This can be done if we also do the RUNNING state work as discussed in the working group meeting @naiveHobo if that is going to be part of your fix :smile:

SteveMacenski:
@naiveHobo did you make any progress on this?

KavenYau:
I increased `server_timeout` to 3000ms for temporary fix the issue. https://github.com/KavenYau/navigation2/commit/565b22af80641432dad1678e428aa93e214bbec.

With using some low performance mechine or high CPU load cases, I guess the number would be higher than 18ms, it's stochastic. This is probably a communication performance issue.

In my opinions, although the timeout is a parameter, it's difficult for me to determine a perfect timeout to set, maybe it's not crucial. I mean slowing down the BT ticking rate could only reduce the number of timeout occurring, 

How about trying to reduce the timeout impact on BT running? or adding threshold of timeout times to BT running?

KavenYau:
I increased `server_timeout` to 3000ms for temporarily fixing the issue. https://github.com/KavenYau/navigation2/commit/87e43c45a5b3fe730ca72c7cd3391c7d2ed596b7

With using some low performance mechine or high CPU load cases, I guess the number would be higher than 18ms, it's stochastic. This is probably a communication performance issue.

In my opinions, although the timeout is a parameter, it's difficult for me to determine a perfect timeout to set, maybe it's not crucial. I mean slowing down the BT ticking rate could only reduce the number of timeout occurring, 

[...]

PR #2305, opened 2021-04-08
Check if goal_status_ is valid.

tzskp1:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | #2303 |
| Primary OS tested on | Ubuntu |
| Robotic platform tested on | gazebo simulation |

---

## Description of contribution in a few bullet points

* I added checking snippet of `goal_handle_` for avoiding race conditions.

---

#### For Maintainers: 
- [x] Check that any new parameters added are updated in navigation.ros.org
- [x] Check that any significant change is added to the migration guide
- [x] Check that any new functions have Doxygen added
- [x] Check that any new features have test coverage
- [x] Check that any new plugins is added to the plugins page
- [x] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

PR #2320, opened 2021-04-22
Use incremental timeouts in spin_until_future_calls during BT execution to handle large server timeouts

naiveHobo:

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2280 |
| Robotic platform tested on | TurtleBot3 Gazebo simulation |

---

* `BtServiceNode` and `BtActionNode` use incremental timeout values in all `rclcpp::spin_until_future_complete` calls while waiting for the server to response to a request when the specified server timeout exceeds the BT execution loop rate. This ensures that `rclcpp::spin_until_future_complete` don't slow down BT execution if the server timeout has to be increased for whatever reason.
* Expose `bt_loop_timeout` and `default_server_timeout` as node parameters.

* Add `bt_loop_timeout` and `default_server_timeout` to navigation.ros.org

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -181,7 +181,7 @@ public:
 
     // if new goal was sent and action server has not yet responded
     // check the future goal handle
-    if (!goal_handle_) {
+    if (future_goal_handle_) {
       auto elapsed = (node_->now() - time_goal_sent_).to_chrono<std::chrono::milliseconds>();
       if (!is_future_goal_handle_complete(elapsed)) {
         // return RUNNING if there is still some time before timeout happens
@@ -193,7 +193,7 @@ public:
           node_->get_logger(),
           "Timed out while waiting for action server to acknowledge goal request for %s",
           action_name_.c_str());
-        goal_handle_.reset();
+        future_goal_handle_.reset();
         return BT::NodeStatus::FAILURE;
       }
     }
@@ -218,7 +218,7 @@ public:
             node_->get_logger(),
             "Timed out while waiting for action server to acknowledge goal request for %s",
             action_name_.c_str());
-          goal_handle_.reset();
+          future_goal_handle_.reset();
           return BT::NodeStatus::FAILURE;
         }
       }
@@ -318,7 +318,9 @@ protected:
         }
       };
 
-    future_goal_handle_ = action_client_->async_send_goal(goal_, send_goal_options);
+    future_goal_handle_ = std::make_shared<
+      std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>>(
+      action_client_->async_send_goal(goal_, send_goal_options));
     time_goal_sent_ = node_->now();
   }
 
@@ -334,21 +336,22 @@ protected:
 
     // server has already timed out, no need to sleep
     if (remaining <= std::chrono::milliseconds(0)) {
-      goal_handle_.reset();
+      future_goal_handle_.reset();
       return false;
     }
 
     auto timeout = remaining > bt_loop_duration_ ? bt_loop_duration_ : remaining;
-    auto result = rclcpp::spin_until_future_complete(node_, future_goal_handle_, timeout);
+    auto result = rclcpp::spin_until_future_complete(node_, *future_goal_handle_, timeout);
     elapsed += timeout;
 
     if (result == rclcpp::FutureReturnCode::INTERRUPTED) {
-      goal_handle_.reset();
+      future_goal_handle_.reset();
       throw std::runtime_error("send_goal failed");
     }
 
     if (result == rclcpp::FutureReturnCode::SUCCESS) {
-      goal_handle_ = future_goal_handle_.get();
+      goal_handle_ = future_goal_handle_->get();
+      future_goal_handle_.reset();
       if (!goal_handle_) {
         throw std::runtime_error("Goal was rejected by the action server");
       }
@@ -390,7 +393,7 @@ protected:
   std::chrono::milliseconds bt_loop_duration_;
 
   // To track the action server acknowledgement when a new goal is sent
-  std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>
+  std::shared_ptr<std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>>
   future_goal_handle_;
   rclcpp::Time time_goal_sent_;
 };
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -49,9 +49,14 @@ public:
     node_ = config().blackboard->template get<rclcpp::Node::SharedPtr>("node");
 
     // Get the required items from the blackboard
+    bt_loop_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_loop_timeout");
     server_timeout_ =
-      config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
+      config().blackboard->template get<std::chrono::milliseconds>("default_server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+
+    // Set spin_until_future_complete timeout
+    timeout_ = server_timeout_ < bt_loop_timeout_ ? server_timeout_ : bt_loop_timeout_;
 
     // Initialize the input and output messages
     goal_ = typename ActionT::Goal();
@@ -174,7 +179,24 @@ public:
       // user defined callback
       on_tick();
 
-      on_new_goal_received();
+      send_new_goal();
+    }
+
+    // if new goal was sent and action server has not yet responded
+    // check the future goal handle
+    if (goal_sent_) {
+      if (!check_future_goal_handle()) {
+        // return RUNNING if there is still some time before timeout happens
+        auto elapsed = (node_->now() - sent_time_).to_chrono<std::chrono::milliseconds>();
+        if (elapsed < server_timeout_) {
+          return BT::NodeStatus::RUNNING;
+        }
+        // if server has taken more time to respond than the specified timeout value return FAILURE
+        RCLCPP_WARN(
+          node_->get_logger(),
+          "Timed out while waiting for action server to respond for %s", action_name_.c_str());
+        return BT::NodeStatus::FAILURE;
+      }
     }
 
     // The following code corresponds to the "RUNNING" loop
@@ -187,7 +209,17 @@ public:
         goal_status == action_msgs::msg::GoalStatus::STATUS_ACCEPTED))
       {
         goal_updated_ = false;
-        on_new_goal_received();
+        send_new_goal();
+        if (!check_future_goal_handle()) {
+          auto elapsed = (node_->now() - sent_time_).to_chrono<std::chrono::milliseconds>();
+          if (elapsed < server_timeout_) {
+            return BT::NodeStatus::RUNNING;
+          }
+          RCLCPP_WARN(
+            node_->get_logger(),
+            "Timed out while waiting for action server to respond for %s", action_name_.c_str());
+          return BT::NodeStatus::FAILURE;
+        }
       }
 
       rclcpp::spin_some(node_);
@@ -257,7 +289,7 @@ protected:
   /**
    * @brief Function to send new goal to action server
    */
-  void on_new_goal_received()
+  void send_new_goal()
   {
     goal_result_available_ = false;
     auto send_goal_options = typename rclcpp_action::Client<ActionT>::SendGoalOptions();
@@ -272,18 +304,42 @@ protected:
         }
       };
 
-    auto future_goal_handle = action_client_->async_send_goal(goal_, send_goal_options);
+    future_goal_handle_ = action_client_->async_send_goal(goal_, send_goal_options);
+    sent_time_ = node_->now();
+    goal_sent_ = true;
+  }
 
-    if (rclcpp::spin_until_future_complete(node_, future_goal_handle, server_timeout_) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+  /**
+   * @brief Function to check response from the action server when a new goal is sent
+   * @return boolen True if future_goal_handle_ returns SUCCESS, False when it returns TIMEOUT
+   */
+  bool check_future_goal_handle()
+  {
+    auto elapsed = (node_->now() - sent_time_).to_chrono<std::chrono::milliseconds>();
+    auto remaining = server_timeout_ - elapsed;
+
+    // server has already timed out, no need to sleep
+    if (remaining <= std::chrono::milliseconds(0)) {
+      return false;
+    }
+
+    auto timeout = remaining > timeout_ ? timeout_ : remaining;
+
+    auto result = rclcpp::spin_until_future_complete(node_, future_goal_handle_, timeout);
+    if (result == rclcpp::FutureReturnCode::INTERRUPTED) {
       throw std::runtime_error("send_goal failed");
     }
 
-    goal_handle_ = future_goal_handle.get();
-    if (!goal_handle_) {
-      throw std::runtime_error("Goal was rejected by the action server");
+    if (result == rclcpp::FutureReturnCode::SUCCESS) {
+      goal_handle_ = future_goal_handle_.get();
+      if (!goal_handle_) {
+        throw std::runtime_error("Goal was rejected by the action server");
+      }
+      goal_sent_ = false;
+      return true;
     }
+
+    return false;
   }
 
   /**
@@ -313,6 +369,18 @@ protected:
   // The timeout value while waiting for response from a server when a
   // new action goal is sent or canceled
   std::chrono::milliseconds server_timeout_;
+
+  // The timeout value for BT loop execution
+  std::chrono::milliseconds bt_loop_timeout_;
+
+  // spin_until_future_complete timeout value
+  std::chrono::milliseconds timeout_;
+
+  // To track the action server response when a new goal is sent
+  std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>
+  future_goal_handle_;
+  bool goal_sent_{false};
+  rclcpp::Time sent_time_;
 };
 
 }  // namespace nav2_behavior_tree
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -182,9 +182,10 @@ public:
     // if new goal was sent and action server has not yet responded
     // check the future goal handle
     if (!goal_handle_) {
-      if (!is_future_goal_handle_complete()) {
+      auto elapsed = (node_->now() - time_goal_sent_).to_chrono<std::chrono::milliseconds>();
+      if (!is_future_goal_handle_complete(elapsed)) {
         // return RUNNING if there is still some time before timeout happens
-        if (time_elapsed_since_goal_sent_ < server_timeout_) {
+        if (elapsed < server_timeout_) {
           return BT::NodeStatus::RUNNING;
         }
         // if server has taken more time to respond than the specified timeout value return FAILURE
@@ -208,8 +209,9 @@ public:
       {
         goal_updated_ = false;
         send_new_goal();
-        if (!is_future_goal_handle_complete()) {
-          if (time_elapsed_since_goal_sent_ < server_timeout_) {
+        auto elapsed = (node_->now() - time_goal_sent_).to_chrono<std::chrono::milliseconds>();
+        if (!is_future_goal_handle_complete(elapsed)) {
+          if (elapsed < server_timeout_) {
             return BT::NodeStatus::RUNNING;
           }
           RCLCPP_WARN(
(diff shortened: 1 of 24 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -257,7 +289,7 @@ class BtActionNode : public BT::ActionNodeBase
   /**
    * @brief Function to send new goal to action server
    */
-  void on_new_goal_received()
```
SteveMacenski, line 260: Please rename this function back to `on_new_goal_received`, to be in the styling of the rest of the file

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -174,7 +179,24 @@ class BtActionNode : public BT::ActionNodeBase
...
       on_tick();
 
-      on_new_goal_received();
+      send_new_goal();
+    }
+
+    // if new goal was sent and action server has not yet responded
+    // check the future goal handle
+    if (goal_sent_) {
+      if (!check_future_goal_handle()) {
```
SteveMacenski, line 188: This logic will bypass on_wait_for_result which in turn will stop checks for path preemption in the FollowPath BT node. Is this block necessary if just below you have the same thing in L213-223?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -49,10 +49,15 @@ class BtActionNode : public BT::ActionNodeBase
...
 
     // Get the required items from the blackboard
+    bt_loop_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_loop_timeout");
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
 
+    // Set spin_until_future_complete timeout
+    timeout_ = server_timeout_ < bt_loop_timeout_ ? server_timeout_ : bt_loop_timeout_;
```
SteveMacenski, line 59: `timeout_` is never really used. `server_timeout_` is in most places. I'm kind of wondering why this is around. If you just used `bt_loop_timeout_` in place of `timeout_` in L326, I don't think anything changes and removes this variable

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: The condition on L208 is true when there's a new goal that has to be sent to the action server while a previous goal is being executed. So L213-223 is used only to immediately check the future goal handle after a new goal has been sent **while a previous goal was being executed**. If `spin_until_future_complete` times out at this iteration, in the next iteration `goal_updated_` would be false and `check_future_goal_handle()` would never be called again. The L187-198 block is the main block that checks the future goal handle at every iteration if the future has not completed yet.

It is my understanding that `on_wait_for_result` is only meant to run when the BT node is waiting for the accepted action to complete, not when the BT node is waiting for the action server to respond to a `send_goal`. Is that right? In that case, `on_wait_for_result` should be bypassed until the future goal handle completes.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: > If you just used `bt_loop_timeout_` in place of `timeout_`

This could result in the total time that the BT node waits for exceeding the specified `server_timeout` value. Say if `bt_loop_timeout` is 10ms and `server_timeout` is 15ms, the node would wait for the server to timeout for 20ms.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: Definitely. Though, just to confirm; this is a protected member function so it's not part of the API and probably doesn't need to follow the `on_` method name styling. I changed the name since now this function is only responsible for sending a new goal, whereas previously it was responsible for handling the complete new goal behavior. Should I revert it back to `on_new_goal_received`?

SteveMacenski: That is a good point, I thought that was something overridable. I'd personally prefer to keep it the same name but I think your suggestion is also valid

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
SteveMacenski: But you still do the subtraction to find if its larger, you just don't need this intermediary variable. `timeout_` is only used in this line https://github.com/ros-planning/navigation2/blob/5084f5617ae65c4973ced077f536263124d78d4e/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L326 which could be replaced as 

```
auto timeout = remaining > bt_loop_timeout_ ? bt_loop_timeout_ : remaining;
```

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
SteveMacenski: > It is my understanding that on_wait_for_result is only meant to run when the BT node is waiting for the accepted action to complete, not when the BT node is waiting for the action server to respond to a send_goal. 

Thinking about this from a client-server perspective, I don't think the client (the BT node) cares about what the state of its request is in (sending goal, processing goal, etc) since for all it cares, once it sends, its being crunched. So I'd assume that would be called too any time its processing. Though knowing what FollowPath does, I'm not sure if having a new goal matters when its still sending the current goal - I don't think we can preempt that easily so maybe you're right we should keep it separate. 

I'm concerned about the `goal_sent_` variable - its not reset on failure / halt / terminal conditions. I think it could cause state to bleed into multiple requests. For instance, its only set back to false once accepted, but if rejected, it remains true which would impact the next request

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: Ah yes, you're right!

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: Yup, that's right. I've missed a few important places where `goal_sent_` should have been reset.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -91,6 +97,18 @@ bool BtActionServer<ActionT>::on_configure()
...
+  // Get parameter for monitoring with Groot via ZMQ Publisher
+  node->get_parameter("enable_groot_monitoring", enable_groot_monitoring_);
+  node->get_parameter("groot_zmq_publisher_port", groot_zmq_publisher_port_);
+  node->get_parameter("groot_zmq_server_port", groot_zmq_server_port_);
+
+  // Get parameters for BT timeouts
+  int timeout;
+  node->get_parameter("bt_loop_timeout", timeout);
+  bt_loop_timeout_ = std::chrono::milliseconds(timeout);
+  node->get_parameter("default_server_timeout", timeout);
```
SteveMacenski, line 109: Should this be `server_timeout` and `bt_loop_rate` (or `bt_loop_duration`)? The bt loop is used to set the running rate, so I'd think that's the best way to semantically describe it. Because we run at a certain rate, it relates to the timeout, but the timeout isn't it's main purpose (to regulate the BT tick rate). `server_timeout` is what we use in other places in the code base to describe the same thing as well

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -272,18 +304,44 @@ class BtActionNode : public BT::ActionNodeBase
...
+  }
 
-    if (rclcpp::spin_until_future_complete(node_, future_goal_handle, server_timeout_) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+  /**
+   * @brief Function to check response from the action server when a new goal is sent
+   * @return boolen True if future_goal_handle_ returns SUCCESS, False when it returns TIMEOUT
+   */
+  bool check_future_goal_handle()
```
SteveMacenski, line 316: Rename maybe to `if_goal_received`, I think semantically it helps explain what's happening here more and what a true vs a false return would actually mean. Maybe not exactly that name, but something similar?

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -107,9 +109,13 @@ class BtServiceNode : public BT::SyncActionNode
    */
   BT::NodeStatus tick() override
   {
-    on_tick();
```
SteveMacenski, line 110: I'm debating this question - do we actually _need_ to change the service node at all? These should be short running computations and the `future_result_` doesn't give you an accepted / declined feedback the way actions do to wait until accepted before continuing (since that ack would actually be the response itself).

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -313,6 +371,15 @@ class BtActionNode : public BT::ActionNodeBase
...
   // new action goal is sent or canceled
   std::chrono::milliseconds server_timeout_;
+
+  // The timeout value for BT loop execution
+  std::chrono::milliseconds bt_loop_timeout_;
+
+  // To track the action server response when a new goal is sent
+  std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>
+  future_goal_handle_;
+  bool goal_sent_{false};
```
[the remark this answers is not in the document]
SteveMacenski: I think some of my confusion in reading this is just naming, some of the variables aren't as descriptive of what they're being used for as I'd like (`goal_sent_` `check_future_goal_handle` come to mind) or what the return of true/false would mean by the function name.

SteveMacenski:
Should as part of this PR we increase the action send_goal timeout so that we remove the issue people are seeing? How high should we make it?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -174,7 +176,25 @@ class BtActionNode : public BT::ActionNodeBase
       // user defined callback
       on_tick();
 
-      on_new_goal_received();
+      send_new_goal();
+    }
+
+    // if new goal was sent and action server has not yet responded
+    // check the future goal handle
+    if (goal_sent_) {
```
SteveMacenski, line 184: Could `goal_sent_` be replaced by a check on `goal_handle_` instead? If its `reset()` after each goal, then it should be a nullpointer while the goal request is being processed without needing to explicitly track this state variable. 

We'd need to make sure in `halt` and all other terminal conditions that the `goal_handle_` is `reset()` so that its only valid when no goal is sent yet or its being sent currently.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
naiveHobo: `bt_loop_duration` sounds good. `bt_loop_rate` suggests this parameter takes a frequency value.
`server_timeout` for an individual BT node can be overridden through the BT node input ports. If the user does not set the `server_timeout` input port on a BT node, it uses the default server timeout value that we put on the blackboard. `default_server_timeout` would make semantic sense to describe the actual purpose of this parameter. What do you think?

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
naiveHobo: I agree. The only thing is that it cloud potentially slow down the tick rate if say a server takes some time to respond or the service fails for whatever reason. That being said, I don't think these changes to the BT service nodes would affect the performance in any way, might as well add them to avoid any potential issues in the future?

naiveHobo:
> Should as part of this PR we increase the action send_goal timeout so that we remove the issue people are seeing? How high should we make it?

I've not been able to reproduce the timeout issues that people are seeing. I'm thinking the timeout would be dependent on the system specifications so difficult to come up with a single value that works every time. 100ms server timeout should be large enough to account for most systems right?

SteveMacenski:
> 100ms server timeout should be large enough to account for most systems right?

Yeah, that seems super high. I would say 20-50ms. One measurement a user made said they were seeing as high as 18ms on their highly loaded system. So 20 I think would cover it and also only slow the BT by at most 1 tick response. Though I like "final" solutions, so going up to 40ms or 50ms seems like a way to just put this to bed (100ms seems like its high enough that might be a real failure to me if the network can't process a request it should be able to do in 10 nanoseconds in 100 milliseconds).

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -272,18 +316,38 @@ class BtActionNode : public BT::ActionNodeBase
         }
       };
 
-    auto future_goal_handle = action_client_->async_send_goal(goal_, send_goal_options);
+    goal_handle_.reset();
```
SteveMacenski, line 319: I don't think this reset is required, its about to be overridden which implicitly destructs the old object

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -272,18 +316,38 @@ class BtActionNode : public BT::ActionNodeBase
...
+    future_goal_handle_ = action_client_->async_send_goal(goal_, send_goal_options);
+    time_goal_sent_ = node_->now();
+    time_elapsed_since_goal_sent_ = std::chrono::milliseconds(0);
```
SteveMacenski, line 322: I'd argue sending the `elapsed_time` as a function input is easier to logically track, but this also works

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -272,18 +316,38 @@ class BtActionNode : public BT::ActionNodeBase
...
+  /**
+   * @brief Function to check if the action server acknowledged a new goal
+   * @return boolean True if future_goal_handle_ returns SUCCESS, False otherwise
+   */
+  bool is_future_goal_handle_complete()
+  {
+    auto timeout = server_timeout_ > bt_loop_duration_ ? bt_loop_duration_ : server_timeout_;
+    auto result = rclcpp::spin_until_future_complete(node_, future_goal_handle_, timeout);
+
+    time_elapsed_since_goal_sent_ =
```
SteveMacenski, line 334: Shouldn't this be computed very first in the `tick`? Also I'm looking at `timeout = ...` and there's no looking at elapsed if less than `server_timeout_` or `bt_loop_duration_` duration remains to spin only the remainder.  I believe you had that before using `remaining` time.

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
SteveMacenski: OK, sounds good. I looked over the implementation and looks good to me

SteveMacenski:
If there's some test we can add to show that this works the way we expect it to with timeouts < loop duration and timeouts > loop duration with a server that blocks for < loop duration and > loop duration, that would be great to make sure. I think we cover the bases here, but this is a subtle thing to reset the state on and I want to be sure that this works as expected, and with multiple requests, if at all possible

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -272,18 +318,44 @@ class BtActionNode : public BT::ActionNodeBase
         }
```
KavenYau, line 318: Should't we check if the `goal_handle_` is `nullptr`? because it seems that you call `goal_handle_.reset()` in the timeout handling segment.

SteveMacenski: If we get a `result`, there shouldn't be any situation where the handle is a nullpointer, this is a callback on success / failure / cancel. If its a nullpointer, we have an issue where we can't then get the `get_goal_id()` in order to validate that this is the proper response to set the flags that the action has completed to return a result to the behavior tree in `tick()`. 

If this is an issue, we need to rethink how we deal with the goal handles in the bt action node.

The `goal_handle_` I don't think is set to anything when it is reset due to timeouts. The future to the goal handle (https://github.com/ros-planning/navigation2/blob/f83612d28ce0db76436cf711071129eb17099c8f/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L321) is still spinning to get valid. The `goal_handle_` from the future isnt set until it spins and gets back a successful goal acceptance (https://github.com/ros-planning/navigation2/blob/f83612d28ce0db76436cf711071129eb17099c8f/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L351) so I think in (all?) most cases, that should be a nullpointer anyway.

@naiveHobo if that is correct, should the checks on `goal_handle_` in some places actually be `future_goal_handle_` and we reset `future_goal_handle_` on timeout? I think we may have been mixing these up at various points. Should it actually be:

```
    if (future_goal_handle_) {
      auto elapsed = (node_->now() - time_goal_sent_).to_chrono<std::chrono::milliseconds>();
      if (!is_future_goal_handle_complete(elapsed)) {
        // return RUNNING if there is still some time before timeout happens
        if (elapsed < server_timeout_) {
          return BT::NodeStatus::RUNNING;
        }
        // if server has taken more time to respond than the specified timeout value return FAILURE
        RCLCPP_WARN(
          node_->get_logger(),
          "Timed out while waiting for action server to acknowledge goal request for %s",
          action_name_.c_str());
        future_goal_handle_.reset();
      }
    }
```

?? We track and reset the `future_goal_handle_` instead? Before we set, we should default the `future_goal_handle_` to nullpointer. Then if we send a goal, it becomes non-null to then spin the future. Once the future is complete (or times out or fails) we should reset the `future_goal_handle_` instead back to nullpointer? I'm not sure anything relating to `send_goal` and `is_future_goal_handle_complete` needs `goal_handle_`, I think it needs `future_goal_handle_` checks, yeah?

SteveMacenski: I think if we remove the `goal_handle_.reset()`'s for this logic in `tick()` and `is_future_goal_handle_complete()`, then the only remaining ones are in `halt` and the one at the end of `tick()` once a final solution is available. **Would that make things work well?** The other 4 could be the `future_goal_handle_.reset()` I think...

(this is overall a subtle issue so correct me if I missed something)

SteveMacenski: At the end of `tick()` I agree, because we wouldn't have gotten there without a `result` in that callback. Though at the end of `halt()` are we _guaranteed_ to have received a result? I'm actually not 100% sure about that. Does the `spin_until_future_complete` spin until the action server accepts the cancellation or until the cancellation is actually complete? 

If it just waits for the cancel to be accepted, in the simple action server (https://github.com/ros-planning/navigation2/blob/9f0e85f312770a63eeac5b37fc4b8745c38187d3/nav2_util/include/nav2_util/simple_action_server.hpp#L135-L149) that doesn't promise that we have actually completed that cancel (I think that happens in https://github.com/ros-planning/navigation2/blob/9f0e85f312770a63eeac5b37fc4b8745c38187d3/nav2_util/include/nav2_util/simple_action_server.hpp#L523 `canceled`, probably)

As such, maybe its best for the goal handle reset to be in end of tick and in the result callback after the quantities are set if the goal IDs match?

naiveHobo: Yup, got it. I'll remove the reset at the end of `halt()`. Although, I don't think we should reset the goal handle in the result callback. It might be useful to have access to the last goal through the goal handle in the `on_success()`, `on_aborted()`, `on_cancelled()` functions after the goal result is available. The reset at the end of `tick()` should ensure that the goal handle is always reset after a goal.

SteveMacenski:
```
/opt/overlay_ws/src/navigation2/nav2_behavior_tree/test/plugins/action/test_bt_action_node.cpp:328
  ticks
    Which is: 6
  9
```

One of the new tests are failing - actually both of the testing runs had different values (6 vs 8)

---

I'm happy with this overall, I don't think I see any other issues needing to be addressed. In light of that, @tzskp1 and @KavenYau might want to take a look / test again to see if it resolves the issues in #2305. However, since those are currently existing issues, as long as this doesn't make it any **worse**, I'm inclined to merge this since we still don't know root cause of those issues and this is supposed to fix something non-trivial. 

@naiveHobo however, his PR should make the CI not flaky anymore, and I'm seeing alot of failures including

```
[bt_navigator-10] [WARN] [1620137230.106779840] [bt_navigator_rclcpp_node]: Timed out while waiting for action server to acknowledge goal request for compute_path_to_pose
[planner_server-8] [INFO] [1620137230.119440375] [global_costmap.global_costmap]: Received request to clear entirely the global_costmap
[bt_navigator-10] [ERROR] [1620137230.122825667] [BehaviorTreeEngine]: Behavior tree threw exception: SyncActionNode MUST never return RUNNING. Exiting with failure.
[bt_navigator-10] [ERROR] [1620137230.131459446] [bt_navigator]: Goal failed
[bt_navigator-10] [WARN] [1620137230.132364221] [bt_navigator]: [navigate_to_pose] [ActionServer] Aborting handle.
[tester_node-13] [INFO] [1620137230.134366128] [nav2_tester]: ?[1;37;44mGoal failed with status code: 6?[0m
[tester_node-13] [ERROR] [1620137230.134978606] [nav2_tester]: ?[1;37;41mTest FAILED?[0m
```

Which is basically the new version of the `send_goal failed` error (and with a new really not good error of that exception from the BT Engine). I think there's something still wrong here -- certainly with state tracking for that new exception -- but also that it didn't fix the error caused by the goal request timing out (did you update the timeout from 10ms to 20ms?)

naiveHobo:
Not quite sure what the issue is now, I've updated all the server timeouts to 20ms. In the tests I added for the BT action node changes, the number of ticks before timeout varies (most likely due to ros clock timing) but I can confirm that the total time that the bt action node waits for the server to acknowledge is always equal to the server timeout value (~1-2ms).

SteveMacenski:
One of the tests is still failing and from the tests that are failing 
```

[tester_node-13] [INFO] [1620163729.442322811] [nav2_waypoint_tester]: Sending goal request...
[tester_node-13] [ERROR] [1620163729.444101914] [nav2_waypoint_tester]: Goal rejected
```

It seems like this has some issues where goals are being rejected now? I don't believe this is happening in main, but I'm not 100% sure of that either. 

The release test otherwise succeeded other than the bt action server new test that genuinely looks to have failed

naiveHobo:
> It seems like this has some issues where goals are being rejected now?

That nav2_waypoint_tester failure you mentioned seems unrelated to this PR, it happened due to a gazebo server failure I believe:

```
[gzserver-1] gzserver: /usr/include/boost/smart_ptr/enable_shared_from_this.hpp:51: boost::shared_ptr<X> boost::enable_shared_from_this<T>::shared_from_this() [with T = gazebo::transport::Node]: Assertion `p.get() == this' failed.
[ERROR] [gzserver-1]: process has died [pid 12244, exit code -6, cmd 'gzserver -s libgazebo_ros_init.so --minimal_comms /opt/overlay_ws/src/navigation2/nav2_system_tests/worlds/turtlebot3_ros2_demo.world'].
```

I just retriggered the CI and everything passed.

SteveMacenski:
One day... one day soon we'll have non-flaky CI again... I keep telling myself to sleep at night :wink:

SteveMacenski (approved):
I'm happy with this! Thanks for all the hard work, this will be really nice to have resolved going into Galactic release :-) 

I'm letting CI to run again but assuming only issues outside of the scope of this work are found, I'll merge. @tzskp1 please test this, it does make about 70% of the changes your PR requests (except for the result callback checks) that might actually resolve the issue you see.

If it does not, we can address your PR separately

SteveMacenski:
@naiveHobo I jumped the gun a little, can you submit a PR to update the docs for the new parameter(s) this added?

PR #2321, opened 2021-04-22
removed name parameter from rviz_launch.py

shonigmann:
---

## Basic Info
This should be a pretty trivial change to avoid warnings when launching `rviz_launch.py` included in the bringup directory (discussed a bit over in ros2/rviz#671). Setting the Node name parameter remaps the name of all 4 nodes launched by `rviz_launch.py` to be the same name. This results in warnings on launch and when listing ros2 nodes. As far as I can tell, removing the name parameter does not affect behavior of the launch script and avoids the node name clash. 

| ------ | ----------- |
| Robotic platform tested on | () |

---

- removed `name='rviz2'` from rviz_launch.py to eliminate duplicate node name warning

---

- This could be extended to other branches that include `rviz_launch.py` (e.g. `foxy-devel`), but I figured I'd raise it for main first to make sure I  didn't overlook anything silly

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/bringup/launch/rviz_launch.py
+++ b/nav2_bringup/bringup/launch/rviz_launch.py
@@ -57,7 +57,6 @@ def generate_launch_description():
         condition=UnlessCondition(use_namespace),
         package='rviz2',
         executable='rviz2',
-        name='rviz2',
         arguments=['-d', rviz_config_file],
         output='screen')
 
@@ -69,7 +68,6 @@ def generate_launch_description():
         condition=IfCondition(use_namespace),
         package='rviz2',
         executable='rviz2',
-        name='rviz2',
         namespace=namespace,
         arguments=['-d', namespaced_rviz_config_file],
         output='screen',


SteveMacenski:
I'm fairly certain that `name` is a required field, did you actually test this locally that this still works? It seems like a better solution would be to do the node name remaps in the code as Michael suggested

shonigmann:
Looking at [ros2/launch_ros/actions/node.py](https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/actions/node.py), name is indeed listed as optional:

        If the name is not given (or is None) then no name is passed to
        the node on creation and instead the default name specified within the
        code of the node is used instead.

`rviz_launch.py` runs fine for me (at least in Foxy) with this change applied. If you'd prefer the more explicit solution though, I'd be happy to close and resubmit with that approach instead.

SteveMacenski (approved):

PR #2328, opened 2021-04-29
Open option to use parent node in lifecycle Manger Client

BriceRenaudeau:
In the process of removing all service nodes (#816).
After removing the service_node from lifecycle Manager (#2267), I propose an option to remove it from lifecycle Manager Clients.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #816 |
| Robotic platform tested on | Wyca robotics simulation and robot|

---

* I added the option to give a parent_node in the `LifecycleMangerClient` constructor to be used instead of creating a "_service_client" node.
* I used the `nav2_util::ServiceClient` for the implementation.

#### For Maintainers: 

### DIFF
--- a/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
+++ b/nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
@@ -28,17 +28,30 @@ using nav2_util::geometry_utils::orientationAroundZAxis;
 
 LifecycleManagerClient::LifecycleManagerClient(
   const std::string & name,
-  const std::string & ns, std::shared_ptr<rclcpp::Node> parent_node)
+  const std::string & ns)
 {
   manage_service_name_ = name + std::string("/manage_nodes");
   active_service_name_ = name + std::string("/is_active");
 
   // Create the node to use for all of the service clients
-  if (parent_node == nullptr) {
-    node_ = std::make_shared<rclcpp::Node>(name + "_service_client", ns);
-  } else {
-    node_ = parent_node;
-  }
+  node_ = std::make_shared<rclcpp::Node>(name + "_service_client", ns);
+
+  // Create the service clients
+  manager_client_ = std::make_shared<nav2_util::ServiceClient<ManageLifecycleNodes>>(
+    manage_service_name_, node_);
+  is_active_client_ = std::make_shared<nav2_util::ServiceClient<std_srvs::srv::Trigger>>(
+    active_service_name_, node_);
+}
+
+LifecycleManagerClient::LifecycleManagerClient(
+  const std::string & name,
+  std::shared_ptr<rclcpp::Node> parent_node)
+{
+  manage_service_name_ = name + std::string("/manage_nodes");
+  active_service_name_ = name + std::string("/is_active");
+
+  // Use parent node for service call and logging
+  node_ = parent_node;
 
   // Create the service clients
   manager_client_ = std::make_shared<nav2_util::ServiceClient<ManageLifecycleNodes>>(
@@ -87,7 +100,7 @@ LifecycleManagerClient::is_active(const std::chrono::nanoseconds timeout)
     node_->get_logger(), "Waiting for the %s service...",
     active_service_name_.c_str());
 
-  if (!is_active_client_->wait_for_service(timeout)) {
+  if (!is_active_client_->wait_for_service(std::chrono::seconds(1))) {
     return SystemStatus::TIMEOUT;
   }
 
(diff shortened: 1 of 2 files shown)


nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -28,17 +28,23 @@ using nav2_util::geometry_utils::orientationAroundZAxis;
 
 LifecycleManagerClient::LifecycleManagerClient(
   const std::string & name,
-  const std::string & ns)
+  const std::string & ns, std::shared_ptr<rclcpp::Node> parent_node)
```
SteveMacenski, line 31: Add this to a new line

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -28,17 +28,23 @@ using nav2_util::geometry_utils::orientationAroundZAxis;
...
 {
   manage_service_name_ = name + std::string("/manage_nodes");
   active_service_name_ = name + std::string("/is_active");
 
   // Create the node to use for all of the service clients
-  node_ = std::make_shared<rclcpp::Node>(name + "_service_client", ns);
+  if (parent_node == nullptr) {
```
SteveMacenski, line 37:
```suggestion
  if (!parent_node) {
```

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -28,17 +28,23 @@ using nav2_util::geometry_utils::orientationAroundZAxis;
 
 LifecycleManagerClient::LifecycleManagerClient(
```
SteveMacenski, line 29: Actually, you should just create a new constructor for `LifecycleManagerClient`. It doesn't make sense for us to have both `name` `ns` and also `node` since the only reason name and namespace are there are to create a node. So you can create another parallel constructor that instead takes in a node

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -87,15 +94,14 @@ LifecycleManagerClient::is_active(const std::chrono::nanoseconds timeout)
...
-    rclcpp::FutureReturnCode::SUCCESS)
-  {
+  try {
+    response = is_active_client_->invoke(request, timeout);
+  } catch (std::runtime_error &) {
     return SystemStatus::TIMEOUT;
   }
 
-  if (future_result.get()->success) {
+  if (response.get()->success) {
```
SteveMacenski, line 104: Why `get()`? Its a shared pointer, you can just access `success`.

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -112,7 +118,7 @@ LifecycleManagerClient::callService(uint8_t command, const std::chrono::nanoseco
     node_->get_logger(), "Waiting for the %s service...",
     manage_service_name_.c_str());
 
-  while (!manager_client_->wait_for_service(std::chrono::seconds(1))) {
+  while (!manager_client_->wait_for_service(timeout)) {
```
SteveMacenski, line 121: Revert this back to 1s, this is specific to the lifecycle management and we want a generous timeout

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
```diff
@@ -123,15 +129,12 @@ LifecycleManagerClient::callService(uint8_t command, const std::chrono::nanoseco
...
     node_->get_logger(), "Sending %s request",
     manage_service_name_.c_str());
-  auto future_result = manager_client_->async_send_request(request);
-
-  if (rclcpp::spin_until_future_complete(node_, future_result, timeout) !=
-    rclcpp::FutureReturnCode::SUCCESS)
-  {
+  try {
+    auto future_result = manager_client_->invoke(request, timeout);
+    return future_result.get()->success;
```
SteveMacenski, line 134: As well as the the `get()`s, these are just shared pointers

SteveMacenski (changes requested):

nav2_lifecycle_manager/src/lifecycle_manager_client.cpp
BriceRenaudeau: No, it is used to create the service client
`manage_service_name_ = name + std::string("/manage_nodes");`
`manager_client_ = node_->create_client<ManageLifecycleNodes>(manage_service_name_);`

BriceRenaudeau: Only the `ns` is used only to create the node.
So it's ok to create a new constructor with `name` and `parent node`

SteveMacenski (approved):

SteveMacenski:
LGTM, suspiciously, 2 of the same tests in both CI test runs failed. I'm retriggering, if it passes I'll merge (if not, might require some looking on your end, might be breaking something)

BriceRenaudeau:
I don't understand the test failures.

PR #2330, opened 2021-05-03
Validate codecov script against checksums.

nuclearsandwich:
After the recent Codecov security incident[1] I've been reviewing
codecov usage across ROS repositories.

This repository is fetching the codecov bash uploader without performing
the recommended validation step.

The validation step does not appear to have been widely explained or
publicised and even the official codecov GitHub action was not
validating the script until the recent security incident.

I have made an attempt to validate the bash uploader here.

[1]: https://about.codecov.io/security-update/

PR #2335, opened 2021-05-10
Fix script error in codecov validation.

nuclearsandwich:
`local` in bash can only be used within a function context.

Fixes the regression reported in https://github.com/ros-planning/navigation2/pull/2330#issuecomment-837032101

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -290,7 +290,7 @@ _steps:
       working_directory: /opt/overlay_ws
       command: |
         curl -s https://codecov.io/bash > codecov
-        local codecov_version=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2)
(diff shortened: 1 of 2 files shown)


ruffsl (approved):
Looks like code coverage upload was successful:
https://codecov.io/github/ros-planning/navigation2/commit/8afc0eedfd864324017f3ce20d403244ebe8132c

SteveMacenski:
Thanks for the help again @nuclearsandwich !

nuclearsandwich:
There are circumstances which prevent me from using Circle CI's GitHub login which is apparently required to see the output of builds there. Was the failing test unrelated to this change?

nuclearsandwich:
> Thanks for the help again @nuclearsandwich !

Apologies for the regression. I was going fast and copy-pasting similar changes in a bunch of different places and clearly missed that I wasn't in a function context.

PR #2339, opened 2021-05-14
Move CI from ROS2 master to Rolling

ruffsl:
Related: #2337

PR #2342, opened 2021-05-17
Rolling ci with repos

wep21:
https://github.com/ros-planning/navigation2/pull/2339 without removing unreleased packages from repos.

### DIFF
--- a/.dockerhub/debug/hooks/build
+++ b/.dockerhub/debug/hooks/build
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-export FROM_IMAGE=osrf/ros2:nightly
+export FROM_IMAGE=ros:rolling
 export FAIL_ON_BUILD_FAILURE=""
 export UNDERLAY_MIXINS="release ccache"
 export OVERLAY_MIXINS="debug ccache coverage-gcc"
--- a/.dockerhub/distro.Dockerfile
+++ b/.dockerhub/distro.Dockerfile
@@ -6,14 +6,14 @@
 #
 # Example build command:
 # export DOCKER_BUILDKIT=1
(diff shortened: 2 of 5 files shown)


ruffsl:
> #2339 without removing unreleased packages from repos.

Everything but `ompl` has indeed been released into rolling, no?

http://packages.ros.org/ros2/ubuntu/dists/focal/main/binary-amd64/Packages

The `.repos` file in #2339 does produce a building Dockerfile.

SteveMacenski:
@ruffsl I agree that it looks like they're release, are we not calling rosdep in the right place now that the underlay workspace isn't needed for those? That is odd, maybe need to apt update?

ruffsl:
> are we not calling rosdep in the right place now that the underlay workspace isn't needed for those? That is odd, maybe need to apt update?

Not sure what your asking. Are you asking if there are still unresolved dependencies? By commenting out the source repos of released rolling packages from the underlay's `.repos` file, the RUN directive invoking rosdep for the overlay workspaces will now instead be the layer where most of nav2's external dependencies will be installed.

SteveMacenski:
@wep21 I appreciate the PR and help, but I think Ruffin's got this one. He knows the CI system for nav2 the best of anyone and is underway.

SteveMacenski:
Closing - Ruffin's PR is merged and building, thanks for the help @wep21!

ruffsl:
@wep21 , could you include 7d8afb2 in a new PR to fix linting. I'd like to see if CI is caching as expected after merging https://github.com/ros-planning/navigation2/pull/2343 .

wep21:
@ruffsl Sure. https://github.com/ros-planning/navigation2/pull/2346

PR #2347, opened 2021-05-18
Update invalid parameter exception handling

jacobperron:
Since https://github.com/ros2/rclcpp/pull/1673, uninitialized parameters are allowed (ie. an exception is not thrown). Instead, if the parameter is not set by the time it is accessed, then an exception is thrown.

Note, this is a decision upstream to change this particular API for the upcoming Galactic release. Ultimately, we're hoping to avoid a deprecation cycle by landing this prior to the release date.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/test/unit/declare_parameter_test.cpp
+++ b/nav2_costmap_2d/test/unit/declare_parameter_test.cpp
@@ -65,8 +65,9 @@ TEST(DeclareParameter, useInvalidParameter)
 
   layer.initialize(&layers, "test_layer", &tf, node, nullptr, nullptr);
 
+  layer.declareParameter("test2", rclcpp::PARAMETER_STRING);
   try {
-    layer.declareParameter("test2", rclcpp::PARAMETER_STRING);
+    std::string val = node->get_parameter("test_layer.test2").as_string();
     FAIL() << "Incorrectly handling test_layer.test2 parameter which was not set";
   } catch (rclcpp::exceptions::ParameterUninitializedException & ex) {
     SUCCEED();
--- a/nav2_costmap_2d/test/unit/declare_parameter_test.cpp
+++ b/nav2_costmap_2d/test/unit/declare_parameter_test.cpp
@@ -68,7 +68,7 @@ TEST(DeclareParameter, useInvalidParameter)
   try {
     layer.declareParameter("test2", rclcpp::PARAMETER_STRING);
     FAIL() << "Incorrectly handling test_layer.test2 parameter which was not set";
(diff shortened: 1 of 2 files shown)


nav2_costmap_2d/test/unit/declare_parameter_test.cpp
```diff
@@ -68,7 +68,7 @@ TEST(DeclareParameter, useInvalidParameter)
   try {
     layer.declareParameter("test2", rclcpp::PARAMETER_STRING);
     FAIL() << "Incorrectly handling test_layer.test2 parameter which was not set";
-  } catch (rclcpp::exceptions::NoParameterOverrideProvided & ex) {
+  } catch (rclcpp::exceptions::ParameterUninitializedException & ex) {
```
ivanpauno, line 71: `declare_parameter` cannot throw `ParameterUninitializedException`, I'm not sure if `declareParameter` is trying using some "declare or get" logic, or sth like that.

ivanpauno: it seems to be doing that https://github.com/ros-planning/navigation2/blob/30b405c58e6d53ba8c96381416bc4679d35a1483/nav2_costmap_2d/src/layer.cpp#L88

the change looks okay to me then

jacobperron: Thanks, nice catch. I'll fix this.

jacobperron: I've moved the exception handling to the place where `get_parameter` is called (8c44967). Though, it's not clear to me how this test was passing originally since it looks like the exception was not being propagated though `Layer::declareParameter`. I'll do some more testing locally.

jacobperron: Nevermind, I was confused. I think this fixes the test (8a33fc2)

SteveMacenski:
What's the ordering of this? Right now `main` is in its Galactic form, would this change be in Galactic in day 1 (e.g. if I released Nav2 **right now**, which is my plan tomorrow, would merging this PR make it fail?). Else, I can plan on merging this after the release.

SteveMacenski (approved):

jacobperron:
I'm planning for this to be in Galactic before the release date. Here's a backport: https://github.com/ros2/rclcpp/pull/1681
I'll make sure a release gets out this week.

SteveMacenski:
OK, I'll merge after that goes in, just to make sure we don't end up in a state where this blocks Nav2's galactic jobs to be released on Day 1

PR #2346, opened 2021-05-18
Repalce deprecated headers of tf2_sensor_msgs

wep21:
### Related PR
https://github.com/ros-planning/navigation2/pull/2342

PR #2350, opened 2021-05-19
Replace deprecated headers

jacobperron:
tf2_geometry_msgs.h was replaced by tf2_geometry_msgs.hpp in https://github.com/ros2/geometry2/pull/418.
tf2_sensor_msgs.h was replaced by tf2_sensor_msgs.hpp in https://github.com/ros2/geometry2/pull/416.

This PR overlaps with #2346, which replaces only `tf2_sensor_msgs.h`.

#### For Maintainers: 


SteveMacenski:
Is this a pre or post galactic PR? (e.g. if I released to Galactic tomorrow, which is my plan, would merging this be the correct status for galactic or after galactic?)

SteveMacenski (approved):

PR #2343, opened 2021-05-17
Update CI to use colcon cache

ruffsl:
This updates the CI to make use of colcon cache, a colcon extension that enables the caching of an entire workspace by tracking the state of package source files. This essentially allows the CI to focus on building and testing only packages that have been modified. For more info on colcon cache: https://github.com/ruffsl/colcon-cache/pull/12 . 

Changes in this PR to incorporate colcon cache include:

* Add cache key from main branch as a fallback cache to restore from
  * now that invalid packages can be detected and rebuilt on demand with colcon cache 
* Cache only the build, install, log folders, not the entire workspace directory
  * as not to include or restore the deterministic checksum files
* Hash underlay repos into checksum, rather than active workspace
  * so that source file changes don't result in missed active workspace cache restore
  * so that source file changes can still bust overlay workspace cache restore
* clean up built and test logic in bash using env
  * to clarify which and how packages are being selected
* Add `cache_test` parameter
  * to bypass caching of tests for nightly jobs
  * so nighy jobs always retest every package
* Don't clean or ignore .git folders from cloned or copied source repos
  * so colcon cache can use revision control in package to track source changes

Pending on: https://github.com/ros-planning/navigation2/pull/2339

PR #2353, opened 2021-05-20
Adjust CI to enable cache sharing with forked PRs

ruffsl:
So for a recent PR https://github.com/ros-planning/navigation2/pull/2349 after merging https://github.com/ros-planning/navigation2/pull/2343 , I noticed that it's cache restore step was not finding the matching cache key provide by the main branch:

## Prior existing cache stored by main branch CI job

```
Creating cache archive...
Uploading cache archive...
Stored Cache to overlay_ws-v1-arch1-linux-amd64-6_85-main-<no value>-C8DP7DcUx71NZcYchPyQBH0ElDOZYfM4Uez8CyHmtNE=-1621429706
  * /opt/overlay_ws/build
  * /opt/overlay_ws/install
  * /opt/overlay_ws/log
https://app.circleci.com/pipelines/github/ros-planning/navigation2/5359/workflows/9c999dde-f6c0-4007-b8a3-87d6e091638f/jobs/19741/parallel-runs/0/steps/0-118
```

## Later failed cache restore from forked PR CI job
```
No cache is found for key: overlay_ws-v1-arch1-linux-amd64-6_85-pull/2349-2349-C8DP7DcUx71NZcYchPyQBH0ElDOZYfM4Uez8CyHmtNE=
No cache is found for key: overlay_ws-v1-arch1-linux-amd64-6_85-main-<no value>-C8DP7DcUx71NZcYchPyQBH0ElDOZYfM4Uez8CyHmtNE=
https://app.circleci.com/pipelines/github/ros-planning/navigation2/5354/workflows/97ef33a5-4b3d-4cce-bbf8-2598807fbf27/jobs/19728/parallel-runs/0/steps/0-116
```

This was because CircleCI does not share caches with forked repos, unless sharing of environment variables is enabled:

> enabling the sharing of environment variables will enable cache sharing between the original repo and all forked builds.
> https://circleci.com/docs/2.0/oss/#caching

We can simply enable sharing of environment variables in the repo's circleci settings, but we currently have two secret tokens stored in the repo's circleci environment. The codecov token and the DockerHub build trigger token. Turns out codecov token should be unnecessary for public repos when using major CI's like CircleCI or GH actions, so we could delete that variable from the current CircleCI environment.

> note that this is a public repo, so a token shouldn't be necessary
> https://github.com/codecov/codecov-circleci-orb/issues/12
> https://about.codecov.io/blog/tokenless-uploads-for-github-actions/

As for the DockerHub build trigger token, given GH actions has a slightly more nuance control for protecting repo secrets:

> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
> https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

I've ported the cron job config from https://github.com/ros-planning/navigation2/pull/2344 into a GH action workflow so we can move the DockerHub secret into GH actions. This also has the added benefit of allowing the use of GH action event triggers to rebuild the CI image if any of the `package.xml` or `.repos` files are changed, given they're subsequently used to pre-install depencies in the CI image.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,32 +265,6 @@ _commands:
             path: << parameters.workspace >>/test_results
         - store_artifacts:
             path: << parameters.workspace >>/test_results
-    check_apt_updates:
-      description: "Check apt updates"
-      parameters:
-        sourcelist:
-          type: string
-      steps:
-        - run:
-            command: |
-              apt-get update \
-                -o Dir::Etc::sourcelist="<< parameters.sourcelist >>"
-              apt-get --simulate upgrade \
-                -o Dir::Etc::sourcelist="<< parameters.sourcelist >>" \
-                | grep "0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded." \
-                && echo "No apt updates" || exit 1
-    trigger_dockerhub_url:
-      description: "Trigger Dockerhub URL"
-      parameters:
-        data:
-          type: string
-      steps:
-        - run:
-            when: on_fail
-            command: |
-              curl -H "Content-Type: application/json" \
-                --data '<< parameters.data >>' \
-                -X POST ${DOCKERHUB_TRIGGER_URL}
 
 _steps:
   pre_checkout: &pre_checkout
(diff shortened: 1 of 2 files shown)


SteveMacenski:
So to be clear, the github action is just for triggering dockerhub to re-build the base image? Maybe I'm missing some small detail, but if this is just on a daily cron job, doesn't dockerhub have that option itself? I see an autobuild option there.

SteveMacenski:
OK got it, thanks for the detailed explanation

Issue #2303, opened 2021-04-08
Simple Action Server Race condition


tzskp1:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- ROS2 Version:
  - Rolling Source
- Version or commit hash:
  - 8a7a4f261ccfa5db68d5add2fa493b32bacc6f33
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue

1. On the arm device, run tb3_simulation_launch.py according to the official wiki.
2. Nav2 goal in rviz2 panel.
3. Wait for minutes.

I met SIGSEGV on an arm device.
backtrace is here.

```
#0  0x0000007fb6f5b6e4 in __GI___pthread_mutex_lock (mutex=0xa0) at pthread_mutex_lock.c:67
#1  0x0000007fb7edaa7c in __gthread_mutex_lock(pthread_mutex_t*) ()
   from /ros/install/behaviortree_cpp_v3/lib/libbehaviortree_cpp_v3.so
#2  0x0000007fb7ee0788 in std::mutex::lock() () from /ros/install/behaviortree_cpp_v3/lib/libbehaviortree_cpp_v3.so
#3  0x0000007fa30eadf8 in std::lock_guard<std::mutex>::lock_guard (__m=..., this=<synthetic pointer>)
    at /usr/include/c++/7/bits/std_mutex.h:162
#4  rclcpp_action::ClientGoalHandle<nav2_msgs::action::ComputePathThroughPoses>::get_status (this=<optimized out>)
    at /ros2_master/install/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp:101
#5  nav2_behavior_tree::BtActionNode<nav2_msgs::action::ComputePathThroughPoses>::should_cancel_goal (this=0x7f7c00a050)
    at /ros/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp:250
#6  nav2_behavior_tree::BtActionNode<nav2_msgs::action::ComputePathThroughPoses>::halt (this=0x7f7c00a050)
    at /ros/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp:223
#7  0x0000007fb7f05234 in std::function<void (BT::TreeNode*)>::operator()(BT::TreeNode*) const ()
   from /ros/install/behaviortree_cpp_v3/lib/libbehaviortree_cpp_v3.so
[...]

PR #2356, opened 2021-05-21
Fix SEGV caused by goal_handle_ which is nullptr while result_callback called

KavenYau:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2303, #2305 |
| Robotic platform tested on | gazebo simulation|

---

- Checking `goal_handle_` if it is `nullptr`.
- If the `future_goal_handle_` is not `nullptr`, try to get `goal_handle_` in `result_callback`.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -312,19 +312,12 @@ protected:
     auto send_goal_options = typename rclcpp_action::Client<ActionT>::SendGoalOptions();
     send_goal_options.result_callback =
       [this](const typename rclcpp_action::ClientGoalHandle<ActionT>::WrappedResult & result) {
-        if (!this->goal_handle_) {
-          if (!future_goal_handle_) {
-            // Shouldn't run into this block scope
-            RCLCPP_WARN(
-              node_->get_logger(),
-              "Goal result for %s available, but it can not get a valid goal_handle_. "
-              "Normally, this message shouldn't be printed.",
-              action_name_.c_str());
-            return;
-          }
-
-          // If the future_goal_handle_ is not nullptr, try to get goal_handle_
-          this->goal_handle_ = future_goal_handle_->get();
+        if (future_goal_handle_) {
+          RCLCPP_DEBUG(
+            node_->get_logger(),
+            "Goal result for %s available, but it hasn't received the goal response yet. "
+            "It's probably a goal result for the last goal request", action_name_.c_str());
+          return;
         }
 
         // TODO(#1652): a work around until rcl_action interface is updated
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -312,6 +312,21 @@ protected:
     auto send_goal_options = typename rclcpp_action::Client<ActionT>::SendGoalOptions();
     send_goal_options.result_callback =
       [this](const typename rclcpp_action::ClientGoalHandle<ActionT>::WrappedResult & result) {
+        if (!this->goal_handle_) {
+          if (!future_goal_handle_) {
+            // Shouldn't run into this block scope
+            RCLCPP_WARN(
+              node_->get_logger(),
+              "Goal result for %s available, but it can not get a valid goal_handle_. "
+              "Normally, this message shouldn't be printed.",
+              action_name_.c_str());
+            return;
+          }
+
+          // If the future_goal_handle_ is not nullptr, try to get goal_handle_
+          this->goal_handle_ = future_goal_handle_->get();
+        }
+
         // TODO(#1652): a work around until rcl_action interface is updated
         // if goal ids are not matched, the older goal call this callback so ignore the result
         // if matched, it must be processed (including aborted)


tzskp1:
I tried to reproduce the problem.
But the problem seems to disappear by upgrade the rclcpp.
Additionally, I observed your patch is working well.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -312,6 +312,21 @@ class BtActionNode : public BT::ActionNodeBase
...
+            RCLCPP_WARN(
+              node_->get_logger(),
+              "Goal result for %s available, but it can not get a valid goal_handle_. "
+              "Normally, this message shouldn't be printed.",
+              action_name_.c_str());
+            return;
+          }
+
+          // If the future_goal_handle_ is not nullptr, try to get goal_handle_
+          this->goal_handle_ = future_goal_handle_->get();
```
KavenYau, line 327: @tzskp1
Perhaps this patch dosen't work, I got a blocking issue on this line.

tzskp1:
Yes, I observed that SEGV error without this patch.

KavenYau:
Do you have a way/unit test to reproduce it quickly?

tzskp1:
No, I have not.
I kept pressing the button patiently until the bug was reproduced...

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
tzskp1: I met the blocking problem too.

SteveMacenski:
> Yes, I observed that SEGV error without this patch.

Does that mean that with this patch you don't see it?

KavenYau:
> does this not solve a problem or introduce a new one?

It's to solve `SEGV`problem, but the solution is infeasible and converts into another problem.

KavenYau:
I think the problem was that it received the previous goal result when waiting for the latest goal response. before the goal response received, `goal_handle_` was `nullptr`. then the `result_callback` called and crashed.

And I also found another problem, if the new sending goal request was a preemption request, and it received the previous goal result when waiting for goal response, due to `goal_handle_` wasn't updated and the `goal_id` was matched, the bt node would return `SUCCESS` and reset `goal_handle_`, though it was an outdated succeeded goal.

@tzskp1 Could you please try the latest commit in your spare time? :smile:

SteveMacenski:
That seems like it should be a debug log, but warn is fine for now for testing (also it would be good to know if it **is** for sure the last goal or we're just guessing it might be). Do we only see this when it is the old goal? I feel like that's something that could be tested with printing some goal IDs periodically to see if it happens in any other situation

tzskp1:
I checked this patch.
Due to hand testing, this patch is probably ok.
The previous patch causes the mismatch of states of navigation2 panel and actual goal handle. this patch seems to solve the mismatch.
Additionaly, I found that rviz2 crashes:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Node has already been added to an executor.
```
But this is just another problem.
Thank you for your help!

SteveMacenski:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Node has already been added to an executor.
```

Yeah, I see that too. I'm not 100% sure if its to do with the nav2 rviz plugin or rviz2 itself. I see the on occasion as well and honestly I don't have an excuse for why I haven't investigated that yet. I'll file a ticket now.

KavenYau:
> This would fix that situation as well, correct?

That's right.

> are we sure that this is only caused by old goals? I thought in the discussion above we talked about this maybe being triggered during the spin_until_future_complete so that would represent the current task.

So far the reason I can see is that it only caused by old goals. 
it would not send goal_result request until receiving goal_response. In other words, it would not receive a result of the new goal before `spin_until_future_complete` returns.

I don't know whether the receiving message order always is consistent with sending message order. If it doesn't guarantee that (Perhaps different `DDS` implementation has different behaviors), the `result` would be ignored and the current goal would never get the `result`. But this is a potential issue which has not appeared yet.

SteveMacenski:
> I don't know whether the receiving message order always is consistent with sending message order. 

After a little bit of looking, no that is not promised yet in ROS2. There is a DDS QoS setting for that, but it has not been exposed to ROS2 (yet). 

> the result would be ignored and the current goal would never get the result. But this is a potential issue which has not appeared yet.

Is there an additional `if(...)` we can make if the `future_goal_handle_` is non-null to check on that case to at least log (if not partially recover) in that case? If the result callback is triggered, the `goal_handle_` is still going to be valid, so I was thinking maybe we have an internal if statement to check on the goal IDs. E.g.

```
        if (future_goal_handle_) {
          if (this->goal_handle_->get_goal_id() == result.goal_id) {
            // do something
          } else {
            // do something else
          }
        }
```

KavenYau:
> ~~the result would be ignored and the current goal would never get the result. But this is a potential issue which has not appeared yet.~~

I have thought about that and I think actually it's not a potential issue, because `rclcpp` could guarantee that it would not send goal_result request until receiving goal_response, unless packet loss occurs. Perhaps the if statement is unnecessary?

SteveMacenski:
> because rclcpp could guarantee that it would not send goal_result request until receiving goal_response

Do you have code to support that? E.g. is that the current behavior of rclcpp or is there something we could add to rclcpp to make that guarantee?

KavenYau:
Yes, it is.

https://github.com/ros2/rclcpp/blob/39bfb30eb0eea70a5863de1aeaa8eb43177ff0f3/rclcpp_action/include/rclcpp_action/client.hpp#L451-L483
https://github.com/ros2/rclcpp/blob/39bfb30eb0eea70a5863de1aeaa8eb43177ff0f3/rclcpp_action/include/rclcpp_action/client.hpp#L723-L753
https://github.com/ros2/rclcpp/blob/39bfb30eb0eea70a5863de1aeaa8eb43177ff0f3/rclcpp_action/include/rclcpp_action/client.hpp#L745
https://github.com/ros2/rclcpp/blob/39bfb30eb0eea70a5863de1aeaa8eb43177ff0f3/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L77
In `async_send_goal`, it calls `send_goal_request` and waits `goal_response_callback` triggered. Once the `goal_response` received, it calls `make_result_aware` which would call `send_result_request` and wait until it receives the result, then it calls`    result_callback_(wrapped_result);`, so that we could receive the result through the callback.


https://github.com/ros2/rclcpp/blob/39bfb30eb0eea70a5863de1aeaa8eb43177ff0f3/rclcpp_action/include/rclcpp_action/client.hpp#L514-L534
https://github.com/ros2/rclcpp/blob/39bfb30eb0eea70a5863de1aeaa8eb43177ff0f3/rclcpp_action/include/rclcpp_action/client.hpp#L725-L728
By the way, another way to get the result is that we call `async_get_result` and `spin_until_future_complete`. But they can't be used on the same `goal_handle`.

SteveMacenski:
Got it - OK, I think that is the last potential issue. Are you happy with this and ready to merge? Any other potential problems we should address?

KavenYau:
Any other potential problems we should address?
So far I haven't seen yet. 

I changed the debug message to `DEBUG` level. And I think it could be merged. 😄

SteveMacenski:
I am very appreciative of your effort on the reliability of actions in nav2 and rclcpp @KavenYau, thanks for another great patch!

PR #2370, opened 2021-05-26
Trigger via env rather than on failure

ruffsl:
To avoid notification noise from failed CI status when apt and ros packages change.


SteveMacenski (approved):
Thanks! I love less emails :-)

PR #2083, opened 2020-11-10
Fix build on macOS (10.x / 11.x)

3wnbr1:
---

## Basic Info

macOS : 11.0.1 (20B5022a)
Build tools : Apple Clang 12.0.x
ROS2 : Foxy (Patch release 3)

---

## Description of contribution in a few bullet points

Building navigation2 on latest macOS 11 (Big Sur) is now possible with a ros2 foxy environment. Previous macOS versions should also work but remains untested. This PR adresses the following fatal build errors (with tests set to OFF) :

#### nav2_util
```
--- stderr: nav2_util
/Users/ewen/ros_ws/navigation2/nav2_util/src/dump_params.cpp:351:33: error: use of undeclared identifier 'basename'
      std::cout << "Usage: " << basename(argv[0]) << "\n";
```

#### nav2_amcl
```
--- stderr: nav2_amcl
/Users/ewen/ros_ws/navigation2/nav2_amcl/src/amcl_node.cpp:404:64: error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char> >') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
      " Verify that data is being published on the %s topic.", scan_topic_);
                                                               ^
/Users/ewen/ros_ws/navigation2/nav2_amcl/src/amcl_node.cpp:413:31: error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char> >') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
      d.nanoseconds() * 1e-9, scan_topic_);
```

#### nav2_behavior_tree
```
--- stderr: nav2_behavior_tree
In file included from /Users/ewen/ros_ws/navigation2/nav2_behavior_tree/plugins/action/reinitialize_global_localization_service.cpp:16:
In file included from /Users/ewen/ros_ws/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/reinitialize_global_localization_service.hpp:20:
/Users/ewen/ros_ws/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp:38:34: error: use 'template' keyword to treat 'get' as a dependent template name
    node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
                                 ^
                                 template
[...]

PR #2375, opened 2021-05-29
Moves changes from PR #2083 to foxy

### LINKED ISSUE #2368: Moving fixes from PR #2083 into foxy-devel
## Bug report

**Required Info:**

- Operating System:
  - macOS 11
- ROS2 Version:
  - Foxy, from source
- Version or commit hash:
  - latest on `foxy-devel`
- DDS implementation:
  - Fast-DDS

Building on macOS was fixed by PR #2083 -- but those changes haven't gotten added to the foxy-devel branch, which is the branch people building for Foxy are expected to build off of. Is there a reason those changes haven't been cherry-picked over? I'd be happy to do it otherwise.

nkalupahana:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Closes #2368 |
| Primary OS tested on | macOS |

---

- Foxy branch now builds on macOS, based on fixes from PR #2083! Proof of build on CI: https://github.com/nkalupahana/ros2-foxy-macos/runs/2698988683?check_suite_focus=true

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -229,7 +229,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Find look ahead distance and point on path and publish
   const double lookahead_dist = getLookAheadDistance(speed);
   auto carrot_pose = getLookAheadPoint(lookahead_dist, transformed_plan);
-  carrot_pub_->publish(std::move(createCarrotMsg(carrot_pose)));
+  carrot_pub_->publish(createCarrotMsg(carrot_pose));
 
   double linear_vel, angular_vel;
 
--- a/nav2_amcl/src/amcl_node.cpp
+++ b/nav2_amcl/src/amcl_node.cpp
@@ -401,7 +401,7 @@ AmclNode::checkLaserReceived()
     RCLCPP_WARN(
       get_logger(), "Laser scan has not been received"
       " (and thus no pose updates have been published)."
-      " Verify that data is being published on the %s topic.", scan_topic_);
+      " Verify that data is being published on the %s topic.", scan_topic_.c_str());
     return;
   }
 
@@ -410,7 +410,7 @@ AmclNode::checkLaserReceived()
     RCLCPP_WARN(
       get_logger(), "No laser scan received (and thus no pose updates have been published) for %f"
       " seconds.  Verify that data is being published on the %s topic.",
-      d.nanoseconds() * 1e-9, scan_topic_);
+      d.nanoseconds() * 1e-9, scan_topic_.c_str());
   }
 }
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -36,11 +36,11 @@ public:
     const BT::NodeConfiguration & conf)
   : BT::ActionNodeBase(xml_tag_name, conf), action_name_(action_name)
   {
-    node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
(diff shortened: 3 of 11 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -229,7 +229,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Find look ahead distance and point on path and publish
   const double lookahead_dist = getLookAheadDistance(speed);
   auto carrot_pose = getLookAheadPoint(lookahead_dist, transformed_plan);
-  carrot_pub_->publish(std::move(createCarrotMsg(carrot_pose)));
```
SteveMacenski, line 232: Change this back - we need to be publishing this as a unique pointer.

SteveMacenski (changes requested):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
nkalupahana: Unfortunately, changing it back will lead to a build error:
```
/Users/runner/ws/src/navigation2/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp:232:24: error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]
  carrot_pub_->publish(std::move(createCarrotMsg(carrot_pose)));
                       ^
/Users/runner/ws/src/navigation2/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp:232:24: note: remove std::move call here
  carrot_pub_->publish(std::move(createCarrotMsg(carrot_pose)));
                       ^~~~~~~~~~                            ~
1 error generated.
```
Full CI Run: https://github.com/nkalupahana/ros2-foxy-macos/runs/2689457060?check_suite_focus=true

SteveMacenski: We publish almost everything using unique pointers, why is this particular publication a problem?

nkalupahana: This is still publishing with unique pointers -- all I've done here is remove `std::move`, as required by the compiler. This is because in C++17, when compiling with clang, the compiler requires that given certain circumstances (which are met here), the code is optimized to construct the object directly in the caller's stack frame, without ever constructing copies. std::move gets in the way of this.

Here's more information on this from a source that can explain it better than I can; see the `-Wpessimizing-move` section: https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c
GCC should also make this optimization since the C++ standard is set to C++17.

Let me know if this explains the change! TL;DR, removing `std::move` allows the compiler to make a more powerful optimization than `std::move`, and this optimization is required by clang on C++17.

SteveMacenski: OK, if you can submit these changes to `main` and they also build, I'd be OK merging this. I don't want to merge this into only `foxy`  though since it wouldn't apply to any future distributions (and our released distro CI isn't set up for non-member changes)

nkalupahana: Got it. Since this is the only change that isn't already in `main`, should I make a PR to main with just this change, then?

SteveMacenski: Yeah (you could also do Galactic if you like)

nkalupahana: Sounds good, I've done `main` and I'll do galactic in a minute.

nkalupahana: #2395 #2396

Issue #2376, opened 2021-05-29
BT Action Nodes crash `bt_navigator` node upon action server call timeout


paucarre:
## Bug report

- Operating System: Ubuntu 20.04
- ROS2 Version: foxy
- Version or commit hash: 2e18dd1cae31807f68746eb1a2c702526a34f4cc
- DDS implementation: Fast-RTPS

#### Steps to reproduce issue
- Use any Behavior Tree Action Node ( BT::ActionNodeBase )
- Call an action with lower-than-needed timeout.

#### Expected behavior
- The action node itself should return a  `BT::NodeStatus::FAILURE` whenever a problem occurs.
- Whenever there is an error, clear and detailed error logs should be provided.
- The ros2 node holding the BT engine ( like `bt_navigator` ) should not die
- The BT engine should keep running the tree (for instance, running a recovery node )

#### Actual behavior
- The whole ros2 node holding the BT engine crashes ( `bt_navigator` ), thus no recovery node is execute but rather the whole BT execution is suddenly halted
- No meaningful logs provided ( only "send_goal failed" )
- `ros2` daemon needs to restart to see the `bt_navigator` node again ( after the `bt_navigator` node is re-launched after it crashes ).
 

## Feature request
- See PR https://github.com/ros-planning/navigation2/pull/2377

SteveMacenski:
I think a better solution is to handle the exception, not remove support for exceptions. Other Bt nodes will throw so we should instead make sure that there isn't crashing when exceptions are thrown in the behavior tree.

paucarre:
Thanks for feedback @SteveMacenski 

I totally agree with you that preventing exceptions from causing major damage is something necessary. 

I'd say though that throwing exceptions, *specifically for BTNodes* , is to me not the most suitable mechanism to notify errors.
These are the reasons why:
[...]

PR #2377, opened 2021-05-29
BtActionNode does not crash bt_navigator upon timing out when calling action server

paucarre:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2376 |
| Robotic platform tested on | Internal mobile robot ( actual physical robot ) |

---

* `BtActionNode`s do not crash `BT engine` nor `ros2 nodes` holding them, as when there is a problem when calling an `action server` a `BT::NodeStatus::FAILURE` is returned instead of throwing an exception.
* When a `BtActionNode` fails, it is able to run a recovery mechanism, integral in `BT` framework.
* `ros2 daemon` is no longer corrupted being unable to discover `bt_navigator` `ros2 node` after relaunch
* Logs are more clear and the logging system is able to display logs from `action servers` 

* No documentation necessary

---

 * Use `return BT::NodeStatus::FAILURE` + `log error` instead of throwing an exception for `BT nodes` 
  
`BT Nodes` can fail, and are meant to allow this, using recovery actions as mechanism resolve these situations. For that, the `BT Engine` does need the `BT Node` to return a `BT::NodeStatus::FAILURE`. Throwing exceptions, will break all the recovery mechanisms the BT provide as the `ros2 node` and `bt engine` will be compromised. See also next point below for details.
 
 * Initialize BT Nodes inputs on `tick()`

I have seen BT Nodes that when compulsory inputs are not provided, the node [crashes in the constructor at runtime](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/condition/transform_available_condition.cpp#L42). This will not only crash the whole `BT engine` and `ros2 node`, but **there is actually nothing wrong at all by a `BT Node` not having values for compulsory inputs during class constructor**. These inputs can be published on the `blackboard` by other `bt nodes` ( for instance the `SetBlackboard` node). This will take place **after** the node construction.

It is therefore not meaningful to get values of inputs in the constructor as it's in the `tick()` method when they shall be loaded. `BT Nodes` do not necessarily have static inputs, they are dynamic by design by means of using the `blackboard`.

Whether a `BT Node` has compulsory inputs is a check that needs to be done in the `tick()` method. And then there is the question as to what to do when these are not provided. From my POV a nice error log together with returning a `BT::NodeStatus::FAILURE` is the right way to do it (not crashing the whole `bt_navigator` using exceptions).

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/package.xml
+++ b/nav2_amcl/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>nav2_amcl</name>
-  <version>0.4.6</version>
+  <version>0.4.7</version>
   <description>
     <p>
       amcl is a probabilistic localization system for a robot moving in
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -213,7 +213,7 @@ public:
       {
         RCLCPP_ERROR(
           node_->get_logger(),
-          "Failed to cancel action server for %s", action_name_.c_str());
+          "Failed to cancel action server \"%s\"", action_name_.c_str());
       }
     }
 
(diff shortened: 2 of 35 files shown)


SteveMacenski:
This fails to build on CI and the developer has not been responsive on the ticket comments https://github.com/ros-planning/navigation2/issues/2376 to finding a solution. I am closing this PR but definitely open to continuing the conversation and reopening it if there is a response.

Issue #2327, opened 2021-04-28
[Reg Pure Pursuit] Support reverse direction


SteveMacenski:
The SMAC planner supports both the Reeds-Shepp & Dubin motion models for ackermann / legged robots.

The Dubin model can only move forward which RPP encapsulates perfectly.  However Reeds-Shepp can allow the robot to reverse at a cusp (and point the opposite direction). So at these points we should move the carrot from the front of the robot to the reverse of the robot in those situations and invert the velocity so we can support reversing as well. 

This figure shows a simple geometric view of the reverse cusping: https://github.com/leggedrobotics/se2_navigation/blob/master/pure_pursuit_core/doc/path_conventions.pdf

As well on the path smoother, do in sections

Timple:
Our (similar) ROS1 planner can travel any path in both directions by indeed moving the carrot to the other side of the robot.

However this doesn't mean it can do reeds-shepp since that would change the target velocity half-way a path. We currently solve this by simply dividing into sections and sending them sequentially to the controller. Is this what you're suggesting here?

SteveMacenski:
Precisely, I’d welcome the contribution if possible! 

Glad great minds think alike with what you built internally 😉

Timple:
Where in the typical **global planner -> local planner** flow would this splitting best occur? It feels like an intermediate step, like **global planner -> plan parser/splitter -> local planner**.
It could also be completely up to the local planner, but this would require the planner to know which parts are supposed to go backwards.

(Our reeds-shepp capable robot uses a global planner which does the splitting into parts already, so we didn't solve this architectural piece yet).

SteveMacenski:
Because the RPP controller allows the reversing, I think this logic belongs there. Other algorithms / controllers might not need / want that separation to handle it. I say put it in RPP directly and if we find a reason to use it in multiple places, we can abstract it out into a library in `nav2_util` for multiple to use.

Timple:
>  if marching forwards in path points, i, and direction is pointing backwards but the last point was pointing forwards, then split (and vise versa). 

This was exactly the code I was trying to prevent. But then again, doesn't seem too bad. Orientation is correct in our paths so no problems there.
We do have (like lots of robots/applications?) different requirements for forward and backward speeds. So an interface like [this](https://github.com/ros-planning/navigation2/blob/22407b7522b4b07cc65a0c3ab865aa4dc7c34282/nav2_msgs/action/BackUp.action#L3) would be welcome, but then for both directions.

[...]

PR #2378, opened 2021-06-01
Adding the capability for the RRP to drive the robot backwards

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2327 |
| Robotic platform tested on | gazebo simulation of [Neobotix MPO-700](https://github.com/neobotix/neo_simulation2) |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -245,25 +245,20 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     curvature = 2.0 * carrot_pose.pose.position.y / carrot_dist2;
   }
 
-  linear_vel = desired_linear_vel_;
+  // Determining the direction of the robot heading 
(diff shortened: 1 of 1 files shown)


SteveMacenski (changes requested):
Please read the discussion in https://github.com/ros-planning/navigation2/issues/2327, this would not be robust. You should be looking at the orientation vectors of the path to determine cusping to change directions. I'm also not sure how you solved the issue of it reversing too early if the carrot is a future-pose relative to the base location being where the cusp is. 

I'm also not sure why you remove rotation to path and reorg the apply constraints function. 

Can you show a gif or video of this working?

SteveMacenski:
This should probably be re-worked -- maybe readded as a draft PR or resubmit a new PR at a later date?

padhupradheep:
Sure, I will look into the points and do the re-work! I think a new PR will be a good idea!

padhupradheep:
Well, I did not take the cusp into the consideration, rather just looked into the position of the carrot. After seeing this [image](https://github.com/leggedrobotics/se2_navigation/blob/master/pure_pursuit_core/doc/path_conventions.pdf), as you said I think my solution will not be robust.

Issue #2042, opened 2020-10-14
Assess parallelism opportunities in Nav2


SteveMacenski:
From GPU, OpenMP, TBB, etc 

- DWB has these `N` critics over `M` trajectories structure that could be parallelized at 2 levels
- AMCL / nav2 localization framework is built on a particle filter, which are independently updated
- Costmap layers can be updated separately and then combined (and sensor processing layers can process multiple measurement readings at once with raycasting/marking)
-  Anything in planning? Not sure because search based. Maybe a sampling planner could better utilize
- Bt navigator recoveries to check for planning/control validity during recovery execution to preempt
- Nav2 dynamic tracking / detection / layer processing
- Any new algorithms 
- transformations of TF for things like pointclouds to a new frame. (e.g. laser projections too)
- Voxel grid: To support the above, we have a voxel grid library which could probably have some internal optimizations with parallel computing to march
- SmacPlanner* (soon): We are collision checking in increments of a motion primitive for the state lattice planner to make sure valid expansions. E.g. we have long motion primitives that cannot be summarized by collision checking the start and end, need to do a few in between and collision checking isn't cheap.
- RPP/Recoveries: We forward project in dt increments into the future given a velocity command in the RPP controller and backup / spin recoveries to collision check. This could be parallelized to collision check them all at once.

simutisernestas:
I've taken a look into costmap layers updates out of curiosity. I'm not quite sure if that's a reasonable assessment, so here is a short description of what I did exactly. As I understand the heaviest stuff in layers is done here ObstacleLayer::updateBounds and my idea were to exploit OpenMPs "parallel for" directive. 

So I've modified the first loop in the LayeredCostmap::updateMap function to support parallel updates. Also, I've stacked more layers onto costmap to make the difference more clear. Changes are available here: https://github.com/simutisernestas/navigation2/commit/f0630c672d2db0c38ad852ac09d26b9216ddf718.

I've observed that map average update time (tb3 simulation, Intel® Core™ i7-3770 CPU @ 3.40GHz×8):
- without "parallel for" ranges from 19ms - 24ms 
- with "parallel for" ranges from 12ms - 18ms

If considering the 5Hz update rate 6ms difference adds up to a 30ms gain per second. I suppose that processing the real-world data could possibly increase layer update time and the effect seen here would be much more visible (for example longer lidar range or bigger pointcloud).

Would be cool to hear a second opinion. :)

SteveMacenski:
[...]

PR #2393, opened 2021-06-07
parallelize costmap bound updates

simutisernestas:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2042 |
| Robotic platform tested on | Gazebo simulation of turtlebot |

---

## Description of contribution

- fun little parallel optimization of costmap bounds update
- i've described this previously in #2042, but this time i've used default config provided in `nav2_bringup`. When the number of layers gets smaller (or layers are very light) improvement slowly vanishes, but still even with 2-3 layers and no filters there is 5-10% improvement overall in  `updateMap` measured here: https://github.com/ros-planning/navigation2/blob/b63b11f8a638faf389d6b99911d95899abe1580b/nav2_costmap_2d/src/costmap_2d_ros.cpp#L411-L415

---

## Future work that may be required

- probably a parallel version of `updateCosts` function call would be more interesting and useful. However, that's a bit more tricky to do safely on that shared member variable `combined_costmap_` here: https://github.com/ros-planning/navigation2/blob/b63b11f8a638faf389d6b99911d95899abe1580b/nav2_costmap_2d/src/layered_costmap.cpp#L208
- any thoughts on how you would go about implementing this one? one obvious way would be to make a copy for each thread, update costs on the copy, and later join them but that seems suboptimal because the costmap copy can become expensive  

#### For Maintainers: 


nav2_costmap_2d/CMakeLists.txt
```diff
@@ -27,6 +27,13 @@ find_package(angles REQUIRED)
 remove_definitions(-DDISABLE_LIBUSB-1.0)
 find_package(Eigen3 REQUIRED)
 
+find_package(OpenMP)
```
SteveMacenski, line 30: inflation_tests.missing_result
whitespace/extra (CMakeLists.txt:31)
whitespace/extra (CMakeLists.txt:32)
whitespace/extra (CMakeLists.txt:33)
whitespace/extra (CMakeLists.txt:34)

You have some linting issues according to CI using the ament cmake linter (there's some other test failures that aren't your fault, I just haven't had time to fix the other linting issues yet).

SteveMacenski:
Conviniently, about 50% of the costmap plugins do the majority of their work in the update bounds section (since you need to know what to extend the bounds by, you have to process information there). So the `updateCosts` function is primarily just copying over the data, the updateBounds probably represents a non-trivial amount of the available optimization. 

I think it would actually be a _better_ use of time to migrate more plugins to do their main "work" in the `updateBounds` function to use this optimization than to try to parallelize `updateCosts` since that is majority copying over which can't be easily parallelized.

nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -145,40 +145,46 @@ void LayeredCostmap::updateMap(double robot_x, double robot_y, double robot_yaw)
   minx_ = miny_ = std::numeric_limits<double>::max();
   maxx_ = maxy_ = std::numeric_limits<double>::lowest();
 
-  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin();
-    plugin != plugins_.end(); ++plugin)
+  #pragma omp parallel reduction(min:minx_) reduction(min: miny_) \
+  reduction(max: maxx_) reduction(max: maxy_) num_threads(2)
```
SteveMacenski, line 149: Is there a way to have it autoscale the number of threads by compute available resources or does it need to be specified?

nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -145,40 +145,46 @@ void LayeredCostmap::updateMap(double robot_x, double robot_y, double robot_yaw)
...
-    (*plugin)->updateBounds(robot_x, robot_y, robot_yaw, &minx_, &miny_, &maxx_, &maxy_);
-    if (minx_ > prev_minx || miny_ > prev_miny || maxx_ < prev_maxx || maxy_ < prev_maxy) {
-      RCLCPP_WARN(
-        rclcpp::get_logger(
-          "nav2_costmap_2d"), "Illegal bounds change, was [tl: (%f, %f), br: (%f, %f)], but "
-        "is now [tl: (%f, %f), br: (%f, %f)]. The offending layer is %s",
-        prev_minx, prev_miny, prev_maxx, prev_maxy,
-        minx_, miny_, maxx_, maxy_,
-        (*plugin)->getName().c_str());
+  #pragma omp for schedule(static, 1)
```
SteveMacenski, line 151: From my understanding, this scheduling makes it so that each thread takes 1 iteration at a time statically. Why specify this and not have it use `auto` or the other options? Genuinely not sure, this might be the right thing to do but asking

nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -145,40 +145,46 @@ void LayeredCostmap::updateMap(double robot_x, double robot_y, double robot_yaw)
...
-    (*filter)->updateBounds(robot_x, robot_y, robot_yaw, &minx_, &miny_, &maxx_, &maxy_);
-    if (minx_ > prev_minx || miny_ > prev_miny || maxx_ < prev_maxx || maxy_ < prev_maxy) {
-      RCLCPP_WARN(
-        rclcpp::get_logger(
-          "nav2_costmap_2d"), "Illegal bounds change, was [tl: (%f, %f), br: (%f, %f)], but "
-        "is now [tl: (%f, %f), br: (%f, %f)]. The offending filter is %s",
-        prev_minx, prev_miny, prev_maxx, prev_maxy,
-        minx_, miny_, maxx_, maxy_,
-        (*filter)->getName().c_str());
+  #pragma omp for nowait schedule(static, 1)
```
SteveMacenski, line 170: Why nowait here? Wouldn't the closed bracket ending the openMP section be right after this to not have any thing to do after the first thread finishes anyway?

SteveMacenski (changes requested):

nav2_costmap_2d/src/layered_costmap.cpp
simutisernestas: The margin of improvement (at least with default number layers) is soo small that launching extra threads becomes more costly than doing it in a serial way. A more reasonable approach would be to check how many (plugins + filters) there are present at runtime and set the number by that.

nav2_costmap_2d/src/layered_costmap.cpp
simutisernestas: Similar to the previous point, if you have a small number of iteration and `auto` sets a schedule for one thread to take all iterations at once this wouldn't make any sense. I'm just cautious about `auto` because I don't know what the compiler will do here in that case. Therefore one plugin or filter update allocated to one thread seems most appropriate because the number is usually less than 5

nav2_costmap_2d/src/layered_costmap.cpp
SteveMacenski: Do the openMP preprocessor statements allow you to do that? I would think it would be set at compile time. Though if you had many plugins (6) on a normal machine with only 4 cores, it would necessarily help. I think at most it should be the number of cores on the computer (which to my understanding of OpenMP is actually the default if not specified) but less if less plugins or some of them are very light weight.

nav2_costmap_2d/src/layered_costmap.cpp
simutisernestas: OpenMP implicitly adds a `barrier` construct after each section in this case `for`. Thus if one thread is finished with a plugin update and another one is still doing its thing the first one will wait for it.

nav2_costmap_2d/src/layered_costmap.cpp
simutisernestas: Wasn't sure about the runtime decision on a number of threads, so i've tried this demo and it actually gives a number of threads requested at runtime.

```
  auto a = random_int();
  
  std::cout << a << '\n';

#pragma omp parallel num_threads(a) default(none)
  {
    int b = omp_get_num_threads();
	printf("%d", b);
  }
```
So let's set that number to be something along the lines:

```
int n_threads = std::min(number_of_cores, (number_of_plugins + number_of_filters))

File: nav2_costmap_2d/src/layered_costmap.cpp (hunk 4 above)
Comment (SteveMacenski) [reply nav2_costmap_2d/src/layered_costmap.cpp:170]: Why does the other one not have `nowait`?

File: nav2_costmap_2d/src/layered_costmap.cpp (hunk 2 above)
Comment (SteveMacenski) [reply nav2_costmap_2d/src/layered_costmap.cpp:149]: Aren't the plugins / filters in different `for` loops? I don't think we'd need to sum them unless the parallelization can continue to the send batch of the filters

File: nav2_costmap_2d/src/layered_costmap.cpp (hunk 3 above)
Comment (SteveMacenski) [reply nav2_costmap_2d/src/layered_costmap.cpp:151]: did you see a demonstrable change for your testing with and without this?

Comment (simutisernestas) [reply nav2_costmap_2d/src/layered_costmap.cpp:151]: i'll get back with some quantitative results on this one, haven't tested with `auto` actually

File: nav2_costmap_2d/src/layered_costmap.cpp (hunk 4 above)
Comment (simutisernestas) [reply nav2_costmap_2d/src/layered_costmap.cpp:170]: At the beginning of the parallel block, all threads start their job without any hesitation. Code snippet below might be clearer:
```
    #pragma omp parallel
    {
        // all thread start from here, no stopping

        for (size_t i = 0; i < 100; i++)
        {
            /* code */
        }
        
        #pragma omp barier // implicitly added after each for block
        // all threads wait until the first loop is finished

        for (size_t i = 0; i < 100; i++)
        {
            /* code */
        }

        #pragma omp barier // implicitly added after each for block
    }

nav2_costmap_2d/src/layered_costmap.cpp
simutisernestas: Yeah, we could have a separate number for plugins & filters

nav2_costmap_2d/src/layered_costmap.cpp
SteveMacenski: (or even just lacking of schedule specification, which I assume is auto.  I think it defaults to the number of cores)

nav2_costmap_2d/src/layered_costmap.cpp
SteveMacenski: I don't understand. Wouldn't you want it to potentially continue on in this case? Wouldn't that be the value of `nowait`? The second `updateBounds` loop has `nowait` but the first doesn't, I'd think the first is where it would have more value. I don't think I understand why its not in both.

simutisernestas: Sorry, my mistake. I thought that `nowait` must be added on the following block in order to let loose the previous block. I'll move `nowait` above.

nav2_costmap_2d/src/layered_costmap.cpp
simutisernestas: Sooo, first things first i've tried switching the schedule to `auto` and it is performing somewhat the same as with static chunk of size one. avg - average in seconds, first-line - key testing conditions:
```
# schedule(static, 1) num_threads(2)
# avg: 0.002816783901452364 from 12187 samples

# schedule(auto) num_threads(2)
# avg: 0.0027575583444915284 from 6425 samples
```
The difference is barely visible, but I would trust the compiler more than myself in this case therefore all following experiments were done with auto.

In order to see the performance with at least one filter, i've added keep out zones in the setup so map update is slightly longer compared to previous ones. This time experimenting with thread number:
```
# schedule(auto) num_threads(2)
# avg: 0.0031912964961487793 from 5323 samples

# int n_cores = omp_get_num_procs();
# n_threads = std::min(n_cores, (int)std::max(plugins_.size(), filters_.size()));
# schedule(auto) num_threads(n_threads)
# avg: 0.0035948433114727134 from 6084 samples

# int n_cores = omp_get_num_procs();
# int n_threads = std::min(n_cores, (int)std::max(plugins_.size(), filters_.size()));
# int n_plugin_threads = std::min((int)plugins_.size(), n_cores);
# int n_filter_threads = std::min((int)filters_.size(), n_cores);
# avg: 0.0032808737608167084 from 6171 samples
```
First one - same setup as previously + keep-out zones, the second one - single `parallel` block to process (plugins+filter) in the same block with `n_threads` as specified, the third one - two blocks for each `for` loop selecting `n_threads` as shown (two parallel blocks because `omp for` does not allow for `num_threads` option). From this, we see that the second one is lagging behind and the remaining two have quite similar execution duration.

SteveMacenski: Got it - so it sounds like we should leave the scheduling to OpenMP, is it worth the core computations in your opinion? ~90 microseconds difference between them is not really worth optimizing for. 

---

From these results, with the final decision on both of those, what's the compute time speed up as seen by the end user (and what's your compute platform, what layers do you have enabled, etc)

simutisernestas: Yeah, I think we can leave scheduling for OpenMP. 

I would go with the third one since the difference is not noticeable and it seems more flexible. I've pushed code with that version. Probably those variables in front of the loop could be made static member variables.

I'll update you on the speed-up in the near future.

simutisernestas: I've got back to the speed-up evaluation. Funny enough, without `DEBUG` log level parallel version is actually underperforming compared to the plain loop... Log level was influencing my measurements previously somehow.

```
plain average update time: 0.001102279375405933
parallel average update time: 0.0010157955731191883
```
These are averaged updated from both local and global cost maps. With this config: https://github.com/simutisernestas/navigation2/blob/parallel_config/nav2_bringup/bringup/params/nav2_params.yaml. I wonder if that would be the case with heavier cost map layers or bigger map.

SteveMacenski: 1.1 ms, why so fast? Oh, is this in the TB3 sandbox world? Yeah, it might be worth trying something larger. This shows an overview of getting nav2 running in the AWS warehouse world https://github.com/ros-planning/navigation2/tree/main/nav2_simple_commander.

simutisernestas: Strangely enough, map updates are even faster on AWS warehouse world demo. I've pulled the latest changes from the repo the code must have been upgraded since the last time I've benchmarked this. Both parallel and serial versions do the job in about the same amount of time. With parallel slightly better, but it's insignificant I would say. I'm testing on `Intel® Core™ i7-3770 CPU @ 3.40GHz × 8` with the same config as mentioned previously.
```
serial avg update time: 0.7053497955298016 ms from 2416 samples
parallel avg update time: 0.6474942507522566 ms from 2991 samples

SteveMacenski: That doesn't surprise me, since you if each thread is doing "more" work, then the overhead of spinning up a new thread is more worth it. It's about 8% it seems, do you know why its lower than what you were seeing before when you reported a ~30% increase in speed? I'm also a little dubious since the updates are so fast. Your other experiment reported 12-20ms in update times (I assume more data) while this is less than 1ms. 1ms seems way too fast to be what you'd see in a practical deployed application so I wonder if in a more realistic setup where things are taking longer if we do see that benefit again.

SteveMacenski:
I'm going to close this for now since we're still trying to characterize the benefits and settings. We're not seeing the kind of improvements we'd like and I don't want to keep holding you hostage to working on this after so many months. Its a good reference starting point for an effort another time, but I think it would be good to get this off the review queue until a more dedicated effort is started another time.

PR #2401, opened 2021-06-09
Cache ccache folder in CI workspace

ruffsl:
Fixes issue with restoring ccache, introduced in https://github.com/ros-planning/navigation2/pull/2343 . This does so by setting the `CCACHE_DIR` to be inside the workspace (i.e. current working directory). This also has the benefit that each workspace now caches it's own ccache directory, and simplifies the number of specific steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,6 @@
 version: 2.1
 
 _commands:
-  common_steps: &common_steps
-    store_ccache_logs: &store_ccache_logs
-      store_artifacts:
-          path: /tmp/ccache.log
   common_commands: &common_commands
     ccache_stats:
       description: "CCache Stats"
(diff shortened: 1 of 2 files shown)


ruffsl:
Example of existing issue in Restore Cache ccache step:

```
Found a cache from build 20042 at ccache-v1-arch1-linux-amd64-6_85-main-<no value>-Aq7BtXYEt7jrMu57mTFIwvjZ0+NimH3EYrtW+noHpow=
Size: 32 B
Cached paths:
  * /tmp/.ccache/build
  * /tmp/.ccache/install
  * /tmp/.ccache/log

Downloading cache archive...
Validating cache...

Unarchiving cache...
```

https://app.circleci.com/pipelines/github/ros-planning/navigation2/5474/workflows/2df6996a-1bbc-4646-ba64-fad1545a0638/jobs/20045/parallel-runs/0/steps/0-105

Issue #2408, opened 2021-06-16
bt_navigator: Multiple action nodes are started with the same names and parameters (pose, poses)


jonatanolofsson:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Galactic binaries
- Version or commit hash:
  - 1.0.6
- DDS implementation:
  - 

#### Steps to reproduce issue
Start nav2 with bt_navigator
```
ros2 node list
```

#### Expected behavior
All nodes have separate names

#### Actual behavior
Two nodes named /bt_navigator_rclcpp_node are started

SteveMacenski:
Please add much more detail to this ticket, I don't understand the issue at all you're reporting and you're using ambiguous language

Issue #2406, opened 2021-06-14
ROS Galactic groot monitoring


kribe48:
## Bug report
In ROS galactic, an additional action server has been added (navigate through poses). However, both servers use the same groot_zmq_publisher/server_port parameter values. This makes it impossible for me to monitor the behavior trees in groot. Am I using it wrong, or is it a bug that needs to be updated?

- Operating System:
ubuntu 20.04
- ROS2 Version:
Galactic
- Version or commit hash:
From source

#### Expected behavior
Able to monitor both navigate through poses and navigate to pose in groot

#### Actual behavior
Not possible to monitor either of them

SteveMacenski:
This is a known issue with Groot, that it only allows a single instance per process. I'd welcome a PR to help resolve this, but as I alluded to [this ticket](https://github.com/ros-planning/navigation2/issues/2386), if this isn't resolved relatively soon, my tentative plan is to strip support for groot out of Nav2 since it seems to be causing issues with critically needed features. 

Destroying and replacing Groot objects every time the navigator changes turns out to not be as straight forward as I thought it would be, so you'd be correct that right now it only works with NavigateToPose actions, not NavigateThroughPoses.

SteveMacenski:
I believe #2409 would resolve these concerns @kribe48. Do you agree?

kribe48:
Yes, I agree!

PR #2409, opened 2021-06-16
Set groot monitoring parameters individually for bt_navigator's pose …

jonatanolofsson:
Set groot monitoring parameters individually for bt_navigator's pose and poses navigators

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #2406, #2408 |
| Primary OS tested on | Ubuntu 20.04 (docker container) |

---

 * Groot monitoring now set up for poses and pose bt_navigator action nodes separately
 * Interface added to allow setting of relevant parameters

* Added six new parameters, so need to add that to default configs and documentation page

---

 * Regression tests could be added

#### For Maintainers: 


SteveMacenski:
All PRs must also be targeted to main and they will be cherry picked back down to released distributions if ABI/API compatible. This helps us make sure that any changes are reflected in all new distributions as well. This PR changes parameter names / defaults so this would not be possible to backport to Galactic. It could though be merged into `main` and you'd get this capability available to you on `rolling` or master builds (maybe also galactic, I haven't checked their compatibility recently)

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -72,6 +72,12 @@ BtNavigator::BtNavigator()
   declare_parameter("global_frame", std::string("map"));
   declare_parameter("robot_base_frame", std::string("base_link"));
   declare_parameter("odom_topic", std::string("odom"));
+  declare_parameter("pose_enable_groot_monitoring", true);
```
SteveMacenski, line 75: If we have these here, we should outright remove the parameter gets / declares in the Bt action node 
- https://github.com/ros-planning/navigation2/blob/8947c9fdca621624fddc49ff40af9d3f29693e81/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp#L62-L70
- https://github.com/ros-planning/navigation2/blob/8947c9fdca621624fddc49ff40af9d3f29693e81/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp#L101-L103

And then we only have the API to set the ports from the application

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -72,6 +72,12 @@ BtNavigator::BtNavigator()
...
+  declare_parameter("pose_groot_zmq_publisher_port", 1670);
+  declare_parameter("pose_groot_zmq_server_port", 1671);
+  declare_parameter("poses_enable_groot_monitoring", true);
```
SteveMacenski, line 78: I think the `enable_groot_monitoring` should remain and apply to both `pose` and `poses`. I think that should be an "all or nothing" parameter.

SteveMacenski (commented):
Overall looks good, just some clean up so that we don't have some parallel APIs people are confused by

nav2_bt_navigator/src/bt_navigator.cpp
jonatanolofsson: Thanks for the comments! I updated the PR with your suggestions. However, I seem to have problem connecting to one of the groot servers, I'll see if I can have a look at it either over the weekend or early next week.

SteveMacenski:
Hey! It looks like your force push here also caused some havoc, see above

SteveMacenski:
Note @jonatanolofsson https://github.com/ros-planning/navigation2/pull/2417 (the default for enable should be false now) please make sure when you resolve your conflicts you also make sure this is false.

jonatanolofsson:
I fixed the target branch and the defaulting to false. 

However, I still can't seem to connect to the poses tree. All printouts I have added (locally) indicate that the monitoring is instantiated correctly and in the same was as the pose monitor, so I don't believe the error is in this code.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
```diff
@@ -86,6 +86,14 @@ class BtActionServer
    */
```
SteveMacenski, line 86: The builds are failing, something isn't linking correctly from these changes

SteveMacenski: Oh, this PR doesn't have the `.cpp` implementation of `setGrootMonitoring` anymore, odd

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -72,6 +72,11 @@ BtNavigator::BtNavigator()
   declare_parameter("global_frame", std::string("map"));
   declare_parameter("robot_base_frame", std::string("base_link"));
   declare_parameter("odom_topic", std::string("odom"));
+  declare_parameter("enable_groot_monitoring", false);
```
SteveMacenski, line 75: Maybe worth dropping the `nt_` prefix? I'm not sure what it adds or that people will understand what that means.

Maybe instead do `pose_bt_...` and `poses_bt_...`?

SteveMacenski (changes requested):
Keep in mind that in this situation, even though you've set different ports, only 1 Groot Monitor object is active at a time. This has nothing to do with ports anymore and just how ZMQ works (it crashes if you try to make 2 instance in the same process). So at instantiation, we make the Pose BT monitored, so if you launch groot, you'd see that.

If you call on the Nav Through Poses, I think you'd probably have to fully exist the Groot system, and then relaunch it with the new ports since its now a completely different object that's been switched to. You may also want to verify that we do indeed switch correctly.

jonatanolofsson:
> Keep in mind that in this situation, even though you've set different ports, only 1 Groot Monitor object is active at a time. This has nothing to do with ports anymore and just how ZMQ works (it crashes if you try to make 2 instance in the same process). So at instantiation, we make the Pose BT monitored, so if you launch groot, you'd see that.

@SteveMacenski Is this why I could only connect to the "pose" tree? If it's impossible to have an active monitor on both simultaneously, then perhaps we should only start one? Where is it that you "make the Pose BT monitored"?

Sorry for the late reply, last week was busy!

SteveMacenski:
I had asked you to test if you started the Nav through poses action and then launched groot to see if you see it then, since only 1 instance is active at any give time, and they have different ports. If you can see it then, then things are working properly, its just that groot needs a restart when a new object / tree are formed and that's not really an "us" problem, that's a problem you'd need  to field with them.

Frankly, I'm at half a mind to remove  groot entirely. Its caused a bunch of headaches and if you're reporting that its not working in a way that makes it useful to you, I'm not sure that the usefulness of it is in general.

SteveMacenski:
@jonatanolofsson any update here? Did this make things work eventually? Else, I'm planning on removing groot support before the next release

albuckley:
A lot of people in the ROS community are not going to be very vocal. I would like to make the claim that dropping support for Groot would be gutting and it shouldn't be a nice-to-have. Developing behaviour with a viewer isn't strictly critical but it is a night and day difference without one. With the other behaviour frameworks seemingly orphaned, it leaves us with Groot as the only option that can potentially work.

I would think there are many people that would be grateful to see this integrated.

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/2627

PR #2410, opened 2021-06-16
Allow separate naming of client action server nodes in bt navigator

### LINKED ISSUE #2408: bt_navigator: Multiple action nodes are started with the same names and parameters (pose, poses)

jonatanolofsson:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2408  |
| Primary OS tested on | Ubuntu 20.04 (docker) |

---

 * New ability to override the default name of action node name
 * Fixes #2408 

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -85,7 +85,7 @@ bool BtActionServer<ActionT>::on_configure()
   // use suffix '_rclcpp_node' to keep parameter file consistency #1773
   auto options = rclcpp::NodeOptions().arguments(
     {"--ros-args",
-      "-r", std::string("__node:=") + node->get_name() + "_rclcpp_node",
+      "-r", std::string("__node:=") + std::string(node->get_name()) + action_name_ + "_rclcpp_node",
       "--"});
   // Support for handling the topic-based goal pose from rviz
   client_node_ = std::make_shared<rclcpp::Node>("_", options);


SteveMacenski:
Please be very explicit about what issue you're facing that this is necessary. Also keep in mind it is our intention before the next distribution release to entirely remove client nodes via https://github.com/ros-planning/navigation2/issues/816. I'm not sure exactly what the tangible issue is that this is trying to resolve.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
```diff
@@ -55,6 +55,33 @@ class BtActionServer
     OnPreemptCallback on_preempt_callback,
     OnCompletionCallback on_completion_callback);
 
+  /**
```
SteveMacenski, line 58: After reading this, I think I understand the intention in the PR. You're seeing that we have multiple BT action servers in the 2 navigators and they share a name so you want to have them be separated. Got it. 

Do you care the names they receive? The client node is sent to the blackboard for use in the different BT node options, and they wouldn't be namespaced under the node at all. As such, I think it would be safe to say "no" we don't care about the specific string name of the node.

If that's the case, a much simpler solution to this issue is to do what TF does https://github.com/ros2/geometry2/blob/ros2/tf2_ros/src/transform_listener.cpp#L48 and just apply a UUID or hex after the topic context name to randomize the name to prevent collisions. That would involve no ABI/API breaking changes and we could safely merge that into both Galactic and `main` this week easily .

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
ruffsl: > apply a UUID or hex after the topic context name to randomize the name to prevent collisions.

Although, nondeterministic node names are a pain for security, as there's no good way to fix the runtime DDS security permissions needed before at design time, while still being node specific.

SteveMacenski: What's the plan for TF on that front (or is there)? they could do something like `tf_impl_listener_<name of node given it it that should be unique>`

jonatanolofsson:
Thank you for the comments! I was assuming that it was better to keep node names as before to the greatest extent (I wasn't sure if others used the same bt action interface). I pushed a far simpler solution now, simply adding the action name to the node name.

SteveMacenski:
It looks like you may have force pushed and is has some problems. I'm seeing a ton of conflicts and the PR contains a ton of code that isn't part of this PR. Please make sure to clean that up, I can't review it in the current condition unfortunately.

jonatanolofsson:
No havoc, I had just rebased onto main. I though you had already changed the target branch, but perhaps github reset it. It's fixed now anyway

SteveMacenski (approved):
Thanks for the help!

anaelle-sw:
Hello!

I encountered some issues since this PR is merged. Here is a short description of the problem:

Our action names often come with namespaces, and so with "/" characters. 
Let's say my action is named "/my_api/navigation/go_to_pose". Since this PR, the bt navigator node's name will be remapped to "bt_navigator/my_api/navigation/go_to_pose_rclcpp_node". But a node name cannot feature some "/" characters, so in my case, the node name cannot be remapped. So the node is just not created and I get this error:
```c
[bt_navigator-57] [ERROR] []: Caught exception in callback for transition 10
[bt_navigator-57] [ERROR] []: Original error: failed to parse arguments: Couldn't parse remap rule: '-r __node:=bt_navigator/my_api/navigation/go_to_pose_rclcpp_node'. Error: Expected lexeme type 1, got 19 at index 20, at /tmp/binarydeb/ros-galactic-rcl-3.1.2/src/rcl/lexer_lookahead.c:238, at /tmp/binarydeb/ros-galactic-rcl-3.1.2/src/rcl/arguments.c:371
[bt_navigator-57] [FATAL] [bt_navigator]: Lifecycle node bt_navigator does not have error state implemented
[lifecycle_manager-54] [ERROR] [lifecycle_manager_navigation]: Failed to change state for node: bt_navigator
[lifecycle_manager-54] [ERROR]  [lifecycle_manager_navigation]: Failed to bring up all requested nodes. Aborting bringup.
```

Actually, this PR is incompatible with the use of namespaces in actions' names. Since  "/" characters are not supported in node names, one cannot assume the action name can just be concatenated to the node name. I guess this issue will happen regurlarly, the use of namespaces for action names seems quite common.

For instance, an easy fix would be replacing all "/" characters within the action name by "_" characters before passing it to the node options for the node name remapping.

Thanks for helping to solve this

SteveMacenski:
I'm not sure what you mean, a node can be namespaced, even the API has that as an option and a get-able field https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node.hpp#L115-L127. But maybe that's just a fluke of what we're talking about, your actual suggestion sounds fine to me.

> For instance, an easy fix would be replacing all "/" characters within the action name by "_" characters before passing it to the node options for the node name remapping.

Are you referring to the act that we have the `action_name_` in the node field? That is the only difference and the only thing that if namespaced would introduce `/` into the name. 

Yes, I would merge a PR to change the `/` to `_` for remapping this name!

anaelle-sw:
>  a node can be namespaced

I agree. But a node has a name and a namespace, that are two distinct entities. 
If you want a node to be named `/my_namespace1/my_namespace2/my_node`, this is what you can do:
```c
auto node = std::make_shared<rclcpp::Node>("my_node", "my_namespace1/my_namespace2");  
```
using this [function](https://github.com/ros2/rclcpp/blob/00f2d563bec3ab16f48fe37fd2d1029263510bd5/rclcpp/include/rclcpp/node.hpp#L101-L104).
This works fine, and your node will have `my_namespace` as its namespace, and `my_node` as its name. And if you run `ros2 node list`, you will be able to see this node with its "full name" `/my_namespace1/my_namespace2/my_node`.


But this is what this PR is trying to do:
```c
auto node = std::make_shared<rclcpp::Node>("my_namespace1/my_namespace2/my_node"); 
```
using this [function](https://github.com/ros2/rclcpp/blob/00f2d563bec3ab16f48fe37fd2d1029263510bd5/rclcpp/include/rclcpp/node.hpp#L89-L91).
This doesn't work. If you try to run this node, you will get this error:
```c
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidNodeNameError'
  what():  Invalid node name: node name must not contain characters other than alphanumerics or '_':
  'my_namespace1/my_namespace2/my_node'
                ^
```

> Are you referring to the act that we have the `action_name_` in the node field? That is the only difference and the only thing that if namespaced would introduce / into the name.

Yes exactly, that's actually the problem. If your `action_name_` features some "/", the `action_name_` cannot be concatenated to the node name. 

> Yes, I would merge a PR to change the `/` to` _` for remapping this name!

Nice, thanks, I will open a PR.
But I am not sure this is the best solution. For instance, instead of adding the `action_name_` to the node name, we could put the `action_name_` as the node namespace. 
Let's say your `action_name` is "/my_api/navigation/go_to_pose", would you rather get the client node named either:
1. "bt_navigator_my_api_navigation_go_to_pose_rclcpp_node"

or

2. "/my_api/navigation/go_to_pose/bt_navigator_rclcpp_node" ("/my_api/navigation/go_to_pose/" being the namespace, "bt_navigator_rclcpp_node" being the name)

At first glance, case 2 seems nicer to me.
I guess we can discuss this directly in the PR

Issue #2412, opened 2021-06-17
[Bug] A default copy assignment is not a shortcut to a copy constructor


topin89:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Foxy Fitzroy
- Version or commit hash:
  - from source: output of `git checkout 40f4c38103f4329694c53d9b4864cb6580c380d4` (current head of foxy-devel)
- DDS implementation:
  - shouldn't matter

#### Steps to reproduce issue

```
// look at actual assembly of
Observation & operator=(const Observation &) = default; // observation.hpp:65
```

#### Expected behavior
It's expected to be a shortcut for explicitly defined copy constructor
#### Actual behavior
It just copies every fields of the class. No `new` is called.
#### Additional information
Pull request is on the way. `main` branch has similar issues, but I can't make a pull request for that right now.

[...]

PR #2413, opened 2021-06-17
Explicit assignment operators for Observation, default assignment is a bug here

topin89:
Also move constructor. Default copy assignment is not just a shortcut to copy constructor, sadly

---

## Basic Info

| ------ | ----------- |
| Ticket this addresses   | #2412) |
| Primary OS tested on | Ubuntu under WSL1 |

---

* Fixed bug with copy assignment being just a copy of all class fields, without proper `new` call for `sensor_msgs::msg::PointCloud2 * cloud_`
* Added move assignment and move constructor, because it's they are called inside `std::vector` with less (de)allocations

Teeeny changes:
* Changed description for copy assignment
* Added description for move assignment and move constructor

---

* I guess similar pulls for `crystal` and `main` braches should be done
* Maybe check if any other default defined functions really shouldn't be

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
@@ -60,9 +60,32 @@ public:
   }
 
   /**
-    * @brief  Explicitly define copy assignment operator for Observation as it has a user-declared destructor
-    */
-  Observation & operator=(const Observation &) = default;
+  * @brief  Copy assignment operator
+  * @param obs The observation to copy
+  */
+  Observation & operator=(const Observation & obs){
+      origin_ = obs.origin_;
+      cloud_ = new sensor_msgs::msg::PointCloud2(*(obs.cloud_));
+      obstacle_range_ = obs.obstacle_range_;
+      raytrace_range_ = obs.raytrace_range_;
+      
+      return *this;
+  }
+
+  /**
+  * @brief  Move assignment operator
+  * @param obs The observation to move
+  */
+  Observation & operator=(Observation && obs){
+      origin_ = std::move(obs.origin_);
+      cloud_ = std::move(obs.cloud_);
+      raytrace_range_ = std::move(obs.raytrace_range_);
+      obstacle_range_ = std::move(obs.obstacle_range_);
+      
+      obs.cloud_ = nullptr; // Prevents data release in the destructor
+      
+      return *this;
+  }
 
   /**
    * @brief  Creates an observation from an origin point and a point cloud
@@ -89,6 +112,17 @@ public:
   {
   }
 
+  /**
+   * @brief  Move constructor
+   * @param obs The observation to move
+   */
+  Observation(Observation && obs)
+  : origin_(std::move(obs.origin_)), cloud_(std::move(obs.cloud_)),
+    obstacle_range_(std::move(obs.obstacle_range_)), raytrace_range_(std::move(obs.raytrace_range_))
+  {
+      obs.cloud_ = nullptr; // Prevents data release in the destructor
+  }
+  
   /**
    * @brief  Creates an observation from a point cloud
    * @param cloud The point cloud of the observation
--- a/nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
@@ -63,28 +63,14 @@ public:
   * @brief  Copy assignment operator
   * @param obs The observation to copy
   */
-  Observation & operator=(const Observation & obs){
-      origin_ = obs.origin_;
-      cloud_ = new sensor_msgs::msg::PointCloud2(*(obs.cloud_));
-      obstacle_range_ = obs.obstacle_range_;
-      raytrace_range_ = obs.raytrace_range_;
-      
-      return *this;
-  }
+  Observation & operator=(const Observation & obs)
+  {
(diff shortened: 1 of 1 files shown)


nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
```diff
@@ -89,6 +112,17 @@ class Observation
   {
   }
 
+  /**
+   * @brief  Move constructor
+   * @param obs The observation to move
+   */
+  Observation(Observation && obs)
```
SteveMacenski, line 119: This seems like significant overkill. These are `double`s, using std::move on them is not necessary as far as I can tell

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
```diff
@@ -60,9 +60,32 @@ class Observation
...
+      raytrace_range_ = obs.raytrace_range_;
+      
+      return *this;
+  }
+
+  /**
+  * @brief  Move assignment operator
+  * @param obs The observation to move
+  */
+  Observation & operator=(Observation && obs){
```
SteveMacenski, line 79: As well here, this is does not seem necessary

SteveMacenski (changes requested):
Overall, I don't see the value in this PR. What are you trying to accomplish that was not possible prior, specifically?

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
```diff
@@ -60,9 +60,32 @@ class Observation
   }
 
   /**
-    * @brief  Explicitly define copy assignment operator for Observation as it has a user-declared destructor
-    */
-  Observation & operator=(const Observation &) = default;
+  * @brief  Copy assignment operator
+  * @param obs The observation to copy
+  */
+  Observation & operator=(const Observation & obs){
```
SteveMacenski, line 66: This looks fine to me, but I don't think any of the other changes are required.

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
topin89: Yes, looks like nothing here needs `std::move` for performance. Mostly here for uniformity and to show that everything is moving here.

> Added move assignment and move constructor, because it's they are called inside `std::vector` with less (de)allocations

That's why I think move constructor (as well as assignment) is needed, so there is no unnecessary PointCloud copies/allocations/deallocations every time some vector needs to grow.

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
topin89: > Added move assignment and move constructor, because it's they are called inside `std::vector` with less (de)allocations

 Just like up there. I'm not sure what exactly needed, constructor or assignment, so why not both?

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
topin89: They're indeed not required, but they improve performance a bit.

topin89:
I guess there's an alternative to those verbose constructors/assignments. I didn't try to compile it, but it looks like
`sensor_msgs::msg::PointCloud2 * cloud_;` can be easily replaced with `sensor_msgs::msg::PointCloud2::UniquePtr cloud_;`.

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
SteveMacenski: We want things to be created as needed and not have dead code in the system. If its not being used and isn't needed, then it should not be added. Its more code then we (I) need to maintain.

nav2_costmap_2d/include/nav2_costmap_2d/observation.hpp
SteveMacenski: We want things to be created as needed and not have dead code in the system. If its not being used and isn't needed, then it should not be added. Its more code then we (I) need to maintain.

SteveMacenski:
Yes, a unique pointer would also work. I'd accept a PR to do that if you like. But please do remove the other unnecessary changes and lets just focus on the copy operator that isn't handling the PC2 correctly.

topin89:
Not sure what exactly should I do, so for now just removed functions unrelated to the bug and fixed linting issues. If I should replace the pointer with `unique_ptr`, I'll change to that. I could also port this to any branches you think it should be ported to, just not sure what to port, how many PRs to do (I mean, I still want to add those move functions). As I understand it, there should be two requests. one for assignment (with or without unique_ptr) and another for move stuff, for every branch.

SteveMacenski (approved):
Please submit against Galactic and Main branches and I can merge the full set of 3 together

If you wished to switch to a unique pointer as well, that would be fine, but not required. I'm OK with it either way.

topin89:
OK, two followup PRs are done: #2425 #2424

> You will also need to submit a PR https://github.com/ros-planning/navigation.ros.org/blob/master/migration/Galactic.rst against this file adding a new section about this change to the migration guide.

But what to submit there? This is just a bugfix, and there's no difference between `observation.hpp` in `galactic` and `main` branches.

I can think of only
> * Various bug fixes

or
> * Fixed a double free bug 

to "Major Improvements", but that's not exactly major.

SteveMacenski (approved):

PR #2419, opened 2021-06-20
Fix code coverage

ruffsl:
This mainly fixes the include filter to only look for the package name in the paths reported, as using package source paths can exclude coverage for files from other packages, such as header files reported using under workspace's install path.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,6 @@
 fixes:
-  - "src/navigation2/::"
+  - "**/src/navigation2/::"
+  - "**/install/::"
 
 ignore:
   - "*/**/test/*"  # ignore package test directories, e.g. nav2_dwb_controller/costmap_queue/tests
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,6 @@
 fixes:
-  - "**/src/navigation2/::"
-  - "**/install/::"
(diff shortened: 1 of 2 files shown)


tools/code_coverage_report.bash
```diff
@@ -42,21 +42,31 @@ mkdir -p ${LCOVDIR}
...
 
 # Capture executed code data.
-fastcov \
+fastcov --lcov \
   -d build \
-  --exclude test/ \
+  --exclude test/ $EXCLUDE_PACKAGES \
   --include $INCLUDE_PACKAGES \
-  --output ${LCOVDIR}/total_coverage.info --lcov
+  --branch-coverage \
```
ruffsl, line 65: Was there a reason we weren't reporting branch coverage before?

codecov.yml
```diff
@@ -1,5 +1,6 @@
 fixes:
-  - "src/navigation2/::"
+  - "**/src/navigation2/::"
```
ruffsl, line 2: I made this use a glob pattern to fix absolute file paths. Not sure why a directory globing wasn't used before.

SteveMacenski:
Merge as you're happy with it

SteveMacenski (approved):

PR #2420, opened 2021-06-22
Improve coverage for NavFn package

lucabonamini:
---

## Basic Info

| ------ | ----------- |

---

* Added invalid start position test to check NavFn planner behaviour

#### For Maintainers: 

### DIFF
--- a/nav2_system_tests/src/system/nav_to_pose_tester_node.py
+++ b/nav2_system_tests/src/system/nav_to_pose_tester_node.py
@@ -380,6 +380,15 @@ def get_testers(args):
     return testers
 
 
+def check_args(expect_failure: str):
+    # Check if --expect_failure is True or False
+    if expect_failure != 'True' and expect_failure != 'False':
+        print('\033[1;37;41m' + ' -e flag must be set to True or False only. ' + '\033[0m')
+        exit(1)
+    else:
+        return eval(expect_failure)
+
+
 def main(argv=sys.argv[1:]):
     # The robot(s) positions from the input arguments
     parser = argparse.ArgumentParser(description='System-level navigation tester node')
@@ -395,6 +404,8 @@ def main(argv=sys.argv[1:]):
 
     args, unknown = parser.parse_known_args()
 
+    expect_failure = check_args(args.expect_failure)
+
     rclpy.init()
 
     # Create testers for each robot
@@ -405,7 +416,7 @@ def main(argv=sys.argv[1:]):
 
     for tester in testers:
         passed = run_all_tests(tester)
-        if str(passed) != str(args.expect_failure):
+        if passed != expect_failure:
             break
 
     for tester in testers:
@@ -414,7 +425,7 @@ def main(argv=sys.argv[1:]):
 
     testers[0].info_msg('Done Shutting Down.')
 
-    if str(passed) != str(args.expect_failure):
+    if passed != expect_failure:
         testers[0].info_msg('Exiting failed')
         exit(1)
     else:
--- a/nav2_system_tests/src/system/nav_to_pose_tester_node.py
+++ b/nav2_system_tests/src/system/nav_to_pose_tester_node.py
@@ -383,6 +383,7 @@ def get_testers(args):
 def main(argv=sys.argv[1:]):
     # The robot(s) positions from the input arguments
     parser = argparse.ArgumentParser(description='System-level navigation tester node')
+    parser.add_argument('-e', '--expect_failure')
     group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument('-r', '--robot', action='append', nargs=4,
                        metavar=('init_x', 'init_y', 'final_x', 'final_y'),
@@ -404,7 +405,7 @@ def main(argv=sys.argv[1:]):
 
     for tester in testers:
         passed = run_all_tests(tester)
-        if not passed:
+        if str(passed) != str(args.expect_failure):
             break
 
     for tester in testers:
@@ -413,7 +414,7 @@ def main(argv=sys.argv[1:]):
 
     testers[0].info_msg('Done Shutting Down.')
 
-    if not passed:
+    if str(passed) != str(args.expect_failure):
         testers[0].info_msg('Exiting failed')
         exit(1)
     else:
--- a/nav2_system_tests/src/system/test_system_launch.py
+++ b/nav2_system_tests/src/system/test_system_launch.py
@@ -103,7 +103,7 @@ def generate_launch_description():
     ])
 
 
-def main(argv=sys.argv[1:]):
+def main1(argv=sys.argv[1:]):
     ld = generate_launch_description()
 
     test1_action = ExecuteProcess(
@@ -119,5 +119,26 @@ def main(argv=sys.argv[1:]):
     return lts.run(ls)
 
 
+def main2(argv=sys.argv[1:]):
+    ld = generate_launch_description()
+
+    test2_action = ExecuteProcess(
+        cmd=[os.path.join(os.getenv('TEST_DIR'), os.getenv('TESTER')),
+             '-r', '-200000.0', '-200000.0', '0.0', '2.0'],
+        name='tester_node',
+        output='screen')
+
+    lts = LaunchTestService()
+    lts.add_test_action(ld, test2_action)
+    ls = LaunchService(argv=argv)
+    ls.include_launch_description(ld)
+    return lts.run(ls)
+
+
 if __name__ == '__main__':
-    sys.exit(main())
+    result1 = main1()
+    result2 = main2()
+    if result1 is False or result2 is False:
+        sys.exit(False)
+    else:
+        sys.exit(True)
--- a/nav2_system_tests/src/system/CMakeLists.txt
+++ b/nav2_system_tests/src/system/CMakeLists.txt
@@ -15,6 +15,23 @@ ament_add_test(test_bt_navigator
     PLANNER=nav2_navfn_planner/NavfnPlanner
 )
 
+ament_add_test(test_bt_navigator_with_wrong_init_pose
+  GENERATE_RESULT_FOR_RETURN_CODE_ZERO
+  COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/test_wrong_init_pose_launch.py"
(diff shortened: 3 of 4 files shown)


nav2_system_tests/src/system/test_system_launch.py
```diff
@@ -119,5 +119,26 @@ def main(argv=sys.argv[1:]):
     return lts.run(ls)
 
 
+def main2(argv=sys.argv[1:]):
```
SteveMacenski, line 122: Can we rename these functions to something more meaningful, like what these tests are testing (what is the new test testing, exactly?)

SteveMacenski (changes requested):

nav2_system_tests/src/system/test_system_launch.py
lucabonamini: The new test verify that *makePlan* function return false if an invalid robot start position is given as input.

SteveMacenski: Then why is the return code `true`? You never check any results to know if the correct thing happened?

lucabonamini: I can't get where you found that exit code. Anyway, you're right, I don't check anywhere that the test result is what I expect (the equivalent of EXPECT_something). It is not clear to me how to do this check in this case, as it is more an integration test than a unit test.

SteveMacenski: yeah but even an integration test needs to check for a given result. The return code is that last statement where you look at `result`. You can add code to the tester node python script to look at more specific items

lucabonamini: Okay, I'll try to make some changes. Meanwhile, thanks for your feedback.

lucabonamini: **nav_to_pose_tester_node.py** can already pass/fails depending on input robot position. So I think that a cleaner way to test if a path cannot be created with the given start position if to create another *test_system_launch.py*, which will have different **cmd** arguments. What do you think about it?

lucabonamini: Alternatively, we could use different configuration files for each test, much like it is now done in *nav2_system_tests/src/system/CMakeLists.txt*.

SteveMacenski: yeah that could be done. Or another input field about "expect failure" so that if you set in a bogus initial pose, you can tell the test you expect everything to fail.

lucabonamini: You mean something like this `test_action = ExecuteProcess(
        cmd=[os.path.join(os.getenv('TEST_DIR'), os.getenv('TESTER')),
             '-r', '-2.0', '-0.5', '0.0', '2.0', False]` ? 

What about having also initial pose as an ENV argument (`test_action = ExecuteProcess(
        cmd=[os.path.join(os.getenv('TEST_DIR'), os.getenv('TESTER')),
             '-r', os.getenv('TEST_INIT_X'), os.getenv('TEST_INIT_Y'), '0.0', '2.0', False]`), so that it's possible to have a specific test (e.g. *test_invalid_start_position.py*) ?

SteveMacenski: I don't think that would work because you'd still also have to set a flag for whether or not you expect the test to be able to pass or fail, and you need logic within the test itself to take that input and do something with it to fail / succeed the test in the appropriate situation. 

You should just have another launch file with the other starting pose so that you can set the new pose as well as whatever flag you want to send to as an input to that `ExecuteProcess` to denote that you expect this to fail

lucabonamini: I can definitely create a new launch file to test a wrong initial pose. However, if a new flag is pass to ExecuteProcess I have to modify nav_to_pose_tester_node.py in order to use it, don't I? Because if I have just to check the result of main() function, there's no need to pass no new flag. Sorry, these python tests are driving me crazy 😕

SteveMacenski: Yes, you'd have to modify it to take another argument for whether you expect it to fail or succeed. If succeed, nothing changes. If fail, then when you're running the tests, you should expect the result to be a fail not a success (and then if it does succeed you should fail the test). Just as there are arguments like `-r` you should add a new one for this expect to succeed that can be grabbed in the `main()` function next to the other arguments and then used where it makes sense to in the file. 

E.g. https://github.com/ros-planning/navigation2/blob/6e0db7a37c196cc088428e7b42025351e64882f7/nav2_system_tests/src/system/nav_to_pose_tester_node.py#L406 passed should be treated inverted when you expect it to fail

lucabonamini: Could you check last commit?

nav2_system_tests/src/system/nav_to_pose_tester_node.py
```diff
@@ -404,7 +405,7 @@ def main(argv=sys.argv[1:]):
 
     for tester in testers:
         passed = run_all_tests(tester)
-        if not passed:
+        if str(passed) != str(args.expect_failure):
```
SteveMacenski, line 408: Don't compare strings, these should be tested as bools (as well below)

nav2_system_tests/src/system/test_system_launch.py
SteveMacenski: Generally looks good!

nav2_system_tests/src/system/nav_to_pose_tester_node.py
lucabonamini: I used strings because `passed != args.expect_failure` didn't work as expected, even if `args.expect_failure` is evaluated correctly. 

What about `passed != eval(args.expect_failure)` ?  It definitely works (I've already test it), but I don't know if it's the best way.

SteveMacenski: You might want to error check before `eval` but that's generally an OK strategy here since this is a test. I'd just verify the string is exactly `True` or `False` and if not, log an error because eval is going to probably fail

lucabonamini: I put the check on `eval()` argument in `get_testers(args)`, so that I can use the `error_msg` already define.

    if args.robot:
        # Requested tester for one robot
        init_x, init_y, final_x, final_y = args.robot[0]
        tester = NavTester(
            initial_pose=fwd_pose(float(init_x), float(init_y)),
            goal_pose=fwd_pose(float(final_x), float(final_y)))
        # Check if --expect_failure is True or False
        if args.expect_failure != 'True' and args.expect_failure != 'False':
            tester.error_msg("-e flag must be set True or False only.")
            exit(1)
        tester.info_msg(
            'Starting tester, robot going from ' + init_x + ', ' + init_y +
            ' to ' + final_x + ', ' + final_y + '.')
        testers.append(tester)
        return testers

    # Requested tester for multiple robots
    for robot in args.robots:
        namespace, init_x, init_y, final_x, final_y = robot
        tester = NavTester(
            namespace=namespace,
            initial_pose=fwd_pose(float(init_x), float(init_y)),
            goal_pose=fwd_pose(float(final_x), float(final_y)))
        # Check if --expect_failure is True or False
        if args.expect_failure != 'True' and args.expect_failure != 'False':
            tester.error_msg("-e flag must be set True or False only.")
            exit(1)
        tester.info_msg(
            'Starting tester for ' + namespace +
            ' going from ' + init_x + ', ' + init_y +
            ' to ' + final_x + ', ' + final_y)
        testers.append(tester)
    return testers

lucabonamini: I create a new **NavTester** method:
```
def check_args(self, expect_failure: str):
  # Check if --expect_failure is True or False
  if expect_failure != 'True' and expect_failure != 'False':
    self.error_msg("-e flag must be set True or False only.")
    exit(1)
```

then call it in `get_testers(args)`:

```
def get_testers(args):
    testers = []

    if args.robot:
        # Requested tester for one robot
        init_x, init_y, final_x, final_y = args.robot[0]
        tester = NavTester(
            initial_pose=fwd_pose(float(init_x), float(init_y)),
            goal_pose=fwd_pose(float(final_x), float(final_y)))
        tester.check_args(args.expect_failure)
        tester.info_msg(
            'Starting tester, robot going from ' + init_x + ', ' + init_y +
            ' to ' + final_x + ', ' + final_y + '.')
        testers.append(tester)
        return testers

    # Requested tester for multiple robots
    for robot in args.robots:
        namespace, init_x, init_y, final_x, final_y = robot
        tester = NavTester(
            namespace=namespace,
            initial_pose=fwd_pose(float(init_x), float(init_y)),
            goal_pose=fwd_pose(float(final_x), float(final_y)))
        tester.check_args(args.expect_failure)
        tester.info_msg(
            'Starting tester for ' + namespace +
            ' going from ' + init_x + ', ' + init_y +
            ' to ' + final_x + ', ' + final_y)
        testers.append(tester)
    return testers
```

SteveMacenski: Or you could just make check args an independent function from the tester and have it return the bool for T/F while its checking (to use then later rather than eval in the `main()`)

lucabonamini: What do you think about last commit?

Even if locally everything works fine, CI build shows this result for *test_bt_navigator_with_wrong_init_pose*:

`The test did not generate a result file.`

SteveMacenski: That means the test crashed, failed, or hung, you should run locally and debug.

lucabonamini: I just checked [CI builds](https://app.circleci.com/pipelines/github/ros-planning/navigation2?branch=pull%2F2420). 

2h ago another build has been triggered somehow, and the error relative to *test_bt_navigator_with_wrong_init_pose* is gone.

Screenshot from 2021-07-12 21-05-43

Screenshot from 2021-07-12 21-05-31

SteveMacenski: I retriggered it just in case it was a transient failure, looks like it was!

SteveMacenski (approved):
Awesome, time to merge!

PR #2427, opened 2021-06-29
Feature addition: capability for the RRP to drive the robot backwards

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2327  |
| Robotic platform tested on | gazebo simulation of neobotix mp-400 |

---

* I added this neat new feature for driving the robot using the RRP planner backwards. This enables car-like robots to efficiently perform the maneuvers using the SMAC planner. 

* Added new parameter "reverse_driving", so need to add that to default configs and documentation page

---

* I tested on a differential drive robot, but it will be great if someone can test it on an Ackermann drive! 

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -198,7 +198,7 @@ protected:
   void applyConstraints(
     const double & dist_error, const double & lookahead_dist,
     const double & curvature, const geometry_msgs::msg::Twist & speed,
-    const double & pose_cost, double & linear_vel);
+    const double & pose_cost, double & linear_vel, double & sign);
 
   /**
    * @brief Get lookahead point
@@ -208,9 +208,20 @@ protected:
    */
   geometry_msgs::msg::PoseStamped getLookAheadPoint(const double &, const nav_msgs::msg::Path &);
 
-  double checkOrientation(const geometry_msgs::msg::PoseStamped & pose);
+  /**
+   * @brief checks for the cusp position
+   * @param pose Pose input to determine the cusp position
+   * @return robot distance from the cusp
+   */
+  double detectPathOrientation(const geometry_msgs::msg::PoseStamped & pose);
 
-  double dot(double q1[], double q2[]);
+  /**
+   * @brief determines the dot product
+   * @param u_x, u_y, represents the vector u
+   * @param v_x, v_y, represents the vector v 
+   * @return robot distance from the cusp
+   */
+  inline double dot(double u_x, double u_y, double v_x, double v_y);
 
   std::shared_ptr<tf2_ros::Buffer> tf_;
   std::string plugin_name_;
@@ -243,7 +254,7 @@ protected:
   double max_angular_accel_;
   double rotate_to_heading_min_angle_;
   double goal_dist_tol_;
-  bool reverse_driving_;
+  bool allow_reversing_;
 
   nav_msgs::msg::Path global_plan_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> global_path_pub_;
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -208,6 +208,10 @@ protected:
    */
   geometry_msgs::msg::PoseStamped getLookAheadPoint(const double &, const nav_msgs::msg::Path &);
 
+  double checkOrientation(const geometry_msgs::msg::PoseStamped & pose);
+
+  double dot(double q1[], double q2[]);
+
   std::shared_ptr<tf2_ros::Buffer> tf_;
   std::string plugin_name_;
   std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
@@ -239,6 +243,7 @@ protected:
   double max_angular_accel_;
   double rotate_to_heading_min_angle_;
   double goal_dist_tol_;
+  bool reverse_driving_;
 
   nav_msgs::msg::Path global_plan_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> global_path_pub_;
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -213,15 +213,7 @@ protected:
    * @param pose Pose input to determine the cusp position
    * @return robot distance from the cusp
    */
-  double detectPathOrientation(const geometry_msgs::msg::PoseStamped & pose);
-
-  /**
-   * @brief determines the dot product
-   * @param u_x, u_y, represents the vector u
-   * @param v_x, v_y, represents the vector v 
-   * @return robot distance from the cusp
-   */
-  inline double dot(double u_x, double u_y, double v_x, double v_y);
+  double findDirectionChange(const geometry_msgs::msg::PoseStamped & pose);
 
   std::shared_ptr<tf2_ros::Buffer> tf_;
   std::string plugin_name_;
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -480,7 +480,7 @@ void RegulatedPurePursuitController::applyConstraints(
     if (unbounded_vel < min_approach_linear_velocity_) {
       approach_vel = min_approach_linear_velocity_;
     } else {
-      approach_vel = approach_vel * velocity_scaling;
+      approach_vel *= velocity_scaling;
     }
     // Use the lowest velocity between approach and other constraints, if all overlapping
     linear_vel = std::min(linear_vel, approach_vel);
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -376,7 +376,7 @@ bool RegulatedPurePursuitController::isCollisionImminent(
   pose_msg.header.frame_id = arc_pts_msg.header.frame_id;
   pose_msg.header.stamp = arc_pts_msg.header.stamp;
 
-  const double projection_time = costmap_->getResolution() / fabs(linear_vel);
+  const double projection_time = costmap_->getResolution() / linear_vel;
 
   geometry_msgs::msg::Pose2D curr_pose;
   curr_pose.x = robot_pose.pose.position.x;
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -581,8 +581,9 @@ double RegulatedPurePursuitController::findDirectionChange(
       global_plan_.poses[pose_id].pose.position.y;
 
     /* Checking for the existance of cusp, in the path, using the dot product
-    and determine it's distance from the robot */
-    if ( (oa_x * ab_x) + (oa_y * ab_y) <= 0) {
+    and determine it's distance from the robot. If there is no cusp in the path,
+    then just determine the distance to the goal location. */
+    if ( (oa_x * ab_x) + (oa_y * ab_y) < 0.0 || pose_id == global_plan_.poses.size() ) {
       auto x = global_plan_.poses[pose_id].pose.position.x - pose.pose.position.x;
       auto y = global_plan_.poses[pose_id].pose.position.y - pose.pose.position.y;
       return hypot(x, y); // returning the distance if there is a cusp
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -118,7 +118,7 @@ void RegulatedPurePursuitController::configure(
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".goal_dist_tol", rclcpp::ParameterValue(0.25));
   declare_parameter_if_not_declared(
-    node, plugin_name_ + ".reverse_driving", rclcpp::ParameterValue(false));
+    node, plugin_name_ + ".allow_reversing", rclcpp::ParameterValue(false));
 
   node->get_parameter(plugin_name_ + ".desired_linear_vel", desired_linear_vel_);
   node->get_parameter(plugin_name_ + ".max_linear_accel", max_linear_accel_);
@@ -145,7 +145,7 @@ void RegulatedPurePursuitController::configure(
   node->get_parameter(plugin_name_ + ".rotate_to_heading_min_angle", rotate_to_heading_min_angle_);
   node->get_parameter(plugin_name_ + ".max_angular_accel", max_angular_accel_);
   node->get_parameter(plugin_name_ + ".goal_dist_tol", goal_dist_tol_);
-  node->get_parameter(plugin_name_ + ".reverse_driving", reverse_driving_);
+  node->get_parameter(plugin_name_ + ".allow_reversing", allow_reversing_);
   node->get_parameter("controller_frequency", control_frequency);
 
   transform_tolerance_ = tf2::durationFromSec(transform_tolerance);
@@ -157,11 +157,13 @@ void RegulatedPurePursuitController::configure(
     use_cost_regulated_linear_velocity_scaling_ = false;
   }
 
-  // Possible to drive in reverse direction if and only if "use_rotate_to_heading" parameter is set to false
+  /** Possible to drive in reverse direction if and only if 
+   "use_rotate_to_heading" parameter is set to false **/
 
-  if(use_rotate_to_heading_)
-  {
-    reverse_driving_ = false;
+  if (use_rotate_to_heading_ && allow_reversing_) {
+    RCLCPP_WARN(logger_, "Both use_rotate_to_heading and allow_reversing parameter cannot "
+      "be set to true. By default setting use_rotate_to_heading true");
+    allow_reversing_ = false;
   }
 
   global_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("received_global_plan", 1);
@@ -236,17 +238,15 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Transform path to robot base frame
   auto transformed_plan = transformGlobalPlan(pose);
 
-  double lookahead_dist = getLookAheadDistance(speed);  
+  double lookahead_dist = getLookAheadDistance(speed);
 
   // Check for reverse driving
-  if(reverse_driving_)
-  {
+  if (allow_reversing_) {
     // Cusp check
-    double orientation_change_dist_= checkOrientation(pose);
+    double orientation_change_dist_ = detectPathOrientation(pose);
 
-    // Calculate the carrot pose based on the lookahead distance and/or the orientation_change_dist
-    if(orientation_change_dist_ < lookahead_dist)
-    {
+    // if the lookahead distance is further than the cusp, use the cusp distance instead
+    if (orientation_change_dist_ < lookahead_dist) {
       lookahead_dist = orientation_change_dist_;
     }
   }
@@ -271,11 +271,11 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Setting the velocity direction
   double sign = 1.0;
 
-  if(reverse_driving_){
-    sign = carrot_pose.pose.position.x >= 0.0 ? 1.0:-1.0;
+  if (allow_reversing_) {
+    sign = carrot_pose.pose.position.x >= 0.0 ? 1.0 : -1.0;
   }
 
-  linear_vel = sign*desired_linear_vel_;
+  linear_vel = sign * desired_linear_vel_;
 
   // Make sure we're in compliance with basic constraints
   double angle_to_heading;
@@ -288,7 +288,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     applyConstraints(
       fabs(lookahead_dist - sqrt(carrot_dist2)),
       lookahead_dist, curvature, speed,
-      costAtPose(pose.pose.position.x, pose.pose.position.y), linear_vel);
+      costAtPose(pose.pose.position.x, pose.pose.position.y), linear_vel, sign);
 
       // Apply curvature to angular velocity after constraining linear velocity
       angular_vel = linear_vel * curvature;
@@ -389,7 +389,7 @@ bool RegulatedPurePursuitController::isCollisionImminent(
     if (i * projection_time > max_allowed_time_to_collision_) {
       break;
     }
-    
+
     i++;
 
     // apply velocity at curr_pose over distance
@@ -441,12 +441,11 @@ double RegulatedPurePursuitController::costAtPose(const double & x, const double
 void RegulatedPurePursuitController::applyConstraints(
   const double & dist_error, const double & lookahead_dist,
   const double & curvature, const geometry_msgs::msg::Twist & curr_speed,
-  const double & pose_cost, double & linear_vel)
+  const double & pose_cost, double & linear_vel, double & sign)
 {
   double curvature_vel = linear_vel;
   double cost_vel = linear_vel;
   double approach_vel = linear_vel;
-  double sign = linear_vel > 0.0 ? 1.0:-1.0;
 
   // limit the linear velocity by curvature
   const double radius = fabs(1.0 / curvature);
@@ -481,19 +480,19 @@ void RegulatedPurePursuitController::applyConstraints(
     if (fabs(unbounded_vel) < fabs(min_approach_linear_velocity_)) {
       approach_vel = min_approach_linear_velocity_;
     } else {
-      approach_vel = fabs(approach_vel)*velocity_scaling;
+      approach_vel = fabs(approach_vel) * velocity_scaling;
     }
     // Use the lowest velocity between approach and other constraints, if all overlapping
     linear_vel = std::min(fabs(linear_vel), fabs(approach_vel));
   }
   // Limit linear velocities to be valid and kinematically feasible, v = v0 + a * dt
-  linear_vel = sign*linear_vel;
+  linear_vel = sign * linear_vel;
   double & dt = control_duration_;
   const double max_feasible_linear_speed = curr_speed.linear.x + max_linear_accel_ * dt;
   const double min_feasible_linear_speed = curr_speed.linear.x - max_linear_decel_ * dt;
   linear_vel = std::clamp(linear_vel, min_feasible_linear_speed, max_feasible_linear_speed);
   linear_vel = std::clamp(fabs(linear_vel), 0.0, desired_linear_vel_);
-  linear_vel = sign*linear_vel;
+  linear_vel = sign * linear_vel;
 }
 
 void RegulatedPurePursuitController::setPlan(const nav_msgs::msg::Path & path)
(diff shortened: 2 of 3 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -154,6 +157,13 @@ void RegulatedPurePursuitController::configure(
     use_cost_regulated_linear_velocity_scaling_ = false;
   }
 
+  // Possible to drive in reverse direction if and only if "use_rotate_to_heading" parameter is set to false
+
+  if(use_rotate_to_heading_)
+  {
```
SteveMacenski, line 163: This will fail linting, please follow the ROS2 style guide and run tests over code before submitting PRs. This can be caught locally by doing so or running ament_cpplint and ament_uncrustify

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +566,44 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
   return transformed_plan;
 }
 
+double RegulatedPurePursuitController::dot(double v1[], double v2[])
+{
+  // Supports only 2D
+  return  (v1[0]*v2[0]) + (v1[1]*v2[1]);
```
SteveMacenski, line 572: This is only used in 1 place, just do it inline in that function

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Why is this even related to rotating to heading? It shouldn't be

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -226,8 +236,21 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
 
-  // Find look ahead distance and point on path and publish
-  const double lookahead_dist = getLookAheadDistance(speed);
+  double lookahead_dist = getLookAheadDistance(speed);  
+
+  // Check for reverse driving
+  if(reverse_driving_)
+  {
+    // Cusp check
+    double orientation_change_dist_= checkOrientation(pose);
```
SteveMacenski, line 245: This function name is not good, I have no idea what its doing. Please rename this function to something inline readable to know what's going on.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -226,8 +236,21 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+
+    // Calculate the carrot pose based on the lookahead distance and/or the orientation_change_dist
+    if(orientation_change_dist_ < lookahead_dist)
```
SteveMacenski, line 248: I do not understand what this is accomplishing

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -245,7 +268,14 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     curvature = 2.0 * carrot_pose.pose.position.y / carrot_dist2;
   }
 
-  linear_vel = desired_linear_vel_;
+  // Setting the velocity direction
+  double sign = 1.0;
+
+  if(reverse_driving_){
```
SteveMacenski, line 274: Why is this called `reverse_driving_`? What is this boolean meant for? If the path contains changes in velocity directions, we must follow it, this is a pure pursuit planner, it cannot deviate

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -245,7 +268,14 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+    sign = carrot_pose.pose.position.x >= 0.0 ? 1.0:-1.0;
```
SteveMacenski, line 275: These semantics would make a whole lot more sense if you just had a flag you could work with that it detected that it should change signs due to a cusp

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -439,30 +470,30 @@ void RegulatedPurePursuitController::applyConstraints(
   }
 
   // Use the lowest of the 2 constraint heuristics, but above the minimum translational speed
-  linear_vel = std::min(cost_vel, curvature_vel);
```
SteveMacenski, line 442: I think in this function this might work, but you should potentially try to slow down a little and make this more human understandable. You get the sign in the main function, apply it to the velocity, and then feed it back into here where you end up doing a bunch of other mechanics with fabs and whatnot. Would there not just be a single place that when you detect a change in direction you could flip the sign? Or a flag you could pass around to functions / areas of code that need it to let it know that its sign is flipped?

SteveMacenski: It would also be good to verify yourself that you indeed do set the signs correctly so that the heuristics and the collision checker are using the correct velocity signed direction (since that will be important for the collision checker to be checking in the right direction of travel)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +566,44 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
+}
+
+double RegulatedPurePursuitController::checkOrientation(const geometry_msgs::msg::PoseStamped & pose )
+{
+  double dot_sign = 0.0;
+  double iter = 1;
+  double cusp_distance = 0.0;
+  bool not_in_cusp = true;
+
+  do
```
SteveMacenski, line 582: Restructure this as a for loop and actually use iterators, `iter` is an indexer and you can easily format this as a `for begin until end, iter++`. for loops are always preferable to while loops since you know the bounded maximum number iterations

SteveMacenski: You can remove the entire `not_in_cusp` logic too since that is just there as a  terminal condition of the while loop, which the for would have just by exiting.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +566,44 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
+  {
+    // We have two vectors for the dot product OA and AB. Determining the vectors.
+    double oa_x = global_plan_.poses[iter].pose.position.x - global_plan_.poses[iter-1].pose.position.x;
+    double oa_y = global_plan_.poses[iter].pose.position.y - global_plan_.poses[iter-1].pose.position.y;
+    double ab_x = global_plan_.poses[iter+1].pose.position.x - global_plan_.poses[iter].pose.position.x;
+    double ab_y = global_plan_.poses[iter+1].pose.position.y - global_plan_.poses[iter].pose.position.y;
+    double OA[] = {oa_x,oa_y};
```
SteveMacenski, line 589: Never do this in C++, we have access to such amazing data structures and tools. Also, there's really not a strong reason to throw  these into an array since you're just doing math operations on them and then throwing them out. Just do your thing with the `_x` and `_y` variables, no need to combine them unless there's something else you're doing like passing them around a bunch or storing them longer term.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +566,44 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
+    double AB[] = {ab_x,ab_y};
+    // Checking for the existance of cusp using the dot product
+    dot_sign = dot(OA, AB) > 0.0 ? 1.0 : -1.0;
```
SteveMacenski, line 592: I'm pretty sure you can change `dot_sign` to `bool path_direction_changed` without any loss of generality and much more clean to read.

SteveMacenski (changes requested):
Good first prototype stab at it! Needs some clean up though in formatting, readability, and a bit of simplification

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: This is the part where I compare the lookahead distance with distance to the cusp. Till the point where the lookahead distance is smaller than the cusp, I continue to use the lookahead distance and once when the distance to the cusp is less than the lookahead distance, then I just use the distance to the cusp as the lookahead distance for holding the carrot to the cusp. By this way, we do not have to unnecessarily have long lookahead distances.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: This is continuation of the first conversation. 

> What is this boolean meant for? 

The boolean here is to determine the position of the carrot, according to which we will be setting the linear velocity. As the position of the carrot gives the information related to the base_link frame. In case if a user decides to skip the reverse driving then, we never want to look at the sign of the carrot, since we already have a functionality that would provide as the rotation to the heading. In this scenario, the boolean will not be necessary in any case.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: Ah this is a cool idea! I will do change it! :)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: I just mean in terms of this code, you end up finding the "sign" in a couple of places that might be better served with a boolean that is just passed to a function to use instead so its not being re-found in multiple places based on derived values

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Can't we just remove that parameter? If the path has a cusp in it, the robot has to reverse to follow it. Otherwise, the robot would be totally stuck. I suppose if it were a diff drive robot, it could follow the carrot into a cusp and then just rotate in place to follow... OK. I'm fine keeping the option to disable reversing, but lets rename this then so its more clear `allow_reversing` parameter?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Yeah I think that's a good call. It might be appropriate at times to use both. Though it would be worth testing on your side if you have both set what actually happens at cusping events

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Got it - makes sense, that comment really doesn't make much sense though. Please update that comment to explain that better / more concisely. `if the lookahead distance is further than the cusp, use the cusp distance instead`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: > I suppose if it were a diff drive robot, it could follow the carrot into a cusp and then just rotate in place to follow...

This indeed sounds to be a fancy and intelligent stuff from a controller. But, curious to know from a user perspective on what the benefit would be for them. 

> OK. I'm fine keeping the option to disable reversing, but lets rename this then so its more clear allow_reversing parameter? 

Agreed! That sounds right.

padhupradheep: OK. I'm fine keeping the option to disable reversing, but lets rename this then so its more clear `allow_reversing` parameter?

Agreed with the name!

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: I gave it a go and when I set both the params to true, all the time it rotates to the heading and goes forward. Here reversing is not considered at all.

SteveMacenski: Got it OK. Then I think you are correct to have these be mutually exclusive. Please update the readme and navigation.ros.org configuration page (https://github.com/ros-planning/navigation.ros.org/blob/master/configuration/packages/configuring-regulated-pp.rst) to mention that in those parameterizations. We should also check on initialization if both are set to true to throw an error saying that is an invalid configuration, only 1 can be on at a time. Then default to rotating in place (since I think of the 2, that's the "safer" default). Do you agree?

padhupradheep: > We should also check on initialization if both are set to true to throw an error saying that is an invalid configuration, only 1 can be on at a time.

That would indeed be a fail proof 

> Then default to rotating in place (since I think of the 2, that's the "safer" default). Do you agree?

Sure, I agree to this point. I will update the documentation in parallel to the changes we discussed here ✌🏽

SteveMacenski: Thanks for this great work! It will be great to have!

padhupradheep: No problem, thank you for the guidance 😃

padhupradheep: > We should also check on initialization if both are set to true to throw an error saying that is an invalid configuration, only 1 can be on at a time. Then default to rotating in place (since I think of the 2, that's the "safer" default).

Here maybe, we could just throw a warn message stating that, both the parameters are set to true, so we are setting the default to rotating in place ?

SteveMacenski: yes

padhupradheep:
ament_uncrustify and ament_cpplint fails for the other part of the code, which was not written as part of this feature.

I will do the cleaning up and give a separate PR!

SteveMacenski:
This feature should also be targeted to `main` not `foxy` we can merge it back into foxy once its done, that might be why some of the linters are failing, those profiles have changed over time. Please retarget to `main`

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -208,6 +208,21 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
+   */
+  double detectPathOrientation(const geometry_msgs::msg::PoseStamped & pose);
+
+  /**
+   * @brief determines the dot product
+   * @param u_x, u_y, represents the vector u
+   * @param v_x, v_y, represents the vector v 
+   * @return robot distance from the cusp
+   */
+  inline double dot(double u_x, double u_y, double v_x, double v_y);
```
SteveMacenski, line 224: Please remove this function. You only use it once and it is so simple you can just do it inline in the actual function that uses it. I think I made this comment too last time :laughing:

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -154,6 +157,15 @@ void RegulatedPurePursuitController::configure(
     use_cost_regulated_linear_velocity_scaling_ = false;
   }
 
+  /** Possible to drive in reverse direction if and only if 
+   "use_rotate_to_heading" parameter is set to false **/
+
+  if (use_rotate_to_heading_ && allow_reversing_) {
+    RCLCPP_WARN(logger_, "Both use_rotate_to_heading and allow_reversing parameter cannot "
```
SteveMacenski, line 164: Mention "disabling reversing"

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -226,8 +238,19 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Transform path to robot base frame
   auto transformed_plan = transformGlobalPlan(pose);
 
-  // Find look ahead distance and point on path and publish
```
SteveMacenski, line 229: Add this comment back?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -245,7 +268,14 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     curvature = 2.0 * carrot_pose.pose.position.y / carrot_dist2;
   }
 
-  linear_vel = desired_linear_vel_;
+  // Setting the velocity direction
+  double sign = 1.0;
+
```
SteveMacenski, line 273: remove this white space

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -245,7 +268,14 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+  if (allow_reversing_) {
+    sign = carrot_pose.pose.position.x >= 0.0 ? 1.0 : -1.0;
+  }
+
+  linear_vel = sign * desired_linear_vel_;
```
SteveMacenski, line 278: Why set linear velocity here at all? Won't everything just be fabs-ed in the applyConstraint function? Then you could remove the fabs-ing you did in that function and just apply the `sign` when required

SteveMacenski: You could largely revert all of the changes you made in applyConstraints before L489

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +565,46 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
+}
+
+double RegulatedPurePursuitController::detectPathOrientation(
+  const geometry_msgs::msg::PoseStamped & pose )
+{
+  double path_direction_changed = 0.0;
+  double cusp_distance = 0.0;
+
+  // Iterating through the global path to determine the position of the cusp
+  for (auto iter = 1; iter <= (int)global_plan_.poses.size(); ++iter) {
```
SteveMacenski, line 581: This is not an iterator, please don't mix those up. This should be `i` or `pose_id` or something.

SteveMacenski: And you can use `unsigned int` rather than `auto` so that you don't need the implicit cast from `size()` (so remove that part)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +565,46 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
+      global_plan_.poses[iter].pose.position.y;
+
+    // Checking for the existance of cusp using the dot product
+    path_direction_changed = dot(oa_x, oa_y, ab_x, ab_y) > 0.0 ? 1.0 : -1.0;
+
+    // Getting the distance of the cusp, if the cusp is found
+    if (path_direction_changed == -1.0) {
+      auto x = global_plan_.poses[iter].pose.position.x - pose.pose.position.x;
+      auto y = global_plan_.poses[iter].pose.position.y - pose.pose.position.y;
+      cusp_distance = sqrt((x * x) + (y * y));
```
SteveMacenski, line 599: We have `hypot` in C++ standard library that will do this for you in a nan-safe way

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +565,46 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
+    double ab_x = global_plan_.poses[iter + 1].pose.position.x -
+      global_plan_.poses[iter].pose.position.x;
+    double ab_y = global_plan_.poses[iter + 1].pose.position.y -
+      global_plan_.poses[iter].pose.position.y;
+
+    // Checking for the existance of cusp using the dot product
+    path_direction_changed = dot(oa_x, oa_y, ab_x, ab_y) > 0.0 ? 1.0 : -1.0;
+
+    // Getting the distance of the cusp, if the cusp is found
+    if (path_direction_changed == -1.0) {
```
SteveMacenski, line 596: We don't actually use `path_direction_changed` for anything other than if detecting if negative. So you can remove that variable entirely and move its logic into the `if()` statement. Also you don't need to look at `-1.0` it just needs to be `< 0.0` so this could be written far more compactly as 

`if (dot(...) < 0.0)`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski, line 599: This could be `return hypot(...)` and then the return at the bottom of this function could be `return 0.0` and that would eliminate the `cusp_distance` variable

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +565,46 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
   return transformed_plan;
 }
 
+inline double RegulatedPurePursuitController::dot(double u_x, double u_y, double v_x, double v_y)
+{
+  // Supports only 2D
+  return  (u_x * v_x) + (u_y * v_y);
+}
+
+double RegulatedPurePursuitController::detectPathOrientation(
```
SteveMacenski, line 574: How about `findDirectionChange` and then the variable above that this function's output is used in be `dist_to_direction_change`

SteveMacenski (changes requested):

SteveMacenski:
All good updates, now just to get things into the C++ stylings and best practices

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
padhupradheep: Lol, I thought, you wanted to use "inline" function!

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Make sure that this is `<` not `<=`. If its 0.0, its not a reverse but a 90 degree move which isn't really a cusp worth doing this in.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Make sure to remove `auto` and replace with `unsigned int`. then you can remove the `(int)`

padhupradheep: oops! sorry!

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: This is kind of the edge case, consider we have a straight path driving in reverse direction. Then the `findDirectionChange` function will return 0.0.

Then this [condition at this line](https://github.com/padhupradheep/navigation2/blob/e1ead123df1e85623c15e8141a26a06afb657109/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L250) gets affected, which means that the lookahead point does not move at all.

padhupradheep:
Should this be a new PR ? I do not see any other way to merge the changes from my branch to `main` without conflicts. Or am I missing a method ?

SteveMacenski:
A new PR would work but you'd also run into the same conflicts if you based these changes on Foxy, but you'd instead run into them locally when you tried to rebase. Copy paste would work... though never the most elegant solution :wink:.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -346,7 +376,7 @@ bool RegulatedPurePursuitController::isCollisionImminent(
   pose_msg.header.frame_id = arc_pts_msg.header.frame_id;
   pose_msg.header.stamp = arc_pts_msg.header.stamp;
 
-  const double projection_time = costmap_->getResolution() / linear_vel;
+  const double projection_time = costmap_->getResolution() / fabs(linear_vel);
```
SteveMacenski, line 379:
```suggestion
  const double projection_time = costmap_->getResolution() / linear_vel;
```

fabs not required anymore

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,19 +480,19 @@ void RegulatedPurePursuitController::applyConstraints(
     if (unbounded_vel < min_approach_linear_velocity_) {
       approach_vel = min_approach_linear_velocity_;
     } else {
-      approach_vel *= velocity_scaling;
```
SteveMacenski, line 453: This line can be reverted to `*=`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -535,6 +565,34 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
   return transformed_plan;
 }
 
+double RegulatedPurePursuitController::findDirectionChange(
+  const geometry_msgs::msg::PoseStamped & pose )
+{
+  // Iterating through the global path to determine the position of the cusp
+   for (unsigned int pose_id = 1; pose_id <= global_plan_.poses.size(); ++pose_id)  {
```
SteveMacenski, line 572: This should be `<` not `<=` since size() of 7 has elements of 0,1,2,3,4,5,6. I think in fixing that you can remove your most recent change of `|| pose_id == global_plan_.poses.size()`, yeah? Or if not, please explain why that's there. If this is actually the last point, then why bother detecting it?

SteveMacenski (approved):
Very good, just last clean up before merging. After these changes, do make sure to test a bit and make sure that you've covered all of the edge cases (reversing, reverse then back forward, doing that multiple times in a single path, etc)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: `|| pose_id == global_plan_.poses.size()`

This is very important. There maybe a scenario, where there is no cusp at all in the path (easiest example will be: imagine a robot driving straight to a goal). In that scenario, we need the distance of the goal from the robot's position, so that [here](https://github.com/padhupradheep/navigation2/blob/3c39b58b14e65642d2dd8040ec812d752e367dc5/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L250), we don't satisfy this condition. Because the return value from this function where the code lies will be 0 and the robot will not move at all.

SteveMacenski: Well the `<=` should definitely be `<`, if we can agree on that we can separate the topics 

---

I see. Would having the `lookahead_dist` be an argument to this function help? So instead of returning `0.0` if no cusp is found, instead it is `lookahead_dist`. That way if you iterate through the set and find no cusp, you exit the loop without hitting that condition and return the initial lookahead distance?

padhupradheep: > Well the <= should definitely be <, if we can agree on that we can separate the topics
Agreed, I think it makes sense. 

>  Honestly you bring up a good point, 0.0 is a bad return value. How about change return 0.0 to return std::numeric_limits<double>::max().

This is a valid point, which I would definitely change!

SteveMacenski:
I'll backport this to Foxy/Galactic once the main one is merged

Issue #2429, opened 2021-07-01
Missing export of required public definition PLUGINLIB__DISABLE_BOOST_FUNCTIONS for downstream packages


devrite:
- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - rolling+galactic(flag depr.)
  - foxy
- Version or commit hash:
  - ddfed98937e1c41cb55b380272cb5d76a6871b6c

#### Steps to reproduce issue
```
docker run --rm -it ros:foxy-ros-base-focal
apt-get update && apt-get install -y ros-foxy-navigation2
grep -r PLUGINLIB__DISABLE_BOOST_FUNCTIONS /opt/ros/foxy/share
```

#### Expected behavior
Similar output as below (e.g., also done by the rviz plugin) for all libs and plugins exported via navigation2. Missing for a lot of plugins. Just do a grep on the source to tree to identify which packages do public set the `PLUGINLIB__DISABLE_BOOST_FUNCTIONS`.

Other packages including, for example, headers from `nav2_costmap_2d` will eventually include pluginlib headers but the definition is not set, thus forcing a downstream package to set this definition itself or falsely enabling the boost features if boost is installed via another dependency already (e.g. if the boost headers are found).
 
#### Actual behavior
```
rviz_common/cmake/rviz_commonExport.cmake:  INTERFACE_COMPILE_DEFINITIONS "PLUGINLIB__DISABLE_BOOST_FUNCTIONS"
rviz_default_plugins/cmake/rviz_default_pluginsExport.cmake:  INTERFACE_COMPILE_DEFINITIONS "PLUGINLIB__DISABLE_BOOST_FUNCTIONS"
nav2_rviz_plugins/cmake/nav2_rviz_pluginsExport.cmake:  INTERFACE_COMPILE_DEFINITIONS "PLUGINLIB__DISABLE_BOOST_FUNCTIONS"
rosbag2_cpp/cmake/export_rosbag2_cppExport.cmake:  INTERFACE_COMPILE_DEFINITIONS "PLUGINLIB__DISABLE_BOOST_FUNCTIONS"
rosbag2_storage/cmake/export_rosbag2_storageExport.cmake:  INTERFACE_COMPILE_DEFINITIONS "PLUGINLIB__DISABLE_BOOST_FUNCTIONS"
```

#### Additional information

[...]

PR #2430, opened 2021-07-01
Add modern cmake export for missing definitions

devrite:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2429 |
| Robotic platform tested on | none|

---

* Switch to modern cmake export for packages using the `PLUGINLIB__DISABLE_BOOST_FUNCTIONS`
* Used export_${PROJECT_NAME} or export_${library_name} as by [foxy documentation/convention](https://docs.ros.org/en/foxy/Guides/Ament-CMake-Documentation.html#building-a-library)
* Updated rviz plugin to this convention
* Removed include_directories for directory `include`
* Instead use `target_include_directories pubic build/install_interface` as this is the only way the dir will get exported

Nothing should be required to change

---

* Thorough porting of all packages to modern cmake
* Consistently using cmake vars or names
* Consistently using public and private in all cmake functions for targets

#### For Maintainers: 

### DIFF
--- a/nav2_controller/CMakeLists.txt
+++ b/nav2_controller/CMakeLists.txt
@@ -17,7 +17,6 @@ find_package(pluginlib REQUIRED)
 nav2_package()
 
 include_directories(
-  include
 )
 
 set(executable_name controller_server)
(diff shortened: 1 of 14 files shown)


devrite:
This pr has become obsolete, as the originating issue is solved upstream for newer distros.

PR #2436, opened 2021-07-06
Wait for goal result after cancelling BT action node (Galactic branch)

anaelle-sw:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses | I didn't find any issue that is currently open, and that would be related to this PR. But I sure can open one if you want me to. The PR #2384 is related, as described below (see **Related PR** section). |
| Primary OS tested on | Ubuntu 20.04, ROS 2 Galactic from binaries |
| Robotic platform tested on | Gazebo simulation of Wyca's robot, and real Wyca's robot |

## Description of the problem fixed by this PR

Hello, 
I fixed a problematic behavior that happens quite often after cancelling a `bt_action_node` (approximately once every 5 cancel requests in our case). 

*What we do*:
- We start a nav2 `BtNavigator`, which controls the action "GoToPose" for instance.
- The BT has several `bt_action_node`s, in order to call sub-actions like `"/follow_path"` for instance.
- When cancelling the "GoToPose" action, all the `bt_action_node`s execute [`halt()` function](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L265-L279), hence they send a cancel request to their respective sub-actions.

*What should happen*:
- The `bt_action_node` should wait for the sub-action to properly end, before returning from the `halt()` function.

*What actually happen*:
- The `bt_action_node` only [waits for the cancel request to return](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L269-L275).
- Hence, when cancelling the `BtNavigator`'s action, the sub-actions' `result_callback`s are not always executed.

## Related PR

* Before #2384 was merged, the next action call after a cancel often results in an instant end of the BT, because the goal result of the previously canceled sub-action is received. Indeed, cancelling the `BtNavigator`'s action halts the `bt_action_node` without waiting for the sub-action goal result. 

* After #2384, if the goal result of the previously cancelled sub-action is received, it is [filtered out](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L315-L321) using the `future_goal_handle_`. So, the next action call does not return instantaneously. But this might not be the cleaner way to solve the problem, since the sub-actions' `result_callback`s are still not always executed. 

## This PR's description

* This PR allows to wait for the sub-actions results, by making the [`bt_action_node` spin](https://github.com/wyca-robotics/navigation2/blob/04261fb7633c69baf709bae739310e37a3322812/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L285) until the sub-action goal result is available, after the cancel request is sent in the `halt()` function. This will force the `result_callback` to be executed after the sub-action is cancelled, allowing to get a clean `BtNavigator`'s action goal result.

* A [timeout](https://github.com/wyca-robotics/navigation2/blob/04261fb7633c69baf709bae739310e37a3322812/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L281) is implemented to prevent from infinitely looping while halting the `bt_action_node`.

* If the sub-action goal result is not available within the timeout, there is an [error output in logs](https://github.com/wyca-robotics/navigation2/blob/04261fb7633c69baf709bae739310e37a3322812/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L288-L292).

* Maybe the timeout can be handled differently
* The case "the goal result is not received after the timeout" may be handled (in a better way than just outputting an error in logs)

Thank you for reviewing this PR.

----

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -275,6 +275,22 @@ public:
       }
     }
 
+    auto start_time = node_->now();
+    while (rclcpp::ok()
+           && !goal_result_available_
+           && node_->now() - start_time < server_timeout_) {
+      RCLCPP_DEBUG(
+        node_->get_logger(),
+        "Waiting for goal result after cancelling, for action %s", action_name_.c_str());
+      callback_group_executor_.spin_some();
+      rclcpp::sleep_for(std::chrono::milliseconds(1));
+    }
+    if (!goal_result_available_) {
+      RCLCPP_ERROR_STREAM(
+        node_->get_logger(),
+        "Failed to get result for action " << action_name_.c_str() << " after cancelling. BT might not finish cleanly. ");
+    }
+
     setStatus(BT::NodeStatus::IDLE);
   }
 


SteveMacenski:
I agree with you in concept about this PR, I'm digging into the code a little to verify this for myself. One thing though is that this should be targeted to the `main` branch and we can then open a PR to the galactic branch once that is merged in. We target all new development to the `main` branch such that it can be reflected in all new distribution releases.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -275,6 +275,22 @@ class BtActionNode : public BT::ActionNodeBase
       }
     }
 
+    auto start_time = node_->now();
+    while (rclcpp::ok()
+           && !goal_result_available_
+           && node_->now() - start_time < server_timeout_) {
```
SteveMacenski, line 281: This part of the condition `node_->now() - start_time < server_timeout_` basically is just giving the cancel future only 1 more set of `server_timeout_` to get a cancel. The spin until future complete in L269 will block for `server_timeout_` waiting on this while loop will also block for another set of `server_timeout_` but is that really what you want to do? Its just giving that now 2 chances of that same length to do so, so you could still run into this issue if it exceeds that time frame. 

I don't think you actually really care about the `goal_result_available_` as much as you care that the `cancel_future` is completed, indicating that the server actually canceled successfully.

SteveMacenski: What is your `server_timeout` set to? We did increase this to I believe 20ms up from 10ms a few months ago when this logic for the server timeout was changed to enable the BT loop duration to move at a consistent rate

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
SteveMacenski: You mention 

> This will force the result_callback to be executed after the sub-action is cancelled, allowing to get a clean BtNavigator's action result.

and point to the `spin_some()`. So is some of this that you actually just want to spin to get a result callback more than anything else? If that' true, we could add a `spin_some()` to `halt()` after cancellation as well. However if the return code on the `future_cancel` is `SUCCESS` then we know it triggered so just 1 `spin_some()` should do the trick without having to do any of the rest of it.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -275,6 +275,22 @@ class BtActionNode : public BT::ActionNodeBase
       }
```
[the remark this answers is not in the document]
anaelle-sw: > When you run into issues, are you seeing that print?

No, we don't see it. 
Actually, the problem we would like to fix is 
**not** that the result of the **cancel** request is not available or has failed (which would be reported by *that print*, indeed), 
**but** that the result of the **goal** request sent to the sub-action is not always available before halting the whole BT. 
I may have not been so clear in the PR description, I will edit it right away in order to replace every "result" reference by either "goal result" or "cancel result".

> If the problem as you describe it is correct, you should be such that would return a non-success return code

I think there is a misunderstanding, because of the result we are discussing here: you mean the **cancel** result (right?), while I meant the **goal** result is *the problem*. (By *the problem*, I mean *when the **goal** result is not available*)
So, if the *non-success return code* you are talking about here is the return code of the result of the **cancel** request: well no, *the problem* happens even if the result of the cancel request is `SUCCESS`. 
Actually, in 100% of the occurrences of *the problem*, the result of the **cancel** request was `SUCCESS`.

>  I think given the code we're adding here, then that error should be reduced to a warning or a debug since we're going to be retrying below.

No, sorry I don't agree, we are not *retrying* anything in this `halt()` function. 
For me to be clearer, let's split the `halt()` function into 3 big steps:
- The first step (already implemented) consists in [sending the **cancel** request (L268)](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L268) and [checking for the result of this **cancel** request (L269)](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L269-L275).
- The second step (the subject of this PR) would consist in [checking for the result of the **goal** request (L279)](https://github.com/wyca-robotics/navigation2/blob/04261fb7633c69baf709bae739310e37a3322812/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L279-L287)
- Finally the third step (already implemented) simply consists in [applying the `IDLE` state to the BT node](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L278).

So the outcomes of the steps 1 and 2 are not related: the first one can end on a success and not the second one, or the opposite, or both can fail, or both can success.
The step 2 (this PR) is here to prevent from reaching step 3 without having the result of the **goal** request available.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -275,6 +275,22 @@ class BtActionNode : public BT::ActionNodeBase
...
     }
 
+    auto start_time = node_->now();
+    while (rclcpp::ok()
```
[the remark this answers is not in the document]
anaelle-sw: I strongly agree with you, the while loop could be simplified. However:

- I think need to keep the timeout in order to prevent infinite looping/spinning

- I would rather not run a *`spin_some()` blindly* too, but we can not use the `future_cancel` to do so. Indeed the variable `future_cancel` can only be used to check on the **cancel** result availability. 
To check **goal** result availability, one can use the variable [`goal_result_available_`](https://github.com/wyca-robotics/navigation2/blob/04261fb7633c69baf709bae739310e37a3322812/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L280) (which is a boolean variable already implemented, not a future).  As this variable is only updated while the `callback_group_executor_` is spinning, the `spin_some()` has been added here to check on the **goal** result availability.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
anaelle-sw: > This part of the condition `node_->now() - start_time < server_timeout_` basically is just giving the cancel future only 1 more set of `server_timeout_` 

Yes, that's right ...

> to get a cancel

... but no, not a cancel. This condition does not aim at getting the **cancel** result, but the **goal** result.
In my understanding, here is what happen when entering the `halt()` function:
1) First step of the `halt()` function, already implemented: The **cancel** request is sent to the sub-action. For a time period of maximum `server_timeout_`, we wait for the **cancel** result. Considering we have no trouble for getting the **cancel** future, there are two possibilities at the end of this first *set of `server_timeout_`*:
  - Case A: The **cancel** result and the **goal** result are available
  - Case B: Only the **cancel** result is result is available. Actually, in my opinion, this is not a exceptional case at all, since the sub-action can automatically accept the cancel requests (so the **cancel** result will be "instantaneously" available after sending the **cancel** request), but this same sub-action can take some time to properly end/stop the on-going **goal** after the cancel acceptation (so the **goal** result will not be available right after the **cancel** acceptation). I think it depends on the behavior you want your sub-action to have in case of a (successful) cancel.
  
2) Second step of the `halt()` function, object of this PR: We want to check for the **goal** result availability before ending the halt of the BT. This second step depends on the case obtained at step 1:
- If Case A: we don't even enter the while loop since the **goal** result is already available.
- If Case B: we enter the while loop. For a time period of maximum `server_timeout_`, we wait for the **goal** result.

So I think that waiting for maximum two *set of `server_timeout_`* for properly ending the sub-action in case of successful cancel is:
- neither only useful for isolated use cases,
- neither too much cautious / a possible slow down cause,

since it strongly depends on the desired behavior of the sub-action when it is being cancelled, and since it seems normal to me to wait for the sub-action to properly ends before halting the BT.

anaelle-sw: > What is your `server_timeout` set to? We did increase this to I believe 20ms up from 10ms a few months ago when this logic for the server timeout was changed to enable the BT loop duration to move at a consistent rate

In our use case, we set the `server_timeout` to 100 ms.

anaelle-sw:
> One thing though is that this should be targeted to the `main` branch and we can then open a PR to the galactic branch once that is merged in. We target all new development to the `main` branch such that it can be reflected in all new distribution releases.

Ok, sorry I didn't target the right branch. As we use Galactic at the moment, and our fork of this repo is based on the Galactic branch, I did think of doing this PR on the `master` branch, my bad.

Here is the PR #2440 targeted on `main` branch. 

Should we continue the discussion here or on the "main" PR?

anaelle-sw:
Both CI don't seem to be completely executed, this is the output:
```c
Build-agent version 1.0.76865-2132ecc2 (2021-07-05T15:58:20+0000)
System information:
 Server Version: 19.03.13
 Storage Driver: overlay2
  Backing Filesystem: xfs
 Cgroup Driver: cgroupfs
 Kernel Version: 4.15.0-1103-aws
 Operating System: Ubuntu 18.04.5 LTS
 OSType: linux
 Architecture: x86_64

Starting container rosplanning/navigation2:main.release
Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.
  image cache not found on this host, downloading rosplanning/navigation2:main.release

Error response from daemon: manifest for rosplanning/navigation2:main.release not found: manifest unknown: manifest unknown
```
I don't know how to solve this. Could you help me with it please?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
anaelle-sw: > So is some of this that you actually just want to spin to get a result callback more than anything else?

From a behavior point of view, what we *want more than anything else* is the sub-action to be cleanly finished before halting the BT. To know if it is indeed cleanly finished, we check on the availability of the sub-action **goal** result, since the **goal** result is available only once the sub-action has ended. Once the sub-action **goal** result is available, the goal result callback is automatically executed. So yes, saying that what we *want more than anything else* is the sub-action to be cleanly finished, is actually saying we *want to spin to get a result callback more than anything else*.

> However if the return code on the future_cancel is `SUCCESS` then we know it triggered so just 1 `spin_some()` should do the trick without having to do any of the rest of it.

No I don't agree with this, there is an explanation in this [previous comment](https://github.com/ros-planning/navigation2/pull/2436#discussion_r665448357)

> As far as I can tell from this PR there are 2 distinct issues you're trying to address (or just 1 and I'm misinterpreting for the second):
> 
>     * `cancel_future` doesn't complete in `server_timeout_` so you want to spin for a bit longer until it actually does successfully (or is that not an issue you're running into, with that error log?)
> 
>     * Because you don't spin after canceling, the callback for result isn't triggered in that same BT run so its messing up the state for the next run, so you're adding in a `spin_some()` to process the cancel result callback before exiting `halt()` to clean things up for the next run.
> If only (B) is your issue, no changes are required to the code above this block and this block can be largely simplified without the `while` to just be `spin_some()` since if `cancel_future` is successful, you're promised to have that waiting in your callback queue. If (A) is your issue, then we should work on combining what you have now with the while loop with spinning the future.

Well, I would said only (B) is an issue, but not exactly. I would reformulate it this way to make it accurate (changes are in bold):
Because you don't spin after canceling, the callback for **goal** result isn't triggered in that same BT run so its messing up the state for the next run, so you're adding in a `spin_some()` to process the **goal (Not cancel)** result callback before exiting `halt()` to clean things up for the next run.

So, I don't really agree with your conclusion about only (B) being an issue. If the `cancel_future` is successful, it doesn't mean the sub-action will have enough time to finish cleanly before halting the BT. It only means that the **cancel** request has been accepted. Which is great, but not enough, since the sub-action might not stop the current **goal** right after having accepting the **cancel** request.

SteveMacenski:
Lets continue on the `main` branch PR but I'll respond to your comments here. It might be good to just close this PR because I can just cherry pick this over to galactic once completed without a new PR. CI isn't quite running on the galactic branch yet, nothing you did. We've been experimenting with new CI technologies and docker environments so things are still in a state of flux.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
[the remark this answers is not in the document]
SteveMacenski: Got it, so it was problem (B) of the 2 I mention in the comments below? If so, then we can resolve this comment, no action required. This applies only to problem (A)

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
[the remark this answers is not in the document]
SteveMacenski: My assertion (which you could test and let me know if I'm wrong) is that when the `future_cancel` returns true after spinning, that you're **guaranteed** to have a result message waiting in your callback queue for a `spin_some()` to pick up. If that is accurate, then there is no need for a `while()` loop at all, therefore no requirement for a timeout from that while loop.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
SteveMacenski: Got it -- see comment in the thread above. I think we're on the same page, but I think we can simplify this if we leverage some internal knowledge about how the actions work. If that turns out not to be true, then we can discuss what to do so we can make sure we process the goal callback

PR #2440, opened 2021-07-07
Wait for goal result after cancelling BT action node

anaelle-sw:
## Basic Info

The PR #2436 was previously opened.
This previous PR #2436 does not target the `main` branch (as it should have), but the `galactic` branch. 
This PR here is exactly the same than #2436, but for the `main` branch

| ------ | ----------- |
| Ticket(s) this addresses | I didn't find any issue that is currently open, and that would be related to this PR. But I sure can open one if you want me to. The PR #2384 is related, as described below (see **Related PR** section). |
| Primary OS tested on | Ubuntu 20.04, ROS 2 Galactic from binaries |
| Robotic platform tested on | Gazebo simulation of Wyca's robot, and real Wyca's robot |

## Description of the problem fixed by this PR

Hello, 
I fixed a problematic behavior that happens quite often after cancelling a `bt_action_node` (approximately once every 5 cancel requests in our case). 

*What we do*:
- We start a nav2 `BtNavigator`, which controls the action "GoToPose" for instance.
- The BT has several `bt_action_node`s, in order to call sub-actions like `"/follow_path"` for instance.
- When cancelling the "GoToPose" action, all the `bt_action_node`s execute [`halt()` function](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L265-L279), hence they send a cancel request to their respective sub-actions.

*What should happen*:
- The `bt_action_node` should wait for the sub-action to properly end, before returning from the `halt()` function.

*What actually happen*:
- The `bt_action_node` only [waits for the cancel request to return](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L269-L275).
- Hence, when cancelling the `BtNavigator`'s action, the sub-actions' `result_callback`s are not always executed.

## Related PR

* Before #2384 was merged, the next action call after a cancel often results in an instant end of the BT, because the goal result of the previously canceled sub-action is received. Indeed, cancelling the `BtNavigator`'s action halts the `bt_action_node` without waiting for the sub-action goal result. 

* After #2384, if the goal result of the previously cancelled sub-action is received, it is [filtered out](https://github.com/ros-planning/navigation2/blob/29c1af087afffb80ddc4ec172015fee2c8a37166/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L315-L321) using the `future_goal_handle_`. So, the next action call does not return instantaneously. But this might not be the cleaner way to solve the problem, since the sub-actions' `result_callback`s are still not always executed. 

## This PR's description

* This PR allows to wait for the sub-actions results, by making the [`bt_action_node` spin](https://github.com/ros-planning/navigation2/blob/13e488392560cd746b94256822f9fe6afe261b25/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L285) until the sub-action goal result is available, after the cancel request is sent in the `halt()` function. This will force the `result_callback` to be executed after the sub-action is cancelled, allowing to get a clean `BtNavigator`'s action goal result.

* A [timeout](https://github.com/ros-planning/navigation2/blob/13e488392560cd746b94256822f9fe6afe261b25/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L281) is implemented to prevent from infinitely looping while halting the `bt_action_node`.

* If the sub-action goal result is not available within the timeout, there is an [error output in logs](https://github.com/ros-planning/navigation2/blob/13e488392560cd746b94256822f9fe6afe261b25/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L288-L292).

* Maybe the timeout can be handled differently
* The case "the goal result is not received after the timeout" may be handled (in a better way than just outputting an error in logs)

Thank you for reviewing this PR.

----

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -275,6 +275,22 @@ public:
       }
     }
 
+    auto start_time = node_->now();
+    while (rclcpp::ok()
+           && !goal_result_available_
+           && node_->now() - start_time < server_timeout_) {
+      RCLCPP_DEBUG(
+        node_->get_logger(),
+        "Waiting for goal result after cancelling, for action %s", action_name_.c_str());
+      callback_group_executor_.spin_some();
+      rclcpp::sleep_for(std::chrono::milliseconds(1));
+    }
+    if (!goal_result_available_) {
+      RCLCPP_ERROR_STREAM(
+        node_->get_logger(),
+        "Failed to get result for action " << action_name_.c_str() << " after cancelling. BT might not finish cleanly. ");
+    }
+
     setStatus(BT::NodeStatus::IDLE);
   }
 


anaelle-sw:
Hello @SteveMacenski, as you suggested, let's continue the discussion here for this PR.
*First part of the discussion is on another closed PR #2436 (the same than this PR but which targeted the wrong branch)*

I would like answer to this [comment](https://github.com/ros-planning/navigation2/pull/2436#discussion_r666439166):
> My assertion (which you could test and let me know if I'm wrong) is that when the `future_cancel` returns true after spinning, that you're guaranteed to have a result message waiting in your callback queue for a `spin_some()` to pick up. That might be a subtle detail I picked up from reading the `rclcpp_action` codebase. If that is accurate, then there is no need for a `while()` loop at all, therefore no requirement for a timeout from that while loop.
>
> That still doesn't help you if the `cancel_future` failed, but that's not the issue we're addressing here and you're screwed either way :smile:

I don't agree with you on that sentence: "when the `future_cancel` returns true after spinning, that you're guaranteed to have a result message waiting in your callback queue for a spin_some() to pick up". I was able to test it with two example nodes:

First node is an action client that send a goal to a client, and after 5 seconds, cancel the goal. Once the cancel sent, the node will wait for the cancel to be accepted, and wait for the goal result:
```c
#include <chrono>
#include <functional>
#include <memory>
#include <string>

#include <rclcpp/rclcpp.hpp>
#include <rclcpp_action/rclcpp_action.hpp>
#include <ros2_examples/action/fibonacci.hpp>

using Fibonacci = ros2_examples::action::Fibonacci;
using GoalHandleFibonacci = rclcpp_action::ClientGoalHandle<Fibonacci>;

class CallbackGroupActionClient: public rclcpp::Node {
public:
  // Constructor
  CallbackGroupActionClient()
  : Node("callback_goup_action_client"){
    // Init callback group
    callback_group_ = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive, false);
    callback_group_executor_.add_callback_group(callback_group_, this->get_node_base_interface());

    // Init client
    action_client = rclcpp_action::create_client<ros2_examples::action::Fibonacci>(
              this->get_node_base_interface(),
              this->get_node_graph_interface(),
              this->get_node_logging_interface(),
              this->get_node_waitables_interface(),
              "action_fibo",
              callback_group_);

    RCLCPP_INFO_STREAM(this->get_logger(), "Node created! ");
    std::cout<<std::endl;
  }

  // Destructor
  ~CallbackGroupActionClient(){
    callback_group_executor_.cancel();
  }

  // Methods
  void result_callback(const GoalHandleFibonacci::WrappedResult& result){
    goal_result_available_ = true;
    RCLCPP_INFO_STREAM(this->get_logger(), "Goal received!");
    std::cout<<std::endl;
  }

  void send_goal_and_cancel(){
    // Check if server available
    if (!action_client->action_server_is_ready()) {
      RCLCPP_WARN_STREAM(this->get_logger(), "Action server not available, skipping call");
      std::cout<<std::endl;
      return;
    }

    // Init goal options
    auto send_goal_options = typename rclcpp_action::Client<Fibonacci>::SendGoalOptions();
    send_goal_options.result_callback = std::bind(&CallbackGroupActionClient::result_callback, this, std::placeholders::_1);

    // Init goal
    goal_result_available_ = false;
    auto goal = std::make_shared<ros2_examples::action::Fibonacci::Goal>();
    goal->order = 100;

    // Send goal
    RCLCPP_INFO_STREAM(this->get_logger(), "Sending goal");
    std::cout<<std::endl;
    auto goal_future = action_client->async_send_goal(*goal, send_goal_options);

    // Cancel goal
    rclcpp::sleep_for(std::chrono::seconds(5)); // Waiting 5 seconds before cancelling goal
    RCLCPP_INFO_STREAM(this->get_logger(), "Cancelling goal");
    std::cout<<std::endl;
    auto cancel_future = action_client->async_cancel_all_goals();

    // Wait until goal result is available
    bool cancel_success = false;
    while (!goal_result_available_ && rclcpp::ok()) {
      RCLCPP_INFO_STREAM(this->get_logger(), "Goal result not available yet... ");
      if (!cancel_success) {
        cancel_success =
            callback_group_executor_.spin_until_future_complete(cancel_future, std::chrono::seconds(1))
            == rclcpp::FutureReturnCode::SUCCESS;
      }
      else {
        rclcpp::sleep_for(std::chrono::seconds(1));
      }
      RCLCPP_INFO_STREAM(this->get_logger(), "Cancel is a success?: " << cancel_success);
      std::cout<<std::endl;
    }
    return;
  }

  // Attributes
  rclcpp_action::Client<Fibonacci>::SharedPtr action_client;
  bool goal_result_available_ = false;
};



int main(int argc, char * argv[]){
  rclcpp::init(argc, argv);

  auto ac = std::make_shared<CallbackGroupActionClient>();

  std::cout << "Begin spinning" << std::endl;
  std::cout<<std::endl;
  while (rclcpp::ok()) {
    ac->send_goal_and_cancel();
  }

  rclcpp::shutdown();
  return 0;
}
```

The second is the server for this action. When a cancel request is received, the node will wait for 5 seconds before accepting the cancel request. Once the cancel request is accepted, the node wait for another 5 seconds before returning the result:
```c
#include <chrono>
#include <functional>
#include <memory>
#include <string>

#include <rclcpp/rclcpp.hpp>
#include <rclcpp_action/rclcpp_action.hpp>
#include <ros2_examples/action/fibonacci.hpp>

using Fibonacci = ros2_examples::action::Fibonacci;
using GoalHandleFibonacci = rclcpp_action::ServerGoalHandle<Fibonacci>;

class CallbackGroupActionServer: public rclcpp::Node {
public:
  // Constructor
  CallbackGroupActionServer()
  : Node("callback_goup_action_server"){
    // Init action server and link it to callback group
    action_server_ = rclcpp_action::create_server<Fibonacci>(
      this->get_node_base_interface(),
      this->get_node_clock_interface(),
      this->get_node_logging_interface(),
      this->get_node_waitables_interface(),
      "action_fibo",
      std::bind(&CallbackGroupActionServer::handle_goal, this, std::placeholders::_1, std::placeholders::_2),
      std::bind(&CallbackGroupActionServer::handle_cancel, this, std::placeholders::_1),
      std::bind(&CallbackGroupActionServer::handle_accepted, this, std::placeholders::_1));

    RCLCPP_INFO_STREAM(this->get_logger(), "Node created! ");
  }

  // Destructor
  ~CallbackGroupActionServer(){
  }

private:
  // Methods
  rclcpp_action::GoalResponse handle_goal(const rclcpp_action::GoalUUID & uuid,  std::shared_ptr<const Fibonacci::Goal> goal){
    RCLCPP_INFO(this->get_logger(), "Received goal request with order %d", goal->order);
    (void)uuid;
    return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
  }

  rclcpp_action::CancelResponse handle_cancel(const std::shared_ptr<GoalHandleFibonacci> goal_handle){
    RCLCPP_INFO(this->get_logger(), "Received request to cancel goal (will take 5 seconds...)");
    std::cout<<std::endl;
    (void)goal_handle;
    rclcpp::sleep_for(std::chrono::seconds(5)); // Waiting 5 seconds before accepting the cancel request
    RCLCPP_INFO(this->get_logger(), "Will cancel goal");
    std::cout<<std::endl;
    return rclcpp_action::CancelResponse::ACCEPT;
  }

  void handle_accepted(const std::shared_ptr<GoalHandleFibonacci> goal_handle) {
    std::thread{std::bind(&CallbackGroupActionServer::executeFibonacciAction, this,std::placeholders::_1), goal_handle}.detach();
    return;
  }
  void executeFibonacciAction(const std::shared_ptr<GoalHandleFibonacci> goal_handle) {
    // Executing action
    RCLCPP_INFO(this->get_logger(), "Executing goal");
    std::cout<<std::endl;
    auto result = std::make_shared<Fibonacci::Result>();
    result->sequence.push_back(5);
    result->sequence.push_back(3);

    for (int i = 1; (i < goal_handle->get_goal()->order) && rclcpp::ok() ; i++) {
      // Compute fibo
      RCLCPP_INFO_STREAM(this->get_logger(), "Computing order " << i);
      std::cout<<std::endl;
      result->sequence.push_back(result->sequence[i] + result->sequence[i-1]);

      // Check if there is a cancel request
      if (goal_handle->is_canceling()) {
        RCLCPP_INFO(this->get_logger(), "Cancelling goal (will take 5 seconds...)");
        std::cout<<std::endl;
        rclcpp::sleep_for(std::chrono::seconds(5)); // Waiting 5 seconds before returning result
        RCLCPP_INFO(this->get_logger(), "Goal is cancelled!");
        goal_handle->canceled(result);
        return;
      }
      rclcpp::sleep_for(std::chrono::seconds(1)); // Waiting 1 second between iterations
    }
  }


  // Attributes
  rclcpp_action::Server<ros2_examples::action::Fibonacci>::SharedPtr action_server_;
};



int main(int argc, char * argv[]){
  rclcpp::init(argc, argv);

  auto as = std::make_shared<CallbackGroupActionServer>();

  std::cout << "Begin spinning" << std::endl;
  std::cout<<std::endl;
  rclcpp::spin(as);

  rclcpp::shutdown();
  return 0;
}
```

If you launch these two nodes in two separated terminals, for the client you will get for the first loop:
```c
[INFO] : Node created! 

Begin spinning

[INFO] : Sending goal
       
[INFO] : Cancelling goal
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 0
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 0
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 0
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 0
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Cancel is a success?: 1
       
[INFO] : Goal result not available yet... 
[INFO] : Goal received!
       
[INFO] : Cancel is a success?: 1
```
So it means that after the client successfully receives the `cancel_future` status, waiting another 5 seconds is necessary to get the goal result anyway, since the server design is "wait 5 seconds before return the result after an accepted cancel request". Here it takes ~6 more `spin_some()` to get the result once the cancel is accepted. So we are not "guaranteed to have a result message waiting in your callback queue for a `spin_some()` to pick up". 

Now, with a more complex and realistic use case, an action server may come with:
- design errors (example: you forgot to abort/cancel/succeed the goal before returning, so no result will be ever return to the client)
- undermined delay for returning the result after the cancel is accepted (example: the server just made a service call before the cancel request, and you want this server to process the service response before it returns the goal result)

For these two main reasons, I think we need to `while()` loop with a timeout in order for the client to get the goal result, but not make it stuck for ever while waiting for the goal result.

SteveMacenski:
Hi, this hasn't been forgotten, just a little busy at the moment and this is something that needs some time for me to experiment / think. I'll try to get back to you before the end of this week.

SteveMacenski:
thanks for holding on -- continuing today

SteveMacenski:
To summarize that, I think then what I was seeing was that after you receive the spinning is done, that means the goal cancel request was **accepted**, but not that it was finished canceling -- got it. The semantics that the action server has to accept the cancel request before `is_canceling()` returns true to enact the cancel itself in the action is bizarre to me. Makes sense for issuing goal requests, but cancels I would just assume would always be welcome to process. Anyhow, that's here nor there. We're on the same page now. There's an argument to be made then we should be using the cancel handle callback better so that the bulk of the cancel work happens there, but that would be a near total redesign of this work. Anyway.

---

Another question worth asking is if there's something we can do in the result callback to mitigate any such issues so that we don't have to wait for that result to come back. We mitigate some of those issues with https://github.com/ros-planning/navigation2/blob/13e488392560cd746b94256822f9fe6afe261b25/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L331-L337 and checking if the goal IDs match https://github.com/ros-planning/navigation2/blob/13e488392560cd746b94256822f9fe6afe261b25/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L342. Is there a way to adjust this callback or another flag we can add / check? That would be overall my preference for reasons I outline in the section below. 

---

Your 2 examples are indeed examples of situations where the some additional waiting is necessary if there is no way for the result callback to be modified to better handle these situations. The major concern I have is of the timeout primarily. The `server_timeout_` is meant to be a timeout based on networking-style delays or retries in a patchy network / low bandwidth. But as you mention, the causes of delays here may not be caused by networking delay but actual time passed while an algorithm finishes up its current task to return cancelled -- or poorly implemented not respecting it. That's an entirely different issue that probably needs its own timeout setting based on the algorithms in play. While that's possible to add, it would it be a headache. 

As such, I'd prefer right now to try to solve this by making the result callback / the BT action client node more "stateless" between runs so that there are no issues even if we don't process the last goal completely. We've made good strides there with resetting futures / checking goal ids / etc but if this is an issue you're running into needing a solution, clearly there's still something not being properly reset locally at a `halt()` such that we just throw away any old request's future callback triggers / future updates.

SteveMacenski:
We're on the same page on the problem statement. I think I've seen that behavior myself in Rviz that I chalked up (wrongly) to DDS being nutty and dropping messages on my crappy laptop. I didn't think that it would be something anyone was actually experiencing in the field, or else someone would say something. Well, you've now said something :laughing: 

---

> But the lines you are referring to just do not allow to start a new BT if there is "something wrong". But this "something wrong" is coming from the BT before. 

This part I don't understand. It would throw out old results, why would that prevent a new BT request? 

I totally understand the idea of wanting to prevent things from going wrong so that its not an issue in the first place. That's good and typically I'd agree with you but in this case, there's no way to know how long to wait for a timeout since it's up to user code that could be either really good or really bad. 

As such, the only way to make sure that it would work even if the users mess things up is to make sure that this can handle incomplete results and allow for new ones to process. That doesn't mean we can't do both though, make a last-ditch attempt at getting a result in the `halt()` sounds good to me, but I don't think we can singularly rely on that to fix our issues since the `server_timeout_` variable is not meant for systemic delays due to user code, only DDS networking delays. So using that here would solve _some_ of the problem, but not _all_ of it, and I am looking for complete solutions. 

That's why I steer towards the question: 'what specifically is not being properly reset / setup for the next run in this case' so we can make sure things are bullet proof from bad-`halts()`. 

So tl;dr
- Yes, lets add a wait at the end of halt for a result for a "reasonable" amount of time. For now, we can use `server_timeout_` as a placeholder. If we do Bullet 2, then we can probably just keep it this way and it will still catch most situations.
- Lets identify what's specifically going wrong on the second run after this situation happens to make sure we can process goals even if the wait at the end of `halt()` is not enough. Even if the timeout was 100s, bad user code could just never implement the cancel, so we need to be able to handle this situation regardless of Bullet 1. 

---

We need a timeout, that's part of the Nav2 design that it is reliable and can handle faults like that. Also, no hardcoded numbers. If we have something, its going to be a parameter.

anaelle-sw:
Nice, we really are on the same page now :) 

---------

> dropping messages on my crappy laptop

Just to let you know, the situation **seems** to happen more frequently as the CPU usage goes up, **in our case**. I don't know if going in this direction can help to identify the origin of the problem.

--------

>> But the lines you are referring to just do not allow to start a new BT if there is "something wrong". But this "something wrong" is coming from the BT before.

> This part I don't understand. It would throw out old results, why would that prevent a new BT request?

Yes, it was not clear at all. I was meaning that it would check for the variables values, so the BT is not "allowed to start immediately". I agree, no BT request is ever prevented because of this situation.

---------

I totally agree with both your last bullet points. Would you rather merge this PR unchanged, and we open a new issue/PR to solve the second point? Or you want us to solve it in this PR? 

At the moment, I don't have the time necessary to identify _'what specifically is not being properly reset / setup for the next run in this case'_ . So I would rather go with merging this PR (since it **partially** solves the problem, as you said), and opening a new issue to document the second point. I sure will help, but when I can spend time on it

SteveMacenski:
> Just to let you know, the situation seems to happen more frequently as the CPU usage goes up, in our case. I don't know if going in this direction can help to identify the origin of the problem.

That's a rational observation. That's what most folks are finding with ROS2 / DDS. What DDS vendor are you using, out of curiosity? It is an issue I believe in all of them, overall though we have had more positive feedback on average around Cyclone for Nav2 use. 

> I totally agree with both your last bullet points. Would you rather merge this PR unchanged, and we open a new issue/PR to solve the second point? Or you want us to solve it in this PR?

I think we need a full solution or at least something more than just this. We can get away with `server_timeout_` if we have a solution to working well. If we have no such solution, then `server_timeout_` is insufficient because it doesn't represent the right 'thing' and would have to be tuned separately from networking related delays in accordance with the user code in play.

anaelle-sw:
> What DDS vendor are you using, out of curiosity? 

We are using Cyclone DDS.
I tested it today, and I can confirm the issue is clearly more frequent with a stressed CPU. In my case, it happens automatically when more than half of the CPUs are used at 100%.

I understand the "timeout problem", and also that you rather merge a real fix.

> I'd argue that its really worth your time at Wyca to solve this problem because this only actually fixes the ~1 standard deviation of the issue. Now rather than happening 1/5 times, now it might happen 1/50 times which is still every single day with even a small fleet. 

I agree, but I cannot just guarantee that this issue will be fixed quickly if only depending on the time I can spend on it, but I will be trying my best :)

Actually, I tried today to understand what's happening. But the only conclusion I can make for now, is that forcing to reset all action-client related variables in the `halt()` function, like this:
```c
goal_ =  typename ActionT::Goal();
goal_handle_.reset();
future_goal_handle_.reset();
result_ = typename rclcpp_action::ClientGoalHandle<ActionT>::WrappedResult();
```
is not helpful at all. So maybe, it is not related to the action client but the action server. I am checking this, and will let you know when finding something interesting.

SteveMacenski:
Huh, I really would have thought it would be due to that. Is it just because the result callback is called _at all_?

SteveMacenski:
Any update?

anaelle-sw:
Hello,
I have some "disappointing" news: I am no longer able to reproduce the issue... Even when getting back to previous versions of nav2 where the issue happened for sure, and checking that the DDS version/implementation is the same than before, and stressing all CPUs... I just cannot make it happen anymore. 
Sorry, I won't be able to know if it is getting any better or worse when trying things in order to solve this issue, until it is possible to reproduce the problem again. 
Maybe we can close this PR if you agree. If anyone encounter this and open an issue about it, I would be happy to help then. But for the moment, there is not much I can do about it :sweat:

anaelle-sw:
> Was there a ticket this was linked to? I can't seem to find it.

I directly opened a PR. I looked for potential related issues before, but didn't find any.

> The 2 PRs in the action side of things that changed

Thanks for mentioning those PRs. I tried to revert both (while still making sure to have the same DDS version and nav2 version than before), but I cannot reproduce the issue (even with stressed CPUs). 

> it was probably handled upstream

You may be right. The thing is that even when building rclcpp from sources and checking out an "old" tag (the one that corresponds to the debian sync we were using when I firstly opened this PR ([9.1.0 tag on galactic branch](https://github.com/ros/rosdistro/releases/tag/galactic%2F2021-05-23))), I am still unable to reproduce. I don't get what's different from then, this is really frustrating...

SteveMacenski:
Sorry about that :frowning: but sounds like we can close this for now then. Feel free to reopen it or comment back if it rears its head again and we can deal with it quickly then!

Steve

PR #2444, opened 2021-07-11
State lattice

jwallace42:
---

## Basic Info

| ------ | ----------- |

This does not address any tickets, just work on the lattice planner. 
Created a new unit test to verify that the parser is working correctly. 

---

Added in functionality to parse in json file of motion primitives to cpp structs

None

---

- Converting motion primitive coordinates to cells and bins 
- Converting the start and goal pose to bins and cells in the smac lattice planner 
- getting projections out of the motion table 

#### For Maintainers: 


nav2_smac_planner/test/test_node_lattice.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "nav2_smac_planner/node_lattice.hpp"
+#include "gtest/gtest.h"
+#include <fstream>
+
+using json = nlohmann::json;
+
+TEST(ParserTest, test_lattice_node)
+{
+    //TODO: Where should the lattice file be placed? 
+    std::string filePath = "/home/josh/nav2_ws/src/navigation2/nav2_smac_planner/test/output.json";
```
SteveMacenski, line 24: if its in nav2_smac_planner, you can use the ament to find the packages' shared path https://docs.ros2.org/foxy/api/ament_index_cpp/ if you install that output file somewhere it can retrieve it

nav2_smac_planner/test/test_node_lattice.cpp
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2021, Samsung Research America
```
SteveMacenski, line 1: This is probably your copyright header :wink:

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -25,6 +25,7 @@
 #include <utility>
 #include <limits>
 #include <string>
+#include <nlohmann/json.hpp>
```
SteveMacenski, line 28: Please add this as a rosdep key to package.xml so that it is installed in the package manager

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -40,7 +41,37 @@ namespace nav2_smac_planner
 class NodeLattice;
 class NodeHybrid;
 
-typedef std::pair<unsigned int, double> LatticeMetadata;
+/**
+ * @struct nav2_smac_planner::LatticeMetadata
+ * @brief A struct of all lattice metadata
+ */
+struct LatticeMetadata
```
SteveMacenski, line 48: You can place the `struct` definitions here in the `types.hpp` file just to keep those all sorted in 1 place

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -40,7 +41,37 @@ namespace nav2_smac_planner
...
+{
+  float turningRadius;
```
SteveMacenski, line 50: Struct member variables should be `snake_case` not `camelCase` by convention in ROS styling. So structs/classes are `CapCap` and then members with in it are `low_low`.

SteveMacenski, line 50: Can we make this `min_turning_radius` to be clear?

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -69,6 +70,36 @@ void LatticeMotionTable::initMotionModel(
...
+    {
+      motionPrimitives.emplace_back(primitives);
+      primitives.clear(); 
+      prevStartAngle = newPrimitive.startAngle;
+    }
+    primitives.emplace_back(newPrimitive);
+  }
+  motionPrimitives.emplace_back(primitives); 
+ 
+  fromJsonToMetaData(j["latticeMetadata"], latticeMetadata);
```
SteveMacenski, line 101: Lets move the metadata first since in the file its also first (and I think generally it helps color the code to understand what's happening under it if we deal with the header first organizationally)

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -69,6 +70,36 @@ void LatticeMotionTable::initMotionModel(
 
   // TODO(Matt) populate num_angle_quantization, size_x, min_turning_radius, trig_values,
   // all of the member variables of LatticeMotionTable
+
+  std::ifstream latticeFile(current_lattice_filepath);
```
SteveMacenski, line 74: There's some more to be done if you look at my comment in https://github.com/ros-planning/navigation2/blob/8e837780fddab09dd6a39bcdf6f3cd31af3fd6cb/nav2_smac_planner/src/node_lattice.cpp#L66-L72. Basically we need to fill in the fields of the Lattice motion table https://github.com/ros-planning/navigation2/blob/8e837780fddab09dd6a39bcdf6f3cd31af3fd6cb/nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp#L80, other than `bin_size` which you can delete since that is not relevant to us right now. In place of `bin_size` we need to have a vector of angles, e.g. `vector<double>` of the length of the number of primitives so we can do something like: `bin_angles[theta]` to find the angle at a certain primitive number (to replace calculations like https://github.com/ros-planning/navigation2/blob/8e837780fddab09dd6a39bcdf6f3cd31af3fd6cb/nav2_smac_planner/src/node_lattice.cpp#L189)

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
jwallace42: Done

jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
jwallace42: Done

nav2_smac_planner/test/test_node_lattice.cpp
jwallace42: Done

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
jwallace42: Done

nav2_smac_planner/test/test_node_lattice.cpp
SteveMacenski: @jwallace42 by the way, you can find an example of this in `nav2_bt_navigator/src/navigators/navigate_to_pose.cpp` using `ament_index_cpp` so you don't have to search around. I was very happy to have found that API myself earlier this year.

jwallace42: Thanks for the example! I am happy you found it too!

jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done. However I did not know what type of state space to use.

nav2_smac_planner/include/nav2_smac_planner/types.hpp
```diff
@@ -87,6 +87,66 @@ struct SmootherParams
...
+/**
+ * @struct nav2_smac_planner::MotionPrimitive
+ * @brief A struct of all motion primitive data
+ */
+struct MotionPrimitive
+{
+  unsigned int trajectory_id;
+  float start_angle; 
+  float end_angle; 
+  float radius;
```
SteveMacenski, line 143: radius -> turning_radius

nav2_smac_planner/include/nav2_smac_planner/types.hpp
```diff
@@ -87,6 +87,66 @@ struct SmootherParams
...
+typedef std::vector<MotionPose> MotionPoses;
+
+/**
+ * @struct nav2_smac_planner::LatticeMetadata
+ * @brief A struct of all lattice metadata
+ */
+struct LatticeMetadata
+{
+  float min_turning_radius;
+  float step_distance;
```
SteveMacenski, line 125: step_distance -> primitive_resolution

grid_separation -> grid_resolution

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -86,15 +85,39 @@ struct LatticeMotionTable
   float num_angle_quantization_float;
   float min_turning_radius;
   float bin_size;
+  std::vector<float> primitive_start_angles;
```
SteveMacenski, line 88: primitive_start_angles -> primitive_headings

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -86,15 +85,39 @@ struct LatticeMotionTable
...
+  std::vector<std::vector<MotionPrimitive>> motionPrimitives;
+  LatticeMetadata latticeMetadata;
 };
 
+/**
+ * @brief convert json to lattice metadata 
+ * @param[in] j json object
+ * @param[out] lattice meta data 
+ */
+void fromJsonToMetaData(const nlohmann::json &j, LatticeMetadata &latticeMetaData);
```
SteveMacenski, line 105: Do these 3 function need to be inside of this object to access local variables? If not, I'd recommend moving them to the utils.hpp file since they're kind of utility-esk

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -81,9 +145,47 @@ LatticeMetadata LatticeMotionTable::getLatticeMetadata(const std::string & latti
...
+{
+  j.at("trajectoryId").get_to(motionPrimitive.trajectory_id);
+  j.at("startAngle").get_to(motionPrimitive.start_angle);
+  j.at("endAngle").get_to(motionPrimitive.end_angle);   
+  j.at("radius").get_to(motionPrimitive.radius);
+  j.at("trajectoryLength").get_to(motionPrimitive.trajectory_length);
+  j.at("arcLength").get_to(motionPrimitive.arc_length);
+  j.at("straightLength").get_to(motionPrimitive.straight_length);
+
+  for(auto& jsonPose : j["poses"])
```
SteveMacenski, line 180: just in styling for the rest of the package, can you do a C-style for loop instead?

nav2_smac_planner/src/node_lattice.cpp
SteveMacenski: Good question, lets set it to DUBIN for now and then I have a bullet on my list of tasks to figure out a way to set params for reversing. We can expand that then. Add a TODO for that in the code as a reminder

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -69,6 +98,41 @@ void LatticeMotionTable::initMotionModel(
 
   // TODO(Matt) populate num_angle_quantization, size_x, min_turning_radius, trig_values,
   // all of the member variables of LatticeMotionTable
+
+  num_angle_quantization = latticeMetadata.number_of_headings;
+  num_angle_quantization_float = static_cast<float>(num_angle_quantization);
+  min_turning_radius = latticeMetadata.min_turning_radius;
```
SteveMacenski, line 104: You can just straight up remove `bin_size` from this object since it doesn't make sense in our context with non-regular sizes

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -69,6 +98,41 @@ void LatticeMotionTable::initMotionModel(
...
+  }
+
+  //TODO: angles in json file need to be converted to radians 
+  for(auto const &start_angle : primitive_start_angles)
+  {
+    trig_values.emplace_back( cos(start_angle), sin(start_angle) );
+  }
 }
 
 MotionPoses LatticeMotionTable::getProjections(const NodeLattice * node)
```
SteveMacenski, line 138: So we have this function getProjections which is what we use to return the set of primitives that can be expanded from the given `node`. In the Hybrid-A* node, we use the `projections` variable but it looks like you instead populate a `motionPrimitives` object that meets basically the same thing.

So my recommendation would be to remove the `projections` variable since you're not using it and you can rename that function `getMotionPrimitives` to use your `motionPrimitives` object instead. You can now populate this function `getProjections` since you have all the information required. The starting node (so you know your starting heading) and the vector of motion primitives at each starting heading to return a `MotionPoses` of primitives. 

If you find that its hard to convert from motionPrimitives to the current `MotionPoses` object, you are welcome to replace `MotionPoses` with a new struct.

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -69,6 +98,41 @@ void LatticeMotionTable::initMotionModel(
...
+    {
+      if( motionPrimitive.radius < largestTurningRadius && motionPrimitive.radius != 0 )
+      {
+        largestTurningRadius = motionPrimitive.radius;
+      }
+    }
+  }
+
+  //NOTE: Note sure what type of state space this should be 
+  state_space = std::make_unique<ompl::base::DubinsStateSpace>( largestTurningRadius );
```
SteveMacenski, line 120: You don't need to find the `largestTurningRadius`, you should use what you already set above for this instead, `min_turning_radius` :wink: that should let you remove the above double nested for loop. The goal is to find the smallest allowable circular radius it can turn in, not the largest (since the "largest" can be infinitely large until its essentially a straight line)

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -59,6 +58,36 @@ void LatticeMotionTable::initMotionModel(
...
+      motionPrimitives.emplace_back(primitives);
+      primitives.clear(); 
+      prevStartAngle = newPrimitive.start_angle;
+    }
+    primitives.emplace_back(newPrimitive);
+  }
+  motionPrimitives.emplace_back(primitives); 
+
+
   // TODO(Matt) read in file, precompute based on orientation bins for lookup at runtime
```
SteveMacenski, line 91: These 3 TODOs can be removed now, you implemented them.

nav2_smac_planner/src/node_lattice.cpp
SteveMacenski: You can also populate the `getLatticeMetadata` function as well (and even use that function in your `init` functions too!)

SteveMacenski (changes requested):
Great! All coming together perfectly, just some comments to shuffle some code around and you've done enough here now that you can fill in the getProjections() and getLatticeMetadata() functions as well!

nav2_smac_planner/include/nav2_smac_planner/types.hpp
jwallace42: Done

nav2_smac_planner/include/nav2_smac_planner/types.hpp
jwallace42: Done

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done, I left a few questions in the code

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
jwallace42: Done, very clever use of inline. I though it was irrelevant because the compiler would handle _inlining_. It was pretty cool to learn that it would resolve functions that are defined twice within a translation unit.

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done, had to comment out a few lines in `getAnalyticPath()` that used the bin size.

nav2_smac_planner/src/a_star.cpp
```diff
@@ -545,10 +545,14 @@ typename AStarAlgorithm<NodeT>::AnalyticExpansionNodes AStarAlgorithm<NodeT>::ge
     node->motion_table.state_space), s(node->motion_table.state_space);
   from[0] = node->pose.x;
   from[1] = node->pose.y;
-  from[2] = node->pose.theta * node->motion_table.bin_size;
+  //TODO: Bin size removed from motion table for state lattice
```
SteveMacenski, line 548: Don't just comment it out -- that'll break every other planner :laughing: 

Instead do the following:
- Make a new function, getRadiansFromBin()
- Make an implementation for Hybrid using bin_size
- Make an implementation for 2D throwing an exception as invalid
- Make an implementation for Lattice by using the primitive headings vector

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -116,11 +188,13 @@ bool NodeLattice::isNodeValid(
 {
   // TODO(steve) if primitive longer than 1.5 cells, then we need to split into 1 cell
   // increments and collision check across them
-  if (collision_checker->inCollision(
-      this->pose.x, this->pose.y, this->pose.theta * motion_table.bin_size, traverse_unknown))
-  {
-    return false;
-  }
+
+  //TODO: removed bin size from motion table, not sutible for lattice planner
```
SteveMacenski, line 192: Same here, don't uncomment, use the primitives headings vector you have to get the orientation at a given `theta` quantization

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -59,29 +58,102 @@ void LatticeMotionTable::initMotionModel(
...
+      cos(primitive_headings[i]), 
+      sin(primitive_headings[i]) );
+  } 
 }
 
-MotionPoses LatticeMotionTable::getProjections(const NodeLattice * node)
+MotionPoses LatticeMotionTable::getMotionPrimitives(const NodeLattice * node)
 {
-  return MotionPoses();  // TODO(Matt) lookup at run time the primitives to use at node
+  //Question: Theta is in bin coords, 0-16 so theta does not have to be shifted?
```
SteveMacenski, line 118: There are some reasonable questions here:
- Should we be returning only the final point, or should we be returning the full primitive (e.g. each of the points making up the primitive translated to the given point)? That might help later for checking for collisions. 
- What are the coordinates of these primitives? Are they in cells or in meters?

SteveMacenski (changes requested):

nav2_smac_planner/src/a_star.cpp
SteveMacenski: See other functions in A* with template overrides for examples

jwallace42: Done. 
The only place I didn't know how to purse was `angle = reals[2] / node->motion_table.bin_size;` inside of `getAnalyticPath`. 
Do we still want to conduct analytic expansion with the lattice planner? I vaguely remember discussing that it way be left out for the first pass.

nav2_smac_planner/src/node_lattice.cpp
jwallace42: Done

nav2_smac_planner/src/node_lattice.cpp
jwallace42: I guess having each node store the primitive might be expensive. Also, each primitive would have to be shifted which would add computation. If we were to store the index of the motion primitive in the node we could then pass in the primitive and pose into the collision checker. 

Then we could potentially do some down-sampling within the collision check to reduce the number of points that we would need to check and shift. 

However, the longest primitive in the json file only has 14 points. So it might not make that much of a difference. 

The coordinates of the json file are in meters and radians. So I converted within the function. 

The primitives are in meters and radius, so

nav2_smac_planner/src/a_star.cpp
```diff
@@ -572,7 +572,9 @@ typename AStarAlgorithm<NodeT>::AnalyticExpansionNodes AStarAlgorithm<NodeT>::ge
   for (float i = 1; i < num_intervals; i++) {
     node->motion_table.state_space->interpolate(from(), to(), i / num_intervals, s());
     reals = s.reals();
-    angle = reals[2] / node->motion_table.bin_size;
+    //TODO: Bin size removed from motion table for state lattice
+    // angle = reals[2] / node->motion_table.bin_size;
+    angle = 0.0;
```
SteveMacenski, line 577: This needs to be converted from 0-2*pi (or maybe -pi to pi?) into an angular bin number to test collisions for. Basically, it needs the opposite of the getRadFromBin and instead getBinFromRad

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -80,19 +80,21 @@ struct LatticeMotionTable
...
+  float grid_resolution;
+  std::vector<float> primitive_headings;
   float change_penalty;
   float non_straight_penalty;
   float cost_penalty;
   float reverse_penalty;
   ompl::base::StateSpacePtr state_space;
   std::vector<TrigValues> trig_values;
   std::string current_lattice_filepath;
+  std::vector<std::vector<MotionPrimitive>> motionPrimitives;
```
SteveMacenski, line 96: motion_primitives

nav2_smac_planner/include/nav2_smac_planner/node_lattice.hpp
```diff
@@ -80,19 +80,21 @@ struct LatticeMotionTable
...
+  LatticeMetadata latticeMetadata;
```
SteveMacenski, line 97: lattice_metadata

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -59,29 +58,104 @@ void LatticeMotionTable::initMotionModel(
...
+  grid_resolution = latticeMetadata.grid_resolution; 
+
+  //TODO: Note sure what type of state space this should be 
+  state_space = std::make_unique<ompl::base::DubinsStateSpace>( min_turning_radius );
+
+  for(unsigned int i = 0; i < motionPrimitives.size(); ++i)
+  {
+    for(unsigned int j = 0; j < motionPrimitives[0].size(); ++j)
+    {
+      //TODO: angles in json file need to be converted to radians
```
SteveMacenski, line 103: This done yet?

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -59,29 +58,104 @@ void LatticeMotionTable::initMotionModel(
...
+      primitive_headings.push_back( motionPrimitives[i][j].start_angle );
+    }
+  }
 
-  // TODO(Matt) populate num_angle_quantization, size_x, min_turning_radius, trig_values,
-  // all of the member variables of LatticeMotionTable
+  //TODO: angles in json file need to be converted to radians
```
SteveMacenski, line 108: This done yet?

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -59,29 +58,104 @@ void LatticeMotionTable::initMotionModel(
...
+  const unsigned int branches = motionPrimitives[node_heading].size();
+  MotionPoses primitive_projection_list; 
+  primitive_projection_list.reserve( branches );
+
+  for( unsigned int i = 0; i < branches; ++i)
+  {
+    const MotionPose end_pose = motionPrimitives[node_heading][i].poses.back(); 
+
+
+    primitive_projection_list.emplace_back(
```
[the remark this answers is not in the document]
SteveMacenski: https://github.com/ros-planning/navigation2/blob/97c2ad55a04f4b78c944f2145166f144337dd802/nav2_smac_planner/src/node_hybrid.cpp#L241-L244 for Hybrid, though I know that the angle here might be different logic, is the final angle in `end_pose._theta` already taking into account the initial heading from `motionPrimitives[node_heading]`?

SteveMacenski:
Work is being manually inputted into state_lattice_2 branch due to the git mess I accidentally created and I don't want to make you deal with all my issues / blocking on small things I can just fix myself. I'll link you in the commit

PR #2448, opened 2021-07-13
add bool arg use_default_callback_group for ServiceClient

gezp:

---

## Basic Info

| Info                       | Please fill out this column               |
| -------------------------- | ----------------------------------------- |
| Ticket(s) this addresses   | #816 based on #2216                       |
| Primary OS tested on       | Ubuntu20.04, ROS 2 Galactic from binaries |
| Robotic platform tested on | None                                      |

---

In PR #2216，`callback_group` and `SingleThreadedExecutor` are created to spin_until_future_complete. they are necessary when we use the method `invoke` of `ServiceClient` in a callback (topic, timer, service) handle by the node, because we can't spin a node already spinning ([rclcpp/test/rclcpp/test_executor.cpp#L261-L266](https://github.com/ros2/rclcpp/blob/1037822a63330495dcf0de5e8f20544375a5f116/rclcpp/test/rclcpp/test_executor.cpp#L261-L266)).

 But in some cases, we don't use the method `invoke` of `ServiceClient` in callback, so we could avoid creating new `callback_group` and `SingleThreadedExecutor` in this cases.

* add a bool arg `use_default_callback_group`.
* when `use_default_callback_group` is true, create client by using default callback group, and use `wait_for()` to block `invoke()`.
* when `use_default_callback_group` is false, create new `callback_group` and `SingleThreadedExecutor`, just as before.

---

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/service_client.hpp
+++ b/nav2_util/include/nav2_util/service_client.hpp
@@ -34,20 +34,32 @@ public:
   * @brief A constructor
   * @param service_name name of the service to call
   * @param provided_node Node to create the service client off of
+  * @param use_default_callback_group Whether to use default callback group. when true,
+  * ServiceClient can't be invoked in ros callback.
   */
   explicit ServiceClient(
     const std::string & service_name,
-    const rclcpp::Node::SharedPtr & provided_node)
-  : service_name_(service_name), node_(provided_node)
+    const rclcpp::Node::SharedPtr & provided_node,
+    bool use_default_callback_group = false)
+  : service_name_(service_name), node_(provided_node),
+    use_default_callback_group_(use_default_callback_group)
   {
-    callback_group_ = node_->create_callback_group(
-      rclcpp::CallbackGroupType::MutuallyExclusive,
-      false);
-    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
-    client_ = node_->create_client<ServiceT>(
-      service_name,
-      rmw_qos_profile_services_default,
-      callback_group_);
+    if (use_default_callback_group_) {
+      client_ = node_->create_client<ServiceT>(
+        service_name,
+        rmw_qos_profile_services_default);
+    } else {
+      callback_group_ = node_->create_callback_group(
+        rclcpp::CallbackGroupType::MutuallyExclusive,
+        false);
+      callback_group_executor_.add_callback_group(
+        callback_group_,
+        node_->get_node_base_interface());
+      client_ = node_->create_client<ServiceT>(
+        service_name,
+        rmw_qos_profile_services_default,
+        callback_group_);
+    }
   }
 
   using RequestType = typename ServiceT::Request;
@@ -78,9 +90,23 @@ public:
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
 
-    if (callback_group_executor_.spin_until_future_complete(future_result, timeout) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+    bool success = true;
+    if (use_default_callback_group_) {
+      std::future_status status;
+      if (timeout < std::chrono::nanoseconds::zero()) {
+        // Block forever until future completed
+        do {
+          status = future_result.wait_for(std::chrono::seconds(1));
+        } while (status != std::future_status::ready);
+      } else {
+        status = future_result.wait_for(timeout);
+      }
+      success = (status == std::future_status::ready);
+    } else {
+      auto status = callback_group_executor_.spin_until_future_complete(future_result, timeout);
+      success = (status == rclcpp::FutureReturnCode::SUCCESS);
+    }
+    if (!success) {
       throw std::runtime_error(service_name_ + " service client: async_send_request failed");
     }
 
@@ -112,9 +138,19 @@ public:
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
 
-    if (callback_group_executor_.spin_until_future_complete(future_result) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+    bool success = true;
+    if (use_default_callback_group_) {
+      std::future_status status;
+      // Block forever until future completed
+      do {
+        status = future_result.wait_for(std::chrono::seconds(1));
+      } while (status != std::future_status::ready);
+    } else {
+      auto status = callback_group_executor_.spin_until_future_complete(future_result);
+      success = (status == rclcpp::FutureReturnCode::SUCCESS);
+    }
+
+    if (!success) {
       return false;
     }
 
@@ -137,6 +173,7 @@ protected:
   rclcpp::Node::SharedPtr node_;
   rclcpp::CallbackGroup::SharedPtr callback_group_;
   rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
+  bool use_default_callback_group_;
   typename rclcpp::Client<ServiceT>::SharedPtr client_;
 };
 
--- a/nav2_util/include/nav2_util/service_client.hpp
+++ b/nav2_util/include/nav2_util/service_client.hpp
@@ -16,7 +16,7 @@
 #define NAV2_UTIL__SERVICE_CLIENT_HPP_
 
 #include <string>
-
+#include <memory>
 #include "rclcpp/rclcpp.hpp"
 
 namespace nav2_util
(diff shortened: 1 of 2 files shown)


nav2_util/test/test_service_client.cpp
```diff
@@ -87,6 +87,30 @@ TEST(ServiceClient, can_ServiceClient_invoke_in_callback)
   ASSERT_EQ(a, 1);
 }
 
+TEST(ServiceClient, can_ServiceClient_use_default_callback_group)
```
SteveMacenski, line 90: I appreciate that you wrote a test for this, thank you

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -137,6 +173,7 @@ class ServiceClient
   rclcpp::Node::SharedPtr node_;
   rclcpp::CallbackGroup::SharedPtr callback_group_;
   rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
```
SteveMacenski, line 175: The executor should be made into a unique pointer. The executor's constructor will call the abstract executor's constructor https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executor.cpp which does a bunch of set up and connects to the network. We'd like to avoid that if possible unless its actually going to be used. So if its a unique pointer, then it will be `nullptr` and not create an object until you create one `make_unique` when it may be used.

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -78,9 +90,23 @@ class ServiceClient
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
 
-    if (callback_group_executor_.spin_until_future_complete(future_result, timeout) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+    bool success = true;
+    if (use_default_callback_group_) {
+      std::future_status status;
```
SteveMacenski, line 95: Why doesn't the `if` statement just mirror the `else` statement's logic, just using the `node_` to spin rather than the `callback_group_executor_`? You should also still use the timeout.

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -112,9 +138,19 @@ class ServiceClient
...
 
-    if (callback_group_executor_.spin_until_future_complete(future_result) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+    bool success = true;
+    if (use_default_callback_group_) {
+      std::future_status status;
+      // Block forever until future completed
+      do {
+        status = future_result.wait_for(std::chrono::seconds(1));
```
SteveMacenski, line 146: As well here, we should just mirror using the node to spin until complete

SteveMacenski (changes requested):

nav2_util/include/nav2_util/service_client.hpp
gezp: we can't use `rclcpp::spin_until_future_complete(node_,future_result,timeout)` , look at https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/executors.hpp#L103-L112, it will create SingleThreadedExecutor . 
when `use_default_callback_group` is true, the default callback group will spin in thread, and we only need wait to use invoke.

gezp: we can't use rclcpp::spin_until_future_complete(node_,future_result,timeout) , look at https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/executors.hpp#L103-L112, it will create SingleThreadedExecutor .
when use_default_callback_group is true, the default callback group will spin in thread, and we only need wait to use invoke.

gezp:
This PR need more discussion，because we want to avoid creating new  `SingleThreadedExecutor` .  `rclcpp::spin_some()`  and `rclcpp::spin_until_future_complete()`  for Node will create new  `SingleThreadedExecutor` within function, so we should avoid using 
 `rclcpp::spin_some()`  and  `rclcpp::spin_until_future_complete()` for `Node`. Anyway,  `ServiceClient `  is more general,can used in any cases.

gezp:
I have a new idea , which uses new arg `use_internal_executor` instead of `use_default_callback_group`

* `use_internal_executor` is true, create new `callback_group` (`automatically_add_to_executor_with_node` is false) and `SingleThreadedExecutor`, use `callback_group_.spin_until_future_complete()` to wait result. (spin and wait)
  * it's general method, which works for any cases.
* `use_internal_executor` is false, only create new `callback_group` (`automatically_add_to_executor_with_node` is true) , use `wait_for()` to wait result (only wait)
  * we should use external  `Excutor`  to spin Node. ( `SingleThreadedExecutor` or `MultiThreadedExecutor`  in a thread )
  * we must use `MultiThreadedExecutor` in `main()`  when we want to use `invoke()` of `ServiceClient ` in callback.

> we can use `invoke()` of `ServiceClient ` in callback whatever `use_internal_executor` is true or false.

nav2_util/include/nav2_util/service_client.hpp
SteveMacenski: This is indeed something I had not thought about before. I did a brief survey of Nav2's use of spin until future is complete and was shocked at the results https://github.com/ros-planning/navigation2/search?p=1&q=spin_until_future_complete. Looking at this list **its not used in Nav2 at runtime**, all places are tests, python utilities, or the rviz panel. The rest have specified executors. That's a good sign, but also really unexpected. 

---

Thinking back to how we got here, this was under the impression that we could spin our input node rather than needing to create a new callback group to `spin_until_future_complete`. If we don't actually need to spin anything (e.g. we can just do some logic based on the future itself) then the entire idea of having a separate executor / callback group is largely moot, isn't it? I'm not 100% sure under the hood how it is that the future would actually get the updates in the service client with the above code if you never spin though and the node isn't spinning in the background, which was the core assumption of this method. If the services are in different processes, there's no possible way to have the client receive its result without spinning somewhere.

--- 

Also, I think you're looking at a different implementation of `spin_until_future_complete`. Here's the version I think you should be using in this case https://github.com/ros2/rclcpp/blob/6806cdf825d9fcdbb4bfe2c6e2a74c4cb8bc5e11/rclcpp/include/rclcpp/executor.hpp#L332. This input is the **future** not the node or executor. That future happens to be **based** on that node (and within it it calls `spin_once_impl(timeout_left);` which will spin on the main node)

SteveMacenski: Good question though - that really did require me to dig in pretty deep. I think the right answer is to use the `spin_until_future_complete` API I linked to, I think that solves all of the issues and you can use my initial suggestion of making them more similar. That implementation largely does what you suggested in this PR but more robustly and includes the spin inside of the function.

gezp: we must use a executor to spin, and the `spin_until_future_complete` API you [linked](https://github.com/ros2/rclcpp/blob/6806cdf825d9fcdbb4bfe2c6e2a74c4cb8bc5e11/rclcpp/include/rclcpp/executor.hpp#L332. ) is a function of `Class Executor`. so we can't avoid creating new  `SingleThreadedExecutor` to spin (`rclcpp::spin_until_future_complete()` will create a new `SingleThreadedExecutor` for each invoke).  

if we want to avoid creating new  `SingleThreadedExecutor`, only when the node has been spinning somewhere. but this will confuse developers (if we don't spin background, the `future` won't update and `invoke()` will block forever).

I can't find a better method than the previous `ServiceClient`, maybe this PR need be closed.

SteveMacenski: Ah, looks like I made a goof, that's in the executor object not the global namespace. 

I agree, maybe we close this for now and document that this is something we'd like to revisit down the line.

SteveMacenski:
I think that works, but that is alot of external requirements for a user of a general utility to remember. That wait_for then assumes the node that coming in is already `spin()`-ing to process callbacks rather than just a node object there for someone to use openly. If its already `spin()`-ing, isn't that why we made the internal executor/callback group in the first place? 

I think we're better off closing this PR and moving onto the next server as you suggested in the comment above. I don't love it, but I think what we have right now is perfectly serviceable unless there are improvements we can make that don't require alot of internal/advanced knowledge.

gezp:
thank you@SteveMacenski ! i will close this PR,  this PR is in vain, but this discussion helps me understand the executor/spin.

SteveMacenski:
This was a good discussion, I'm glad you went down that line of thinking, I think we both learned something here and that's all we can ask for :smile:

PR #2454, opened 2021-07-16
Reduce map saver nodes

gezp:
---

## Basic Info

-------

As described in [#816](https://github.com/ros-planning/navigation2/issues/816), `rclcpp_node_` in `class MapSaver` need be removed, you can find more details(other nodes which need be removed) in [#816 (comment)](https://github.com/ros-planning/navigation2/issues/816#issuecomment-876061677)

refer to [#2334](https://github.com/ros-planning/navigation2/pull/2334),I use callback group/ executor combo instead of `rclcpp_node_` in `saveMapTopicToFile()`.

- create new `CallbackGroup` for `map_sub`
- create new `SingleThreadedExecutor` for new `CallbackGroup` 
- wait map subscription by using `executor.spin_some()` in `while()`

## Test

colcon test

```
colcon test --packages-select nav2_map_server

# Result: 0 errors, 0 failures, 0 skipped
```

turtlebot3 demo test

```
# Server node mode test
ros2 launch nav2_map_server map_saver_server.launch.py
ros2 service call /map_saver/save_map nav2_msgs/srv/SaveMap "{map_topic: map, map_url: my_map, image_format: pgm, map_mode: trinary, free_thresh: 0.25, occupied_thresh: 0.65}"
# CLI mode test
ros2 run nav2_map_server map_saver_cli -t map -f my_map2

# Result: my_map and my_map2(.yaml and .pgm) will be generated correctly.
```

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/src/map_saver/map_saver.cpp
+++ b/nav2_map_server/src/map_saver/map_saver.cpp
@@ -42,7 +42,7 @@ using namespace std::placeholders;
 namespace nav2_map_server
 {
 MapSaver::MapSaver()
-: nav2_util::LifecycleNode("map_saver", "", true)
+: nav2_util::LifecycleNode("map_saver", "")
 {
   RCLCPP_INFO(get_logger(), "Creating");
 
--- a/nav2_map_server/src/map_saver/map_saver.cpp
+++ b/nav2_map_server/src/map_saver/map_saver.cpp
@@ -186,20 +186,29 @@ bool MapSaver::saveMapTopicToFile(
         map_msg = msg;
       };
 
-    // Add new subscription for incoming map topic.
-    // Utilizing local rclcpp::Node (rclcpp_node_) from nav2_util::LifecycleNode
-    // as a map listener.
     rclcpp::QoS map_qos(10);  // initialize to default
     if (map_subscribe_transient_local_) {
       map_qos.transient_local();
       map_qos.reliable();
       map_qos.keep_last(1);
     }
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
What was your process of  testing this PR continues to work as expected?

wjwwood (approved):
This looks ok to me.

You might get a cleaner solution using a future and `spin_until_future_complete()`. I could draft something up if you're unsure what I mean.

The logic around `while (rclcpp::ok())` and `(now() - start_time) > timeout` is just really similar to what `spin_until_future_complete` does already.

gezp:
according to suggestions of @wjwwood , i use `future` and `spin_until_future_complete()` to solve this.

SteveMacenski (approved):
Can you test this locally to verify no unexpected results? CI tests generally passed, but I don't trust CI alone for big changes to core tools. 

Admittedly I'm not an expert in promise/future-style programming but yeah this looks nice and can be a good reference in the future. There might be some other places we can make use of this mechanic in Nav2.

gezp:
yes, before PR, i have tested  this locally with turtlebot3 navigation2 demo, and files (.yaml and .pgm) were generated correctly.

SteveMacenski:
After reviewing some more info about futures/promises, they're not dissimilar to other things we have in Nav2 and other projects I've worked on for dealing with async tasks (but better). Not sure why it took me this long to see how they work under the hood.

PR #2459, opened 2021-07-19
improve SimpleActionServer

gezp:

---

## Basic Info

-------

improve `SimpleActionServer` 
* support two args  `options` and `group`  for action server ([rclcpp_action/create_server.hpp](https://github.com/ros2/rclcpp/blob/master/rclcpp_action/include/rclcpp_action/create_server.hpp#L57-L68))
* remove template `<typename nodeT = rclcpp::Node>` for `class SimpleActionServer`, and template the constructor. (like [NodeThread](https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/node_thread.hpp))

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -24,6 +24,7 @@
 
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp_action/rclcpp_action.hpp"
+#include "nav2_util/node_thread.hpp"
 
 namespace nav2_util
 {
@@ -54,7 +55,7 @@ public:
    * @param action_name Name of the action to call
    * @param execute_callback Execution  callback function of Action
    * @param server_timeout Timeout to to react to stop or preemption requests
-   * @param spin_thread  Whether to spin with a dedicated thread internally
+   * @param spin_thread Whether to spin with a dedicated thread internally
    */
   template<typename NodeT>
   explicit SimpleActionServer(
@@ -78,7 +79,7 @@ public:
    * @param action_name Name of the action to call
    * @param execute_callback Execution  callback function of Action
    * @param server_timeout Timeout to to react to stop or preemption requests
-   * @param spin_thread  Whether to spin with a dedicated thread internally
+   * @param spin_thread Whether to spin with a dedicated thread internally
    */
   explicit SimpleActionServer(
     rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_interface,
@@ -117,9 +118,9 @@ public:
       rcl_action_server_get_default_options(),
       callback_group_);
     if (spin_thread_) {
-      auto executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
-      executor_->add_callback_group(callback_group_, node_base_interface_);
-      executor_thread_ = std::make_unique<std::thread>([&]() {executor_->spin();});
+      auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+      executor->add_callback_group(callback_group_, node_base_interface_);
+      executor_thread_ = std::make_unique<nav2_util::NodeThread>(executor);
     }
   }
 
@@ -129,8 +130,7 @@ public:
   ~SimpleActionServer()
   {
     if (spin_thread_) {
-      executor_->cancel();
-      executor_thread_->join();
+      executor_thread_.reset();
     }
   }
   /**
@@ -514,8 +514,7 @@ protected:
   typename rclcpp_action::Server<ActionT>::SharedPtr action_server_;
   bool spin_thread_;
   rclcpp::CallbackGroup::SharedPtr callback_group_{nullptr};
-  rclcpp::executors::SingleThreadedExecutor::SharedPtr executor_;
-  std::unique_ptr<std::thread> executor_thread_;
+  std::unique_ptr<nav2_util::NodeThread> executor_thread_;
 
   /**
    * @brief Generate an empty result object for an action type
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -56,6 +56,7 @@ public:
    * @param execute_callback Execution  callback function of Action
    * @param server_timeout Timeout to to react to stop or preemption requests
    * @param spin_thread Whether to spin with a dedicated thread internally
+   * @param options Options to pass to the underlying rcl_action_server_t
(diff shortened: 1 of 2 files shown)


SteveMacenski:
Removed William, need to chat more with Zhenpeng before its a good use of his time.

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -54,19 +54,22 @@ class SimpleActionServer
...
+  template<typename NodeT>
   explicit SimpleActionServer(
-    typename nodeT::SharedPtr node,
+    NodeT node,
     const std::string & action_name,
     ExecuteCallback execute_callback,
     CompletionCallback completion_callback = nullptr,
-    std::chrono::milliseconds server_timeout = std::chrono::milliseconds(500))
+    std::chrono::milliseconds server_timeout = std::chrono::milliseconds(500),
+    bool spin_thread = false)
```
SteveMacenski, line 66: What happened to the options with the default argument?

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -103,9 +113,26 @@ class SimpleActionServer
       action_name_,
       std::bind(&SimpleActionServer::handle_goal, this, _1, _2),
       std::bind(&SimpleActionServer::handle_cancel, this, _1),
-      std::bind(&SimpleActionServer::handle_accepted, this, _1));
+      std::bind(&SimpleActionServer::handle_accepted, this, _1),
+      rcl_action_server_get_default_options(),
+      callback_group_);
+    if (spin_thread_) {
+      auto executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
```
SteveMacenski, line 120: Remove `auto`, that will instantiate a local executor_ copy, not initialize the class variable.

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -103,9 +113,26 @@ class SimpleActionServer
...
+      executor_->add_callback_group(callback_group_, node_base_interface_);
+      executor_thread_ = std::make_unique<std::thread>([&]() {executor_->spin();});
```
SteveMacenski, line 122: Using the `utils` node thread constructor you added in the last PR would be a good use here as well

SteveMacenski (changes requested):

nav2_util/include/nav2_util/simple_action_server.hpp
gezp: it's same as the previous one if `spin_thread` is false, in another word, `SimpleActionServer` use default callback group and need spin in `main()`.

nav2_util/include/nav2_util/simple_action_server.hpp
gezp: this is my mistake. if i use `NodeThread` , we don't need global variable  `executor_ ` , just store shared_ptr of executor in `NodeThread` .

SteveMacenski:
@wjwwood this is ready for you now

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -103,9 +114,25 @@ class SimpleActionServer
       action_name_,
       std::bind(&SimpleActionServer::handle_goal, this, _1, _2),
       std::bind(&SimpleActionServer::handle_cancel, this, _1),
-      std::bind(&SimpleActionServer::handle_accepted, this, _1));
+      std::bind(&SimpleActionServer::handle_accepted, this, _1),
+      rcl_action_server_get_default_options(),
+      callback_group_);
+    if (spin_thread_) {
+      auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
```
SteveMacenski, line 121: Please make the executor a class variable. I can foresee a time when we want to be able to access this within the action server's execution or make a `getExecutor()` method to mess with things during runtime. While its not strictly required today, it seems like a bad idea to have a thing running for which the program cannot access its executor after initialization.

SteveMacenski (changes requested):

nav2_util/include/nav2_util/simple_action_server.hpp
SteveMacenski: Yeah, sorry I'm asking why you removed the constructor argument for `options`. I think that was the right thing to do so that someone could override some of the action server's options if they wanted to.

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -103,9 +114,25 @@ class SimpleActionServer
...
+    }
   }
 
+  /**
+   * @brief A destructor
+   */
+  ~SimpleActionServer()
+  {
+    if (spin_thread_) {
+      executor_thread_.reset();
```
wjwwood, line 133: I think you need to join the thread right? Or is that handled in the destructor of the node thread class? This is kind of redundant. The unique pointer will be destroyed when this destructor finishes.

nav2_util/include/nav2_util/simple_action_server.hpp
gezp: yes, i agree with you. the type of `options` is `rcl_action_server_options_t` struct, which contains variables about server qos, etc ([rcl_action/action_server.h#L43-L61](https://github.com/ros2/rcl/blob/414fdbd479b83f2b8af8235ad1fa3aeba2b0027f/rcl_action/include/rcl_action/action_server.h#L43-L61)). i will add it.

nav2_util/include/nav2_util/simple_action_server.hpp
gezp: executor_thread_ is type of `NodeThread`, thread will be  joined in destructor of `NodeThread`, https://github.com/ros-planning/navigation2/blob/main/nav2_util/src/node_thread.cpp#L41-L45

wjwwood: Right, so this code block (and the whole destructor) isn't really needed. Right?

gezp: yes, i think executor_thread_ will auto destruct when SimpleActionServer destructs. but i refer some code like https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L70. 
anyway, i will remove it. i will add it again if need.

wjwwood (approved):

SteveMacenski (approved):
LGTM, submit a PR to add to the list in navigation.ros.org migration guide

Issue #2373, opened 2021-05-28
Controller/planner server segfault on clear local/global costmap


charlielito:
## Bug report

- Operating System:
  - Docker + Ubuntu 20.04, 10th gen i7
- ROS2 Version:
  - Galactic binaries
- Version or commit hash:
  - Galactic branch: 4b3391ee01993e97cc8a566363581108d1f291da
- DDS implementation:
  - cyclone 

#### Steps to reproduce issue
I am running gazebo with a custom map and custom robot. I am able to navigate, and while doing it, sometimes randomly the controller server dies when requested to clear the local costmap. Also sometimes the planner server dies when requested to clear the global costmap. I have the same set up as in foxy, just changed a little the `nav2_params.yaml` to meet the new API. I am also using the same default BT file as in foxy branch `navigate_w_replanning_and_recovery.xml`. In foxy everything runs well and it never dies.

<details>
  <summary>nav2_params.yaml</summary>

```yaml
amcl:
  ros__parameters:
    use_sim_time: True
    alpha1: 0.25
    alpha2: 0.2
    alpha3: 0.15
    alpha4: 0.2 
    alpha5: 0.2 # for omnidirectional robot_model_type only
    base_frame_id: "chassis"
    beam_skip_distance: 0.5
    beam_skip_error_threshold: 0.9
    beam_skip_threshold: 0.3
    do_beamskip: false
    global_frame_id: "map"
    lambda_short: 0.1
    laser_likelihood_max_dist: 2.0
    laser_max_range: 40.0
    laser_min_range: -1.0 # -1 laser’s reported minimum
    laser_model_type: "likelihood_field"
    max_beams: 60
    max_particles: 2000
    min_particles: 500
    odom_frame_id: "odom"
    pf_err: 0.05
    pf_z: 0.99
    recovery_alpha_fast: 0.0
    recovery_alpha_slow: 0.0
    resample_interval: 1
    robot_model_type: "differential"
    save_pose_rate: 0.5
    sigma_hit: 0.2
    tf_broadcast: true
    transform_tolerance: 1.0
    update_min_a: 0.2
    update_min_d: 0.25
    z_hit: 0.5
    z_max: 0.05
    z_rand: 0.5
    z_short: 0.05
    scan_topic: scan

amcl_map_client:
  ros__parameters:
    use_sim_time: True

amcl_rclcpp_node:
  ros__parameters:
    use_sim_time: True

bt_navigator:
  ros__parameters:
[...]

PR #2460, opened 2021-07-19
Fix for voxel grid out-of-range issues due to loss of precision during floating-point calculations

AlexeyMerzlyakov:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2373 |
| Robotic platform tested on | TB3 simulation at shared in ticket pre-recorded dataset |

---

* Fix for floating point precision loss due to double->int->double conversions which are leading to incorrectly calculated data array ranges leading to out-of-array addressing causing segfaults on some configurations.
* (Additionally) Correct shift length reduction by min range components instead of absolute min range value.

### DIFF
--- a/nav2_voxel_grid/include/nav2_voxel_grid/voxel_grid.hpp
+++ b/nav2_voxel_grid/include/nav2_voxel_grid/voxel_grid.hpp
@@ -226,18 +226,6 @@ public:
     double x1, double y1, double z1, unsigned int max_length = UINT_MAX,
     unsigned int min_length = 0)
   {
-    int dx = int(x1) - int(x0);  // NOLINT
-    int dy = int(y1) - int(y0);  // NOLINT
-    int dz = int(z1) - int(z0);  // NOLINT
-
-    unsigned int abs_dx = abs(dx);
-    unsigned int abs_dy = abs(dy);
-    unsigned int abs_dz = abs(dz);
-
-    int offset_dx = sign(dx);
-    int offset_dy = sign(dy) * size_x_;
-    int offset_dz = sign(dz);
-
     // we need to chose how much to scale our dominant dimension, based on the
     // maximum length of the line
     double dist = sqrt((x0 - x1) * (x0 - x1) + (y0 - y1) * (y0 - y1) + (z0 - z1) * (z0 - z1));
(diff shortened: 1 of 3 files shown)


SteveMacenski (approved):

SteveMacenski:
Retriggering CI, not sure why it failed to build on some caching issues. Anything else you want to do before this is merged? I'll cherry pick to galactic / foxy (does foxy have this?) and do an immediate release, this is a big enough fix we should get something out to users.

SteveMacenski:
@ruffsl there's an error in caching

```
mv: cannot stat '/root/.ccache': No such file or directory

Exited with code exit status 1

CircleCI received exit code 1

```

ruffsl:
Please rebase from the main branch to ensure the CI config for this PR is up to date, e.g. with fixes from https://github.com/ros-planning/navigation2/pull/2403 .

SteveMacenski:
You changed the length calculation? Can you explain geometrically what's happening there now?

AlexeyMerzlyakov:
Previous calculations of `length = (unsigned int)(scale * abs_dx) - abs(min_shift_x);` still may produce incorrect results for some cases due to incorrect `min_shift_x` truncating. This will be also the cause of loss of precision thus we need to calculate everything in doubles up to the end and then transfer them to ints.

Geometrical example why I've corrected shift calculation in https://github.com/ros-planning/navigation2/pull/2460/commits/d97bf5500f33c107e62cbf1ee1b1c00eca514ab4 is shown in the picture below:
P10720-211736_r

SteveMacenski:
I know you were on vacation, let me know if there's any update on this crashing issue, its pretty high priority blocking a new sync

AlexeyMerzlyakov:
@SteveMacenski , yes I am back online and then buried oneself in summer school lectures work recent days. Considering the priority of this issue I will continue the investigation as soon as possible.

AlexeyMerzlyakov:
Force-pushed with updated solution. The initial problem as it was supposed was incorrect boundaries compensation calculations made in 13f518ac. However, when I've changed `dx`/`dy`/`dz` types from int to double, I also broke the raytraceLine()->bresenham3D() magic based on suppose that we are operating in integer space of array values. Thus, I resorted to a different solution: to correctly compensate `x0`, `y0` and `z0` values in the space and run `bresenham3D()` algorithm towards these compensated `x0_min`/`y0_min`/`z0_min` values. In other words, the initially working `bresenham3D()` is running on newly updated `(x0_min,y0_min,z0_min)...(x1,y1,z1)` boundaries as if they are passed from outside w/o any compensating logic breaking.

For verification you also can compare this change with latest state before the regression: https://github.com/ros-planning/navigation2/compare/7e284a3f02a1b889cadfb6ab402a8278337cf040..332e9c48454ffa998dd3bf8e2345cd3473783adf#diff-d4260e8fd3cbbe31986833c7e04aeebf742a967c6ee0c7854e48733688dfb44f and find out a `voxel_grid.hpp` in there (maybe there could be a comparison of a single file between two revisions, but I am not good at GitHub UI)

Also, adding a comprehensive regression test for checking the edges of `raytraceLine()->bresenham3D()` operation (writing this test exactly helped me to came to this solution).

SteveMacenski (approved):
Needs verification of fix, but looks plausible to me

Issue #2433, opened 2021-07-02
Fix broken tests created after Galactic release


SteveMacenski:
Since the Galactic release, our CI has been introduced with a number of failing tests. We fixed the trivial ones of linter profile changes, but there are still several nav2_utils tests that are failing that probably just need to be tweaked to pass again.

Goal: Look over the failures in CI, resolve them, and merge into the `main` branch

SteveMacenski:
e.x. https://app.circleci.com/pipelines/github/ros-planning/navigation2/5602/workflows/02427c64-ce4a-4e0b-af82-db0f574413c9/jobs/20522

karl-schulz:
Hi,
is this issue the reason why the Nav2 binaries for galactic are not yet available? Or are they just not released yet for other reasons?

`sudo -H apt-get install -y -qq ros-galactic-nav2-core` 
yields
`E: Unable to locate package ros-galactic-nav2-core`

SteveMacenski:
We're just waiting on a Galactic sync. I barely missed the first release of Galactic so some of the packages made it through and some did not. They're going to do a sync in the next few days so you should be able to get binaries then, the test failures in CI didn't block that.

SteveMacenski:
We have flaky tests now, which is a slightly different problem. The core of this has been addressed and CI is turning over

Issue #1710, opened 2020-05-09
Working list of algorithms to include


SteveMacenski:
Hi all,

I have a working list of planners / controllers that I've been compiling and I wanted to share them here for potential ideas if anyone's interested in developing / adapting one in particular. Feel free to ping me about this and I'll help however I can

Planners
- [x] A* - holonomic robots or robots small relative to environment
- [x] Dij. - holonomic robots or robots small relative to environment
- [x] ~OMPL - non-holonomic and non-circular footprint robots, potential formulation to include dynamic obstacles **[IN PROGRESS]**~ state lattice and hybrid-A* supersedes 
- [ ] GPS / cartesian route following - taking in a list of semantic GPS waypoints and issuing a path based on it to follow exactly. https://github.com/ros-planning/navigation2/issues/1631#issuecomment-730009390. Or more  generally a teach and repeat for non-GPS too. Teach via recording while moving or manual annotation. 
- [x] Hybrid-A* - non-holonomic and non-circular robots with smoothed paths
- [x] ~Vornoi / potential field planner - for a robot to stay far, far away from things as its primary goal~ NavFn, Theta*, and other planners can be tuned to highly penalize costed regions. While these are easy to implement, the quality is low so its not a good addition for a modern stack. Very 1990s. 
- [x] Theta* - doing straight line planning based on LoS
- [x] State Lattice - non-holonomic and non-circular robots, useful in analog situations as Hybrid-A* but for arbitrary motion models (omni, diff, non-cars, forklifts) Hybrid-A* meets this need, might be adding state lattice template node to smac 
- [x] non-zero via-point planning (e.g. calling a continuous, kinematically feasible planner from A->1->2->3->…->B and then adding them all to a single path) -- route server #2229 ? Or new action interface in planner server to have multiple via-points to plan (!)

Controllers
- [x] DWA - compute trajectories based on weighted heuristics
- [x] TEB - compute trajectories based on optimal bands 
- [ ] MPPI / MPC /Nonlinear MPC- compute trajectories by model simulation supporting high speeds. Exact path follower / dynamic obstacles in formulation **in progress**
- [x] ~LQR/iLQR/CiLQR - "MPC-lite" is the easiest way to describe it supporting high speeds~ MPPI supersedes 
- [x] Pure Pursuit - following a carrot 

Recoveries
- [x] Spin - spin in place
- [x] Backup - back up from pose
[...]

PR #2462, opened 2021-07-20
Add costmap_thread_.reset() on the destructor of ControllerServer class

harderthan:
## Basic Info

| ------ | ----------- |

---

Add `costmap_thread_.reset()` on the destructor of ControllerServer class
I referred [this line](https://github.com/ros-planning/navigation2/blob/21a44bcb559682f67289ac72571206597c00855b/nav2_planner/src/planner_server.cpp#L70) of `nav2_planner` package.

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/nav2_controller.cpp
+++ b/nav2_controller/src/nav2_controller.cpp
@@ -71,6 +71,7 @@ ControllerServer::~ControllerServer()
   progress_checker_.reset();
   goal_checkers_.clear();
   controllers_.clear();
+  costmap_thread_.reset();
 }
 
 nav2_util::CallbackReturn


SteveMacenski (approved):

SteveMacenski:
LGTM, rerunning CI, a suspicious number of system tests failed. If that happens again, you might have introduced a regression.

On another note, I see you have an MPC controller on your github repo -- I'd love to hear about that if you're open to sharing!

harderthan:
@SteveMacenski, thank you for your approval.

The MPC Controller repo is from [other](https://github.com/Geonhee-LEE/mpc_ros).
I just integrated it with nav2 plugin. I'd like to develop it to support vehicle kinematics model.

And, I'm considering to develop nav2 plugins of nav2_planner or nav2_controller using well-known approach like mpc. I would like to ask your advice, such as suggestion with paper or links, If there are approaches under consideration.

Thanks.

SteveMacenski:
Remaining test failures I'm less suspicious about given the state of #2433, merging!

---

Got it- I think I've seen that github project before but for some reason I haven't looked at it further. I'd be happy if you were to develop new planners / controllers for Nav2! I don't have much generic advice, since it largely depends on your needs (for a company) or interests (as a student or researcher) but I have a few thoughts depending on what road you want to follow. There's also this ticket https://github.com/ros-planning/navigation2/issues/1710 which covers my general idea of all of the algorithms I'd have to have in Nav2. 

In terms of MPC, I think there are 2 classes worth having an implementation of:
- Path tracking, which I think that work does, where it uses MPC to track a path given a model -- but doesn't contain any additional constraints for obstacles
- Local trajectory planning, where it uses MPC to track a path given a model, but also has some constraints to be able to have DWB or TEB like dynamic obstacle avoidance capabilities letting it deviate from the path when necessary. 

A single planner could probably do both with different settings, but I don't know architecturally if they'd be similar or not to make that practical.

PR #2411, opened 2021-06-16
Nav2 Simple (Python3) Commander Library

SteveMacenski:
Adding in a python3 library that can be used to create autonomy applications with Nav2 abstracting away the ROS2 and Action specifics. 

TODO
- [x] external review
- [x] Navigation.ros.org migration guide and configuration guides  https://github.com/ros-planning/navigation.ros.org/pull/179
- [x] announce to community  https://discourse.ros.org/t/nav2-new-python3-nav2-api-to-build-applications/20954/2
- [x] Handle aws warehouse world release so this can be deployed to binaries https://github.com/ros-planning/navigation2/issues/2414

PR #2466, opened 2021-07-20
Python string format

Timple:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/pull/2411#issuecomment-863561181 |

---

As per @SteveMacenski's request: https://github.com/ros-planning/navigation2/pull/2411#issuecomment-863561181

---

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
+++ b/nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
@@ -123,7 +123,7 @@ def generate_launch_description():
     # Define commands for launching the navigation instances
     nav_instances_cmds = []
     for robot in robots:
-        params_file = LaunchConfiguration(robot['name'] + '_params_file')
+        params_file = LaunchConfiguration(f"{robot['name']}_params_file")
 
         group = GroupAction([
             IncludeLaunchDescription(
--- a/nav2_simple_commander/test/test_flake8.py
+++ b/nav2_simple_commander/test/test_flake8.py
@@ -21,5 +21,5 @@ import pytest
 def test_flake8():
     rc, errors = main_with_errors(argv=[])
     assert rc == 0, \
-        f'Found {len(errors)} code style errors / warnings:\n' + \
+        'Found %d code style errors / warnings:\n' % len(errors) + \
         '\n'.join(errors)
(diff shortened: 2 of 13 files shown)


nav2_simple_commander/test/test_flake8.py
```diff
@@ -21,5 +21,5 @@
 def test_flake8():
     rc, errors = main_with_errors(argv=[])
     assert rc == 0, \
-        'Found %d code style errors / warnings:\n' % len(errors) + \
```
SteveMacenski, line 24: The `test/` entries are autogenerated by colcon and aren't to be hand modified.

SteveMacenski (changes requested):
Other than this one file, LGTM. Though I am not a python developer primarily, so I'd like a review from someone that does that more than I

nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
```diff
@@ -123,7 +123,7 @@ def generate_launch_description():
     # Define commands for launching the navigation instances
     nav_instances_cmds = []
     for robot in robots:
-        params_file = LaunchConfiguration(robot['name'] + '_params_file')
+        params_file = LaunchConfiguration(f"{robot['name']}_params_file")
```
klintan, line 126: I would probably try to keep single / double quotes consistent across the code base. If not for any other specific reason, this discussion could be valuable https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python (but my personal preference is just to make sure to be consistent, given there are no double / single quotes in the text itself)

klintan (commented):
Looks good to me :) one minor comment.

nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
SteveMacenski: Oh yeah, @timple make sure you're running the linters and this passes in the python ones

Timple: @klintan: Agreed on the consistency, changed.

@SteveMacenski: I doubt these kind of CI errors are related to the python strings?
```
/opt/overlay_ws/src/navigation2/nav2_dwb_controller/dwb_critics/test/obstacle_footprint_test.cpp:246
  critic->lineCost(3, 3, 0, 50)
    Which is: 85
  50
```

`ament_pep257` succeeds locally.

SteveMacenski: No, other issues, have been fixed now but your fork is probably outdated. Merging! Thanks!

SteveMacenski (approved):

Issue #2389, opened 2021-06-04
Proper use of Gazebo / model spawner for more modular model / world use


SteveMacenski:
Right now, we have the model and the world all in 1 world file, rather than having a world file that we spawn a gazebo model into.

There's a `spawn_entity.py` and a multirobot version in Nav2 that we should be using instead to spawn robots within a world so that these things are not tied together. This would let someone change the robot or world more easily to spawn (different robots, in different places in the world, or in different worlds).

I think the gazebo gzserver needs to be the ros factory so file instead to use the model spawner, and any other things required (separate URDF file with gazebo plugins from the .world file)

sathak93:
hi steve

i have manually input world file and spawned robot model sdf file and changed gazebo state to factory.
![Screenshot from 2021-07-16 16-40-31](https://user-images.githubusercontent.com/57533054/125938978-6977242a-459b-4fbe-ac22-95ca34d3900c.png)
i debugged it  to  some extent.
the only difference between the launch file with spawn entity and combined model + world is 
the layered  costmaps  are not published 
the if block in 
https://github.com/ros-planning/navigation2/blob/5a7027536068183602f10b419bd711fafdad6784/nav2_costmap_2d/src/costmap_2d_ros.cpp#L422 is always becomes false as the current time is always zero
![Screenshot from 2021-07-16 16-47-28](https://user-images.githubusercontent.com/57533054/125940897-466fbe16-5b05-4e55-aeda-2182e397b4e6.png)
![Screenshot from 2021-07-16 16-49-46](https://user-images.githubusercontent.com/57533054/125940926-7ebd84c1-959a-44fc-ab81-5b78cb03bfe1.png)

can you comment on  why auto current_time = now ()  is not updated or anything i did wrong?

sathak93:
making the if block  true publishes global and local costmaps

SteveMacenski:
I'm not sure, did it work before you changed the model spawning / gazebo? These seem unrelated but clearly must be somehow... maybe the gazebo .so file isn't publishing the clock in the same way?

sathak93:
@SteveMacenski  i found  the problem it seems that libgazebo_ros_factory.so  should be used in addition to libgazebo_ros_init.so instead of replacing it. 
[...]

PR #2473, opened 2021-07-23
Spawn turtlebot3 separately

sathak93:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (#2389) |
| Primary OS tested on | (Ubuntu 20.04 ROS 2 galactic  )  |
| Robotic platform tested on | (gazebo simulation of turtlebot) |

---

separated gazebo robot model file from  gazebo world file and spawn tb3 using  nav2 gazebo spawner.

---

#### For Maintainers: 


nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -149,9 +151,21 @@ def generate_launch_description():
...
         output='screen',
-        parameters=[{'use_sim_time': use_sim_time}],
-        remappings=remappings,
-        arguments=[urdf])
+        parameters=[{'use_sim_time': use_sim_time,
+                     'robot_description': robot_description}],
+        remappings=remappings)
+
+    start_gazebo_spawner_cmd = Node(
+        package='nav2_gazebo_spawner',
```
SteveMacenski, line 159: Question: can you use the gazebo spawn model instead? https://github.com/ros-simulation/gazebo_ros_pkgs/blob/231a7219b36b8a6cdd100b59f66a3df2955df787/gazebo_ros/scripts/spawn_model Can you find the major differences between the Nav2 one and the Gazebo one? I believe we made that before Gazebo had that feature and we ported it from ROS 1 so it very well might be the same or so similar we  can remove that. Less code here is always good news

SteveMacenski (commented):
LGTM, question on spawning, I have to say, I'm really surprised that the URDF works for gazebo spawning. I'm actually note clear where in your spawn command you give the system your robot file to spawn? I didn't think the URDF we had in Nav2 contained the gazebo plugin descriptions, I thought that was in the `waffle.model` file.

BTW if we're not longer using that, we can remove `waffle.model` from the repo as well (though I think you may not be using the right robot for gazebo spawning... so maybe still need parts of that)

sathak93:
@SteveMacenski i only used nav2_gazebo_spawner pkg to input the waffle.model file like it is in multi tb launch file.
I thought i should not be edit waffle.urdf
If you want i will edit waffle.urdf with gazebo tags.  I had  tried that before but there was some problem. i don't remember. I will look into it.
Spawn_entity.py has different ways to spawn the robot. How do you want to spawn     
It.? As a file input or as  robot_description topic subscription.

nav2_bringup/bringup/launch/tb3_simulation_launch.py
sathak93: sure we can use. there is no major difference between them. it only additionally checks is the model has differential drive.

Yousseflah:
To answer @SteveMacenski's question about spawning, yes indeed the urdf in nav2 is not complete, but in the start_gazebo_spawner_cmd, there is the following argument `'--turtlebot_type', 'waffle',`  and the nav2_gazebo_spawner package takes the model from the [turtlebot3_gazebo](https://github.com/ROBOTIS-GIT/turtlebot3_simulations/blob/galactic-devel/turtlebot3_gazebo/models/turtlebot3_waffle/model.sdf) package depending on the turtlebot_type argument you specified, and in this package (turtlebot3_gazebo) there is a complete sdf model with all the required plugins.
You guys decide whether to copy/paste the model into nav2_bringup ( + convert it to urdf maybe !) or you can always rely on the turtlebot3_gazebo package to get the sdf.

And for the use of gazebo spawner model directly, do not forget to test it with multi robot senario or at least using namespace to be sure that all nodes/topics/services are namespaced during the spawn.

nav2_bringup/bringup/launch/tb3_simulation_launch.py
SteveMacenski: OK sweet, just remove the nav2 version then and replace with gazebo ros's! That's great to hear!!

SteveMacenski:
>  and the nav2_gazebo_spawner package takes the model from the turtlebot3_gazebo package depending on the turtlebot_type argument you specified, and in this package (turtlebot3_gazebo) there is a complete sdf model with all the required plugins.

I see, can you add a local file containing the robot's gazebo model that is used instead? It just helps that we can add a parameter for `robot_sdf` to load other robots, not just TB3's. The default will be TB3's for now, but we want to be able to customize it for Nav2 or replace with other robots easily. Relying on Gazebo's internal registry of robots is limited.

sathak93:
i didn't see your inline comment. i will change to spawn entity  from gazebo_ros and commit soon

nav2_bringup/bringup/launch/tb3_simulation_launch.py
sathak93: i tried it. it only works without namespace.  tb_differential_drive plugin sends tf to /tf even with namespace. i tried to remap it using remap rule it didn't work out.
Screenshot from 2021-07-29 22-41-56
i also tried to  bring the block of nav2_spawner code that remaps it to launch,  namespace being runtime variable it doesn't work.
do you have any suggestion to make it work.

SteveMacenski: Did you try the actual nav2_spawner and not just bringing back in the block of code? 

If they both fail, then I don't know what to tell you, its probably not relevant to this PR. Sounds like a ticket to file with the turtlebot3 repo. Did this used to work when TB3 used the ROS2 diff drive plugin? Did you try simply replacing theirs with the generic one? I think that's what we use in the Nav2 version of the TB3 SDF file in https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/bringup/worlds/waffle.model#L516

There's many reasons we do NOT use the TB3 supplied files and why long term we are likely to entirely stop supporting it in Nav2. Its just a minefield of lazy engineering and bad user support

sathak93: nav2_gazebo_spawner works fine. we can finish this PR using it. you earlier said that to try spawn with spawn_entity.py of gazebo_ros. it doesn't work when we spawn with namespace. it has problem with diff_drive_node it remap cmd_vel and odom topic with robot1 namespace but not tf see the above fig. nav2_gazebo_spawner avoid that by https://github.com/ros-planning/navigation2/blob/e96450db3a425d8eb54737dffc09eef0bb1e1391/nav2_bringup/nav2_gazebo_spawner/nav2_gazebo_spawner/nav2_gazebo_spawner.py#L87

nav2_gazebo_spawner stays for now. 
shall i finish this pr i have some changes in rviz files and param file

SteveMacenski: OK makes sense!

nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
```
SteveMacenski, line 13: Linting
```
E231 (./launch/tb3_simulation_launch.py:183:50) - nav2_bringup.flake8

missing whitespace after ',':
    robot_model_sdf = os.path.join(tb3_gazebo_dir,'models', 'turtlebot3_waffle', 'model.sdf')

```

nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
```diff
@@ -100,26 +100,12 @@ def generate_launch_description():
...
-    # Start Gazebo with plugin providing the robot spawing service
+    # Start Gazebo with plugin providing the robot spawning service
     start_gazebo_cmd = ExecuteProcess(
-        cmd=[simulator, '--verbose', '-s', 'libgazebo_ros_factory.so', world],
+        cmd=[simulator, '--verbose', '-s', 'libgazebo_ros_init.so',
+                                     '-s', 'libgazebo_ros_factory.so', world],
         output='screen')
 
-    # Define commands for spawing the robots into Gazebo
-    spawn_robots_cmds = []
```
SteveMacenski, line 109: This is the multi-robot launch file, why was this all deleted? Don't we need need something to spawn the 2 different robots?

SteveMacenski: I see you're forwarding info below, but if we spawned all `N` robots here, couldn't we just have their navigation systems and 1 simulator launch in the other file?

I suppose I just don't understand why changes needed to be made structurally so that we could just separate out the world and robot model files. Could the pose be read in as a array vector rather than independent X, Y,and Z components? Because we also need to consider orientation as well.

sathak93: sorry i didn't  get notification to reply quickly

the pose can be read as vector if you suggest.

mutli_tb3_simulation_launch.py  file uses tb3_simulation_launch.py https://github.com/ros-planning/navigation2/blob/792893a5b56256cff8e9d4fb581e71d23892b5d6/nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py#L142
before there is no nav2_gazebo_spawner_node in tb3_simulation_launch file.  in order to seperate  world and robot nav2_gazebo_spawner _node  should be created in tb3_simulation_launch file. i thought its neat to use this node  in multi_tb3 file as launch configuration input.
  we  can also  remove the file [spawn_tb3_launch.py](https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/bringup/launch/spawn_tb3_launch.py) since it contains a single node .

@SteveMacenski  if you prefer the earlier structure . the node can be ignored with if condition (simulator) and the structure can be restored.

nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -141,6 +165,8 @@ def generate_launch_description():
         cwd=[launch_dir], output='screen')
 
     urdf = os.path.join(bringup_dir, 'urdf', 'turtlebot3_waffle.urdf')
+    with open(urdf, 'r') as infp:
```
SteveMacenski, line 168: Is this necessary? It looks before like we just gave `urdf` as an argument to the robot state publisher and it just handled it itself? Just asking the question

SteveMacenski, line 168: Is this necessary? It looks before like we just gave `urdf` as an argument to the robot state publisher and it just handled it itself? Just asking the question

nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
SteveMacenski: Let me think about this (aloud), these are just be spitballing ideas here. I don't know what the best solution is, but lets chat it over and use the ones we like the best. I don't think there's an objective right answer to this one

- We do what you suggest in the PR now, have `N` of the Tb3 launch files called each spawning its own robot with simulation set to false. We should add a comment in the code in this case that the spawning is happening when navigation/simulation is coming up.
- We have an option in the Tb3 sim file for whether or not to spawn a robot, set true default and use false for the multirobot case and have that multirobot launch file use the spawner in a for loop 

Of those two options, I like what you have now better. What do you think?

---

A separate topic, but what if we did all the finding robot SDF file stuff in the spawner launch file to abstract that from these launch files (e.g. https://github.com/ros-planning/navigation2/pull/2473/files#diff-f2cb98e53f28476ba70befa88ddc8279de87edf8b4bd6de915f1dbd844281252R182-R183)? That way that launch file is self contained in finding and spawning the robot (probably with a launch configuration to override).

sathak93: i am inclined to first option thats why i removed spawn block in multi_tb3_launch file. now the single tb3_simulation_launch file has two launch description includes bringup for nav2 related nodes. rviz to start rviz everything else in it is specific to robot.
i think its more user friendly to beginners to try change robot urdf, sdf, world, map and  simulator
 multi _launch file just starts gazebo and iterates for N robots.

about the separate topic why create separate launch file for spawning only how about we club simulator specific nodes or robot + simulator +robot_state _publisher node since if one change urdf, sdf  also need to be changed. also it would be more convenient to change simulator like ignition , webot etc

just my thoughts

sathak93: converting the pose as a vector in launch argument is not straight forward. i am finding  how to do it. do you came across any reference?

nav2_bringup/bringup/launch/tb3_simulation_launch.py
sathak93: there is a warning if we gave urdf as an argument.
[WARN 1626405835.205918208] [robot_state_publisher]: No robot_description parameter, but command-line argument available.  Assuming argument is name of URDF file.  This backwards compatibility fallback will be removed in the future. (RobotStatePublisher() at /tmp/binarydeb/ros-galactic-robot-state-publisher-2.4.3/src/robot_state_publisher.cpp:85)  

to remove this the urdf should be converted to robot description parameter

nav2_bringup/bringup/launch/multi_tb3_simulation_launch.py
SteveMacenski: OK. 

On the vector question, I think it would be good to look at the TF static broadcaster as an example, they take in the exact same kind of vector we're talking about.

nav2_bringup/bringup/launch/tb3_simulation_launch.py
SteveMacenski: Got it, thanks for fixing that

sathak93:
hey @SteveMacenski  please review the changes.
by adding <remapping>  /tf:=tf  <remapping /> tag in tb3_diff_drive plugin, we can now use spawn_entity.py it works with and without namespace.

on initial pose, we cannot use  a single launch arg to define six pose variables. As launch arg can only of type string, initialized on runtime and cannot be modified in the same file. i could not find any other ways to do.  does the current pose definition looks good?

SteveMacenski:
Please pull in the latest changes to the main branch to fix CI

nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -149,16 +163,29 @@ def generate_launch_description():
...
         namespace=namespace,
         output='screen',
-        parameters=[{'use_sim_time': use_sim_time}],
-        remappings=remappings,
-        arguments=[urdf])
+        parameters=[{'use_sim_time': use_sim_time,
+                     'robot_description': robot_description}],
+        remappings=remappings)
+
+    robot_model_sdf = os.path.join(bringup_dir, 'worlds', 'waffle.model')
```
SteveMacenski, line 170: Make this into a launch configuration too that defaults to this file path so people can specify other robots

nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -149,16 +163,29 @@ def generate_launch_description():
...
+
+    start_gazebo_spawner_cmd = Node(
+        package='gazebo_ros',
```
SteveMacenski, line 173: Sweet! If we can actually get rid of the nav2_gazebo_spawner package now? If so, feel free to delete it in this PR now!

SteveMacenski (changes requested):
Great! Just need 
- Docs to make clear the "remap hack" we set up for multirobot
- remove the custom spawner if its not needed anymore
- add a configuration for which model file to use in case someone wants another one

nav2_bringup/bringup/README.md
```diff
@@ -134,3 +134,10 @@ In RVIZ:
 * Make sure all transforms from odom are present. (odom->base_link->base_scan)
 * Localize the robot using “2D Pose Estimate” button.
 * Send the robot a goal pose using “2D Nav Goal” button.
+
+Note:
+* nav2_gazebo_spawner pkg inside nav2_bringup directory is deleted.
+* use of nav2_gazebo_spawner to spawn the robot in gazebo is not recommended any more. instead use spawn_entity.py 
+of gazebo_ros to spawn the robot.
+* gazebo should be started with both libgazebo_ros_init.so and libgazebo_ros_factory.so to work correctly
+* spawn_entity node could not remap /tf and /tf_static to tf and tf_static in the launch file yet. instead it should be done as remapping argument <remapping>/tf:=tf</remapping>  <remapping>/tf_static:=tf_static</remapping> under ros2 tag in each plugin which publishes transforms. it is essential if the nodes are started with namespace and also in multi robot launch to differentiate the tf's of the different robot.
```
SteveMacenski, line 143:
```suggestion
* spawn_entity node could not remap /tf and /tf_static to tf and tf_static in the launch file yet, used only for multi-robot situations. Instead it should be done as remapping argument <remapping>/tf:=tf</remapping>  <remapping>/tf_static:=tf_static</remapping> under ros2 tag in each plugin which publishs transforms in the SDF file. It is essential to differentiate the tf's of the different robot. 
```

SteveMacenski:
I can't tell you how much I love deleting code :heart:

nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -47,6 +47,14 @@ def generate_launch_description():
...
     headless = LaunchConfiguration('headless')
     world = LaunchConfiguration('world')
+    pose = {'x': LaunchConfiguration('x_pose', default='-2.00'),
+            'y': LaunchConfiguration('y_pose', default='-0.50'),
+            'z': LaunchConfiguration('z_pose', default='0.01'),
+            'R': LaunchConfiguration('roll', default='0.00'),
+            'P': LaunchConfiguration('pitch', default='0.00'),
+            'Y': LaunchConfiguration('yaw', default='0.00')}
+    robot_name = LaunchConfiguration('robot_name')
+    robot_sdf_file = LaunchConfiguration('robot_sdf_file')
```
SteveMacenski, line 57:
```suggestion
    robot_sdf = LaunchConfiguration('robot_sdf')
```

nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -125,9 +133,19 @@ def generate_launch_description():
...
+        default_value=os.path.join(bringup_dir, 'worlds', 'world_only.model'),
         description='Full path to world model file to load')
 
+    declare_robot_name_cmd = DeclareLaunchArgument(
+        'robot_name',
+        default_value='turtlebot3_waffle',
+        description='name of the robot')
+
+    declare_robot_sdf_file_cmd = DeclareLaunchArgument(
+        'robot_sdf_file',
```
SteveMacenski, line 145:
```suggestion
        'robot_sdf',
```

nav2_bringup/bringup/launch/tb3_simulation_launch.py
```diff
@@ -149,16 +169,27 @@ def generate_launch_description():
...
+                     'robot_description': robot_description}],
+        remappings=remappings)
+
+    start_gazebo_spawner_cmd = Node(
+        package='gazebo_ros',
+        executable='spawn_entity.py',
+        output='screen',
+        arguments=[
+            '-entity', robot_name,
+            '-file', robot_sdf_file,
```
SteveMacenski, line 182:
```suggestion
            '-file', robot_sdf,
```

nav2_system_tests/src/system/test_multi_robot_launch.py
```diff
@@ -52,22 +52,22 @@ def generate_launch_description():
 
     # Launch Gazebo server for simulation
     start_gazebo_cmd = ExecuteProcess(
-            cmd=['gzserver', '-s', 'libgazebo_ros_factory.so',
-                 '--minimal_comms', world],
+            cmd=['gzserver', '-s', 'libgazebo_ros_init.so',
```
sathak93, line 55: @SteveMacenski  this change fixes the multi_robot system test. with few minor changes and i managed to make the test passed locally. shall i commit  the changes regarding to it also in this PR?

SteveMacenski (approved):

nav2_system_tests/src/system/test_multi_robot_launch.py
SteveMacenski: I'm not sure what you mean? Are these changes needed before this PR can go in? Maybe it would make sense for me to approve and merge this and then you submit a new PR with those proposed changes that we can look at. That way we can get this work completed and discuss that separately if there's any concerns.

sathak93: @SteveMacenski  i will open new pr on it later.
 
i cannot think of anything else to do in this pr. i tried the changed launch files on foxy, galactic , rolling. nav2  succeeded on all.  there was some error on rviz2 material rendering in rolling. 
Screenshot from 2021-08-19 06-30-39.
i don't think it is related to this pr.

you can merge this. if you are okay with it.

thanks. it's my pleasure to work with you on my first PR. i am eager to learn more  and contribute.

SteveMacenski: It was very much so appreciated, thanks for your hard work! Another thing in the follow up PR might be to remove the `bringup` directory (e.g. we have `nav2_bringup/bringup/<pkg stuff>` since we no longer the nav2_gazebo_spawner in that directory. 

Let me know if you're interested in working on another ticket / project! I'd definitely appreciate it :smile:

sathak93: sure.
i thought to do it in this pr . since  I'm  new to github things i was feared that doing it might mess things up.  i will create a follow up PR. more PR's will make my github profile look good.

I am not a CS major . i have completed post graduate in mechanical design.  Robotics and ROS made me interested towards programming. i love to learn and work in c++. i am currently free i will find another project that is suitable to me soon most probably in nav2.

Issue #2332, opened 2021-05-10
Regulated Purepursuit is slow


vimalrajayyappan:
FollowPath:
      plugin: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
      desired_linear_vel: 7.2
      max_linear_accel: 2.5
      max_linear_decel: 2.5
      lookahead_dist: 1.8 #0.6
      min_lookahead_dist: 1.0 #0.3
      max_lookahead_dist: 2.0 #0.9
      lookahead_time: 3.0 #1.5
      rotate_to_heading_angular_vel: 1.8
      transform_tolerance: 0.1
      use_velocity_scaled_lookahead_dist: false
      min_approach_linear_velocity: 0.05
      use_approach_linear_velocity_scaling: true
      max_allowed_time_to_collision: 1.0
      use_regulated_linear_velocity_scaling: true
      use_cost_regulated_linear_velocity_scaling: false
      regulated_linear_scaling_min_radius: 0.9
      regulated_linear_scaling_min_speed: 0.25
      use_rotate_to_heading: true
      rotate_to_heading_min_angle: 0.785
      max_angular_accel: 3.2

The tuning parameters mentioned above are the one I'm using. It may look wierd, but I tried giving the max values possible to check any reflecting changes. But I couldnt find any. The purepursuit is running at its slow speed as it used to be. Am I missing anything/ Can you please help me on that?

SteveMacenski:
What do you mean "runs too slowly"? I can get this to run at 1khz easily on my 7th gen i5. Please provide _much_ more detail about exactly what your issue is, what you have tried to solve it, and any relevant compute / robot platform details.

SteveMacenski:
@vimalrajayyappan can you give more context / info?

SteveMacenski:
@vimalrajayyappan we'll have to close this ticket in about a week if we don't receive a response

fmrico:
Hi, 

While running the experiments with the pure pursuit in January, I experienced an effect that made the robot move slow, even if I configured the params to move faster. Maybe it is what @vimalrajayyappan describes.

I will try to remember the reason and the fix. I remember something related to the existence of the `odom`  topic. Do you remember @jginesclavero?

Best

vimalrajayyappan:
Yes, exactly friend :). Kindly help @fmrico and @SteveMacenski  :).

vimalrajayyappan:
[...]

PR #2474, opened 2021-07-23
Fix for the seg-fault that occurred when RPP was tested at high speed

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2332  |
| Robotic platform tested on | mp-400 from Neobotix |

---

* Fixed the segfault issue that occurred when the robot was set to drive at high speed using RPP. 

NA

---

NA

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -232,6 +232,7 @@ protected:
   std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
   nav2_costmap_2d::Costmap2D * costmap_;
   rclcpp::Logger logger_ {rclcpp::get_logger("RegulatedPurePursuitController")};
+  rclcpp::Clock::SharedPtr clock_;
 
   double desired_linear_vel_, base_desired_linear_vel_;
   double lookahead_dist_;
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -43,7 +43,7 @@ void RegulatedPurePursuitController::configure(
 {
   auto node = parent.lock();
   if (!node) {
-    throw std::runtime_error("Unable to lock node!");
+    throw nav2_core::PlannerException("Unable to lock node!");
   }
 
   costmap_ros_ = costmap_ros;
@@ -51,6 +51,7 @@ void RegulatedPurePursuitController::configure(
   tf_ = tf;
   plugin_name_ = name;
   logger_ = node->get_logger();
+  clock_ = node->get_clock();
 
   double transform_tolerance = 0.1;
   double control_frequency = 20.0;
@@ -314,7 +315,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
   // Collision checking on this velocity heading
   if (isCollisionImminent(pose, linear_vel, angular_vel)) {
-    throw std::runtime_error("RegulatedPurePursuitController detected collision ahead!");
+    throw nav2_core::PlannerException("RegulatedPurePursuitController detected collision ahead!");
   }
 
   // populate and return message
@@ -437,13 +438,13 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double 
 {
   unsigned int mx, my;
 
-  auto convertWorldToMap = costmap_->worldToMap(x, y, mx, my);
-  if (!convertWorldToMap) {
+  if (!costmap_->worldToMap(x, y, mx, my)) {
     RCLCPP_WARN_THROTTLE(
       logger_, *(clock_), 30000,
       "The dimensions of the costmap is smaller, proceed at your own risk");
     return false;
   }
+
   unsigned char cost = costmap_->getCost(mx, my);
 
   if (costmap_ros_->getLayeredCostmap()->isTrackingUnknown()) {
@@ -456,12 +457,14 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double 
 double RegulatedPurePursuitController::costAtPose(const double & x, const double & y)
 {
   unsigned int mx, my;
-  auto convertWorldToMap = costmap_->worldToMap(x, y, mx, my);
-  if (!convertWorldToMap) {
-    RCLCPP_WARN_THROTTLE(
-      logger_, *(clock_), 30000,
-      "The dimensions of the costmap is smaller, proceed at your own risk");
-    return 0.0;
+
+  if (!costmap_->worldToMap(x, y, mx, my)) {
+    RCLCPP_FATAL_ONCE(
+      logger_,
+      "The dimensions of the costmap is smaller, the controller failed and so the robot cannot proceed further");
+    throw nav2_core::PlannerException(
+            "RegulatedPurePursuitController detected poor configuration of local costmap!");
+    return std::numeric_limits<double>::max();
   }
 
   unsigned char cost = costmap_->getCost(mx, my);
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -461,7 +461,7 @@ double RegulatedPurePursuitController::costAtPose(const double & x, const double
   if (!costmap_->worldToMap(x, y, mx, my)) {
     RCLCPP_FATAL_ONCE(
       logger_,
-      "The dimensions of the costmap is smaller, the controller failed and so the robot cannot proceed further");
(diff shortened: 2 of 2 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -436,8 +436,14 @@ bool RegulatedPurePursuitController::isCollisionImminent(
 bool RegulatedPurePursuitController::inCollision(const double & x, const double & y)
 {
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
 
+  auto convertWorldToMap = costmap_->worldToMap(x, y, mx, my);
+  if (!convertWorldToMap) {
```
SteveMacenski, line 441: These can be simply replaced to `if (!costmap_->worldToMap(x, y, mx, my))` since you don't make use of `convertWorldToMap` anywhere else than that statement. Also below

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -232,6 +232,7 @@ class RegulatedPurePursuitController : public nav2_core::Controller
   std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
   nav2_costmap_2d::Costmap2D * costmap_;
   rclcpp::Logger logger_ {rclcpp::get_logger("RegulatedPurePursuitController")};
+  rclcpp::Clock::SharedPtr clock_;
```
SteveMacenski, line 235: Below `  logger_ = node->get_logger();` in L53 place the get_clock so that we get the clock from the node. Right now I don't think you initialize this anywhere...

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,7 +456,13 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
 double RegulatedPurePursuitController::costAtPose(const double & x, const double & y)
 {
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
+  auto convertWorldToMap = costmap_->worldToMap(x, y, mx, my);
+  if (!convertWorldToMap) {
```
SteveMacenski, line 460: This function is only used when the x,y represents our robot pose, this should never be out of bounds.

SteveMacenski (changes requested):
Good first stab, only a couple of simple tweaks and this can be merged

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: So I think these changes in this function can be reverted?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: Ah yeah ! That is a nice way to do it !

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
padhupradheep: I think it is not a good idea. If you remember the second issue that I brought up in #2332 , these changes actually solves that issue. 

>  This function is only used when the x,y represents our robot pose, this should never be out of bounds.

Explaining it further: With the parameters set in that issue, the current position of the robot can be sometimes outside of the local costmap. Then again, we have a seg fault. In order to avoid it, we check for the bounds.

SteveMacenski: Please give me an example flow of data that this would prevent?  What I see is:
- `costAtPose` is only used as an argument for `applyConstraints()`
- the `pose` given to `applyConstraints` is `pose`, which in that function I track to the argument `pose` https://github.com/ros-planning/navigation2/blob/8595b870dd66ef709101786a13eb86a5fc889e76/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L243
- that pose represents the robot's current pose -- so there's no look ahead applied and my assumption is that at all times the robot should be on the costmap

SteveMacenski: That seems.... bad. I know you set it to 1x1 to prove a point, but what if this was a more normal size like 5x5 or a more reasonable update rate like 5hz? We should cover the crash, you bring up a good point, but those are some extraordinarily poor configurations :laughing:. I think we just need to change the logging error message

SteveMacenski:
Any update?

padhupradheep:
> Any update?

Expect a commit for the above reviews, on this friday.

SteveMacenski:
Pull in master or rebase to get CI turning over, has to do with a stale branch from changes we made to sync with Rolling

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,7 +457,15 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
...
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
+
+  if (!costmap_->worldToMap(x, y, mx, my)) {
+    RCLCPP_FATAL_ONCE(
+      logger_,
+      "The dimensions of the costmap is smaller, the controller failed and so the robot cannot proceed further");
+    throw nav2_core::PlannerException(
+            "RegulatedPurePursuitController detected poor configuration of local costmap!");
+    return std::numeric_limits<double>::max();
```
SteveMacenski, line 467: Nothing after a `throw` is executed, so you can remove this return line.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,7 +457,15 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
 double RegulatedPurePursuitController::costAtPose(const double & x, const double & y)
 {
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
+
+  if (!costmap_->worldToMap(x, y, mx, my)) {
+    RCLCPP_FATAL_ONCE(
```
SteveMacenski, line 462: I would not make this a `ONCE`, this should be a `FATAL` warning every single time it gets triggered

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -436,7 +437,13 @@ bool RegulatedPurePursuitController::isCollisionImminent(
 bool RegulatedPurePursuitController::inCollision(const double & x, const double & y)
 {
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
+
+  if (!costmap_->worldToMap(x, y, mx, my)) {
+    RCLCPP_WARN_THROTTLE(
+      logger_, *(clock_), 30000,
+      "The dimensions of the costmap is smaller, proceed at your own risk");
```
SteveMacenski, line 444:
```suggestion
      "The dimensions of the costmap is too small to successfully check for collisions as far ahead as requested. Proceed at your own risk, slow the robot, or increase your costmap size.");
```

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,7 +457,15 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
 double RegulatedPurePursuitController::costAtPose(const double & x, const double & y)
 {
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
+
+  if (!costmap_->worldToMap(x, y, mx, my)) {
+    RCLCPP_FATAL_ONCE(
+      logger_,
+      "The dimensions of the costmap is smaller, the controller failed and so the robot cannot proceed further");
```
SteveMacenski, line 464:
```suggestion
      "The dimensions of the costmap is too small to fully include your robot's footprint, thusly the robot cannot proceed.");
```

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,7 +457,15 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
...
+    throw nav2_core::PlannerException(
+            "RegulatedPurePursuitController detected poor configuration of local costmap!");
```
SteveMacenski, line 466:
```suggestion
            "RegulatedPurePursuitController: Dimensions of the costmap are too small to encapsulate the robot footprint at current speeds!");
```

SteveMacenski (changes requested):
Mostly just wording changes to be more explicit

padhupradheep:
created new PR #2526

Issue #956, opened 2019-07-19
Re-enable runtime configuration of key parameters


crdelsey:
#### Feature description
The lifecycle changes disabled all runtime reconfiguration of parameters. However, certain key parameters need to be dynamically adjusted to allow users to easily tune the stack for their use cases.

These key parameters need to be identified and support added to make them changeable while in the active state.

bpwilcox:
![image](https://user-images.githubusercontent.com/22353511/63456175-31258800-c403-11e9-9cce-0a3f99160191.png)
Here is a class diagram I am proposing to address support for dynamic parameters in the stack.

bpwilcox:
https://github.com/ros2/rclcpp/pull/829 submitted to add additional support for parameter event subscriptions for dynamic parameter callbacks

SteveMacenski:
Major TODO list
- [x] DWB
- [ ] costmap layers / costmap_2d
- [ ] navfn **in progress** https://github.com/ros-planning/navigation2/pull/2181
- [ ] smac planner (2x)
- [ ] amcl
- [ ] regulated pure pursuit

Minor TODO list (lower impact, unlikely to be actually needed to change for tuning)
- [ ] waypoint follower
- [ ] BT navigator
- [ ] lifecycle manager
- [ ] Task servers: planner, recovery, controller server
- [ ] map server

doisyg:
Moving the discussion from https://github.com/ros-planning/navigation2/issues/2565

 
>If you could offer a code snippet of how one of the simple changes would look like, that would help. You don't have to do every parameter in some node, just a couple to get the idea of what you're proposing looks like. The changing parameters APIs of ROS 2 keeps changing so my guess is that wasn't on the mind when we started. I know there's an API for registering callbacks for individual parameters (which I think this is asking for) but there's also a parameter filter object floating around too https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1ParameterEventsFilter.html. There was also a design discussion around topic filters for parameters so that it only triggers callbacks for the node of interest instead of all nodes, but that's not tangible yet.

Short term solution of a callback triggered only when a parameter is changed on the hosting node:
In the header:
```
  // Dynamic parameters handler
  OnSetParametersCallbackHandle::SharedPtr dyn_params_handler;

  /**
   * @brief Callback executed when a paramter change is detected
   * @param parameters list of changed parameters
   */
  rcl_interfaces::msg::SetParametersResult
[...]

PR #2475, opened 2021-07-23
[SmacPlanner2D] make tolerance parameter dynamic

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/956|
| Robotic platform tested on | Wyca sim) |

---

I am starting to experiment with **SmacPlanner2D** and in order to work in our workflow, I need the tolerance parameter to be dynamic, which is what this PR is about.

---

Should we make other paremeters dynamic in **SmacPlanner2D** ? Though most of them would require re-initialization of several objects

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
@@ -88,6 +88,12 @@ public:
     const geometry_msgs::msg::PoseStamped & goal) override;
 
 protected:
+  /**
+   * @brief Callback executed when a paramter change is detected
+   * @param event ParameterEvent message
+   */
+  void on_parameter_event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
+
   std::unique_ptr<AStarAlgorithm<Node2D>> _a_star;
   GridCollisionChecker _collision_checker;
   std::unique_ptr<Smoother> _smoother;
@@ -113,12 +119,6 @@ protected:
   // Subscription for parameter change
   rclcpp::AsyncParametersClient::SharedPtr _parameters_client;
   rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr _parameter_event_sub;
-
-  /**
-   * @brief Callback executed when a paramter change is detected
-   * @param event ParameterEvent message
-   */
-  void on_parameter_event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
 };
 
 }  // namespace nav2_smac_planner
--- a/nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
@@ -100,6 +100,16 @@ protected:
   bool _downsample_costmap;
   rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>::SharedPtr _raw_plan_publisher;
   double _max_planning_time;
+
+  // Subscription for parameter change
+  rclcpp::AsyncParametersClient::SharedPtr parameters_client_;
+  rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr parameter_event_sub_;
+
+  /**
+   * @brief Callback executed when a paramter change is detected
+   * @param event ParameterEvent message
+   */
+  void on_parameter_event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
 };
 
 }  // namespace nav2_smac_planner
--- a/nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
@@ -18,6 +18,7 @@
 #include <memory>
 #include <vector>
 #include <string>
+#include <mutex>
 
 #include "nav2_smac_planner/a_star.hpp"
 #include "nav2_smac_planner/smoother.hpp"
@@ -100,10 +101,18 @@ protected:
   bool _downsample_costmap;
   rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>::SharedPtr _raw_plan_publisher;
   double _max_planning_time;
+  bool _allow_unknown;
+  int _max_iterations;
+  int _max_on_approach_iterations;
+  SearchInfo _search_info;
+  std::string _motion_model_for_search;
+  MotionModel _motion_model;
+  std::mutex _mutex;
+  rclcpp_lifecycle::LifecycleNode::WeakPtr _node;
 
   // Subscription for parameter change
-  rclcpp::AsyncParametersClient::SharedPtr parameters_client_;
-  rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr parameter_event_sub_;
+  rclcpp::AsyncParametersClient::SharedPtr _parameters_client;
+  rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr _parameter_event_sub;
 
   /**
    * @brief Callback executed when a paramter change is detected
--- a/nav2_smac_planner/src/smac_planner_2d.cpp
+++ b/nav2_smac_planner/src/smac_planner_2d.cpp
@@ -375,7 +375,7 @@ void SmacPlanner2D::on_parameter_event_callback(
     }
 
     // Re-Initialize costmap downsampler
-    if (reinit_a_star) {
+    if (reinit_downsampler) {
       if (_downsample_costmap && _downsampling_factor > 1) {
         auto node = _node.lock();
         std::string topic_name = "downsampled_costmap";
--- a/nav2_smac_planner/src/smac_planner_2d.cpp
+++ b/nav2_smac_planner/src/smac_planner_2d.cpp
@@ -25,6 +25,8 @@
 namespace nav2_smac_planner
 {
 using namespace std::chrono;  // NOLINT
+using rcl_interfaces::msg::ParameterType;
+using std::placeholders::_1;
 
 SmacPlanner2D::SmacPlanner2D()
 : _a_star(nullptr),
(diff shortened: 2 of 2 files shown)


nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
```diff
@@ -100,6 +100,16 @@ class SmacPlanner2D : public nav2_core::GlobalPlanner
   bool _downsample_costmap;
   rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>::SharedPtr _raw_plan_publisher;
   double _max_planning_time;
+
```
SteveMacenski, line 103: Hi,

Certainly appreciate this and helps get one of the algorithms done for run time reconfiguration. Something though is that it would be great if this also enabled runtime reconfiguration of any parameter here that can be reconfigured. Some of the parameters are going to be set on initialization and are therefore not reconfigurable, but some are used at runtime and can therefore be reconfigured.

SteveMacenski: - _downsampling_factor
- _max_planning_time

Seem like the only 2 I can find

SteveMacenski (changes requested):

nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
doisyg: `_downsampling_factor` is used to initialize the `CostmapDownsampler`
https://github.com/ros-planning/navigation2/blob/5c61644651c4eab882b042e073d0f5964f03a501/nav2_smac_planner/src/smac_planner_2d.cpp#L139-L145

I see only `_max_planning_time` okay to modify at runtime in the same way as `tolerance`.
But we could allow reconfigurability of all parameters if we trigger re-initialization of the relevant objects. But it will require a mechanism ensuring that no planning is currently being computed. What do you think ?

SteveMacenski: Its also used in the call to down sample https://github.com/ros-planning/navigation2/blob/b519dece8b5f1268e8bb16ca9b1b04fa06c75e34/nav2_smac_planner/src/smac_planner_2d.cpp#L215 

Frankly, I don't even remember what I was thinking where I wanted to have that in the configuration argument if its also in the `downsample()` function argument. It might be able to be safely removed.

SteveMacenski: > But we could allow reconfigurability of all parameters if we trigger re-initialization of the relevant objects

Certainly, yes. We could add a mutex lock to prevent planning during these reconfiguration phases.

doisyg: Alright, I will take care of it with mutex lock when I will have a minute.

SteveMacenski: curious -- are you guys using Smac 2D?

doisyg: I started evaluating it (on a real robot/environment) today ;)
Not yet in production though but it is a possibility

SteveMacenski: Let me know if that changes. The Hybrid-A* version is actually faster in many cases -- and it is kinematically feasible so no plans starting at a 60 deg angle from your starting orientation.

doisyg:
Last commit adds full dynamic configuration, though I am not a huge fan of the nested if / else if on the on_parameter_event_callback.

nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
doisyg: I thought the 2D version was more appropriate to small circular base ?
Plus at first inspection, there is no tolerance parameter on the Hybrid-A* version.
But I am already a fan of the path cleanness the 2D version produce, despite the reduced efficiency (we compute global path only once in our use case and do recomputation only in case of local planner/controller failure)

SteveMacenski: > I thought the 2D version was more appropriate to small circular base ?

Sure, but its not kinematically feasible, the Hybrid-A* planner is. The "tolerance" isn't as important because if there is a way to get to the goal, the analytic expansions make it much easier to make sure its possible. Use what is better for your situation, but I don't think you should dismiss Hybrid-A* out of hand. No more 'rotate to heading' to follow a path, the path will always start from the robot's heading smoothly.

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -285,6 +293,100 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
...
+      _a_star->initialize(
+        _allow_unknown,
+        _max_iterations,
+        _max_on_approach_iterations,
+        0.0 /*unused for 2D*/,
+        1.0 /*unused for 2D*/);
+    }
+
+    // Re-Initialize costmap downsampler
+    if (reinit_a_star) {
```
SteveMacenski, line 378: Should this be reinit downsampler?

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -285,6 +293,100 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
   return plan;
 }
 
+void SmacPlanner2D::on_parameter_event_callback(
```
SteveMacenski, line 296: There's a slightly cleaner way you can handle this that was added in Galactic and newer
- https://github.com/ros2/demos/pull/486/files
- https://docs.ros.org/en/rolling/Tutorials/Monitoring-For-Parameter-Changes-CPP.html

Using the `ParameterEventHandler`, you can add callbacks for specific parameters rather than 1 big callback to rule them all. While I think this is probably the right direction, I would be open if you thought differently before asking you to do that instead. But if you agree that would be cleaner / nicer to work with, that's what I would suggest doing.

SteveMacenski: It looks like you might have to store the callback handle (?) which could make this annoying to deal with. I'm uncertain if you have to store that if it that would be nice in case you needed it.

SteveMacenski (changes requested):
LGTM, I see what I think is a typo and a question on if we should do it this way or using the newer ParameterEventHandler API

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: Definitely a typo, fixed, thanks!

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: Hum, it doesn't seem ideal either: need to store one handle per custom callback plus in our case we would have to define 3 custom callback: 1. for tolerance, 2. for the a_star group of params, and 3. for the downsampler group of param.
I would leave it like that. I am not a super fan of the current approach but the new one doesn't satisfy me more

nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
doisyg: Okay, I will have a look a some point. For our current use case (circular robot), 'rotate to heading' is perfectly acceptable.

nav2_smac_planner/src/smac_planner_2d.cpp
SteveMacenski: OK sounds good

nav2_smac_planner/include/nav2_smac_planner/smac_planner_2d.hpp
```diff
@@ -100,6 +101,24 @@ class SmacPlanner2D : public nav2_core::GlobalPlanner
...
+  // Subscription for parameter change
+  rclcpp::AsyncParametersClient::SharedPtr _parameters_client;
+  rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr _parameter_event_sub;
+
+  /**
+   * @brief Callback executed when a paramter change is detected
+   * @param event ParameterEvent message
+   */
+  void on_parameter_event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
```
SteveMacenski, line 121: Place this function above the variables

SteveMacenski (changes requested):

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -285,6 +293,100 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
...
+            "valid options are MOORE, VON_NEUMANN, DUBIN, REEDS_SHEPP.",
+            _motion_model_for_search.c_str());
+        }
+      }
+    }
+  }
+
+  // Re-init if needed with mutex lock (to avoid re-init while creating a plan)
+  if (reinit_a_star || reinit_downsampler) {
+    std::lock_guard<std::mutex> lock_reinit(_mutex);
```
SteveMacenski, line 364: Last thing I see, since you are changing a couple of values that can actually change live, this mutex lock needs to happen either earlier in the function or in those specific `if` statements above. As we noted in the thread above, there are 2 (I believe) params that can be changed live and used without reinitalization. So its possible this function could be in the middle of execution to change that value while trying to be read for use.

SteveMacenski (changes requested):

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: I believe the two parameters that can be changed without re-init are `_tolerance` and `_max_planning_time`, all the others are needed either for `_a_star` or `_costmap_downsampler`. Hence by putting the lock there, I use it only if a re-init is needed.

SteveMacenski: huh? If you change `_tolerance` and `_max_planning_time` you don't need to reinit, yes. But since those are *independent* of the A* / costmap downsampler, they can be changed in this callback at the same time that a planning request is processing. So it could be trying to access that memory at the same time its being overwritten. So we need a lock in the `if` statement where you set those 2 variables in particular since they are not constrained by the A* / downsampler objects that are locked already.

Does that make sense?

- Planner is called at the same time as a reconfigure service
- If any others are changed, doesn't matter since the planning request will finish before giving up the mutex for the reinitialization to happen
- If this is changed, it can be corrupt a `get` while this is `setting` them, since used live in the code and not just within the A*/downsampler objects.

doisyg: All right, I completely overlooked this potential race condition.
Basically all params needs protection so I placed the lock at the top of the param change callback

SteveMacenski (approved):

SteveMacenski:
Last thing is a new section in the migration guide mentioning this has dynamic reconfigure parameters now https://navigation.ros.org/migration/Galactic.html

doisyg:
https://github.com/ros-planning/navigation.ros.org/pull/192

PR #2479, opened 2021-07-27
Reduce controller server nodes

gezp:

---

## Basic Info

-------

As described in [#816](https://github.com/ros-planning/navigation2/issues/816), `rclcpp_node_` in `class ControllerServer ` need be removed, you can find more details(other nodes which need be removed) in [#816 (comment)](https://github.com/ros-planning/navigation2/issues/816#issuecomment-876061677)

after [PR#2459](https://github.com/ros-planning/navigation2/pull/2459) merged, we can create `SimpleActionServer` with argument `spin_thread` which could spin server with dedicated thread.

- set `spin_thread` true when create `action_server_` .

## Test
colcon test
```bash
colcon test --packages-select nav2_controller
```
* success
---

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/nav2_controller.cpp
+++ b/nav2_controller/src/nav2_controller.cpp
@@ -32,7 +32,7 @@ namespace nav2_controller
 {
 
 ControllerServer::ControllerServer()
-: LifecycleNode("controller_server", "", true),
+: nav2_util::LifecycleNode("controller_server", ""),
   progress_checker_loader_("nav2_core", "nav2_core::ProgressChecker"),
   default_progress_checker_id_{"progress_checker"},
   default_progress_checker_type_{"nav2_controller::SimpleProgressChecker"},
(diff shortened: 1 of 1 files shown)


nav2_controller/src/nav2_controller.cpp
```diff
@@ -266,7 +270,6 @@ ControllerServer::on_cleanup(const rclcpp_lifecycle::State & state)
   odom_sub_.reset();
   vel_publisher_.reset();
   speed_limit_sub_.reset();
-  action_server_.reset();
```
SteveMacenski, line 269: Readd this reset, its important for lifecycle bring up and shut downs

SteveMacenski (changes requested):

nav2_controller/src/nav2_controller.cpp
gezp: the line265 and line269 are same, i just remove the duplicated one.

nav2_controller/src/nav2_controller.cpp
```diff
@@ -195,8 +195,12 @@ ControllerServer::on_configure(const rclcpp_lifecycle::State & state)
 
   // Create the action server that we implement with our followPath method
   action_server_ = std::make_unique<ActionServer>(
-    rclcpp_node_, "follow_path",
-    std::bind(&ControllerServer::computeControl, this));
+    shared_from_this(),
+    "follow_path",
+    std::bind(&ControllerServer::computeControl, this),
+    nullptr,
+    std::chrono::milliseconds(500),
```
SteveMacenski, line 202: I didn't ask in the Planner PR and I should have -- is this 500ms the same behavior as before or did this value change?

gezp: before PR, the default value was used, and 500ms is the default value, so it's same as before.

SteveMacenski (approved):

Issue #2467, opened 2021-07-21
Set BT::NodeStatus based on service result msg (class BtServiceNode)


philison:
**Required Info:**

- Operating System:
  - Ubuntu 20.04.2 LTS
- ROS2 Version:
  - Galactic binaries
- Version or commit hash:
  - 5a7027536068183602f10b419bd711fafdad6784
- DDS implementation:
  - 

#### Feature description

It is the goal to set the BT::NodeStatus of a behavior tree leave node based on the result send from the service server.
More specifically, I tried to implement a _Condition-Node_ that makes a request to a service server via the ROS2 service interface and sets its BT::NodeStatus based on the result.. Therefore, the BtServiceNode class ([_bt\_service\_node.cpp_](https://github.com/ros-planning/navigation2/blob/c417e2fd267e1dfa880b7ff9d37aaaa7b5eab9ca/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp)) seemed to me to be the best suited class to use. However, in contrast to the BtActionNode class ([_bt\_action\_node.cpp_](https://github.com/ros-planning/navigation2/blob/0e470415f0d6b2672e350e2cc9271002b143cb10/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp)) it missed the virtual _on\_success()_ function, that can be used to define the custom BT::NodeStatus behavior based on the service's answer. The current BtServiceNode class sets the BT::NodeStatus solely based on the success of the service request.

I am not sure if I missed a class that implements this feature, if there is a simpler solution or if I should just use the BT::ConditionNode base class with a normal service client after all. 

#### Implementation considerations

The proposed implementation simply adds a call to the virtual _on\_success()_ function in the _check\_future()_ function (Similar to the BtActionNode).

The changes do not change do not affect the default behavior of the BtServiceNode as the default definition of _on\_success()_ does still simply return BT::NodeStatus::SUCCESS.

If the proposed implementation is an appropriate modification I could create a pull request.

Changes:
https://github.com/angsa-robotics/navigation2/blob/033cdc49b4c106f2747772f2081fca47bee6e9ec/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L150-L187

Original:
https://github.com/ros-planning/navigation2/blob/0764e5af5b23ea7a08629876792a8da1465ef261/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L145-L171

[...]

PR #2481, opened 2021-07-27
Modify the BtServiceNode to include an on_success call.

philison:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2467 |
| Robotic platform tested on | custom differential drive robot |

---

* Modified the `bt_service_node` to include an `on_success()` function, to process the service answer. 
* The additional `on_success()` function can be used to process the results send from the service server, to, for example, set the BT::NodeStatus of the behavior tree leave node

* The `on_success()` function was added. However, I am not sure if this needs to be documented somewhere. 

---

* So far I can't think of an additional on-function, which could be useful and which is not yet implemented.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -144,10 +144,10 @@ public:
 
   /**
    * @brief Function to perform some user-defined operation upon successful
-   * completion of the action. Could put a value on the blackboard.
-   * @return BT::NodeStatus Returns SUCCESS by default, user may override return another value
+   * completion of the service. Could put a value on the blackboard.
+   * @return BT::NodeStatus Returns SUCCESS by default, user may override to return another value
    */
-  virtual BT::NodeStatus on_success()
+  virtual BT::NodeStatus on_completion()
   {
     return BT::NodeStatus::SUCCESS;
   }
@@ -168,7 +168,7 @@ public:
       rc = callback_group_executor_.spin_until_future_complete(future_result_, server_timeout_);
       if (rc == rclcpp::FutureReturnCode::SUCCESS) {
         request_sent_ = false;
-        BT::NodeStatus status = on_success();
+        BT::NodeStatus status = on_completion();
         return status;
       }
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -168,7 +168,7 @@ public:
       rc = callback_group_executor_.spin_until_future_complete(future_result_, server_timeout_);
       if (rc == rclcpp::FutureReturnCode::SUCCESS) {
         request_sent_ = false;
-        BT::NodeStatus status = on_success();   
+        BT::NodeStatus status = on_success();
         return status;
       }
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -143,6 +143,16 @@ public:
   }
 
   /**
+   * @brief Function to perform some user-defined operation upon successful
+   * completion of the action. Could put a value on the blackboard.
+   * @return BT::NodeStatus Returns SUCCESS by default, user may override return another value
+   */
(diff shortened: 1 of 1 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -142,6 +142,16 @@ class BtServiceNode : public BT::ActionNodeBase
   {
   }
 
+  /**
```
SteveMacenski, line 145: Please add to migration guide a new section on the Galactic page that we added this virtual function to allow for setting of codes after services return

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -142,6 +142,16 @@ class BtServiceNode : public BT::ActionNodeBase
   {
```
SteveMacenski, line 142: Linting failures:

```
whitespace/end_of_line [4] (/opt/overlay_ws/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp:171) - nav2_behavior_tree.cpplint

Line ends in whitespace.  Consider deleting these extra spaces.

include/nav2_behavior_tree/bt_service_node.hpp - nav2_behavior_tree.uncrustify

Diff with 5 lines
--- include/nav2_behavior_tree/bt_service_node.hpp
+++ include/nav2_behavior_tree/bt_service_node.hpp.uncrustify
@@ -171 +171 @@
-        BT::NodeStatus status = on_success();   
+        BT::NodeStatus status = on_success();

```

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -142,6 +142,16 @@ class BtServiceNode : public BT::ActionNodeBase
...
   }
 
+  /**
+   * @brief Function to perform some user-defined operation upon successful
+   * completion of the action. Could put a value on the blackboard.
+   * @return BT::NodeStatus Returns SUCCESS by default, user may override return another value
+   */
+  virtual BT::NodeStatus on_success()
```
SteveMacenski, line 150: How do you feel about on_completion instead? It might make more sense from a application developer's perspective since you don't want to imply that the service necessarily did succeed just by returning

SteveMacenski (changes requested):
On a higher level, I think this should be merged in. Just a few things to clean up

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
philison: I totally agree with you. I just named it `on_success()` to be consistent with the implementation in other BT-Nodes, like the BtActionNode: https://github.com/ros-planning/navigation2/blob/2035819e0d69fb2e5aea80dd06f668911dca9048/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L146
Therefore, a change there, might be necessary to keep the implementations consistent ?

philison: I changed it to on_completion in the new commit.

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
philison: Are you referring to this [file](https://github.com/ros-planning/navigation.ros.org/blob/1278dcc676a8ea2c1b9207351a49e68ccb50f1d2/migration/Foxy.rst) ?
Would the best way to edit it be to create a fork and than PR the changes ?

SteveMacenski: https://github.com/ros-planning/navigation.ros.org/blob/master/migration/Galactic.rst this file

SteveMacenski (approved):
LGTM, will merge once the Galactic docs PR is opened to match! Thanks for the contribution!

philison:
I opened the the [Galactic docs PR](https://github.com/ros-planning/navigation.ros.org/pull/195).
Thanks for your support and thanks to @karl-schulz for getting me to write my first proper contribution. It was a pleasure.

SteveMacenski:
One small change to that PR to remove a build warning and then OK to merge!

SteveMacenski:
Thanks for the contribution!

PR #2483, opened 2021-07-28
Reduce amcl_node nodes

gezp:

---

## Basic Info

-------

As described in [#816](https://github.com/ros-planning/navigation2/issues/816), `rclcpp_node_` in `class AmclNode` need be removed, you can find more details(other nodes which need be removed) in [#816 (comment)](https://github.com/ros-planning/navigation2/issues/816#issuecomment-876061677)

- After the PR (https://github.com/ros2/message_filters/pull/59) was merged, we can use `message_filters::Subscriber` in `LifecycleNode`. and update `tools/underlay.repos` to sync the feature.
- There is no blocked callback in `AmclNode`, so just use `shared_from_this()` instead of `rclcpp_node_` for `Transform` and `MessageFilter`

# Test

colcon test
```bash
colcon test --packages-select nav2_amcl

```

- success

---

- i notice that the constructor of `TransformListener` has a argument `spin_thread` with default value `true`, which create a internal node to subscribe tf topic and a dedicated thread to spin this node. 
- i think  the best way is to use new callback group and executor with dedicated thread  so that we could avoid creating internal node which is overhead. i will open a issue for `ros2/geometry` to discuss these.

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -105,6 +105,12 @@ protected:
   // respond until we're in the active state
   std::atomic<bool> active_{false};
 
+  // Dedicated callback group and executor for services and subscriptions in AmclNode,
+  // in order to isolate TF timer used in message filter.
+  rclcpp::CallbackGroup::SharedPtr callback_group_;
+  rclcpp::executors::SingleThreadedExecutor::SharedPtr executor_;
+  std::unique_ptr<nav2_util::NodeThread> executor_thread_;
+
   // Pose hypothesis
   typedef struct
   {
@@ -165,7 +171,8 @@ protected:
    * @brief Initialize incoming data message subscribers and filters
    */
   void initMessageFilters();
-  std::unique_ptr<message_filters::Subscriber<sensor_msgs::msg::LaserScan>> laser_scan_sub_;
+  std::unique_ptr<message_filters::Subscriber<sensor_msgs::msg::LaserScan,
+    rclcpp_lifecycle::LifecycleNode>> laser_scan_sub_;
   std::unique_ptr<tf2_ros::MessageFilter<sensor_msgs::msg::LaserScan>> laser_scan_filter_;
   message_filters::Connection laser_scan_connection_;
 
--- a/nav2_amcl/src/amcl_node.cpp
+++ b/nav2_amcl/src/amcl_node.cpp
@@ -59,7 +59,7 @@ namespace nav2_amcl
 using nav2_util::geometry_utils::orientationAroundZAxis;
 
 AmclNode::AmclNode(const rclcpp::NodeOptions & options)
-: nav2_util::LifecycleNode("amcl", "", true, options)
+: nav2_util::LifecycleNode("amcl", "", false, options)
 {
   RCLCPP_INFO(get_logger(), "Creating");
 
@@ -236,7 +236,8 @@ nav2_util::CallbackReturn
 AmclNode::on_configure(const rclcpp_lifecycle::State & /*state*/)
 {
   RCLCPP_INFO(get_logger(), "Configuring");
-
+  callback_group_ = create_callback_group(
+    rclcpp::CallbackGroupType::MutuallyExclusive, false);
   initParameters();
   initTransforms();
   initParticleFilter();
@@ -245,7 +246,9 @@ AmclNode::on_configure(const rclcpp_lifecycle::State & /*state*/)
   initPubSub();
   initServices();
   initOdometry();
-
+  executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+  executor_->add_callback_group(callback_group_, get_node_base_interface());
+  executor_thread_ = std::make_unique<nav2_util::NodeThread>(executor_);
   return nav2_util::CallbackReturn::SUCCESS;
 }
 
@@ -316,6 +319,8 @@ nav2_util::CallbackReturn
 AmclNode::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
 {
   RCLCPP_INFO(get_logger(), "Cleaning up");
+
+  executor_thread_.reset();
 
   // Get rid of the inputs first (services and message filter input), so we
   // don't continue to process incoming messages
@@ -1426,13 +1431,14 @@ AmclNode::initTransforms()
   RCLCPP_INFO(get_logger(), "initTransforms");
 
   // Initialize transform listener and broadcaster
-  tf_buffer_ = std::make_shared<tf2_ros::Buffer>(rclcpp_node_->get_clock());
+  tf_buffer_ = std::make_shared<tf2_ros::Buffer>(get_clock());
   auto timer_interface = std::make_shared<tf2_ros::CreateTimerROS>(
-    rclcpp_node_->get_node_base_interface(),
-    rclcpp_node_->get_node_timers_interface());
+    get_node_base_interface(),
+    get_node_timers_interface(),
+    callback_group_);
   tf_buffer_->setCreateTimerInterface(timer_interface);
   tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
-  tf_broadcaster_ = std::make_shared<tf2_ros::TransformBroadcaster>(rclcpp_node_);
+  tf_broadcaster_ = std::make_shared<tf2_ros::TransformBroadcaster>(shared_from_this(), true);
 
   sent_first_transform_ = false;
   latest_tf_valid_ = false;
@@ -1442,11 +1448,18 @@ AmclNode::initTransforms()
 void
 AmclNode::initMessageFilters()
 {
-  laser_scan_sub_ = std::make_unique<message_filters::Subscriber<sensor_msgs::msg::LaserScan>>(
-    rclcpp_node_.get(), scan_topic_, rmw_qos_profile_sensor_data);
+  auto sub_opt = rclcpp::SubscriptionOptions();
+  sub_opt.callback_group = callback_group_;
+  laser_scan_sub_ = std::make_unique<message_filters::Subscriber<sensor_msgs::msg::LaserScan,
+      rclcpp_lifecycle::LifecycleNode>>(
+    shared_from_this(), scan_topic_, rmw_qos_profile_sensor_data, sub_opt);
 
   laser_scan_filter_ = std::make_unique<tf2_ros::MessageFilter<sensor_msgs::msg::LaserScan>>(
-    *laser_scan_sub_, *tf_buffer_, odom_frame_id_, 10, rclcpp_node_, transform_tolerance_);
+    *laser_scan_sub_, *tf_buffer_, odom_frame_id_, 10,
+    get_node_logging_interface(),
+    get_node_clock_interface(),
+    transform_tolerance_);
+
 
   laser_scan_connection_ = laser_scan_filter_->registerCallback(
     std::bind(


SteveMacenski:
@ruffsl it looks like the underlay failed to build and use the message filters package in the underlay.repos file. I see in the CI job it recognizes that its there, but I don't see any logging that it was built

nav2_amcl/include/nav2_amcl/amcl_node.hpp
```diff
@@ -57,6 +57,9 @@ namespace nav2_amcl
  */
 class AmclNode : public nav2_util::LifecycleNode
 {
+  template<class M>
+  using Subscriber = message_filters::Subscriber<M, rclcpp_lifecycle::LifecycleNode>;
```
SteveMacenski, line 61: Lets use this in line instead of `using` since its only used 1 time and I don't want people to glance over that code thinking its just a normal subscription

ruffsl:
The underlay appears to have built correctly, but it seems that a dependency was added without also amending the respective package.xlm and CMake files. Also I think the message filter package already ships with rolling, so you could leave it commented out and the repos file, and let rosdep install it from apt for the overlay workspace.

SteveMacenski:
> There is no blocked callback in AmclNode, so just use shared_from_this() instead of rclcpp_node_ for Transform and MessageFilter

AMCL does have the callback for the subscriber, `laserReceived`which [does a bit of work](https://github.com/ros-planning/navigation2/blob/1ef65cbb8e672dbbeadbecefb9692b9f139185d3/nav2_amcl/src/amcl_node.cpp#L624-L735). 

> i think the best way is to use new callback group and executor with dedicated thread so that we could avoid creating internal node which is overhead. i will open a issue for ros2/geometry to discuss these.

Agreed, for now, I think its worth letting TF do its own thing and control its own node. I looked over these files https://github.com/ros2/geometry2/blob/ros2/tf2_ros/src/transform_listener.cpp#L68-L91 and I think there's a really reasonable change that can be made here to solve our problems.
- Have the TransformListener take in a node in the constructor. 
- If spin_thread, then create the executor and handle all the same way
- If not spin_thread, then its on the application to spin

gezp:
@ruffsl could you give me more details? i don't know how to fixed the CI error.

ruffsl:
> What do you mean? No new packages were added. We just want to use a newer master version of message_filters until it is synced with Rolling in order to continue moving this ticket forward. Changes were made we want to exploit.

Ah, ok. I'm on my phone, so I didn't browes the code outside of the PR changes, and just assumed that the additional package to the repos file was still undelared in the manifests. Never the less, the underlay seems to be building fine:

```
Starting >>> message_filters
]0;colcon cache [0/1 done] [1 ongoing]Finished <<< message_filters [0.01s]
]0;colcon cache [1/1 done] [0 ongoing]
Summary: 1 package finished [0.22s]
BUILD_UNFINISHED: message_filters
BUILD_FAILED:
BUILD_INVALID: message_filters
BUILD_PACKAGES: message_filters
Starting >>> message_filters
]0;colcon build [0/1 done] [1 ongoing][Processing: message_filters]
[Processing: message_filters]
[Processing: message_filters]
Finished <<< message_filters [1min 47s]
]0;colcon build [1/1 done] [0 ongoing]
Summary: 1 package finished [1min 48s]
```

@gezp , what distro or source are you developing this PR with locally? Are you using the latest Rolling release, or are you building everything from ros2 master? Check if the Nav2 project Dockerfile builds fine locally for your PR:

```
docker build --tag my/navigation2:tag --target tester .
```

wjwwood:
I don't know if tf2 has been updated to make use of the fact that you can add callback groups to dedicated executors yet. You might want to bring it up on the geometry2 issue tracker. Or provide a pull request to that effect (rather than it having its own node). I don't use tf2 often enough to answer your architectural question about the buffer interface and how it should or shouldn't depend on ROS.

SteveMacenski:
Ok @gezp can you open the PR to implement what I describe above for the listener and open a ticket on the listener / buffer node usages? We can continue the discussion with Tully over there. 

For now, lets leave TF alone to do its own thing and fix it over there instead.

gezp:
@SteveMacenski I openedn a ticket https://github.com/ros2/geometry2/issues/440 ,and https://github.com/ros2/geometry2/pull/442 for listener.

gezp:
@ruffsl  Thank you, i will wait for the PR (https://github.com/ros/rosdistro/pull/30350) to be merged so that i could use the feature which i need on the latest Rolling release.

SteveMacenski:
@gezp those don't get released to rolling immediately, it requires a sync. The last sync was only a couple of weeks ago so it might be more than another month before we get another one. We should add the master to underlay and get this building so we can keep moving forward

gezp:
I guess the `message filter`  is a basic package  in ros so we can't overlay this package, but  i use a weird way so that i could use the newest `message filter`  from source on my local computer.

I install ROS2 Galactic form binary in my local computer , and delete `message filter` package in `/opt/ros/galactic`  before `source /opt/ros/galactic/local_setup.bash`,  then i git clone the newest `message filter`  from github, and build it so that i can use it in `nav2`. anyway i can build nav2 with this PR on my local computer with the newest `message filter` package, but i don't think this is a good way, and i'm not sure  if there is any problem after i delete files.

@ruffsl  i can't wait for next ROS rolling version ,and i don't  know how to pass CI , could you give me some suggestions?

ruffsl:
> is a basic package in ros so we can't overlay this package

I'm not sure I understand why the message filter package can't be overlay'ed in this particular case. It being a core package shouldn't it self prevent this. One can overload an existing packe by including it in a workspace, where upon setup, the package sourced from the top most overlay should take environmental president over any duplicates in the underlay workspaces.

Complications can occur when trying to link to dependencies that were compiled against a different version of the common dependency from a deeper underlay, but in this case, from how you described your current development with monkey patching galactic and overlay confined to navigation2, that doesn't seem to be an issue already.

Is there anything else you're including in your underlay besides the master branch of the message filter package, such as master branches of the message filter package's own dependencies?

gezp:
> Is there anything else you're including in your underlay besides the master branch of the message filter package, such as master branches of the message filter package's own dependencies?

I only put `message_filters` in my underlay workspace locally. 

currently, I try use Docker to build nav2. I add `message_filters` in `underlay.repos` and add some cmds before [line70](https://github.com/ros-planning/navigation2/blob/main/Dockerfile#L70) in Dockerfile 
```Dockerfile
# berore . /opt/ros/$ROS_DISTRO/setup.sh
# remove message_filters
RUN rm -rf /opt/ros/rolling/include/message_filters && \
    rm -rf /opt/ros/rolling/share/message_filters &&\
    rm -rf /opt/ros/rolling/libmessage_filters.so 
```
then, i can build my PR successfully. so do i need change CI ? or do you have a better way? @ruffsl

SteveMacenski:
mhm, that is complicated. I think the only real solution would be building it all from absolute scratch and replace the message filters with our branch at that point. That's probably not realistic for us to do in CI, so the only realistic option I see is to wait for a new rolling sync on this PR and just move on to other  topics in the meantime while we wait.

SteveMacenski:
https://discourse.ros.org/t/preparing-for-rolling-sync-2021-08-05/21701

Good news! A sync is happening tomorrow

gezp:
the new release of ROS Rolling is available [ros rolling/2021-08-06](https://discourse.ros.org/t/new-packages-for-ros-2-rolling-ridley-2021-08-06/21737) , but the docker image `ros:rolling` is not updated (https://hub.docker.com/_/ros) , i noticed you are contributor of [osrf/docker_images](https://github.com/osrf/docker_images), could you help me let CI use the latest ROS Rolling? @ruffsl

ruffsl:
The CI image should be kept up to date using this action that checks for
apt upgrades.

https://github.com/ros-planning/navigation2/runs/3277796714?check_suite_focus=true

I'm not sure why this hasn't triggered the image rebuild in the next step,
I suspect there might still be a syntax issue with bash or GitHub action
error in the config file.

SteveMacenski:
This is building now -- but I think we are still blocked here with respect to TF or no?

gezp:
i don't think  we should wait TF. because tf should spin with self's thread (to avoid delay or block), we don't consider the implementation of tf (use internal Node or callback group+executor). if TF is updated, we just need a small update in Nav2 (use another constructor of TF listener)

SteveMacenski:
Wouldn't the services / map / initial pose callbacks block the laser scan thread? As this PR stands.

I think I need to revisit the TF code. I'm not concerned about giving the buffer the clock from this node, but I do want to understand what the create Timer ROS function is doing with the timer / base interface.

gezp:
there is no block in services / map / initial pose callbacks 
services:
* callback of `global_loc_srv` _ (`globalLocalizationCallback`): call `pf_init_model()`  and set some variables, so the callback won't be blocked
* callback of `nomotion_update_srv_` (`nomotionUpdateCallback`) : only set a variable, so the callback won't be blocked

subscriptions:
* callback of  `initial_pose_sub_` (`initialPoseReceived`): call `handleInitialPose()`  , it contains `lookupTransform()`, but tf use internal nodes with dedicated thread, so this callback won't be blocked
* callback of `map_sub_` (`mapReceived` ) : call `handleMapMessage()` (free map memory and convert msg to map) , so the callback won't be blocked

SteveMacenski:
I'm not what you mean by "won't be blocked" if its deallocating memory and copying over an entire map, that's not necessarily a fast process. That could take ~100ms for a large map which is "blocking" to me. Let me look at this tomorrow. TF is sensitive so delaying messages that much could start with a bunch of warnings thrown all over the place from both the TF subscriber as well as internally to the message filter rejecting laser scans because the TF buffer didn't have the messages yet due to blockages. I'm far more concerned about TF than anything else, one of the reasons I want TF fixed up first and foremost.

gezp:
> I'm not what you mean by "won't be blocked" 

i mean there is no deadlock, but there maybe is delay. 

> TF is sensitive so delaying messages

 tf use a dedicated thread (use internal Node), so tf won't have delay.

but `laser_scan_sub_` of MessageFilter maybe have a delay, i don't know whether `laser_scan_sub_`  is sensitive so delaying messages, if so, there are two options to avoid delay
* (1) create a separated callback group and executor with dedicated thread  for `laser_scan_sub_` 
* (2) create a separated callback group  for `laser_scan_sub_`  , and  use multi threaded executor for `AmclNode`

according to the previous experience, option1 may be better , what do you think about ? @SteveMacenski

gezp:
> I think I need to revisit the TF code. I'm not concerned about giving the buffer the clock from this node and the TF broadcaster is just making a publisher with the node, but I do want to understand what the create Timer ROS function is doing with the timer / base interface.

the `timer_interface` is used by `Buffer::waitForTransform()`  , which is asynchronous and will return `future`. if we use wait_for(future), the timer will be used for timeout to avoid waiting forever.
i noticed `Buffer::waitForTransform()` is used in `tf2_ros::MessageFilter` , but i don't look at the code to understand how it works.

SteveMacenski:
So I see a few things here that seem like they need to be flushed out / justified more
- The TF2 buffer creates a timer using this node. My understanding of timers in ROS 2 (which could be wrong) is that they are created like callbacks on the clock so its essentially treated as a callback like a subscriber. This is usually really nice if you want to run something periodically that also has subscribers pouring in data. In the TF situation though, the TF callbacks could be blocked by, or blocking laser scan data 
- The services can then be called which would block TF from getting its information on the timer 
- Then the laser scan callbacks via the message filters wouldn't have the messages required and drop laser scans

gezp:
timer and tf listener use different thread  (diffenrent node) to spin in parallel.

* the callback of timer：use `cancelTransformableRequest()`  to cancel TransformableRequest which is timeout.
* the callback of tf listener : use `setTransform()` to set new transform  (invoke `testTransformableRequests()` in the end)

mutex `transformable_requests_mutex_` is used in `cancelTransformableRequest()` and `testTransformableRequests()` ([geometry2/tf2/src/buffer_core.cpp#L1349](https://github.com/ros2/geometry2/blob/bd125c5af32e2c63c70b7e0ba88e57094714b9b8/tf2/src/buffer_core.cpp#L1344) ), so there is no race condition.  

in my opinion, current PR is not worse than the previous , but there is a problem that laser scan callbacks maybe block tf listener, which existed before, i don't know how to solve it. what need i do in this PR next ?

SteveMacenski:
Yeah it was an issue before, but its different now because the services are involved. If you followed my example, the main concern I had was longer running things (services) blocking long enough to cause a flood of message filter warnings / errors. The TF/laser scan issue is also there, but isn't concerning to me because what you mentioned, its been like that for about 2 years. The no-update service is pretty trivial and not an issue but the global relocalization one is a bit more so. 

Can we separate the services onto another executor? Do you think that's a rational move (it would probably need then some shared resource protections)? Not the best solution in my mind though. 

The best solution to me is to make TF isolated from the rest of the node's interfaces so that it isn't being interfered with based on user services or data callbacks. If that was the case, then we shouldn't have an issue with message filters not being able to transform and we'd just be dropping laser scans in the case of long-running services which is the right behavior.

gezp:
> Can we separate the services onto another executor? Do you think that's a rational move (it would probably need then some shared resource protections)? Not the best solution in my mind though.

In my understanding，i want to do like this.

* let all services and subscriptions in `AmclNode` use a new callback group, and create a dedicated thread to spin it.
* let timer use the default callback group

all services and subscriptions use a same callback group, so we needn't shared resource protections. what do you think about this approach? @SteveMacenski 

i'm not sure i understand what you say, the callback of laser scan callback may be handled in tf (in another thread), which is existed before, so do i need protect shared resource between laser scan callback and others callbacks in `AmclNode`?

SteveMacenski:
yes, though as I said below, I think the best solution is sufficiently isolating TF

gezp:
do you mean `global_loc_srv_`  callback maybe block  `laser_scan_sub_` ?  if so, how about moving the `laser_scan_sub_` onto another executor ?

SteveMacenski:
It doesn't have to do with that, it has to do with the services blocking **tf** and the message filter.

gezp:
sorry, the message filter of tf is difficult to understand for me. in my understanding, when receive a laser_scan message:
*  if it's transformable , the laser_scan message will be handled Immediately (call `laserReceived`).
*  otherwise, the message will add to a map (`transformable_callbacks_`), then in the callback of tf listener, use `testTransformableRequests()` to check and process this map ( if item of map is transformable , call `laserReceived` for this item), in addition,  timers will be used to remove item  which is timeout from the map.

and `global_loc_srv_` callback is only  called when initialize global localization, which is not constant.  if it blocks laser_scan callback, just drop out  laser_scan message, it's not a big issue, because laser_scan subscriber is constant. i don't understand why need a separate executor for `global_loc_srv_`.

gezp:
sorry, the message filter of tf is difficult to understand for me. in my understanding, when receive a laser_scan message:

* if it's transformable , the laser_scan message will be handled Immediately (call `laserReceived`).
* otherwise, the message will add to a map (`transformable_callbacks_`), then in the callback of tf listener, use `testTransformableRequests()` to check and process this map ( if item of map is transformable , call `laserReceived` for this item), in addition, timers will be used to remove item which is timeout from the map.

if  you want to isolate TF timer , i have a idea to do it :
* let all services and subscriptions use a new callback group with dedicated executor and thread.
* let timer use default callback group. 

in my opinion, we needn't  isolate TF timer, because timer is only used to cancel TransformableRequests (maybe wrong), it isn't a issue to cancel with delay.

SteveMacenski:
- https://github.com/ros2/geometry2/blob/ros2/tf2_ros/src/buffer.cpp#L193 uses the timer we give it, which would be blocked by a long running service call or topic callback
- https://github.com/ros2/geometry2/blob/ros2/tf2_ros/include/tf2_ros/message_filter.h#L426 uses `waitForTransform` within the message filter in AMCL (and costmap)
- `add` is called every time a new message is thrown into the filter https://github.com/ros2/geometry2/blob/ros2/tf2_ros/include/tf2_ros/message_filter.h#L576

Thusly, a blocked executor from a service could easily mess up the message filter's ability to function, throwing exceptions from timeouts and returning lots of errors/warnings to the logs.

gezp:
I have updated PR to isolate TF timer : 

* let all services and subscriptions use  a new callback group with dedicated executor and thread.
* let timer use default callback group.

please review again. @SteveMacenski

SteveMacenski:
What you suggested is not unreasonable, but I'd like to focus on https://github.com/ros2/geometry2/issues/446 to actually fix the underlying issue. I'd be curious if you could find a clean way to fix this in TF itself (maybe take in a node and have the timer interfaces in its own separate callback group / executor background thread) like we did with the other TF PR.

That would have a significantly larger impact on the community longer term preventing issues like this from reappearing elsewhere

SteveMacenski (approved):

SteveMacenski:
@gezp where are we on this? Can you ping the ticket / PR that is blocking this work from being completed? I really don't want this to continue sitting around when it could be resolved

gezp:
@SteveMacenski , it's blocked by https://github.com/ros2/geometry2/pull/447.

SteveMacenski:
OK - I poked them and it should be merged within a day

SteveMacenski:
Merged!

gezp:
CI failed,  we need wait for a new rolling sync to get feature https://github.com/ros2/geometry2/pull/447.

gezp:
CI passed, please review this pr again. @SteveMacenski

SteveMacenski (approved):

SteveMacenski:
Thanks! Nice to get this in finally :smile: I think there's just 1 more pending to complete this all. Thanks for being so on top of it once we got back on Rolling!

PR #2484, opened 2021-07-28
Update regulated_pure_pursuit_controller.cpp

SteveMacenski:
https://answers.ros.org/question/380584/regulatedpurepursuitcontroller-detected-a-collision-ahead-but-robot-doesnt-stop/?comment=383417#post-id-38341

Backport to all distributions.

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -43,7 +43,7 @@ void RegulatedPurePursuitController::configure(
 {
   auto node = parent.lock();
   if (!node) {
-    throw std::runtime_error("Unable to lock node!");
+    throw nav2_core::PlannerException("Unable to lock node!");
   }
 
   costmap_ros_ = costmap_ros;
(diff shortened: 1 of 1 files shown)


Kaetera:
Hello (I'm the author of the original ROS Q&A question).

As we are on Foxy I replaced manually the `std::runtime_error` by `nav2_core::PlannerException`.
The exception is indeed caught (verified by putting a RCLCPP_INFO() in the catch block of nav2_controller.cpp::ControllerServer::computeControl(), but we already had the error print)

But this does not solves the initial issue. My test case still fails,. 
The test cases shown below also fails: 
- Robot is idle, not moving, no velocity commands
- Place an obstacle close to the robot, visible in the costmap, so close that it will hit it in the few seconds of motion
- Put a goal behind the obstacle,
- The robot will start to move and hit the obstacle, this time even without detecting "collision ahead".

This scenario is shown with the two screenshot below, with two inflation radius sets in the local costmaps (one 0.4m, other 0.6m greater than the greater dimension of the robot). 

Non-null velocity commands are published nor "collision ahead" error message despite the collision is obvious (bottom terminal shows `ros2 topic echo /cmd_vel`):

RPP patched test 1
RPP patched test 2

Kaetera:
Ok my bad.

I just realize that the collision is detected when the extremity of the collision arc hit a region of 'high' cost (cyan area, ie. the obstacle inflated by the minimal dimension of the robot).

So my cases are a problem of tuning max_allowed_time_to_collision w.r.t to the size of the robot.

image

SteveMacenski:
> So my cases are a problem of tuning max_allowed_time_to_collision w.r.t to the size of the robot.

Exactly, looks like you found it on your own :smile:

Issue #2482, opened 2021-07-27
[Smac 2D planner] Path end point has always a 0 orientation


doisyg:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 
- ROS2 Version:
  - Galactic bin
- Version or commit hash: last bin for ros2, src sync with main for nav2
- DDS implementation:
  - Cyclone

#### Steps to reproduce issue

Ask the planner using the Smac 2D plugin  to generate a path to a goal with a specific orientation
#### Expected behavior
Then end point of the path should have the same orientation as the goal sent
#### Actual behavior
Then end point of the path has a 0 orientation

SteveMacenski:
This is a 2D A*, that's not possible the same way its not for any other 2D A* including NavFn

doisyg:
> > Ask the planner using the Smac 2D plugin to generate a path to a goal with a specific orientation
> 
> This is a 2D A*, that's not possible the same way its not for any other 2D A* including NavFn or any other planner for that matter. Its very typical as you know to rotate to heading once within tolerance of the goal with a TEB/DWB/RPP/etc. That is common behavior. To get an approach angle, you need a kinematically feasible planner, like Hybrid-A*...
> 
> The `0` is the same `0` that all others provide. I think what you're pointing out is that we actually do populate orientations for the other path points in the plan based on the tangent of the path. We should populate the final point as well, which I'd merge a PR to do [(its in the smoother)](https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/include/nav2_smac_planner/smoother.hpp#L236). But note that NavFn and anything you used before also provided `0`s as well for the path.
> 
> It should be easy to just add to the end goal point the final orientation

NavFn and everything I used before populates the end point with the goal orientation (or something close), that's the only way the local planner can have access to the goal orientation.
Here is what I observe:

> With SmacPlanner2D
> [planner_server]: Found valid path of size 23 to (11.05, 3.50, 1.91)
> [planner_server]: Path end pose: (11.03, 3.48, 0.00)
> 
> With SmacPlannerHybrid
> [planner_server]: Found valid path of size 49 to (11.11, 3.63, 3.13)
> [planner_server]: Path end pose: (11.13, 3.63, 3.05)
> 
> With NavfnPlanner
> [planner_server]: Found valid path of size 35 to (10.80, 3.94, -1.94)
> [planner_server]: Path end pose: (10.80, 3.94, -1.94)

[...]

Issue #2478, opened 2021-07-27
[Smac planner] Ignore goal orientation option


doisyg:
#### Feature description
I would like to ask the planner to generate a path without taking into account the orientation of the goal pose. In other word, to arrive to the goal position as naturaly as possible and without final rotation or added constraint on the trajectory.
Is it already possible and I missed something ?
If not, what would be the place to add this feature ? A hacky/nasty way with the 2D planner is to replace the final orientation by the orientation of the point of the path before.

SteveMacenski:
Which Smac Planner do you mean? 

The 2D planner is just a 2D grid search A*, it doesn't know anything about orientations, so I think that is already true. You can verify yourself that we don't even input the orientation information to the A* algorithm in the 2D case. 

The Hybrid-A* planner's selling point is kinematic feasibility, so if that were to be offered in Hybrid-A*, it would have to be a unique parameter or only leveraged by the analytic expansion. I'm not sure realistically how to get that done without actually costing more CPU than just specifying an orientation

doisyg:
> Which Smac Planner do you mean?
Ideally both but I am currently focused on the 2D planner

> The 2D planner is just a 2D grid search A*, it doesn't know anything about orientations, so I think that is already true. You can verify yourself that we don't even input the orientation information to the A* algorithm in the 2D case.

I just realize that more or less: a path computed with the 2D planner outputs a end pose with a 0 orientation, no matter which orientation was sent in the goal! This is, I believe a major issue to use it as such as a standard planner! Opening a separated issue: https://github.com/ros-planning/navigation2/issues/2482

SteveMacenski:
O..k... closing this then?

gomandr:
Hi,

I'm confronted with the case where I don't have any requirement on the goal orientation and specifying one beforehand should be avoided (large turning radius w.r.t cell size). While I don't see how to have a "pure" orientation free goal implemented easily in Hybrid-A*, I tried the following modifications:

- Add an `ignore_goal_orientation` parameter in `motion_table`,
- In `NodeHybrid::getDistanceHeuristic`, recompute the goal orientation as being the "bearing" from the current node, e.g something along the line of `to[2] = std::atan2(to[1] - from[1], to[0] - from[0]);`,
- In the `AStarAlgorithm<NodeT>::getAnalyticPath`, same as above.

It's working as expected, but maybe I'm missing something important here. Do you see any pitfalls in doing that ? 

If you think it's a viable solution, I could open a PR with this.

PR #2488, opened 2021-08-02
[All 2D planners] Fix last pose orientation, fix small path and add ignore_goal_orientation parameter

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2482, https://github.com/ros-planning/navigation2/issues/2478 and https://github.com/ros-planning/navigation2/issues/2517|

---

1. Fix goal orientation being ignored and always set to 0 (https://github.com/ros-planning/navigation2/issues/2482) by copying the goal orientation to the last pose of the path
2. Thanks to 2., new `ignore_goal_orientation` parameter (defaults to **false**) to enable the behavior discussed here https://github.com/ros-planning/navigation2/issues/2478, i.e. no orientation discontinuities between path last pose and path (last - 1) pose
3. Fix small path corner cases

* Description of the fixed default behavior in the migration guide ?

---

#### For Maintainers: 


nav2_smac_planner/include/nav2_smac_planner/smoother.hpp
```diff
@@ -243,6 +243,8 @@ class Smoother
       theta = atan2(dy, dx);
       path.poses[i].pose.orientation = orientationAroundZAxis(theta);
     }
+    // Populate last point
+    path.poses[path.poses.size() - 1].pose.orientation = orientationAroundZAxis(theta);
```
SteveMacenski, line 247: how about path.poses.back().pose.orientation?

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -80,6 +80,9 @@ void SmacPlanner2D::configure(
   nav2_util::declare_parameter_if_not_declared(
     node, name + ".max_on_approach_iterations", rclcpp::ParameterValue(1000));
   node->get_parameter(name + ".max_on_approach_iterations", _max_on_approach_iterations);
+  nav2_util::declare_parameter_if_not_declared(
+    node, name + ".ignore_goal_orientation", rclcpp::ParameterValue(false));
+  node->get_parameter(name + ".ignore_goal_orientation", _ignore_goal_orientation);
```
SteveMacenski, line 85: No need for a parameter here, this should always just be done. There's no situation I can think of where this isn't the right thing to do.

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -292,6 +295,10 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
     _smoother->smooth(plan, costmap, time_remaining);
   }
 
+  if (!_ignore_goal_orientation && plan.poses.size() > 0) {
+    plan.poses[plan.poses.size() - 1].pose.orientation = goal.pose.orientation;
```
SteveMacenski, line 299: This can go in the Hybrid as well!

SteveMacenski (changes requested):

nav2_smac_planner/src/smac_planner_2d.cpp
SteveMacenski: and `back()`

nav2_smac_planner/include/nav2_smac_planner/smoother.hpp
doisyg: cleaner indeed

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: I did not did into the hybrid version yet but it according to my experiment it is not needed for it to respect the goal orientation: https://github.com/ros-planning/navigation2/issues/2482#issuecomment-888133212

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: With this parameter set to false (default), the goal pose is respected, which is the classical behavior of NavFn and other planner, i.e last path pose orientation = goal pose orientation
When it is set to true, the last pose has the same orientation as the n - 1 pose, which is a behavior that I sometimes need (i.e. send the robot to a `position` without orientation constraints, as opposite to send the robot to a `pose` with orientation constraints. And apparently I am not the only one, see https://github.com/ros-planning/navigation2/issues/2478#issuecomment-891834153, though it is for the hybrid version.

nav2_smac_planner/src/smac_planner_2d.cpp
SteveMacenski: Got it, makes sense

nav2_smac_planner/src/smac_planner_2d.cpp
SteveMacenski: Weird, I'm not regularly getting notifications about this ticket, sorry about that. 

For Hybrid-A*, there will not (and impossible to be) a version of that planner without orientations, it is by nature kinematically feasible which means you need a full kinematic state to hit. That person is misinformed about path planning. 

> With this parameter set to false (default), the goal pose is respected, which is the classical behavior of NavFn and other planner, i.e last path pose orientation = goal pose orientation

I think that's the right behavior for 2D always. If you don't want orientation constraints, why are you not just setting your goal checker plugin to have a tolerance of 2*PI? Any orientation is valid then. We should not have the path planner just "ignore" valid state information. I know for the 2D case its pretty silly because there's no orientations in the algorithm, but if you request `10 deg`, even if you don't really mean it, the planner should just take that at face value to return a goal at `10 deg`. 

I think the place to ignore that information is the controller server. The goal checker can be set to 2*PI to make it just care about the position and not at all about the orientation. That would result in the same behavior and much more robust to changes / path tracking errors (e.g. if you set to n-1 but then approached at a slightly different angle using a modeling based local planner, then it would try to rotate to n-1's heading)

doisyg: > If you don't want orientation constraints, why are you not just setting your goal checker plugin to have a tolerance of 2*PI?

Well, because I use a local planner/controller which looks way ahead of its current pose and is allowed to go away from the global path in order to time optimize its trajectory.  Hence even with a angle tolerance of 2*PI, the final pose orientation will have an effect on the trajectory computed by my local planner/controller. But better explained with gifs:

1. Goal orientation pointed left of screen and hence equal to the final global path pose orientation, 2*PI angle tolerance, 0.02m distance tolerance
without_ignore_orientation

2. Goal orientation pointed left of screen BUT ignore_goal_orientation set to true so final global path pose orientation equal n-1 pose orientation , 2*PI angle tolerance, 0.02m distance tolerance
with_ignore_orientation

> We should not have the path planner just "ignore" valid state information

I agree that this should definitely not be the default behavior but there is the use case of wanting to go to a `Position` instead of a `Pose`

> I think the place to ignore that information is the controller server. The goal checker can be set to 2*PI to make it just care about the position and not at all about the orientation. That would result in the same behavior and much more robust to changes / path tracking errors (e.g. if you set to n-1 but then approached at a slightly different angle using a modeling based local planner, then it would try to rotate to n-1's heading). So the real solution is in the goal checker plugin, which if you do, makes this moot.

I am not sure what is the right place for it, but as shown above, the goal checker is not the way to solve this.
I believe there is an use case for it (or at least I have it) and to me it will nicely fit in the 2D planner as an optional behavior (as it is an orientation ignorant algorithm anyway). However if you feel that it is better to leave it out of the planner, the cost for me to put that end pose orientation modification somewhere else, like in my custom planner server, is small, and I will remove it from this PR.

SteveMacenski: That's a fair situation. Mhm. I do wonder what is best here. Thematically, it seems wrong for us to just "ignore" the goal requested orientation, and if we did, it would seems even a weirder behavior to default it to the approach angle rather than just being `0.0` as the default value of a `double`. How does the controller handle just a default `0.0`?

But at the same time, I can appreciate the situation. The other half of my brain then says "ok, then let's solve this for the general case" which would then be somewhere in the planner server so that it could apply to any planner and not specifically just the Smac planner (and only the 2D version of that, no less). 

Neither of these situations are the options you're describing as desirable though. I agree for your predictive controller you'd need to manipulate the planner's approach angle as well as the goal checker for task termination. 

So ideally, I'd want it in the planner server - which doesn't sound like it would help you much if you have your own planner server. I'd also want to mess with the naming to be more clear to what it is doing since it's not really "ignoring" it as much as "setting" it to the approach angle. 

---

Just curiosity, is that TEB or something else? Why the separate planner server?

doisyg: > That's a fair situation. Mhm. I do wonder what is best here. Thematically, it seems wrong for us to just "ignore" the goal requested orientation, and if we did, it would seems even a weirder behavior to default it to the approach angle rather than just being `0.0` as the default value of a `double`. How does the controller handle just a default `0.0`?

The controller treats it as a 0 deg orientation request

> But at the same time, I can appreciate the situation. The other half of my brain then says "ok, then let's solve this for the general case" which would then be somewhere in the planner server so that it could apply to any planner and not specifically just the Smac planner (and only the 2D version of that, no less).

The thing is, for "smart" planners, like the hybrid, where the goal pose orientation has an influence on many poses before the last, the trick of copying the n-1 orientation is not working. Putting such a parameter in the planner server would imply that all planners can make use of it, which for some planner is complex (or maybe impossible). Or do you see a way of having it in the planner server that will not confuse users in case the planner they use is not implementing this feature ? 
I like the idea of putting it in the planner itself because it is immediately clear if this feature is supported: the parameter is present, it is, the parameter is absent, it is not. But I don't have a strong opinion on this.


> Neither of these situations are the options you're describing as desirable though. I agree for your predictive controller you'd need to manipulate the planner's approach angle as well as the goal checker for task termination.
> 
> So ideally, I'd want it in the planner server - which doesn't sound like it would help you much if you have your own planner server. I'd also want to mess with the naming to be more clear to what it is doing since it's not really "ignoring" it as much as "setting" it to the approach angle. This topic will probably come up again, so I'm not entirely sure what the right answer is, maybe add that parameter to this and NavFn / Theta* so all the 2D planners have this option? That seems less.. great than having the planner server take it on, but what if you have multiple planners and only want this behavior with 1? There are other parameters like `unknown` and `tolerance` in each multiple planners implement, so maybe this could just be another?

Now that you mention it, yes, that could be another parameter like `unknown` and `tolerance` and that's an argument for putting it in the planners.
About the name we can brainstorm. But I would like it to convey the idea that we tell the planner to not use the last goal orientation as a constraint to compute its path, which make it more general than copying the approach angle. compute_without_goal_orientation_constraint ? 2d_only_goal ? xy_only_goal ? 

> I know that's not a very clear "do this" answer. The only thing I can clearly state is changing the name. Otherwise, if it goes into Smac2D as a param, it should also go in NavFn / Theta* to match. If it goes into planner server, then no need. What do you think?

I prefer it in the planner, and I can add it the NavFn, but I will let you take the decision.

> Just curiosity, is that TEB or something else? Why the separate planner server?

Yes, TEB based. Not entirely separate planner server, it still inherits from nav2_planner::PlannerServer and we have a couple of overload mainly for dealing with computing path transparently when the robot is docked, maintaining multiple costmaps in parallel with different options and custom feedback/result messages

SteveMacenski: `use_final_approach_orientation` would be my naming suggestion.

BTW any parameter changes needs a complementary update to navigation.ros.org for any packages changed. Also adding to the migration guide this new parameter and what it does (those 2 gifs would be helpful illustrations).

Lets do this, lets add this parameter to all of the 2D planners. If true, does the approach angle. If false, sets it to the requested goal orientation. So that's Navfn, Smac2D, and Theta*.

SteveMacenski: A small test in Smac2D and Theta* at minimum would be nice to make sure this is working when set

doisyg: All right, I will try to attend asap

doisyg: Ok, added to NavFn and Theta*. Theta* had also the same end pose orientation always set to 0 issue as Smac2D, fixed.
Tests where added to test_planner_pluggins.cpp and I changed the name from `test_planner_plugin_failures` to `test_planner_plugins`. It was easy to do with the **nav2_system_tests::NavFnPlannerTester** class, which btw can be used to test other planner than NavFn, so maybe we should change its name ?
Let me know if you want another iteration, and if that's good, I will prepare the documentation PR

nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -85,6 +85,9 @@ NavfnPlanner::configure(
   node->get_parameter(name + ".use_astar", use_astar_);
```
SteveMacenski, line 85: Need PRs to add params to the configuration guide / migration guide

nav2_smac_planner/include/nav2_smac_planner/smoother.hpp
```diff
@@ -243,6 +243,8 @@ class Smoother
       theta = atan2(dy, dx);
       path.poses[i].pose.orientation = orientationAroundZAxis(theta);
     }
+    // Populate last point
+    path.poses.back().pose.orientation = orientationAroundZAxis(theta);
```
SteveMacenski, line 247: For the Hybrid-A* planner though, I don't think this would be overridden for the final goal orientation this way. I think that needs to be done in the Hybrid-A* code if you want to add this for the benefit of the 2D code.

nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -284,6 +288,24 @@ NavfnPlanner::makePlan(
     // extract the plan
     if (getPlanFromPotential(best_pose, plan)) {
       smoothApproachToGoal(best_pose, plan);
+
+      size_t plan_size = plan.poses.size();
+      if (use_final_approach_orientation_ && plan_size > 1) {
```
SteveMacenski, line 293: This would have to be > 2 in order to do the approach_pose.

What would you do if you had a path that came back with only 1 or 2 poses? It would then return the actual angle instead making your TEB planner backout to hit that orientation goal.

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -292,6 +295,10 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
     _smoother->smooth(plan, costmap, time_remaining);
   }
 
+  if (!_use_final_approach_orientation && plan.poses.size() > 0) {
+    plan.poses[plan.poses.size() - 1].pose.orientation = goal.pose.orientation;
```
SteveMacenski, line 299: `back()`

nav2_smac_planner/test/test_smac_2d.cpp
```diff
@@ -70,6 +71,20 @@ TEST(SmacTest, test_smac_2d)
   } catch (...) {
   }
 
+
+  goal.pose.position.x = 0.0;
```
SteveMacenski, line 75: Where's the test in this? Also lots of white space that will cause the linter to be unhappy

SteveMacenski (changes requested):

nav2_navfn_planner/src/navfn_planner.cpp
SteveMacenski: That will happen, and happen often, with any reasonably large scaled deployment.

nav2_smac_planner/test/test_smac_2d.cpp
doisyg: merge artefact, tests are in `nav2_system_tests` `test_planner_plugins`

nav2_navfn_planner/src/navfn_planner.cpp
doisyg: This is working for a path with only 2 poses (plan_size = 2)
However I am not sure to see what should happen for a path with one pose only. We probably don't want the robot to move at all if `use_final_approach_orientation=true` (i.e not rotate if it doesn't have to change its x,y position), hence we need our only pose to have the same orientation as the current robot orientation. Hence the start pose orientation. Does it make sense ?

SteveMacenski: yes,  the starting pose orientation makes sense, we should cover that edge case

SteveMacenski:
@doisyg any update?

doisyg:
I'll try to wrap up this week, but turns out the corner case with start to goal distance very small (or 0) should be taken care of properly too. Currently the 2d planner fails in that case. I will make additional tests for this case and check the behavior of the other 2d planners.
I propose the following:
- in case start(x,y) = goal(x,y), return a path with a single pose equal to the start/goal pose (and if the orientation is not the same, use start or goal orientation depending on use_final_approach_orientation)
- in case start to goal distance is small (but we have to define what is small, I believe it depends on the resolution of the planner), return a path with two poses: start for the first and goal for the second. Hence we let the local planner decide if it big enough to initiate a move according to its tolerance. As for the orientation, again, start or goal orientation for the goal orientation (I hesitate with filling it to the start to goal vector orientation but it doesn't really make sense if the distance is small)

SteveMacenski:
I think https://github.com/ros-planning/navigation2/issues/2517 asks that same request as well that I approved.

doisyg:
> I think #2517 asks that same request as well that I approved that same concept. Just needs to be applied to all of the planners to make that behavior consistent.

Indeed

SteveMacenski:
I'm sure he wouldn't mind if you did it instead if you need it

nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -284,6 +315,33 @@ NavfnPlanner::makePlan(
...
     if (getPlanFromPotential(best_pose, plan)) {
       smoothApproachToGoal(best_pose, plan);
+
+      // Override last pose orientation to match goal if use_final_approach_orientation=false
+      // (default).
+      // If use_final_approach_orientation=true, interpolate the last pose orientation from the
+      // previous pose.
+      // And deal with corner case of plan of length 1
+      size_t plan_size = plan.poses.size();
+      if (plan_size == 1) {
```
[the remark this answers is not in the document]
doisyg: hum, then we need to nest everything in a `if (plan_size > 0)` (because we don't want to interfere with the planner returning a null path):

>       if (plan_size > 0){
>         if (use_final_approach_orientation_) {
>           
>         } else { 
>           
>         }
>       }

or to avoid adding a block, but less clearer :

>       if (use_final_approach_orientation_ && plan_size > 0) {
>         
>       } else if (!use_final_approach_orientation_ && plan_size > 0){ 
>         
>       }

What do you prefer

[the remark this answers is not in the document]
SteveMacenski: That's fair, I wasn't super explicit

```
if (use_final_approach_orientation_) {
  size_t plan_size = plan.poses.size();
  if (plan_size == 1) {
    plan.poses.back().pose.orientation = start.orientation;
  } else if (plan_size > 1) {
    double dx, dy, theta;
    auto last_pose = plan.poses.back().pose.position;
    auto approach_pose = plan.poses[plan_size - 2].pose.position;
    // Deal with the case of NavFn producing a path with two equal last poses
    if (std::abs(last_pose.x - approach_pose.x) < 0.0001 &&
      std::abs(last_pose.y - approach_pose.y) < 0.0001 && plan_size > 2)
    {
      approach_pose = plan.poses[plan_size - 3].pose.position;
    }
    dx = last_pose.x - approach_pose.x;
    dy = last_pose.y - approach_pose.y;
    theta = atan2(dy, dx);
    plan.poses.back().pose.orientation =
      nav2_util::geometry_utils::orientationAroundZAxis(theta);
  }
```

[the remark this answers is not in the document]
SteveMacenski: I think that's alot more easily readable, no?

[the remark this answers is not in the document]
doisyg: Definitly, thx!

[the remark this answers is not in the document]
doisyg: I remove the override of the goal orientation in NavFn and Theta*, only needed for Smac2d and modify the comments to reflect it

SteveMacenski (approved):

Issue #2494, opened 2021-08-08
 [SmacPlanner2D] Planner fails when goal is close to start


doisyg:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu
- ROS2 Version:
  - galactic
- Version or commit hash:
  - ros last bin and nav2 src sync
- DDS implementation:
  - Cyclone

#### Steps to reproduce issue
Ask the planner to compute a path with a goal which has the same position as the start (independently from the orientation).

#### Expected behavior
The planner should succeed and output a valid path, even if it contains only one pose.

#### Actual behavior
The planner always fail when the the start x and y are close to the goal x and y

doisyg:
Proposed fix: https://github.com/ros-planning/navigation2/pull/2495

PR #2495, opened 2021-08-09
[Smac2d] Accept path of length 1

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2494 |

---

Solves https://github.com/ros-planning/navigation2/issues/2494 . 
Fully tested with smac2d planner. However as a_star is used for the hybrid version, it may have implications that I don't foresee (I did a basic sanity check with the hybrid though)

---

#### For Maintainers: 


SteveMacenski (approved):
LGTM

Issue #2515, opened 2021-08-17
smac_planner bad_alloc on short paths


Aposhian:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - rolling
- Version or commit hash:
  - 1.0.6
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
Add the following unit test to https://github.com/ros-planning/navigation2/blob/1261768dd97253ba7a4bf58b7022be2dbeec7e38/nav2_smac_planner/test/test_a_star.cpp
```
TEST(AStarTest, test_se2_single_pose_path)
{
  nav2_smac_planner::SearchInfo info;
  info.change_penalty = 0.1;
  info.non_straight_penalty = 1.1;
  info.reverse_penalty = 2.0;
  info.minimum_turning_radius = 8;  // in grid coordinates
  unsigned int size_theta = 72;
  info.cost_penalty = 1.7;
  nav2_smac_planner::AStarAlgorithm<nav2_smac_planner::NodeHybrid> a_star(
    nav2_smac_planner::MotionModel::DUBIN, info);
  int max_iterations = 100;
  float tolerance = 10.0;
  int it_on_approach = 10;
  int num_it = 0;

  a_star.initialize(false, max_iterations, it_on_approach, 401, size_theta);

  nav2_costmap_2d::Costmap2D * costmapA =
    new nav2_costmap_2d::Costmap2D(100, 100, 0.1, 0.0, 0.0, 0);

  std::unique_ptr<nav2_smac_planner::GridCollisionChecker> checker =
    std::make_unique<nav2_smac_planner::GridCollisionChecker>(costmapA, size_theta);
  checker->setFootprint(nav2_costmap_2d::Footprint(), true, 0.0);

  // functional case testing
  a_star.setCollisionChecker(checker.get());
  a_star.setStart(10u, 10u, 0u);
  a_star.setGoal(11u, 10u, 0u);
  nav2_smac_planner::NodeHybrid::CoordinateVector path;
  EXPECT_TRUE(a_star.createPath(path, num_it, tolerance));

  // Check that the path is length one
  EXPECT_EQ(path.size(), 1u);

  delete costmapA;
}
```

#### Expected behavior
Planner plans a path of with a single pose in it.

#### Actual behavior
```
[ RUN      ] AStarTest.test_se2_single_pose_path
unknown file: Failure                         
[...]

PR #2516, opened 2021-08-17
Smac planner bad alloc

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2515 |

---

Fixed bug that allowed a `std::bad_alloc` to be thrown by providing a `-1`, cast to `INT_MAX`, to `std::vector::reserve`

None

---

Single pose paths and zero length paths are still nonfunctional, but that is for another ticket.

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Waiting on the AOK from CI

Issue #2333, opened 2021-05-10
Obstacle_layer plugin cause high CPU usage in controller_Server and planner_server(sometime over 120% ) 


wilsionhuang:
## Bug report 
 Obstacle_layer plugin cause high CPU usage in controller_Server and planner_server

**Required Info:**

- Operating System
  -  Ubuntu 20.04
- ROS2 Version
  - Foxy Binaries
- Version or commit hash
  -  foxy version and build From source
- DDS implementation
  -   Fast-RTPS

#### Steps to reproduce issue

```
1. launch the laser node  and navigation2 software stack
2. keep the robot standstill or do task of moving point A to piont B
3. open htop in new terminal and you can find the cpu of controller_server and planner_server is unstable and high(sometimes the crest value may over 120%)
```

#### Expected behavior
The cpu usage of controller_server and planner_server  is stable and not high 
#### Actual behavior
The cpu usage of controller_server  and planner_server is unstable and increase abruptly over 120%
![Screenshot from 2021-05-11 15-39-34](https://user-images.githubusercontent.com/26588883/117783995-a7bc8600-b275-11eb-877b-d3f40a21e832.png)

#### Temporary solution
1. Transform the scan data(laser_node) to pointcloud and then transform the pointcloud data from the laser_link to map before publish
2. Transform scan data from laser_link to map directly

#### The result of temporary solution
the cpu usage of controller_server and planner_server  is decreased and keeping stable obviously(not increase abruptly)
![Screenshot from 2021-05-11 16-32-07](https://user-images.githubusercontent.com/26588883/117784774-7db79380-b276-11eb-890c-5b50f943d9e3.png)

#### Three points may cause this problem and Debug
1. projector_.transformLaserScanToPointCloud(message->header.frame_id, *message, cloud, *tf_) cause the most cpu in callback function -- through transform the scan to pointcloud data  and not transform data from laser_link to map before publish, the bug is still alive
2.  buffer->bufferCloud(cloud) -- ongoing debug 
3. QOS  buffer size is lager -- decrease the buffer size, the problem is still remaining

#### Co-woker
@tal-grossman @Andersdu @duanyunxin

Although the temporary solution can fix this problem, we still could not find the point which cause this problem in obstacle_layer plugins. If anyone have ideas with this issue, we are hopefully Looking forward to your reply. 
Thanks.

SteveMacenski:
[...]

PR #2518, opened 2021-08-17
[ObstacleLayer] Use message_filter timeout

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2333 |
| Robotic platform tested on | Elodie Wyca|

---

Use a timeout on the obstacle_layer message_filter. And use `transform_tolerance` to set its value since buffered message from the message_filter older than `transform_tolerance` would have been discarded later.
This effectively solves the controller_server being stuck with a very high CPU usage after few tf extrapolation errors

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -252,7 +252,7 @@ void ObstacleLayer::onInitialize()
 
       std::shared_ptr<tf2_ros::MessageFilter<sensor_msgs::msg::PointCloud2>> filter(
         new tf2_ros::MessageFilter<sensor_msgs::msg::PointCloud2>(
-          *sub, *tf_, global_frame_, 50, rclcpp_node_));
+          *sub, *tf_, global_frame_, 50, rclcpp_node_, tf2::durationFromSec(transform_tolerance)));
 
       filter->registerCallback(
         std::bind(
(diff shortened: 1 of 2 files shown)


nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -218,7 +218,7 @@ void ObstacleLayer::onInitialize()
 
       std::shared_ptr<tf2_ros::MessageFilter<sensor_msgs::msg::LaserScan>> filter(
         new tf2_ros::MessageFilter<sensor_msgs::msg::LaserScan>(
-          *sub, *tf_, global_frame_, 50, rclcpp_node_));
+          *sub, *tf_, global_frame_, 50, rclcpp_node_, tf2::durationFromSec(transform_tolerance)));
```
SteveMacenski, line 221: You mind adding this to the other message filters? There's another in this file and 1 in AMCL. There's also one in SLAM Toolbox / STVL each if you're still users of those

SteveMacenski (changes requested):

nav2_costmap_2d/plugins/obstacle_layer.cpp
SteveMacenski: Edit: You were ahead of me :laughing:

doisyg: already done for STVL (as yes, I am still using it), I am checking the others

SteveMacenski: there's 2 of these in obstacle layer; one for PC2 and another for laser scans

doisyg: Can't believe I missed the PC2 one, done now. Troubleshooting this problem really got the best of me.
However, I don't use SLAM toolbox and I am not familiar with its code, so I will let you fix it there if that's fine.

SteveMacenski (approved):

SteveMacenski:
Waiting on CI

PR #2526, opened 2021-08-20
Hot fix rrp slow

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2332  |
| Robotic platform tested on | mp-400 from Neobotix |

---

* Fixed the segfault issue that occurred when the robot was set to drive at high speed using RPP. 

NA

---

NA

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -441,7 +441,9 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double 
   if (!costmap_->worldToMap(x, y, mx, my)) {
     RCLCPP_WARN_THROTTLE(
       logger_, *(clock_), 30000,
-      "The dimensions of the costmap is smaller, proceed at your own risk");
+      "The dimensions of the costmap is too small to successfully check for"
+      "collisions as far ahead as requested. Proceed at your own risk, slow the robot, or"
+      "increase your costmap size.");
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -450,7 +459,16 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
 double RegulatedPurePursuitController::costAtPose(const double & x, const double & y)
 {
   unsigned int mx, my;
-  costmap_->worldToMap(x, y, mx, my);
+
+  if (!costmap_->worldToMap(x, y, mx, my)) {
+    RCLCPP_FATAL(
+      logger_,
+      "The dimensions of the costmap is too small to fully include your robot's footprint"
```
SteveMacenski, line 466: Needs a space and a comma, the line extension doesn't add a space between the words

SteveMacenski: as well below

SteveMacenski: and above

SteveMacenski (changes requested):

SteveMacenski:
@padhupradheep last thing here is just string formatting, then good to go!

padhupradheep:
> @padhupradheep last thing here is just string formatting, then good to go!

Done @SteveMacenski

SteveMacenski (approved):

SteveMacenski:
Thanks!!

PR #2534, opened 2021-08-25
Change default timeout of BT navigator

Frederic98:
Changed the default timeout of BT navigator. The previous timeout of 10ms caused ROS to crash if there was some latency in the network - for example when setting a new goal, as seen in https://github.com/ros2/ros2/issues/1074.
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (https://github.com/ros2/ros2/issues/1074) |
| Primary OS tested on | (Ubuntu 20.04 Foxy) |
| Robotic platform tested on | (Custom robot) |

---

Change default timeout of BT navigator
Timeout was 10ms. This was causing a crash if there was a tiny bit of latency in the network, or the CPU was overloaded by other tasks. Changed to 500ms to prevent crashing the whole robot.

Might want to update the documentation to describe the parameter, and it's effect on the navigator.

---

None, as far as I can see

#### For Maintainers: 


SteveMacenski:
Please change your own configuration files. This PR will not be accepted, this is a parameter for you to be able to tune to your own system’s needs.

Issue #2453, opened 2021-07-16
[Recovery] Backup never terminating when robot is stuck


karl-schulz:
## Bug report

**Required Info:**

- Operating System:
  - Any
- ROS2 Version:
  - Galactic
- Version or commit hash:
  - master or 5a70275 
- DDS implementation:
  - Any

#### Steps to reproduce issue

Run any robot simulation using a backup recovery. Example behavior tree XML:

````
<root main_tree_to_execute="MainTree">
  <BehaviorTree ID="MainTree">
    <RecoveryNode number_of_retries="3" name="NavigateRecovery">
      <PipelineSequence name="SimpleBT">
        <RateController hz="4.0">
          <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased"/>
        </RateController>
        <FollowPath path="{path}" controller_id="FollowPath"/>
      </PipelineSequence>
      <ReactiveFallback name="RecoveryFallback">
        <GoalUpdated/> 
        <RoundRobin name="RecoveryActions">
          <BackUp backup_dist="0.65" backup_speed="0.15"/>
        </RoundRobin>
      </ReactiveFallback>
    </RecoveryNode>
  </BehaviorTree>
</root>
````

Now, block the robot from behind to it cannot move backwards. The backup movement will be stuck at:

    [controller_server-3] [ERROR] [1626426808.548096680] [controller_server]: Failed to make progress
    [controller_server-3] [WARN] [1626426808.548129815] [controller_server_rclcpp_node]: [follow_path] [ActionServer] Aborting handle.
    [recoveries_server-4] [INFO] [1626426808.570189476] [recoveries_server]: Attempting backup

And navigation never terminates and keeps publishing`cmd_vel` = `... linear: x: -0.15  ...`.

In the real world, this can be triggered by multiple things:
* Failure in localization (and the robot will move backwards forever, not good)
* Being blocked by a non-visible object
* Motor failure, robot being broken and not moving despite non-negative `cmd_vel`

#### Expected behavior

The recovery fails and the navigator returns an error.

#### Actual behavior

The navigator is stuck in the backup node on RUNNING.

#### Proposed Solution

Provide an (optional) argument `timeout` and check if it expired in the `BackUp::onCycleUpdate`. If the timeout expired, return FAILURE.

If you want, I could create a PR.

SteveMacenski:
That sounds reasonable, I'd merge that.

SteveMacenski:
[...]

PR #2536, opened 2021-08-26
navigation: Modify BackUp behavior to respect a demanded 'backup_dura…

philison:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2453 |
| Robotic platform tested on | custom differential drive robot |

---

* The BackUp movement will be terminated as soon as the elapsed time exceeds the `time_allowance`, defined in the `time_allowance` input port of the `BackUp` behavior tree node.
* During the `onRun` call the `back_up_time_end_` time is calculated by adding the `time_allowance` to the current time
* If the `rem_time` (difference between `time_allowance` and the current time) becomes smaller or equal to zero during the cycle updates, the robot will be stopped and the `BackUp` behavior will be exited with `FAILURE`
* If the time_allowance is smaller than zero, this restriction will be ignored and the BackUp movement will continue until the robot reaches the desired goal position
* This commit changes all of the `BackUp` behavior tree node instances in the BT example xml's to make use of the newly added feature

* Update [BackUp Doc](https://github.com/ros-planning/navigation.ros.org/blob/48b9615821bad5b2ffec3a507b0e43a8c92ffe14/configuration/packages/bt-plugins/actions/BackUp.rst)
* Update [nav_to_pose example](https://github.com/ros-planning/navigation.ros.org/blob/48b9615821bad5b2ffec3a507b0e43a8c92ffe14/behavior_trees/trees/nav_to_pose_recovery.rst)
* Update [nav_through_poses example](https://github.com/ros-planning/navigation.ros.org/blob/48b9615821bad5b2ffec3a507b0e43a8c92ffe14/behavior_trees/trees/nav_through_poses_recovery.rst)

---

* Make sure that the newly added feature is included in the test procedure
* Set the default value of the `time_allowance` to a negative value, preventing unintended behavior changes ??? (or select a more appropriate default value)

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/back_up_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/back_up_action.hpp
@@ -54,7 +54,8 @@ public:
     return providedBasicPorts(
       {
         BT::InputPort<double>("backup_dist", 0.15, "Distance to backup"),
-        BT::InputPort<double>("backup_speed", 0.025, "Speed at which to backup")
+        BT::InputPort<double>("backup_speed", 0.025, "Speed at which to backup"),
+        BT::InputPort<double>("time_allowance", 10.0, "Allowed time for reversing")
       });
   }
 };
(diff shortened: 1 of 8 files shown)


SteveMacenski:
Any update?

PR #2540, opened 2021-08-29
Enabled runtime configuration of parameters for Hybrid A*

lucabonamini:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #956  |
| Primary OS tested on | Ubuntu 20.04|
| Robotic platform tested on | Personal Laptop |

---

* Added runtime configuration for Hybrid A* planner parameters
---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
@@ -107,7 +107,7 @@ protected:
   bool _downsample_costmap;
   int _downsampling_factor;
   double _angle_bin_size;
-  int _angle_quantizations;
+  unsigned int _angle_quantizations;
   bool _allow_unknown;
   int _max_iterations;
   SearchInfo _search_info;
--- a/nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
@@ -87,21 +87,41 @@ public:
     const geometry_msgs::msg::PoseStamped & goal) override;
 
 protected:
+  /**
+   * @brief Callback executed when a parameter change is detected
+   * @param event ParameterEvent message
+   */
+  void on_parameter_event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
+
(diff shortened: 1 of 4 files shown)


nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
```diff
@@ -87,21 +87,41 @@ class SmacPlannerHybrid : public nav2_core::GlobalPlanner
...
   std::string _global_frame, _name;
+  float _lookup_table_dim;
   float _tolerance;
+  bool _downsample_costmap;
   int _downsampling_factor;
-  unsigned int _angle_quantizations;
   double _angle_bin_size;
-  bool _downsample_costmap;
-  rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>::SharedPtr _raw_plan_publisher;
+  int _angle_quantizations;
```
SteveMacenski, line 110: You changed the type of this, it should be unsigned

SteveMacenski:
Otherwise LGTM

nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
lucabonamini: But then I cast to `unsigned int` as it was done before

SteveMacenski: I'd just have this keep the types consistent, there are really easy errors within this library that can be introduced if the signed / unsigned ness are messed up, so I'm pretty careful about it.

SteveMacenski: It can be an int from the parameter, but see how it was being used before, it gets  cast immediately and used as unsigned from there on out

SteveMacenski: You had deleted that logic for some reason

lucabonamini: Got it. Fixed in last commit.

SteveMacenski (approved):
Awesome! Great job, looks perfect to me

PR #2542, opened 2021-08-31
Add new test for nav2_regulated_pure_pursuit

sathak93:
---

## Basic Info

| ------ | ----------- |

---

add unit test for findDirectionChange function in nav2_regulated_pure_pursuit

---

#### For Maintainers: 


nav2_regulated_pure_pursuit_controller/test/test_regulated_pp.cpp
```diff
@@ -144,6 +150,26 @@ TEST(RegulatedPurePursuitTest, createCarrotMsg)
   EXPECT_EQ(rtn->point.z, 0.01);
 }
 
+TEST(RegulatedPurePursuitTest, findDirectionChange)
```
SteveMacenski, line 153: You should also test inputs to it that don't have a reverse and make sure that output is right too (that it doesn't have it)

SteveMacenski (changes requested):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -624,7 +624,7 @@ double RegulatedPurePursuitController::findDirectionChange(
   const geometry_msgs::msg::PoseStamped & pose)
 {
   // Iterating through the global path to determine the position of the cusp
-  for (unsigned int pose_id = 1; pose_id < global_plan_.poses.size(); ++pose_id) {
+  for (unsigned int pose_id = 1; pose_id < global_plan_.poses.size() - 1; ++pose_id) {
```
sathak93, line 627: since  pose_id +1 is checked at current step, the loop should end at last-1 pose. without it if block was executed most of the time as the variables were contaminated.

SteveMacenski (approved):
Awesome,  thanks and nice catch!

Issue #2538, opened 2021-08-27
[nav2_behavior_tree] ROS2 Master: Failed to build due to API changes in rclcpp/include/rclcpp/client.hpp


panikgit:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04
- ROS2 Version: Rolling sources
- Version or commit hash: 5a6d4da4d2effe4572132cdb65fb68989d2149f2
- DDS implementation: Cyclone DDS

#### Steps to reproduce issue
Build nav2 for Main Branch Development by following official documentation [https://navigation.ros.org/build_instructions/index.html](url)

#### Expected behavior
Build should finish successfully.

#### Actual behavior
Failed to build nav2_behavior_tree:
```
Starting >>> nav2_costmap_2d
[Processing: nav2_behavior_tree, nav2_costmap_2d]                                                                      
--- stderr: nav2_behavior_tree
In file included from /home/admin/nav2_ws/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/reinitialize_global_localization_service.hpp:20,
                 from /home/admin/nav2_ws/src/navigation2/nav2_behavior_tree/plugins/action/reinitialize_global_localization_service.cpp:16:
/home/admin/nav2_ws/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp: In instantiation of ‘BT::NodeStatus nav2_behavior_tree::BtServiceNode<ServiceT>::tick() [with ServiceT = std_srvs::srv::Empty]’:
/home/admin/nav2_ws/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp:117:18:   required from here
[...]

PR #2543, opened 2021-08-31
Fix deprecated type conversion (#2538)

panikgit:
---
## Basic Info

| Info | Please fill out this column
| ----------- | ----------- |
| Ticket(s) this addresses   | #2538 |

---

Fix build failure (#2538) by replacing implicit deprecated type conversion from `rclcpp::Client::FutureAndRequestId` to `std::shared_future` with call to convenience method `FutureAndRequestId::share()`

#### For Maintainers: 


padhupradheep:
I think the rolling release has been made and now it would be the right time to merge this PR! @SteveMacenski

SteveMacenski:
retriggering CI

SteveMacenski:
Thanks kindly!

Issue #2147, opened 2021-01-08
Create nav2 dynamic composition


SteveMacenski:
Something to contain manually composed nav2 system nodes so that even though the launch system doesn't deal with component-lifecycle nodes well, we can still get the benefits of both (even if not as flexible). This will be good enough for most users as they're not going to be swapping out the task servers. 

Per idea / request from Emerson in the WG meeting

SteveMacenski:
e.x. https://github.com/ros2/demos/blob/master/composition/src/manual_composition.cpp

NithishkumarS:
> Something to contain manually composed nav2 system nodes so that even though the launch system doesn't deal with component-lifecycle nodes well, we can still get the benefits of both (even if not as flexible). This will be good enough for most users as they're not going to be swapping out the task servers.
> 
> Per idea / request from Emerson in the WG meeting
So ideally where should this composition file be located?

NithishkumarS:
So ideally where should this composition file be located?

SteveMacenski:
In a new package, `nav2_composition_bringup`

gezp:
i will do this work in the next few weeks. i make a plan list at first , and any suggestions are welcome.

- [ ] fix export dependency and library
- [ ] add argument (`node_name`,  `node_option`, etc) for Nav2 Nodes.
- [ ] create `nav2_composition` package,  and compose all nav2 nodes in  a single `main()`.
- [ ] add test, and benchmark improvements in performance.
- [ ] improve `nav2_composition` by using `zero copy` between publisher and subscription (if time allots)

gezp:
It provides a composition-based bringup method, which could replace launching Nav2 nodes separately in package `nav2_bringup`, so it's not necessary to create a new package `nav2_composition`, just put it in package `nav2_bringup`.

gezp:
i try to implement `dynamically composed bringup` for Nav2, but i failed because of some problems, i will share my experience:

* first, i try to register Nav2 nodes as the component(for exmaple, register `AmclNode` : https://github.com/gezp/navigation2/blob/dynamic_composed_bringup/nav2_amcl/src/amcl_node.cpp#L1373-L1378)
* then, launch these component in `ComposableNodeContainer` (https://github.com/gezp/navigation2/blob/dynamic_composed_bringup/nav2_bringup/launch/composed_bringup_launch.py#L103-L167)

there are some problems about loading ROS parameters :

[...]

PR #2546, opened 2021-09-06
create new package nav2_composition

gezp:

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |(tickets:  [#2147](https://github.com/ros-planning/navigation2/issues/2147)) |
| Primary OS tested on | (Ubuntu20.04, ROS 2 Galactic from binaries)  |
| Robotic platform tested on |  (gazebo simulation of turtlebot3)   |
---

create a composition-based bringup package for Nav2, in other word, compose all Nav2 nodes in a single process instead of launching these nodes separately, which is useful for embedded systems users that need to make optimizations due to harsh resource constraints.

* create multiple nodes in a single `main()` (manual composition).
* create a bunch of single threaded executors to spin these nodes.

you can find some details about performance test(cpu and memory usage) here
* https://discourse.ros.org/t/nav2-composition/22175

---

try to enable intra-process comms until it's stable in ROS2, which are still kind of under development currently, you can find more detials:
* https://github.com/ros2/rclcpp/issues/1753
* https://github.com/ros2/rclcpp/issues/1750

#### For Maintainers: 

### DIFF
--- a/nav2_composition/CMakeLists.txt
+++ b/nav2_composition/CMakeLists.txt
@@ -1,0 +1,50 @@
+cmake_minimum_required(VERSION 3.5)
+project(nav2_composition)
+
+find_package(ament_cmake REQUIRED)
+find_package(nav2_common REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(nav2_map_server REQUIRED)
+find_package(nav2_amcl REQUIRED)
+find_package(nav2_controller REQUIRED)
+find_package(nav2_planner REQUIRED)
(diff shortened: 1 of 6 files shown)


gezp:
It provides a composition-based bringup method, which could replace launching Nav2 nodes separately in package `nav2_bringup`, so it's not necessary to create a new package `nav2_composition` , just put it in package `nav2_bringup`. 

The PR should be closed and new PR https://github.com/ros-planning/navigation2/pull/2555 is created.

Issue #2374, opened 2021-05-28
Allow planning once and only replan if path has been invalidated


SteveMacenski:
It wouldn't be too hard to create a BT node, add to the planner server an action, or make it part of the main planning task (parameterized) to validate a path.

- March through path
- Check costs on path
- If now in collision, or "sufficiently" more cost than the original path, return failure
- If fails, trigger plannning

SteveMacenski:
Merged!

PR #2548, opened 2021-09-07
added skeleton for is valid path

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Added service to planner server and a client to a new conditional BT node to determine if the global path is no longer feasible. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/CMakeLists.txt
+++ b/nav2_behavior_tree/CMakeLists.txt
@@ -96,6 +96,9 @@ list(APPEND plugin_libs nav2_initial_pose_received_condition_bt_node)
 add_library(nav2_is_battery_low_condition_bt_node SHARED plugins/condition/is_battery_low_condition.cpp)
 list(APPEND plugin_libs nav2_is_battery_low_condition_bt_node)
 
+add_library(nav2_is_path_valid_condition_bt_node SHARED plugins/condition/is_path_valid_condition.cpp)
+list(APPEND plugin_libs nav2_is_path_valid_condition_bt_node)
+
 add_library(nav2_reinitialize_global_localization_service_bt_node SHARED plugins/action/reinitialize_global_localization_service.cpp)
 list(APPEND plugin_libs nav2_reinitialize_global_localization_service_bt_node)
 
--- a/nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
+++ b/nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
@@ -1,0 +1,52 @@
+
+#include <string>
+#include "nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp"
+
+
+namespace nav2_behavior_tree
+{
+ValidPathCondition::ValidPathCondition(
+  const std::string & condition_name,
+  const BT::NodeConfiguration & conf)
+: BT::ConditionNode(condition_name, conf)
+{
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+}
+
+BT::NodeStatus ValidPathCondition::tick()
+{
+
+  // Grab the path
+  nav_msgs::msg::Path path;
+  getInput("path", path);
+
+  auto request = std::make_shared<nav2_msgs::srv::IsPathValid::Request>();
+
+  request->path = path; 
+
+  //To be implemented
+  auto result = client->async_send_request(request);
+  // Wait for the result.
+  if (rclcpp::spin_until_future_complete(node_, result) ==
+    rclcpp::FutureReturnCode::SUCCESS)
+  {
+    RCLCPP_DEBUG(node_->get_logger(), "Got response");
+
+    if( result.get()->response.data )
+    {
+      return BT::NodeStatus::SUCCESS;
+    }
+    return BT::NodeStatus::FAILURE;
+  } else {
+    RCLCPP_DEBUG(node_->get_logger(), "No response");
+    return BT::NodeStatus::FAILURE;
+  }
+}
+
+} // namespace nav2_behavior_tree
+
+#include "behaviortree_cpp_v3/bt_factory.h"
+BT_REGISTER_NODES(factory)
+{
+  factory.registerNodeType<nav2_behavior_tree::ValidPathCondition>("ValidPathCondition");
+}
--- a/nav2_msgs/srv/IsPathValid.srv
+++ b/nav2_msgs/srv/IsPathValid.srv
@@ -1,0 +1,5 @@
+#Determine if the current path is still valid
+
+nav_msgs/Path path
+---
+std_msgs/Bool response
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -510,4 +510,14 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
   }
 }
 
+void PlannerServer::isPathValid(const std::shared_ptr<nav2_msgs::srv::IsPathValid::Request> request, 
+                   std::shared_ptr<nav2_msgs::srv::IsPathValid::Response> response)
+{
+  //Need to implement
+  nav_msgs::msg::Path path = request->path;
+  std_msgs::msg::Bool msgData;
+  msgData.data = true; 
+  response->response = msgData; 
+}
+
 }  // namespace nav2_planner
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
@@ -1,0 +1,51 @@
+#ifndef NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__VALID_PATH_CONDITION_HPP_
+#define NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__VALID_PATH_CONDITION_HPP_
+
+#include "behaviortree_cpp_v3/condition_node.h"
+#include "rclcpp/rclcpp.hpp"
+#include "nav2_msgs/srv/is_path_valid.hpp"
+
+namespace nav2_behavior_tree
+{
+/**
+ * @brief A BT::ConditionNode that returns SUCCESS
+ * if the path is still valid and returns FAILURE otherwise
+ */
+class ValidPathCondition : public BT::ConditionNode
+{
+public:
+  /**
+   * @brief A constructor for nav2_behavior_tree::ValidPathCondition
+   * @param condition_name Name for the XML tag for this node
+   * @param conf BT node configuration
+   */
+  ValidPathCondition(
+    const std::string & condition_name,
+    const BT::NodeConfiguration & conf);
+
+  ValidPathCondition() = delete;
+
+  /**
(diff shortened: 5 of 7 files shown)


SteveMacenski:
https://github.com/ros-planning/navigation2/issues/2374 please link the appropriate ticket this is to support

nav2_msgs/srv/IsPathValid.srv
```diff
@@ -0,0 +1,5 @@
+#Determine if the current path is still valid
+
+nav_msgs/Path path
+---
+std_msgs/Bool response
```
SteveMacenski, line 5: new line at EOF

Also, `response` to something more telling, `is_valid` maybe? 

Also, maybe good to list back the poses in the path that were invalid, might help. I think it would be good to list back the pose index's that failed the check

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -0,0 +1,52 @@
...
+    return BT::NodeStatus::FAILURE;
+  }
+}
+
+} // namespace nav2_behavior_tree
+
+#include "behaviortree_cpp_v3/bt_factory.h"
+BT_REGISTER_NODES(factory)
+{
+  factory.registerNodeType<nav2_behavior_tree::ValidPathCondition>("ValidPathCondition");
```
SteveMacenski, line 51: "IsPathValid"

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -0,0 +1,52 @@
+
+#include <string>
+#include "nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp"
+
+
+namespace nav2_behavior_tree
+{
+ValidPathCondition::ValidPathCondition(
```
SteveMacenski, line 8: rename is IsPathValidCondition

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -0,0 +1,52 @@
...
+  getInput("path", path);
+
+  auto request = std::make_shared<nav2_msgs::srv::IsPathValid::Request>();
+
+  request->path = path; 
+
+  //To be implemented
+  auto result = client->async_send_request(request);
+  // Wait for the result.
+  if (rclcpp::spin_until_future_complete(node_, result) ==
```
SteveMacenski, line 30: See the timeout logic for action nodes -- if we spin, we should have a timeout in case of networking failures (don't want to block indefinitely)

SteveMacenski: Is "RUNNING" a valid state for condition nodes? If so, might want to consider also the timeout based on the behavior tree tick rate so that we can tick the tree at a constant rate and spin again until the final timeout is reached in behavior tree spin rate increments.

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -0,0 +1,52 @@
+
+#include <string>
+#include "nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp"
+
```
SteveMacenski, line 4: The spacing in these file are weird, but that's not really the point in this PR so I'll not mention them all

nav2_msgs/srv/IsPathValid.srv
SteveMacenski: This can also just be bool, no need for std_msgs/Bool. That way its `.response` not `.response.data`

SteveMacenski (changes requested):

nav2_msgs/srv/IsPathValid.srv
jwallace42: Okay, so I would have a new is_valid data type that holds a boolean and the invalid poses.

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Yeah.. that is way better

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Conditional BT nodes can only return SUCCESS or FAILURE. Would this better suited as action node to allow for RUNNING to be returned when waiting for the service(or fail)?

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: I will fix this

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
SteveMacenski: This should be fast, but once you have it implemented, it would be good to benchmark how long it takes. If its in line with the tick rate, lets just keep it a condition. I think semantically that makes more sense. If it takes too long though, we might want to weight the benefits / costs.

nav2_msgs/srv/IsPathValid.srv
SteveMacenski: Not a new struct or something, just include multiple fields in the response. `bool` for the fail/success and a `int[]` for the indices

jwallace42: okay cool

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Done

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Done

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Done

nav2_msgs/srv/IsPathValid.srv
jwallace42: Done

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -96,6 +96,9 @@ list(APPEND plugin_libs nav2_initial_pose_received_condition_bt_node)
 add_library(nav2_is_battery_low_condition_bt_node SHARED plugins/condition/is_battery_low_condition.cpp)
 list(APPEND plugin_libs nav2_is_battery_low_condition_bt_node)
 
+add_library(nav2_is_path_valid_condition_bt_node SHARED plugins/condition/is_path_valid_condition.cpp)
```
SteveMacenski, line 99: Make sure to add this lib to the list in BT navigator and the configuration yaml.

nav2_msgs/srv/IsPathValid.srv
```diff
@@ -0,0 +1,6 @@
+#Determine if the current path is still valid
+
+nav_msgs/Path path
+---
+bool is_valid
+int32[] invalid_pose_indices
```
SteveMacenski, line 6: Look over the git view, you can see a few files need lines at EOF

nav2_planner/src/planner_server.cpp
```diff
@@ -62,6 +62,7 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
 
   // Launch a thread to run the costmap node
   costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);
+
```
SteveMacenski, line 65: remove this line added between the object and `}`

nav2_planner/src/planner_server.cpp
```diff
@@ -510,4 +517,24 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
   }
 }
 
+void PlannerServer::isPathValid(
+  const std::shared_ptr<nav2_msgs::srv::IsPathValid::Request> request,
+  std::shared_ptr<nav2_msgs::srv::IsPathValid::Response> response)
+{
+  unsigned int threshold = 10;
```
SteveMacenski, line 524: This seems like you should be using the pre-defined costmap types for LETHAL if just straight up checking if in collision now.

nav2_planner/src/planner_server.cpp
```diff
@@ -510,4 +517,24 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+  unsigned int mx = 0;
+  unsigned int my = 0;
+
+  response->is_valid = true;
+  for (unsigned int i = 0; i < request->path.poses.size(); ++i) {
+    costmap_->worldToMap(
+      request->path.poses[i].pose.position.x, request->path.poses[i].pose.position.y,
+      mx, my);
+    if (static_cast<double>(costmap_->getCost(mx, my)) > threshold) {
```
SteveMacenski, line 533: You're comparing a double to an unsigned int

nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_for_invalid_path.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+  This Behavior Tree replans the global path when it becomes invalid
+-->
+
+<root main_tree_to_execute="MainTree">
```
SteveMacenski, line 5: It would be nice if this was setup with the recovery stuff as well (e.g. modify the current default BT and add in the changes for IsPathValid and save as a new file). That way out of the box someone gets the good nav BT stuff and this behavior at the same time versus a simplified demo BT

nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_for_invalid_path.xml
```diff
@@ -0,0 +1,17 @@
...
+  <BehaviorTree ID="MainTree">
+    <PipelineSequence name="NavigateWithReplanning">
+        <Sequence>
+            <IsPathValid>
```
SteveMacenski, line 9: This still wouldn't update except for the rate controller passing. I think in this situation, you should remove the rate controller, no? That way if it `SUCCEED` then we shouldn't plan, but anytime we `FAIL` we do. I think then the parent node here would be a reactive fallback node instead, yes? Since the compute path should only be done if we fail, and pass when we succeed.

SteveMacenski (changes requested):
We had some discussion about also failing if the cost "changed by enough" vs only had a lethal entry in it. Doesn't necessarily need to be part of this PR, but  something to think about. I'd put this out of scope for the current PR though. 

Also, I'm a little concerned about what happens if a path point already passed (e.g. we're at pt 100) like pt 35 and it becomes LETHAL. Since we've already passed it, that shouldn't really invalidate the path. That begs the question if this should prune the path to only check parts "ahead" of the robot (potentially by a parameterization that's defaulted to `true` so we can still check the full path if someone actually wants that)

SteveMacenski:
I think you need to pull in main / rebase in order to build. There was a breaking API change a few weeks ago to Rolling

Issue #1899, opened 2020-07-30
Collision safety node 


SteveMacenski:
While this is an area I've typically not found belonging to navigation, this has come up so many times maybe its worth just including it. The goal is to make a node that takes in raw sensor data from a lidar (2D, 3D, I suppose a depth camera, just something that  gives ranges in the forward motion direction) and either stop or slow the robot to avoid collisions. 

We've added heartbeat server alive + lifecycle transition support and this falls into the same category of safety awareness / promises we can round it off with (e.g. we now know within a timeout everything is running properly and activated, this will now ensure no collisions using raw safety data which nicely rounds off the minimum needs for a safe robot regardless of algorithm for planning and control or localization quality).

In most industrial situations, you might use something like a SICK lidar that has this capability built in and is safety rated. For smaller robots falling under less strict functional safety requirements and budgets, a laser like this may be out of scope and needing a solution based on the ROS sensors they have available. It is not a suitable safety certification alternative, but will give you comparable non-certified capability as long as the sensor and ROS driver are publishing information.

There are 2 major options (probably do both and give the user option of which to use):
- Define a bounding box and a point threshold and if > N points appear in this box in front of the robot, stop it. Usually with a larger bounding box in front of the stop box to warn of a collision and slow the maximum speeds.
- Look at sensor data and look at the closest few points. With the current speed, estimate the time to collision to those points. If the time is < N seconds (1, 5, 10, etc), slow the robot until its = N seconds. The effect here would be to make the robot always N seconds from a collision and continuously scale down the speeds.  

This can be trivially made in about a day:
- Transform data from source frame to base_link
- Take in user-generated box sizes
- Check if data in boxes -> react
- Take current odometry speed, do basic kinematics to tell the time to collision. If < N, limit speed. 

Useful for both autonomous navigation and as an important element in an assisted teleop system (does this replace the assisted part?)

mikeferguson:
I would actually suggest not doing this - for two reasons:

1. To my knowledge, there aren't any standards out there that define "less strict functional safety requirements" for smaller robots. Sure, people are doing it - but largely because they've decided they can assume and insure out the liability related. Sure you've got a license that says "no warranty" - but most of these licenses haven't actually gone through the courts, especially with regards to limited liability claims.
[...]

PR #2554, opened 2021-09-07
Navigation Safety Node - Summer 2021

Spnetic-5:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (tickets: #1899 ) |
| Primary OS tested on | (Ubuntu20.04, ROS 2 Galactic) |
| Robotic platform tested on | (Not tested yet) |

---

Creation of a ROS 2 node that will prevent robot collision based on lidar and/or RGBD data
* converted the scan data into pointcloud2 & then transformed it into the base_frame parameter value using TF.
* created a new utility function called makeVectorPointsFromString under the nav2_util package, which takes a string of points as an input parameter returns a vector of points.
* created function for defining a safety polygon for the robot using input parameters.
*  made multiple Laserscan subscribers to multiple topics which are passed as the vector of strings parameter.
* created detectPoints function to detect if the detected points in pointcloud are inside the safety polygon or not.

---

* I think there might be some optimizations to be made in the current code.
* need to test the working of detectPoints function for checking if it covers all the edge cases.
* use the zone_action param value to trigger a reaction using another controller node.
 
#### For Maintainers: 


nav2_safety_nodes/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+
+nav2_package()
+
+link_libraries(stdc++fs)
+
+include_directories(
+  include
+)
+# prevent pluginlib from using boost
+# target_compile_definitions(nav2_safety_nodes_core PUBLIC "PLUGINLIB__DISABLE_BOOST_FUNCTIONS")
```
SteveMacenski, line 34: remove commented out code

nav2_safety_nodes/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+)
+
+add_executable(nav2_safety_node
+  src/main.cpp
+)
+
+target_link_libraries(nav2_safety_node
+  nav2_safety_node_core
+)
+
```
SteveMacenski, line 67: 1 space between each thing -- apply for the full file. Lots of irregular spacing below

nav2_safety_nodes/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+if(BUILD_detect_testING)
+  find_package(ament_lint_auto REQUIRED)
+  # the following line skips the linter which checks for copyrights
+  # uncomment the line when a copyright and license is not present in all source files
+  set(ament_cmake_copyright_FOUND TRUE)
+  # the following line skips cpplint (only works in a git repo)
+  # uncomment the line when this package is not in a git repo
+  set(ament_cmake_cpplint_FOUND TRUE)
+  ament_lint_auto_find_detect_test_dependencies()
+
```
SteveMacenski, line 105: This is where you would `add_directory(test)` for adding your tests

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
+#ifndef NAV2_SAFETY_NODES__NAV2_SAFETY_NODE_HPP_
```
SteveMacenski, line 1: Needs a apache 2.0 copyright header

nav2_safety_nodes/package.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>nav2_safety_nodes</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="saurabhpowar1823@gmail.com">ssp</maintainer>
+  <license>TODO: License declaration</license>
```
SteveMacenski, line 8: Apache 2.0 for the license, add me as a maintainer

SteveMacenski: ALso make the version match the versions in the current packages in `main`

SteveMacenski: Add a description too in `<description>`

nav2_safety_nodes/package.xml
```diff
@@ -0,0 +1,35 @@
...
+  <depend>sensor_msgs</depend>
+  <depend>geometry_msgs</depend>
+  <depend>visualization_msgs</depend>
+  <depend>laser_geometry</depend>
+  <depend>nav2_util</depend>
+  <depend>pluginlib</depend>
+  <depend>tf2</depend>
+  <depend>tf2_geometry_msgs</depend>
+  <depend>tf2_ros</depend>
+  <depend>tf2_sensor_msgs</depend>
```
SteveMacenski, line 25: Fix irregular spacing -- also verify that you have all of your dependencies listed here.

nav2_safety_nodes/src/main.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  auto node = std::make_shared<nav2_safety_nodes::SafetyZone>();
+  rclcpp::spin(node->get_node_base_interface());
+  rclcpp::shutdown();
+
+  return 0;
+}
```
SteveMacenski, line 14: needs space at EOF

nav2_util/include/nav2_util/string_utils.hpp
```diff
@@ -23,22 +25,35 @@ namespace nav2_util
...
+std::vector<std::vector<float>> parseVVF(const std::string & input, std::string & error_return);
+
+/**
+ * @brief Make the vector from the given string.
+ * @param str_pts
+ * @param vector_pts
+ * Format should be bracketed array of arrays of floats, like so: [[1.0, 2.2], [3.3, 4.2], ...]
+ *
+ */
+bool makeVectorPointsFromString(const std::string & safety_zone_str,
```
SteveMacenski, line 54: These variable names should be more general for the general approach that this function is. Its just a param string in and a set of points out

nav2_util/src/string_utils.cpp
```diff
@@ -45,4 +51,117 @@ Tokens split(const string & tokenstring, char delimiter)
   return tokens;
 }
 
+
+/** @brief Parse a vector of vector of floats from a string.
+ * @param input
+ * @param error_return
+ * Syntax is [[1.0, 2.0], [3.3, 4.4, 5.5], ...] */
+std::vector<std::vector<float>> parseVVF(const std::string & input, std::string & error_return)
```
SteveMacenski, line 59: Please verify for me: this is an exact  copy paste from costmap 2d, right?

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+#include <memory>
+#include <string>
+#include <vector>
+#include <queue>
+
+#include "rclcpp/rclcpp.hpp"
+#include "visualization_msgs/msg/marker.hpp"
+
+// #include "tf2/convert.h"
```
SteveMacenski, line 12: remove commented code

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+#include "sensor_msgs/msg/point_cloud2.hpp"
+#include "sensor_msgs/point_cloud2_iterator.hpp"
+#include "laser_geometry/laser_geometry.hpp"
+#include "geometry_msgs/msg/polygon_stamped.hpp"
+#include "geometry_msgs/msg/point.hpp"
+#include "nav2_msgs/msg/speed_limit.hpp"
+#include "nav2_util/lifecycle_node.hpp"
+#include "nav2_util/string_utils.hpp"
+namespace nav2_safety_nodes
```
SteveMacenski, line 27: add proper spacing between namespace, includes, and class definitions. See other files for examples.

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+    SafetyZone();
+    /**
+     * @brief A destructor for nav2_safety_nodes::SafetyZone class
+     */
+    ~SafetyZone();
+
+    /**
+     * @brief Configures member variables
+     *
+     * Initializes action server for "follow_waypoints"
```
SteveMacenski, line 44: I don't think this is follow_waypoints :wink:

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+    nav2_util::CallbackReturn on_cleanup(const rclcpp_lifecycle::State & state) override;
+    /**
+     * @brief Called when in shutdown state
+     * @param state Reference to LifeCycle node state
+     * @return SUCCESS or FAILURE
+     */
+    nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
+
+    /**
+       * @brief Get the pose of the robot in the global frame of the costmap
```
SteveMacenski, line 75: indentation

SteveMacenski: Also this is all commented out --- remove?

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+       * @return True if the pose was set successfully, false otherwise
+       */
+    // bool getRobotPose(geometry_msgs::msg::PoseStamped & global_pose);
+    
+    // The Logger object for logging
+    rclcpp::Logger logger_{rclcpp::get_logger("nav2_safety_nodes")};
+
+protected:
+    // The local node
+    std::vector<geometry_msgs::msg::Point> safety_zone;
```
SteveMacenski, line 86: Missing traling underscore for a class member variable

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+    int zone_num_pts_{0};
+    std::string base_frame_;   ///< The frame_id of the robot base
+    double tf_tolerance_{};
+    std::vector<std::string> scan_topics_;
+    std::string speed_limit_topic_;
+
+    /**
+     * @brief Initialize required ROS transformations
+     */
+    void initTransforms();
```
SteveMacenski, line 104: All functions go above all variables

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+    std::shared_ptr<tf2_ros::Buffer> tf2_;
+    std::shared_ptr<tf2_ros::TransformListener> tf2_listener_;
+
+    // Used to project laser scans into point clouds
+    laser_geometry::LaserProjection projector_;
+    std::queue<sensor_msgs::msg::PointCloud2> sensor_data;
```
SteveMacenski, line 110: missing the trailing underscore

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+       * @brief Get the pose of the robot in the global frame of the costmap
+       * @param global_pose Will be set to the pose of the robot in the global frame of the costmap
+       * @return True if the pose was set successfully, false otherwise
+       */
+    // bool getRobotPose(geometry_msgs::msg::PoseStamped & global_pose);
+    
+    // The Logger object for logging
+    rclcpp::Logger logger_{rclcpp::get_logger("nav2_safety_nodes")};
```
SteveMacenski, line 82: Why is this public?

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+     */
+    void timerCallback();
+
+    /**
+     * @brief  A callback to handle buffering LaserScan messages
+     * @param message The message returned from a message notifier
+     */
+    void laserCallback(const sensor_msgs::msg::LaserScan::SharedPtr message);
+    
+    double dotProduct(const Eigen::Vector3d &pt1,
```
SteveMacenski, line 137: Why new line?

SteveMacenski: Also linting, there are mising spaces

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,154 @@
...
+        const Eigen::Vector3d &pt2);
+
+    /**
+     * @brief  Function for detecting if point is inside or outside of safety zone
+     * @param message Pointcloud & safetyzone
+     * @return Number of points Inside
+     */
+    int detectPoints(
+      const sensor_msgs::msg::PointCloud2 &cloud,
+      std::vector<geometry_msgs::msg::Point> safety_zone);
```
SteveMacenski, line 147: Should be a reference, also the cloud reference is missing a space

nav2_safety_nodes/launch/nav2_safety_node.launch.py
```diff
@@ -0,0 +1,44 @@
+from launch import LaunchDescription
+from launch.actions import DeclareLaunchArgument
+from launch.actions import ExecuteProcess
+from launch.substitutions import LaunchConfiguration
+from launch_ros.actions import Node
+
+import yaml
+
+
+def generate_launch_description():
```
SteveMacenski, line 10: What is this launch file for?

SteveMacenski: Looks like it should be removed

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+{
+
+SafetyZone::SafetyZone()
+: nav2_util::LifecycleNode("SafetyZone", "", false)
+{
+  logger_ = get_logger();
+  RCLCPP_INFO(logger_, "Creating Safety Polygon");
+
+  // Vector of string for multiple LaserScan topics
+  const std::vector<std::string> scan{};
```
SteveMacenski, line 21: This variable should be called `scan_topics` and it should contain a single entry `scan` so by default it uses the scan topic if none other are given

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  declare_parameter("zone_action", rclcpp::ParameterValue(0.0));
+  declare_parameter("zone_priority", rclcpp::ParameterValue(1));
+  declare_parameter("zone_num_pts", rclcpp::ParameterValue(1));
+  declare_parameter("base_frame", std::string("base_link"));
+  declare_parameter("tf_tolerance", rclcpp::ParameterValue(0.01));
+  declare_parameter("scan_topics", scan);
+  declare_parameter("speed_limit_topic", rclcpp::ParameterValue("speed_limit"));
+}
+
+SafetyZone::~SafetyZone()
```
SteveMacenski, line 33: If you do nothing here, then remove this from the cpp file and for the destructor in the hpp file use `= default` for the default destructor behavior

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  return nav2_util::CallbackReturn::SUCCESS;
+}
+
+nav2_util::CallbackReturn
+SafetyZone::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
+{
+  RCLCPP_INFO(logger_, "Deactivating");
+  safety_polygon_pub_->on_deactivate();
+  point_cloud_pub_->on_deactivate();
+  speed_limit_pub_->on_deactivate();
```
SteveMacenski, line 63: Timer should be cancelled

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+nav2_util::CallbackReturn
+SafetyZone::on_shutdown(const rclcpp_lifecycle::State &)
+{
+  RCLCPP_INFO(logger_, "Shutting down");
+  return nav2_util::CallbackReturn::SUCCESS;
+}
+
+// Get Parameters
+void
+SafetyZone::getParameters()
```
SteveMacenski, line 87: Delete this function and move its contents into `on_configure`

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+{
+  RCLCPP_DEBUG(logger_, " getParameters");
+  std::string speed_limit_topic;
```
SteveMacenski, line 90: This variable is unused, remove

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+{
+  RCLCPP_INFO(logger_, "Shutting down");
+  return nav2_util::CallbackReturn::SUCCESS;
+}
+
+// Get Parameters
+void
+SafetyZone::getParameters()
+{
+  RCLCPP_DEBUG(logger_, " getParameters");
```
SteveMacenski, line 89: remove this logging, replace with an INFO message that has actual sentence structure, not just debugging artifacts.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  // If the safety_polygon has been specified, it must be in the correct format
+  if (safety_polygon_ != "" && safety_polygon_ != "[]") {
+    // Polygon parameter has been specified, polygon -> point vector(safety_zone)
+    std::vector<geometry_msgs::msg::Point> safety_zone;
+    /// string of polygon points and returns a polygon vector
+    nav2_util::makeVectorPointsFromString(safety_polygon_, safety_zone);
+  } else {
+    // Polygon provided but invalid, so stay with the radius
+    RCLCPP_ERROR(
+      logger_, "The safety_polygon is invalid: \"%s\" :) ",
```
SteveMacenski, line 110: make this a fatal error and `throw` an exception

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+      safety_polygon_.c_str());
+  }
+}
+
+void
+SafetyZone::initTransforms()
+{
+  RCLCPP_INFO(get_logger(), "initTransforms");
```
SteveMacenski, line 118: Delete this function and move its contents into on_configure

On logging, same with configure, just have a log for in on configure, and I'm setting up. Something simple.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  auto timer_interface = std::make_shared<tf2_ros::CreateTimerROS>(
+    get_node_base_interface(), get_node_timers_interface());
+  tf2_->setCreateTimerInterface(timer_interface);
+  tf2_listener_ = std::make_unique<tf2_ros::TransformListener>(*tf2_);
+}
+
+// Publishers and subscribers
+void
+SafetyZone::initPubSub()
+{
```
SteveMacenski, line 130: Delete function, move to on activate, remove log, add a log for for activating.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  RCLCPP_INFO(logger_, "initPubSub");
+  // Create the publishers and subscribers
+  safety_polygon_pub_ = create_publisher<geometry_msgs::msg::PolygonStamped>(
+    "published_polygon", rclcpp::SystemDefaultsQoS());
+  // Pointcloud publisher
+  point_cloud_pub_ = create_publisher<sensor_msgs::msg::PointCloud2>(
```
SteveMacenski, line 136: `point_cloud_pub_` is unused, delete from cpp and  header files

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  RCLCPP_INFO(logger_, "Subscribing to scan topics");
+  for (unsigned int i = 0; i < scan_topics_.size(); i++) {
+    scan_subscribers_.push_back(
+      create_subscription<sensor_msgs::msg::LaserScan>(
+        scan_topics_[i].c_str(), rclcpp::SystemDefaultsQoS(),
+        std::bind(&SafetyZone::laserCallback, this, std::placeholders::_1)));
+  }
+  // Timer -> 10hzs
+  timer_ = create_wall_timer(
+    100ms, std::bind(&SafetyZone::timerCallback, this));
```
SteveMacenski, line 153: Add a parameter for `update_frequency` that should be an `double` in the units of seconds, defaulting to 0.10 for 100ms. Then use chrono to convert that to a duration that you can use here. Check elsewhere in the codebase for examples.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+    sensor_data.push(*cloud);
+  }
+}
+
+double
+SafetyZone::dotProduct(
+  const Eigen::Vector3d & pt1,
+  const Eigen::Vector3d & pt2)
+{
+  return pt1[0] * pt2[1] - pt1[1] * pt2[0];
```
SteveMacenski, line 191: Use eigens `dot()` function for vector 3d. Remove this function

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+      RCLCPP_INFO(logger_, "No");
+    }
+  }
+  return points_inside;
+}
+
+void
+SafetyZone::timerCallback()
+{
+  std::unique_ptr<nav2_msgs::msg::SpeedLimit> msg =
```
SteveMacenski, line 238: Only create this message inside of the `if` condition, no need to allocate memory if not being used yet.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+    std::make_unique<nav2_msgs::msg::SpeedLimit>();
+  while (!sensor_data.empty()) {
+    if (detectPoints(sensor_data.front(), safety_zone) >= zone_num_pts_) {
+      msg->header.frame_id = base_frame_;
+      msg->header.stamp = clock_->now();
+      msg->speed_limit = zone_action_;
```
SteveMacenski, line 244: Is this a proper way to populate that message?

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  return points_inside;
+}
+
+void
+SafetyZone::timerCallback()
+{
+  std::unique_ptr<nav2_msgs::msg::SpeedLimit> msg =
+    std::make_unique<nav2_msgs::msg::SpeedLimit>();
+  while (!sensor_data.empty()) {
+    if (detectPoints(sensor_data.front(), safety_zone) >= zone_num_pts_) {
```
SteveMacenski, line 241: I messed up here. We need to move the `detectPoints` to set a local variable `num_points_in_zone` to the function so that multiple sensor readings can contribute to this count. Then the if statement should check if `num_points_in_zone` is larger than the zone num pts.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const sensor_msgs::msg::LaserScan::SharedPtr message)
+{
+  // project the laser into a point cloud
+  const sensor_msgs::msg::PointCloud2::SharedPtr cloud;
+
+  try {
+    projector_.projectLaser(*message, *cloud);
+  } catch (...) {
+    RCLCPP_WARN(
+      logger_,
```
SteveMacenski, line 168: Line line extended? this can all be on 1 line

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+    100ms, std::bind(&SafetyZone::timerCallback, this));
+  RCLCPP_INFO(logger_, "Subscribed to scan topics");
+}
+
+void
+SafetyZone::laserCallback(
+  const sensor_msgs::msg::LaserScan::SharedPtr message)
+{
+  // project the laser into a point cloud
+  const sensor_msgs::msg::PointCloud2::SharedPtr cloud;
```
SteveMacenski, line 162: This doesn't need to be a shared pointer, just make it stack memory, e.g. `sensor_msgs::msg::PointCloud2`, it also doesn't and shouldn't be `const`.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+      "Unable to project laser into cloud");
+  }
+
+  // Transform the cloud if necessary
+  if (!base_frame_.empty() && cloud->header.frame_id != base_frame_) {
+    try {
+      *cloud = tf2_->transform(*cloud, base_frame_, tf2::durationFromSec(tf_tolerance_));
+      sensor_data.push(*cloud);
+    } catch (tf2::TransformException & ex) {
+      RCLCPP_ERROR_STREAM(logger_, "Transform failure: " << ex.what());
```
SteveMacenski, line 178: Don't use STREAM -- there's nothing wrong with it, just by convention we don't use it in Nav2. Use the `%s` instead.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const Eigen::Vector3d & pt2)
+{
+  return pt1[0] * pt2[1] - pt1[1] * pt2[0];
+}
+
+// Function for detecting if cloud points are inside safety polygon or not
+int
+SafetyZone::detectPoints(
+  const sensor_msgs::msg::PointCloud2 & cloud,
+  std::vector<geometry_msgs::msg::Point> safety_zone)
```
SteveMacenski, line 198: All arguemnts to all functions should be const references

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+        count_same_side_results++;
+      } else if (x < 0) {
+        count_same_side_results--;
+      } else {
+        on_edge++;
+      }
+    }
+    // checking if point is on same side of all edges
+    if (count_same_side_results == n || count_same_side_results == (-1) * n || on_edge == 2) {
+      RCLCPP_INFO(logger_, "Yes");
```
SteveMacenski, line 226: remove these logs

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+SafetyZone::timerCallback()
+{
+  std::unique_ptr<nav2_msgs::msg::SpeedLimit> msg =
+    std::make_unique<nav2_msgs::msg::SpeedLimit>();
+  while (!sensor_data.empty()) {
+    if (detectPoints(sensor_data.front(), safety_zone) >= zone_num_pts_) {
+      msg->header.frame_id = base_frame_;
+      msg->header.stamp = clock_->now();
+      msg->speed_limit = zone_action_;
+      speed_limit_pub_->publish(std::move(msg));
```
SteveMacenski, line 245: In the if statement, add a log saying that more points than N were detected within the zone, slowing the robot by action M

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const int n = safety_zone.size();
+  sensor_msgs::PointCloud2ConstIterator<float> iter_x(cloud, "x");
+  sensor_msgs::PointCloud2ConstIterator<float> iter_y(cloud, "y");
+  sensor_msgs::PointCloud2ConstIterator<float> iter_z(cloud, "z");
+  // iterating through cloud points
+  for (; iter_x != iter_x.end(); ++iter_x, ++iter_y, ++iter_z) {
+    double px = *iter_x, py = *iter_y, pz = *iter_z;
+    // iterating through polygon points
+    for (int i = 0; i < n; ++i) {
+      geometry_msgs::msg::Point a = safety_zone[i];
```
SteveMacenski, line 211: Name these variables much, much better. a,b,x mean absolutely nothing

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+      geometry_msgs::msg::Point b = safety_zone[(i + 1) % n];
+      Eigen::Vector3d affine_segment = {b.x - a.x, b.y - a.y, b.z - a.z};
+      Eigen::Vector3d affine_point = {px - a.x, py - a.y, pz - a.z};
+      double x = dotProduct(affine_segment, affine_point);
+      if (x > 0) {
+        count_same_side_results++;
+      } else if (x < 0) {
+        count_same_side_results--;
+      } else {
+        on_edge++;
```
SteveMacenski, line 221: Count the edge points as inside the polygon for all situations

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+      }
+    }
+    // checking if point is on same side of all edges
+    if (count_same_side_results == n || count_same_side_results == (-1) * n || on_edge == 2) {
```
SteveMacenski, line 225: This  can just be `abs(count_same_side_results) == n` to handle both the positive and negative cases. 

Also, what's the on_edge == 2 condition?

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,250 @@
+#include <chrono>
```
SteveMacenski, line 1: Package needs a readme file with some basic information like
- What is this
- What does it do
- What are the parameters
- How does it work
- Any user advice required

This is a good example https://github.com/ros-planning/navigation2/tree/main/nav2_smac_planner though its much, much longer than what you need to do here. This is considerably simpler.

SteveMacenski (changes requested):

SteveMacenski:
@Spnetic-5 you need to rebase this PR from galactic to main.

.... more realistically due to you beginning programming, you should just copy paste this entire code from your current branch to another branch that's based on main

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: When `on_edge == 2` point lies on one of the sides of polygon, I have considered all such point as inside of polygon!

SteveMacenski: Couldn't you just make one of the conditions `>=` or `<=` to include this case in their analysis?

nav2_util/src/string_utils.cpp
Spnetic-5: Yes!

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: I don't know the reason but the thing is when I used the inbuilt `dot()` function, I was getting different results for dot product, which in turn resulted in false detection of points, so I continued using the function created.

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: No, If I do so then `abs(count_same_side_results) == n` this condition fails, because it becomes `n+2` for inside and remains n for outside.

nav2_safety_nodes/include/nav2_safety_nodes/nav2_safety_node.hpp
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) 2021
```
SteveMacenski, line 1: Add your name to all of these copyright lines. It should be `Copyright (c) 2021 <your name>`

nav2_safety_nodes/package.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>nav2_safety_nodes</name>
+  <version>1.0.6</version>
+  <description>A Navigation safety node server</description>
+  <maintainer email="stevenmacenski@gmail.com">Steve Macenski</maintainer>
```
SteveMacenski, line 7: Add another maintainer, yourself, in addition to me

nav2_safety_nodes/package.xml
```diff
@@ -0,0 +1,33 @@
...
+  <depend>tf2_sensor_msgs</depend>
+
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>nav2_lifecycle_manager</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
+</package>
```
SteveMacenski, line 33: line at EOF, look at the git diff and add all missing to all files

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+  // If the safety_polygon has been specified, it must be in the correct format
+  if (safety_polygon_ != "" && safety_polygon_ != "[]") {
+    // Polygon parameter has been specified, polygon -> point vector(safety_zone)
+    std::vector<geometry_msgs::msg::Point> safety_zone_;
+    /// string of polygon points and returns a polygon vector
+    nav2_util::makeVectorPointsFromString(safety_polygon_, safety_zone_);
+  } else {
+    // Polygon provided but invalid, so stay with the radius
+    throw std::invalid_argument("The safety_polygon is invalid");
```
SteveMacenski, line 72: use runtime_error for all exceptions

nav2_safety_nodes/src/nav2_safety_node.cpp
SteveMacenski: That is because you use the Vector3d type, try the Vector2d type instead. The result is different because you populated the `z` components of the vector which are being used in the computation.

nav2_safety_nodes/src/nav2_safety_node.cpp
SteveMacenski: No, keep the on edge, but reduce the `count_same_side_results == n || count_same_side_results == (-1) * n` to `bs(count_same_side_results) == n`.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// Function for detecting if cloud points are inside safety polygon or not
+int
+SafetyZone::detectPoints(
+  const sensor_msgs::msg::PointCloud2 & cloud,
+  const std::vector<geometry_msgs::msg::Point> & safety_zone)
+{
+  int points_inside = 0;
+  int count_same_side_results = 0, on_edge = 0;
```
SteveMacenski, line 185: You don't reset count_same_side_results or on_edge between point evaluations, this would only work for a single first point!

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  const int n = safety_zone.size();
```
SteveMacenski, line 186: Remove this variable and use `safety_zone.size()` inline where required.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+using namespace std::chrono_literals;
+
+namespace nav2_safety_nodes
+{
+
+SafetyZone::SafetyZone()
+: nav2_util::LifecycleNode("SafetyZone", "", false)
+{
+  logger_ = get_logger();
+  RCLCPP_INFO(logger_, "Creating Safety Polygon");
```
SteveMacenski, line 32:
```suggestion
  RCLCPP_INFO(logger_, "Creating Safety Zone Polygon Node");
```

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  base_frame_ = get_parameter("base_frame").as_string();
+  tf_tolerance_ = get_parameter("tf_tolerance").as_double();
+  scan_topics_ = get_parameter("scan_topics").as_string_array();
+  speed_limit_topic_ = get_parameter("speed_limit_topic").as_string();
+
+  // If the safety_polygon has been specified, it must be in the correct format
+  if (safety_polygon_ != "" && safety_polygon_ != "[]") {
+    // Polygon parameter has been specified, polygon -> point vector(safety_zone)
+    std::vector<geometry_msgs::msg::Point> safety_zone_;
+    /// string of polygon points and returns a polygon vector
```
SteveMacenski, line 68: Remove this comment

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  zone_num_pts_ = get_parameter("zone_num_pts").as_int();
+  base_frame_ = get_parameter("base_frame").as_string();
+  tf_tolerance_ = get_parameter("tf_tolerance").as_double();
+  scan_topics_ = get_parameter("scan_topics").as_string_array();
+  speed_limit_topic_ = get_parameter("speed_limit_topic").as_string();
+
+  // If the safety_polygon has been specified, it must be in the correct format
+  if (safety_polygon_ != "" && safety_polygon_ != "[]") {
+    // Polygon parameter has been specified, polygon -> point vector(safety_zone)
+    std::vector<geometry_msgs::msg::Point> safety_zone_;
```
SteveMacenski, line 67: This should not be instantiated here, it is a class variable, remove this line

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+    /// string of polygon points and returns a polygon vector
+    nav2_util::makeVectorPointsFromString(safety_polygon_, safety_zone_);
+  } else {
+    // Polygon provided but invalid, so stay with the radius
```
SteveMacenski, line 71: This comment makes no sense, remove

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  return nav2_util::CallbackReturn::SUCCESS;
+}
+
+nav2_util::CallbackReturn
+SafetyZone::on_activate(const rclcpp_lifecycle::State & /*state*/)
+{
+  RCLCPP_INFO(logger_, "Activating");
+  // Create the publishers and subscribers
+  safety_polygon_pub_ = create_publisher<geometry_msgs::msg::PolygonStamped>(
+    "published_polygon", rclcpp::SystemDefaultsQoS());
```
SteveMacenski, line 91: Rename topic to "safety_zone"

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  // Multiple Laserscan subscribers
+  scan_subscribers_.reserve(scan_topics_.size());
+  RCLCPP_INFO(logger_, "Subscribing to scan topics");
+  for (unsigned int i = 0; i < scan_topics_.size(); i++) {
+    scan_subscribers_.push_back(
+      create_subscription<sensor_msgs::msg::LaserScan>(
+        scan_topics_[i].c_str(), rclcpp::SystemDefaultsQoS(),
+        std::bind(&SafetyZone::laserCallback, this, std::placeholders::_1)));
+  }
+  // Timer -> 10hzs
```
SteveMacenski, line 104: This isn't 10hz anymore

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  auto update_frequency_s = std::chrono::duration<double>(update_frequency_);
+  auto update_frequency_ms = std::chrono::duration_cast<std::chrono::milliseconds>(update_frequency_s);
```
SteveMacenski, line 106: You don't need to cast from s to ms to then give to the timer, you just need the chrono duration conversion above

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  safety_polygon_pub_ = create_publisher<geometry_msgs::msg::PolygonStamped>(
+    "published_polygon", rclcpp::SystemDefaultsQoS());
+  // Velocity publisher
+  speed_limit_pub_ = create_publisher<nav2_msgs::msg::SpeedLimit>(
+    speed_limit_topic_, rclcpp::QoS(10));
+  // Multiple Laserscan subscribers
+  scan_subscribers_.reserve(scan_topics_.size());
+  RCLCPP_INFO(logger_, "Subscribing to scan topics");
+  for (unsigned int i = 0; i < scan_topics_.size(); i++) {
+    scan_subscribers_.push_back(
```
SteveMacenski, line 99: Create a log in each of these saying what scan topics were subscribed to

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+      create_subscription<sensor_msgs::msg::LaserScan>(
+        scan_topics_[i].c_str(), rclcpp::SystemDefaultsQoS(),
+        std::bind(&SafetyZone::laserCallback, this, std::placeholders::_1)));
+  }
+  // Timer -> 10hzs
+  auto update_frequency_s = std::chrono::duration<double>(update_frequency_);
+  auto update_frequency_ms = std::chrono::duration_cast<std::chrono::milliseconds>(update_frequency_s);
+  timer_ = create_wall_timer(
+  update_frequency_ms, std::bind(&SafetyZone::timerCallback, this));
+  RCLCPP_INFO(logger_, "Subscribed to scan topics");
```
SteveMacenski, line 109: Remove this log

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+  speed_limit_pub_->on_activate();
+
+  return nav2_util::CallbackReturn::SUCCESS;
+}
+
+nav2_util::CallbackReturn
+SafetyZone::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
+{
+  RCLCPP_INFO(logger_, "Deactivating");
+  safety_polygon_pub_->on_deactivate();
```
SteveMacenski, line 120: Reset the subscriptions too for the laser scanner

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,232 @@
...
+SafetyZone::laserCallback(
+  const sensor_msgs::msg::LaserScan::SharedPtr message)
+{
+  // project the laser into a point cloud
+  sensor_msgs::msg::PointCloud2 cloud;
+
+  try {
+    projector_.projectLaser(*message, cloud);
+  } catch (...) {
+    RCLCPP_WARN(logger_, "Unable to project laser into cloud");
```
SteveMacenski, line 153: there should be a return after that error message so that the function returns. If it can't transform, it cannot be used.

SteveMacenski (changes requested):

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: Yes that I have done!

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: No, it isn't working either with Vector2d, giving different results

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: I referred some other packages, I'm not getting what's exactly missing here

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: We need to calculate same_side results for every point in pcl so we need to reset those variables to zero for each point, I think I should move these two lines inside first for loop, right?

nav2_safety_nodes/src/nav2_safety_node.cpp
SteveMacenski: why?

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,229 @@
...
+  // Multiple Laserscan subscribers
+  scan_subscribers_.reserve(scan_topics_.size());
+  for (unsigned int i = 0; i < scan_topics_.size(); i++) {
+    RCLCPP_INFO(logger_, "Subscribing to %s topic", scan_topics_[i].c_str());
+    scan_subscribers_.push_back(
+      create_subscription<sensor_msgs::msg::LaserScan>(
+        scan_topics_[i].c_str(), rclcpp::SystemDefaultsQoS(),
+        std::bind(&SafetyZone::laserCallback, this, std::placeholders::_1)));
+  }
+  auto update_frequency_s_ = std::chrono::duration<double>(update_frequency_);
```
SteveMacenski, line 101: `update_frequency_s_` should not have an underscore, it is not a class variable.

nav2_safety_nodes/src/nav2_safety_node.cpp
SteveMacenski: You put it in cleanup, not in deactivate. They are created in activate, so they need to be destroyed in deactivate.

SteveMacenski (changes requested):
Mostly looks good, still have some open issues left

nav2_safety_nodes/src/nav2_safety_node.cpp
SteveMacenski: yes

nav2_safety_nodes/README.md
```diff
@@ -0,0 +1,13 @@
+# Navigation Safety Node
+
+The Navigation Safety Node is a safety watchdog node to ensure the robot is acting properly and not about to collide with an obstacle. Typical safety-rated lidars will contain “safety zones” whereas if any sensor points are located in a box around the lidar, then the lidar will send a signal to the robot to stop due to a potential collision.
+
+The node sits below the navigation stack but above the robot controller to do the following:
+
+* Take in the current command velocity from navigation and the most recent laser or RGBD scan.
```
SteveMacenski, line 7: This is not what the node does that you have implemented below.

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: This is the comparison between results for 5 test points, dot product of each point with all 5 edges of polygon using both
| dotProduct  | dot() |
| ------------- | ------------- |
| 0.75 0.5 0 0 0.25  | 0.25 0.25 1 0 -0.25  |
| -0.1 0.1 0.9 0.4 0.2  | 0.2 -0.2 0.2 0.45 0.6  |
| 0.25 0 0 0.5 0.75  | 0.75 0.25 0 0 0.25  |
| 0 0.25 1 0.25 0   | -0 -0.25 0.5 0.5 0.5  |
| 0.35 0.3 0.4 0.2 0.25   | 0.25 0.05 0.6 0.2 0.15  |

nav2_safety_nodes/README.md
Spnetic-5: Yes, I thought I'm supposed to readme for what overall package does!

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: I have done that

nav2_safety_nodes/src/nav2_safety_node.cpp
SteveMacenski: I'm not asking about how its different, I'm asking why its different, please assess it.

nav2_safety_nodes/README.md
SteveMacenski: But it doesn't do that, looking over the PR.

nav2_safety_nodes/src/nav2_safety_node.cpp
```diff
@@ -0,0 +1,229 @@
...
+      if (dot > 0) {
+        count_same_side_results++;
+      } else if (dot < 0) {
+        count_same_side_results--;
+      } else {
+        on_edge++;
+      }
+    }
+    // checking if point is on same side of all edges
+    if (abs(count_same_side_results == int(safety_zone.size())) || on_edge == 2) {
```
SteveMacenski, line 206: This is wrong. Can you see why its wrong?

SteveMacenski (changes requested):

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: Yes, My mistake

nav2_safety_nodes/README.md
Spnetic-5: Yes, I'll update the readme again

nav2_safety_nodes/src/nav2_safety_node.cpp
Spnetic-5: It's because in our dot product function we aren't actually calculating dot product it just substitutes the point in the equation of the line segment for determining which side of the line segment does the point lie.

SteveMacenski:
Closing due to lack of response

Issue #2422, opened 2021-06-23
Regulated pure pursuit use footprint not point costs


stevemartinov:
- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - ROS2 Foxy, binaries
- Version or commit hash:
  - ros-foxy-navigation2
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
Just run the planner with default configurations but increse the footprint of the robot to 1 x 1.5 meters.

#### Expected behavior
Robot should navigate and if there is an obstacle next to it, trigger the collision check to true and stop the robot

#### Actual behavior
When visualizing the arc, it actually starts from the base_link (middle of the robot) and the length of the arc does not go beyond the footrpint as the footprint is so long. Therefore, in many cases the planner fails to detect the collision

#### Additional information

I am not sure whether it is a bug as I was not able to properly understand. In README the picture has some lookeahead radius and I presume that the arc should start from the footprint and not from the base link, otherwise, we would need to set the lookahead time to a very large number for the arc to check the collisions beyound the footprint
----

SteveMacenski:
What is your path planner? If you planner isn't generating kinematically feasible paths (e.g. paths your robot can literally drive and be collision free driving) then any Pure Pursuit controller will have problems because they're pure path following (pure "pursuit"). They only see the path and have interest in following that path. 

We also check for collision https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L404-L416 but assumes you have your costmap properly setup with a footprint and inflation layer. We use this to forward project in time to make sure there is no collision in the immediate future https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L373-L397 based on this pure position 2D costmap collision check. 

It is possible to improve this by using the footprint and using the full SE2 footprint checking, but it really depends on "why" your collisions are happening to know if that would help at all.

stevemartinov:
@SteveMacenski I am using a simple global planner (Djikstra, old ROS1 implementation) but I am not sure whether it generates kinematically feasable paths, do you have one in mind that does that?

Yes, I saw that function and I think I can implement a collision checking incorporating a whole footprint. I have a question though, in your function:
[...]

PR #2556, opened 2021-09-08
Footprint collision checking for nav2_regulated_pure_pursuit

sathak93:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2422 |

---

collision checking in nav2_regulated_pure_pursuit is updated to based on footprint cost instead of point cost

---

#### For Maintainers: 


nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -183,13 +189,20 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
+   * @param traverse_unknown ignore UNKNOWN cost (255) as obstuctle
    * @return Whether in collision
    */
-  bool inCollision(const double & x, const double & y);
-
+  bool inCollision(
+    const double & x,
+    const double & y,
+    const double & theta,
+    const nav2_costmap_2d::Footprint & footprint_spec,
```
SteveMacenski, line 204: Remove footprint / traverse unknown, those are both available in class variables that this function can access.

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -260,12 +273,17 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
+  double footprint_cost_;
+  double possible_inscribed_cost_{-1};
 
+  nav2_costmap_2d::Footprint unoriented_footprint_;
   nav_msgs::msg::Path global_plan_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> global_path_pub_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<geometry_msgs::msg::PointStamped>>
   carrot_pub_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> carrot_arc_pub_;
+  std::unique_ptr<nav2_costmap_2d::FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>
```
SteveMacenski, line 285: trailing underscore, not leading

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -260,12 +273,17 @@ class RegulatedPurePursuitController : public nav2_core::Controller
   double rotate_to_heading_min_angle_;
   double goal_dist_tol_;
   bool allow_reversing_;
+  double footprint_cost_;
```
SteveMacenski, line 276: There's no need to have this be a class variable

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -260,12 +273,17 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
+  double possible_inscribed_cost_{-1};
```
SteveMacenski, line 277: Remove this and all the logic around it -- that's specific to planning because we're checking upwards of 1M nodes, here its like 20, so can just check use the actual footprint for all of them.

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -260,12 +273,17 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
 
+  nav2_costmap_2d::Footprint unoriented_footprint_;
```
SteveMacenski, line 279: Remove this, you can get it from the costmap when needed

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -447,13 +464,27 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
     return false;
   }
 
-  unsigned char cost = costmap_->getCost(mx, my);
+  // if the robot is even potentially in an inscribed collision
+  footprint_cost_ = _collision_checker->pointCost(wx, wy);
```
SteveMacenski, line 468: Remove all the logic around pointCost for the same rational I gave on `possible_inscribed_cost_`, we just need to check the footprint only here!

SteveMacenski (changes requested):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
sathak93: what about this one <https://github.com/ros-planning/navigation2/blob/2e6594ca9d7f6652c914f5b5961324bfc15b9e82/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L473>

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -383,10 +389,12 @@ bool RegulatedPurePursuitController::isCollisionImminent(
...
 
   // check current point is OK
-  if (inCollision(robot_pose.pose.position.x, robot_pose.pose.position.y)) {
+  if (inCollision(
+      robot_pose.pose.position.x, robot_pose.pose.position.y,
+      tf2::getYaw(robot_pose.pose.orientation)))
+  {
     return true;
   }
-
```
SteveMacenski, line 389: Readd the 2 spaces (here and L413) please, just a styling thing so I can parse this easier in the future, nothing technically wrong with it.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -434,11 +442,15 @@ bool RegulatedPurePursuitController::isCollisionImminent(
...
 
-bool RegulatedPurePursuitController::inCollision(const double & x, const double & y)
+bool RegulatedPurePursuitController::inCollision(
+  const double & x,
+  const double & y,
+  const double & theta)
 {
-  unsigned int mx, my;
+// Assumes setFootprint already set
+  unsigned int wx, wy;
```
SteveMacenski, line 451: Revert to `mx` and `my` the `m` means "map" and the `w` means world. Since this is converting world coordines (x,y) into map coordinates (mx,my) they should use the `m` version

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -447,13 +459,14 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
     return false;
   }
 
-  unsigned char cost = costmap_->getCost(mx, my);
-
-  if (costmap_ros_->getLayeredCostmap()->isTrackingUnknown()) {
-    return cost >= INSCRIBED_INFLATED_OBSTACLE && cost != NO_INFORMATION;
-  } else {
-    return cost >= INSCRIBED_INFLATED_OBSTACLE;
+  double footprint_cost_ = collision_checker_->footprintCostAtPose(
```
SteveMacenski, line 462: No trailing underscore the footprint cost, not a class member variable anymore

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -178,6 +178,12 @@ void RegulatedPurePursuitController::configure(
   global_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("received_global_plan", 1);
   carrot_pub_ = node->create_publisher<geometry_msgs::msg::PointStamped>("lookahead_point", 1);
   carrot_arc_pub_ = node->create_publisher<nav_msgs::msg::Path>("lookahead_collision_arc", 1);
+
+  // initialize collision checker and set costmap
+  collision_checker_ = std::make_unique<nav2_costmap_2d::
+      FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>(costmap_);
+  collision_checker_->setCostmap(costmap_);
+  traverse_unknown = true;
```
SteveMacenski, line 186: Since this is a class variable, needs a trailing underscore. Also populate it from `costmap_ros_->getLayeredCostmap()->isTrackingUnknown())` rather than setting it to always true

SteveMacenski (changes requested):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: I know its not used, but in case its used later, we don't want it to be unclear and cause a bug

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Nope, thanks for asking, keep that one. that's for the heuristics, that's OK as is due to inflation

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -447,13 +459,14 @@ bool RegulatedPurePursuitController::inCollision(const double & x, const double
...
   }
 
-  unsigned char cost = costmap_->getCost(mx, my);
-
-  if (costmap_ros_->getLayeredCostmap()->isTrackingUnknown()) {
-    return cost >= INSCRIBED_INFLATED_OBSTACLE && cost != NO_INFORMATION;
-  } else {
-    return cost >= INSCRIBED_INFLATED_OBSTACLE;
+  double footprint_cost_ = collision_checker_->footprintCostAtPose(
+    x, y, theta, costmap_ros_->getRobotFootprint());
```
SteveMacenski, line 463: I will need to look into it in the next round of reviewing before merging, but figured it would be good to express this concern now in case it has merit to change sooner: Check what frame of reference the footprint cost at pose expects. What you're giving it is world coordinates, make sure this function expects world coordinates

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
sathak93: where to add it .  any change made on it makes uncrustify test fail

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
sathak93: setting the variable to this function in the configure method fails basic api tests. so i removed the variable and get the value  directly in incollision function

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: L389, where this comment is located. If you look at the diff, there was a space after the `}` and before the comment.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -434,8 +441,12 @@ bool RegulatedPurePursuitController::isCollisionImminent(
   return false;
 }
 
-bool RegulatedPurePursuitController::inCollision(const double & x, const double & y)
+bool RegulatedPurePursuitController::inCollision(
+  const double & x,
+  const double & y,
+  const double & theta)
 {
+// Assumes setFootprint already set
```
SteveMacenski, line 449: remove this comment

SteveMacenski (changes requested):

SteveMacenski:
@sathak93 last couple of changes are just styling, LGTM otherwise!

SteveMacenski (approved):

SteveMacenski:
As you can see now from the diff, the only lines changed are basically just the essentials to make the modifications you require! For reference, that's typically what I look for in PRs, I don't like erroneous changes unless it specifically fixes a bug, otherwise makes reviewing a bit difficult. 

Can you verify to me that you've tested and validated that this works properly? This is a big change to the software so it would be good to do some testing to validate that it does what you expect and projects the footprint "n" seconds, stops if in collision, and detects that collision properly.

I checked the collision detection frames logic and it looks right to me, but it would be good to verify via testing and validate yourself that projection stops when it should and doesn't when it should not.

SteveMacenski:
Needs
- [ ] new logic for stopping collision detection at the goal in predictive loop
- [ ] docs on param conservative changes on the proximity heuristic due to new collision detection method

Kaetera:
Hello, if I may bring a suggestion:
I've been experimenting with a variation of this PR based on the first commits (thanks for it!)
I think we shouldn't [fully skip the collision checking when at goal](https://github.com/ros-planning/navigation2/pull/2556/files#diff-029657b749b8dce55e34c4e407738963328af3e004095a0167d3925ad4f506cbR443-R450): indeed when on the goal point and is spinning to align with the goal orientation, rectangular robots can hit an obstacle on their side. This happen often in my office (quite constrained environment).

On the contrary, I even guaranteed a minimal resolution for the check in rotation:
```
 // also ensure minimum resolution for rotation, because it seems that current collision checker only checks 
  // collision with the edges of the footprint, not with points inside it. So projection step too far in the 
  // future may miss points crossing the edge.
  double proj_time_step_angular; 
  double costmap_angular_res = costmap_->getResolution() / max_robot_radius_;  // so that robot radius * angular res = costmap linear res
  if (fabs(angular_vel) < 0.01) {
    proj_time_step_angular = costmap_angular_res / 0.01;
  }
  else {
    proj_time_step_angular = costmap_angular_res / fabs(angular_vel); 
  }

  double projection_time_step = std::min(proj_time_step_linear, proj_time_step_angular);
```

What was the need to disable collision check at goal ?
THe case I encountered is. eg. if the goal is close to a wall. When still far from the goal the robot approach at full speed and RPP detects a future collision, but in practice RPP will slow down the robot when closer to goal if `use_approach_linear_velocity_scaling = true`
Maybe we would need something more "predictive" to account for that RPP will slow down the robot when approaching goal (If I look at it I'll made a PR)

sathak93:
Thanks for suggestion.
This pr changes 1. Cost at pose Calculation earlier a single point at center of robot is used now 16 point at circumference of robot is used to calculate it.
   This greatly increases collision avoidance behaviour of the robot
2. @SteveMacenski  want to reduce the number of warnings raised when the goal is near to obstacle. We tried number of ways. We are currently at omit predictive collision checking at costmap resolution 0.05 before the goal.

This pr took long time. I intended to finish it by next week.
 As far i know currently there is no predictive collision checking for  in place rotation of robot. 
For non circular that may be problem. i think you should create a new issue or PR regarding this. We can discus /contribute there.

Kaetera:
Ok, I'll propose PR for collision check in rotation and/or for handling the collision check when approaching the goal if I find something, once this PR will be fully done :)

SteveMacenski:
Sorry, Mergify is going insane right now and I don't know why. You can safely just ignore its annoyances.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -288,10 +293,18 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     (carrot_pose.pose.position.x * carrot_pose.pose.position.x) +
     (carrot_pose.pose.position.y * carrot_pose.pose.position.y);
 
+  // similarly find dist^2 to half of look_ahead_point to smooth the curvature
+  auto mid_pose = getLookAheadPoint(lookahead_dist / 2.0, transformed_plan);
```
SteveMacenski, line 297: As I mentioned in Slack, please revert all of the mid_pose stuff, that's probably going to break the stability analysis of the algorithm and you can accomplish a very similar (or exactly the same?) thing just by lowering the lookahead distance to begin with. This is trying to fix an unrelated issue to the collision checking on approach to the goal and using the full SE2 footprint. I think you've mentioned this happens regardless of these changes (eg you see this behavior even when using just point-costs), so its an unrelated topic that is probably just from poor defaults for this particular environment.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -408,28 +424,38 @@ bool RegulatedPurePursuitController::isCollisionImminent(
...
     // apply velocity at curr_pose over distance
     curr_pose.x += projection_time * (linear_vel * cos(curr_pose.theta));
     curr_pose.y += projection_time * (linear_vel * sin(curr_pose.theta));
     curr_pose.theta += projection_time * angular_vel;
 
+    dx = global_plan_.poses.back().pose.position.x - curr_pose.x;
+    dy = global_plan_.poses.back().pose.position.y - curr_pose.y;
+    dist_to_goal = std::hypot(dx, dy);
+
+    if (dist_to_goal < costmap_->getResolution()) {
```
SteveMacenski, line 447: Before in our reviews, you had a condition like `dist_to_goal < desired_linear_vel_ * max_allowed_time_to_collision_` based on velocity and time rather than distance. I think your initial intuition was the right way to go here, if I may explain.

If the current pose projected forward by the current velocity doesn't actually intersect with the goal pose, then this condition would never be hit. I think that's why you saw changes when you made this 2x and 3x the costmap resolution so that if there's some error of the  general region, it hits it OK. 

However, if you measure the distance to the goal from the `curr_pose` and then compare it to the current `projection_time * velocity`, what you're obtaining is the distance traversed on the current heading as compared to the distance to the goal. That would make it so that even if the current velocity didn't pass through the goal pose, we'd still terminate after the theoretical minimum distance to achieve the goal.

SteveMacenski (changes requested):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: If you lower the lookahead distances, doesn't that just fix the problem as well? Keep in mind these are just defaults I set in testing in maps while developing the algorithm, there's nothing special about them.

SteveMacenski:
> indeed when on the goal point and spinning to align with the goal orientation

@Kaetera that is a good point, it really should be more about distance and angular distance! You mention an alternative approach, can you share that with us?

> What was the need to disable collision check at goal ?

We're getting collision warnings on approach to goals near obstacles because projection time * current velocity will be a longer distance than the goal is away from the robot, causing collision checking to occur after the robot would have otherwise stopped at the goal. If close to obstacles, then it might hit an obstacle in the after-goal checking. 

I don't think that's an issue if `use_approach_linear_velocity_scaling = true` since we slow proportionally, but if its `false`, we don't. How have you been testing @sathak93? This might actually be a good argument to make this no longer a parameterization but a strict requirement and part of the algorithm, what do you think? I would support removing that as a parameter and making it a requirement if that would solve an issue!

Kaetera:
My analysis was:
0. Digging into the collision checker, I understood it only checks if the contour of the footprint collides
1. In RPP, the time resolution of the projection is computed so that at each step the footprint "slides" only on adjacent costmaps cells, for linear motion. But it did not account for angular motion, so if the robot is spinning, we could "miss" some adjacent checks.

So this is what I added (to the initial changes of this PR when I ported manually the first commit that just bring up SE2 check):
Determine the minimal projection time
- According to both linear and angular speed
- And so that the most outer point of the footprint, that is the point moving linearly fastest for a given angular speed (v = radius *  angular speed), moves only on an adjacent costmap cell at each projection step

_/regulated_pure_pursuit_controller.hpp_
_Added protected member max_robot_radius_ _
```
  double max_robot_radius_;
```

_In regulated_pure_pursuit_controller.cpp::RegulatedPurePursuitController::configure()
Just get the max radius using footprint.cpp's method:_
```
  double dummy_min_robot_radius;  
  calculateMinAndMaxDistances(costmap_ros_->getRobotFootprint(), dummy_min_robot_radius, max_robot_radius_);
```

_An then in regulated_pure_pursuit_controller.cpp::RegulatedPurePursuitController::isCollisionImminent()
Compute time resolution for angular an linear speed, then project with the smaller one:_
```
bool RegulatedPurePursuitController::isCollisionImminent(
  const geometry_msgs::msg::PoseStamped & robot_pose,
  const double & linear_vel, const double & angular_vel)
{
  ...
  pose_msg.header.frame_id = arc_pts_msg.header.frame_id;
  pose_msg.header.stamp = arc_pts_msg.header.stamp;

  // prevent division by zero
  double proj_time_step_linear;
  if (fabs(linear_vel) < 0.01) {
    proj_time_step_linear = costmap_->getResolution() / 0.01;
  }
  else {
    proj_time_step_linear = costmap_->getResolution() / linear_vel;
  }
  
  // also ensure minimum resolution for rotation, because current collision checker only checks 
  // collision with the edges of the footprint, not with points inside it. So projection step too far in the 
  // future may miss points crossing the edge.
  double proj_time_step_angular; 
  double costmap_angular_res = costmap_->getResolution() / max_robot_radius_;  // so that robot radius * angular res =  costmap linear res
  if (fabs(angular_vel) < 0.01) {
    proj_time_step_angular = costmap_angular_res / 0.01;
  }
  else {
    proj_time_step_angular = costmap_angular_res / fabs(angular_vel); 
  }

  double projection_time_step = std::min(proj_time_step_linear, proj_time_step_angular);
 
  geometry_msgs::msg::Pose2D curr_pose;
  curr_pose.x = robot_pose.pose.position.x;
  curr_pose.y = robot_pose.pose.position.y;
  curr_pose.theta = tf2::getYaw(robot_pose.pose.orientation);

  int i = 1;
  while (true) {
    // only forward simulate within time requested
    if (i * projection_time_step > max_allowed_time_to_collision_) {
      break;
    }
    
    i++;

    // apply velocity at curr_pose over distance
    curr_pose.x += projection_time_step * (linear_vel * cos(curr_pose.theta));
    curr_pose.y += projection_time_step * (linear_vel * sin(curr_pose.theta));
    curr_pose.theta += projection_time_step * angular_vel;

   ...
```

NB: I post the changes manually here since because of some manual ports and other patches I'm few commits late on w.r.t to the last version (I don't usually work with git)
This was added to @sathak93 's commit of September 15th, I did not port the new modifications since it did the job.

SteveMacenski:
> But it did not account for angular motion, so if the robot is spinning, we could "miss" some adjacent checks. ... According to both linear and angular speed

I don't want to hold up this particular PR too much longer since Sathak has been very patient with my feedback, but I would immediately accept a follow on PR to resolve that, because that is indeed an issue. Would that be something you'd be willing to do? Merging this PR would not introduce any additional error that wasn't already in place, so I'm OK with that.  

However, I'm not following your logic about how you obtain `costmap_angular_res`, how is that principled?

What I was thinking was to find the minimum angle `theta_min` for which the robot with max radius `r_max` would have to leave a cell of resolution `resolution`. So this sets up an isosceles triangle where the 2 equal lengthed sides are `r_max` and the `resolution` is the side length opposite of the angle `theta_min`. If we split the triangle in half to get a right triangle, the side lengths are `r_max` `resolution/2` and `h`. Using `sin(opposite / hypot)` we find that `theta_min/2 = sin(resolution/2 / r_max)`, which means that `theta_min = 2*sin(resolution/2 / r_max)`. With that, we know the angular distance, using the angular velocity, we can find the timestep needed to obtain that. 

But I think that logic should only be used for **pure rotations** for the rotate to headings. Linear Velocity should be used when we're actually moving. So the logic for the collision checker could be updated to have a bool for whether to use linear / angular velocities depending on when its being called (e.g. in a rotate to heading situation or a path tracking situation)

sathak93:
yes. isCollisionImminent function should be extended with logic to separate pure rotation and motion

SteveMacenski (approved):

SteveMacenski:
Awesome! @sathak93 @Kaetera can we collaborate on a solution for the rotate at heading collision detection and reducing premature collision detections at goals near obstacles?

SteveMacenski:
Extended conversation in https://github.com/ros-planning/navigation2/issues/2670 for the remaining issues

PR #2559, opened 2021-09-13
Updates to Nav2 Theta Star Planner docs

Anshu-man567:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | NA |

---

This PR basically does the following:
- removing 'w_heuristic_cost_' from Nav2 Theta Star Planner as a user configurable parameter
  - updates to nav2_theta_star_planner/README.md to remove its reference wherever applicable
  - updates to nav2_theta_star_planner/src/theta_star_planner.cpp to set w_heuristic_cost as std::min(1.0, w_euc_cost_)
- fixed an incorrect check of whether the start and goal pose are the same in nav2_theta_star_planner/src/theta_star_planner.cpp, this was leading to runtime errors

removing the mention of w_heuristic_cost from Nav2 Theta Star Planner's docs, as it no longer user configurable

---

update the docs available on the website for the Nav2 Theta Star Planner

#### For Maintainers: 

### DIFF
--- a/nav2_theta_star_planner/README.md
+++ b/nav2_theta_star_planner/README.md
@@ -51,7 +51,6 @@ The parameters of the planner are :
 - ` .how_many_corners ` : to choose between 4-connected and 8-connected graph expansions, the accepted values are 4 and 8
 - ` .w_euc_cost ` : weight applied on the length of the path. 
 - ` .w_traversal_cost ` : it tunes how harshly the nodes of high cost are penalised. From the above g(neigh) equation you can see that the cost-aware component of the cost function forms a parabolic curve, thus this parameter would, on increasing its value, make that curve steeper allowing for a greater differentiation (as the delta of costs would increase, when the graph becomes steep) among the nodes of different costs.
-- `w_heuristic_cost ` : it has been provided to have an admissible heuristic; it is **not** a user configurable parameter
 Below are the default values of the parameters :
 ```
 planner_server:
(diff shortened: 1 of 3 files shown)


nav2_theta_star_planner/README.md
```diff
@@ -51,8 +51,7 @@ The parameters of the planner are :
 - ` .how_many_corners ` : to choose between 4-connected and 8-connected graph expansions, the accepted values are 4 and 8
 - ` .w_euc_cost ` : weight applied on the length of the path. 
 - ` .w_traversal_cost ` : it tunes how harshly the nodes of high cost are penalised. From the above g(neigh) equation you can see that the cost-aware component of the cost function forms a parabolic curve, thus this parameter would, on increasing its value, make that curve steeper allowing for a greater differentiation (as the delta of costs would increase, when the graph becomes steep) among the nodes of different costs.
-- ` .w_heuristic_cost ` : it has been provided to have an admissible heuristic
-
+- `w_heuristic_cost ` : it has been provided to have an admissible heuristic; it is **not** a user configurable parameter
```
SteveMacenski, line 54: Should this be removed here if not a parameter?

SteveMacenski (changes requested):

nav2_theta_star_planner/README.md
SteveMacenski: Update needs to be made to remove it from navigation.ros.org as well

Anshu-man567: I just put it in, cause I do use it in my explanation of the cost function

SteveMacenski: I think you can just explain it there instead (that there's this thing around that is set as 1.0 or set to the same value, if less than 1.0)

Anshu-man567: Ok so I should be changing the comment to state that it is set as min(1.0, w_euc_cost_), right?

SteveMacenski: yeah

SteveMacenski: @Anshu-man567 can you do this soon? This is actually blocking the galactic sync in https://github.com/ros-planning/navigation2/pull/2561

SteveMacenski (approved):

SteveMacenski:
YOu have 2 tests failing

```
/opt/overlay_ws/src/navigation2/nav2_system_tests/src/planning/test_planner_plugins.cpp:63
The difference between tf2::getYaw(path.poses.back().pose.orientation) and -3.14159265358979323846 is 3.1415926535897931, which exceeds 0.01, where
tf2::getYaw(path.poses.back().pose.orientation) evaluates to 0,
-3.14159265358979323846 evaluates to -3.1415926535897931, and
0.01 evaluates to 0.01.
/opt/overlay_ws/src/navigation2/nav2_system_tests/src/planning/test_planner_plugins.cpp:63
The difference between tf2::getYaw(path.poses.back().pose.orientation) and -3.14159265358979323846 is 3.1415926535897931, which exceeds 0.01, where
tf2::getYaw(path.poses.back().pose.orientation) evaluates to 0,
-3.14159265358979323846 evaluates to -3.1415926535897931, and
0.01 evaluates to 0.01.
```

Anshu-man567:
Well the issue seems to arise from the fact that I dont give orientations which the robot needs to reach and hence the difference in angles. I dont really want to orientational bounds at the waypoints (waypt) but if you want it to be the case, I might as well consider the orientation at the waypt[i] to be the atan(waypt[i+1].y - waypt[i].y, waypt[i+1].x - waypt[i].x)
Should I do it though?

Anshu-man567:
I do not push the orientation of the start and goal poses, that is what created the issue, it should be fixed now

SteveMacenski (approved):

SteveMacenski:
a suspicious number of tests failed, just trying again

Anshu-man567:
Ohh sorry, I missed this comment, is it working now?

SteveMacenski:
yeah I reran and got it OK

PR #2570, opened 2021-09-23
removed redundant params in smac example

relffok:
Tiny documentation change in the smac planner README. Looks like those two parameters were overlooked when adding the Lattice Node example. The [docu example](https://navigation.ros.org/configuration/packages/configuring-smac-planner.html) does not hold the redundant params.


SteveMacenski (approved):

SteveMacenski:
Thanks!

Issue #2226, opened 2021-03-09
Assistive Teleop


SteveMacenski:
Create an assisted teleop planner/controller/BT such that someone can teleop a robot and make it do things but use costmap information to make sure it doesn't run into anything or go somewhere it shouldn't. Maybe have an elevated permission BT branch so that someone can send commands to collide if appropriate.  

Potentially make it semi-autonomous as well 

Node take in output of joystick, teleop pad, etc and test that command til `N` seconds in the future for validity. That way always `N` seconds from collision at the given speed and vector commanded. Do so in the BT or as a planning plugin?

Or how nav1 did it https://github.com/ros-planning/navigation_experimental/blob/melodic-devel/assisted_teleop

Do a bit of research, find the options, select the most sensible for the behavior tree and nav2 structure. 

Might be useful to have a teleop global recovery for a user to joystick a robot out of a bad situation and hit "ok" to go back to autonomous navigation via the BT.

SteveMacenski:
3 major modalities to support
- standalone teleop node
- Bt node / recovery
- Main task of the robot to do telepresence 

As such, should probably contain a core library that is then used to create an independent node, a server for BT recovery, and a BehaviorTree XML specifically configured to do solely assisted teleop. 

This should be based on the local costmap and local costmap footprint definitions (possibly a different footprint possible?)

SteveMacenski:
Merging imminent, waiting  on resolution to https://github.com/ros-planning/navigation2/issues/3014 before merging but it is approved and ready to rock and roll (just don't want to release a safety feature when it is known to not work due to another unrelated bug)

PR #2575, opened 2021-09-29
Added Assisted Teleop feature in Navigation2

MOLOCH-dev:
---

## Basic Info
Signed-off-by: Anushree Sabnis sabnisanushree@gmail.com
| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #2226 ) |

---

* I added Assistive Teleop as a feature in Nav2. It currently takes a **Pure Collision Rejection** approach to Assistive Teleop, as mentioned in the [linked discourse topic](https://discourse.ros.org/t/nav2-assisted-teleop-call-for-opinions/21380).
* In this, a teleop input velocity command is tested for collisions by projecting its footprint till `N` seconds in the future
* If a collision is detected within `N` seconds, velocity is scaled down so the robot again remains `N` seconds away from a collision
* Assisted Teleop is a Recovery Plugin that hosts its own action server
* Added four new parameters : `projection_time`, `linear_velocity_threshold`, `input_vel_topic` and `cmd_vel_topic`
* `projection_time` : Maximum allowed time to collision (`N`)
* `linear_velocity_threshold` : Upon collision detection within `projection_time`, if the scaled-down velocity of the robot is below this threshold, the robot is stopped.
* `cmd_vel_topic` : Assisted Teleop publishes velocity to this topic.
* `input_vel_topic` : Assisted Teleop subscribes to this topic to receive velocity commands

* Added new parameters, so need to add that to default configs and documentation page
* Need to add documentation for Assisted Teleop, including an example behavior tree configuration
---

* Update README to include Assisted Teleop as a Recovery
* I need to create a Behaviour Tree Node for Assisted Teleop
* I need to potentially implement more ways for carrying out Assisted Teleop, exploring a mixture of Pure Collision Avoidance and Rejection approaches.

#### For Maintainers: 


SteveMacenski:
- [ ] scale angular velocities proportionally as well so we follow the same vector of travel
- [ ] handle omni robots

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+AssistedTeleop::updatePose()
+{
+  if (!nav2_util::getCurrentPose(
+      current_pose, *tf_, global_frame_, robot_base_frame_,
+      transform_tolerance_))
+  {
+    RCLCPP_ERROR(logger_, "Current robot pose is not available.");
+    return false;
+  }
+  projected_pose.x = current_pose.pose.position.x;
```
SteveMacenski, line 137: This also should not be a global variable, this is a bit confusing

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+bool
+AssistedTeleop::checkCollision()
+{
+  double hypot = std::hypot(speed_x, speed_y, angular_vel_);
+  const double dt =
+    (hypot != 0) ? (costmap_ros_->getResolution() / std::fabs(hypot)) : projection_time_;
+  int loopcount = 1;
+
+  while (true) {
+    if (updatePose()) {
```
SteveMacenski, line 167: You should definitely not be updating your reference pose in this loop. You know where you started, you should be using that alone and then projecting forward from it

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  if (go) {
+    if (!checkCollision()) {
+      RCLCPP_INFO(logger_, "Reducing velocity by a factor of %.2f", scaling_factor);
+    }
+    moveRobot();
+  }
+}
+
+bool
+AssistedTeleop::updatePose()
```
SteveMacenski, line 128: You can probably just delete this function and move the `getCurrentPose` call into `checkCollision()`

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+}
+
+void
+AssistedTeleop::projectPose(
+  double speed_x, double speed_y,
+  double angular_vel_, double projection_time)
+{
+  // Project Pose by time increment
+  projected_pose.x += projection_time * (
+    speed_x * cos(projected_pose.theta));
```
SteveMacenski, line 151: This is incorrect, you're not contributing the component of Y velocity to X based on the angle, and vise versa.

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  projected_pose.y += projection_time * (
+    speed_y * sin(projected_pose.theta));
+  projected_pose.theta += projection_time *
+    angular_vel_;
+}
+
+bool
+AssistedTeleop::checkCollision()
+{
+  double hypot = std::hypot(speed_x, speed_y, angular_vel_);
```
SteveMacenski, line 161: Rename this variable -- also there's no need for angular velocity in this expression. You're computing the time offsets for costmap resolution increment checks. The unit analysis does not work out for rad/s to be related to meters.

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  const double dt =
+    (hypot != 0) ? (costmap_ros_->getResolution() / std::fabs(hypot)) : projection_time_;
```
SteveMacenski, line 163: if hypot == 0, then you'd only just check 1 increment. Why even bother with this expression then?

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  int loopcount = 1;
+
+  while (true) {
+    if (updatePose()) {
+      double time_to_collision = loopcount * dt;
```
SteveMacenski, line 168: Why not just have this look keep track of the acumulated time (time_to_collision) and then if there's no collision, just pass through the command. Else, use that time_to_collision for computing the scale? There's no need to update the scale every iteration until there's a problem.

Also, that should not be a class member variable, also confusing to read.

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+AssistedTeleop::vel_callback(const geometry_msgs::msg::Twist::SharedPtr msg)
+{
+  speed_x = msg->linear.x;
+  speed_y = msg->linear.y;
+  angular_vel_ = msg->angular.z;
+  if (go) {
+    if (!checkCollision()) {
+      RCLCPP_INFO(logger_, "Reducing velocity by a factor of %.2f", scaling_factor);
+    }
+    moveRobot();
```
SteveMacenski, line 123: Pass values to this, don't do it implicitly with member variables being changed, that's bad design

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  } else {
+    vel_pub_->publish(std::move(cmd_vel));
+  }
+}
+
+void AssistedTeleop::execute()
+{
+  RCLCPP_INFO(logger_, "Attempting %s", recovery_name_.c_str());
+
+  if (!enabled_) {
```
SteveMacenski, line 208: remove enabled

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  // Log a message every second
+  {
+    auto node = node_.lock();
+    if (!node) {
+      throw std::runtime_error{"Failed to lock node"};
+    }
+
+    auto timer = node->create_wall_timer(
```
SteveMacenski, line 222: does this actually work? The timer would go out of scope immediately, I highly doubt this would log this statement continuously during use

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+      1s,
+      [&]()
+      {RCLCPP_INFO(logger_, "%s running...", recovery_name_.c_str());});
+  }
+
+  auto start_time = steady_clock_.now();
```
SteveMacenski, line 228: You locked the node in this function, why not just do `node->now()` and remove the clock variable?

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+        logger_, "Received a preemption request for %s,"
+        " however feature is currently not implemented. Aborting and stopping.",
+        recovery_name_.c_str());
+      stopRobot();
+      result->total_elapsed_time = steady_clock_.now() - start_time;
+      action_server_->terminate_current(result);
+      return;
+    }
+
+    auto current_point = std::chrono::steady_clock::now();
```
SteveMacenski, line 260: use the node or timer, don't mix time sources

nav2_msgs/action/AssistedTeleop.action
```diff
@@ -0,0 +1,8 @@
+#goal definition
+builtin_interfaces/Duration time
+---
+#result definition
+builtin_interfaces/Duration total_elapsed_time
+---
+#feedback
+builtin_interfaces/Duration time_left
```
SteveMacenski, line 8: remove this, its not really necessary, we know the time left based on the request.

nav2_recoveries/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,300 @@
...
+  projected_pose.x += projection_time * (
+    speed_x * cos(projected_pose.theta));
+  projected_pose.y += projection_time * (
+    speed_y * sin(projected_pose.theta));
+  projected_pose.theta += projection_time *
+    angular_vel_;
+}
+
+bool
+AssistedTeleop::checkCollision()
```
SteveMacenski, line 159: Rename this function, this is actually doing the main body of assisted teleop, its not just checking for collisions but also scaling velocities to avoid collisions as well.

This should return a velocity in the end to use for sending to the robot, not a boolean.

SteveMacenski (changes requested):
Good first draft, I'll have some more review comments after you fix these, but these are necessary to get into the bigger architectural patterns

nav2_recoveries/plugins/assisted_teleop.cpp
MOLOCH-dev: so instead of speed_x and speed_y, the resultant of the 2 is needed (the hypot) for projection, right?

SteveMacenski: This is in the odometric frame. The velocities are in base frame. You need to take the Y component of the velocity and apply it to the position in X by the angle relative to the frame and vise versa

nav2_recoveries/plugins/assisted_teleop.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+  // Action Server
+  std::shared_ptr<ActionServer> action_server_;
+
+  // Publishers and Subscribers
+  std::shared_ptr<nav2_costmap_2d::CostmapTopicCollisionChecker> collision_checker_;
+  std::shared_ptr<tf2_ros::Buffer> tf_;
+  rclcpp_lifecycle::LifecyclePublisher<geometry_msgs::msg::Twist>::SharedPtr vel_pub_;
+  rclcpp::Subscription<geometry_msgs::msg::Twist>::SharedPtr vel_sub_;
+  std::unique_ptr<nav2_costmap_2d::CostmapSubscriber> costmap_sub_;
```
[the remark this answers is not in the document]
SteveMacenski: Why do you need the costmap sub at all?

SteveMacenski:
On testing:
- Create the object and call `moveRobot` with an example message, verify that it was received on the right topic with the right information based on a changed output velocity parameter
- Create an example costmap to publish and call `computeVelocity` to test a few velocity commands, some that should be in collision, which it should back out of, and some that are not, that should be passed through, check the return command
- Call stopRobot and check works to return on the right topic and 0 call
- Create a system test to bring up the system in simulation and call the assisted teleop action and test that it works fine / timeout works / no collisions

PR #2576, opened 2021-09-30
[Draft] Set parameters result example

doisyg:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Wyca's Elodie|

---

As discussed here https://github.com/ros-planning/navigation2/issues/956

### DIFF
--- a/nav2_smac_planner/src/smac_planner_2d.cpp
+++ b/nav2_smac_planner/src/smac_planner_2d.cpp
@@ -147,7 +147,7 @@ void SmacPlanner2D::configure(
 
   _raw_plan_publisher = node->create_publisher<nav_msgs::msg::Path>("unsmoothed_plan", 1);
 
-  // Add callback for dynamic parametrs
+  // Add callback for dynamic parameters
   dyn_params_handler = node->add_on_set_parameters_callback(
     std::bind(&SmacPlanner2D::dynamicParametersCallback, this, _1));
 
--- a/nav2_smac_planner/src/smac_planner_2d.cpp
+++ b/nav2_smac_planner/src/smac_planner_2d.cpp
@@ -147,15 +147,9 @@ void SmacPlanner2D::configure(
 
   _raw_plan_publisher = node->create_publisher<nav_msgs::msg::Path>("unsmoothed_plan", 1);
 
-  // Setup callback for changes to parameters.
-  _parameters_client = std::make_shared<rclcpp::AsyncParametersClient>(
-    node->get_node_base_interface(),
-    node->get_node_topics_interface(),
-    node->get_node_graph_interface(),
-    node->get_node_services_interface());
-
-  _parameter_event_sub = _parameters_client->on_parameter_event(
-    std::bind(&SmacPlanner2D::on_parameter_event_callback, this, _1));
+  // Add callback for dynamic parametrs
+  dyn_params_handler = node->add_on_set_parameters_callback(
+    std::bind(&SmacPlanner2D::dynamicParametersCallback, this, _1));
 
   RCLCPP_INFO(
     _logger, "Configured plugin %s of type SmacPlanner2D with "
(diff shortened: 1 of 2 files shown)


nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -340,45 +334,46 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
   return plan;
 }
 
-void SmacPlanner2D::on_parameter_event_callback(
-  const rcl_interfaces::msg::ParameterEvent::SharedPtr event)
+rcl_interfaces::msg::SetParametersResult
+SmacPlanner2D::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
 {
+  auto result = rcl_interfaces::msg::SetParametersResult();
```
SteveMacenski, line 340: err, umm maybe `rcl_interfaces::msg::SetParametersResult result` instead? I think you've been writing some python code recently :wink:

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -340,45 +334,46 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
...
+  RCLCPP_INFO_STREAM(_logger, "Param callback, size: " << parameters.size());
```
SteveMacenski, line 341: Please replace with `RCLCPP_INFO` call instead using `%i`. Its a stylistic thing, I like to keep logging statements consistent within a project such that we can easily scan through code and not have to think too much. This is really simple, but that's how the creep starts :-)

SteveMacenski (changes requested):
LGTM, just small changes. Should we be doing anything to check if the event is a declaration (and to block an update in that case?)

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: hum, at least from cpp 11 using `auto` is pretty convenient, and I was just reproducing rclcpp style:
https://github.com/ros2/rclcpp/blob/d5ec2580805c3abd8b6fecaf3d09040e00e9f15e/rclcpp/src/rclcpp/parameter_service.cpp#L86
But no problem to use `rcl_interfaces::msg::SetParametersResult result` instead if you want

nav2_smac_planner/src/smac_planner_2d.cpp
doisyg: Debug artefact, removing it

doisyg:
> Should we be doing anything to check if the event is a declaration (and to block an update in that case?)

I believe registering the callback after all the declaration is enough, no ? Plus I don't see how to check if it is a declaration in this callback

nav2_smac_planner/src/smac_planner_2d.cpp
SteveMacenski: Do `rcl_interfaces::msg::SetParametersResult result;`, its more compact. Clearly whoever wrote that was thinking in python. Its the same thing without `auto` or `()`, making it more compact. There's no reason to use auto for the default constructor.

doisyg: :man_facepalming: nice catch, did not realize it. Will change at next push

doisyg:
> Into the on activate state transition? That way its definitely after parameter declarations (and this is via a network interface, it shouldn't be allowed in the inactive states). Probably belongs there anyway.

I did not think of this before, but indeed, that looks like the perfect place to put parameters callback registration. I guess the handler should be reset in the deactivate transition too. Changes pushed

SteveMacenski:
I didn't really think about it either until we started this discussion. Yeah, it should be reset in deactivate

SteveMacenski (approved):

Issue #2064, opened 2020-10-28
Consider adding shim rotate-to-path controller plugin


SteveMacenski:
Per https://github.com/ros-planning/navigation2/issues/2035 this has come up in now 5 different places I've worked / hobby projects / other users.

It would be useful to have a local trajectory planner plugin that consists of 2 steps:
- Checking the angle between the robot's current orientation and the path, when using a holonomic planner on a non-holonomic robot. If larger than threshold, rotate to path orientation
- Once below threshold, continue with the loaded local trajectory plugin (like DWB or TEB)

SteveMacenski:
This has now **again** come up in another context. This clearly deserves action. Maybe another option is to include a rotate to orientation controller plugin itself for use. That way it could be called independently from a BT node (like rotate to path orientation then give to controller; or rotate to goal after XY tol hit) or used in conjunction with other controller plugins. 

Another thought is to have it have both rotate to path orientation and rotate to goal at end as the shim, and the middle of it when not in XY tolerance or at start of path will be given to the "real" plugin shimmed (DWB, etc). Then controller plugins wouldn't need to handle the rotate to goal itself - but if it were to - then the second-level shim would just return true and do nothing.

shrijitsingh99:
I too have implemented this functionality in projects in the past, so I support this.
Having it as a BT node seems like a good way to go. Any potential downsides?

> Another thought is to have it have both rotate to path orientation and rotate to goal at end as the shim, and the middle of it when not in XY tolerance or at start of path will be given to the "real" plugin shimmed (DWB, etc). Then controller plugins wouldn't need to handle the rotate to goal itself - but if it were to - then the second-level shim would just return true and do nothing.

This is not quite clear to me. Could you expand on this?

SteveMacenski:
I don't think a BT node is the objectively right option for everything. I think the best option here is to make a class that does a rotate to relative heading task. Then we  can use that class to implement a `rotate to heading` BT node, `rotate to node` controller plugin, add to pure pursuit planner or optionally to DWB, and the shim node (any or all).

Sure, what I meant is that typically we have 3 stages in this situation:
- Rotate to path heading, if you have a holonomic planner like NavFn, it can create a path that's not in the orientation of the front of the robot. So first step is rotating to path heading (roughly)
- The normal local trajectory planning
- Then rotating to goal, once you've reached your cartesian tolerance for goal. Then rotate to match the right heading. 

So we could have a controller plugin that does the following:
[...]

PR #2579, opened 2021-10-04
Shim controller

melkobrodov:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2064 |

---

* I added shim controller plugin

* I added parameters, need to document them in Configuration Guide section
* Also I added tutorial in General Tutorials

---

#### For Maintainers: 


nav2_bringup/params/nav2_sc_params.yaml
```diff
@@ -0,0 +1,341 @@
+amcl:
```
SteveMacenski, line 1: Remove this param file

nav2_shim_controller/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+find_package(geometry_msgs REQUIRED)
+find_package(nav2_costmap_2d REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(nav_msgs REQUIRED)
+find_package(nav2_util REQUIRED)
+find_package(nav2_core REQUIRED)
+find_package(tf2 REQUIRED)
+
+nav2_package()
+set(CMAKE_CXX_STANDARD 17)
```
SteveMacenski, line 21: This shouldn't be necessary, what C++17 feature are you using?

nav2_shim_controller/include/nav2_shim_controller/shim_controller.hpp
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) 2020 Shrijit Singh
```
SteveMacenski, line 1: This package is missing a readme file explaining what it is, how to use it, why you want to use it, parameters in it, etc.

nav2_shim_controller/include/nav2_shim_controller/shim_controller.hpp
```diff
@@ -0,0 +1,213 @@
...
+    const geometry_msgs::msg::PoseStamped & pose,
+    const geometry_msgs::msg::Twist & velocity);
+
+  std::shared_ptr<tf2_ros::Buffer> tf_;
+  std::string plugin_name_;
+  std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
+  nav2_costmap_2d::Costmap2D * costmap_;
+  rclcpp::Logger logger_ {rclcpp::get_logger("ShimController")};
+
+  double desired_linear_vel_;
```
SteveMacenski, line 189: Many of these parameters should not exist, there's no need for this param, or look ahead dist, or most of them below. This controller only needs a param for max angular velocity / acceleration for rotating to heading / path.

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+void ShimController::cleanup()
+{
+  RCLCPP_INFO(
+    logger_,
+    "Cleaning up controller: %s of type"
+    " shim_controller::ShimController",
+    plugin_name_.c_str());
+
+  if (default_plugin) {default_plugin->cleanup();}
```
SteveMacenski, line 125: Expand onto multiple lines (and for all cases of this below) to match styling of the rest of the program.

SteveMacenski: Also need to `reset()` the pointer, I believe in cleanup.

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+geometry_msgs::msg::TwistStamped ShimController::computeVelocityCommands(
+  const geometry_msgs::msg::PoseStamped & pose,
+  const geometry_msgs::msg::Twist & speed,
+  nav2_core::GoalChecker * goal_checker)
+{
+  goal_checker_ = goal_checker;
+  return calcCmdVel(pose, speed);
+}
+
+geometry_msgs::msg::TwistStamped ShimController::calcCmdVel(
```
SteveMacenski, line 158: Put this function into `computeVelocityCommands`, there's no real reason to separate this out.

SteveMacenski: Then goal_checker doesn't need to be a class member anymore

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+{
+  global_plan_ = path;
+  if (default_plugin) {default_plugin->setPlan(global_plan_);}
+}
+
+void ShimController::setSpeedLimit(
+  const double & speed_limit,
+  const bool & percentage)
+{
+  if (speed_limit == nav2_costmap_2d::NO_SPEED_LIMIT) {
```
SteveMacenski, line 295: This should all just be a pass through to the sub-plugin, there's no need for any of this logic here.

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+  RCLCPP_DEBUG(
+    logger_, "speed = %f, speed_cur = %f, angle = %f, calc angle = %f",
+    angular_vel,
+    curr_speed.twist.angular.z,
+    angle_to_path,
+    deceleration_angle);
+}
+
+geometry_msgs::msg::PoseStamped ShimController::getLookAheadPoint(
+  const double & lookahead_dist,
```
SteveMacenski, line 268: This function is unneeded, remove

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+    } else {
+      // Speed limit is expressed in absolute value
+      desired_linear_vel_ = speed_limit;
+    }
+  }
+  if (default_plugin) {default_plugin->setSpeedLimit(speed_limit, percentage);}
+}
+
+nav_msgs::msg::Path ShimController::transformGlobalPlan(
+  const geometry_msgs::msg::PoseStamped & pose)
```
SteveMacenski, line 311: This function is unneeded, remove

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+  if (transformed_plan.poses.empty()) {
+    throw nav2_core::PlannerException("Resulting plan has 0 poses in it.");
+  }
+
+  return transformed_plan;
+}
+
+bool ShimController::transformPose(
+  const std::string frame,
```
SteveMacenski, line 373: This function is unneeded, remove

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+  geometry_msgs::msg::Twist vel_tolerance;
+  if (goal_checker_ && goal_checker_->getTolerances(pose_tolerance, vel_tolerance)) {
+    goal_dist_tol_ = pose_tolerance.position.x;
+    goal_yaw_tol_ = tf2::getYaw(pose_tolerance.orientation);
+  } else {
+    RCLCPP_WARN(logger_, "Unable to retrieve goal checker's tolerances!");
+  }
+
+  // Transform path to robot base frame
+  auto transformed_plan = transformGlobalPlan(pose);
```
SteveMacenski, line 173: Remove all RPP logic, I don't see why that should be in the file that simply rotates a robot to its current path / goal headings

SteveMacenski (changes requested):
Code looks clean for the most part, just need to pare it down to the necessary minimum components and simplify it. Our goal is just a plugin that rotates a robot to the rough path heading once a new path comes in, and when at the goal pose. Otherwise, give it to the controller plugin to track the path. Overall, this should be a relatively simple passthrough file with some block of rotate to heading logic.

nav2_shim_controller/src/shim_controller.cpp
```diff
@@ -0,0 +1,397 @@
...
+    desired_linear_vel_ = max_linear_vel_;
+  } else {
+    if (percentage) {
+      // Speed limit is expressed in % from maximum speed of robot
+      desired_linear_vel_ = max_linear_vel_ * speed_limit / 100.0;
+    } else {
+      // Speed limit is expressed in absolute value
+      desired_linear_vel_ = speed_limit;
+    }
+  }
```
AlexeyMerzlyakov, line 306: Where is this logic used? It looks, like `desired_linear_vel_` is unused at all: we could just call `default_plugin->setSpeedLimit()`, I think

AlexeyMerzlyakov (commented):
I would like agree with Steve's main comment: it looks like we could do without `transformGlobalPlan()/getLookAheadPoint()` PRR's functions as knowing current robot pose and global path should be enough for decisions making to rotate and won't break any current logic, as currently it relies on predicted `carrot_pose` on one step ahead which could be calculated as tangent to global path.

nav2_shim_controller/src/shim_controller.cpp
SteveMacenski: The only time you need to do any of this logic is when a new path comes in. So in the set plan function, you could set a flag that it has changed (or compare the paths to see that there is a change) to then trigger the "check if I even need to rotate to rough heading" logic. If not, just pass through and reset the flag. If so, rotate to heading. Once rotated within the tolerance, reset the flag as pass through to the internal controller plugin.

SteveMacenski:
Hi,

Unfortunately, this is needed a bit more immediately for a project so I'm planning on tackling this this week in order to get it done and merged in before year's end. I'll leave this PR open in case you get a chance to fix it before I'm done, but I'm targeting Friday for a PR to resolve this feature request.

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/2718

PR #2591, opened 2021-10-09
Cleaned up version of is path valid

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2374 |

---

- Added service on planner server to determine if a path is valid by checking the current path against the cost map. 
- Added conditional BT node to call the service 

Need to add to BT documentation

---

Allow the threshold for determine re-planning to adjusted.  

#### For Maintainers: 


SteveMacenski:
Pull in main or rebase on main, there was an API change to rolling you need to introduce to build

SteveMacenski:
On your questions

> Figure out why new goals cannot be set.

I think you mean for preemption right? It works ok if you send a goal, it achieves the goal, and you send another, yeah? Probably because you're only replanning on the `isPathValid` condition. The other would replan at 1hz or after N meters regardless, so if a new goal came in, it would use that immediately in the next cycle. Here, there is no "out". We use the `GoalUpdated` in the recovery sub-BT in the main navigation tree to resolve that issue. If a new goal is updated, it can force a recomputation. You can add that in your BT logic for replanning to make it force replan on new goal update, regardless if the current plan is invalid or not.

Probably would look something like 
```
<containerA>
  <containerB>
    <GoalUpdated/>
    <IsPathValid/>
  </containerB>
  <ComputePathToPose/>
</containerA>
```

So that `containerB` forces the recomputation of `ComputePathToPose` if either a new goal is received or the path is invalid. No need for the distance controller in this situation, since if the path is invalid or there is a new goal, we want to compute something. It might be nice to have some controller wrapped around the compute path to pose action so that if we're constantly invalidating it (due to dynamic obstacles harassing the robot), there's some minor delay to action. Maybe a rate controller set to something  small, like 5hz, so that there's a 200ms delay so we're not thrashing the CPU with recomputation requests 

> Figure how to truncate the current path as the bot moves toward the goal. I am not if you have any suggestions for this, (what files should I be looking into)

This is a more interesting question. I don't actually have a really bullet-proof answer off of the top of my head. Here are some options:
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp I made this for the navigation through poses behavior tree to truncate old points as we passed them so we didn't continuously recompute paths to passed locations. I admit this is pretty fragile and I  wouldn't recommend it, longer term I'd like to remove this with a "better" method, but maybe part of this PR can be to improve the actual mechanics this uses. Why this is so fragile is that if the path follower controller deviates too far from the original path, it will fail to remove old points. So that radius is set pretty liberally where its used. 
- Within the is path valid action in the planner server, you can take the robot's current pose and find the closest point to the robot (if multiple are the same, take the oldest one) and then remove everything before it for checking collision. Requires a little vector math, but nothing too challenging. 
- Google some papers or other software robotics projects to see how they go about this. There must be better methods than what I can come up with as a knee-jerk action on this topic

jwallace42:
> I think you mean for preemption right? It works ok if you send a goal, it achieves the goal, and you send another, yeah?

Preemption and sending another goal after the first one is reached does not work. 

When I send a new goal after the first is reached I get the following. 
send_new_goal
It seems to instantly reach the goal instantly. I am currently not quite sure what is happening. 



> The other would replan at 1hz or after N meters regardless, so if a new goal came in, it would use that immediately in the next cycle. Here, there is no "out".
Yep. I will implement a BT with the solution you suggested above. 





> This is a more interesting question. I don't actually have a really bullet-proof answer off of the top of my head. Here are some options:

I will try to start looking for papers to come up with a long term solution.

SteveMacenski:
I don't see any reason that it should be auto-succeeding, other than the goal isn't actually getting to the planner, so its using the last goal, and therefore it does auto-succeed because it is where it is. 

I can see why preemption would fail, your BT doesn't allow for it since https://github.com/ros-planning/navigation2/blob/0246e99fdc3e63953884f0eba6cfde7069c2e6f9/nav2_bt_navigator/behavior_trees/navigate_w_replanning_distance_if_invalid_path.xml#L5 would still return true and never get passed to the planner. You need to put both that condition and `isGoalUpdated` into a container that will pass if either is true to allow for preemptions.

I'm wondering if its autosucceeding because `isPathValid` on the 2nd run is using the same `{path}`, which may still be valid, then going to the controller which says "I'm already at the end" and then passing? Or if `isPathValid`'s path is empty and that doesn't auto-return FAILURE within the BT node if the path is empty (that seems like a condition to add to the BT node).

jwallace42:
PXL_20211020_174232464

Hey Steve, 

I attached the an image of the BT I am using above. After a bit more experimenting it seems that the goal updated condition is not working as I expected. When I select a new nav2 goal in rviz I expected to see success for goal updated. On closer inspection I found that BT::NodeStatus::IDLE was constantly being triggered and the node could never return SUCCESS. 

`  if (status() == BT::NodeStatus::IDLE) 
{
    config().blackboard->get<std::vector<geometry_msgs::msg::PoseStamped>>("goals", goals_);
    config().blackboard->get<geometry_msgs::msg::PoseStamped>("goal", goal_);
  }`  

When I force that node to always return SUCCESS everything works as expected.

SteveMacenski:
General note: I think from this, you want reactive https://www.behaviortree.dev/sequencenode/ so that it always restarts fresh each cycle no matter the previous state. This is true of both sequence and fallback. I don't think that'll fix your problem, but it might. 

The first run should not trigger the goal updated thing because it wasn't "updated" after the original request. The original request is what triggers the tree to start, so that's not a change (hence why the IDLE state sets the goal variables to reference again). It's the second time you update the goal (e.g. once the robot is already navigating to the original goal) that it should trigger. Is that not working? 

The goal updated is working in a nested reactive fallback in the BTs https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_through_poses_w_replanning_and_recovery.xml#L17 so I don't think the fact that its nested under a control flow node is the reason (though it could be). I think maybe changing to reactive would do the trick? Also perhaps my clarification above would help?

jwallace42:
> It's the second time you update the goal (e.g. once the robot is already navigating to the original goal) that it should trigger. Is that not working?

Yep. So I switched the Sequence and Fallback control node to their reactive version and continued to see the same behavior as before. The goal updated node is still stuck in idle. 

I looked through the behavior tree source code and found that the only times a node is idle is on construction and when it is halted. I am a bit confused on why the this node keeps getting set back to idle. I am going to keep digging.

SteveMacenski:
How goes it? Anything I can help answer?

jwallace42:
Hey Steve, just spent some time this morning digging in. I found out why this behavior tree won't work. 

Lets go through the following 
IsPathValid status = IDLE
GoalUpdated status = IDLE
1) The first tick though the tree goes down to IsPathValid -> SUCCESS status gets set to SUCCESS
2) The next node in the reactive sequence gets hit GoalUpdated -> FAILURE and get inverted to return SUCCESS
The status of this node gets set to FAILURE
3) Since all the children returned success, the reactive sequence halts all children nodes. This in turn, sets the node status of each child to IDLE. 

As you can see the GoalUpdated node will always been stuck in the idle state and not update correctly. 

I not quite sure how to proceed from here. We have to somehow check the combined status of both the GoalUpdated and IsPathValid without using a sequence because we cannot halt the node.

SteveMacenski:
Is that then implying that all of these in the BT XML aren't actually working? https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml#L29

I see that these all use `ReactiveFallback` not Sequence. I think you could get the exact same behavior in your BT if you invert all of the logic (e.g. swap the fallback and sequence nodes, and put the invert on the is path valid rather than is goal updated). Would that fix the issue?

I see that all of the controllers (e.g. RateController, DistanceController, etc) have the same `status() == BT::NodeStatus::IDLE` condition, but I believe (?) they're working, but maybe we've stumbled onto something showing there's some real failures going on. The only thing I know is working right is Rate because we use that in the default tree under a [PipelineSequence](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/control/pipeline_sequence.cpp) for controlling the planning rate (which you can visually see works). The pipeline sequence also calls haltChildren, which in turn calls [this](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/725eba7e0abbc704284dec393706e5404f1472e3/src/control_node.cpp#L51) to set to idle. So I'm not really sure

jwallace42:
> 

So I just ran the navigate_to_pose_w_replanning_and_recovery.xml and added printouts to the GoalUpdated node. I switched the goal pose multiple times mid way through and saw no output from the GoalUpdated node. So I don't think it is working as expected in this BT. Hopefully I am missing something!

SteveMacenski:
I'm taking a look across the board at what BT nodes use the idle state to initialize like this.
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/condition/distance_traveled_condition.cpp#L46-L54
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/condition/goal_updated_condition.cpp#L31-L35
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/condition/time_expired_condition.cpp#L39-L43
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/decorator/distance_controller.cpp#L53-L64
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/decorator/rate_controller.cpp#L36-L41
- https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/decorator/single_trigger_node.cpp#L33-L35

On this list, I think the `DistanceTraveledCondition`, `TimeExpiredCondition`, `DistanceController`, and `RateController` are OK since they are all relative models from an initial time point or pose and looking for changes over that range. Then if they are reset because the parent is done (sets all to `IDLE`) and triggered again, this should be doing what its supposed to do still but just restarts from 0.

I'm worried about `GoalUpdatedCondition` and `SingleTrigger`. I wanted `SingleTrigger` to be triggering once per BT run, but I suppose it runs once per BT node context (e.g. if under a control flow once per time until it returns a non-RUNNING status). That might be OK? I suppose that is the BT-y way of doing things if all the states of the lower nodes are `IDLE` after a control flow node returns a terminal status. I would like to have a global option that only runs again on the next BT run, but I'm honestly not sure how that would be possible.

It occurs to me now that it might be good to have a `RunID` on the blackboard, so that each time we ask the BT to do a navigation run, we have some ID attached to it for debugging / being able to tell if we're in a different navigation request. I think that could help us. 

The goal updated is certainly a problem. I think that can be solved though. What if we update the node to be like

```
BT::NodeStatus GoalUpdatedCondition::tick()
{
  config().blackboard->get<int>("run_id", run_id); 
  if (first_time || run_id != run_id_) {
    config().blackboard->get<std::vector<geometry_msgs::msg::PoseStamped>>("goals", goals_);
    config().blackboard->get<geometry_msgs::msg::PoseStamped>("goal", goal_);
    first_time = false;
    run_id_ = run_id;
  }

  std::vector<geometry_msgs::msg::PoseStamped> current_goals;
  config().blackboard->get<std::vector<geometry_msgs::msg::PoseStamped>>("goals", current_goals);
  geometry_msgs::msg::PoseStamped current_goal;
  config().blackboard->get<geometry_msgs::msg::PoseStamped>("goal", current_goal);

  if (goal_ != current_goal || goals_ != current_goals) {
    goal_ = current_goal;
    goals_ = current_goals;
  }

}

```

Then if we're in the same BT control flow node context when the goal changes, it should work. If we leave that Bt control flow node context (e.g. the control flow returns `SUCCESS` and resets everything to IDLE) and reenter, then we retain the same state of `goal_` and `goals_`. Between BT execution runs, we can also reset the goal/goals if the ID is updated. We'd have to update the ID every new navigation or preemption request.

jwallace42:
Hey Steve, 

Sorry for the delay, I had a bunch of job interviews these past weeks. 



> It occurs to me now that it might be good to have a `RunID`

I agree. I think a `RunID` would be very useful for debugging and would work for a goal updated condition. I have a similar reservation to you about having `RunID` connected to a node but if it is limited to logging and the goal updated condition  it should be okay. 


> Something else we could do is store state info on the blackboard with unique keys per instance and then shared keys things we want shared across all instance.

Could you elaborate on this idea. What state information are we storing? The Node status state? 

> What if we updated it to have a variable `first_time` that triggers only the first time the BT node is called that replaces the check for IDLE?

How would the `first_time` var get reset? I am imaging that we tick the tree and hit the condition node which would set `first_time` to false. But how does this var get reset back to true?

 


> Am I correct in saying the first probably works OK?

Yes.


> The 2nd and 3rd probably don't work at all since they're only in existence for the period of time it takes to clear the costmap.

I also agree with this. 



> Did you see logging if you preempted with a new goal during recovery behaviors?

I think I may have gotten the sequence wrong when trying to test this tree. I will have to check it out again.

SteveMacenski:
>  Something else we could do is store state info on the blackboard with unique keys per instance and then shared keys things we want shared across all instance.

I mean we have the blackboard where we can store variables that persist. So we could have some variables we store that persist between runs / sessions / contects if we like that the BT node could get access to. I propose that there are 2 different types of variables we could store:
- Unique: variables that are  for a specific instance of the BT node and used only by that specific instance of the BT node (eg each instance of the goal updated condition can store some info for later use)
- Shared: variables that are for a specific type of BT node and used / shared by all instances of that specific type of BT node (eg all goal updated conditions nodes share some state info) 

There's a situation if you have multiple `GoalUpdatedCondition`s that you may want them all to reset if any of them see the update such that each one doesn't trigger an exit due to a goal change already processed by another instance -- if there's not some global state. But at the same time, there may be other instances of `GoalUpdatedCondition` that purely want to think about its own context (e.g. the recovery instance) and doesn't care about retaining global state bc its only goal is to preempt if it sees a change while actively in recovery. 

But maybe this is all here-nor-there. I am just proposing options and potential problems with that option as I was trying to think through the problem. 

Maybe the best solution is here potentially (?) another BT node entirely, not `GoalUpdatedCondition` which works in the local context. There are a few options including a new condition or decorator node. Maybe we should have a `GlobalGoalUpdatedCondition` that retains its own global state on the blackboard, that could solve the issue. A similar thing could be true for a decorator node, but I think those BT semantics are less appealing. What if we have a node that:
- Is essentially the `GoalUpdatedCondition` node
- With a random number generator in the constructor that stores a variable about its state on the blackboard as `<random_num>/state` 
- Uses that on retriggering to set the state
- Updates on the global setting, if the goal changes anywhere, ever to trigger

>  Am I correct in saying the first probably works OK? --> Yes

OK, lets remove the other 2 then if they're useless. 

> I think I may have gotten the sequence wrong when trying to test this tree. I will have to check it out again.

OK that would be helpful to know, since that's pretty important

jwallace42:
Hey Steve, 

> Maybe we should have a `GloballyUpdatedGoalCondition`

So I wrote the global updated goal node today and the path gets re planned correctly whenever a new goal is placed. 

However, another issue came up. It seem that the 'FollowPathAction' does not follow the new path that is planned. Instead it continues to follow the old path. This seems like an issue with how the new path is either placed on the blackboard or read off the blackboard but I really have no idea. I will spend some time tomorrow looking into this issue. 

Have you seem anything like this before? 


> * remove GoalUpdatedCondition from the navigation subtree that clearly don't do anything useful
Done and everything appears to be working well.

jwallace42:
[path_replanning.zip](https://github.com/ros-planning/navigation2/files/7669838/path_replanning.zip)
I added a video above and push up my changes. I am continuing to look into what is happening.

SteveMacenski:
What does the terminal logging say?

jwallace42:
[path_replan-2021-12-07_12.09.34.zip](https://github.com/ros-planning/navigation2/files/7670301/path_replan-2021-12-07_12.09.34.zip)
I attached a video with the logs and added a print out in the on_tick method in the FollowPathAction node.

SteveMacenski:
That's really hard to tell from the video. I don't see any of the print outs, but there's alot of noise in there from rviz. Were there FollowPath prints? If not, I'd add prints to the controller server + planner server to see where the message is being lost.

jwallace42:
So it looks like the `ComputePathToPose` fails. It gets ticked but the `on_success` method never gets triggered so the new path never gets put onto the blackboard. So then `FollowPathAction` will never get re ticked.

SteveMacenski:
Got it - that's odd that it fails! Why's that?

jwallace42:
> Got it - that's odd that it fails! Why's that?

Not sure yet! Digging into it to see what may be happening!

jwallace42:
So it appears to be something on the client end. The action is successful on the planner server, but as it goes to the BT client something happens that causes the BT node to return failure. I am going dig into the client side code tomorrow.

jwallace42:
The issue is `goal_result_available_` never gets set to true. I think this may have something to do with the node only getting ticked once. 

The `goal_result_available_` gets set to true when the result for the goal is received.  
https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L234

I think that the since the `ComputePoseToPath` node only ever gets ticked once there is not enough time for the action server to get the result. 

The printouts confirm my hypotheses by showing the tick function getting hit returning and then the server returns success. So We need the `ComputePoseToPath` to be ticked again.

jwallace42:
The first thought is to add a decorator node that will tick the `ComputePathToPose` until it returns SUCCESS or FAILURE. Something similar to `RetryNode`

SteveMacenski:
> I think this may have something to do with the node only getting ticked once.

If it returned the `RUNNING` status, then in your tree it should be being reticked until it returns a status. It's not being "retried" as much as RUNNING in the BT means that the tree should be ticking that over and over until it returns a status. Then that status is communicated up the tree via the sequence / fallback nodes to then impact their statuses (and so on) for the flow of logic. 

I think your `ReactiveFallback` should be a `Fallback` instead so that it persists state in this situation. That should fix the issue since it won't retry the other things to fail, it will go directly to the last RUNNING state node before doing other things. https://www.behaviortree.dev/fallbacknode/

I think your sequence is OK as reactive, since its only about whether to trigger based on new goals / paths valid

jwallace42:
It works!!! Okay, so I should try to integrate into the navigate to pose BT?

SteveMacenski:
Can you push your code first (and deal with the merge conflict)? Doing some reviews on the simplified version would be great and we can get in the final tree before merging. 

Action list:
- [ ] Push updates - please also include an updated BT visualization Groot or [this script](https://github.com/ros-planning/navigation2/blob/main/tools/bt2img.py) can help
- [ ] Reviews (test coverage, code quality items, comments, etc) 
- [ ] Final BT update

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
```diff
@@ -12,15 +12,13 @@
           <RecoveryNode number_of_retries="1" name="ComputePathToPose">
             <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased"/>
             <ReactiveFallback name="ComputePathToPoseRecoveryFallback">
-              <GoalUpdated/>
```
SteveMacenski, line 15: Can you also remove this from the navigate through poses BT?

nav2_behavior_tree/plugins/condition/goal_updated_condition.cpp
```diff
@@ -24,13 +24,16 @@ GoalUpdatedCondition::GoalUpdatedCondition(
   const BT::NodeConfiguration & conf)
 : BT::ConditionNode(condition_name, conf)
 {
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
 }
 
 BT::NodeStatus GoalUpdatedCondition::tick()
 {
+  RCLCPP_INFO(node_->get_logger(), "Goal updated: Ticked");
```
SteveMacenski, line 32: Probably should remove the logging now :smile:

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+  std::vector<geometry_msgs::msg::PoseStamped> current_goals;
+  config().blackboard->get<std::vector<geometry_msgs::msg::PoseStamped>>("goals", current_goals);
+  geometry_msgs::msg::PoseStamped current_goal;
+  config().blackboard->get<geometry_msgs::msg::PoseStamped>("goal", current_goal);
+
+  if (goal_ != current_goal || goals_ != current_goals) {
+    goal_ = current_goal;
+    goals_ = current_goals;
+    RCLCPP_INFO(node_->get_logger(), "Goal updated: SUCCESS");
```
SteveMacenski, line 49: Probably remove this logging as well

SteveMacenski: (and others in this file)

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -0,0 +1,65 @@
...
+: BT::ConditionNode(condition_name, conf)
+{
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+  client_ = node_->create_client<nav2_msgs::srv::IsPathValid>("is_path_valid");
+}
+
+BT::NodeStatus IsPathValidCondition::tick()
+{
+  if (status() == BT::NodeStatus::IDLE) {
+    RCLCPP_INFO(node_->get_logger(), "Is Path Valid: IDLE");
```
SteveMacenski, line 35: Removed logging if working :-)

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +516,52 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+
+  response->is_valid = true;
+
+  if (request->path.poses.empty() ) {
+    response->is_valid = false;
+  } else {
+    geometry_msgs::msg::PoseStamped current_pose;
+    unsigned int closest_pose_index = 0;
+    if (costmap_ros_->getRobotPose(current_pose)) {
+      tf2::Vector3 curr_position(current_pose.pose.position.x,
```
SteveMacenski, line 534: You can use https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/geometry_utils.hpp#L82 instead so we're not working with `tf2` types that we don't have control over. Use `PoseStamped` instead and use the nav2_util function to get the distance. Then you don't need to convert the types.

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +516,52 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+          request->path.poses[i].pose.position.z);
+
+        if (tf2::tf2Distance(curr_position, path_point) < closest_distance) {
+          closest_pose_index = i;
+          closest_distance = tf2::tf2Distance(curr_position, path_point);
+        }
+      }
+    }
+
+    for (unsigned int i = closest_pose_index; i < request->path.poses.size(); ++i) {
```
SteveMacenski, line 552: Add a comment here that we're using the closest point to the end that way we don't check passed points if they're occupied now, because we passed them already so we don't care.

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +516,52 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+  } else {
+    geometry_msgs::msg::PoseStamped current_pose;
+    unsigned int closest_pose_index = 0;
+    if (costmap_ros_->getRobotPose(current_pose)) {
+      tf2::Vector3 curr_position(current_pose.pose.position.x,
+        current_pose.pose.position.y,
+        current_pose.pose.position.z);
+      // Find index of pose closest to the robot
+      float closest_distance = 0;
+      for (unsigned int i = 0; i < request->path.poses.size(); ++i) {
```
SteveMacenski, line 539: To make this a little more robust, may I suggest that you also track the last distance(s) as we march through the path, it should be getting closer and then further from the robot (or if its the start, just further). So if we see the trend that its getting further, we can stop checking and exit.

That helps with the case that there are loops in the path and the incorrect point is selected because it is floating-point error closer to the robot than the actual point of interest (which is a problem right now for the controllers that use a similar technique as this)

nav2_behavior_tree/plugins/condition/goal_updated_condition.cpp
jwallace42: haha. Done

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
jwallace42: Done

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Done

nav2_planner/src/planner_server.cpp
jwallace42: Done.

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
jwallace42: Done

SteveMacenski:
Push the updates to take a look?

jwallace42:
Just pushed. I still need to add on the groot visualization stuff. After that it should be good.

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2021 Joshua Wallace
```
SteveMacenski, line 1: This node needs unit testing

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
```diff
@@ -0,0 +1,60 @@
...
+//
```
SteveMacenski, line 2: Both nodes need entries in the README in this package to explain what they are and what they do.

Also, in the https://navigation.ros.org/plugins/index.html and https://navigation.ros.org/migration/Galactic.html and https://navigation.ros.org/configuration/packages/configuring-bt-xml.html

nav2_bt_navigator/behavior_trees/navigate_w_replanning_distance_if_invalid_path.xml
```diff
@@ -0,0 +1,16 @@
+<root main_tree_to_execute="MainTree">
```
SteveMacenski, line 1: `navigate_w_replanning_only_if_path_becomes_invalid.xml`

nav2_planner/src/planner_server.cpp
```diff
@@ -27,6 +27,7 @@
 #include "builtin_interfaces/msg/duration.hpp"
```
SteveMacenski, line 27: Check the CI linters https://app.circleci.com/pipelines/github/ros-planning/navigation2/6576/workflows/0be4de15-e62c-4770-bb9f-f931edc6fd6d/jobs/24252. You can reproduce locally with `ament_cpplint` and `ament_uncrustify`

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +517,72 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+  const std::shared_ptr<nav2_msgs::srv::IsPathValid::Request> request,
+  std::shared_ptr<nav2_msgs::srv::IsPathValid::Response> response)
+{
+  unsigned int mx = 0;
+  unsigned int my = 0;
+
+  response->is_valid = true;
+
+  if (request->path.poses.empty() ) {
+    response->is_valid = false;
```
SteveMacenski, line 530: you can add a `return` here and then the rest of this function doesn't need to be in an `else` statement, which makes it easier to read

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +517,72 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+      float closest_distance = std::numeric_limits<float>::max();
+      float current_distance = std::numeric_limits<float>::max();
+      float prev_distance = std::numeric_limits<float>::max();
+      unsigned int increasing_distance_counter = 0;
+      const unsigned int MAX_INCREASE = 3;
+      geometry_msgs::msg::Point current_point;
+      current_point = current_pose.pose.position;
+      for (unsigned int i = 0; i < request->path.poses.size(); ++i) {
+        geometry_msgs::msg::Point path_point;
+        path_point.x = request->path.poses[i].pose.position.x;
```
SteveMacenski, line 544: `request->path.poses[i].pose.position` is a `geometry_msgs::msg::Point`, you can just do

`geometry_msgs::msg::Point path_point = request->path.poses[i].pose.position`

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +517,72 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+  } else {
+    geometry_msgs::msg::PoseStamped current_pose;
+    unsigned int closest_point_index = 0;
+    if (costmap_ros_->getRobotPose(current_pose)) {
+      float closest_distance = std::numeric_limits<float>::max();
+      float current_distance = std::numeric_limits<float>::max();
+      float prev_distance = std::numeric_limits<float>::max();
+      unsigned int increasing_distance_counter = 0;
+      const unsigned int MAX_INCREASE = 3;
+      geometry_msgs::msg::Point current_point;
```
SteveMacenski, line 540: `geometry_msgs::msg::Point current_point = current_pose.pose.position;` just do all in 1 line

nav2_planner/src/planner_server.cpp
```diff
@@ -510,7 +517,72 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+        current_distance = nav2_util::geometry_utils::euclidean_distance(
+          current_point,
+          path_point);
+
+        if (current_distance < closest_distance) {
+          closest_point_index = i;
+          closest_distance = current_distance;
+        }
+
+        if (current_distance > prev_distance) {
```
SteveMacenski, line 557: `prev_distance` isn't updated in the loop, I don't think this actually works...

SteveMacenski, line 557: Why don't you revert this logic and I'll take care of it as part of https://github.com/ros-planning/navigation2/issues/2622 when we have a solution I like for that problem to apply to all of the places at once. Sorry about that, just do the normal marching through to find the closest point.

SteveMacenski: Why don't you revert this logic and I'll take care of it as part of #2622 when we have a solution I like for that problem to apply to all of the places at once. Sorry about that, just do the normal marching through to find the closest point.

jwallace42: Done.

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_planner/src/planner_server.cpp
jwallace42: Done

jwallace42:
> * Push updates - please also include an updated BT visualization Groot or [this script](https://github.com/ros-planning/navigation2/blob/main/tools/bt2img.py) can help
> 
> *

nav2_bt_navigator/behavior_trees/navigate_w_replanning_distance_if_invalid_path.xml
jwallace42: Done

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
jwallace42: The current README in the nav2_behavior_tree does not have any documentation about specific BT nodes. I am assuming you would like me to start a new section?

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
jwallace42: Done

nav2_planner/src/planner_server.cpp
jwallace42: Done, however I am still getting errors about missing test results for the nav2 system tests. Not really sure how to handle those.

jwallace42:
I successfully integrated into `navigate_to_pose_w_replanning_and_recovery.xml`. 
However, when attempting to integrate into `navigate_through_poses_w_replanning_and_recovery.xml` the BT tree was throwing the following error that you have see before https://github.com/BehaviorTree/BehaviorTree.CPP/issues/271.

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
jwallace42: Also added to the documentation and submitted a pull request https://github.com/ros-planning/navigation.ros.org/pull/256

Kaetera:
Hi,
(thanks for this valuable PR !)
The declaration of the new BT nodes  IsPathValid and GlobalUpdatedGoal seem missing in the nav2_tree_nodes.xml palette. 
This will prevent any BT that have this nodes to be visualized with Groot (as in https://github.com/ros-planning/navigation2/issues/2643) :)

jwallace42:
> Hi, (thanks for this valuable PR !) The declaration of the new BT nodes IsPathValid and GlobalUpdatedGoal seem missing in the nav2_tree_nodes.xml palette. This will prevent any BT that have this nodes to be visualized with Groot (as in #2643) :)
> 
> EDIT: Oops sorry, just seen that you mentioned if here few comments above

Nope, you were correct. I just added it in.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
+      BT::InputPort<std::chrono::milliseconds>("server_timeout")
```
Kaetera, line 61: Hi,
Maybe there should be a default value set for the timeout (e.g.5-10ms as suggested in previous comments).
Here the default is 0ms, and when experimenting with this PR, it took me a little time to realize that IsPathValid was always failing because of that.

nav2_planner/src/planner_server.cpp
```diff
@@ -522,6 +529,56 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+    /**
+     * The lethal check starts at the closest point to avoid points that have already been passed
+     * and may have become occupied
+     */
+    for (unsigned int i = closest_point_index; i < request->path.poses.size(); ++i) {
+      costmap_->worldToMap(
+        request->path.poses[i].pose.position.x,
+        request->path.poses[i].pose.position.y, mx, my);
+      if (static_cast<unsigned int>(costmap_->getCost(mx, my)) >=
+        nav2_costmap_2d::LETHAL_OBSTACLE)
```
Kaetera, line 574: Hi, 
Shouldn't we check `== nav2_costmap_2d::LETHAL_OBSTACLE` instead of `>=` ?
Because NO_INFORMATION (255) > LETHAL_OBSTACLE (254), I think IsPathValid will not be usable with planners that plans in unknown regions (e.g. to do environment exploration)

Just a suggestion: when experimenting with this PR, I ended to check against `cost == LETHAL_OBSTACLE || cost == INSCRIBED_INFLATED_OBSTACLE` because:
-  If the path point is in the inscribed region then the robot will already collide with the obstacle if following the path, so the path is actually "obstructed"
- I especially wanted to retrigger planning when people crosses the path. But the legs seen by my lidars are represented only by one or two cells on my 0.1m resolution costmap, so the legs had to be exactly on the path to trigger replanning. Using the INSCRIBED_INFLATED_OBSTACLE made it more sensitive and efficient for this application

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
SteveMacenski: Just website is fine then, sorry my notebook on boilerplate responses is outdated :laughing:

nav2_planner/src/planner_server.cpp
SteveMacenski: @jwallace42 you can ignore the system tests backup/spin tests, those are just flaky, but the rest should work. Looks like at least 1 linting problem. 

You should also rebase / pull in main. It'll speed up CI and make the diff easier to process

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
SteveMacenski: Agreed, what's the default used elsewhere?

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2021 Joshua Wallace
```
SteveMacenski, line 1: You added a file ` tools/navigate_w_replanning_only_if_path_becomes_invalid.png.png `, that should be removed, but that would be great to add to the Nav2 docs PR. You could add a new page to https://navigation.ros.org/behavior_trees/index.html with the walk through (albeit simple)

nav2_planner/src/planner_server.cpp
```diff
@@ -167,6 +168,12 @@ PlannerServer::on_activate(const rclcpp_lifecycle::State & state)
     it->second->activate();
   }
 
+  is_path_valid_service_ = shared_from_this()->create_service<nav2_msgs::srv::IsPathValid>(
+    "is_path_valid",
+    std::bind(
+      &PlannerServer::isPathValid, this,
+      std::placeholders::_1, std::placeholders::_2));
+
   auto node = shared_from_this();
```
SteveMacenski, line 177: If you move this line above the create_service line, you can use `node` :smile:

nav2_planner/src/planner_server.cpp
```diff
@@ -522,6 +529,56 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
...
+void PlannerServer::isPathValid(
+  const std::shared_ptr<nav2_msgs::srv::IsPathValid::Request> request,
+  std::shared_ptr<nav2_msgs::srv::IsPathValid::Response> response)
+{
+  unsigned int mx = 0;
+  unsigned int my = 0;
+
+  response->is_valid = true;
+
+  if (request->path.poses.empty() ) {
```
SteveMacenski, line 541:
```suggestion
  if (request->path.poses.empty()) {
```

nav2_planner/src/planner_server.cpp
```diff
@@ -522,6 +529,56 @@ PlannerServer::publishPlan(const nav_msgs::msg::Path & path)
   }
 }
 
+void PlannerServer::isPathValid(
+  const std::shared_ptr<nav2_msgs::srv::IsPathValid::Request> request,
+  std::shared_ptr<nav2_msgs::srv::IsPathValid::Response> response)
+{
+  unsigned int mx = 0;
```
SteveMacenski, line 536: Move these closer to where they are used, L568-ish or so

nav2_planner/src/planner_server.cpp
SteveMacenski: I think it should be `== LETHAL || (!unknown_allowed && UNKNOWN)` type logic. That way if it becomes unknown and we don't allow for unknown travel that case is covered.

Where LETHAL is INSCRIBED_INFLATED_OBSTACLE and LETHAL_OBSTACLE, both. 

So something like 

```
if (!allow_unknown && cost == UKNWOWN) {
  return false
} else if (cost == LETHAL || cost == INSCRIBED) {
  return false
}

```

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
```diff
@@ -10,19 +10,21 @@
       <PipelineSequence name="NavigateWithReplanning">
```
SteveMacenski, line 10: Ah ok, so revert this file (except the removals of the goal updated bits we talked about before). Instead, create a **new** file that contains a full BT with your only replan on invalidation with a descriptive name. 

e.g. `navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml`

SteveMacenski: We want to preserve the option for people that do want constant replanning

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_planner/src/planner_server.cpp
jwallace42: Done

jwallace42: :)

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
jwallace42: Done!

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
jwallace42: The default is 20ms and it gets declared in the 

https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp on line 111. 

This got set correctly on my blackboard so I am not sure why @Kaetera is seeing it get set to 0ms. 

Do we want to add another parameter to nav2_params.yaml for service_timeout?

nav2_planner/src/planner_server.cpp
jwallace42: Added in this logic, not sure how to grab the allow_unknow_ parameter. I will need access to the active planner plugin.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
SteveMacenski: No, you need to add a default to the InputPort, right now there's no default value in it so it would use the initialized value of 0. Here's an example of a input port with a default set https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/back_up_action.hpp#L56

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
SteveMacenski: This image should still be removed from the branch

SteveMacenski:
See CI you have linting issues https://app.circleci.com/pipelines/github/ros-planning/navigation2/6664/workflows/cf09cfbf-4bed-4d3d-aa78-b24f5fd020b3/jobs/24612

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
SteveMacenski: Add as a comment in this PR but remove from source code

nav2_planner/src/planner_server.cpp
jwallace42: Done

nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.cpp
jwallace42: Removed

SteveMacenski: Still there...

SteveMacenski: Or you didn't push updates?

jwallace42: Sorry not yet! 

Just working through the comments now.

SteveMacenski: No worries :laughing: I was just too quick. I'll check back tomorrow afternoon

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
jwallace42: So, it looks like std::to_string has no def for chrono types which makes sense. 

/opt/ros/rolling/include/behaviortree_cpp_v3/basic_types.h:138:26: error: no matching function for call to ‘to_string(std::chrono::duration<long int, std::ratio<1, 1000> >&)’
  138 |     return std::to_string(value);
  
  This prevents the default being set. I spent some time trying to figure it out yesterday did not have any luck.

Kaetera: Got same problem, I ended to make this param an int to get the value as number of milliseconds, and then convert it later in an std::chrono duration (milliseconds or seconds, we should use the same unit as used in other BT nodes)

jwallace42: That is a possible solution but quite fragile. It would depend on people correctly converting and would result in a lot of file changes. I think a better solution would be to just have the timeout in the xml. Maybe a note also in the code too?

Kaetera: Actually when using a node with such param in the XML BTree, all params are entered as strings, so it does not really matters if it's cast to an int, float or chrono in the code. The user just has to know the unit expected, a good naming and documentation should be sufficient, like "server_timeout_ms". We already do rely on the documentation/naming for all the other nodes, e.g. to enter the `backup_dist` of the BackUp recovery node in meters, so I think it's ok to do the same for this timeout.
It involves only changing one or two lines in is_path_valid_condition.hpp and is_path_valid_condition.cpp (unless I missed some tests scripts that set this parameter).

SteveMacenski: That's what we use in all other contexts, https://github.com/ros-planning/navigation2/blob/779e3b60b9237374a76ed42233b116786715c575/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/spin_action.hpp#L57 have some seconds / milliseconds and it can be converted to chrono, this isn't fragile (or if it is, then its just as fragile as the ROS parameter systems which also doesn't support Chrono types in the yaml; nothing does for IO parsing...).

SteveMacenski: https://github.com/ros-planning/navigation2/blob/806029d0652c2618400f13fe5822f8ad671e1371/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L56-L58

Similarly, you could use the global service timeout set on the blackboard instead. Then have the input port be used **only if set** and so you don't need to specifically add a default. So you could keep this code the same without a default but then get the default from the  blackboard. And this code also does seem to show use of chrono types since we defined serialization/deserialization overrides for it in other header files we include.

jwallace42: Okay, so I switched it to a int and tested. Everything looks good. 

So deserialization does not seem to be an issue only serialization. 

https://github.com/ros-planning/navigation2/blob/806029d0652c2618400f13fe5822f8ad671e1371/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L56-L58

Also, this is a little odd. It seems like we are setting server_timeout twice?

SteveMacenski: Its looking from 2 sources but only uses the best value.

The blackboard has a generic timeout used by everything that wasn't explicitly set. Then we have a default-free version in BT nodes that use the timeout so they can be individually overrided. So what's happening is we use the global default from the blackboard **unless** the local server timeout port as a non-default value to set (which getInput handles). 

And its using Chrono types :-)

SteveMacenski: @jwallace42 please use the same strategy here: https://github.com/ros-planning/navigation2/blob/806029d0652c2618400f13fe5822f8ad671e1371/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L56-L58 with the same variables and default to blackboard values.

SteveMacenski:
Please move your images out of the branch (e.g. delete them from source) and put in a comment here so that we can retain them in the PR but not add them to our source code.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
+      BT::InputPort<int>("server_timeout_ms", 10, "Timeout for waiting for service response [ms]")
```
SteveMacenski, line 61: See comment in other thread - use the same variable as the others and don't set a default, use the blackboard server timeout variable as default.

nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml
```diff
@@ -0,0 +1,19 @@
...
+-->
+<root main_tree_to_execute="MainTree">
+  <BehaviorTree ID="MainTree">
+    <PipelineSequence name="NavigateWithReplanning">
+      <Fallback> 
+        <ReactiveSequence>
+          <Inverter>
+            <GlobalUpdatedGoal/>
+          </Inverter>
+          <IsPathValid path="{path}"/>
```
SteveMacenski, line 12: Question - Should this be throttled? How often is this running (at the tick rate of 100hz)?

SteveMacenski:
We're not done with this PR yet, but I would think it would be interesting to have a mixed implementation of this. Where we replan at some super slow rate (every 20 seconds or so) but also have the replanning when needed part of the BT there. That way we can eliminate oscillation behaviors and still get some replanning to refine over time -- but still update when strictly needed b/c its invalid.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
jwallace42: Done

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
jwallace42: Done

nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml
jwallace42: Done

jwallace42:
Yeah, sorry for the delay. 

A mixed implementation would be really cool and I don't think it would take much to implement. 

Everything should be ready to go now.

SteveMacenski:
There are a few test failures:

```
test_behavior
nav2_behavior_tree.IsPathValidTestFixture

unknown file
C++ exception with description "Blackboard::get() error. Missing key [server_timeout]" thrown in the test body.

behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml
nav2_bt_navigator.xmllint

Diff with 911 lines
/opt/overlay_ws/src/navigation2/nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml:19: parser error : Opening and ending tag mismatch: RateController line 7 and PipelineSequence
    </PipelineSequence>
                       ^
/opt/overlay_ws/src/navigation2/nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml:20: parser error : Opening and ending tag mismatch: PipelineSequence line 7 and BehaviorTree
  </BehaviorTree>
                 ^
/opt/overlay_ws/src/navigation2/nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml:21: parser error : Opening and ending tag mismatch: BehaviorTree line 7 and root
</root>
       ^
/opt/overlay_ws/src/navigation2/nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml:21: parser error : EndTag: '</' not found
</root>
       ^

TestBTXMLFiles
nav2_system_tests.BehaviorTreeTestFixture

/opt/overlay_ws/src/navigation2/nav2_system_tests/src/behavior_tree/test_behavior_tree_node.cpp:211
  bt_handler->loadBehaviorTree(entry.path().string())
    Which is: false
  true
```

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -0,0 +1,61 @@
...
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+  client_ = node_->create_client<nav2_msgs::srv::IsPathValid>("is_path_valid");
+}
+
+BT::NodeStatus IsPathValidCondition::tick()
+{
+  nav_msgs::msg::Path path;
+  getInput("path", path);
+
+  server_timeout_ = config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
```
SteveMacenski, line 37: You can move these 2 lines getting the server timeout to the constructor of the BT node so its only done once on construction

SteveMacenski:
Please also stick the images of the updated BTs in your PR's main text above so that its there for later reference for onlookers, then this should be good to go (after the small code change + tests passing)!

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml
```diff
@@ -0,0 +1,44 @@
+
+<!--
+  This Behavior Tree replans the global path only if the path becomes invalid and it also has
+  recovery actions specific to planning / control as well as general system issues.
+  This will be continuous if a kinematically valid planner is selected.
+-->
+<root main_tree_to_execute="MainTree">
```
SteveMacenski, line 7: I think you need to update your documentation PR (I already merged it, so open another with just the text change) for the new XML file here with the throttling & description of why there is throttling (so we only check on a reasonable rate in case there are cascading issues we don't want to have it replan at 100hz)

jwallace42: This was already correct in the documentation. The file that got updated is navigate_w_replanning_only_if_path_becomes_invalid.xml which is not in the documentation because it is not one of the main BTs.

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
jwallace42: Done

SteveMacenski:
This LGTM, anything else we need to do before merging?

SteveMacenski (approved):

jwallace42:
> This LGTM, anything else we need to do before merging?

Nope. It should be good.

SteveMacenski:
Thanks!! 

Now the BT update would be cool to see, make sure to test it and make sure that it fixes the problem we're setting out to.

- Replan only when required due to invalidity (throttled to some rate so we don't do it at 100hz in a bad situation) -- lets say 1 second?
- But also just replan on occasion to refine the path -- lets say 10-20 seconds? What do you say? I'm just throwing that number out of a hat.

afonsofonbraga:
Hi, @jwallace42  and @SteveMacenski! 
I was really happy to see this PR because this solves exactly the issue I'm facing when there is an obstacle right in front of the robot and the global planner cannot decide the shortest path, switching directions constantly.
However, I have tried to use the BT navigate_w_replanning_only_if_path_becomes_invalid but I get the error after following the path for a few seconds:
```
[bt_navigator-4] [INFO] [1651005827.201578852] [bt_navigator]: Begin navigating from current location to (5.21, -4.27)
[controller_server-1] [INFO] [1651005827.444162413] [controller_server]: Received a goal, begin computing control effort.
[controller_server-1] [ERROR] [1651005837.801537887] [controller_server]: Extrapolation Error: Lookup would require extrapolation into the past.  Requested time 4627.527000 but the earliest data is at time 4628.097000, when looking up transform from frame [map] to frame [odom]
[controller_server-1] 
[controller_server-1] [ERROR] [1651005837.801679195] [controller_server]: Global Frame: odom Plan Frame size 199: map
[controller_server-1] 
[controller_server-1] [ERROR] [1651005837.801727095] [controller_server]: Could not transform the global plan to the frame of the controller
[controller_server-1] [WARN] [1651005837.801857656] [controller_server]: [follow_path] [ActionServer] Aborting handle.
```
And after receiving this message the controller stops. This happens using the TEB controller. To solve it, I had to constantly update the global plan (every 5 seconds for example) but I'm not sure if the best solution for it. The BT I created to do it was:
```
<!--
  This Behavior Tree replans the global path only if the path becomes invalid
-->
  <BehaviorTree ID="MainTree">
      <RateController hz="1.0">
        <Fallback> 
          <ReactiveSequence>
            <Inverter>
              <GlobalUpdatedGoal/>
            </Inverter>
            <IsPathValid path="{path}"/>
          </ReactiveSequence>
        </Fallback>
      </RateController>
      <RateController hz="0.2">
      </RateController>
      <FollowPath path="{path}" controller_id="TEB"/>
    </PipelineSequence>
  </BehaviorTree>
</root>
```

Since you guys wanted to test in different scenarios, I thought this was some relevant information to share

SteveMacenski:
Huh, this is interesting. I'm curious if you find that to be true as well for DWB or RPP controllers. I'm just trying to suss out exactly where the issue is to help.

I imagine this doesn't happen right after you send a path, so when was that path computed? How long after that do you start to see that error? 

I think this is something we can do something about. @jwallace42 might be a good medium sized project too since it seems like you're the code owner for this awesome new feature! I imagine its the TF buffer not being large enough / data expiring so it can't find the look up to transform the path into the current time stamp. That seems like something we can work around with a little clever thought & thinking about the semantic meaning of the frames.

SteveMacenski:
I'd recommend seeing what TEB does differently than DWB / RPP in terms of path handling. But it sounds like this is an issue to file in TEB not here -- we are not maintainers of that over here and its in a separate repository.

Issue #2566, opened 2021-09-16
[Costmap2DROS] dynamic parameters


doisyg:
## Feature request
I have a personal interest in making the parameter `robot_radius` of `Costmap2DROS` dynamic (it is already in my fork). I could make a PR with only this parameter dynamic but this could also be the opportunity of making the others dynamic too.
Out of the following:
https://github.com/ros-planning/navigation2/blob/e5a4c844c748233dc4d3039b1ea5809c81dac45f/nav2_costmap_2d/src/costmap_2d_ros.cpp#L90-L116
which one should we make dynamic ?

SteveMacenski:
It is already possible, the footprint subscriber is there to update based on dynamic changes to a robot's size https://github.com/ros-planning/navigation2/blob/e5a4c844c748233dc4d3039b1ea5809c81dac45f/nav2_costmap_2d/src/costmap_2d_ros.cpp#L176-L179. You would have to specify the circle yourself though on publication (e.g. polygon that makes a circle) rather than just defining a radius, but you can use the costmap_2d function `makeFootprintFromRadius` in your application to generate it. 

In terms of which of these should be dynamic, I'd use ROS1 as a starting point https://github.com/ros-planning/navigation/blob/noetic-devel/costmap_2d/cfg/Costmap2D.cfg and use your best judgement for the rest.

SteveMacenski:
@doisyg should this be rolled into the main dynamic parameters ticket?

PR #2592, opened 2021-10-11
Add dynamic parameters do Costmap2DROS

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/956  and https://github.com/ros-planning/navigation2/issues/2566 |

---

As discussed https://github.com/ros-planning/navigation2/issues/2566, this PR makes dynamic the parameters that were in ros1: https://github.com/ros-planning/navigation/blob/noetic-devel/costmap_2d/cfg/Costmap2D.cfg

Only `update_frequency` was left out as it requires `std::thread * map_update_thread_` to be reinit, but this can be debated.
I tried to reproduce the priority order of what sets the footprint : from topic > footprint string > robot_radius, hence the messy change callback.
The `std::vector<geometry_msgs::msg::Point> ` from the topic needs to be saved when it is set in order to be able to apply a changing `footprint_padding` on it.

Side notes not directly related to this PR:

- With the current implementation, applying padding on a round footprint distorts it to a rounded square
- Default value of `footprint_padding` is 0.0 in the code and 0.01 in the documentation

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/test/integration/dyn_params_tests.cpp
+++ b/nav2_costmap_2d/test/integration/dyn_params_tests.cpp
@@ -17,7 +17,7 @@
 
 #include "gtest/gtest.h"
 #include "rclcpp/rclcpp.hpp"
-#include "tf2_geometry_msgs/tf2_geometry_msgs.h"
+#include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
 #include "nav2_costmap_2d/costmap_2d_ros.hpp"
 #include "tf2_ros/transform_broadcaster.h"
 


doisyg:
>  All of the costmap_2d parameters that could be changed (non-plugin)?

For completeness, list of parameters that are currently NOT dynamic (they could theoritically be made dynamic too but it requires some re-initialization of objects):

- `always_send_full_costmap` -> needs costmap_publisher_ reinit
- `global_frame` -> needs layered_costmap_  costmap_publisher_ tf_buffer_ reinit
- `lethal_cost_threshold` , `map_topic` , `observation_sources` , `trinary_costmap` , `unknown_cost_value` , `clearable_layers` -> not used in costmap2DROS itself but in pluggins
- `plugins` -> needs layered_costmap_ reinit
- `filters` -> needs layered_costmap_ reinit
- `robot_base_frame` -> needs tf_buffer_ reinit
- `rolling_window` -> needs layered_costmap_ reinit
- `track_unknown_space` -> needs layered_costmap_ reinit

As for tests, any suggestions ?

SteveMacenski:
@mergifyio refresh

SteveMacenski:
Feel free to ignore the mergify stuff, I was just testing that everything was set up correctly. It now is

moritzboeker:
@doisyg and @SteveMacenski , thank you for implementing your solution! This is very useful since I was looking for a way to dynamically set the footprint of our articulated robot comprising a front and rear carriage with an articulated joint in between them. Unfortunately, I am forced to use ROS melodic or noetic. So I started to take your changes and implement them into a **ROS1 navigation fork, but realized that I cannot depend on rcl_interfaces. Is there a way to get this running also on ROS1?**

Kind regards,
Moritz

PR #2607, opened 2021-10-15
adding dynamic params to RPP and cleaningup style in others

SteveMacenski:
@doisyg let me know if you see anything wrong


SteveMacenski:
- [ ] linting
- [ ] test coverage

SteveMacenski:
@lucabonamini you wrote the dynamic parameters tests for smac, I copied that same format here for RPP but they're failing. Can you take a look at this and tell me what I'm missing? For some reason, I seriously cannot find my error why these are failing to set

SteveMacenski:
Weird - works now, there was a param that was misnamed that without atomically created an exception that I was able to isolate. Once fixed, atomically worked for all of them -- I suppose it just rejects all param updates atomically if any are incorrect causing the issue. Good to know. Working now

PR #2609, opened 2021-10-15
Modified the projection_time calculation

Naraharirahul:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here ##2594) |
| Robotic platform tested on | (On turtlebot waffle in turtlebot world simulation) |

---

I have updated the `projection_time` variable calculation method by adding the `fabs() ` to the `linear_vel` which was missing. This change now enables the robot to move in the reverse direction.

#### For Maintainers: 


SteveMacenski:
@Mergifyio backport foxy-devel

PR #2624, opened 2021-10-18
Stopping networking-related BT failures from failing task, rather fail specific BT node

SteveMacenski:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2376  |

---

* Adding try/catch loop around `is_future_goal_handle_complete` to catch node-specific exception failures rather than propagating them to the root tree
* Allows to fail the BT node for networking related failures rather than failing the entire navigation task (then goes to recoveries) 

#### For Maintainers: 


SteveMacenski:
https://github.com/ros-planning/navigation.ros.org/pull/222 Docs PR

Issue #2621, opened 2021-10-18
RPP produces discontinuous angular velocities when the linear one is limited


dpm-seasony:
## Bug report

**Required Info:**

- Operating System:
  - Kubuntu 20.04
- ROS2 Version:
  - Galactic binaries
- Version or commit hash:
  - 8faa0f0550027edc7bf2640f23cff13b2af60a8f
- DDS implementation:
  - default

#### Steps to reproduce issue

```
Controller server parameters:
`controller_server:
  ros__parameters:
    use_sim_time: True
    controller_frequency: 20.0
    min_x_velocity_threshold: 0.001
    min_y_velocity_threshold: 0.001
    min_theta_velocity_threshold: 0.001
    progress_checker_plugin: "progress_checker"
    goal_checker_plugin: "goal_checker"
    controller_plugins: ["FollowPath"]
    # Progress checker parameters
    progress_checker:
      plugin: "nav2_controller::SimpleProgressChecker"
      required_movement_radius: 0.5
      movement_time_allowance: 30.0
    # Goal checker parameters
    goal_checker:
      plugin: "nav2_controller::SimpleGoalChecker"
      xy_goal_tolerance: 0.25
      yaw_goal_tolerance: 0.15
      stateful: True
    # DWB parameters
    FollowPath:
      plugin: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
      desired_linear_vel: 0.25
      max_linear_accel: 0.3
      max_linear_decel: 0.3
      lookahead_dist: 0.9
      min_lookahead_dist: 0.3
      max_lookahead_dist: 1.5
      lookahead_time: 1.5
      rotate_to_heading_angular_vel: 0.175
      transform_tolerance: 0.1
      use_velocity_scaled_lookahead_dist: true
      min_approach_linear_velocity: 0.05
      use_approach_linear_velocity_scaling: true
      max_allowed_time_to_collision: 5.0
      use_regulated_linear_velocity_scaling: true
      regulated_linear_scaling_min_radius: 0.7
      regulated_linear_scaling_min_speed: 0.05
      allow_reversing: true
      use_rotate_to_heading: false
      rotate_to_heading_min_angle: 0.35
      max_angular_accel: 0.05
      use_cost_regulated_linear_velocity_scaling: false`




```

#### Expected behavior
max_linear_accel and max_linear_decel (maybe max_angular_accel?) limit accelerations so that the twist control commands have a smooth shape and continuous shape.

[...]

PR #2631, opened 2021-10-20
removing kinematic limiting from RPP

SteveMacenski:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2621 |
| Robotic platform tested on | seasony bot |

---

* Removing kinematic limiting in RPP after finding it creates discontinuities
* This should be handled by a lower level velocity smoother anyway 

- [x] Update migration docs that params are gone
- [x] Update docs with parameters out of RPP config guide

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/test/test_regulated_pp.cpp
+++ b/nav2_regulated_pure_pursuit_controller/test/test_regulated_pp.cpp
@@ -406,8 +406,6 @@ TEST(RegulatedPurePursuitTest, testDynamicParameter)
       rclcpp::Parameter("test.min_lookahead_dist", 6.0),
       rclcpp::Parameter("test.lookahead_time", 1.8),
       rclcpp::Parameter("test.rotate_to_heading_angular_vel", 18.0),
-      rclcpp::Parameter("test.max_linear_accel", 0.5),
-      rclcpp::Parameter("test.max_linear_decel", 0.5),
       rclcpp::Parameter("test.min_approach_linear_velocity", 1.0),
       rclcpp::Parameter("test.max_allowed_time_to_collision", 2.0),
       rclcpp::Parameter("test.cost_scaling_dist", 2.0),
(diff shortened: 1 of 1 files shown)


vinnnyr:
+1 for this change. These parameters were causing issues in our robot. Specifically the issues were related to selecting parameters to help RPP overcome a deadband region in our machine controller, while not gunning the acceleration, which was incredibly difficult to do.

doisyg:
Docs should be updated to reflect that change, I got confused believing accel limits were still in place:
- https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/README.md

doisyg:
What's the rational behind removing the linear accel limits and not the angular limit ?

doisyg:
Last spam on this merge. For a robot with a very little linear decel capability, how to not overshoot the goal position ? By playing with the `lookahead_dist` and scaling params ?

SteveMacenski:
I updated the readme.

> For a robot with a very little linear decel capability, how to not overshoot the goal position ? By playing with the lookahead_dist and scaling params ?

Perhaps, yeah the scaling parameters would be good. This had not considered much robots having unusually low decel capabilities, so perhaps there are modifications that can be made to support. https://github.com/ros-planning/navigation2/issues/2815 and its complimentary PR have some such change that I haven't gotten to reviewing yet, but may be of interest?

doisyg:
Thanks for the pointer, the PR addresses precisely my needs. I may end up testing it very soon on a heavy platform (~500kg). However only on straight line paths. Will report directly there https://github.com/ros-planning/navigation2/issues/2815

SteveMacenski:
OK. I don't have an issue with their general idea, its the scale of changes due to the new library I'm not super familiar with that I'm a little weary about. And where else we should maybe be leveraging that.

Issue #2442, opened 2021-07-08
[Smac Smoother] handle reverse direction reeds sheep models as individual segments


SteveMacenski:
Right now the smoother will smooth the entire path, meaning that it will smooth the cusp in direction changes. That probably isn't good. Instead we should be treating each segment in a velocity direction separately to retain the cusp and directional changes.

dpm-seasony:
Could you confirm that the PR  #2660 solves this issue?

salmagro:
I have experienced a similar effect, and I would like to clarify if it corresponds to the same issue treated here.

The function `updateApproximatePathOrientations` recalculate the heading based on the order of the XY coordinates without considering the driving direction.

This generates a continuous path to drive forward, provoking the misbehavior of the local planner.

![example1](https://user-images.githubusercontent.com/52298571/142390772-ffc33f1a-5254-4873-924c-5fcf7ce3c885.png)
![example2](https://user-images.githubusercontent.com/52298571/142390775-f4331ff3-8101-4f77-991c-b3df8f2b0990.png)

SteveMacenski:
@salmagro try testing the PR #2660. It should work, I've asked for more programming style updates. Let us know if that fixes your issue

salmagro:
@SteveMacenski thank you for the information. Yes, PR #2660 solved the problem with the wrong heading. However, I have experienced a bug described in https://github.com/ros-planning/navigation2/pull/2660#issuecomment-974087387 .

SteveMacenski:
It's my intention to address this tomorrow and get something merged in hopefully before the end of the week. The issue with the smoother making the plans potentially no longer kinematically feasible remains, but at least reversing will not be negatively impacted anymore (e.g. I can stop commenting out the smoother when I do tests in reverse).

 I'll plan on addressing, or asking someone else to address, the issue with feasibility as a separate topic covered in #2690. I want to start working on these problems as individuals so we can make incremental progress.

SteveMacenski:
I have a working solution to this, but discovered some quirks with the State Lattice planner that need to be resolved before I can merge it. We need the primitive markers to be some distance apart so I can use the dot product to discover cusps. Right now, it over-fires on some situations that I'm trying to fully analyze so that I can fix them all.

For now since I'm in a bit of a waiting cycle, I'm also going to try and tackle the smoother feasibility challenge, which might also encompass re-adding the parameter for smoothing or not so that it can be disabled.

SteveMacenski:
[...]

PR #2660, opened 2021-11-04
[Smac Smoother] Avoid smoothing cusps

dpm-seasony:
---

## Basic Info

| Info | Please fill out this column  |
| ------ | ----------- |
| Ticket(s) this addresses   | #2442 |
| Robotic platform tested on | Seasony bot simulation |

---

* Path segments with different velocity directions are smoothed separately to retain cusps

---

#### For Maintainers: 


nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -340,8 +340,23 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
 #endif
 
   // Smooth plan
-  if (num_iterations > 1 && plan.poses.size() > 6) {
-    _smoother->smooth(plan, costmap, time_remaining);
+  // Check if there are any cusps and if so, split path to avoid smoothing cusps
+  if (num_iterations > 1) {
+    std::vector<unsigned int> cusps_id = findDirectionChange(plan);
+    if (cusps_id.size() < 3 && plan.poses.size() > 6) {
```
SteveMacenski, line 346: Why `cusps_id.size() < 3`?

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -340,8 +340,23 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
...
+        _smoother->smooth(plan, costmap, time_remaining);
+    } else {
+      for (unsigned int i = 1; i < cusps_id.size(); ++i) {
+        if (cusps_id[i] - cusps_id[i - 1] > 6) {
```
SteveMacenski, line 350: This is the right idea, but I think this should be in the smoother class's `smooth` instead of the planner implementation. It's a specific niche of this specific smoother, not a planner-level concern.

nav2_smac_planner/include/nav2_smac_planner/utils.hpp
```diff
@@ -104,6 +104,36 @@ inline double findCircumscribedCost(std::shared_ptr<nav2_costmap_2d::Costmap2DRO
   return result;
 }
 
+/**
+ * @brief Checks for the cusp(s) position(s)
+ * @param path Path in which to look for cusps
+ * @return Cusps' indices plus initial and final path poses' indices
+ */
+inline std::vector<unsigned int> findDirectionChange(const nav_msgs::msg::Path & path)
```
SteveMacenski, line 112: Need a test for this

SteveMacenski (changes requested):
In general seems fine, just needs to move the logic around a little.

How well does this work, functionally? There's a bug in the smoother that makes it so that the final on-approach for the path segment might be discontinuous (e.g. if the path is 20 long, the 19-20 line segment will be fine but the 18-19 can be discontinuous due to heavy deflection in the smoothing process. I lock in 19-20 so that the final heading can be respected from the request, which results in that slight discontinuity in smoothness).

 This is not really much of an issue for the full system, since as we approach the goal, the path gets shorter on reaplanning so that when we're within a few meters of the goal, this will smooth out properly since not much more deflection from the root path is happening anymore (as opposed to before where there's alot of curves and path length potentially to be smoothed out).

But if you're separating out the segments here, then the issue will appear at every single direction change point, not just at the end. The same rational applies that it might not be important, but it could be more significant if its happening on important points like directional changes. 

e.g. https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/include/nav2_smac_planner/smoother.hpp#L118 why we do length - 2. 19 and 20 would stay statically grounded. It could also be that I did largely fix that and I don't remember, but if you don't see it, then maybe I did fix it?

The answer here might be a fix in that, which could potentially be easy, but I didn't find one and it practically didn't matter - the way it was being applied before.

nav2_smac_planner/src/smac_planner_hybrid.cpp
dpm-seasony: The vector cusps_id always contains at least the indices of the initial and final poses, thus if no cusps changes of direction were found the vector size would be 2. I added some comments in the code regarding this

nav2_smac_planner/src/smac_planner_hybrid.cpp
dpm-seasony: I've moved this to the smoother class' smooth as you suggested

nav2_smac_planner/include/nav2_smac_planner/utils.hpp
dpm-seasony: Where should the test be added? test_smac_hybrid, test_nodehybrid or somewhere else?

dpm-seasony:
For me it works rather well. Without this PR I basically could not use the smoother because I need to be sure that the cusps are kept in the path. I don't think I've seen the discontinuity issue that you mentioned.

nav2_smac_planner/include/nav2_smac_planner/utils.hpp
SteveMacenski: You can add a new file `test_utils.cpp` if you like

nav2_smac_planner/src/smac_planner_hybrid.cpp
SteveMacenski: Why return the first and last points? Why would those be considered cusps?

SteveMacenski:
Can you show a few screen grabs of this working?

nav2_smac_planner/include/nav2_smac_planner/smoother.hpp
```diff
@@ -72,8 +73,33 @@ class Smoother
     nav_msgs::msg::Path & path,
     const nav2_costmap_2d::Costmap2D * costmap,
     const double & max_time,
-    const bool do_refinement = true)
+    const bool do_refinement = true, const bool single_dir_segment = false)
   {
+    if (!single_dir_segment) {
```
SteveMacenski, line 78: This all technically works... but I would avoid recursion if not required. This could just be made to put the core smoothing logic below into a separate `smoothSegment` function and then just loop over it with the various segments and attach them in the end, without need for recursion.  The `do_refinement` was necessary to have recursion in unfortunately, but this can be modeled more clearly without it.

SteveMacenski:
@dpm-seasony a couple of reorg changes and this should be good to go!

Also I wanted to ask you: I'm working on some planning stuff with Smac and seeing that if I enable backward expansion, sometimes the analytic expansions will find a path that involves driving backwards for a somewhat long period of time. This will happen if you give it a goal where it needs to back into the final approach but has visibility early on. Is that a problem to you? That seems like that could potentially be a problem if robots are driving backwards across large parts of a warehouse, but maybe if you have enabled the possibility of going backwards its OK because you have sensor coverage there too?

nav2_smac_planner/include/nav2_smac_planner/smoother.hpp
dpm-seasony: I thought about doing that but went for recursion to minimize the amount of changes. I can reorganize the logic once I find a bit of spare time

dpm-seasony:
For me moving backwards long distances is not really an issue as our robot has perception coverage on the rear as well. Nonetheless I would indeed prefer to have mostly forward motions simply because it looks more natural.
I do think that having the possibility of limiting the analytic expansion could be useful

SteveMacenski:
Awesome, its in the schedule to have that in by end of the year (likely end of Nember)

salmagro:
I have tested this MR in order to solve https://github.com/ros-planning/navigation2/issues/2442#issuecomment-972702195.

The problem of having the wrong heading is solved. However, after a few iterations, the `planner_server` node is dying. 
Here you could see the `call stack`. Looks that is related to the smooth function.

2021-11-19_13-49_bug_stack

SteveMacenski:
That's a weird place to die, can you inspect why that died? Assuming your costmap is actually valid from a real test, I can't see why mx, my, or the size would be wrong

SteveMacenski:
@dpm-seasony it appears that you are potentially too busy to bring this home. I'll plan on taking your approach and my comments into account and implement a full solution for your review this week. I hope you'll take a look when I link it back.

 I greatly appreciate your time and help in starting the ball rolling on this issue. I look forward to other contributions in the future! This is not a reflection of your work - its a reflection of my panicking about getting some of these bugs squashed before 2022 :laughing:

dpm-seasony:
> @dpm-seasony it appears that you are potentially too busy to bring this home. I'll plan on taking your approach and my comments into account and implement a full solution for your review this week. I hope you'll take a look when I link it back.
> 
> I greatly appreciate your time and help in starting the ball rolling on this issue. I look forward to other contributions in the future! This is not a reflection of your work - its a reflection of my panicking about getting some of these bugs squashed before 2022 laughing

Indeed I had been super busy last weeks. Fortunately it seems that I'll be able to squeeze some spare time next week for reviewing the full solution

SteveMacenski:
New PR open https://github.com/ros-planning/navigation2/pull/2727

Issue #2673, opened 2021-11-16
Severe warning of nav2_amcl at runtime 


gezp:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 20.04
- ROS2 Version:
  - Rolling binaries
- Version or commit hash: 
  - `main` branch
- DDS implementation:
  - default DDS

#### Steps to reproduce issue
```
ros2 launch nav2_bringup tb3_simulation_launch.py
```

#### Expected behavior
* no severe warning
#### Actual behavior
```bash
[amcl-7] Warning: class_loader.impl: SEVERE WARNING!!! A namespace collision has occured with plugin factory for class nav2_amcl::DifferentialMotionModel. New factory will OVERWRITE existing one. This situation occurs when libraries containing plugins are directly linked against an executable (the one running right now generating this message). Please separate plugins out into their own library or just don't link against the library and use either class_loader::ClassLoader/MultiLibraryClassLoader to open.
[amcl-7]          at line 253 in /opt/ros/rolling/include/class_loader/class_loader_core.hpp
[amcl-7] Warning: class_loader.impl: SEVERE WARNING!!! A namespace collision has occured with plugin factory for class nav2_amcl::OmniMotionModel. New factory will OVERWRITE existing one. This situation occurs when libraries containing plugins are directly linked against an executable (the one running right now generating this message). Please separate plugins out into their own library or just don't link against the library and use either class_loader::ClassLoader/MultiLibraryClassLoader to open.
[amcl-7]          at line 253 in /opt/ros/rolling/include/class_loader/class_loader_core.hpp
[amcl-7] Warning: class_loader.ClassLoader: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded.
[amcl-7]          at line 127 in /tmp/binarydeb/ros-rolling-class-loader-2.1.2/src/class_loader.cpp

```
#### Additional information

reason:
* motion model is refactored as a plugin library `motions_lib`, and `amcl_core` is also a plugin library (rclcpp component), `target_link_libraries(amcl_core  motions_lib)`  causes this severe warning.
* AmclNode should load motion model at runtime, so  `amcl_core`  shouldn't link this plugin library.

[...]

PR #2675, opened 2021-11-16
unlink motions_lib plugin library in nav2_amcl

gezp:

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | ( #2673) |

---

* remove API  `nav2_amcl::MotionModel::createMotionModel()` ,  use `pluginlib::ClassLoader` load MotionModel in `AmclNode` directly.
* remove redundant library `motion_model` (as same as `motions_lib`)
* remove `motions_lib` from `target_link_libraries` of `amcl_core`.

---

#### For Maintainers: 


nav2_amcl/CMakeLists.txt
```diff
@@ -68,14 +68,6 @@ set(dependencies
   pluginlib
 )
 
-add_library(motion_models SHARED
```
SteveMacenski, line 71: I think we still need a library for use and be installed? It just shouldn't be `target_link_libraries` against the main executables

gezp: `motion_models` library is same as `motions_lib` , you could find here : https://github.com/ros-planning/navigation2/blob/464bfffec0d81fcc14b4b27ddf30ee83e418436b/nav2_amcl/src/motion_model/CMakeLists.txt#L1-L16

poornimajd: Yes,sorry I had just missed that part and created redundant code in CMakelists.txt.

SteveMacenski:
Retriggering CI, but every single system test has failed, which indicates to me that AMCL isn't coming up right (unless transient)

SteveMacenski:
test_localization failed, rerunning, that's an unusual test to have fail and seems as though it may relate...

SteveMacenski (approved):

SteveMacenski:
Thanks @gezp !

PR #2681, opened 2021-11-17
Amcl updated code

poornimajd:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | ( #2673) |

---

Fixed the severe warnings due to the redundant code in CMakelists.txt .
Removed the motions_lib library from CMakelists.txt

#### For Maintainers: 


nav2_amcl/CMakeLists.txt
```diff
@@ -89,7 +90,7 @@ ament_target_dependencies(${library_name}
 )
 
 target_link_libraries(${library_name}
-  map_lib motions_lib sensors_lib motion_models
+  map_lib sensors_lib motion_models
```
SteveMacenski, line 93: I'm almost entirely sure that `motion_models` shouldn't be linked here, that was the crux of the issue, the pluginlib library should not be linked against the use library.

SteveMacenski (changes requested):

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/2675's solution is closer, closing this PR

Issue #2658, opened 2021-11-02
CostmapTopicCollisionChecker footprint tilted/shifted


afrixs:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Galactic binaries
- Version or commit hash:
  - 0f6f49a3633de66db7cdf5f60ecbb951fea5883f
- DDS implementation:
  - Eclipse Cyclone DDS

#### Steps to reproduce issue
 1. Make robot move, e.g.
```
ros2 topic pub /cmd_vel geometry_msgs/msg/Twist "linear:
   x: 0.0
   y: 0.0
   z: 0.0
 angular:
   x: 0.0
   y: 0.0
   z: 1.0"
```
 2. Use a component which uses CostmapTopicCollisionChecker, e.g.
```
ros2 launch nav2_bringup navigation_launch.py
```
```
ros2 action send_goal /backup nav2_msgs/action/BackUp "target:
   x: 1.0
   y: 0.0
   z: 0.0
 speed: 0.2
 time_allowance:
   sec: 10
   nanosec: 0"
```
 3. Visualize footprints returned by `CostmapTopicCollisionChecker::getFootprint()` (needs some code editing)

#### Expected behavior
Footprints should be aligned with the checked poses.

#### Actual behavior
Footprints have an angular/positional offset.

#### Implementation considerations
Offset is caused because timestamp is being ignored in tf lookup. The stamp used for creating the footprint message should be used for transform lookup in `CostmapTopicCollisionChecker::unorientFootprint()`, i.e.:
 1. Relevant pose stamp must not be discarded [when publishing footprint message](https://github.com/ros-planning/navigation2/blob/0f6f49a3633de66db7cdf5f60ecbb951fea5883f/nav2_costmap_2d/src/costmap_2d_ros.cpp#L470) (`pose.header.stamp` needs to be used instead of `now()`)
 2. The stamp must be retrievable from FootprintSubscriber. This needs discussion, since there are multiple ways to achieve it. For example it could be done as follows:
[...]

PR #2687, opened 2021-11-18
CostmapTopicCollisionChecker tf stamp fix

robotechvision:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2658 |
| Robotic platform tested on | (RTV Caster, gazebo simulation of RTV Caster) |

---

* Fixed footprint stamp rewriting by Costmap2DROS
* Added header retrieval to FootprintSubscriber
* Moved footprint "unorienting" from CostmapTopicCollisionChecker to Footprint subscriber and made it use footprint stamp for tf lookup
* Removed footprint_timeout from FootprintSubscriber
* Updated FootprintSubscriber API and added doxygen description to getFootprintRaw and getFootprintInRobotFrame methods

* Removed `global_frame` parameter from smoother server, so need to remove that from default configs and documentation page

---

* Check for other places where tf lookup has possibly been used improperly

#### For Maintainers: 


SteveMacenski:
You have tests failing, in particular the collision test isn't a flaky test. If you pull in main as well, I recently fixed a bunch of flaky tests and at least release builds should be totally green now.

nav2_costmap_2d/include/nav2_costmap_2d/footprint_subscriber.hpp
```diff
@@ -60,10 +60,15 @@ class FootprintSubscriber
     std::vector<geometry_msgs::msg::Point> & footprint,
     rclcpp::Duration & valid_footprint_timeout);
   /**
-   * @brief Returns an oriented robot footprint without timeout
+   * @brief Returns an oriented robot footprint using timeout specified in constructor
```
SteveMacenski, line 63:
```suggestion
   * @brief Returns an oriented robot footprint using timeout in argument
```

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
```diff
@@ -92,25 +95,27 @@ double CostmapTopicCollisionChecker::scorePose(
...
   Footprint footprint;
-  if (!footprint_sub_.getFootprint(footprint)) {
+  rclcpp::Time stamp(0L, clock_->get_clock_type());
+  if (!footprint_sub_.getFootprint(footprint, stamp)) {
     throw CollisionCheckerException("Current footprint not available.");
   }
 
   Footprint footprint_spec;
-  unorientFootprint(footprint, footprint_spec);
+  unorientFootprint(footprint, footprint_spec, stamp);
```
SteveMacenski, line 104: You add stamp, which is just Time(0), and pass in this function to `getCurrentPose` where you added the stamp argument, which defaults to Time() if not provided. 

Which was the previous behavior without an argument at all, I don't think you ever use `getCurrentPose` which a non Time() value to have any change. Maybe worth reverting those changes? I see changes but no use of the features they would enable.

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -466,8 +466,7 @@ Costmap2DROS::updateMap()
       layered_costmap_->updateMap(x, y, yaw);
 
       auto footprint = std::make_unique<geometry_msgs::msg::PolygonStamped>();
-      footprint->header.frame_id = global_frame_;
-      footprint->header.stamp = now();
+      footprint->header = pose.header;
```
SteveMacenski, line 469: So far, I'm not seeing any actual changes to behavior / code other than right here, where the stamp goes from now() to Time(0)

nav2_costmap_2d/src/footprint_subscriber.cpp
```diff
@@ -68,9 +68,19 @@ FootprintSubscriber::getFootprint(
     return false;
   }
 
+  stamp = footprint_stamp;
```
robotechvision, line 71: This is the difference, the stamp is now filled with actual footprint stamp inside getFootprint().

nav2_costmap_2d/include/nav2_costmap_2d/footprint_subscriber.hpp
robotechvision: But there's no timeout argument here. This method uses FootprintSubscriber's own footprint_timeout_.

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
robotechvision: Actually the stamp is retrieved inside `footprint_sub_.getFootprint(footprint, stamp)`. Yeah, the line 98 should be `rclcpp::Time stamp = clock_->now();` instead of Time(0) so that the footprint timeout check can be performed. That means, stamp is an in-out parameter of `getFootprint` method (in for timeout check, out for getting actual footprint timeout). I agree the usage of parameters of this method is not very intuitive and should be documented more in detail (I needed to study the code too to get a picture of how it actually works) or the API should be changed somehow (that's why I requested discussion in the issue report).

nav2_costmap_2d/src/costmap_2d_ros.cpp
robotechvision: It doesn't go to Time(0). It's filled with actual pose stamp by tf inside `getRobotPose()` method.

SteveMacenski:
I’ll look at the sp. comments in the morning, but did want to mention quickly that tests are failing so something’s not quite right. We’d need to fix that before merging regardless. 

Also, pull in main or rebase, the release tests should be totally green. If there’s a flaky debug test that wouldnt be concerning. I’m working through that issue this week.

robotechvision:
Could you please have a look at the test results? I fixed the tests and they work on my machine but some infrastructure problems are reported here and an absolutely unrelated fibonacci test is failing

SteveMacenski:
There was something odd in the CI infra, please push some commit, rebase, etc in order to retrigger it, I'm unable to do it from the admin terminal

nav2_costmap_2d/src/costmap_2d_ros.cpp
SteveMacenski: So `pose` gets its header from getting the robot pose, which is calling https://github.com/ros-planning/navigation2/blob/cf10abebfee60fb5cf09df6836b24594b68c083c/nav2_util/src/robot_utils.cpp#L49, does this actually populate a header stamp? Because you hand it a default constructed pose stamped (https://github.com/ros-planning/navigation2/blob/cf10abebfee60fb5cf09df6836b24594b68c083c/nav2_util/src/robot_utils.cpp#L49) and I'm not sure TF will actually populate that. Please verify that

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
```diff
@@ -96,25 +99,27 @@ double CostmapTopicCollisionChecker::scorePose(
 Footprint CostmapTopicCollisionChecker::getFootprint(const geometry_msgs::msg::Pose2D & pose)
 {
   Footprint footprint;
-  if (!footprint_sub_.getFootprint(footprint)) {
+  rclcpp::Time stamp = clock_->now();
+  if (!footprint_sub_.getFootprint(footprint, stamp)) {
```
SteveMacenski, line 103: All https://github.com/ros-planning/navigation2/blob/cf10abebfee60fb5cf09df6836b24594b68c083c/nav2_costmap_2d/src/footprint_subscriber.cpp#L94 does is call https://github.com/ros-planning/navigation2/blob/cf10abebfee60fb5cf09df6836b24594b68c083c/nav2_costmap_2d/src/footprint_subscriber.cpp#L85 which has the `now()` call in it, why change the use here? Does this buy us something? This looks like the usecase for your new `getFootprint()` function, but I'm not seeing how this helps in synchronizing, it looks like its just shuffling those lines around rather than using an existing API

SteveMacenski: Oh I see, its so you can share the same `now()` timestamp with `unorientFootprint()`. Is there anywhere where the old just `getFootprint(footprint)` API is still used and wouldn't be able to be updated to use the new one specifying `now()` as an argument? If not, delete that old function and replace all its uses with this one.

SteveMacenski (changes requested):

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
```diff
@@ -96,25 +99,27 @@ double CostmapTopicCollisionChecker::scorePose(
 Footprint CostmapTopicCollisionChecker::getFootprint(const geometry_msgs::msg::Pose2D & pose)
```
SteveMacenski, line 99: Question: From a software separations of concern standpoint, should be we move this function https://github.com/ros-planning/navigation2/blob/cf10abebfee60fb5cf09df6836b24594b68c083c/nav2_costmap_2d/src/costmap_topic_collision_checker.cpp#L101-L111 to the footprint subscriber itself? It seems like a footprint subscriber's job to take in the footprint and give out a reasonable one given a timestamp of request. There may be users of `FootprintSubscriber` without the topic collision checker that would want a refined & transformed solution out of the box.

nav2_costmap_2d/src/costmap_2d_ros.cpp
afrixs: Of course it works like that. I'm not going to formulate any proofs here for verification, please spend some time expanding your horizons yourself with these basic ROS elements, it will be beneficial for you and this project.

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
afrixs: Yes, this is a good idea. I'm going to put it there along with other cleanup and readability updates of FootprintSubscriber.

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
afrixs: No, it's not the same `now()`, `stamp` is being updated inside `getFootprint(footprint, stamp)`. After 3 unsuccessful attempts of explaining these few lines I think it's time to do a proper cleanup of FootprintSubscriber and remove everything that's misleading and useless (namely `footprint_timeout` check) + to write a proper doxygen, so that FootprintSubscriber is less confusing and easy to understand.

nav2_costmap_2d/src/costmap_2d_ros.cpp
SteveMacenski: There is no reason to get nasty. There is a massive amount of code I have to keep track of spanning most robotics domains, its not a reasonable expectation that I remember off the top of my head how every single function is internally implemented at all times, even having read TF2 many times. 

You're adding a reliance on an internal implementation detail of an external library. I'm simply asking for verification.

nav2_costmap_2d/src/costmap_topic_collision_checker.cpp
SteveMacenski: I agree, this was from an old Intel design when they were trying to have an independent environmental model and use the footprint topic as more of a word-of-god type thing synchronized with the costmaps for use across the system. This is code we haven't seriously looked at in about 2 years because it was refactored out, now only used by the recovery behaviors which weren't as context sensitive. 

For context, the original design intent was to make sure that if the footprint wasn't received in synchronous with the costmap (before the topic sychronizers were ported to ROS 2) that we'd fail planning or control because a dropped footprint message could be fatal in a bad situation if assumed to be paired with a costmap update. But now that we don't use this tool anymore in that context, the timeout is nonsensical as its morphed to have a refinement process to get it into the correct frame in an updated timestamp.

nav2_recoveries/src/recovery_server.cpp
```diff
@@ -73,21 +73,19 @@ RecoveryServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
   tf_->setCreateTimerInterface(timer_interface);
   transform_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_);
 
-  std::string costmap_topic, footprint_topic;
+  std::string costmap_topic, footprint_topic, robot_base_frame;
+  double transform_timeout;
   this->get_parameter("costmap_topic", costmap_topic);
   this->get_parameter("footprint_topic", footprint_topic);
-  this->get_parameter("transform_tolerance", transform_tolerance_);
+  this->get_parameter("transform_timeout", transform_timeout);
```
SteveMacenski, line 80: transform_tolerance is the standardized parameter name for transform timeout across all of the packages, please change that name back

SteveMacenski (changes requested):
These changes look good to me. The only subjective question I'd have is whether we should add the `transformFootprint()` into the footprint subscriber as well, rather than in the costmap collision checker, so that a user of the footprint subscriber could get the raw, unoriented in base frame, or transformed footprint.

nav2_smoother/src/nav2_smoother.cpp
```diff
@@ -84,23 +83,21 @@ SmootherServer::on_configure(const rclcpp_lifecycle::State &)
...
   transform_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_);
 
-  std::string costmap_topic, footprint_topic;
-  double transform_tolerance;
+  std::string costmap_topic, footprint_topic, robot_base_frame;
+  double transform_tolerance, footprint_tolerance;
   this->get_parameter("costmap_topic", costmap_topic);
   this->get_parameter("footprint_topic", footprint_topic);
   this->get_parameter("transform_tolerance", transform_tolerance);
+  this->get_parameter("footprint_tolerance", footprint_tolerance);
```
SteveMacenski, line 91: This should be removed, I believe you depreciated this

SteveMacenski:
Remaining things are just adding a migration guide section update. I believe with removing `footprint_tolerance` there are no other parameter updates so no need to update any of the guides.

nav2_costmap_2d/src/costmap_2d_ros.cpp
afrixs: I'm sorry, I got tired of repetitive persuasion about that the update actually does something. If simple 'yes' was enough for you then I could have gone just with it.

afrixs:
Regarding the question, it could be a nice bonus to provide such a function, but I personally think it doesn't need to be under the responsibility of the subscriber, what will be done next with the footprint. Especially when it would save just one simple line of code. I updated `transformFootprint` method so that it can take the same variable as input and output, making this subsequent transform simpler and less error-prone.

nav2_smoother/src/nav2_smoother.cpp
```diff
@@ -49,7 +49,6 @@ SmootherServer::SmootherServer()
     "footprint_topic",
     rclcpp::ParameterValue(
       std::string("global_costmap/published_footprint")));
-  declare_parameter("global_frame", rclcpp::ParameterValue(std::string("map")));
```
SteveMacenski, line 52: This parameter is removed, that's OK, we just have to remove it from the config guide documentation. But before we do that, do you think the plugins will want to know this? If so, maybe we keep it in knowing that it'll be used later.

SteveMacenski (changes requested):
Last question, then good to be merged

nav2_smoother/src/nav2_smoother.cpp
afrixs: You mean that the parameter would remain declared so that plugins could use it? Well, unlike in recovery server, I don't think a usual smoother would get use of it. Smoothers get all input data (path poses, costmap) in one frame (+ footprint in robot frame but there's no further need for tf lookup, it can be transformed directly to path poses), work with them in one frame and return the results in the same frame. If there was a special smoother (e.g. which considers other robots' footprints), it can still declare its own parameter (or get the global frame from the header of an oriented footprint).

BTW, when I found this PR unmerged I took liberty to do one final minor optimization that came to mind, hopefully it's OK.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_topic_collision_checker.hpp
```diff
@@ -64,31 +56,29 @@ class CostmapTopicCollisionChecker
...
 protected:
-  /**
-   * @brief Set a new footprint
-   */
-  void unorientFootprint(const Footprint & oriented_footprint, Footprint & reset_footprint);
   /**
    * @brief Get a footprint at a set pose
    */
-  Footprint getFootprint(const geometry_msgs::msg::Pose2D & pose);
+  Footprint getFootprint(const geometry_msgs::msg::Pose2D & pose, bool fetch);
```
SteveMacenski, line 73: I think `fetch` should be removed here or needs to default to `true` like the others so default behavior gets the latest footprint. A comment in the collision checker regarding these `fetch` and `fetch_costmap_and_footprint` would be nice to tell users why this is there (so if looping over many poses, using the same costmap / footprint since the costmap specifically would incur a pretty gnarly copy if working on the global costmap).

SteveMacenski (changes requested):

nav2_costmap_2d/include/nav2_costmap_2d/costmap_topic_collision_checker.hpp
SteveMacenski: Basically, just want this context to be made clear to future readers

nav2_costmap_2d/include/nav2_costmap_2d/costmap_topic_collision_checker.hpp
```diff
@@ -69,8 +78,12 @@ class CostmapTopicCollisionChecker
 protected:
   /**
    * @brief Get a footprint at a set pose
+   *
+   * @param pose Pose to get footprint at
+   * @param fetch_latest_footprint Defaults to true. When checking with multiple poses at once,
+   * footprint should be fetched in the first check but fetching can be skipped in consequent checks for speedup
    */
-  Footprint getFootprint(const geometry_msgs::msg::Pose2D & pose, bool fetch);
+  Footprint getFootprint(const geometry_msgs::msg::Pose2D & pose, bool fetch_latest_footprint);
```
SteveMacenski, line 86:
```suggestion
  Footprint getFootprint(const geometry_msgs::msg::Pose2D & pose, bool fetch_latest_footprint = true);
```

I think you just need the default field and LGTM.

SteveMacenski:
Sorry about Circle, maybe some update came on the servers, I'm not sure why just the debug builds would be failing. 

I proposed a change above, I'd just merge it in and then merge in the PR, but wanted you to approve it. If you approve it and commit the suggestion, I'll merge this PR on Monday.

SteveMacenski (approved):

SteveMacenski:
Ugh, I really don't mean this to be the never ending PR but it failed a linting check

```
Diff with 7 lines
--- include/nav2_costmap_2d/costmap_topic_collision_checker.hpp
+++ include/nav2_costmap_2d/costmap_topic_collision_checker.hpp.uncrustify
@@ -86 +86,3 @@
-  Footprint getFootprint(const geometry_msgs::msg::Pose2D & pose, bool fetch_latest_footprint = true);
+  Footprint getFootprint(
+    const geometry_msgs::msg::Pose2D & pose,
+    bool fetch_latest_footprint = true);

```

If you haven't fixed it by the time I get to this Monday morning during work hours, I'll just merge it and quickly submit the patch to fix. I don't want to have this be hung up any longer for something so small.

Issue #2674, opened 2021-11-16
Live groot monitoring error messages 


SteveMacenski:
```
[composed_bringup-8] [PublisherZMQ] Publisher quitting.
[composed_bringup-8] [PublisherZMQ] just died. Exeption Context was terminated
```

SteveMacenski:
Due to loading default tree on activation, and not waiting for first request

PR #2696, opened 2021-11-19
removing live groot monitoring from Nav2

SteveMacenski:
There have been near-endless tickets of issues regarding live groot monitoring, so we'll be removing it from Nav2 since it really hasn't worked out well since we introduced it. 

It was a great feature idea but never quite panned out, especially after we introduced multiple navigator types in the BT Navigator server. The issue we run into primarily is that Zero-MQ prevents users from producing multiple logger types in the same process. Since BT nav has multiple servers, the swapping between them for viewing has never had a clean hand off causing folks to file tickets or have nasty logs appear or ZMQ crashes in the background. The BT.CPP client for this doesn't allow us to have a clean shutdown process so we're left with hoping that ZMQ properly handles the situation, which it rarely does. Further, Groot only supports visualizing one type of tree at a time so for applications often switching between navigator types, its not possible to use a single groot client, causing great frustration.

So, what I propose here is to remove live monitoring of the BT from Nav2. **We can still use Groot to modify, visualize, and generally work with behavior trees**, the only thing being removed is to live view the executing behavior tree as Nav2 is currently executing it (it used to light up the boxes of the current nodes). This was of dubious value anyhow, since the tree ticks so fast its difficult to visualize and get meaningful insights into things as the system is moving _so_ quickly.  

Ex. tickets

- https://github.com/ros-planning/navigation2/issues/2406
- https://github.com/ros-planning/navigation2/issues/2674

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/behavior_tree_engine.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/behavior_tree_engine.hpp
@@ -84,24 +84,6 @@ public:
     BT::Blackboard::Ptr blackboard);
 
   /**
-   * @brief Add groot monitor to publish BT status changes
-   * @param tree BT to monitor
-   * @param publisher_port ZMQ publisher port for the Groot monitor
-   * @param server_port ZMQ server port for the Groot monitor
-   * @param max_msg_per_second Maximum number of messages that can be sent per second
-   */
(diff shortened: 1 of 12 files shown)


elyarzv:
I have created a behavior tree in ROS1. I am able to have live tree monitoring the only problem is that is doesn't reflect the status of lthe last node and consequetly the whole tree.

Issue #2659, opened 2021-11-02
Smac Hybrid Planner ignores non-lethal costs around the end of the plan


zoltan-lengyel:
**Facts**
The smac hybrid planner periodically attempts to do an analytic solution to finish the plan as it expands _(related function: AStarAlgorithm<NodeT>::tryAnalyticExpansion)_.
When there is a cached analytic solution, the planner checks if there is any collision on this analytic path _(DUBIN curve)_.
If there is a collision, the planner discards the analytic solution and tries it again a few expansion later.
If there is no collision, the plan is finished using the analytic solution, regardless of the obstacle costs it crosses.
_(Technically, the planner may pick a different analytic solution further away due to a refining process, but this is not relevant for the point I am trying to make.)_

**Issues**
The analytic solution can make the plan go through really high cost areas at the last 10 meters or even more, because its only criteria for success is not colliding.
This can make the plan go really close to obstacles.
In our navigation system the costs in the costmap are not only the result of obstacle inflation, they serve navigation purposes as well, like making the robots follow a preferred line. These navigation constraints are also ignored along the analytic solution.

**Ideas**
I tried simply limiting the distance the analytic solution can be triggered from, to **minimum_turning_radius * 3**.
I did not change the lookup_table_size, because I still wanted to use the analytic solutions to influence the heuristics _(NodeHybrid::getDistanceHeuristic)._
This resolved my issues fairly well. Best case performance degraded of course in simple cases, but the worst case performances for complicated planning cases were not affected significantly _(some special cases significantly improved even)_.
In my case losing some of the extremely quick results is a perfectly acceptable tradeoff for stricter cost awareness around the end of the plan, so I would be glad to have some configuration options for this.
Another idea I had was to disregard the analytic solution only if it was long and crossing high obstacle costs. The reason I didn't like this idea very much, because it doesn't help much if the goal is close to an obstacle, which happens very often in my use cases.

**Questions**
Do you have any suggestions on how to resolve this issue in a better or nicer way?
Do you think the smac hybrid planner should have some configuration options to improve this situation at the cost of performance?

afrixs:
I'm supporting this feature request since we experience better planning results too when the analytic expansion is somewhat constrained. Despite the fact that the analytic expansion is certainly needed to provide a reasonable final path section (as discretized node search often misses the goal), some constraining is needed to prevent dangerous cost-unaware maneuvers, especially in tight spaces around the goal.

[...]

PR #2697, opened 2021-11-20
adding max analytic expansion length

SteveMacenski:
https://github.com/ros-planning/navigation2/issues/2659

- [x] test for performance
- [x] fix unit tests
- [x] Change to parameter in meters and used in expander in cell coords
- [x] dynamic params

### DIFF
--- a/nav2_smac_planner/README.md
+++ b/nav2_smac_planner/README.md
@@ -115,6 +115,7 @@ planner_server:
       cost_travel_multiplier: 2.0         # For 2D: Cost multiplier to apply to search to steer away from high cost areas. Larger values will place in the center of aisles more exactly (if non-`FREE` cost potential field exists) but take slightly longer to compute. To optimize for speed, a value of 1.0 is reasonable. A reasonable tradeoff value is 2.0. A value of 0.0 effective disables steering away from obstacles and acts like a naive binary search A*.
       angle_quantization_bins: 64         # For Hybrid nodes: Number of angle bins for search, must be 1 for 2D node (no angle search)
       analytic_expansion_ratio: 3.5       # For Hybrid/Lattice nodes: The ratio to attempt analytic expansions during search for final approach.
+      analytic_expansion_max_length: 3.0    # For Hybrid/Lattice nodes: The maximum length of the analytic expansion to be considered valid to prevent unsafe shortcutting
       minimum_turning_radius: 0.40        # For Hybrid/Lattice nodes: minimum turning radius in m of path / vehicle
       reverse_penalty: 2.1                # For Reeds-Shepp model: penalty to apply if motion is reversing, must be => 1
       change_penalty: 0.20                # For Hybrid nodes: penalty to apply if motion is changing directions, must be >= 0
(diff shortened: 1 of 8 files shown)


nav2_smac_planner/src/analytic_expansion.cpp
```diff
@@ -132,6 +132,14 @@ typename AnalyticExpansion<NodeT>::AnalyticExpansionNodes AnalyticExpansion<Node
 
   float d = node->motion_table.state_space->distance(from(), to());
 
+  // If the length is too far, exit. This prevents unsafe shortcutting of paths
+  // into higher cost areas far out from the goal itself, let search to the work of getting
+  // close before the analytic expansion brings it home. This should never be smaller than
+  // 4-5x the minimum turning radius being used, or planning times will begin to spike.
+  if (d > _search_info.analytic_expansion_max_length) {
```
afrixs, line 139:
```suggestion
  if (d*_collision_checker->getCostmap()->getResolution() > _search_info.analytic_expansion_max_length) {
```
Didn't check if it has changed in recent updates but `d` was in map cell size units instead of meters. In that case default and readme values should be reevaluated accordingly.

nav2_smac_planner/README.md
```diff
@@ -115,6 +115,7 @@ planner_server:
       cost_travel_multiplier: 2.0         # For 2D: Cost multiplier to apply to search to steer away from high cost areas. Larger values will place in the center of aisles more exactly (if non-`FREE` cost potential field exists) but take slightly longer to compute. To optimize for speed, a value of 1.0 is reasonable. A reasonable tradeoff value is 2.0. A value of 0.0 effective disables steering away from obstacles and acts like a naive binary search A*.
       angle_quantization_bins: 64         # For Hybrid nodes: Number of angle bins for search, must be 1 for 2D node (no angle search)
       analytic_expansion_ratio: 3.5       # For Hybrid/Lattice nodes: The ratio to attempt analytic expansions during search for final approach.
+      analytic_expansion_max_length: 3.0    # For Hybrid/Lattice nodes: The maximum length of the analytic expansion to be considered valid to prevent unsafe shortcutting
```
zoltan-lengyel, line 118: I would also mention that it is in meters, and that it should be scaled together with minimum_turning_radius.

SteveMacenski: Added

nav2_smac_planner/src/analytic_expansion.cpp
SteveMacenski: Yup, sorry mentioned in the ticket https://github.com/ros-planning/navigation2/issues/2659#issuecomment-974570505 but will push commit shortly that changes

SteveMacenski:
Pushed updates
- Support for dynamic params
- Fix CI linting / tests with change
- Using meters for parameter and converting to grid cells for A* to use

nav2_smac_planner/README.md
SteveMacenski: It's also more detailed in the configuration guide on the website in the docs PR that will be merged at the same time https://github.com/ros-planning/navigation.ros.org/pull/234/files#diff-4b80e4229af9d6c9cb965894da700179259b3dc1925f28641a3cc024c29adeeeR91

SteveMacenski:
@afrixs @zoltan-lengyel take a look again, I think it should satisfy both of your requests. 

I did some performance testing and I'm very happy with the quality of paths with the defaults even - very little if any skating close to obstacles of approach!

afrixs:
Looks good! Just noting that before putting it into meters the value 3.0 meant 3 cells, i.e. 0.15 meters, i.e. 0.374*minimum_turning_radius. So if you were satisfied with the results before the change, it means that it can be still OK to have the value far lower than 4-5x minimum_turning_radius (our tests have shown that 1.5-2.2x is quite good). However, it's subjective and it's just about the note and suggestion not to push the value too low is in place.

SteveMacenski:
I did my testing after updating to have the correct units, since its REP 103 that states standardized ROS parameter units are SI. It was a brief oversight that I didn't have that and quickly figured it out while initially testing (and seeing unit test number of iterations spike). 

The default of 3m is ~7x the default turning radius of 0.4m for Hybrid / Lattice files, well above the recommended 4x from my rough testing. I've also seen lower than 4x do OK, but then occasionally it spikes up, so I give it a little more latitude. The defaults should be a balance of good out of the box performance for many situations (then dug into to optimize for a particular application).

PR #2642, opened 2021-10-22
Refactored amcl code for modularization of motion models(main branch)

poornimajd:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | add tickets here #2552 |
| Primary OS tested on | Ubuntu 20.04 |
| Robotic platform tested on | turtlebot3 |

---

## Description of contribution in a few bullet points
- The amcl source code has been slightly modified and a plugin package for motion model has been created.

- Pluginlib is used to make the differential and omni model as plugins.Major files changed are related to the motion_model folder.

- This functionality will enable a developer to test his/her own custom motion model by just creating it as a plugin and adding the name in the nav2_params.yaml file.
-  Code is tested on the ros2 foxy branch.

## Description of documentation updates required from your changes

- The parameter robot_model_type in nav2_params.yaml file is to be taken as the fully-qualified type of the plugin class.For example - robot_model_type: "nav2_amcl::DifferentialMotionModel" if the name of the custom motion model class is "DifferentialMotionModel" and the namespace is "nav2_amcl".

- To test any custom motion model ,just add the motion_model.hpp file to the source code of the motion model implemented as a plugin.

---
#### For Maintainers: 
- [ ] Check that any new parameters added are updated in navigation.ros.org
- [ ] Check that any significant change is added to the migration guide
- [ ] Check that any new functions have Doxygen added
- [ ] Check that any new features have test coverage
- [ ] Check that any new plugins is added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

PR #2700, opened 2021-11-22
Remove "wrong"/unnecessary dependency

destogl:
## Problem
`ament_cmake_ros` dependency is added in #2642. This breaks build in our repository where navigation2 is dependency.

## Solution
Looking at the changed in #2642 I am not sure why this dependency is added, or it is at least added wrongly because there should equivalent in the package.xml file and there is none.

Based on my experience working on 10+ CPP repositories with ROS2 I think the dependency is not necessary, therefore propose to remove it.


SteveMacenski (approved):

SteveMacenski:
Toggled to get CI to trigger, assuming it passes, I'll merge

SteveMacenski:
Thanks!

destogl:
Thank you for the prompt response!

Issue #2670, opened 2021-11-11
RPP Warnings at goals near obstacles on approach


SteveMacenski:
@kaetera @sathak93

SteveMacenski:
Any ideas on this?

sathak93:
I am busy with other stuff. Next week i can work on this.

sathak93:
Since you tried  RPP with footprint collision checking . What do you think about the warnings?

SteveMacenski:
I think there were 2 topics
- When we rotate to heading, we need the collision checker to do orientation-based discretization, not linear velocity based
- When we're approaching the goal that is very close to an obstacle, we need something to help with reducing warnings

I can work on the first one this week and have a PR ready for you both to take a look over, but its still a little bit of an open question how to handle the warnings of collision when approaching the goal / stop collision checking at the goal pose.

SteveMacenski:
@sathak93 @kaetera https://github.com/ros-planning/navigation2/pull/2701 please review!

Kaetera:
I thinks it's ok, to keep a general behaviour similar to the current one.
In the original issue you said: "I think that logic should only be used for pure rotations for the rotate to headings".
But here are my thoughts:
1. If the purpose of the collision checking in RPP is to prevent a collision in case "something goes wrong" and that the robot is no more driven by RPP but e.g. just by its inertia, then projecting the poses in the future based on the current velocity makes sense. But currently we project using the velocity command, not the current velocity, and ignore the angular velocity except for pure rotation, so this check is approximative.
2. On the other hand, if the purpose is to check that there is no obstacle on the path (e.g. if the global planner plans every 1s, then an obstacle can appear on the path between two replanning), then we should project the poses along the path.

My feeling is that the current implementation is a mix between 1. and 2.
And since we check the collision in rotation only in pure rotation, solution 1. looks incomplete since it does not covers the case of turning in curves (cf. my awesome drawing below)
If the purpose is 2. then we could implement a collision check inspired by MPC methods, that would account for that RPP will reduce the command near goal, in curves, etc. Then:
- It would naturally handle the "near goal case" (since obstacles beyond the goal are not on the path)
- Avoid the similar "curve case": e.g. if the path does a 90° turn with a small radius, the current implementation triggers collision detections, what is a little annoying in constrains environments like mine where since the robot stops frequently 

[...]

PR #2701, opened 2021-11-23
[RPP] remove velocity scaling option (now required) and enable orientation sp projection times

SteveMacenski:
https://github.com/ros-planning/navigation2/issues/2670

- [x] update docs for param change https://github.com/ros-planning/navigation.ros.org/pull/236
- [x] Migration guide param gone https://github.com/ros-planning/navigation.ros.org/pull/236
- [x] TODO get radius

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -412,8 +412,9 @@ bool RegulatedPurePursuitController::isCollisionImminent(
     // theta_min = 2.0 * sin ((res/2) / r_max)
     // via isosceles triangle r_max-r_max-resolution,
     // dividing by angular_velocity gives us a timestep.
+    double max_radius = costmap_ros_->getLayeredCostmap()->getCircumscribedRadius();
     projection_time =
-      2.0 * sin((costmap_->getResolution() / 2) / 0.25 /*TODO radius*/) / fabs(angular_vel);
+      2.0 * sin((costmap_->getResolution() / 2) / max_radius) / fabs(angular_vel);
   } else {
     // Normal path tracking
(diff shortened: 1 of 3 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -409,7 +404,20 @@ bool RegulatedPurePursuitController::isCollisionImminent(
...
+    // Equation finds the angular distance required for the largest
+    // part of the robot radius to move to another costmap cell:
+    // theta_min = 2.0 * sin ((res/2) / r_max)
+    // via isosceles triangle r_max-r_max-resolution,
+    // dividing by angular_velocity gives us a timestep.
+    projection_time =
+      2.0 * sin((costmap_->getResolution() / 2) / 0.25 /*TODO radius*/) / fabs(angular_vel);
+  } else {
+    // Normal path tracking
+    projection_time = costmap_->getResolution() / fabs(linear_vel);
```
Kaetera, line 420: Shouldn't we check that ```linear_vel``` is not null to prevent division-by-zero ?

Kaetera (approved):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: There is no time that is possible if you look at the code. There are tons of minimum parameterizations set such that it cannot be 0.0 ever, plus if it were 0.0, `0.0 * curvature` would be nonsensical so the entire algorithm would fail to operate

SteveMacenski: I suppose it could happen if people misset the parameters, I'll make a note of that in the configuration guide.

Issue #2698, opened 2021-11-22
Theta star planner crashes


daisukes:
Hello, I have been trying to use the theta star planner, which is very fast, and I love it very much!!
I found a bug that caused a crash when the planner could not make a plan (return zero-length poses).
Please let me know if you want to make a PR (or two for galactic/main branches)

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - galactic binary release
- Version or commit hash:
  - galactic 3ac88193c0b52c7e6e71bbf9de8818979f5996c9
- DDS implementation:
  - Fast-RTPS (does not matter)

#### Steps to reproduce issue
- provide a goal where the theta star planner cannot produce a plan

#### Expected behavior
- fail without a crash

#### Actual behavior
- crashes at theta_star_planner.cpp#L119
- length of `global_path.poses` can be zero
https://github.com/ros-planning/navigation2/blob/d8f50da2ed59879112105ac1488e9139ee7f3c71/nav2_theta_star_planner/src/theta_star_planner.cpp#L118-L119

SteveMacenski:
Seems like a simple fix! Can you submit a PR to check if the size is `0` and if so, return early? No need for any of that orientation logic if its completely empty

PR #2703, opened 2021-11-23
Theta star planner fix

daisukes:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2698 |
| Primary OS tested on |  Ubuntu |
| Robotic platform tested on | test code |

---

* fix the bug #2698, check if a plan is generated to avoid crash
* add test case if the goal is unsafe


SteveMacenski (approved):

SteveMacenski:
Thanks for adding a complimentary test!

PR #2704, opened 2021-11-24
adding dynamic parameters support for controller, planner, and waypoint servers

SteveMacenski:
- [x] needs unit test coverage
- [x] add mutex
https://github.com/ros-planning/navigation.ros.org/pull/242

---

@doisyg I believe that we do not require the mutexes in most of the dynamic reconfigure callbacks. The dynamic parameter service is tied to the `node` given to the algorithms, which is `shared_from_this()` from the servers. That's the same executor that is being used for the `ActionServer` in the controller / planner servers.That implies that all of the algorithm reconfiguration functions _couldn't_ run at the same time as the action, only between action calls. That means that we could remove the mutexes because they would not be able to run at the same time.

Do you agree? If so, I'll update this PR to remove all of the mutexes in Smac, DWB, Theta*, RPP, NavFn, and goal / progress checkers. Costmap (?) I'll need to look at more specifically, I think that does actually use another callback group

---

This PR leaves only AMCL and the costmap layers as non-dynamically reconfigurable. https://github.com/ros-planning/navigation2/issues/956

### DIFF
--- a/nav2_bringup/src/composed_bringup.cpp
+++ b/nav2_bringup/src/composed_bringup.cpp
@@ -19,7 +19,7 @@
 #include "rclcpp/rclcpp.hpp"
 #include "nav2_map_server/map_server.hpp"
 #include "nav2_amcl/amcl_node.hpp"
-#include "nav2_controller/nav2_controller.hpp"
+#include "nav2_controller/controller_server.hpp"
 #include "nav2_smoother/nav2_smoother.hpp"
 #include "nav2_planner/planner_server.hpp"
 #include "nav2_recoveries/recovery_server.hpp"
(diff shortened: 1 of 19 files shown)


doisyg:
Are you sure ? I know that I can modify parameters of my controller (a forked version of TEB) while the `FollowPath` action of the controller_server is running (and that's actually my desired behavior)

SteveMacenski:
Yeah, I tracked it down and its called as part of `set_parameter` by the application that's setting the dynamic parameter. I can't quite tell from a quick 5 minute look where that's actually happening, but I tracked down where the callbacks are being triggered and its happening on setting parameters. My guess is that's the mechanism the dynamic parameters is using, though I couldn't immediately find that connection.

Mutexes are required :+1:

SteveMacenski:
Adding test coverage :+1:

Issue #2692, opened 2021-11-18
costmap_2d::VoxelLayer::raytraceFreeSpace fails with because clearing_endpoints_ is empty


m2-farzan:
## Bug report

**Required Info:**

- Operating System:
Archlinux
- ROS2 Version:
Galactic (built from source) (gcc 11.1.0) (repos file) (galactic branch of https://github.com/ros2/ros2) (HEAD: c7c318e)
- Nav2 Version or commit hash:
Galactic (built from source) (gcc 11.1.0) (git clone) (galactic branch of this repo) (HEAD: d8f50da2ed59879112105ac1488e9139ee7f3c71)
- DDS implementation:
  - rmw_fastrtps_cpp

#### Steps to reproduce issue

```
ros2 launch nav2_bringup tb3_simulation_launch.py slam:=True
```

#### Expected behavior
Launch a gazebo simulation with turtlebot3 in a closed environment, with Nav2 running in slam mode. User should be able to mark "2D pose estimate" and "Nav2 Goal" in rviz, and the robot should move to target position.

#### Actual behavior
nav2_controller node fails at startup with no error message. As a consequence, "Goal was rejected by server" is printed when user selects a goal in rviz

#### Additional information
I recompiled everything with debug info (`-DCMAKE_BUILD_TYPE=RelWithDebInfo`) and got this backtrace:
```c++
#0  0x00007ffff74a0d22 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff748a862 in abort () from /usr/lib/libc.so.6
#2  0x00005555555be5fa in std::__replacement_assert (__file=<optimized out>, 
    __line=<optimized out>, __function=<optimized out>, __condition=<optimized out>)
    at /usr/include/c++/11.1.0/x86_64-pc-linux-gnu/bits/c++config.h:504
#3  0x00007ffff7c6dcf8 in std::vector<unsigned char, std::allocator<unsigned char> >::front (
    this=0x7fffb00042c8) at /usr/include/c++/11.1.0/bits/stl_vector.h:1121
#4  std::vector<unsigned char, std::allocator<unsigned char> >::front (this=0x7fffb00042c8)
    at /usr/include/c++/11.1.0/bits/stl_vector.h:1121
[...]

PR #2705, opened 2021-11-24
Always initialize clearing endpoints in voxel layer

m2-farzan:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2692 |
| Primary OS tested on | (Ubuntu, Archlinux) |
| Robotic platform tested on | (turtlebot3 in gazebo) |

---

* With this fix we avoid calling `std::vector::front()` on an empty vector, which is undefined behavior and causes glibc asserts to fail in runtime. See the ticket for more info.

[none]

---

* Can be backported to galactic. Older releases are not affected.

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Will merge after CI

Issue #2711, opened 2021-11-26
AMCL crashes with assertion after initialpose received


AlexeyMerzlyakov:
AMCL crashes with `i < set_a->sample_count` assertion while "initialpose" message being received. This appears from time to time in such system tests as `test_speed_filter_global` during the Circle-CI integration. The failure logs are as follows:
```
[amcl-8] [INFO] [1637842857.912058601] [amcl]: initialPoseReceived
[amcl-8] [WARN] [1637842857.912216894] [amcl]: Failed to transform initial pose in time (Lookup would require extrapolation into the future.  Requested time 14.438000 but the latest data is at time 14.410000, when looking up transform from frame [base_footprint] to frame [odom])
[amcl-8] [INFO] [1637842857.912258118] [amcl]: Setting pose (14.438000): -2.000 -0.500 0.000
[amcl-8] amcl: /opt/overlay_ws/src/navigation2/nav2_amcl/src/pf/pf.c:381: pf_update_resample: Assertion `i < set_a->sample_count' failed.
[ERROR] [amcl-8]: process has died [pid 25975, exit code -6, cmd '/opt/overlay_ws/install/nav2_amcl/lib/nav2_amcl/amcl --ros-args -r __node:=amcl --params-file /tmp/tmpln7ui15p -r /tf:=tf -r /tf_static:=tf_static'].
```

This also could be reproduced locally in CI Docker container running the selected test manually.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04 focal
- ROS2 Version:
  - ROS2 Rolling (20211105 version)
- Version or commit hash:
  - 89bbb6d37ffffe3adfe7e09d45b556b84c1bf985
- DDS implementation:
  - Cyclone DDS (CI default), Fast DDS

#### Steps to reproduce issue
Describing steps to reproduce the problem on local Docker CI container:
1. Download docker CI Docker image and make a container from it:
```
$ docker pull ghcr.io/ros-planning/navigation2:main
$ docker image list
REPOSITORY                         TAG                IMAGE ID       CREATED        SIZE
ghcr.io/ros-planning/navigation2   main               6ade87d6aaba   2 days ago     3.49GB
$ docker run -t -d --name nav2 6ade87d6aaba
```

2. Start Docker container:
```
$ docker start nav2
$ docker exec -it nav2 /bin/bash
```

3. Build nav2 stack at Docker side:
```
# source /opt/ros/rolling/setup.bash
[...]

PR #2712, opened 2021-11-29
Prohibit of simultaneous usage of callbacks accessing to pf_ object

AlexeyMerzlyakov:
Prohibit simultaneous usage of `pf_` shared object by possible callbacks. This will eliminate inconsistent data within `pf_`, (such as `pf_->sets[]->samples[]`) causing crashes of AMCL during "initialpose" and "laserscan" messages receiving at the same time (described in #2711)

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2711 |
| Primary OS tested on | Ubuntu 20.04 focal |
| Robotic platform tested on | Nav2 integrated tests |

### DIFF
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -238,6 +238,7 @@ protected:
    */
   static pf_vector_t uniformPoseGenerator(void * arg);
   pf_t * pf_{nullptr};
+  std::mutex pf_mutex_;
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

SteveMacenski:
Thanks for finding and fixing that so quickly!

SteveMacenski:
Note: This won't be an issue once we move AMCL over to a single Node object. This happens because `rclcpp_node_` is used for a separate executor from the services

PR #2750, opened 2021-12-20
support dynamic composed bringup

gezp:

---

## Basic Info

| Ticket(s) this addresses   | #2147                                    |
| Primary OS tested on       | Ubuntu20.04, ROS 2 Rolling from binary   |
| Robotic platform tested on | gazebo simulation of turtlebot3          |

---

use dynamic composition instead of manual composition which exists in Nav2 currently.

* register `SmootherServer` node as component.
* create `container` in `bringup_launch.py`.
* use `LoadComposableNodes` and `ComposableNode` to load Nav2 node in `localization_launch.py` and `navigation_launch.py`.
* remove manual composition.

performance test
| CPU: Intel(R) i7-8700 (6Cores 12Threads), Memory: 32GB | cpu(%) | memory(%) |
| ------------------------------------------------------ | ------ | --------- |
| normal multiple processes                              | 40     | 0.69      |
| manual composition (only for navigation servers)      | 31     | 0.40      |
| dynamic composition (current PR, use `component_container_mt`)     | 39     | 0.23      |

dynamic composition consumes lower memory(saves ~70%),  but consumes the similar cpu as normal multiple processes, due to the bad performance of  `MultiThreadedExecutor`.

---

* there might be some optimizations to be made by using `component_container_isolated` instead of `component_container_mt`, the feature (https://github.com/ros2/rclcpp/pull/1781) was merged, but it isn't available now, we can use this method after the next rolling sync.

#### For Maintainers: 


nav2_bringup/launch/bringup_launch.py
```diff
@@ -108,6 +108,15 @@ def generate_launch_description():
             condition=IfCondition(use_namespace),
             namespace=namespace),
 
+        Node(
+            condition=IfCondition(use_composition),
+            name='nav2_container',
+            package='rclcpp_components',
+            executable='component_container_mt',
```
SteveMacenski, line 115: I thought we didn't want to use the mt container and instead use the isolated one you added?

nav2_bringup/launch/localization_launch.py
```diff
@@ -80,28 +140,15 @@ def generate_launch_description():
...
-
-        Node(
-            package='nav2_lifecycle_manager',
-            executable='lifecycle_manager',
-            name='lifecycle_manager_localization',
-            output='screen',
-            parameters=[{'use_sim_time': use_sim_time},
-                        {'autostart': autostart},
-                        {'node_names': lifecycle_nodes}])
+        DeclareLaunchArgument(
```
SteveMacenski, line 143: Add these declares up top like the format of other files so we keep all the parameterization stuff grouped together

nav2_bringup/launch/navigation_launch.py
```diff
@@ -80,60 +190,15 @@ def generate_launch_description():
...
-
-        Node(
-            package='nav2_lifecycle_manager',
-            executable='lifecycle_manager',
-            name='lifecycle_manager_navigation',
-            output='screen',
-            parameters=[{'use_sim_time': use_sim_time},
-                        {'autostart': autostart},
-                        {'node_names': lifecycle_nodes}]),
+        DeclareLaunchArgument(
```
SteveMacenski, line 193: As well here

SteveMacenski (changes requested):
LGTM, just some small stuff for consistency

nav2_bringup/launch/bringup_launch.py
gezp: yes, but `component_container_isolated` isn't available from binary now, maybe we can wait for next Rolling sync.

gezp:
updated performance test

| CPU: Intel(R) i7-8700 (6Cores 12Threads), Memory: 32GB       | cpu(%) | memory(%) |
| ------------------------------------------------------------ | ------ | --------- |
| normal multiple processes                                    | 44     | 0.76      |
| manual composition ( only for navigation servers)            | 37     | 0.40      |
| dynamic composition (use `component_container_mt`, current PR) | 40     | 0.23      |
| dynamic composition (use `component_container_isolated` , from source code) | 38     | 0.23      |

i re-run all test cases,  the performance of dynamic composition which uses `component_container_isolated`  is the best, which consumes lower memory(saves ~70%),  and lower  cpu (saves ~13%) than normal multiple processes. 

btw, the cpu is measured by `psutil`, and is constantly changing during navigation, so the value of cpu is only approximate.

SteveMacenski:
Interesting, do you know why it is that manual composition isn't as good as dynamic composition? I would have assumed the opposite

SteveMacenski:
Got it, something to note and make sure we find a way around for experiments later for the paper. We should make sure they're as equivalent as possible. e.g. we have a `main` that just manually has the AMCL node in it that we couldn't do in nav2 main branch due to the need to support both SLAM and AMCL in parallel for generalized use. The metrics we capture for the paper should compare composition, manual composition, and dynamic composition of the same identical system.

Please make the changes to the code I requested above and I can approve this and we can merge it in once there's a rolling sync. This is a really big contribution @gezp ! Thanks for sticking with it and making contributions across the ecosystem. This is definitely a long term impactful project.

SteveMacenski:
I believe rolling was updated!

gezp:
yes, i can use `component_container_isolated` from binary on my local computer, i have updated it.

SteveMacenski (approved):

SteveMacenski:
Anything else we need to consider before merging in?

Please submit a PR to the documentation website to the migration guide, somewhere near the top (https://navigation.ros.org/migration/Galactic.html) talking about changing to dynamic composition and the performance improvement it represents (a table or something would be nice showing the performance achieved) .

SteveMacenski:
@gezp I believe all that's blocking this is the documentation PR that should be very short.

gezp:
i created a PR for migration docs in `navigation.ros.org`,  and updated the `README` in `nav2_bringup` (removing the information about manual composition).

SteveMacenski (approved):

SteveMacenski (approved):

Issue #2498, opened 2021-08-10
Revitalization of cost-aware smoother as smoother plugin


robotechvision:
﻿## Feature request

#### Feature description
Although stated that new cost-aware penalty functions push the plans away from high-cost spaces and feasibility checks are performed, we still find some serious problems with the recently updated approach. Most importantly
 * Current implementation of obstacle heuristics and traversal cost computation breaks the A* requirement of admissibility and consistency of heuristic functions. E.g. heuristic can clearly overestimate the cost of reaching the goal because costmap penalty is not included in `getTraversalCost` while it is included in `getObstacleHeuristic`. However I'm afraid that this is not the only problem and returning `motion_table.cost_penalty * normalized_cost` back to `getTraversalCost` will not be sufficient to correct the problem (I can elaborate if requested). Although in some simple test scenarios the results may look better, efficiency and optimality of the search method is not guaranteed anymore and our experiments have shown that the inconsistency problem is not present only at theoretical level but in practice as well.
 * Feasibility checks avoid collisions but path can still be unnecessarily close to collisions. Analytic path search used by the planner can be quite greedy and the resulting path is often dangerous. B-Spline smoother is not able to make it safer. Additionally non cost-aware smoother can itself bring planned path closer to obstacles.
 * Other minor issues (possibly easy to fix, e.g. #2442 )

#### Implementation considerations
During our analysis and experiments with the planner we found solutions which bring more consistency, simplicity and time efficiency to the original smoother, due to lack of which it has been deprecated:
 * We used a simpler ceres::Problem with AutoDiffCostFunctions instead of the original ceres::GradientProblem, avoiding failures on wolfe conditions and complicated manual definitions of jacobians
 * We added `ceres::BiCubicInterpolator<ceres::Grid2D<u_char>>` for costmap data since ceres requires continuous differentiable functions
 * Instead of optimizing the whole path at once we optimize a short fixed-length section around the robot (e.g. 4 meters behind and 6 meters in front of the robot). Thence the optimization is performed in a separate behavior tree node which is called periodically (e.g. every 2 meters)
 * Additionally we updated smoothing and path poses orientation computation to be aware of motion direction changes (forward/reverse)
[...]

PR #2753, opened 2021-12-22
Ceres smoother

robotechvision:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2498 |
| Robotic platform tested on | (hardware RTV Caster, gazebo simulation of RTV Caster) |

---

* Original Ceres smoother from Smac planner put into operational state by simplifying to AutoDiff and fixing errors
* Added support for forward/reverse motion (e.g. Reeds-Shepp model)
* Added cubic interpolator for costmap cost
* Curvature cost simplified and generalized to be valid for segments with non-equal lengths
* Added option to add extra costmap cost near cusps (forward/reverse direction changes) which are difficult to be smoothed properly using basic cost parameters
* Added option to read costmap cost from different points of robot, which can be useful for arbitrary robot shapes (e.g. diff-drive with caster wheels has footprint offset backwards against rotation center)
* Added option to downsample path prior to smoothing (for speedup) and upsample back to original (or even higher) density using cubic bezier
* Detailed tests written to check for smoother attributes to be maintained when optimizing/modifying smoother in the future
* Added TruncatePathLocal BT action to pick a section from the global plan so that smoothing can be performed with a good quality in a reasonable time (can be used also for other purposes)

Note: We will check for reviews soonest on 2022-01-10 so no need to hurry with them. Merry Christmas

* Added new smoother plugin (CeresCostawareSmoother), it should be added to docs
* Added new BT plugin (TruncatePathLocal), it should be added to docs

---

* Try to find an optimizer configuration which converges faster
* Try to write down proper derivatives which can speed up optimization significantly, maintaining (or improving) smoothing quality (see tests)

#### For Maintainers: 


nav2_ceres_costaware_smoother/package.xml
```diff
@@ -0,0 +1,30 @@
...
+  <depend>libceres-dev</depend>
+
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+    <!-- <nav2_core plugin="${prefix}/nav2_ceres_costaware_smoother.xml" /> -->
```
SteveMacenski, line 28: This should be added

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/truncate_path_local_action.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "behaviortree_cpp_v3/action_node.h"
+#include "tf2_ros/buffer.h"
+
+namespace nav2_behavior_tree
+{
+
+/**
+ * @brief A BT::ActionNodeBase to shorten path by some distance
+ */
+class TruncatePathLocal : public BT::ActionNodeBase
```
SteveMacenski, line 33: Add the new node to the list https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/nav2_tree_nodes.xml so that users can use Groot with it

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/ceres_costaware_smoother.hpp
```diff
@@ -0,0 +1,104 @@
...
+    nav_msgs::msg::Path & path,
+    const rclcpp::Duration & max_time) override;
+
+protected:
+  std::shared_ptr<tf2_ros::Buffer> tf_;
+  std::string plugin_name_;
+  std::shared_ptr<nav2_costmap_2d::CostmapSubscriber> costmap_sub_;
+  rclcpp::Logger logger_ {rclcpp::get_logger("CeresCostawareSmoother")};
+
+  std::unique_ptr<nav2_ceres_costaware_smoother::Smoother> _smoother;
```
SteveMacenski, line 97: Make these 3 last variables have a trailing underscores (e.g. `smoother_`) like the previous 4 class members

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother.hpp
```diff
@@ -0,0 +1,377 @@
...
+    a[0] = p3[0] - p0[0] - c[0] - b[0];
+    a[1] = p3[1] - p0[1] - c[1] - b[1];
+
+    p[0] = a[0] * mu * mu * mu + b[0] * mu * mu + c[0] * mu + p0[0];
+    p[1] = a[1] * mu * mu * mu + b[1] * mu * mu + c[1] * mu + p0[1];
+
+    return p;
+  }
+
+  bool _debug;
```
SteveMacenski, line 370: Change to trailing underscore so consistent within the package across all files

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
```diff
@@ -0,0 +1,284 @@
...
+#include "Eigen/Core"
+#include "nav2_costmap_2d/costmap_2d.hpp"
+#include "nav2_ceres_costaware_smoother/options.hpp"
+#include "nav2_ceres_costaware_smoother/utils.hpp"
+
+namespace nav2_ceres_costaware_smoother
+{
+
+template<typename T>
+struct DoubleOnlyLogger
```
SteveMacenski, line 38: This doesn't appear to be used, can this be removed?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
```diff
@@ -0,0 +1,284 @@
...
+          (T)_costmap_resolution;
+        T value;
+        _costmap_interpolator->Evaluate(interp_pos[1] - (T)0.5, interp_pos[0] - (T)0.5, &value);
+
+        r += (T)weight * (T)_params.cost_check_points[i + 2] * value * value;
+      }
+    }
+  }
+
+  const Eigen::Vector2d _original_pos;
```
SteveMacenski, line 272: trailing underscore please

SteveMacenski: Actually, you can just make the small number of `variable_`'s into `_variables`'s, it'll be easier for you than to change all of these all over the place for just 3

SteveMacenski (commented):
I was out on PTO until today and heading to a  trade show this week. I wanted to do a quick once over before I left to give you feedback on the code-related things. I'll give this a deep dive when I get back on the functional bits. 

Overall, it looks good though. 

Can we brainstorm a few different name options? `ceres` is just the library we used to implement the smoother, but we could have used one of many, so its not a very good descriptor to the method itself. 

`nav2_cost_aware_smoother`? Something else? 

Also, I see you used dense QR, these are the options I found that gave the best runtime performance way back when https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/test/deprecated/smoother.hpp#L62-L65. How did you select these options for default? Did you play around with other options? I found that changing up these methods had a big impact on runtime performance and very little change in overall outcome. The smoothing problem will naturally have one (or more) of the techniques its best suited for. Just wanting to make sure we selected the best option, since I did notice sometimes up to 10x speed ups or more with some method testing

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/ceres_costaware_smoother.hpp
SteveMacenski: Actually, you can just make the small number of variable_'s into _variables's, it'll be easier for you than to change all of these all over the place for just 3

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother.hpp
SteveMacenski: Actually, you can just make the small number of variable_'s into _variables's, it'll be easier for you than to change all of these all over the place for just 3

SteveMacenski:
@robotechvision just checking in on this!

Steve

afrixs:
Hi, I'm back from PTO too, thank you for your comments.

Yeah, I didn't like the name too but haven't spent much time until now figuring a better one. I think there are multiple constraints/criteria currently (and potentially more to be added in future) so `nav2_cost_curvature_smoother` would be incomplete. `nav2_cost_aware_smoother` expresses its major advantage compared to other current smoothers but it can lose its meaning when other cost-aware smoothers are implemented. I think `nav2_constrained_smoother` is more appropriate but maybe 'criteria/path qualities' would be more exact than 'constraints'. I also think that problem optimization could be mentioned in the name since it is basic method the smoother uses. So I would propose something like `nav2[_path][_quality|_criteria]_optimizing_smoother`.

On the solver type: I will need some time to dive back into it to check but I remember the line search method kept failing in many situations, alerting that Wolfe Constraints are being broken. After several attempts to fix the failures I decided to start over from the most simple and least coding/math error-prone point - the AutoDiff cost functions. As far as I know the line-search problem solvers don't support AutoDiff so I needed to change the solver. I remember I played with multiple options and this one worked the best from what I tried but certainly I didn't try every available solver configuration, that's why I added the performance-optimization tasks into **Future work that may be required**. However, when used properly (on a truncated path, using path downsampling and a reasonable `max_iterations` parameter), the smoother can be used in its current form for real-time navigation (computation time < 0.15s) + there have already been demand by users for a cost-aware smoother, so I think that it's OK to select a good enough option for now and leave finding the best one for future (frankly, currently it's not among our company's priorities so I don't know when I'll find time for it).

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
```diff
@@ -0,0 +1,252 @@
...
+   * @param original_path Original position of the path node
+   * @param next_to_last_length_ratio Ratio of next path segment compared to previous.
+   *  Negative if one of them represents reversing motion.
+   * @param reversing Whether the path segment after this node represents reversing motion.
+   * @param costmap A costmap to get values for collision and obstacle avoidance
+   * @param params Optimization weights and parameters
+   * @param costmap_weight Costmap cost weight. Can be params.costmap_weight or params.cusp_costmap_weight
+   */
+  SmootherCostFunction(
+    const Eigen::Vector2d & _original_pos,
```
SteveMacenski, line 56: No underscore in the constructor argument

SteveMacenski:
Hi,

Finally, I've cleared enough from my plate to be able to give some serious thought in this PR. I sincerely apologize for the delay. The Smac Planner work took priority I think on both of our sides then I got pulled into a bunch of papers, then working on some of the 2022 goals. I've made enough headway I can start to iterate with you on this much more quickly.

I still have some open comments from January that should be dealt with on styling you can see above. 

On naming: I like `nav2_constrained_smoother` because its abstract enough and it is a constrained optimization problem (albeit, soft constraints, not hard constraints, but that's common). The cost-awareness is a contraint in the optimization problem and so is the path curvature. Conceptually (and I'm **not** suggesting we do this) we could make the cost functions plugins so that different constraints could be added by users. Just adding `optimizer` or `optimizing` isn't very descriptive to much else, but its not a bad qualifier if we had others in it as well. I just don't want to make something that ends up with too long of a name. `Ceres` definitely shouldn't appear in it though, since that's just  the library of implementation. 

On solver types: Let me look at this again now that I have fresh eyes. If nothing else, it may just be worth making it parameterized. 

A more detailed look / comments will follow this comment.

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+bool CeresCostawareSmoother::smooth(nav_msgs::msg::Path & path, const rclcpp::Duration & max_time)
+{
+  if (path.poses.size() < 2) {
+    return true;
+  }
+
+  auto costmap = costmap_sub_->getCostmap();
+  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> lock(*(costmap->getMutex()));
```
SteveMacenski, line 95: There's no need to lock this since there is no one else using the costmap (or updating it in the background, since there are no costmap plugin layers at play in this server).

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
```diff
@@ -0,0 +1,156 @@
...
+      if (i == 0) {
+        start_dir = orientation;
+      }
+    }
+  }
+
+  _smoother_params.max_time = max_time.seconds();
+
+  // Smooth plan
+  if (!_smoother->smooth(path_world, start_dir, end_dir, &*costmap, _smoother_params)) {
```
SteveMacenski, line 125: I think `costmap.get()` would be the same effect and more clear

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
```diff
@@ -0,0 +1,156 @@
...
+  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> lock(*(costmap->getMutex()));
+
+  // populate smoother input with (x, y, forward/reverse dir)
+  std::vector<Eigen::Vector3d> path_world;
+  path_world.reserve(path.poses.size());
+  Eigen::Vector2d start_dir;
+  Eigen::Vector2d end_dir;
+  for (size_t i = 0; i < path.poses.size(); i++) {
+    auto & pose = path.poses[i].pose;
+    tf2::Quaternion q;
```
SteveMacenski, line 104: This can be significantly simplified using tf2's `getYaw()` method. If you give it a quaternion, it will spit out the angle theta of the Z rotation for mobile robot needs. Example: https://github.com/ros-planning/navigation2/blob/fee433ff76b1fe2997043bec16b81568fbead49e/nav2_recoveries/plugins/spin.cpp#L87

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
```diff
@@ -0,0 +1,156 @@
...
+    double angle = q.getAngle() * sign(q.getZ());
+    Eigen::Vector2d orientation(cos(angle), sin(angle));
+    if (i == path.poses.size() - 1) {
+      path_world.emplace_back(pose.position.x, pose.position.y, path_world.back()[2]);
+      end_dir = orientation;
+    } else {
+      auto & pos_next = path.poses[i + 1].pose.position;
+      Eigen::Vector2d mvmt(pos_next.x - pose.position.x, pos_next.y - pose.position.y);
+      bool reversing = _smoother_params.reversing_enabled && orientation.dot(mvmt) < 0;
+      path_world.emplace_back(pose.position.x, pose.position.y, reversing ? -1 : 1);
```
SteveMacenski, line 115: The reversing logic is a little bit convoluted, it would be simpler to do something like what we do here: https://github.com/ros-planning/navigation2/blob/fee433ff76b1fe2997043bec16b81568fbead49e/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L646-L673 to use the dot product of the robot poses to find cusp point of directional changes.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/utils.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+template<typename T>
+inline int sign(T a)
+{
+  return a == 0 ? 0 : a > 0 ? 1 : -1;
+}
+
+template<typename T>
+inline Eigen::Matrix<T, 2, 1> arcCenter(
+  Eigen::Matrix<T, 2, 1> p1,
```
SteveMacenski, line 34: Naming these variables a little more meaningfully would help alot in the understanding of the code. I think this is point i+1, point i, and point i - 1, but the actual names don't make that clear

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
```diff
@@ -0,0 +1,252 @@
...
+  {
+    return new ceres::AutoDiffCostFunction<SmootherCostFunction, 4, 2, 2, 2>(this);
+  }
+
+  void setCostmapWeight(double costmap_weight)
+  {
+    _costmap_weight = costmap_weight;
+  }
+
+  double costmapWeight()
```
SteveMacenski, line 84: `getCostmapWeight()` would be nice

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
```diff
@@ -0,0 +1,252 @@
...
+  inline void addCostResidual(
+    const double & weight,
+    const Eigen::Matrix<T, 2, 1> & pt,
+    const Eigen::Matrix<T, 2, 1> & pt_p1,
+    const Eigen::Matrix<T, 2, 1> & pt_m1,
+    T & r) const
+  {
+    if (_params.cost_check_points.empty()) {
+      Eigen::Matrix<T, 2,
+        1> interp_pos = (pt - _costmap_origin.template cast<T>()) / (T)_costmap_resolution;
```
SteveMacenski, line 209: The line breaks here are odd, why not keep all the type on 1 line and use the line extension after the `=`?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
```diff
@@ -0,0 +1,252 @@
...
+      T value;
+      _costmap_interpolator->Evaluate(interp_pos[1] - (T)0.5, interp_pos[0] - (T)0.5, &value);
+      r += (T)weight * value * value;  // objective function value
+    } else {
+      Eigen::Matrix<T, 2, 1> dir = tangentDir(
```
SteveMacenski, line 214: Can you explain a bit what's going on with these "cost check points"?

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "nav2_util/robot_utils.hpp"
+#include "nav_msgs/msg/path.hpp"
+#include "tf2_ros/create_timer_ros.h"
+
+#include "nav2_behavior_tree/plugins/action/truncate_path_local_action.hpp"
+
+namespace nav2_behavior_tree
+{
+
+TruncatePathLocal::TruncatePathLocal(
```
SteveMacenski, line 32: Does this need to be in this same PR? It would be good to discuss this in another PR just to keep all of this focused on the smoother. I'm not 100% sure why this is needed, but it would be good to review / discuss the need separately

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
```diff
@@ -0,0 +1,169 @@
...
+    node->get_parameter(name + ".minimum_turning_radius", minimum_turning_radius);
+    max_curvature = 1.0f / minimum_turning_radius;
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "w_curve", rclcpp::ParameterValue(30.0));
+    node->get_parameter(local_name + "w_curve", curvature_weight);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "w_cost", rclcpp::ParameterValue(0.015));
+    node->get_parameter(local_name + "w_cost", costmap_weight);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "w_cost_cusp", rclcpp::ParameterValue(0.04));
```
SteveMacenski, line 62: Can you explain this weight a bit, why its needed, its effect, and where it is applied?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
```diff
@@ -0,0 +1,169 @@
...
+    node->get_parameter(local_name + "w_cost_cusp", cusp_costmap_weight);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "cusp_zone_length", rclcpp::ParameterValue(-1.0));
```
SteveMacenski, line 65: Ditto

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
```diff
@@ -0,0 +1,169 @@
...
+      node, local_name + "input_downsampling_factor", rclcpp::ParameterValue(1));
+    node->get_parameter(local_name + "input_downsampling_factor", input_downsampling_factor);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "output_upsampling_factor", rclcpp::ParameterValue(1));
+    node->get_parameter(local_name + "output_upsampling_factor", output_upsampling_factor);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "reversing_enabled", rclcpp::ParameterValue(true));
+    node->get_parameter(local_name + "reversing_enabled", reversing_enabled);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "keep_goal_orientation", rclcpp::ParameterValue(true));
```
SteveMacenski, line 86: What situation would we not want to keep the goal / start orientation?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
```diff
@@ -0,0 +1,169 @@
...
+    node->get_parameter(local_name + "w_dist", distance_weight);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "w_smooth", rclcpp::ParameterValue(15000.0));
+    node->get_parameter(local_name + "w_smooth", smooth_weight);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "cost_check_points", rclcpp::ParameterValue(std::vector<double>()));
+    node->get_parameter(local_name + "cost_check_points", cost_check_points);
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "input_downsampling_factor", rclcpp::ParameterValue(1));
+    node->get_parameter(local_name + "input_downsampling_factor", input_downsampling_factor);
```
SteveMacenski, line 78: How about `path_downsampling_factor` the `input` bit is a little confusing.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
```diff
@@ -0,0 +1,169 @@
...
+    nav2_util::declare_parameter_if_not_declared(
+      node, local_name + "output_upsampling_factor", rclcpp::ParameterValue(1));
+    node->get_parameter(local_name + "output_upsampling_factor", output_upsampling_factor);
```
SteveMacenski, line 81: How about `path_upsampling_factor` for the same reasons

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother.hpp
```diff
@@ -0,0 +1,396 @@
...
+
+  /**
+   * @brief Initialization of the smoother
+   * @param params OptimizerParam struct
+   */
+  void initialize(const OptimizerParams params)
+  {
+    _debug = params.debug;
+
+    _options.linear_solver_type = ceres::DENSE_QR;
```
SteveMacenski, line 63: How does `SPARSE_NORMAL_CHOLESKY` perform? I know some settings are so bad that its not worth exposing them to users, but just checking if we should strictly prevent users from selecting another legitimate option

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother.hpp
```diff
@@ -0,0 +1,396 @@
...
+   */
+  bool smooth(
+    std::vector<Eigen::Vector3d> & path,
+    const Eigen::Vector2d & start_dir,
+    const Eigen::Vector2d & end_dir,
+    const nav2_costmap_2d::Costmap2D * costmap,
+    const SmootherParams & params)
+  {
+    // Path has always at least 2 points
+    CHECK(path.size() >= 2) << "Path must have at least 2 points";
```
SteveMacenski, line 96: Should we make this `throw` instead? Will the main `nav2_smoother` server catch if the algorithm throws during execution?

SteveMacenski (commented):
Overall, LGTM, I skimmed over `[nav2_ceres_costaware_smoother/smoother.hpp](https://github.com/ros-planning/navigation2/pull/2753/files#diff-d237e6158151d23832965fb29cd112aa21aa3f09a966907ab7b1aca076a56a46)` since alot of that logic is related to the cusp handling I requested some clarification regarding. Not worth trying to pull that apart without more context.

But otherwise, I went through the rest of the files and except where noted, this is really nice looking and not much I would ask for change -- mostly superficial.

Can you also mention in the readme a few examples of paths being smoothed with the costmap being shown & runtime metrics alongside that?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
afrixs: this was done by uncrustify

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
afrixs: e.g. when smoothing a path from SmacPlanner2D when goal checker's `goal_tolerance_yaw` is `.inf`. I.e. when we don't care about current nor final robot orientation, we just want the path to be smooth and far from obstacles. Sticking on the orientations can introduce minor artifacts to final path.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
SteveMacenski: Sometimes it picks odd locations to make the break. It would also be acceptable by the formatting to do `Eigen::Matrix<T, 2,1> interp_pos =` on one line and have the rest be the line extension. It doesn't do well with templates / commas for breaking lines up.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
SteveMacenski: Would it be reasonable to restrict this smoother to only kinematically aware plans, given the curvature constraint term? If a plan was generated via another method, I'm not entirely sure how reliably this could / would create a path that followed that constraint if it is wildly deviating from it in the first place (or in some cases where the path itself went through a space where it is impossible to have the curvature constraint followed).

If it can be applied elsewhere, I think that's a reasonable point to make -- but for feasible paths its not possible to create said path without orientations being meaningful for obvious reasons. 

> when smoothing a path from SmacPlanner2D... Sticking on the orientations can introduce minor artifacts to final path.

This would be a good point to make in documentation

SteveMacenski: For reference, I'm working on pulling out the current simplified smoother from Smac and making a new plugin that would be able to support other classes of paths where we don't care about the curvature constraints, if that changes your view on this at all.

SteveMacenski:
I merged your docs PR, I'll fix the remaining items from the last review and add in the "simple" smoother plugin from https://github.com/ros-planning/navigation2/pull/2875, which is just a port from the simple Smac smoother into this plugin form.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
afrixs: I see no problem with the curvature constraint - it can be set to zero for "infeasible" planner inputs (or `minimum_turning_radius` can be set to zero with the same effect). I can add a test to check for an edge scenario (path with positions 0,0 -> 1,0 -> 0,0) whether some harmful `Inf`s or `NaN`s are produced, but even if it happened, it would be an easy-to-fix issue, so I think that narrowing the smoother to feasible inputs only is not required (a note about setting the parameters to zero can be added to docs).

A more difficult problem (however solvable - I have ideas for that but that's for another PR) would be an input path with multiple poses at one location (like the inplace rotations generated by diff-drive/omni `SmacPlannerLattice` planners). This can be mentioned in docs until it is resolved.

afrixs: However the 2D input was just an example, even with a feasible path producing planner you don't always need to finish with a specific orientation. The smoother can be used to take care of that and smooth out the final rotation maneuver.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
afrixs: It is a compensation for not using a `footprint` polygon. Simply said: the robot can pick the cost from other places than its `base_link`. E.g. a diff drive robot with 2 actuated wheels in front and 2 caster wheels in back has its `base_link` frame between the front wheels, however an effective cost center is in the middle of the robot.

However (as noted [here](https://github.com/robotechvision/navigation2/blob/bd4ec4c87eb0c4383a35736171c0841b4dda84f9/nav2_ceres_costaware_smoother/README.md?plain=1#L25)), it's quite hard to make profit of it in a reasonable time, but since the "reasonable time" is subjective (depends on use, robot, CPU, situation, other parameters), I think it's appropriate to keep the option. The parameter could be marked as `experimental.cost_check_points` if the note is not expressive enough.

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
afrixs: It can be discussed after closing the smoother reviews (although I would keep it in the same PR if it's not a problem - that would be less work for me). Just shortly about the need: the smoother really needs an upper bound for the path length to provide good results in a reasonable time and this simple BT node takes care of that.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
afrixs: Let the [tests](https://github.com/robotechvision/navigation2/blob/bd4ec4c87eb0c4383a35736171c0841b4dda84f9/nav2_ceres_costaware_smoother/test/test_costaware_smoother.cpp#L721) speak for me here. Since the cusps are usually surrounded by rotations, the optimizer tends to stick in a local minima, not moving the cusp point away from obstacles. A higher cost value is needed to help him out.

This picture (cyan - input path, purple - path smoothed with no cusp zone/cost, green - path smoothed with a higher cusp cost) shows how wobbliness is eliminated and cusp is pushed into a safer place when `w_cost_cusp` and `cusp_zone_length` are used.
image

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
afrixs: It has been empirically found that the cost around cusp needs to increase gradually (linearly) towards the cusp. `cusp_zone_length` parameter defines the area (and thus the slope as well) of increase

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
afrixs: It is a little bit more convoluted but much more robust. Frankly, I know this piece of code and have had bad experiences with it. (Haven't raised an issue though yet, I preferred returning back to our own controller plugin, since there was more than this to change in the controller to suit our needs).

If you think about it, a path cusp is a point where a forward movement direction changes to backward (that's what our implementation checks for). Although usually, but not necessarily (e.g. when sampling is not dense enough), it shows a >90° break in the path (that's what the RPPC implementation checks for). I understand that our attitude requires orientations to be assigned in the input path (or `reversing_enabled` parameter to be turned off) but that's what usually planners which can produce backward motions do provide, so eliminating the requirement is not worth stepping back with the robustness.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
SteveMacenski: OK - we can put a pin in this item, I don't think it needs to be changed, I just need to keep the note here so I can go back through and really understand this in detail.

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
SteveMacenski: I figured that would be something we could add to the SmoothPath action API to add the max length to send to the smoother. That seems like a more natural fit, no?

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
SteveMacenski: OK - I think more than the actual algorithm used its the lack of standard tools used to implement it. More than I have an issue with what this code does, my concern is that the meaning / semantics of what you're trying to accomplish are lost to the passive reader. 

```
    tf2::Quaternion q;
    tf2::fromMsg(pose.orientation, q);
    double angle = q.getAngle() * sign(q.getZ());
    Eigen::Vector2d orientation(cos(angle), sin(angle));
    ...
    orientation.dot(mvmt)
    reversing ? -1 : 1
```
Can this be simplified / use more standard tools (or add comments?) so that the intention of what's happening is clear?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
SteveMacenski: I suppose my question is: if given a circular robot, obviously you can just use the center point and check if its >= inscribed cost. For a non-circular robot, we could make the smoother server give a footprint subscriber object to the smoother algorithm so it can directly access the footprint itself. 

Would it not be more correct to use that, or is this setup because using the footprint itself is too expensive? If the case is that its too expensive to check each point, I can understand that and it would make sense to add a `simplified_footprint_polygon` parameter or something to specify that with the note. 

Also, why weight different points differently? Why not simply take the maximum cost of the points sampled of the footprint?

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
SteveMacenski: Got it, this would be good graphic / explanation to add to the readme of the package in explaining those parameters. I figured as much was true, but this is a  different way than I would have gone about it - works though.

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
afrixs: I think you are looking at an outdated commit... the tf2::getYaw is currently used there (reason I used getAngle and getZ was that this was one of the first things I did in ROS2 so I didn't know about all the misc utility methods, so I just sticked to class member methods, which were all at one place, considering that planner output paths are 2D...)

Adding clarifying comments to other mentioned operations...

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
afrixs: > On the parameters, the optimization_length and optimization_length_backwards should probably be parameters of your smoother plugin rather than attributes of the server (since not every smoother will have that option / make use of it).

I'm sorry but we have already discussed that and these are your words from Slack. You asked me to remove the max length from the smoother server and we agreed upon adding the truncation BT node.

In addition, what profit would be from gluing two independent useful lego parts into one? The node can be used in BT for various other tasks. Actually we use it in our robots for further collision checking in front of robot, i.e. something like this:

```
                <TruncatePathLocal input_path="{path_coarse}" output_path="{path_coarse_local}"
                  distance_forward="8.0" distance_backward="4.0" robot_frame="base_footprint"/>
                <SmoothPath unsmoothed_path="{path_coarse_local}" smoothed_path="{smoothed_path}" smoother_id="SmoothPath" check_for_collisions="false"/>
                <TruncatePathLocal input_path="{smoothed_path}" output_path="{smoothed_path_for_check}"
                  distance_forward="3.5" distance_backward="0.0" robot_frame="base_footprint"/>
                <CheckPathValidity path_in="{smoothed_path_for_check}"/>
```

Note: SmoothPath with a VoidSmoother plugin (doing nothing with the path) and `check_for_collisions="true"`  can be used instead of CheckPathValidity (the name was used just for readability)

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
afrixs: As you guessed, -1 means "disabled", i.e. all points use `w_cost` only. My decision was to disable it so that during tuning users first focus on `w_cost` and enable the cusp handling afterwards, choosing a reasonable `w_cost_cusp` value, relatively to their tuned `w_cost`. The cusp cost is enabled in the `README.md` example, I just wanted to spare users who tune from scratch (and don't read whole docs) from confusions like "I turned `w_cost` totally off, why the path is still being pushed away from obstacles at some places?" (note: it can also happen that `w_cost_cusp` remains lower than `w_cost` if not adjusted properly, again, causing confusions).

Just telling my thread of thoughts, I have no problem to change the default if you decide.

afrixs: Just came to mind, that a `w_cost_cusp_multiplier` could be used instead of the absolute parameter. This would prevent confusions and thus the cusp cost could be enabled by default with no problems.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother.hpp
afrixs: Seems like it is quite usable. For now, I just ran tests with `SPARSE_NORMAL_CHOLESKY` (not trying to tune) and it seems to perform similarly in most tests, and even better (and much faster) in the cusp tests. I can add a parameter to pick between these two solver types and make `SPARSE_NORMAL_CHOLESKY` default

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
afrixs: Updated the picture with shorter arrow lengths, so now there should be less clutter. The purple path is continuous, but apparently worse than the green one (note: `w_curve` was set to a very low value here to make other parameters show up, so turning radius is not relevant here).

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
afrixs: > Rename to simplified footprint, if I'm correctly understanding that this is what this is

It is more of footprint centroid than the actual footprint boundary points.

> Use normal footprint given by the smoother server, unless that's too computational

I wanted to let users pick the point (or theoretically multiple points in case of a very bizarre robot shape), but in practice we could simplify the usage and compute centroid of the footprint polygon.

> Why not use the max cost of all the points, versus weighting them?

If we choose to compute the centroid ourselves, then this is irrelevant because there would be just one point. If not, then I would go with the weighted sum. The reason is simply a parable to physics: since the obstacle repelling can be understood as a pressure applied on a part of the robot, the pressure results in a force (higher force for larger parts - that's why the weights are there) and `sum` is used instead of `max` when applying multiple forces on an object. Note: the forces are applied on different points, this is not included in this parable, but is "detected and considered" by the optimizer (a separate residual field for each centroid would maybe help the optimizer, but frankly, I haven't paid too much effort in tuning the optimizer for these bizarre-shaped multi-centroid robots)

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
SteveMacenski: You are correct, yes this is much more clear now, thank you.

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
SteveMacenski: Fair enough, I'll go with you on that then

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
SteveMacenski: Ah I see, so its more like below, where the footprint is the polygon, the points are the check points, and the radius is the inflation lethal radius: 

image

Conceptually if it was less extreme, you could use some single non-`(0,0)` point that is the centroid of the polygon, but I think in practice if the center of the robot isn't good (circular, square), you probably would need to be working with a few points (rectangular, odd shaped) to deal with the edges. 

I think then the name / process makes sense. You do a good job warning the user about potential side effects but it might be nice to add a sentence or two about what these represent more explicitly. For example, I'd remove the word "footprint" from any explanation and add a blob about how these are a minimal set of points around the robot to check for collision at. A diagram would be illustrative.

If we sum them all would this term blow out all the others, relative to point costs? Though I agree with your explanation that max probably doesn't do the job well. Would averaging the weighted costs of multiple points have the desired outcome? That may or may not help with the local minima issue you describe. Just a suggestion to try, if it doesn't work as well, we can stick with this.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
SteveMacenski: Got it - OK I like this behavior and this makes sense.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/options.hpp
SteveMacenski: I like that idea - a multiple on the cost parameter to apply at cusps. 

I think it should also be defaulted to "on" as long as the defaults you provide in the source code are "reasonable" for out of the box users. If folks don't need to change the parameters (or change them very much) its probably not going to be come an issue. If users want to tune from absolute scratch, then this isn't the only package they're going to get wacky results in if they don't understand what the dials do.

SteveMacenski:
OK - we can do the BT node at the same time, reviewed that and went deeper into the code after you explained the cusp stuff to me

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
afrixs: Yes, it's exactly like you understand it now. I'm removing the "footprint" word from the comment and adding a link to the configuration where a detailed clarification with a diagram will reside. Btw your diagram made me realize that even a (not so bizarre) rectangular shape can yield significantly from having multiple cost check points. Test result was improved by 1.6% when using 2 points (front/back) instead of one in the middle. Thank you!

An average is just a sum divided by a constant, so it will not solve the problem (however using a faster `SPARSE_NORMAL_CHOLESKY` solver enables the cost check points to give good results in < 150ms time, so it is less of a problem now). But if you consider `avg` to be better here than `sum`, I have no problem to change it. Currently when you use e.g. `w_cost: 0.02` and three cost check points with weights 1, 2 and 1, the total cost weight is effectively `(1+2+1)*0.02 = 0.08`. Using `avg` (i.e. automatically normalizing weights so that their sum equals 1), could separate the problem of tuning `w_cost` against other costs and problem of "defining" a good robot shape.

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/utils.hpp
```diff
@@ -0,0 +1,114 @@
...
+ * @result Tangential line direction.
+ * Note: the sign of tangentDir is undefined here, should be assigned in post-process
+ * depending on movement direction. Also, for speed reasons, direction vector is not normalized.
+ */
+template<typename T>
+inline Eigen::Matrix<T, 2, 1> tangentDir(
+  Eigen::Matrix<T, 2, 1> pt_prev,
+  Eigen::Matrix<T, 2, 1> pt,
+  Eigen::Matrix<T, 2, 1> pt_next,
+  int forced_dot_sign = 0)
```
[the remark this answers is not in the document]
afrixs: Yes, it seems like historically all the calls have been replaced by the "informed" ones

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
```diff
@@ -0,0 +1,158 @@
...
+  // populate smoother input with (x, y, forward/reverse dir)
+  std::vector<Eigen::Vector3d> path_world;
+  path_world.reserve(path.poses.size());
+  Eigen::Vector2d start_dir;
+  Eigen::Vector2d end_dir;
+  for (size_t i = 0; i < path.poses.size(); i++) {
+    auto & pose = path.poses[i].pose;
+    double angle = tf2::getYaw(pose.orientation);
+    Eigen::Vector2d orientation(cos(angle), sin(angle));
+    if (i == path.poses.size() - 1) {
```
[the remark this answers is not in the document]
afrixs: These situations would not be a problem. The `reversing` actually speaks about the direction of the segment _after_ the point. There is no segment after the last point (or at least smoother has no information about it), so the `reversing` of the last point is irrelevant. Actually I think it wouldn't be a problem to keep the value undefined, but the "undefined values" alarm inside me led me to add those lines which pick the less problematic `reversing` value (one without a cusp). Can add a comment about it.

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
```diff
@@ -0,0 +1,158 @@
...
+bool CeresCostawareSmoother::smooth(nav_msgs::msg::Path & path, const rclcpp::Duration & max_time)
+{
+  if (path.poses.size() < 2) {
+    return true;
+  }
+
+  // populate smoother input with (x, y, forward/reverse dir)
+  std::vector<Eigen::Vector3d> path_world;
+  path_world.reserve(path.poses.size());
+  Eigen::Vector2d start_dir;
```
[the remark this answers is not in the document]
afrixs: The orientations of other poses are going to be changed during optimization anyways, so the smoother forgets them and then extract them from the path shape. However start/goal segments are not going to be changed (unless requested by `keep_start/goal_orientation` params), so the smoother can keep this (often quite valuable) information and use it in the output paths. E.g. the goal orientation defined by the last path segment direction can differ slightly (or quite much e.g. in the case of 2D planners) from the one requested by user which could cause robot to finish with an undesired orientation

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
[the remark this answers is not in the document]
SteveMacenski: A comment would help :+1:

nav2_ceres_costaware_smoother/src/ceres_costaware_smoother.cpp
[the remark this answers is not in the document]
SteveMacenski: Good point - a comment to that effect in the code would be appreciated

nav2_ceres_costaware_smoother/include/nav2_ceres_costaware_smoother/smoother_cost_function.hpp
SteveMacenski: No worries, just trying to understand! Sometimes an outside perspective can help (I know it does for me).

I leave `avg` vs `max` up to your best judgement and results, I'm just offering the suggestion that it might give you the behavior you'd like but balance better with other terms. I think the tuning problem is a good point but I wouldn't want to do it if it comes at the expense of performance / path smoothness away from obstacles. If we do leave it as `max` a note about retuning may be necessary when changing the number of these points would be good.

SteveMacenski:
Glancing over the open review comments, there aren't many really significant change requests - mostly just documentation / clarification in comments and a few little changes based on utilities we have in `nav2_utils` that would make it easier to understand / use the same methods for similar actions. Great work on improving this! I really wanted to get this with analytical derivatives, but I suppose as you found as well it was difficult to get that to work well.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/truncate_path_local_action.hpp
```diff
@@ -0,0 +1,99 @@
...
+        "robot_frame", "base_link",
+        "Robot base frame id"),
+      BT::InputPort<double>(
+        "transform_tolerance", 0.2,
+        "Transform lookup tolerance"),
+      BT::InputPort<geometry_msgs::msg::PoseStamped>(
+        "pose", "Manually specified pose to be used"
+        "if overriding current robot pose"),
+      BT::InputPort<double>(
+        "angular_distance_weight", 0.2,
```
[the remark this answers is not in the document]
afrixs: This attitude could be quite helpful, but it would require to use the path pruning which is performed by the controllers, i.e. assigning whole path to a field when a goal is received and then trimming it step-by-step. As opposed to controller server/plugins, this node is not an action server, it's not even stateful, it simply gets the current path from the blackboard every time and returns a trimmed portion of it.

Using `angular_distance_weight` has solved the loop problem for us, making it possible to keep the node that simple. However, if required, it is possible to mimic the stateful behavior of RPP/DWB. I'd prefer the behavior as an option though, since non-statefulness has its advantages too (no risk of trimming off something that should not be trimmed in case of a temporal anomaly in the inputs, without a chance to return it back), i.e. to use the current non-stateful behavior if `max_robot_pose_search_dist` is `.inf`.

So if decided to implement the path pruning and `max_robot_pose_search_dist`, my idea would be to keep track of
 * `last_robot_pose_index_` to be used as a `begin` for computing the `closest_pose_upper_bound` (this would mimic the path pruning)
 * keep track of `last_path_` to compare it against current `path`. If the path has changed, reset `last_robot_pose_index_` to 0 (this mimics the [`setPlan()`](https://github.com/ros-planning/navigation2/blob/4234a6ea27de2094dde98a5f928522d72b139227/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L631) method of RPP/DWB which basically resets all trimming)

It's not a 100% elegant solution, so it's worth considering, but should be quite easy to implement.

SteveMacenski:
The remaining comments are w.r.t. documentation for the website / readme and the Truncate BT node. The smoother is A-OK with me now.

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
```diff
@@ -0,0 +1,147 @@
...
+    path.poses.begin(), path.poses.end(),
+    [&pose,
+    angular_distance_weight](const geometry_msgs::msg::PoseStamped & ps) {
+      return poseDistance(pose, ps, angular_distance_weight);
+    });
+
+  // expand forwards to extract desired length
+  double length = 0;
+  auto end = current_pose - path.poses.begin();
+  while (static_cast<int>(end) < static_cast<int>(path.poses.size() - 1) &&
```
SteveMacenski, line 91: This could be just `end < path.poses.end()` since this is an iterator, no need to calculate positions.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/truncate_path_local_action.hpp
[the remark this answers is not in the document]
SteveMacenski: I'm not sure I understand, you're processing the path here anyway, if you grab the path from the blackboard and create a local variable with the path value, you can do whatever you want. You have a `path` and a `robot_pose`, this is no different inputs than those have. I'm not suggesting you override anything, I'm suggesting you remove the `angular_distance_weight` and the custom `poseDistance` that uses it and instead use the method https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L670-L682 to find the closest point on the path, since that style is mirrored in RPP, DWB, and I believe in 1 other place to do that same operation.

Then once you have the point in the path closest to the robot, you can use the integrated distance `nav2_util` (maybe needs to be modified for inputs or create a new instance of it) to find the segment backward and forward of interest based on your parameters, rather than implementing those inline in the BT node. I think integrating N distance backward or forward from a point on the path could be a useful general utility for other applications, don't you think?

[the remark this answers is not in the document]
afrixs: Yeah, it's a great new util for this purpose, I'm already using it in my code, just haven't pushed that yesterday, since this discussion hasn't been resolved yet... No need to add a special "reverse" case, it works well with `std::reverse_iterator`s

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+inline void TruncatePathLocal::updatePath(bool reset_path_pruning_on_change)
+{
+  if (reset_path_pruning_on_change) {
+    nav_msgs::msg::Path new_path;
+    getInput("input_path", new_path);
+
+    // detect for changes
+    bool changed = true;
```
SteveMacenski, line 110: I believe just `new_path != path_ -> changed = true` would suffice. I believe that would iterate through the internal structures.

SteveMacenski: The entire block from L109-121 can just be 

`if (new_path != path_) {path_ = new_path; closest_pose_detection_begin_ = path_.poses.begin();} else {...}`

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+      }
+    }
+
+    if (changed) {
+      path_ = new_path;
+      closest_pose_detection_begin_ = path_.poses.begin();
+    }
+  } else {
+    getInput("input_path", path_);
```
SteveMacenski, line 126: This seems to do essentially the same thing as in L121 above

- sets path_ to new_path
- sets `closest_pose_detection_begin_` to `path_` begin

vs 

- gets the `path_` from the port, which is identical to `new_path`, since new_path was never modified
- sets `closest_pose_detection_begin_` to `path_` begin

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
```diff
@@ -0,0 +1,178 @@
...
+  output_path.poses = std::vector<geometry_msgs::msg::PoseStamped>(
+    backward_pose_it.base(), forward_pose_it);
+  setOutput("output_path", output_path);
+
+  return BT::NodeStatus::SUCCESS;
+}
+
+inline void TruncatePathLocal::updatePath(bool reset_path_pruning_on_change)
+{
+  if (reset_path_pruning_on_change) {
```
SteveMacenski, line 105: Overall, I don't see why this function exists, it can be trivially replaced with `getInput("input_path", path_);` in the main `tick`  function

SteveMacenski: Even if the path hasn't changed, it doesn't hurt just to reset `path_` with itself.

SteveMacenski (changes requested):
This has gotten pretty convoluted in the truncation node. I think it would be better if we handled the BT node in another PR so that we can ship the actual smoother work which is completed except for the documentation elements on navigation.ros.org.  There are probably going to be a few more iterations required given the current state of this BT node but the smoother is approved. 

If only all users wanted to do roughly-exact path tracking instead of dynamic behaviors... it would make all of our lives so much easier... :laughing:

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
SteveMacenski: I feel like I'm missing something subtle here

afrixs: Yes, it can be simplified. I just wanted to leave out comparisons of headers, thinking that there's no need to reset pruning if poses don't actually change. But maybe I just overthinked it...

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
afrixs: ok, the if/else practically spared just one copying of the path, which is not a big deal after all... can simplify it for readability

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
afrixs: actually it would need to be
```
  nav_msgs::msg::Path new_path;
  getInput("input_path", new_path);
  if (!path_pruning || new_path != path_) {
    path_ = new_path;
    closest_pose_detection_begin_ = path_.poses.begin();
  }
```
Keeping previous value of `closest_pose_detection_begin_` if path hasn't changed makes the difference (it enables path pruning to take effect).

On whether it shoud be in a separate method: the `tick()` just started taking a bit of spaghetti-look, so I separated out some closed actions (`updatePath`, `getRobotPose`). If `updatePath` gets pruned to those six lines above, I have no problem putting it back into `tick()

nav2_behavior_tree/plugins/action/truncate_path_local_action.cpp
```diff
@@ -0,0 +1,178 @@
...
+  const double angular_distance_weight)
+{
+  double dx = pose1.pose.position.x - pose2.pose.position.x;
+  double dy = pose1.pose.position.y - pose2.pose.position.y;
+  tf2::Quaternion q1;
+  tf2::convert(pose1.pose.orientation, q1);
+  tf2::Quaternion q2;
+  tf2::convert(pose2.pose.orientation, q2);
+  double da = angular_distance_weight * std::abs(q1.angleShortestPath(q2));
+  return std::sqrt(dx * dx + dy * dy + da * da);
```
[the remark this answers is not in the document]
afrixs: Yes, it can be still helpful. It can help e.g. picking the correct pose near cusps (where robot often finds itself spatially closer to segment _after_ the cusp, but still is not rotated enough, so technically it is still _before_ the cusp) or it can be used for loop problems if user decides (due to API elegancy flaws/statefulness risk reasons I stated in comments above) to use the non-stateful approach.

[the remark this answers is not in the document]
SteveMacenski: OK, an inline comment explaining the value would be  great. Else 10 years from now someone will file this as a bug and I want an explicit mention that this is to deal with orientation changes due to looping / cusps in the paths

SteveMacenski (approved):
LGTM, I think at this point its just on documentation / adding in the diagrams / descriptions we've previously discussed in the still-open discussion points (you may need to expand the shortened view git has due to the number of comments on the PR). The check list on the top of the PR for maintainers goes through the documentation elements that be updated + the diagrams we previously discussed. 

I did another run through of the smoother code and I have no new comments / requests. This all looks ready to go to me! 

If you rebase to main / pull it in, CI should turn over with fewer errors.

SteveMacenski:
Any update? I think its just about an hour of writing up the documentation around this for the website! Docs help other users and gives you a better platform for users to use this work if they can understand how it works / what dials they have to play with (more users = more people to find issues = more fixes without our explicit effort)

afrixs:
Agreed, I'll get to it today or tomorrow

SteveMacenski:
Thanks for your dedication and help in getting this in! It's much appreciated and one of the new crown jewels in Nav2!

afrixs:
Thank you very much for your notes and experiences, I think we've together found a solution I'm satisfied with much more than with our original version.

Issue #2690, opened 2021-11-18
[Smac Planner]: Obstacle Heuristic continue expanding to similarly costed regions


mangoschorle:
I am experiencing unintuitive results regarding the SMAC Hybrid planner.
It appears to me that the planner's exploration is excessively dominated by distance-, rather than obstacle-costs.

I have set up a toy environment, where the distance-wise shortest path is obstructed by a large high-cost (251) area. 

I would expect the planner to favor the (distance-wise) slightly longer path alternative path with no high-cost obstacle in between. However, the planner chooses to traverse the high-cost region.

This behavior persists irrespective of the parameterized cost_penalty. 

![Screenshot from 2021-11-18 14-53-02](https://user-images.githubusercontent.com/94619337/142429755-2d76992b-5d2f-45ca-b94b-590d004d3f0d.png)

When shifting the starting position slightly (towards the alternative path without the high-cost are) the solution changes.

Issue:
The planner produces results, which seem to be heavily biased towards distance rather than costmap-cost. The cost_penalty coefficient has no apparent influence (neither had the analytic_expansion_ratio).

![Screenshot from 2021-11-18 13-37-34](https://user-images.githubusercontent.com/94619337/142421062-9ead525c-3a72-4c62-b594-686935555462.png)

Questions
Am I missing an apparent property of the SMAC planner, is this the expected behavior, or may there actually be some issues with the two heuristics.
**Required Info:**

- Operating System:
  - Used the provided docker- development environment on a Ubuntu 20.04 host

#### Steps to reproduce issue
I have attached the maps and the parameterization 
[parameter_and_map.zip](https://github.com/ros-planning/navigation2/files/7563140/parameter_and_map.zip)

#### Expected behavior
Planner balances geometrical distance with costmap-cost and is affected by cost_penalty 

#### Actual behavior
Planner solution is heavily biased towards short solutions

SteveMacenski:
A couple of things

First see https://github.com/ros-planning/navigation2/issues/2659 which meets the thrust of your ticket. This is known and has a known solution that is being worked on (probably?). Towards the end, it will just plow through higher cost situations if there's a valid solution because that's what analytic expansion means.

Now there are 2 more "actual" clarifications I have to make from your toy situation. This is where a _real_ situation you would see success in but because you've set up an unrealistic  setting, you're getting unrepresentative results.

A.

[...]

Issue #2725, opened 2021-12-03
[Smac] Optimizing performance


SteveMacenski:
Starting a new thread on performance considerations from https://github.com/ros-planning/navigation2/issues/2710 with @afrixs.

I believe we've talked about a couple of options

- Moving some of the A* logic into the NodeT's which speeds  things up for some reason I don't fully (personally) understand from the description, but definitely worth looking into
- In-situ costmap downsampling and caching so we don't downsample the entire costmap each cycle (see analysis of valgrind output https://github.com/ros-planning/navigation2/issues/2710#issuecomment-983821267)
- Increasing Hybrid-A* primitive lengths, make use of sub-primitive collision checking that State Lattice setup. Should reduce wobbling / # expansions.
- Downsample distance heuristic to every other grid cell and double the distance like we do for the obstacle heuristic

Happy to discuss others as well, these were just the ones I found from scanning, if you have others, I'd love to add to the list. I think we're on the same page on the Downsampling value / implementation potential. 

Can you give some more explanation on the `g_cost` and "helper node" in the NodeT? The blurb in https://github.com/ros-planning/navigation2/issues/2710#issuecomment-983821267 didn't give me a good understanding about why that would speed things up. I re-read and I think I understand the mechanics, there may be a way to compromise to keep alot of the A* logic in the A* file if I'm understanding correctly, but I'm not confident that I am.

PR #2765, opened 2022-01-13
Smac planner admissible obstacle heuristic

robotechvision:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2690 |
| Robotic platform tested on | RTV Caster, gazebo simulation of RTV Caster |

---

* Added a consistent and admissible obstacle heuristic computation solution which uses Differential A* instead of a simple dynamic BFS
* Added `use_min_cost_neighbor` attribute to CostmapDownsampler to ensure that heuristic doesn't exceed real cost and thus remains admissible
* Added `obstacle_heuristic_admissible` parameter to switch between Differential A* and dynamic BFS methods
* Added `retrospective_penalty` parameter to compensate planner speed drop with admissible heuristic (which produces slightly lower heuristic values in general because of being truly optimal, resulting in significantly more expansions of the 3D search). Setting this parameter to a non-zero value makes the search inadmissible again but in a controlled way (i.e. still avoiding [situations like this](https://github.com/ros-planning/navigation2/issues/2690#issue-1057366322) caused by error-driven inadmissibility). Can be used to speed up the inadmissible heuristic search too.

* Added new parameters (`obstacle_heuristic_admissible`, `retrospective_penalty`), so need to add them to default configs and documentation page

---

* Nothing comes to mind...

#### For Maintainers: 


nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -492,6 +515,126 @@ float NodeHybrid::getObstacleHeuristic(
...
+
+  const std::vector<int> neighborhood = {1, -1,  // left right
+    size_x_int, -size_x_int,  // up down
+    size_x_int + 1, size_x_int - 1,  // upper diagonals
+    -size_x_int + 1, -size_x_int - 1};  // lower diagonals
+
+  const int goal_x = floor(goal_coords.x / 2.0);
+  const int goal_y = floor(goal_coords.y / 2.0);
+  const int goal_index = goal_y * size_x + goal_x;
+  if (astar_2d_first_run) {
```
SteveMacenski, line 555: I believe (or perhaps not start?) that you have all of the information required to do the `astar_2d_queue.emplace_back(...)` in `resetObstacleHeuristic`. That would let you remove the `astar_2d_first_run` flag and streamline this function. That's what the original implementation did, e.g.

```
obstacle_heuristic_queue.emplace(
      ceil(goal_y / 2.0) * sampled_costmap->getSizeInCellsX() + ceil(goal_x / 2.0));
```

after it resized all of the data structures and reset them. That would keep the 2 implementations at least relatively analog.

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -492,6 +515,126 @@ float NodeHybrid::getObstacleHeuristic(
...
+    astar_2d_queue.emplace_back(
+      distanceHeuristic2D(
+        goal_index, size_x, start_x,
+        start_y), goal_index);
+
+    // initialize goal cell with a very small value to differentiate it from 0.0 (~uninitialized)
+    // the negative value means the cell is in the open set
+    obstacle_heuristic_lookup_table[goal_index] = -0.00001f;
+  } else {
+    for (auto & n : astar_2d_queue) {
```
SteveMacenski, line 567: Marching through every cell here is costly. I'm not understanding this loop. If we've already started the table and we're still executing the same planning / goal run, why is this being updated? 

The goal pose (e.g. where we start) is still the same, so for any given cell, the heuristic value would continue to be the same. We're continuing to expand the heuristic front because we were asked to give an H value for a cell that wasn't previously assigned in the last expansion. All of the prior expanded cells should still be valid since its the same planning request. Right?

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -492,6 +515,126 @@ float NodeHybrid::getObstacleHeuristic(
...
+      }
+    }
+
+    if (c.second == start_index) {
+      break;
+    }
+  }
+
+  // costs are doubled due to downsampling
+  return 2.0 * starting_cost;
```
SteveMacenski, line 635: The fact that `starting_cost` is a reference is a little subtle. Mind changing this to be a little more explicit `obstacle_heuristic_lookup_table[start_index];` or adding a little comment here that `starting_cost` is a reference so its updated from the search process (I don't mind which)?

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -492,6 +515,126 @@ float NodeHybrid::getObstacleHeuristic(
...
 }
 
+inline float distanceHeuristic2D(
+  const int idx, const int size_x, const int goal_x,
+  const int goal_y)
+{
+  int x = std::abs(idx % size_x - goal_x);
+  int y = std::abs(idx / size_x - goal_y);
+  float & table_value = NodeHybrid::astar_2d_h_table[y * size_x + x];
+  return table_value < 0.0f ? table_value = std::hypotf(x, y) : table_value;
```
SteveMacenski, line 525: Just curious - how much does caching this L2 norma speed things up (ballpark)? If you don't have numbers handy that's OK as long as you know it actually does speed things up.

nav2_smac_planner/README.md
```diff
@@ -121,6 +121,7 @@ planner_server:
       change_penalty: 0.0                 # For Hybrid nodes: penalty to apply if motion is changing directions, must be >= 0
       non_straight_penalty: 1.20          # For Hybrid nodes: penalty to apply if motion is non-straight, must be => 1
       cost_penalty: 2.0                   # For Hybrid nodes: penalty to apply to higher cost areas when adding into the obstacle map dynamic programming distance expansion heuristic. This drives the robot more towards the center of passages. A value between 1.3 - 3.5 is reasonable.
+      retrospective_penalty: 0.01         # For Hybrid/Lattice nodes: penalty to prefer later maneuvers before earlier along the path. Saves search time since earlier nodes are not expanded until it is necessary
```
SteveMacenski, line 124: A note that it should be between 0-1.0 would be good

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -314,7 +321,7 @@ float NodeHybrid::getTraversalCost(const NodePtr & child)
 
   float travel_cost = 0.0;
   float travel_cost_raw =
-    NodeHybrid::travel_distance_cost +
+    NodeHybrid::travel_distance_cost * motion_table.travel_distance_reward +
```
SteveMacenski, line 324: Can you give a bit more of an expanded explanation of this and what values you find are "reasonable"? Was the drop in speed so significant we need new flags to artificially speed it up? Where's that compute time going to, specifically?  

What's the expansion pattern look like now?

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: I suspect this is not actually required

SteveMacenski (changes requested):
Really good - I think I found 2 areas we can improve speed if I'm correct and a few other areas just need clarity. If we work on these points, I see no reason for a second review, everything else was spot on.

nav2_smac_planner/src/costmap_downsampler.cpp
```diff
@@ -140,7 +142,10 @@ void CostmapDownsampler::setCostOfCell(
       if (my >= _size_y) {
         continue;
       }
-      cost = std::max(cost, _costmap->getCost(mx, my));
+      cost = _use_min_cost_neighbor ? std::min(cost, _costmap->getCost(mx, my)) : std::max(
+        cost, _costmap->getCost(
```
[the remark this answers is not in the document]
afrixs: yeah, `ament_uncrustify` sometimes does weird things...

nav2_smac_planner/src/costmap_downsampler.cpp
```diff
@@ -126,7 +128,7 @@ void CostmapDownsampler::setCostOfCell(
   const unsigned int & new_my)
 {
   unsigned int mx, my;
-  unsigned char cost = 0;
+  unsigned char cost = _use_min_cost_neighbor ? 255 : 0;
```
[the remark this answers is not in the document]
afrixs: This is not a default value, it's a starting value of finding a minimum/maximum of the cells being downsampled and needs to be the lowest/highest `unsigned char` value so that `std::max`/`std::min` takes effect

nav2_smac_planner/src/smac_planner_lattice.cpp
```diff
@@ -83,6 +83,11 @@ void SmacPlannerLattice::configure(
   nav2_util::declare_parameter_if_not_declared(
     node, name + ".cache_obstacle_heuristic", rclcpp::ParameterValue(false));
   node->get_parameter(name + ".cache_obstacle_heuristic", _search_info.cache_obstacle_heuristic);
+  nav2_util::declare_parameter_if_not_declared(
```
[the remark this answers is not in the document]
afrixs: Actually I don't see a reason for keeping the old implementation, I left the option there just because you were quite resistive against the change in our [discussion](https://github.com/ros-planning/navigation2/issues/2498) because of computation speed. Our tests have, however, shown that the new heuristic computation is in general faster (and more importantly it yields correct results) but there are some theoretical situations where the old can be faster so I'm leaving the decision about deleting the old upon you.

If needed for benchmarking, you can check out our [smac_planner_debug](https://github.com/robotechvision/navigation2/tree/smac_planner_debug) branch which contains some heuristic computation time logs, but maybe you have your own tools.

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -401,13 +412,25 @@ void NodeHybrid::resetObstacleHeuristic(
     // must reset values for non-constructed indices
     std::fill_n(
       obstacle_heuristic_lookup_table.begin(), obstacle_size, 0.0);
+    if (motion_table.obstacle_heuristic_admissible) {
+      astar_2d_h_table.resize(size, -1.0f);
+      // must reset values for non-constructed indices
+      std::fill_n(
```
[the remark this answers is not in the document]
afrixs: This is just a relative euclidean distance cache whose values are not going to be changed with a different goal. However, I'm going to measure whether the whole caching thing has any sense so maybe `astar_2d_h_table` will be deleted.

nav2_smac_planner/src/node_hybrid.cpp
afrixs: This was done just because of the start position missing in `resetObstacleHeuristic`. No problem to change the method signature and put it there instead.

nav2_smac_planner/src/node_hybrid.cpp
afrixs: Yes, this is the most questionable element in the new implementation so I'm going to elaborate a bit:

First thing - the heuristic algorithm finds path costs in the opposite direction, i.e. from the goal pose to node. So the goal is the start and current node is the target. All of the prior expanded cell costs are valid since its the same planning request but the 2D A* cell priority is composed of 

**travel cost from start** (start is the goal position so this cost remains the same) + **euclidean distance heuristic towards target** (target is the currently requested node so this **changes**)

The second term causes, unfortunately, that reprioritization is needed. I've done quite a wide research, searching for an algorithm that addresses this problem (path search from an unchanging position to a changing position) better and found multitude of algorithms which address similar problems (moving target replanning/pursuit path search) but every seemingly usable one of them assumed the target doesn't move too fast, which is not our case because our targets (~current requested nodes) can jump all around the OPEN set during the main A* search and such big jumps would significantly impact the computational speed of the algorithms (they would add the complexity of around `O(q*log(q)*r)` where `q` is the queue size and `r` is number of requested unvisited cells). Queue reprioritization, used in the currently proposed algorithm, has a [linear complexity](https://en.cppreference.com/w/cpp/algorithm/make_heap) and thus adds just `O(q*r)`.

I originally thought about making a hybrid search which uses one of the moving target algorithms for small jumps of the target cell (currently requested node) and the queue reprioritization for larger jumps and yes, I think it is possible, e.g. by combining a lazy requeuing method (described in D* Lite paper) with the currently used queue reprioritization (described in Differential A* paper). However, after implementing the reprioritization part, I did some benchmarking which showed that it already works very well and fast in practice, occupying just about 20% of the whole heuristic computation so I decided not to implement the lazy requeuing.

However, situations which need the lazy requeuing to be solved in real-time can theoretically occur but since the update is a little bit heavier, I would wait with it until (if at all) they actually appear in practice.

nav2_smac_planner/src/node_hybrid.cpp
afrixs: Will add a comment and rename `starting_cost` to `requested_cell_cost`

nav2_smac_planner/src/node_hybrid.cpp
afrixs: Well, this caching is maybe just an artifact from my experiments (presuming there may be more to compute than just `hypotf`) and it's possible that it does not bring a speedup. I will test it and remove if unnecessary/detrimental.

nav2_smac_planner/src/node_hybrid.cpp
afrixs: > Was the drop in speed so significant we need new flags to artificially speed it up?

It depends on situation and other-than-costmap penalties but in some situations I recorded computation time rise from 114ms to 583ms so it can be quite significant (although ["turning traversal penalties off"](https://github.com/ros-planning/navigation2/issues/2498#issuecomment-966579662) can make the difference smaller).

> Where's that compute time going to, specifically?

There happen to be more node expansions in the main A* search. Reason is that since 2D heuristic search now returns an optimal 2D path cost, the resulting H value is lower than the H value of a sub-optimal path found by the inadmissible search. Not by much but even a small difference introduces a large difference in expansions count (as you must have already experienced too [when retuning penalties](https://github.com/ros-planning/navigation2/issues/2498#issuecomment-966579662)). Here you can see a situation where decreasal of H value by 1.33% (largest H value: 1293 -> 1276) raises the number of expansions (the rainbow-colored band) by 561% (46984 -> 263829).
Screenshot from 2022-01-12 17-26-39
Screenshot from 2022-01-12 17-28-16

> Can you give a bit more of an expanded explanation of this?

The reason of the above (H-value/expansions-count) gain being so large is that the search wants to find an optimal path and thus every little other-than-costmap traversal penalty and every little deviation from the 3D path from the 2D optimal path makes the search algorithm ask "Could a change made earlier along the path make the current deviation/penalty smaller?". Therefore in that situation it usually expands a large number of nodes all along the path trying to find the answer (~a better path). However, in a vast majority of situations the solution is somewhere near the "problematic node" and therefore a question arises: "Is it worth to look for an answer so far backwards?". Introducing retrospective penalty gives an option to the search to ask such question and thus spare time not expanding a large number of nodes which have just a very small chance to improve the path.

Of course, it doesn't affect the ability of finding a path, it just gradually prioritizes "later/longer" search branches before "earlier/shorter" ones.

Technically it doesn't increase the cost of all earlier nodes, instead it decreases the cost of newly added ones.

> And what values you find are "reasonable"?

It depends on how much speedup you need and how difficult your map is.

Of course, when this penalty is turned on, the search loses any guarantees for the path to be optimal. If you use a Reeds-Shepp model and your goal is e.g. to reverse into a long narrow passage, inside which it's difficult to rotate (although possible but very problematic and turn should be chosen just in case there are no other options), you will want a value of 0.0, because whenever planner finds the robots should have turned backwards, it needs to go back all along the path to plan a rotation before the entrance to the passage. If a retrospective penalty was used, a later turn inside the passage could be prefered and robot could get stuck doing the dangerous rotation.

If situations like the above one are not present in your environment/use-case, you can penalize the retrospective a bit. According to our tests, value 0.01 decreased number of expansions of the example above down to 35044 expansions (less than with the original inadmissible search), while still dealing pretty well with the passage reversal situation described in the paragraph above.
Screenshot from 2022-01-13 08-18-36

Following is an illustrative example with value 0.2. You can see how the rainbow-colored band expands only near the problematic parts. The search still deals very well with problems along the path, however the value is maybe quite marginal, making it possible in some situations to prefer paths which are problematic to execute.
Screenshot from 2022-01-13 09-00-25

nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
```diff
@@ -38,6 +38,17 @@ namespace nav2_smac_planner
...
+typedef std::pair<float, int> AStar2DElement;
+struct AStar2DComparator
+{
+  bool operator()(const AStar2DElement & a, const AStar2DElement & b) const
+  {
+    return a.first > b.first;
+  }
+};
+
+typedef std::vector<AStar2DElement> AStar2DQueue;
```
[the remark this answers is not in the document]
afrixs: The reason I had to use `std::vector` and heap operations is the need to assign new priorities to cells (see comment below), i.e. iterating over all elements in linear time which is not supported by `std::priority_queue` + reordering the queue subsequently which is not supported too.

nav2_smac_planner/src/costmap_downsampler.cpp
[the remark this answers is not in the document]
SteveMacenski: Good point, thanks

nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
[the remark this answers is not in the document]
SteveMacenski: Got it, for now will discuss below on those points. Will re-circle up to this to see if still relevant after.

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: OK, its just a very large data structure to have around so if there's no speed up, that would decrease memory usage _alot_.

nav2_smac_planner/src/smac_planner_lattice.cpp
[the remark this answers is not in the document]
SteveMacenski: Got it, do you have some relative numbers to share on the compute time saved (on average)? It's obviously map / location dependent, but some ballparks to start would be nice. I can then use that to see if I can get similar numbers in my maps to verify.

If its not too much more compute, or even better yet, _less_  compute, then I'd be happy to have my implementation striped out completely. Lets leave in for now so I can do that benchmarking more easily side-by-side. Let me know on the metrics you see.

nav2_smac_planner/src/node_hybrid.cpp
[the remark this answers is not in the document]
SteveMacenski: I'm unclear on this - if you later check if the values of `astar_2d_h_table` are `< 0` to know if they're valid to use, you would need to reset them all to be negative between goals. If you had 2 different goal locations on the same map, this would change in meaning to need to be reconstructed, no?

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: If we're going to remove the current heuristic implementation, there's no reason to change the signature since there wouldn't be 2 implementations to try to keep analog. We can keep it this way right now. I'll recircle back to this comment after we benchmark performance and come to a final decision about whether to remove it.

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: I see, the need to re-prioritize makes sense after I very carefully went through it. A comment at this block of code about that would be nice since that is also subtle.

nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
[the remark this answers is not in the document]
SteveMacenski: I understand the need now, this is good to stay as is.

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: > in some situations I recorded computation time rise from 114ms to 583ms so it can be quite significant  ... raises the number of expansions (the rainbow-colored band) by 561% (46984 -> 263829).

:grimacing: yeeek! When adding in the non-0 values, does this get back in line with the ~114 ms number? I see the expansions go back in line, just checking that expansions + new H computation still is in the same ballpark. 

A value of 0.2 would be 20% off which would seems high. It doesn't surprise me that you see some issues if you discount expansions by 20% that it would cause some non-optimal issues. It does surprise me that even just 1% bounds things so quickly. 

If you were to pick a "reasonable" default value, what would it be? Sounds like between 0.01 - 0.20.

nav2_smac_planner/src/smac_planner_lattice.cpp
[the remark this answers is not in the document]
afrixs: I have tested it some time ago and for an 80m long path from one map corner to another both computations lasted around 40 ms. 4x speedup came with a path (25m: 16ms vs 4ms) leading to a goal which is in the middle of the map since the inadmissible implementation searches in every direction from goal while the new 2D A* implementation searches from goal mostly towards starting node

nav2_smac_planner/src/node_hybrid.cpp
afrixs: > When adding in the non-0 values, does this get back in line with the ~114 ms number?

Did you mean non-0 value of `retrospective_penalty`? Yes, it brings the computation down to around 120 ms or even lower

> It does surprise me that even just 1% bounds things so quickly

As I said, there was around 1.33% decrease of heuristic values with the new admissible computation so 1% brings the expansion number back (+ even lower since the retrospective penalization here is continuous, taking effect all along path, which makes it more effective than the one introduced by an inadmissible heuristic, which was error-driven and thus occurred presumably mostly around obstacle corners etc.).

> If you were to pick a "reasonable" default value, what would it be?

For me it is OK to be 0.01 (you know I like "more optimal paths and longer planning time" more than "less optimal paths and shorter planning time") or even lower. But since this is just one test/comparison, we could use some reserve, e.g. 0.015. But I would like you to find the proper value because our priorities differ here.

SteveMacenski: Got it - OK. If you could set the value in your PR to have a default value in the plugins of `0.015` that would be great. Since that puts thing back in line, **I'd support removing my existing implementation** here. Which means you can delete that stuff and change this to being the only obstacle heuristic.

nav2_smac_planner/src/smac_planner_lattice.cpp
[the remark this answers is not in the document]
SteveMacenski: Got it -- I think then the only thing remaining in this line item is the `This is missing from the readme list of variables.`

SteveMacenski:
Hi,

Just quickly following back up on this. Any update on these changes? I've paused on the Ceres PR just so I don't overload you with things and get 1 PR merged in at a time.

nav2_smac_planner/src/node_hybrid.cpp
afrixs: I retested the planner with more paths and decided to use 0.025 as default to keep expansions number < 50000 (=> planning time < 120ms) for all 50 m long paths I tested with...

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -402,72 +415,89 @@ void NodeHybrid::resetObstacleHeuristic(
...
+    mx_idx = c.second - (my_idx * size_x);
 
     // find neighbors
     for (unsigned int i = 0; i != neighborhood.size(); i++) {
-      new_idx = static_cast<unsigned int>(static_cast<int>(idx) + neighborhood[i]);
-      cost = static_cast<float>(sampled_costmap->getCost(idx));
-      travel_cost =
-        ((i <= 3) ? 1.0 : sqrt_2) + (((i <= 3) ? 1.0 : sqrt_2) * cost_penalty * cost / 252.0);
-      current_accumulated_cost = last_accumulated_cost + travel_cost;
+      new_idx = static_cast<unsigned int>(static_cast<int>(c.second) + neighborhood[i]);
```
mangoschorle, line 492: @SteveMacenski @afrixs new_idx is int so if the result of static_cast<unsigned int>() exceeds the max of int, new_idx will be negative and the cost lookup segfaults.

In may case:
 c.second = 135
 neighborhood[i] = -407
 this leads to new_idx = -272.
 The cast in getCost to unsigned int makes the lookup at an array index well out of bounds. -> Segfault

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: @afrixs did you ultimately find this didn't speed anything up (even when having to reprioritize the queue on occasion?)

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -402,72 +415,89 @@ void NodeHybrid::resetObstacleHeuristic(
...
+  obstacle_heuristic_queue.reserve(
+    sampled_costmap->getSizeInCellsX() * sampled_costmap->getSizeInCellsY());
+
   // Set initial goal point to queue from. Divided by 2 due to downsampled costmap.
-  std::queue<unsigned int> q;
-  std::swap(obstacle_heuristic_queue, q);
-  obstacle_heuristic_queue.emplace(
-    ceil(goal_y / 2.0) * sampled_costmap->getSizeInCellsX() + ceil(goal_x / 2.0));
+  const unsigned int size_x = sampled_costmap->getSizeInCellsX();
+  const unsigned int goal_index = floor(goal_y / 2.0) * size_x + floor(goal_x / 2.0);
```
SteveMacenski, line 424: You changed `ceil` to `floor` here, was that intentional?

SteveMacenski: As well below in L440-441, was there something wrong with the current behavior?

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -402,72 +415,89 @@ void NodeHybrid::resetObstacleHeuristic(
...
+  const unsigned int start_x = floor(node_coords.x / 2.0);
+  const unsigned int start_index = start_y * size_x + start_x;
+  const float & requested_node_cost = obstacle_heuristic_lookup_table[start_index];
+  if (requested_node_cost > 0.0f) {
     // costs are doubled due to downsampling
-    return 2.0 * starting_cost;
+    return 2.0 * requested_node_cost;
   }
 
-  // If not, expand until it is included. This dynamic programming ensures that
```
SteveMacenski, line 427: Please readd this comment block, I think its useful to show people that this is dynamic programming and caching

nav2_smac_planner/src/node_hybrid.cpp
afrixs: Yes, I added some time logs and there was no speed up at all.

nav2_smac_planner/src/node_hybrid.cpp
afrixs: @mangoschorle Did you check this with my latest updates from yesterday? I moved the `cost = static_cast<float>(sampled_costmap->getCost(new_idx));` line into the `if (new_idx > 0 && new_idx < size_x * size_y)` block and I see no reason here for Segfaults any more. Also `new_idx` is `unsigned int`, not `int`.

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -402,72 +415,89 @@ void NodeHybrid::resetObstacleHeuristic(
...
+  const float sqrt_2 = sqrt(2);
 
-    if (idx == start_index) {
-      // costs are doubled due to downsampling
-      return 2.0 * last_accumulated_cost;
+  while (!obstacle_heuristic_queue.empty()) {
+    ObstacleHeuristicElement c = obstacle_heuristic_queue.front();
+    std::pop_heap(
+      obstacle_heuristic_queue.begin(), obstacle_heuristic_queue.end(),
+      ObstacleHeuristicComparator{});
```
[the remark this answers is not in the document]
afrixs: These things are handled by optimizer but can put it outside among other variables...

nav2_smac_planner/src/node_hybrid.cpp
afrixs: Yes, I think that `ceil` was there for some mysterious historical reasons or something, because it is just wrong and points to an incorrect cell. Nodes use [`static_cast<unsigned int>`](https://github.com/ros-planning/navigation2/blob/85cf934de1c0cf8d335f8d46440a96636410b8ee/nav2_smac_planner/src/collision_checker.cpp#L97) which is equivalent of `floor` for positive numbers and downsampler uses [`coord*2 + (0 or 1)`](https://github.com/ros-planning/navigation2/blob/85cf934de1c0cf8d335f8d46440a96636410b8ee/nav2_smac_planner/src/costmap_downsampler.cpp#L136) which is again inverse of `floor`, not `ceil`. Also verified by logs during analysis and debugging.

nav2_smac_planner/src/node_hybrid.cpp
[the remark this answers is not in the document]
afrixs: + changing the variable idx, since `c.first` is never used

nav2_smac_planner/src/node_hybrid.cpp
mangoschorle: @afrixs My bad. I hadn't looked at the new version.
Just a question:
The neighbourhood doesn't work on costmap edges, or am I wrong. The +1 might actually not be the cell to the right but in the next line on the very left?

afrixs: There's a check for the edges, see [these lines](https://github.com/ros-planning/navigation2/blob/85cf934de1c0cf8d335f8d46440a96636410b8ee/nav2_smac_planner/src/node_hybrid.cpp#L504)

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -415,59 +436,74 @@ float NodeHybrid::getObstacleHeuristic(
...
     for (unsigned int i = 0; i != neighborhood.size(); i++) {
       new_idx = static_cast<unsigned int>(static_cast<int>(idx) + neighborhood[i]);
-      cost = static_cast<float>(sampled_costmap->getCost(idx));
-      travel_cost =
-        ((i <= 3) ? 1.0 : sqrt_2) + (((i <= 3) ? 1.0 : sqrt_2) * cost_penalty * cost / 252.0);
-      current_accumulated_cost = last_accumulated_cost + travel_cost;
 
       // if neighbor path is better and non-lethal, set new cost and add to queue
-      if (new_idx > 0 && new_idx < size_x * size_y && cost < INSCRIBED) {
+      if (new_idx >= 0 && new_idx < size_x * size_y) {
```
mangoschorle, line 501: new_idx is unsigned_int. This is a redundant check

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: OK

nav2_smac_planner/src/node_hybrid.cpp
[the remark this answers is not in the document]
SteveMacenski: Thanks :-)

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: This is the only remaining open comment @mangoschorle is this solved to your satisfaction to "resolve" it?

mangoschorle: Its good! I have zero experience with github. Do I need to press some "resolved" button, or are you asking because you didn't see any reply from my side?

SteveMacenski (approved):

SteveMacenski:
This looks good to me. @mangoschorle what do you think? 

Once he approves, I'll play with it for a bit locally, but I see no reason this wouldn't be merged by the end of this week.

SteveMacenski:
On documentation, 

- the new parameters need to be added to the State Lattice / Hybrid-A* configuration files https://navigation.ros.org/configuration/packages/smac/configuring-smac-hybrid.html and https://navigation.ros.org/configuration/packages/smac/configuring-smac-lattice.html
-  Add a bullet (or two) on the migration guide entry for smac https://navigation.ros.org/migration/Galactic.html
-

mangoschorle:
> This looks good to me. @mangoschorle what do you think?
> 
> Once @mangoschorle approves, I'll play with it for a bit locally, but I see no reason this wouldn't be merged by the end of this week.
> 
> Great quality, as always. Very happy to have your contribution and critique.

@SteveMacenski I integrated this in our setup and let the planner have a go on several of our system tests. I am very happy with the results.

mangoschorle (approved):

nav2_smac_planner/src/node_hybrid.cpp
SteveMacenski: Usually, yeah the "Resolve conversation" button. Don't worry I got it, just wanted to make sure this was fully settled.

SteveMacenski:
I did some quick testing, I'm happy with this performance. 

I will need to do a little tuning on the defaults again to get the performance to be best out of the box with the penalty functions and analytic expansion ratio. I don't have the cycles right at this exact moment to do that tuning so I don't want to merge this unless I can same-day turn around and tune it so there's little time between those two events.

@robotechvision as far as I'm concerned this is approved and practically merged, I super appreciate the detailed analysis and work you've done here and very well implemented. It's going to sit in this state for a short time until I retune and then merge. 

 If you had some time, I'd appreciate some information about some of the optimizations you've made in the Smac Optimizations ticket https://github.com/ros-planning/navigation2/issues/2725 so we can discuss adding those points as well to further improve performance.

SteveMacenski:
Did some tuning on State Lattice, I'm happy with it. Hybrid's a little more complex and I need to do more back-to-back comparisons

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -108,6 +108,9 @@ void SmacPlannerHybrid::configure(
   nav2_util::declare_parameter_if_not_declared(
     node, name + ".cost_penalty", rclcpp::ParameterValue(2.0));
   node->get_parameter(name + ".cost_penalty", _search_info.cost_penalty);
+  nav2_util::declare_parameter_if_not_declared(
+    node, name + ".retrospective_penalty", rclcpp::ParameterValue(0.025));
```
SteveMacenski, line 112:
```suggestion
    node, name + ".retrospective_penalty", rclcpp::ParameterValue(0.015));
```

SteveMacenski:
@robotechvision I'm working on a academic paper about Smac in order to have a detailed explanation of the framework, methodology, and planners in it that is citable. 

I had 2 questions for you
- I would like to include you and your contribution in the acknowledgements. Is that OK / it would be good to know specifically who all I should give a shout out to that's helped on your end
- I am looking for a figure to show the heuristic. You've included a few in the past but I wasn't sure if you had one from the final form of this PR that could be used to demonstrate the heuristic in your warehouse/store/etc world (with some runtime metric / number of expansions it represents). I was thinking potentially the heuristic field expanded where the color is proportional to the heuristic cost and the final path (smoothed) overlayed on top of it. Is that something that would be easy to create or that you have from your debugging efforts?

robotechvision:
Greetings @SteveMacenski ,
 * it's great you are writing an academic paper about the planner, we'll be gladly included in it. The team members participated are: Ing. Matej Vargovčík and Ing. Peter Pásztó, PhD. from RoboTech Vision company (diacritical marks are not obligatory if it is a problem)
 * adding screenshots with rviz and benchmark outputs from different planning situations, file names are in format `codename-rviz/bench-retrospective_penalty_used.png`. The lighter rainbow indicates expanded heuristic cells, while the darker inverted rainbow indicates expanded `NodeHybrid`s. The positive effect of retrospective penalty is most significant in "smoke" and "flower" situations, the negative effect is visible in the "jug" situation - resulting path is suboptimal and dangerous to execute when a high (20%) `retrospective_penalty` is used, although a much safer path exists (found when `retrospective_penalty` is set to 0)

[smac-article-screenshots.zip](https://github.com/ros-planning/navigation2/files/8304696/smac-article-screenshots.zip)

robotechvision:
Just noticed the message in Slack that the article's already been released, thanks for including us :) Keeping the previous message undeleted, since the images could be useful e.g. to demonstrate pros and cons of retrospective penalty in the docs

PR #2780, opened 2022-01-21
Bugfix tf lookup of goal pose in nav2_controller

ErikOrjehag:
Signed-off-by: Erik Örjehag <erik@orjehag.se>

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | (gazebo simulation of other diff drive robot) |

* Move tf lookup of goal pose in nav2_controller from `setPlannerPath` to `isGoalReached`. This is needed because the local planner will otherwise have a much more recently transformed goal compared to the goal checker causing the local planner to stop when it has reached its goal but the goal checker will not trigger (we use pure pursuit). If setPlannerPath is called with 1Hz this is barely noticeable but we have a different behavior tree that only plans when we have a new goal (and in the future we'll try to check if the plan in illegal) and then this is a big problem.

Screenshot_2022-01-21_10-43-48

Maybe others will run into a similar issue once this (https://github.com/ros-planning/navigation2/pull/2591#issuecomment-947945404) is in.

None

* I tested this change by overlaying nav2_controller in our project repo (container running galactic) but our project does automatic code formatting so the changes in this PR is just copy pasted from our project into a fork of mine and I was not able to build the navigation2 repo, hoping the CI will complain if it doesn't build.
* Ideally I would like to have this change in galactic as well but I don't know the process.

#### For Maintainers: 

### DIFF
--- a/nav2_controller/include/nav2_controller/controller_server.hpp
+++ b/nav2_controller/include/nav2_controller/controller_server.hpp
@@ -256,7 +256,7 @@ protected:
   double failure_tolerance_;
 
   // Whether we've published the single controller warning yet
-  geometry_msgs::msg::Pose end_pose_;
+  geometry_msgs::msg::PoseStamped end_pose_;
 
   // Last time the controller generated a valid command
   rclcpp::Time last_valid_cmd_time_;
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

SteveMacenski:
Rerunning testing job to get a green light on CI

SteveMacenski:
Thanks for the contribution! I think that may actually resolve another seemingly related problem with another contributors' project. Thanks for the explanation and the clear implementation.

PR #2782, opened 2022-01-21
[costmap2d] make robot_base_frame param dynamic

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/956 and https://github.com/ros-planning/navigation2/pull/2592|

---

Turns out than in addition to the previously added dynamic parameters https://github.com/ros-planning/navigation2/pull/2592, I needed also the robot_base_frame param to be dynamic

Mention in migration guide

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -654,22 +654,20 @@ Costmap2DROS::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameter
         // First, make sure that the transform between the robot base frame
         // and the global frame is available
         std::string tf_error;
-
         RCLCPP_INFO(get_logger(), "Checking transform");
-        rclcpp::Rate r(2);
-        while (rclcpp::ok() &&
-          !tf_buffer_->canTransform(
-            global_frame_, parameter.as_string(), tf2::TimePointZero, &tf_error))
+        if (!tf_buffer_->canTransform(
+            global_frame_, parameter.as_string(), tf2::TimePointZero,
+            tf2::durationFromSec(1.0), &tf_error))
         {
-          RCLCPP_INFO(
+          RCLCPP_WARN(
             get_logger(), "Timed out waiting for transform from %s to %s"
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -650,6 +650,28 @@ Costmap2DROS::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameter
...
         }
+      } else if (name == "robot_base_frame") {
+        // First, make sure that the transform between the robot base frame
+        // and the global frame is available
+        std::string tf_error;
+
+        RCLCPP_INFO(get_logger(), "Checking transform");
+        rclcpp::Rate r(2);
+        while (rclcpp::ok() &&
+          !tf_buffer_->canTransform(
```
SteveMacenski, line 661: This concerns me a little. I totally agree that this is needed, but in the situation that it can't transform because of a typo, does this lock up the system really badly if the node is spinning on this callback for awhile? 

I'd be concerned about safety, this seems almost like something we may want to have only changable in a completely transitioned inactive state. 

I could be fine with this, but I would like you to characterize what happens when this callback goes wrong so we know the symptoms

doisyg: We should definitely discuss this point and the desired behavior! I guess in its current state it is blocking indefinitely if there is a typo.
I propose to check the transform and if it is not available after a certain timeout (to discuss but it should be short), reject the parameter change (`result.successful = false;`) and leave the original value of the param `robot_base_frame` unchanged. 
What do you think ?

SteveMacenski: That seems very reasonable to me. Maybe give it 1 full second to make it work?

doisyg: Yes!

SteveMacenski (approved):

PR #2783, opened 2022-01-21
[OdomSubscriber] make odom_topic param dynamic

doisyg:
---

## Basic Info

| ------ | ----------- |

---

Make the odom_topic parameter of OdomSubscriber dynamic. I have a personal need of changing at runtime the odometry source used by nav2, so sharing here.
TBD: relevance and test.

Mention in migration guide

---

#### For Maintainers: 

### DIFF
--- a/nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/odom_subscriber.hpp
+++ b/nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/odom_subscriber.hpp
@@ -45,6 +45,8 @@
 #include "rclcpp/rclcpp.hpp"
 #include "nav2_util/node_utils.hpp"
 
+using rcl_interfaces::msg::ParameterType;
+
 namespace nav_2d_utils
 {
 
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Linter

> Add #include <vector> for vector<>

nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/odom_subscriber.hpp
```diff
@@ -91,9 +97,38 @@ class OdomSubscriber
...
+    return result;
+  }
+
   rclcpp::Subscription<nav_msgs::msg::Odometry>::SharedPtr odom_sub_;
   nav_2d_msgs::msg::Twist2DStamped odom_vel_;
   std::mutex odom_mutex_;
+  nav2_util::LifecycleNode::WeakPtr nh_;
+  std::string odom_topic_;
+  // Dynamic parameters handler
+  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr dyn_params_handler;
```
SteveMacenski, line 131: trailing underscore for class member

SteveMacenski:
This seems really quite odd. It's not odd enough that I wouldn't merge it, so I would question the design of a system that requires it and perhaps we can find you a better solution. 

Why not use the typical topic muxing methodology? Have a `output_odom` topic that you have something that toggles your preferred odometry source to publish to and have the system subscribe to that? Runtime changing subscribers breaks a bit of the lifecycle readiness guarantees you get when we create the network interfaces before the active state. Further, you still need to have some autonomy code that is changing this dynamic parameter, that could be trivially changed to just pushing to the `output` topic instead and no need to change any subscriptions at runtime. 

Also, needs test coverage on dynamic parameters.

doisyg:
Hum, a topic muxer would indeed make more sense. Let me reflect on that, but I believe you're right and this is not needed.

doisyg:
For reference, the generic topic muxer node has not been ported to ROS2 yet, see: https://github.com/ros-tooling/topic_tools/issues/3

doisyg:
Closing, after testing, a custom odom muxer makes much more sense.

Issue #2781, opened 2022-01-21
planner_server dies when clearing costmap: Segmentation fault in nav2_voxel_grid


janx8r:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - galactic from apt
- Version or commit hash:
  - a1e8ea3f578f6edc9a6048d3b94c6c2cb5755694 (branch galactic from this repo, but same for ros-galactic-navigation2 from apt)
- DDS implementation:
  - ?

#### Steps to reproduce issue
I use the package turtlebot3_navigation2 to let a simulated turtlebot navigate in an environment with many dynamic obstacles (simulated humans). In long simulations, where after reaching the goal a new goal is set again and again, this error happens sometimes (about once in 10 tries). Subjectively, the error usually happens when many dynamic obstacles come close to the robot.

#### Expected behavior
The planner_server should not crash when clearing the costmap.

#### Actual behavior
In various situations (but probably not all situations) where a costmap (both local and global) is to be cleared, planner_server crashes.
4s later all nodes bonded to it are stopped. The robot does not move anymore.

#### Additional information
Here are the outputs from all nodes except planner_server:
```
[controller_server-13] [INFO] [1642770375.384868426] [controller_server]: Passing new path to controller.
[controller_server-13] [INFO] [1642770376.434876989] [controller_server]: Passing new path to controller.
[controller_server-13] [INFO] [1642770377.435171565] [controller_server]: Passing new path to controller.
[controller_server-13] [INFO] [1642770378.484868835] [controller_server]: Passing new path to controller.
[controller_server-13] [INFO] [1642770379.484867185] [controller_server]: Passing new path to controller.
[lifecycle_manager-18] [INFO] [1642770385.536995074] [lifecycle_manager_navigation]: Have not received a heartbeat from planner_server.
[lifecycle_manager-18] [ERROR] [1642770385.537065639] [lifecycle_manager_navigation]: CRITICAL FAILURE: SERVER planner_server IS DOWN after not receiving a heartbeat for 4000 ms. Shutting down related nodes.

```

Here are the gdb outputs of planner_server:
```
[WARN] [1642770380.484649057] [planner_server]: GridBased: failed to create plan with tolerance 0.50.
[WARN] [1642770380.484748077] [planner_server]: Planning algorithm GridBased failed to generate a valid path to (4.00, 4.00)
[...]

PR #2784, opened 2022-01-25
Fixing the issue #2781: raytraceLine with same start and end point no…

janx8r:
… longer causes segmentation fault

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2781 |
| Primary OS tested on | Ubuntu 20.04, ROS galactic |
| Robotic platform tested on | Webots 2022a simulation of Turtlebot3 |

---

* in rare cases raytraceLine causes a segmentation fault
* this occurs if start and end point of the line are the same and the length of the line is therefore 0
* my changes now check if the length is greater than 0
* if the length is not greater than 0, some variables are set so that the voxel (= start point = end point) is processed anyway (i.e. either marked or deleted, depending on the context)

* none

---

#### For Maintainers: 

### DIFF
--- a/nav2_voxel_grid/include/nav2_voxel_grid/voxel_grid.hpp
+++ b/nav2_voxel_grid/include/nav2_voxel_grid/voxel_grid.hpp
@@ -232,12 +232,22 @@ public:
     if ((unsigned int)(dist) < min_length) {
       return;
     }
-    double scale = std::min(1.0, max_length / dist);
+    double scale, min_x0, min_y0, min_z0;
+    if(dist > 0.0){
+      scale = std::min(1.0, max_length / dist);
 
-    // Updating starting point to the point at distance min_length from the initial point
-    double min_x0 = x0 + (x1 - x0) / dist * min_length;
-    double min_y0 = y0 + (y1 - y0) / dist * min_length;
-    double min_z0 = z0 + (z1 - z0) / dist * min_length;
+      // Updating starting point to the point at distance min_length from the initial point
+      min_x0 = x0 + (x1 - x0) / dist * min_length;
+      min_y0 = y0 + (y1 - y0) / dist * min_length;
+      min_z0 = z0 + (z1 - z0) / dist * min_length;
+    }
+    // dist can be 0 if [x0, y0, z0]==[x1, y1, z1]. In this case only this voxel should be processed.
+    else{
+      scale = 1.0;
+      min_x0 = x0;
+      min_y0 = y0;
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
LGTM, waiting on Alexey which might still be a few days / a week. Sorry for the delay, just want to be transparent.

AlexeyMerzlyakov (commented):
LGTM: this unique point in this case will be marked by `bresenham3D()` algorithm rather than just ignoring it.

I'd also like to suggest to enhance already existing testcase by this corner situation, to track that this works correctly. Something like below:
```
diff --git a/nav2_voxel_grid/test/voxel_grid_bresenham_3d.cpp b/nav2_voxel_grid/test/voxel_grid_bresenham_3d.cpp
index 10ff23dd..49aa5830 100644
--- a/nav2_voxel_grid/test/voxel_grid_bresenham_3d.cpp
+++ b/nav2_voxel_grid/test/voxel_grid_bresenham_3d.cpp
@@ -49,6 +49,10 @@ public:
     ASSERT_TRUE(off < size_);
     data_[off] = val;
   }
+  inline unsigned int operator()(unsigned int off)
+  {
+    return data_[off];
+  }
 
 private:
   uint32_t * data_;
@@ -122,6 +126,29 @@ TEST(voxel_grid, bresenham3DBoundariesCheck)
   }
 }
 
+TEST(voxel_grid, bresenham3DSamePoint)
+{
+  const int sz_x = 60;
+  const int sz_y = 60;
+  const int sz_z = 2;
+  const unsigned int max_length = 60;
+  const unsigned int min_length = 0;
+  nav2_voxel_grid::VoxelGrid vg(sz_x, sz_y, sz_z);
+  TestVoxel tv(vg.getData(), sz_x, sz_y);
+
+  // Initial point
+  const double x0 = 2.2;
+  const double y0 = 3.8;
+  const double z0 = 0.4;
+
+  unsigned int offset = int(y0) * sz_x + int(x0);
+  unsigned int val_before = tv(offset);
+  // Same point to check
+  vg.raytraceLine(tv, x0, y0, z0, x0, y0, z0, max_length, min_length);
+  unsigned int val_after = tv(offset);
+  ASSERT_FALSE(val_before == val_after);
+}
+
 int main(int argc, char ** argv)
 {
   testing::InitGoogleTest(&argc, argv);
```

SteveMacenski:
@[janx8r](https://github.com/janx8r) It sounds like if you make those quick changes we can merge this!

Issue #2596, opened 2021-10-13
Add default BT logic for patience near goal


SteveMacenski:
If the robot is "close" to the goal and its path is blocked:
- Wait some time to see if it becomes unblocked
- Then replan to the goal
- repeat N times

The general idea is in blocked aisleways in warehouses / retail stores or general navigation where something temporary is blocking the robot's path. If in a narrow aisle and something in temporarily in the way, don't go out of the aisle, up another, and around, simply wait a hot second to see if it clears up (someone getting something in the aisle, transient leaving the aisle, picker, forklift, etc) 

CC  @DLu

padhupradheep:
Implemented the first version of this feature, Added a `obstacle_clearance_time` parameter to the `FollowPath` BT-Node, which would eventually allow us to specify the waiting time, if we see the obstacle. Below, we can see 2 possible scenarios:

1.  Obstacle does not clear at all, with `obstacle_clearance_time` to be 3 seconds: 
![image](https://user-images.githubusercontent.com/20242192/140328433-f5fcdf8e-0e24-4916-a652-22a26eb10396.png)

2. Obstacle clears (I just delete the obstacle in gazebo - Laziness at it's peak) and you can see the robot pass through the (could have been ideally the) same path:
![wiObstacle](https://user-images.githubusercontent.com/20242192/140328869-e09664f7-d7e3-4fc6-948b-dc7e02ccd0d6.gif)

Thoughts?

SteveMacenski:
That looks pretty good to me! What I had in mind was to create a new Behavior Tree node plugin that would do this instead, but implementing in the controller server could work as well. I think the benefit of it in the behavior tree is more configurability (e.g. be able to use the primitive node you create in other contexts than this) and would be a bit more "structured" to the autonomy application to have an explicit pause <if condition is met>. Our example is closness to goal, but someone could manipulate a BT node to set other conditions for the robot to pause (like in a dynamic setting).

padhupradheep:
> What I had in mind was to create a new Behavior Tree node plugin that would do this instead, but implementing in the controller server could work as well. I think the benefit of it in the behavior tree is more configurability (e.g. be able to use the primitive node you create in other contexts than this) and would be a bit more "structured" to the autonomy application to have an explicit pause . Our example is closness to goal, but someone could manipulate a BT node to set other conditions for the robot to pause (like in a dynamic setting).

[...]

PR #2787, opened 2022-01-26
adding an abstract BTNode for action cancellation

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2596  |
| Robotic platform tested on | MPO_700 simulation |

---

Adding an abstract BT node for handling and cancelling various servers. Also added a BT node derived from it for cancelling the controller server

---

Create new BT nodes for cancelling:
* planner server
* recover server
* costmap ...

#### For Maintainers: 


nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) 2022 Pradheep Padmanabhan, Neobotix GmbH
```
SteveMacenski, line 1: Requires test coverage.

Also, I think (?) the copyright is probably to your employer if you did this on company time / resources. We assign by copyright holder not author (so mine all say Samsung Research)

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,183 @@
...
+  : BT::ActionNodeBase(xml_tag_name, conf), action_name_(action_name)
+  {
+    node_ = config().blackboard->template get<rclcpp::Node::SharedPtr>("node");
+    callback_group_ = node_->create_callback_group(
+      rclcpp::CallbackGroupType::MutuallyExclusive,
+      false);
+    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
+
+    // Get the required items from the blackboard
+    bt_loop_duration_ =
```
SteveMacenski, line 56: You don't actually use this variable

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,183 @@
...
+  static BT::PortsList providedPorts()
+  {
+    return providedBasicPorts({});
+  }
+
+  /**
+   * @brief Function to perform some user-defined operation on tick
+   * Could do dynamic checks, such as getting updates to values on the blackboard
+   */
+  virtual void on_tick()
```
SteveMacenski, line 124: Is there a need for this if we're simply canceling a goal? There are no fields associated with cancelation so there's nothing for a user to specify

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,183 @@
...
+  {
+  }
+
+  /**
+   * @brief The other (optional) override required by a BT action.
+   */
+  void halt() override
```
SteveMacenski, line 131: I don't believe you have to override this if you have not state you need to reset

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,183 @@
...
+  {
+    // setting the status to RUNNING to notify the BT Loggers (if any)
+    setStatus(BT::NodeStatus::RUNNING);
+
+    // user defined callback
+    on_tick();
+
+    // Cancel all the goals specified before 10ms from current time
+    // to avoid async communication error
+    rclcpp::Time goal_expiry_time_(rclcpp::Clock().now().nanoseconds() - 10000000,
```
SteveMacenski, line 149: Or simply `rclcpp::Time var_name = node_->now() - 10ms` (or something like that) if you use the chrono types. No need to specify the clock type, the `node_`'s clock will have that depending on sim or real world time.

SteveMacenski: Also don't need the underscore at the end, its not a class member

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,183 @@
...
+  rclcpp::Node::SharedPtr node_;
+  rclcpp::CallbackGroup::SharedPtr callback_group_;
+  rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
+
+  // The timeout value while waiting for response from a server when a
+  // new action goal is canceled
+  std::chrono::milliseconds server_timeout_;
+
+  // The timeout value for BT loop execution
+  std::chrono::milliseconds bt_loop_duration_;
```
SteveMacenski, line 178: this can be removed if not being used

SteveMacenski (changes requested):
Generally looks like the right approach, just needs a little polishing!

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
padhupradheep: Even, there was nothing I could think of. Just left it out of confusion for your review. Will remove it.

SteveMacenski: Ok! Airing on the side of configurable is good, I appreciate that. Always happy to readd it if we can even think of something someone might want to do with it.

nav2_behavior_tree/test/plugins/action/test_bt_cancel_action_node.cpp
```diff
@@ -0,0 +1,199 @@
...
+      <root main_tree_to_execute = "MainTree" >
+        <BehaviorTree ID="MainTree">
+            <Fibonacci/>
+        </BehaviorTree>
+      </root>)";
+
+  tree_ = std::make_shared<BT::Tree>(factory_->createTreeFromText(xml_txt, config_->blackboard));
+  auto send_goal_options = rclcpp_action::Client<test_msgs::action::Fibonacci>::SendGoalOptions();
+  auto goal_msg = test_msgs::action::Fibonacci::Goal();
+
```
SteveMacenski, line 168: This never actually 1) cancels a goal or 2) tests the behavior tree node abstraction you added above. That or any derivative of it needs to be inputted into the behavior tree to even call cancel and exercise the code you added. That's what we need test coverage over. We know that ROS action servers work by this point in the test coverage. Its about exercising and making sure the specific code you added is functioning properly

SteveMacenski (changes requested):

nav2_behavior_tree/test/plugins/action/test_bt_cancel_action_node.cpp
padhupradheep: okay, gottcha. Let me come up with a better one.

padhupradheep: Ah, I already have my cancel control node and a test for it. 

> so that your tests have tangible complete nodes to test in the BT itself.

Does that mean we can simply avoid avoid a test node for the abstract class? 

BTW, the abstract class test was just inspired from the `test_bt_action_node`. Maybe I just misunderstood the test coverage as you said.

SteveMacenski: Yeah, if the control node derives from this abstraction, then it would exercise this code

padhupradheep: But wait the Fibonacci node already derives from the abstraction. Maybe, the naming should have been `CancelFibonacci `

Let me quickly explain the steps I do in the test: 

* I created a derivative of the `bt_cancel_action_node` in a dummy `Fibonacci` BT node (maybe the naming confused). 
* Create a server for the Fibonacci action
* Create a BT consisting of dummy Fibonacci node, and sent a dummy goal through `async_send_goal()`
* Finally, I just execute the tick for the BT consisting of the  Fibonacci node derived from  `bt_cancel_action_node` then we just cancel the goal through the Fibonacci node (again I named it wrongly, it should have been something like `CancelFibonacci`). Then for the test, I just check if the node returns `SUCCESS` 

We have the abstraction derived right? Sorry for troubling you again. Just wanted to clarify, so that I can make much better tests for someother issue, someother time.

padhupradheep:
Added a `cancel_control` node that derives from the abstraction of `bt_cancel_action_node`

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -0,0 +1,169 @@
...
+    action_client_->wait_for_action_server();
+  }
+
+  /**
+   * @brief Any subclass of BtCancelActionNode that accepts parameters must provide a
+   * providedPorts method and call providedBasicPorts in it.
+   * @param addition Additional ports to add to BT port list
+   * @return BT::PortsList Containing basic ports along with node-specific ports
+   */
+  static BT::PortsList providedBasicPorts(BT::PortsList addition)
```
SteveMacenski, line 101: https://github.com/ros-planning/navigation2/blob/63feaf70780d1d41a694d6b53aadc6005cdb7ce2/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/back_up_action.hpp#L52-L61

Do something in this style rather than the 2 functions, its a little expanded and not in the style of the other ones

SteveMacenski: Then you don't need to implement one of these functions, I don't believe

nav2_behavior_tree/test/plugins/action/test_controller_cancel_node.cpp
```diff
@@ -0,0 +1,167 @@
...
+  : TestActionServer("follow_path")
+  {}
+
+protected:
+  void execute(
+    const typename std::shared_ptr<rclcpp_action::ServerGoalHandle<nav2_msgs::action::FollowPath>>
+    goal_handle)
+  {
+    while (!goal_handle->is_canceling()) {
+      // waiting here until goal cancels
```
SteveMacenski, line 39: Add a sleep here, a while empty condition will explode and consume all available resources. A sleep for 10ms will give the scheduler time to do other tasks

SteveMacenski (changes requested):
Great improvement in the tests! Just a little cleanup and this is good to merge

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
padhupradheep: Unfortunately, that will not be possible because, `back_up_action` was derived from the `bt_action_node`. Also, on top of the `providedBasicPorts`, we can find a comment stated as follows

https://github.com/ros-planning/navigation2/blob/63feaf70780d1d41a694d6b53aadc6005cdb7ce2/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L96-L101

padhupradheep:
Thank you very much for all the patient reviews.. I have added all the necessary navigation.ros.org documentation in [Pr: 267](https://github.com/ros-planning/navigation.ros.org/pull/267).

nav2_behavior_tree/test/plugins/action/test_controller_cancel_node.cpp
```diff
@@ -0,0 +1,165 @@
...
+  CancelControllerServer()
+  : TestActionServer("follow_path")
+  {}
+
+protected:
+  void execute(
+    const typename std::shared_ptr<rclcpp_action::ServerGoalHandle<nav2_msgs::action::FollowPath>>
+  )
+  {
+    std::this_thread::sleep_for(10ms);
```
SteveMacenski, line 38: ... readd your while loop, in the body of the while loop, add a sleep. So `while (...) {sleep}`

SteveMacenski (changes requested):

SteveMacenski:
https://app.codecov.io/gh/ros-planning/navigation2/blob/d9f6a5b7059326204c19e4d5b7f5dc1177afc41b/nav2_behavior_tree/plugins/action/controller_cancel_node.cpp

it doesn't look like your node is properly registering for some reason. Also you have linting issues

```
build/namespaces [5] (/opt/overlay_ws/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp:27)
nav2_behavior_tree.cpplint

Do not use namespace using-directives.  Use using-declarations instead.

whitespace/blank_line [3] (/opt/overlay_ws/src/navigation2/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp:164)
nav2_behavior_tree.cpplint

Redundant blank line at the end of a code block should be deleted.
```

padhupradheep:
> it doesn't look like your node is properly registering for some reason.

Strange! Let me see, what's causing it.

padhupradheep:
Did further code coverage analysis locally and found that, even for the other BT's the percentage of code coverage will be lesser and the node will not be registered, if we do not run the `nav2_system_tests`. This is an obvious case, because for each of the BT nodes, we do the node registration in their corresponding tests using a factory and that is the reason.

Here are some local results comparing the line coverage for the `nav2_behavior_tree` package before and after running the tests for the `nav2_system_tests`:

before_bt_cov
after_bt_cov

Hope my points make sense. Let me come with the next step of actions hopefully tonight or earlier next week. I think it is better we add the test coverage to the `nav2_system_tests` in this PR itself, so we do not have to think about it anymore and concentrate on the next tasks.

SteveMacenski:
Ah, try I forgot you must add these to each of the YAML files in nav2-bringup and nav2 system tests so that they can be dynamically loaded if a BT uses it. You also need to add it to the BT navigator's default set of plugin libs. See the yaml files for the styling it appears in, that may be the reason.

There's something odd here given that every other BT node has these lines covered even if they're not being used actively in system tests. I think maybe because its not being loaded? Or something you missed in the CMake?

SteveMacenski:
I'm looking at our combined coverage from codecov we use to track our coverage metrics. https://app.codecov.io/gh/ros-planning/navigation2 

I don't really mind where the coverage comes from, but your nodes need to have it before we can merge this PR. Please add in the yaml files and defaults.

padhupradheep:
Let me figure out a way, need some more time in this.

padhupradheep:
Now it's 83.33%

SteveMacenski (approved):

SteveMacenski:
Yup!  The 1 line is OK, I just need to know that something is trying to dynamically load it so that we know its all registered properly (and that the defaults in the main yaml file / bt navigator have all of the options built it) This is good to go! We can continue in the other PR :+1:

PR #2790, opened 2022-01-27
FIx wrong  path segment selection in RPP and DWB

### LINKED ISSUE #2622: [RPP/DWA/TEB] When NavThroughPose has an intersecting loop, it can occasionally pick the wrong path segment
I suspect its for one of a few reasons
- If it does "skip", its because the path is fully contained within the local costmap, so the pruning doesn't get rid of the future-looping section and then the "cross" point later in the path is technically closer to the lookahead requested point than the next sequential point based on some very small amount closer (<1mm)
- The fact that it jumps to that part once the robot hits that "cross" point makes me think its actually an issue in transforming the path, where the point closest to the robot becomes the other side of the "cross" and so it uses that (<1mm).

There's a couple of potential fixes:
- When we [find the points closest to the robot to start transforming the path from](https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L590-L595), there should be some tolerance, like 0.01 or something, that if there are multiple points within that distance range of each other, use the first in the path rather than the closest one. That should help it pick the right part of the cross
- Rather than only [checking when it leaves the costmap to prune til](https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L598-L602), also add a path distance element, once the path length from the closest pose is larger than N meters, also prune past those points. That way if there's a cross in the local costmap, as long as we pick the first point, we should prune out the rest

I suppose [this part](https://github.com/ros-planning/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L625) in erasing the passed parts should be a solution if there is no replanning (so we don't infinitely loop cauased by selecting the "first" part of the intersection in the path). So I think doing those 2 points should cover that edge case completely.

I'm pretty sure that DWA and TEB would have this exact same problem, but its something we can do something about in RPP much more easily.

sathak93:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #2622 |
| Robotic platform tested on | gazebo simulation  turtlebot) |

---

* This PR fix #2622 bug in RPP and DWB controller by slicing the global plan path using integrated path distance.

---

* isGoalreached() in goal checker  should be improvised.
* fix this bug in TEB.

#### For Maintainers: 


SteveMacenski:
Should I take a look at this or wait until you work on the DWB part?

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -86,6 +87,22 @@ inline double euclidean_distance(
...
+ * @return double L2 distance
+ */
+inline double euclidean_distance(
+  const geometry_msgs::msg::Pose2D & pos1,
+  const geometry_msgs::msg::Pose2D & pos2)
+{
+  double dx = pos1.x - pos2.x;
+  double dy = pos1.y - pos2.y;
+
+  return std::sqrt(dx * dx + dy * dy);
```
SteveMacenski, line 103: hypot would be better, it handles nan/inf conditions better

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -107,6 +124,25 @@ inline Iter min_by(Iter begin, Iter end, Getter getCompareVal)
...
+ * Find first element in iterator that is greater integrated distance than comparevalue
+ */
+template<typename Iter, typename Getter>
+inline Iter first_element_beyond(Iter begin, Iter end, Getter getCompareVal)
+{
+  if (begin == end) {
+    return end;
+  }
+  Getter dist = 0.0;
+  for (Iter it = begin; it != end - 1; it++){
```
SteveMacenski, line 137: There are some linting problems here, check CircleCI to find them. You can reproduce locally if you run `ament_cpplint` and `ament_uncrustify` to have the right ROS2 formatting style

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -467,47 +458,62 @@ DWBLocalPlanner::transformGlobalPlan(
...
-    [&](const auto & global_plan_pose) {
-      return getSquareDistance(robot_pose.pose, global_plan_pose) > sq_transform_end_threshold;
-    });
+  // Find the first pose in the end of the plan that's further than transform_end_threshold
+  // from the robot using integrated distance
+  auto transformation_end =
+    nav2_util::geometry_utils::first_element_beyond(
+    transformation_begin, global_plan_.poses.end(), transform_end_threshold);
+
+  for (auto it = transformation_begin + 1; it != transformation_end - 1; ++it) {
```
SteveMacenski, line 506: I'm not quite understanding why this was added. It looks like a dot product finding a directional change but not really a comment about why or what problem its solving

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -606,20 +601,22 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
   // First find the closest pose on the path to the robot
   auto transformation_begin =
     nav2_util::geometry_utils::min_by(
-    global_plan_.poses.begin(), global_plan_.poses.end(),
+    global_plan_.poses.begin(), lookahead_element,
     [&robot_pose](const geometry_msgs::msg::PoseStamped & ps) {
       return euclidean_distance(robot_pose, ps);
     });
 
-  // Find points definitely outside of the costmap so we won't transform them.
```
SteveMacenski, line 617: I think this was the right implementation, since we want to find the first point that's outside of the statically set costmap size. This isn't an integral distance but a real-constraint distance dependent on the costmap size. This could potentially not prune quite to the end of the costmap (dropping necessary data for longer look-ahead points) for wavy paths. There's no real benefit to culling a few points more off the end of the rolling costmap window.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -267,15 +267,10 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Find look ahead distance and point on path and publish
   double lookahead_dist = getLookAheadDistance(speed);
 
-  // Check for reverse driving
-  if (allow_reversing_) {
```
SteveMacenski, line 271: Why remove the allow_reversing function? That is necessary so that we only do this if its appropriate to be rotating in place due to a sudden change in pose, that shouldn't normally be turned on unless letting the robot move backwards is allowable by that parameterization

SteveMacenski (changes requested):

sathak93:
This bug https://github.com/ros-planning/navigation2/issues/2622 is due to either of 

1. min_by function used to pick [transformation_begin](https://github.com/ros-planning/navigation2/blob/5ca6c511120d4dd612d14fd944fca7eba5e19b83/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L610) pose - when the path overlaps, sometimes it picks the later poses than immediate ones which leads  to deleting of  poses. in this pr  i changed the range from global plan begin to look_ahead point  instead of global plan begin to global plan end  so that the immediate ones will be always selected.

2. find_if algorithm in [getlookaheadpoint](https://github.com/ros-planning/navigation2/blob/9c6fc02f95e8c51212cb7b2af12f12fb81c43b3b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L373) picks wrong pose if the transformed plan has some poses in front of the robot and some back to the robot. in this pr, look_ahead point in progressively reduced when nearing the cusp using the already defined function for reversing behavior. And  the transformed_plan ending  is selected using the integrated path distance instead of absolute distance

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
sathak93: if the transformed_plan is very large and has poses ahead and behind the robot as in overlapping case, the getlookahead point may select the point behind the robot making the robot leave some path and turns back immediately.
for this it is trimmed to a particular distance from transform begin using integrated distance. 

if we need this to be remain unchanged,  we can modify and get the lookahead point using integrated distance.

SteveMacenski: OK that makes sense, but then this value shouldn't be derivative of the costmap size as it currently is e.g.

```
  const double max_transform_dist = max_costmap_dim * costmap->getResolution() / 2.0;
```

This should probably now be a parameterized distance

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Again, makes sense thanks for the clarification.

SteveMacenski:
@Aposhian please let us know your thoughts :smile:

Aposhian:
Here is the difference between the stock `galactic` controller, and building this PR for rolling. The green panels are the global plan (which is being updated by a custom planner), and the `base_link` TF can be seen cruising along, with the lookahead point being updated in red.
current-controller
n
When I try this PR, the robot approaches the initial lookahead point, and then stops.
new-controller

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Aposhian: I agree that finding the closest point needs to have an upper bound, so that it doesn't select overlapping poses at a later part of the path. Also noting that we are implicitly relying on the lower bound given by pruning the global plan after receiving it, so we don't process the same segments twice.

Aposhian: I agree with @SteveMacenski that we shouldn't prune by lookahead distance: this part of the controller happens before the lookahead distance is computed, right?

I think that we should simply prune by the minimum of the cusp distance and the costmap size. That guarantees that you won't be able to get portions of the path that have looped around.

SteveMacenski:
I think this may relate to the fact you're giving it a hilariously small path, this is handling the looping for you so you shouldn't need to give it something so small. Though this is clearly an issue, are you sure that the controller is getting those path updates? 

@sathak93 why would this be?

sathak93:
@SteveMacenski  i will check this pr for very small path. also i think it is best to validate RPP of this using main branch rolling. RPP in galactic branch and in rolling branch  has many differences. using rolling we can focus problems related to this pr.

SteveMacenski:
I think @Aposhian did use rolling's version of RPP to see that issue. I think he was just highlighting the difference with a baseline (no?)

Aposhian:
Yes, the short path that is being given has been our workaround to the bug that we are trying to fix here :smile:, so that it something that we should hopefully be able to remove, but I don't think that this fix should cause a regression for short paths. I will test on nav2 `main` to compare as well, and will try removing our intermediary path segment selection to see how this fix addresses our problem.

Aposhian:
Here is the performance against the `main` branch: it doesn't exhibit the stalling behavior that I am seeing with this PR.
nav2-main-controller

Aposhian:
I have made some modifications on my own fork. But it has occured to me: if we implement an upper bound to prevent jumping to future points, then we are implicitly increasing our reliance on the lower bound provided by the `global_plan` pruning. What this means that if we make the change this way, it may break backward compatibility with some users' expectations by removing the controllers capability to "snap" to any point on the path, provided by the current behavior of just grabbing the closest pose to the robot.

"snapping" is fine and easy for paths that don't overlap, but with overlapping paths, you need to make sure that you always prioritize earlier portions of the path over later portions, and it may be hard to do that in a deterministic but also intuitive way.

There are a few ways I think we could solve this problem:
- add a parameter to give users control on whether they want "snapping" or being able to do overlapping paths (nonideal)
  - this would essentially work by either starting the local plan at the first pose in the global plan (enforced lower bound) or using the current behavior of snapping to the absolute closest waypoint
- instead of finding the absolute closest waypoint, maybe find the minimum (by index) waypoint within a predefined "search radius". This would support snapping (within reason)
  - this poses the following tradeoff: a greater search radius provides a larger snapping distance, but may not snap to the closest waypoint
- use a heuristic of minimum waypoint index and distance between robot and pose to determine which waypoint to snap to.

Aposhian:
Take the following global plan for example:
blue are poses in the global plan, and red is the robot pose.
snapping
e

Now, assuming that nav2 has been incrementally pruning the start of the global plan, in this situation, the controller should snap to pose `2`. But how do you describe that as an algorithm? Closest pose won't work, since that would snap to pose `10`. If you do a simple search radius paired with minimum waypoint id, then if `search_radius = 4`, then it will snap to `2`. But if `search_radius = 5`, then by the same algorithm, it would snap to pose `1`, which is not ideal.

Aposhian:
Another option would be be to get the closest waypoint between the start of `global_plan` and the 180 degree turnaround. That should snap to pose `2`, and that should support many, but not all cases where people want to snap to paths that don't overlap.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Yes, this happens before any of the lookahead logic is applied. this is just pruning the path forward to only consider things in a local window forward

Aposhian:
Sorry if my comments were confusing, since I was bringing up a few different alternatives. I do agree that parameterized closest pose is better than using a search radius, which is not as robust. I think the bound should be up to the local costmap OR the nearest turnaround point.
>This is where maybe your wording is confusing me. It should always select the path point that it was tracking in time = t-1 plus 1-N waypoint markers depending on the speed and density of the path. It should be going sequentially.

I suppose most users would not experience this "snapping" behavior since they use planners that always start the global plan from the current robot pose. Our planner doesn't necessarily do this, and this is why we have observed the "snapping" behavior since it just grabs the closest pose.

My modifications are here: https://github.com/ros-planning/navigation2/pull/2812

SteveMacenski:
@sathak93 thank you kindly for jump starting this discussion and developing the core ideas to solve this problem. It was very clever and didn't occur to me and a few others so we really appreciate you jumping in and I hope to have your help again in the future!

I'm going to close this ticket since #2812 supersedes with a slight variation on your technique to work better with another user's experiences to make sure this can work for everyone's needs

Issue #2415, opened 2021-06-18
Additional features for the Python3 Simple Navigator Commander


SteveMacenski:
- [x] calling the controller 
- [x] Add default arguments for full action API to functions (that don't necessarily need to be provided, but if they are, use them. example: behavior tree xmls for nav requests)
- [x] Allowing the wait for navigation to bring up to have a configurable set of nodes to wait for (and default to the ones currently present) 
- [x]  Call common recoveries
- [ ] Constructor to take in optional namespaces/topics to use for creating the clients and subscribers
- [ ] setting parameters

jacobwaller:
To add on to 

> Allowing the wait for navigation to bring up to have a configurable set of nodes to wait

as far as I can see, simple navigator does not work if SLAM-Toolbox is doing the localization because of its reliance on the nav2_amcl node

SteveMacenski:
Yeah, that's part of what that should do, have options for the nodes to wait on and not assume which are in place if the user can provide better estimates.

oldtopos:
Do not understand "calling the controller" and "Call common recoveries"

wait for navigation refers to function "waitUntilNav2Active" ?

ComputePathToPose and ComputePathThroughPoses take planner_id and use_start, should these be given overrides similar to bt override ?

SteveMacenski:
> Do not understand "calling the controller" and "Call common recoveries"

Its the analog to [getPlan](https://github.com/ros-planning/navigation2/blob/main/nav2_simple_commander/nav2_simple_commander/robot_navigator.py#L207-L233) by calling the planner server, but for local trajectory planning via the controller server. See the `FollowPath.action` in `nav2_msgs`: https://github.com/ros-planning/navigation2/tree/main/nav2_msgs/action

Call common recoveries is the same, have `wait()` `spin()` `backup()` APIs calling their appropriate action servers (see the `nav2_msgs` link above).

With planning/control/recoveries, a Python application can have essentially full control over Nav2 without the Behavior Tree to create a simple application if they like or call servers to compute particular values for debug / application needs.

tonynajjar:
I'm working on adding wait, spin and backup. Spin and backup make sense to me but I'm not sure how wait is expected to work in a python script. As far as I understand in a BT, the wait node would stall the BT. In a python script, wait behave like a time.sleep()?

SteveMacenski:
[...]

PR #2792, opened 2022-01-29
Add recoveries to simple commander

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Part of [2415](https://github.com/ros-planning/navigation2/issues/2415) |
| Robotic platform tested on | Gazebo Simulation |

---

* I added wait and backup action client calls
* Fixed a typo in NavigationResult Enum

* Added docs in this PR and https://github.com/ros-planning/navigation.ros.org/pull/268

---

#### For Maintainers: 


tonynajjar:
I don't think the tests fail because of my changes

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -155,6 +160,47 @@ def followWaypoints(self, poses):
         self.result_future = self.goal_handle.get_result_async()
         return True
 
+    def spin(self, spin_dist=1.57, time_allowance=10):
```
SteveMacenski, line 163: Can you add an example python file for these in use? 

Also, might be worth changing the API to `isComplete()` from `isNavComplete()`  since its being used in so many more contexts now (and update the other files that use it)

tonynajjar: Sure I can add a demo using spin and wait calls. Do you have an idea for a meaningful demo scenario using wait and spin? Otherwise I would make a simple script that uses them (e.g navigate in front of a wall, backup, spin, navigate somewhere else). then again that might just look like a cheap/fake way of demo-ing recoveries.

SteveMacenski: That's fine to me, its more about showing how to use them / get info on feedback than making the demo hyper realistic.

SteveMacenski:
Can you rename `isNavComplete()` to `isTaskComplete()`  and `cancelNav()` to `cancelTask()` since we now use those functions for any running action which now includes recoveries? You'd need to update uses of them in the demo code & documentation

tonynajjar:
Oops, missed that. Sure will do!

nav2_simple_commander/nav2_simple_commander/demo_recoveries.py
```diff
@@ -0,0 +1,104 @@
...
+# limitations under the License.
+
+from geometry_msgs.msg import PoseStamped
+from nav2_simple_commander.robot_navigator import BasicNavigator, NavigationResult
+import rclpy
+from rclpy.duration import Duration
+
+
+"""
+Basic stock inspection demo. In this demonstration, the expectation
```
SteveMacenski, line 23: Please update this comment to correspond to what the content of this file is

nav2_simple_commander/nav2_simple_commander/demo_recoveries.py
```diff
@@ -0,0 +1,104 @@
...
+        i += 1
+        feedback = navigator.getFeedback()
+        if feedback and i % 5 == 0:
+            print(
+                f'Estimated time of arrival to destination is: \
+                {Duration.from_msg(feedback.estimated_time_remaining).nanoseconds / 1e9}'
+            )
+
+    # Robot hit a dead end, back it up
+    navigator.backup(backup_dist=0.5, backup_speed=0.1)
```
SteveMacenski, line 67: Add some prints before backing up so in the terminal they see a "starting to back up..." and spin messages for context of what's happening on their screen

nav2_simple_commander/nav2_simple_commander/demo_recoveries.py
```diff
@@ -0,0 +1,104 @@
...
+        feedback = navigator.getFeedback()
+        if feedback and i % 5 == 0:
+            print(f'Spin angle traveled: {feedback.angular_distance_traveled}')
+
+    result = navigator.getResult()
+    if result == NavigationResult.SUCCEEDED:
+        print('Dead end confirmed! Returning to start...')
+    elif result == NavigationResult.CANCELED:
+        print('Recovery was canceled. Returning to start...')
+        exit(1)
```
SteveMacenski, line 91: Exit(1) doesn't look like returning to start :wink:

tonynajjar: Looks like you found another bug 
https://github.com/ros-planning/navigation2/blob/d83f9d995326f3aa1fb76c7cbe879ab783d95955/nav2_simple_commander/nav2_simple_commander/demo_inspection.py#L87

Will fix it.

tonynajjar:
Regarding the renaming of isNavComplete and cancelNav, two questions:

1- Am I right that it should not be renamed in the [migration to Humble doc](https://github.com/tonynajjar/navigation.ros.org/blob/master/migration/Galactic.rst) because this won't be part of Humble?

2- What about NavigationResult Enum? Shouldn't it also be renamed to TaskResult?

SteveMacenski:
> 1- Am I right that it should not be renamed in the [migration to Humble doc](https://github.com/tonynajjar/navigation.ros.org/blob/master/migration/Galactic.rst) because this won't be part of Humble?

That should be updated for the new API you're making :smile: since `main` is going to roll into `Humble` when that is released later this year

> 2- What about NavigationResult Enum? Shouldn't it also be renamed to TaskResult?

Sure! Good catch

tonynajjar:
Please re-check along with https://github.com/ros-planning/navigation.ros.org/pull/273

SteveMacenski:
Awesome merged!

I think the next best one is 

>  Allowing the wait for navigation to bring up to have a configurable set of nodes to wait for (and default to the ones currently present)

Basically,  the wait for navigation to initialize function should by default want for BT nav and AMCL, but you should allow the user to specify others in case they're not using AMCL or another highlevel navigator name. So I think you should add 2 fields that default to BT nav and AMCL `navigator` and `localizer` so that the user can specify others if they would like to wait for

tonynajjar:
Why is it considered that Nav2 is active when BT nav and AMCL are active? Are these deterministically the last nodes to spawn?

Issue #2796, opened 2022-02-02
Couldn't load critics when launching in namespace


Fellfalla:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04
- ROS2 Version: Galactic and installed via `apt`
- Version or commit hash: 1.0.8-1focal.20211223.041721
- DDS implementation: default

#### Steps to reproduce issue

Run `ros2 launch bringup navigation.launch.py namespace:=ns` with following launch file and the default `nav2_params.yaml`

```
# Source: https://github.com/ros-planning/navigation2 @ d1ed5e466286898abebbc1bf80dd51e009670fe5

import os

from ament_index_python.packages import get_package_share_directory

from launch import LaunchDescription
from launch.actions import DeclareLaunchArgument, SetEnvironmentVariable
from launch.substitutions import LaunchConfiguration
from launch_ros.actions import Node
from nav2_common.launch import RewrittenYaml


def generate_launch_description():
    # Get the launch directory
    bringup_dir = get_package_share_directory('bringup')

    namespace = LaunchConfiguration('namespace')
    use_sim_time = LaunchConfiguration('use_sim_time')
    autostart = LaunchConfiguration('autostart')
    params_file = LaunchConfiguration('params_file')
    default_bt_xml_filename = LaunchConfiguration('default_bt_xml_filename')
    map_subscribe_transient_local = LaunchConfiguration('map_subscribe_transient_local')

    lifecycle_nodes = ['controller_server',
                       'planner_server',
                       'recoveries_server',
                       'bt_navigator',
                       'waypoint_follower']

    # Map fully qualified names to relative ones so the node's namespace can be prepended.
    # In case of the transforms (tf), currently, there doesn't seem to be a better alternative
    # https://github.com/ros/geometry2/issues/32
    # https://github.com/ros/robot_state_publisher/pull/30
    # TODO(orduno) Substitute with `PushNodeRemapping`
    #              https://github.com/ros2/launch_ros/issues/56
    # remappings = [('/tf', 'tf'),
    #               ('/tf_static', 'tf_static')]
    remappings = []

    # Create our own temporary YAML files that include substitutions
    param_substitutions = {
        'robot_base_frame': [namespace, '_base_link'],
        'odom_frame_id': [namespace, '_odom'],
        'local_costmap.local_costmap.ros__parameters.global_frame': [namespace, '_odom'],
        'recoveries_server.ros__parameters.global_frame': [namespace, '_odom'],
[...]

PR #2798, opened 2022-02-03
fix namespaces for bringup

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2796 |

---

* Added the namespace parameter as part of the Node. 

---

#### For Maintainers: 


padhupradheep:
This is a fix mainly for the people who wants to add a namespace and do the `navigation_launch.py` as a standalone.

SteveMacenski:
How does this interact with using the bringup's push into namespace? https://github.com/ros-planning/navigation2/blob/galactic/nav2_bringup/bringup/launch/bringup_launch.py#L80-L82 that should be doing all the heavy lifting here. I think adding that there would apply a double namespace `/ns/ns/planner_server`. How can we make sure that doesn't occur?

padhupradheep:
> that should be doing all the heavy lifting here. I think adding that there would apply a double namespace /ns/ns/planner_server

Nope, it's indeed not pushing twice. I can confirm that.

SteveMacenski:
From `bringup` being set as `namespace` so its both being pushed in `bringup` and set in `navigation`'s Node() call? That's surprising behavior

padhupradheep:
> From bringup being set as namespace so its both being pushed in bringup and set in navigation's Node() call?
Atleast with the tests from this PR, that's how it is working.

padhupradheep:
I feel something like this is happening as shown in the below lines:

https://github.com/ros2/launch_ros/blob/835b40c47cfd84083a510baea510052148d3d406/launch_ros/launch_ros/utilities/namespace_utils.py#L59-L60

SteveMacenski:
Where is the absolute namespace with the leading `/` in all of this? I don't see any arguments with that to get that potential outcome.
https://github.com/ros2/launch_ros/blob/d1c497878274a4ea6a1c1865fe74e395b961ba1f/test_launch_ros/test/test_launch_ros/actions/test_push_ros_namespace.py#L61-L64

 this test case for the Push API in launch looks to me what's happening. We have a node namespace set in `Node()` and a Push namespace showing we should get both. Unless I'm misinterpreting what `node_ns` is supposed to be otherwise (what's defined in code?)

padhupradheep:
> Just to verify you are also setting use_namespace to true along with setting the namespace too so the push + the namespace field in the Node() are used. 

Oops, I missed the `use_namespace` args. When I use it, I get a double namespace as you showed /ns/ns/planner_server . Damn, sorry for the confusion, I missed the above argument and wasted some time on it.

SteveMacenski:
No worries at all, this is why we ask questions. I appreciate you taking a look at this issue to help find a solution! 🙂

sounds like we can close this unless you had another idea? Perhaps if we Push namespace we should set the namespace we pass to empty string? Or maybe do this outright and remove the use of Push? The major use of that is for the multirobot case

padhupradheep:
> Perhaps if we Push namespace we should set the namespace we pass to empty string?

This can be an option for us, which leads us to the removal of `use_namespace` parameter in `nav2_bringup.`

SteveMacenski:
OK, I think there's just 2 things you need to make sure works
- Setting namespace from CLI works
- Using the multirobot launch API, it continues to throw everything in the appropriate namespace

padhupradheep:
On further research, removing `use_namespace` from the `nav2_bringup` will have a severe consequence on RViz launch for multi robot navigation. Especially here:

https://github.com/ros-planning/navigation2/blob/7f0e08746915b7ad64caf1e2c008c4109a981367/nav2_bringup/launch/rviz_launch.py#L57

 I would suggest, we keep the stones unturned.

SteveMacenski:
OK - understood. Closing (unless you don't mean that, happy to reopen!)

PR #2802, opened 2022-02-04
Adding a new BT node Is_path_longer_on_approach

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2596 |
| Robotic platform tested on | Gazebo Simulation of Neobotix MPO-700 |

---

* New BT Nodes - IsPAthLongerOnApproach
* Added Tests
* Updated the default `navigate_to_pose_w_replanning_and_recovery` with the feature request discussed in #2596 

* Explain the newly developed BT node and it's parameters
* Explain the updated BT
* Add these to the navigation plugins page, migration guide, and add new pages in the behavior tree configuration guide for the newly created node

---

#### For Maintainers: 


SteveMacenski:
I think you know the drill on docs, see the table above in the PR description for what docs to add for this in the website

nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.cpp
```diff
@@ -0,0 +1,73 @@
...
+inline BT::NodeStatus IsPathLongerOnApproach::tick()
+{
+  getInput("path", new_path_);
+  getInput("prox_leng", prox_leng_);
+  getInput("length_factor", length_factor_);
+
+  // Check if the old path is not same with the current one
+  if (new_path_ != old_path_ && old_path_.poses.size() != 0 && new_path_.poses.size() != 0) {
+    // Calculate and compare the old and the new path length, given the goal proximity
+    if ( (nav2_util::geometry_utils::calculate_path_length(new_path_, 0) >= prox_leng_) &&
```
SteveMacenski, line 42: This if condition has too much going on in it and its hard to follow. Break this down into a  few well-named `bool` variables and check the correct bool states in the if condition.

SteveMacenski, line 42: Wouldn't it be `nav2_util::geometry_utils::calculate_path_length(new_path_, 0) < prox_leng_` so that we only apply it when we're within the proximity to the goal?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.hpp
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<nav_msgs::msg::Path>("path", "Path to follow"),
+      BT::InputPort<double>(
+        "prox_leng", 1.0,
```
SteveMacenski, line 53: I'd think we'd want to apply this far before 1m away. Maybe 5m? What do you think is a good default?

Also the description isn't very clear, I'd do something like "Integrated path length proximity to the goal pose to apply the behavior (m)"

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.hpp
```diff
@@ -0,0 +1,78 @@
...
+        "Proximity length (m) for the path to be longer on approach"),
+      BT::InputPort<double>(
+        "length_factor", 1.0,
```
SteveMacenski, line 56: Maybe do something like 2.0, 1.0 seems like small variations would trigger this

SteveMacenski (changes requested):

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
```diff
@@ -14,10 +14,20 @@
             <ClearEntireCostmap name="ClearGlobalCostmap-Context" service_name="global_costmap/clear_entirely_global_costmap"/>
           </RecoveryNode>
         </RateController>
-        <RecoveryNode number_of_retries="1" name="FollowPath">
```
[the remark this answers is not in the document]
padhupradheep: Sure, we can have a separate BT. For now, I've tested it remotely. But, I will do verify it on the robot, maybe by end of this week.

nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.cpp
padhupradheep: Yes, you are right. I will update it.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.hpp
padhupradheep: anywhere around 3-5m will be a good choice. So, as a default 3m sounds good.

padhupradheep:
FYI, created a new BT called `navigate_to_pose_w_replanning_monitoring_and_recovery.xml` with the above node and the cancel BT ! Maybe it should be `navigate_to_pose_w_replanning_patience_and_recovery.xml`

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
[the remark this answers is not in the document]
SteveMacenski: OK let me know how that goes!

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_monitoring_and_recovery.xml
```diff
@@ -0,0 +1,46 @@
...
+          </RecoveryNode>
+        </RateController>
+        <ReactiveSequence name="MonitorAndFollowPath">
+          <IsPathLongerOnApproach new_path="{path}" prox_leng="0.0" length_factor="1.0">
+            <RetryUntilSuccessful num_attempts="1">
+              <SequenceStar name="CancelingControlAndWait">
+                <CancelControl name="ControlCancel"/>
+                <Wait wait_duration="5"/>
+              </SequenceStar>
+            </RetryUntilSuccessful>
```
SteveMacenski, line 24: Please base this BT off of the most recent ones, there are excessive `GoalUpdated` checks that don't do anything. Pull in `main`

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<nav_msgs::msg::Path>("new_path", "New Planned Path"),
+      BT::InputPort<nav_msgs::msg::Path>("old_path", "Old Planned Path"),
```
SteveMacenski, line 52: Remove this port, there is no reason for this, it should be stored internally. Then the single remaining input port should be `path` like every other port API

nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.cpp
```diff
@@ -0,0 +1,98 @@
...
+  // given the goal proximity and check if the new path is longer
+  if (isPathUpdated(new_path_, old_path_) && isRobotInGoalProximity(old_path_, prox_leng_) &&
+    isNewPathLonger(new_path_, old_path_, length_factor_) && !first_time_)
+  {
+    const BT::NodeStatus child_state = child_node_->executeTick();
+    switch (child_state) {
+      case BT::NodeStatus::RUNNING:
+        return BT::NodeStatus::RUNNING;
+      case BT::NodeStatus::SUCCESS:
+        first_time_ = true;
```
SteveMacenski, line 80: Wouldn't this be trivially overridden as `false` in L67 because the old path is still populated? I don't think this line has any effect and can be removed

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.hpp
padhupradheep: I added it mainly for the purpose of testing the failure case, without having an `old_path` it is highly impossible to test the failure case.

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_monitoring_and_recovery.xml
padhupradheep: Sure

padhupradheep: `main` also has the `GoalUpdated` checks. I added the BT based on the recent changes in the `main`.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.hpp
SteveMacenski: Why not just run the tree twice? I don't think you should be adding things like that purely for testing. There's got to be work arounds for testing like making a new BT node that derives from this that you can set the path before executing to have something in it beforehand.

SteveMacenski:
Have you been testing this? I've been seeing a few issues that makes me think that perhaps you haven't tested this yet since there were booleans of the wrong result, setting of state like `first_time_` that has no impact, etc. Please make sure to slow down and only submit a PR or updates **once you have it tested working**, else it just uses up a bunch of time for me to point out clear logical errors.

padhupradheep:
> Have you been testing this? I've been seeing a few issues that makes me think that perhaps you haven't tested this yet since there were booleans of the wrong result, setting of state like first_time_ that has no impact, etc.

Okay, now I tested both on the real robot and simulation, which seems to work fine. Now I also updated the test case scenario and I'm now pretty sure that we can let know the community(unless, if you spot any minors... ) and get feedback from them.

SteveMacenski:
Sorry, this got a little lost in my queue, is this tested in its current form and ready for another look?

nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.cpp
padhupradheep: Just to give you a clarity, this has been already taken care and it is not relevant anymore.

padhupradheep:
Yes, should be good for a check now.

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_monitoring_and_recovery.xml
padhupradheep: maybe this point I didn't get you. I did use the recent changes for the BT.

padhupradheep:
Just retriggering so that you can have a look.

padhupradheep:
A gentle reminder to have a look at this PR. I would love to take this off my list..  @SteveMacenski

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_goal_patience_and_recovery.xml
```diff
@@ -0,0 +1,49 @@
...
+    <RecoveryNode number_of_retries="6" name="NavigateRecovery">
+      <PipelineSequence name="NavigateWithReplanning">
+        <RateController hz="1.0">
+          <RecoveryNode number_of_retries="1" name="ComputePathToPose">
+            <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased"/>
+            <ClearEntireCostmap name="ClearGlobalCostmap-Context" service_name="global_costmap/clear_entirely_global_costmap"/>
+          </RecoveryNode>
+        </RateController>
+        <ReactiveSequence name="MonitorAndFollowPath">
+          <IsPathLongerOnApproach path="{path}" prox_leng="0.0" length_factor="1.0">
```
SteveMacenski, line 21: This is also semantically difficult to read. The `is_` prefix is used for condition nodes, but this is a decorator. I went a little on a rabbit chance to explain why we need to invert this logic but its actually the case that this is correct but the name is deceptive that this is not a condition node, but a decorator. 

I think `Is` needs to be replaced from the node's name, file name, and anywhere else it appears, to just being `PathLongerOnApproach`

nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.cpp
```diff
@@ -0,0 +1,94 @@
...
+        return BT::NodeStatus::SUCCESS;
+      case BT::NodeStatus::FAILURE:
+        return BT::NodeStatus::FAILURE;
+      default:
+        return BT::NodeStatus::FAILURE;
+    }
+  }
+  first_time_ = false;
+  old_path_ = new_path_;
+  return BT::NodeStatus::SUCCESS;
```
SteveMacenski, line 85: https://github.com/ros-planning/navigation2/blob/9b674ff43726b599b327155ec83fe5ba0dd8d50e/nav2_behavior_tree/plugins/decorator/speed_controller.cpp#L108 the other decorators just return `status()` in this situation. 

They also set it to running https://github.com/ros-planning/navigation2/blob/9b674ff43726b599b327155ec83fe5ba0dd8d50e/nav2_behavior_tree/plugins/decorator/speed_controller.cpp#L75 

In general, your implementation is fishy w.r.t. the other decorators, for example the most simple: https://github.com/ros-planning/navigation2/blob/9b674ff43726b599b327155ec83fe5ba0dd8d50e/nav2_behavior_tree/plugins/decorator/rate_controller.cpp

SteveMacenski (changes requested):

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_goal_patience_and_recovery.xml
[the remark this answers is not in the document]
padhupradheep: I just left it for the user to decide the value. The reason was because initially we were planning to update it in the default BT right ? Anyways I can update it.

nav2_behavior_tree/plugins/decorator/is_path_longer_on_approach.cpp
padhupradheep: I have seen that as well. But this decorator is a bit unique to the rest, this would be obviously success on the condition if the path is longer. That’s the main reason why, I also didn’t insist to change the name, as it is kinda conditional_decorator. I think this is the best possible solution with the current state of the BT API.

padhupradheep: I have seen that as well. But this decorator is a bit unique to the rest, this would be obviously success on the condition if the path is longer. I think this is the best possible solution with the current state of the BT API.

padhupradheep: I have seen that as well. But this decorator is a bit unique to the rest, this would be obviously success on the condition if the path is longer. The only place, where running occurs is in L73. That happens when you are waiting. So this particular node, does not have to do the return in that way, I think.

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_goal_patience_and_recovery.xml
```diff
@@ -0,0 +1,49 @@
+
+<!--
+  This Behavior Tree replans the global path periodically at 1 Hz and it also has
+  recovery actions specific to planning / control as well as general system issues. 
+  This will be continuous if a kinematically valid planner is selected.
```
SteveMacenski, line 5: This line is not part of this BT?

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_goal_patience_and_recovery.xml
```diff
@@ -0,0 +1,49 @@
...
+  Additionally it has a feature to wait near the goal to have a new valid plan 
+  if the old plan becomes invalid before trying to drive significantly
```
SteveMacenski, line 7: Its not that its invalid, its that it just asked the robot to drive around the "long way" vs waiting to see if the obstacle at the goal pose moves

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
```diff
@@ -0,0 +1,94 @@
...
+      case BT::NodeStatus::SUCCESS:
+        first_time_ = true;
+        return BT::NodeStatus::SUCCESS;
+      case BT::NodeStatus::FAILURE:
+        return BT::NodeStatus::FAILURE;
+      default:
+        return BT::NodeStatus::FAILURE;
+    }
+  }
+  first_time_ = false;
```
SteveMacenski, line 83: The last couple of things I'm wondering about are these global setters in L83/84.

We wouldn't change the `old_path_` if the `if()` condition was met to trigger the children. For the case of `RUNNING`, that has the nice result of retriggering to keep ticking without specifically storing the state of the child between runs. However, for the `SUCCESS` or `FAILURE` terminal condition, wouldn't it be necessary that the new path is ingested? 

Else, wouldn't this behavior just go in a cyclical loop that would never terminate if it actually did have to go all the way around because the obstacle in the way didn't move? Did you see that in your testing?

 `first_time_` I'm still not fully understanding what it is you're trying to accomplish with its use. Please let me know why this is around and what this is trying to solve. The check in the `if()` + change to `true` if success, and `false` at this line  created a behavior I'm not sure is sensible but I want to know what you are thinking about it. It feels to me like this should be default to `false` on initialization, then set to `true` once its been called once, and then only set back to false when the state returns to `IDLE`. Moreover, it seems like a proxy for `old_path_` essentially, where if it returns to IDLE we just clear those poses so that its empty again and then the `isPathUpdated()` check would handle it

padhupradheep: > We wouldn't change the old_path_ if the if() condition was met to trigger the children. For the case of RUNNING, that has the nice result of retriggering to keep ticking without specifically storing the state of the child between runs. However, for the SUCCESS or FAILURE terminal condition, wouldn't it be necessary that the new path is ingested?

We are using `ReactiveSequence` for this branch which would technically be reticking the `PathLongerOnApproach` nodes all the times. To quickly give you a short background on `ReactiveSequence`, it will tick the second child, if and only if the first child is set to SUCCESS. Also note that the first child will be reticked all the time. A nice example is given `[here](https://www.behaviortree.dev/sequencenode/)`. 

So during the state while robot is waiting, which technically here makes the BT node return a running state, remember the controller_server has been already cancelled, meaning the FollowPath is not in the RUNNING state any more. This BTNode gets reticked all the time, at this point obviously the new path gets ingested, which would eventually go into the if condition that you specify. The `FollowPath` node will be ticked if and only if the `PathLongerOnApproach` returns SUCCESS. So we have two scenarios for the SUCCESS to occur, first one is the obvious one without the obstacle. The second one is after you see the obstacle, wait for it either to see if it gets cleared, in this case as already said above, the `ReactiveSequence` makes sure that our decorator node gets reticked and compares new path (L61) with the lastly stored old path before the obstacle (L84) to see if the obstacle has cleared or not. If cleared, well and good the if condition fails and the SUCCESS is returned as seen in the L85. If not, it waits and returns the SUCCESS in L74. 

The only scenario where the FAILURE could happen is only when any of the decorator's children or grandchildren fail ( :-D )

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_goal_patience_and_recovery.xml
padhupradheep: This was taken from the default.

SteveMacenski: OK... update the description to describe your tree :smile:

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
SteveMacenski: I do not follow this explanation. If the node ticks its children due to passing the `if()`  condition, then the `old_path_` is never set with the new value for the duration of the children ticking process. At the end when it eventually does return a SUCCESS or FAILURE, I think the `old_path_` needs to be set to the current path so that there is now a reset in behavior that we process now based on the new path information. 

Else, then if the next path comes in after this has all gone down, it will forever be stuck in that loop because `old_path_` is never updated with the new going-around-the-obstacle longer path, if eventually the result is that we **have** to take the long path around, since you check proximity with the `old_path_` that ever was updated. You would have to have an intermediary `tick()` that did not pass the `if()` statement in order for the `old_path_` to be reset, which wouldn't happen if we were still within N meters of the goal and referencing the original `old_path_` that got us into that situation in the first place.  

To me at least, it seems clear that you need to set `old_path_ = new_path_;` before all FAILURE or SUCCESS conditions in the `if()` so that it resets the execution of this behavior to be rebased on the new information that we have to go around. 

Is there something I'm misunderstanding?

Or is this where that `first_time_` thing comes in and starts to muddle things up. Why is that even there? If we simply delete the reset `first_time_ = false;` at L83 and move `old_path_ = new_path_;`  before L76, L78, and L80 (terminal conditions) doesn't that make `first_time_` actually be about only the first time and clear up my concern as well?

padhupradheep: I think we are getting on to the same page now. 

> If the node ticks its children due to passing the if() condition, then the old_path_ is never set with the new value for the duration of the children ticking process. 

Exactly, that is what we want when the robot cancels and waits to see if there is a new shorter path. If we update the old_path_ at the point where we wait to see if the new path is shorter, then the old path (which is longer) get's updated by the global planner (remember we are not going to cancel the planner server). This would indeed have a chain of consequences leading to the failures for the case where we check for significantly longer path lengths. 

> To me at least, it seems clear that you need to set old_path_ = new_path_; before all FAILURE or SUCCESS conditions in the if() so that it resets the execution of this behavior to be rebased on the new information that we have to go around.

This is indeed an option. I had this idea initially and decided not to do it because of the code re-use. Meanwhile I had abused the usage of `first_time_` I think. I will update the PR.

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
```diff
@@ -0,0 +1,96 @@
...
+inline BT::NodeStatus PathLongerOnApproach::tick()
+{
+  getInput("path", new_path_);
+  getInput("prox_len", prox_len_);
+  getInput("length_factor", length_factor_);
+
+  // Check if the path is updated and valid, compare the old and the new path length,
+  // given the goal proximity and check if the new path is longer
+  if (isPathUpdated(new_path_, old_path_) && isRobotInGoalProximity(old_path_, prox_len_) &&
+    isNewPathLonger(new_path_, old_path_, length_factor_) && first_time_)
```
SteveMacenski, line 68: Shouldn't this be `!first_time_`?

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
```diff
@@ -0,0 +1,96 @@
...
+         length_factor * nav2_util::geometry_utils::calculate_path_length(
+    old_path, 0);
+}
+
+inline BT::NodeStatus PathLongerOnApproach::tick()
+{
+  getInput("path", new_path_);
+  getInput("prox_len", prox_len_);
+  getInput("length_factor", length_factor_);
+
```
SteveMacenski, line 64: I think you need to set the state of this node to running when running, and then you can check `if (IDLE) {first_time_ = true}` to reset between execution runs. e.g.

```
if (IDLE) {first_time_ = true}
setStatus(RUNNING)
```

Also `first_time_` defaults to `false` in your hpp file, shouldn't that be `true` and then it gets set to `false` in L86?

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
padhupradheep: > I think you need to set the state of this node to running when running, and then you can check if (IDLE) {first_time_ = true} to reset between execution runs.

I'm just not sure about the ideal scenario. But on a hindsight, we can save a feature request in the issue tracker! 

> Also first_time_ defaults to false in your hpp file, shouldn't that be true and then it gets set to false in L86?

Yepp! Solves the above conversation as well.

SteveMacenski (changes requested):

SteveMacenski (approved):
Please verify functionality after changes, but LGTM now (finally :sweat_smile: )

padhupradheep:
> Hey, I really liked your gifs in your documentation PR, that was really awesome! This actually feeds well potentially into a follow up project if you're interested. Right now, most of our gifs / images are from the turtlebot3 sandbox world. As you can imagine this is not very convincing of a real-world application to first-time readers. It would be nice if we could index the set of gifs / images we have and update them to use a more (or actually) real environment like what you showed in your gifs that is far more compelling. Further, there are places that it would be great to have videos / images / gifs where they are also missing that would be valuable for explanation.

Sure, I will be really interested in this project. I will be having additional resource in a month's time. So you can really count on me for this. We can also create a completely new environment if needed.

SteveMacenski:
Please specifically check that doing the BT multiple executions things reset properly

padhupradheep:
https://raw.githubusercontent.com/neobotix/neo_simulation2/main/maps/neo_workshop.pgm

padhupradheep:
https://github.com/neobotix/neo_simulation2/blob/main/maps/neo_workshop.pgm

padhupradheep:
> Please specifically check that doing the BT multiple consecutive navigation requests things reset properly and work

I tested all the functionalities several times, after the changes you had requested. I'm pretty confident on them now.

SteveMacenski:
OK, just waiting on docs. I'll do another look through once the docs are in order just to make sure (since we've found a number of issues, I want to second check myself :wink: ) but I think this should certainly be done before the end of the month

padhupradheep:
https://user-images.githubusercontent.com/20242192/158563579-02725da4-2b8e-4fcc-b124-da27aff24db0.mp4

Taking one step better than the GiF, a video showing both the possible cases. 

While you do the second check, this video might help in understanding what's happening.

SteveMacenski:
@padhupradheep want to post that video to discourse?

padhupradheep:
Sure, you are free to do that and post it on discourse or wherever you want. 😃

padhupradheep:
Damn, this is now in conflict. Let me try to resolve it.

padhupradheep:
Seems that CI is not triggered. May be someone needs to trigger it?

padhupradheep:
okay, dummy commit re-triggered. I hope this is a good way to go about.

SteveMacenski (approved):

SteveMacenski:
Congrats! This is a big and nice effort to get in. I'm happy to have some demo solutions to these problems and I would love moving forward we could integrate this work + Josh's mixed replanning work into a single BT with #2877 to make a BT for default that fixes the vast majority of behavioral problems for users!

padhupradheep:
Thank you very much for the help and advise Steve. I will be around, and glad to help further. 😃

PR #2811, opened 2022-02-08
Add optional node names to wait

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Part of [2415](https://github.com/ros-planning/navigation2/issues/2415) |
| Robotic platform tested on | Gazebo simulation of Waffle |

---

* Added possibility to wait for custom navigator and localizer in waitUntilNav2Active

https://github.com/ros-planning/navigation.ros.org/pull/275

---

#### For Maintainers: 

### DIFF
--- a/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
+++ b/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
@@ -267,10 +267,10 @@ class BasicNavigator(Node):
 
     def waitUntilNav2Active(self, navigator='bt_navigator', localizer='amcl'):
         """Block until the full navigation system is up and running."""
-        self._waitForNodeToActivate(navigator)
         self._waitForNodeToActivate(localizer)
         if localizer == 'amcl':
             self._waitForInitialPose()
+        self._waitForNodeToActivate(navigator)
         self.info('Nav2 is ready for use!')
         return
 
--- a/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
+++ b/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
@@ -265,11 +265,11 @@ class BasicNavigator(Node):
         else:
             return TaskResult.UNKNOWN
 
-    def waitUntilNav2Active(self):
+    def waitUntilNav2Active(self, navigator='bt_navigator', localizer='amcl'):
         """Block until the full navigation system is up and running."""
-        self._waitForNodeToActivate('amcl')
+        self._waitForNodeToActivate(navigator)
         self._waitForInitialPose()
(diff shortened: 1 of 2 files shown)


tonynajjar:
Following up on the conversation [here](https://github.com/ros-planning/navigation2/pull/2792#issuecomment-1032967041)


To me, waitForNodeToActivate sounds like a functionality ROS should provide and not a navigation library so I would keep it as private and let the user implement his own version if needed (there are probably other ways than through /get_state) or willingly use a private function. No strong opinion though, the last word is yours here.

tonynajjar:
I wouldn't think we need to update a demo to using these params because we can't demonstrate anything other than the default

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -265,11 +265,11 @@ def getResult(self):
         else:
             return TaskResult.UNKNOWN
 
-    def waitUntilNav2Active(self):
+    def waitUntilNav2Active(self, navigator='bt_navigator', localizer='amcl'):
         """Block until the full navigation system is up and running."""
-        self._waitForNodeToActivate('amcl')
+        self._waitForNodeToActivate(navigator)
         self._waitForInitialPose()
```
SteveMacenski, line 271: If the localizer is non-AMCL, then it might not make sense to wait for the initial pose (e.g. if in SLAM). I think that check should be done if its AMCL, else to not do it.

SteveMacenski (changes requested):

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -268,8 +268,9 @@ def getResult(self):
     def waitUntilNav2Active(self, navigator='bt_navigator', localizer='amcl'):
         """Block until the full navigation system is up and running."""
         self._waitForNodeToActivate(navigator)
-        self._waitForInitialPose()
         self._waitForNodeToActivate(localizer)
+        if localizer == 'amcl':
```
SteveMacenski, line 272: Needs to be in the original order 🙂but otherwise LGTM

tonynajjar: How am I messing up those small PRs :sweat_smile:. Reverted the order but why does it matter?

SteveMacenski (approved):

SteveMacenski:
I think the remaining python APIs I had in mind where remote server parameter setting (see code block in the issue comment) and allowing specification of all of the action request fields. 

Are there any other things you think a user might want to do that we’re not yet supporting?

PR #2812, opened 2022-02-11
Bound closest pose by costmap and turnaround point

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2622  |

---

- based on https://github.com/ros-planning/navigation2/pull/2790
- bound closest pose by local costmap and turnaround point
- don't prune by lookahead distance

None, I think

---

Add a bunch of automated tests to verify this behavior.

#### For Maintainers: 


SteveMacenski:
> Our planner doesn't necessarily do this, and this is why we have observed the "snapping" behavior since it just grabs the closest pose.

I'd be interested to hear a little about this. The impression I had was that it is pre-recorded and just being played back. That would mean the initial point could be off, but after it starts going things are good, no? Is there a situation you want to play back from not the start that you wouldn't be defining "where along" to start to be able to feed it just the pre-pruned path?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -604,22 +600,29 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
   const double max_costmap_dim = std::max(costmap->getSizeInCellsX(), costmap->getSizeInCellsY());
-  const double max_transform_dist = max_costmap_dim * costmap->getResolution() / 2.0;
+  const double max_costmap_extent = max_costmap_dim * costmap->getResolution() / 2.0;
+  const double max_transform_dist = std::min(max_costmap_extent, dist_to_direction_change);
+
+  // TODO(aposhian): for long paths, this might get computationally expensive to do all the time,
+  // and we should be able to cache the results of this distance accumulation
+  auto closest_pose_upper_bound =
+    nav2_util::geometry_utils::first_element_beyond(
+    global_plan_.poses.begin(), global_plan_.poses.end(), dist_to_direction_change);
```
SteveMacenski, line 610: You use `dist_to_direction_change` here, wouldn't this be `max_transform_dist`? What if the distance to direction change is infinite (e.g. there is no direction change).

SteveMacenski (commented):
@sathak93 please review and give us your thoughts

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: I think that makes sense to use in both locations. That way we only search from point 0 -> costmap edges for the starting point and then we only transform point 0 -> costmap edges for the purposes of control.

Aposhian:
>I'd be interested to hear a little about this. The impression I had was that it is pre-recorded and just being played back. That would mean the initial point could be off, but after it starts going things are good, no? Is there a situation you want to play back from not the start that you wouldn't be defining "where along" to start to be able to feed it just the pre-pruned path?

This behavior isn't very important to our use-case since we plan to handle those sorts of things at the planner level. It just something that we observed in the nav2 RPP behavior, and was kind of convenient, but also not ideal since it was non-deterministic.

SteveMacenski (commented):
This LGTM, just 2 polishing touches. I'd like @sathak93 to review though to give feedback to make sure this meets his need as well

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -604,22 +600,29 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
   // We'll discard points on the plan that are outside the local costmap
   nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
   const double max_costmap_dim = std::max(costmap->getSizeInCellsX(), costmap->getSizeInCellsY());
-  const double max_transform_dist = max_costmap_dim * costmap->getResolution() / 2.0;
+  const double max_costmap_extent = max_costmap_dim * costmap->getResolution() / 2.0;
+  const double max_transform_dist = std::min(max_costmap_extent, dist_to_direction_change);
+
+  // TODO(aposhian): for long paths, this might get computationally expensive to do all the time,
```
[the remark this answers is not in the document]
Aposhian: Sure, and I suppose that it would be more complex to implement the caching logic, especially where the global plan can be updated.

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -467,47 +458,62 @@ DWBLocalPlanner::transformGlobalPlan(
...
   } else {
-    sq_transform_end_threshold = sq_dist_threshold;
+    transform_end_threshold = dist_threshold;
   }
 
-  // Find the first pose in the plan that's less than sq_transform_start_threshold
+  // Find the first pose in the global plan that's further than prune distance
+  // from the robot using integrated distance
+  auto prune_point = nav2_util::geometry_utils::first_element_beyond(
+    global_plan_.poses.begin(), global_plan_.poses.end(), prune_dist);
```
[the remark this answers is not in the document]
Aposhian: I haven't made any modifications to @sathak93 's work with the dwb controller, and I haven't tested the dwb controller, but I could go ahead and make that change.

SteveMacenski:
Also don't worry about CI. Rolling is in a transitory state with the movement to 22.04 for Humble

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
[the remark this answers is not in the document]
sathak93: it's LGTM. @Aposhian  Does this changes solves lookahead point being not updated problem in your robot.

SteveMacenski:
Want to move this from a draft to a full PR?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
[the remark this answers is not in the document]
SteveMacenski: Didn't you add the dot product at the end of the transform function in DWB for that?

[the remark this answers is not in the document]
sathak93: No. dot product is to find the cusp. RPP has it already but DWB doesn't .  the above problem was caused by critics especially it was  triggered when the this https://github.com/ros-planning/navigation2/blob/70e7948afb45350e45c96b683518c1e6ba3f6010/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L222 updates and passes new path to controller.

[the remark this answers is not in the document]
SteveMacenski: Thank you, that's right

Aposhian:
I would still like to add some unit tests for `transformGlobalPlan` before taking it out of a draft PR.

SteveMacenski:
I can certainly appreciate that! 😁

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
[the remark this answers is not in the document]
Aposhian: Yes, it does fix that issue.

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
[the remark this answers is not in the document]
Aposhian: @SteveMacenski I had trouble finding where the default prune distance is set: I am not familiar with the dwb controller enough.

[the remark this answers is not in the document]
SteveMacenski: https://github.com/ros-planning/navigation2/blob/70e7948afb45350e45c96b683518c1e6ba3f6010/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L88-L90 :smile:

Aposhian:
I finally tested on my full system (in simulation) and it didn't solve the problem. I realized that this is because `findDirectionChange` doesn't find just any direction change: it _specifically_ finds "cusps", defined to be a pose where the the previous pose, and the next pose form an acute angle. This doesn't catch the case where you do a 180 with a smooth turn that doesn't involve any acute angles. I'm going to try to modify accordingly.

Aposhian:
My first idea for how to solve this problem is to look at orientation change relative to the starting position. The only downside of this is that it makes the RPP controller orientation aware: currently it completely ignores orientation in the path given to it.
Something like this:
```CPP
std::vector<geometry_msgs::msg::PoseStamped>::iterator findPathUpToTurn(
  const nav_msgs::msg::Path & path,
  double angle)
{
  tf2::Quaternion starting_orientation;
  tf2::fromMsg(path.poses.begin()->pose.orientation, starting_orientation);
  return std::find_if(
    path.poses.begin(), path.poses.end(),
    [](const auto& pose) {
      tf2::Quaternion quat;
      tf2::fromMsg(pose.pose.orientation, quat);
      return quat.angleShortestPath(starting_orientation) > angle;
    }
  );
}
```

SteveMacenski:
@Aposhian can we solve 1 problem at a time? The method of finding where paths change direction is certainly something we can talk about and refine, but that doesn't seem directly related to the topic at hand on fixing overlapping / crossed paths.

Aposhian:
I believe that the direction change code is integral to solving the problem of intersecting paths. I think there are ways of solving it fairly simply. I can definitely provide an illustration in a bit to explain why I think that is.

SteveMacenski:
I don't want both in the same PR though, I think these are decoupled problems. I think we can merge this solution first which would work with any planner given in Nav2 (which will produce the needed paths for this to be detectable). 

Then circle back around to how to better detect directional changes since you guys use another method.

Aposhian:
Here is an illustration of what I am talking about. When there is a loop that is fully contained in the costmap, you are not guaranteed to prune out poses that are closer to the starting pose if you only look at distance. In this particular case, you probably would be fine by using the costmap max extent integrated over the path, but not if the costmap was 2x larger, or the loop was 2x smaller.

So our goal is prune to path to:
- something that fits entirely in the local costmap
- something that doesn't have any poses from later on in the path close to the robot pose

Although it is occurring to me now: the only reason why the second point is a requirement is because the algorithm for selecting the lookahead point is based on pure euclidean distance. If selecting the lookahead point was done with integrated, rather than absolute, distance, then we wouldn't care if the transformed plan had overlapping poses. In the case of a simple loop like in this picture, if the lookahead distance was greater than the circumference of the loop, then it would shortcut the loop, but that would be the desired behavior. Whereas in the current implementation, if the transformed plan contains a loop like this, it will shortcut the loop even if the lookahead distance is very small, which is undesired.
direction-change

Aposhian:
To clarify, the current algorithm on `main` does:
1. `transformGlobalPlan` into the robot frame and prune to the local costmap (in order to avoid collisions)
2. Set lookahead distance based on speed and lookahead time, bounded by max and min lookahead distance
3. Set the lookahead point to the first point in the transformed plan that is further than the lookahead distance

And so far in this PR, we have been modifying `transformGlobalPlan` to:
- bound the start of the transformed plan to something within reason that doesn't chop off portions of the path
- bound the end of the transformed plan such that step (3) doesn't grab points later on in the path.

I propose that instead of coming up with bounds for the start and end of the transformed plan we should do the following:
- Always start the transformed plan at the start of the global plan. We already rely on this assumption to bound it. Why not run with it?
- Select the lookahead point based on integrated distance rather than absolute distance

Aposhian:
I just realized that we are relying on the "closest pose" search to prune the start of the path for subsequent iterations, so we will need some sort of that.

SteveMacenski:
> When there is a loop that is fully contained in the costmap, you are not guaranteed to prune out poses that are closer to the starting pose if you only look at distance. In this particular case, you probably would be fine by using the costmap max extent integrated over the path, but not if the costmap was 2x larger, or the loop was 2x smaller.

Didn't we cover this in the last ticket? I said we could just add a parameterization that we can recommend to set as the costmap size for the length forward to prune by. That way you can tune it knowing your costmap size vs turning characteristics to only look ahead the appropriate amount, but typically the costmap size is a good baseline (1.5-4m). 

> If selecting the lookahead point was done with integrated, rather than absolute, distance, then we wouldn't care if the transformed plan had overlapping poses

I believe we again covered this. If we prune the path by looking at N meters ahead **of integrated path length** to prune, then the lookahead point even if using the euclidean distance is max bounded by that integrated path distance used to prune it. That's how the implementation worked. 

So I'm really not sure what the issue is. Add a parameter to bound the forward pruning by either cusp distance or that (rather than costmap size). I think that should fully resolve your concerns without any other changes. 

 ---

None of this appears to have to do with the finding a cusp point problem (please revert that part of this PR).

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -602,29 +605,31 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
...
-  const double max_transform_dist = std::min(max_costmap_extent, dist_to_cusp);
-
+  // This bound is based on the assumption that the robot can't have traveled very far
+  // since the last iteration, and that when the global_plan was first passed, the robot
+  // was intended to start at the beginning.
+  // lookahead distance should be a pretty good proxy for this bound.
   auto closest_pose_upper_bound =
     nav2_util::geometry_utils::first_element_beyond(
-    global_plan_.poses.begin(), global_plan_.poses.end(), max_transform_dist);
+    global_plan_.poses.begin(), global_plan_.poses.end(), lookahead_distance);
```
SteveMacenski, line 614: Revert this back to `max_transform_dist` and have it be computed to be either distance to direction change **or** max parameterization rather than the costmap size.

SteveMacenski (changes requested):

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: We should not be pruning the total path available to the algorithm by the lookahead distance, that's way too short

Aposhian: Keep in mind this is not the pruning distance: this is the upper bound for finding the beginning of the transformed path. So this upper bound should be relatively short.

SteveMacenski: I don't think the lookahead distance though is big enough. This works fine if you're talking about a robot with a large look ahead distance, but if its short & the controller is running slow (10hz or something) due to limited compute, I could see cycles where that would find invalid past points because you hit the bound.

I think the role of this should be the largest bound reasonably possible. That's kind of why I would support this being a parameterization

SteveMacenski: For both locations where `max_transform_dist` is used (representing currently min of costmap size and directional change) could be replaced with the same parameter where instead of costmap size, now its a parameterized size. I think then using the updated `max_transform_dist` would work for both cases: (1) pruning the path forward and (2) finding the max to search for the closest point on the robot. 

What do you think?

Aposhian:
Here, I'll try to resummarize a timeline of what I have tried to clarify for both you and me (I think I might be getting confused as well), and where my thinking has gone:
1. My original problem was:
    - regardless of local costmap size, intersecting paths would result in the robot shortcutting to a later portion of the path. This was caused by:
      - `transformGlobalPlan` choosing the closest pose in `global_plan_` to start the transformed plan, regardless of where it was in the path.
2. @sathak93 came up with a solution to:
    - find the start of the transformed plan by choosing the closest point, but bounding it to be one of the points between the start of `global_plan_` up to `lookahead_dist_` _from the previous iteration_ integrated along the path.
    - find the end of the transformed plan by integrating the path length up to `max_transform_dist` (min of distance to cusp and the costmap max extent)
    - always try to find cusps, even if `allow_reversing_` is `false`, and bound the lookahead distance to be no greater than the distance to the first cusp.
3. I tested @sathak93 's solution on a straight path, and it did not work.
    - https://github.com/ros-planning/navigation2/pull/2790#issuecomment-1032158123
    - **I still don't exactly understand why this didn't work**
4. I modified @sathak93's solution to bound the selection of the beginning transformation by `max_transform_dist` (min of distance to cusp and the costmap max extent) instead of `lookahead_distance_` (from the previous iteration)
    - This had the desired effect of not breaking the robot moving on a straight path
    - This just borrowed using the distance to cusp directly, rather than getting it from `lookahead_distance_`
    - https://github.com/Aposhian/navigation2/commit/6e5c45bf827c870c3f95eb7d82379d96b4fb369c
    - https://github.com/Aposhian/navigation2/commit/0d5ff9825ba8bf50dec934ba40dd40f87ff09e17
5. On my robot, I would like to increase the size of the local costmap so I can get a further lookahead distance for less reactive control. But then I realized that if I do that, then it will be possible for turns to be self contained inside the costmap, so they would not necessarily get pruned simply by using the max costmap extent.
    - I thought that this should be solved by using the minimum distance to the cusp, because I thought that `findDirectionChange` included finding a 180 degree turn, which it does not.
    - https://github.com/ros-planning/navigation2/pull/2812#issuecomment-1047483822
    - The idea being is that if you prune up to where you turn around, you can never make it back to where you started.
6. I brought up a way to _actually_ calculate where the 180 degree turn is based on the orientations in the path, but didn't implement this idea:
    - https://github.com/ros-planning/navigation2/pull/2812#issuecomment-1047502219
7. Then I thought: why are we spending all this effort trying to make the transformed plan not have any intersecting poses? It is just because `getLookaheadPoint` only looks at euclidean distance. So then I thought: if I change `getLookaheadPoint` to use integrated distance, then it doesn't _matter_ if the transformed plan has intersections: we just have to worry about keeping it inside of the costmap.
    - We still need to have some bound for selecting the starting pose for the transform based on euclidean distance to the robot, since this what keeps the start of the global plan pruned.
      - If this bound is too small, then the robot can race past the start of the global plan between iterations and not be able to find the start for the next transformed plan
      - If this bound is too large, then we come back to the original problem of selecting a pose that is closer to the robot, but skips a significant portion of the path.
      - I rather arbitrarily chose this to be the lookahead distance. I have not tested this yet.

Aposhian:
So, for a while, I was thinking that there were _two_ root causes of the problem: the starting pose selection for the transformation _and_ the lookahead point selection on the path wrapping around. I now recognize that only the first is actually a problem, so I don't need to worry about the cases where the plan loops around, as long as I don't choose that for the start of the transform.

Aposhian:
Which means I don't even think we need to use integrated distance to find the end of the transformed path: we can just use the costmap. We just need integrated distance for finding the starting point.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -604,21 +607,27 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
   // We'll discard points on the plan that are outside the local costmap
   nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
   const double max_costmap_dim = std::max(costmap->getSizeInCellsX(), costmap->getSizeInCellsY());
-  const double max_transform_dist = max_costmap_dim * costmap->getResolution() / 2.0;
+  const double max_costmap_extent = max_costmap_dim * costmap->getResolution() / 2.0;
```
Aposhian, line 610: @SteveMacenski we have a choice on whether to use the max costmap extent, which could include points that are actually outside of the costmap, or we could use the min costmap extent, which could exclude points that are actually in the costmap. If we wanted to be exact, we would have to transform the whole path first, and then prune based on the costmap bounding box in the robot frame.

Aposhian:
I have now tested this most recent implementation on my simulation, and it works! :tada: Thanks for the input @SteveMacenski. I will try to make sure everything is cleaned up here, and I will still try to get those unit tests in.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -105,6 +105,9 @@ void RegulatedPurePursuitController::configure(
     node, plugin_name_ + ".max_angular_accel", rclcpp::ParameterValue(3.2));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".allow_reversing", rclcpp::ParameterValue(false));
+  declare_parameter_if_not_declared(
+    node, plugin_name_ + ".max_distance_between_iterations",
+    rclcpp::ParameterValue(10000.0));  // something really big
```
Aposhian, line 110: I tried using `std::numeric_limits<double>::max` here, but for some reason it made `rclcpp::ParameterValue` think it was a `bool`? We can either make the default really big: which means that most users will not see a change in behavior with this PR, or we can make it something "sensible" that will fix intersecting paths for most users.

Aposhian: @SteveMacenski what do you think of the parameter name `max_distance_between_iterations`?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -267,15 +274,11 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   // Find look ahead distance and point on path and publish
   double lookahead_dist = getLookAheadDistance(speed);
 
-  // Check for reverse driving
```
Aposhian, line 270: I think now that the cusp distance is not being used in `transformGlobalPlan`, we should be able to revert this so that cusps are only worried about when `allow_reversing_` is true? Or is there a reason to keep it like this @sathak93 ?

SteveMacenski:
Sorry, its taken me a hot minute to get through all of those commits / messages. It sounds like from your comment we're roughly on the same page again, but then I'm not sure why the files for `path_utils` are still in this PR? 

It would help me greatly in looking this over if I can do a 1:1 on what we had in Sathak's PR to the changes now to help wrap my head around this. There are now 39 commits and a number of decisions and reversals that I need some help getting back to the core foundation, concisely :sweat_smile:.  

I would prefer `findDirectionChange` over `findCusp` since the Cusp is a specific concept we use to find directional changes in the path. I'd like the names of the functions to describe what its doing with respect to the task at hand, which is finding the directional change in the path, which as an implementation detail uses some cusp mathematics. That way if we change the specific method of directional change detection later on, its still a valid name of the task being accomplished.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: I don't understand the comment, sorry, can you elaborate?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Part of the point here is not to transform any thing we don't have to, so I don't think a pre-transform would be a good idea. 

I think this is probably fine as is, I don't think there's a compelling reason to over-complicate this, no? Even if its slightly over, its unlikely to be used as a look ahead point. Even if it is used, the collision checker will check up until the edge of the costmap and if its off the costmap, will return a warning

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: For instance, we just changed this to a parameter from the costmap size due to issues you were having, but in general, I think setting this to the costmap size is a good **maximum bound**, I wouldn't suggest any more than that, but at times, it might be worthwhile to have it less than that if you have very irregular/curvy paths or an outstandingly large local costmap.

Aposhian:
@SteveMacenski I finished my unit tests for `transformGlobalPlan`, so I'm ready for this to be considered for merging. I haven't really touched anything on the DWB controller, since I don't have a robot (in sim or real) setup for that. But I am pretty confident in the RPP changes.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Aposhian: @sathak93 made this original change here to limit the lookahead distance to the closest cusp, regardless of whether reversing is enabled or not. I'm not sure if there is still a justification for this change: should we revert it to how it was?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Aposhian: Ok sounds good.

Aposhian: I will add a comment explaining that a little more.

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -107,6 +124,25 @@ inline Iter min_by(Iter begin, Iter end, Getter getCompareVal)
   return lowest_it;
 }
 
+/**
+ * Find first element in iterator that is greater integrated distance than comparevalue
+ */
+template<typename Iter, typename Getter>
+inline Iter first_element_beyond(Iter begin, Iter end, Getter getCompareVal)
```
Aposhian, line 131: I still wonder if there is a more descriptive name for this function, something that clarifies that we are using integrated distance. Maybe `find_after_integrated_distance`?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
sathak93: sorry for late reply. i think it is essential to make the robot to fully reach the pose point, when the path ahead  and the path behind the pose is overlapping.

sathak93: is this tested for overlapping case?

sathak93: > So, for a while, I was thinking that there were _two_ root causes of the problem: the starting pose selection for the transformation _and_ the lookahead point selection on the path wrapping around. I now recognize that only the first is actually a problem, so I don't need to worry about the cases where the plan loops around, as long as I don't choose that for the start of the transform.

for the overlapping path case lookahead point selection will surely cause the problem, it is still required to switch to integrated path distance

Aposhian: I think it won't be a problem since the lookahead point selection uses the first point outside the lookhead distance. Even if the path loops back around, and even overlaps, those poses right by the robot shouldn't be selected, right?

SteveMacenski:
Hi,

Back from PTO, should be responsive on this on a day-to-day basis now.

On `path_utils`, if its only there for testing, place in the `test` directory, we have many "helpers" for the tests that make them easier. I think this applies to that. I'd be happy to pull it back out if/when it is used in some of the functional run-time code. 

> which can also be interpreted as "direction"?

I'd be OK renaming `findDirectionChange` to something else, don't get me wrong. I just want to make sure it remains abstract from the implementation and describes what it is trying to accomplish. If you have another name you'd prefer, I'd be open to hearing it. I really can't think of one :laughing: 

Lastly, there's a few open comments throughout the thread above yet:
- Changing the DWB pruning default
- Sathak's comment on the DWB dot product loop https://github.com/ros-planning/navigation2/pull/2812#discussion_r805105875
- This thread on the reversing / lookahead point https://github.com/ros-planning/navigation2/pull/2812#discussion_r814492371

nav2_util/include/nav2_util/geometry_utils.hpp
SteveMacenski: How about `first_element_after_integrated_distance` or `first_after_integrated_distance`? That's alot of words, but the most clear combining the two. I think the "first" word (or one like it) is important to communicate its returning the first single item that meets the criteria, not any arbitrary one or a set of them. `find_` I think could be misinterpreted as potentially returning all.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -105,6 +105,9 @@ void RegulatedPurePursuitController::configure(
     node, plugin_name_ + ".max_angular_accel", rclcpp::ParameterValue(3.2));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".allow_reversing", rclcpp::ParameterValue(false));
+  declare_parameter_if_not_declared(
+    node, plugin_name_ + ".max_distance_between_iterations",
```
SteveMacenski, line 109: Note: I myself am terrible with naming things. I don't like this variable name, its not really very descriptive without reading alot of documentation or the code itself.

How about: `max_robot_pose_search_dist`? 

I think either way we'll need a healthy few sentences in documentation, but at least then describes "why" this is here and what it might be doing. I would recommend when we get to docs making sure to note that we do not recommend changing this from the default unless you have full loops or criss-crosses within your local costmap size, which you may then need to reduce proportionally to those sizes

Aposhian: Yeah I think `max_robot_pose_search_dist` is just fine. I can change it to that.

Aposhian: Where would be the best place to put those docs? In the README.md for `nav2_regulated_pure_pursuit_controller`?

Aposhian:
Ok, I have:
- Changed the DWB pruning default to 2.0 meters
- Added notes to `nav2_regulated_pure_pursuit_controller` README about `max_robot_pose_search_dist`

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -467,47 +458,62 @@ DWBLocalPlanner::transformGlobalPlan(
...
-    [&](const auto & global_plan_pose) {
-      return getSquareDistance(robot_pose.pose, global_plan_pose) > sq_transform_end_threshold;
-    });
+  // Find the first pose in the end of the plan that's further than transform_end_threshold
+  // from the robot using integrated distance
+  auto transformation_end =
+    nav2_util::geometry_utils::first_element_beyond(
+    transformation_begin, global_plan_.poses.end(), transform_end_threshold);
+
+  for (auto it = transformation_begin + 1; it != transformation_end - 1; ++it) {
```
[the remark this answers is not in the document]
Aposhian: @sathak93 can you explain on what is left to do here?

Aposhian:
@SteveMacenski on moving `path_utils`, are you suggesting to move it into `nav2_util/test` or `nav2_regulated_pure_pursuit_controller/test`? Also, I have some unit tests for `path_utils` itself (https://github.com/ros-planning/navigation2/blob/cd3ee1c1fd11dacb1840ffb38f0c515130a88971/nav2_util/test/test_path_utils.cpp): would those just go in the same directory?

nav2_util/include/nav2_util/geometry_utils.hpp
Aposhian: I like `*_after_integrated_distance`. I'm not as concerned about `find` in this context, since in most programming languages (including C++), `find` means finding the first match. But I think `first_after_integrated_distance` is just fine.

Aposhian:
Maybe a better name for `findDirectionChange` would be `findLinearVelocitySignChange`? or just `findVelocitySignChange`? That way it is clear that it is the velocity that is changing, not orientation.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Yeah, there should be a table with parameters in the readme you can add too. Also, once this is done, we'll need to update docs on navigation.ros.org that would also reflect it

nav2_util/include/nav2_util/geometry_utils.hpp
SteveMacenski: OK, lets do that

SteveMacenski:
Lets put them all in RPP until we have a reason to branch into `nav2_utils`

Aposhian:
Ok, I have moved the path_utils into `nav2_regulated_pure_pursuit_controller/test` and renamed `findCusp` to `findVelocitySignChange`. The only outstanding comments are about whether or not to revert the changes about using the velocity sign change based on whether reversing is enabled or not. @sathak93 do you still have outstanding concerns about that?

SteveMacenski:
I'd like to request @sathak93 do a full review and provide comments given the current state. Other than the 2 items he was addressing, I'm good with this as is so it would be a good opportunity.

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -467,47 +458,62 @@ DWBLocalPlanner::transformGlobalPlan(
...
-  // from the robot
-  auto transformation_end = std::find_if(
-    transformation_begin, end(global_plan_.poses),
-    [&](const auto & global_plan_pose) {
-      return getSquareDistance(robot_pose.pose, global_plan_pose) > sq_transform_end_threshold;
-    });
+  // Find the first pose in the end of the plan that's further than transform_end_threshold
+  // from the robot using integrated distance
+  auto transformation_end =
+    nav2_util::geometry_utils::first_after_integrated_distance(
```
SteveMacenski, line 503: The RPP version of this logic uses `find_if` here: 

```
  auto transformation_end = std::find_if(
    transformation_begin, global_plan_.poses.end(),
    [&](const auto & pose) {
      return euclidean_distance(pose, robot_pose) > max_costmap_extent;
    });
```

I'm not really sure if it matters one way or another, but shouldn't we keep them consistent? I know this is from @sathak93 's PR so maybe he can let us know if there's a reason for the change here or if we can homologate them

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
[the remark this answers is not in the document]
SteveMacenski: I also don't necessarily agree this should be added here. That was important for RPP. I know this is for creating the behavior of the robot going all the way to the end at a cusp, but at best this should be a parameterized option not the default. The nature of DWB is a scoring technique and if the scoring says to cut short, that's the nature of the controller that it deviates from the path. Though I agree that having the option is valuable so I'd be fine with having this be an option that is usable, but defaulted to off. 

Also, what is left to do here? This looks pretty complete to me?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: So I think this is in the same vein as the other comment that is still pending. If I understand this correctly:
- @Aposhian thinks we can remove this since we're no longer limited 
- @sathak93 wants to keep this so that the robot always gets to exactly the cusp point and then turns around to fully track the path

I think I agree with @Aposhian for now to have it cut the path short in that situation if not reversing. I think that is reasonably expected behavior for a non-feasible transition point to have a little shortness applied. But, I would also be open to adding a parameterization to this to enable it either way but I think that should be as part of another PR to not block this work from resolving 80% of the problem today. If @sathak93 you're open to submitting a PR to make a parameter to enable exact-full-path behavior, I'd be happy to review / merge it at a later date.

Aposhian:
I have changed RPP to only check cusp distance when reversing is enabled again.

SteveMacenski:
OK, if you removed https://github.com/ros-planning/navigation2/pull/2812/files#diff-3a9280dd976248532d945d2664b4732c22a2e9cdb598d2240710caf4b22495eaR506-R516 that would make these 2 analogous and I'd approve.

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
sathak93: this was same as that in RPP in my PR. @Aposhian  changed it in  RPP now. i think it is better to make them consistent.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
sathak93: @SteveMacenski I am open to follow up on this and contribute.

sathak93:
LGTM

SteveMacenski (approved):

Issue #2815, opened 2022-02-13
[RPP] Jerk limited trajectory generation


angstrem98:
## Feature request
Regulated Pure Pursuit controller is an attractive controller due to it's simplicity to configure and robustness.
It used to have acceleration limiting which did not work properly and has been removed. There is some deceleration near goal implemented but it requires large lookahead values in order to prevent harsh deceleration.
RPP controller should have motion profile generation with limited acceleration and jerk that allows it to smoothly accelerate from standstill and decelerate when approaching the goal.

I'm already working on this feature and opening this issue as part of the procedure to open a PR as noted in navigation2 documentation.

Issue #2633 is closely related, but it is an external velocity smoother. We need to know when to start braking inside the controller in order to reach the goal accurately.

angstrem98:
Comparing before and after results.  
Sending custom differential drive robot in Webots from point A to point B with original implementation results in commanded linear velocity as shown on image
## Before:  
![original_rl](https://user-images.githubusercontent.com/44432285/153780515-3cc0807e-0612-42d6-b750-5b43479196ca.png)

New implementation in the linked Pull Request produced following commanded linear velocity:
## After:
![jerk_limited_rl](https://user-images.githubusercontent.com/44432285/153780550-ef52922f-024d-437a-811e-7f13542a545a.png)

---
We can notice that the new commanded velocity has "S" shaped curve - jerk limited.

angstrem98:
Motion profile generation is done using ```ruckig``` package (it is used inside ```MoveIt2```).
Imagine a robot with significant mass driving at moderate/higher speeds towards the goal. We want it to smoothly come to a stop at the goal point. Velocity smoother #2633 could help us to smooth acceleration, but deceleration near goal with current implementation would cause overshoots. Maybe it is not noticeable with TurtleBot which moves at relatively low speeds, but tests with custom robot moving at 0.4 m/s and above showed overshoots with original implementation.

In order to know when to start braking, one ```ruckig``` instance works Position mode and tracks remaining path length.
While following the path, another  instance works in velocity mode and sets the angular velocity command. 
When we need to rotate to path heading or rotate to goal angle, proportional controller is used to achieve accurate angle control. Tests showed easy achievement of 1 degree tolerance in goal checker.

SteveMacenski:
I wanted to give you a note that this is still on my review queue. I've been slammed the last few weeks and I'm now getting my footing back to review these bigger PRs that have been sitting around.

[...]

Issue #2633, opened 2021-10-20
Nav2 Velocity Smoother


vinnnyr:
https://github.com/ros-planning/navigation2/pull/2631 will remove acceleration / deceleration limits from RPP due to some issues. With the idea that these kind of things should be handled by a downstream velocity smoother anyways. 

Is there a home for a velocity smoother in Nav2, either as a tutorial, or a full implementation?

This proposed smoother could set (this list is just an initial set of ideas):
- acceleration limit 
- deceleration limit
- jerk limit 
- velocity deadband (do not command a velocity between x1 m/s and x2 m/s)

vinnnyr:
On another note, I am not sure why this issue did not pre-populate with the issue template -- perhaps bc I am using the iOS app... Apologies, I will edit to fit the template soon as possible.

wilcobonestroo:
> Aren't most of those taking odometry messages https://github.com/kobuki-base/velocity_smoother/blob/devel/src/velocity_smoother.cpp#L161 to check and limit the next velocity command by? That would make most sense to me.

I see that the Kobuko package has 3 options for input: **none**, **odometry** and **commanded velocities**. Odometry makes sure that the new `cmd_vel` value is within bounds compared to the currently estimated odometry. The commanded velocities looks at the previous `cmd_vel` that was send and assumes that the robot follows these commands. I don't understand the none option yet, but I will have a look at their code to see what it actually does. I think it makes sense to at least have the options for **odom** and for **commanded velocities** as input.

In the three implementations mentioned above I only see speed limits and acceleration limits. I think we can add jerk limits as an option.

@vinnnyr Can you describe some use cases or scenarios where you use the deadband issue? E.g. what are typical values? Is there one band or multiple? What should happen if the orginal cmd_vel is in the deadband?

I can pick up this issue and work on a nav2 velocity smoother. Should I put it in its own package or add it to an existing one? What would be a good name for the package? Simply `nav2_velocity_smoother`? There is another folder called `nav2_smoother`

vinnnyr:
Jerk Limits
--
- Yeah so I can concede that jerk limits when in **odometry** mode might not make a lot of sense due to localization noise.

Deadband
--
The velocity smoother in the version of Regulated Pure Pursuit (RPP) was in **odometry** mode. 
[...]

PR #2816, opened 2022-02-13
[RPP] Jerk limited trajectory generation

angstrem98:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (#2815) |
| Robotic platform tested on | (Custom differential drive robot in Webots, TurtleBot 3 Gazebo simulation) |

---

* Generate jerk limited trajectories using ```ruckig``` package
* Provide accurate goal angle control using proportional controller
* Smooth stopping near goal while respecting kinematic limits due to motion profiling

* Add new parameters to documentation, remove unused

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -30,6 +30,7 @@
 #include "nav2_util/odometry_utils.hpp"
 #include "nav2_util/geometry_utils.hpp"
 #include "geometry_msgs/msg/pose2_d.hpp"
+#include "ruckig/ruckig.hpp"
 
 namespace nav2_regulated_pure_pursuit_controller
 {
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -238,6 +238,13 @@ protected:
   double findDirectionChange(const geometry_msgs::msg::PoseStamped & pose);
 
   /**
+   * @brief Normalizes angle in range [-pi, pi]
+   * @param angle Angle to normalize
+   * @return Normalized angle
+   */
+  double angleNormalize(double angle);
+
+  /**
    * @brief Callback executed when a parameter change is detected
    * @param event ParameterEvent message
    */
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -158,9 +158,11 @@ protected:
   /**
    * @brief Whether robot should rotate to final goal orientation
    * @param carrot_pose current lookahead point
+   * @param lookahead_dist current lookahead distance
    * @return Whether should rotate to goal heading
    */
-  bool shouldRotateToGoalHeading(const geometry_msgs::msg::PoseStamped & carrot_pose);
+  bool shouldRotateToGoalHeading(
+    const geometry_msgs::msg::PoseStamped & carrot_pose, const double & lookahead_dist);
 
   /**
    * @brief Create a smooth and kinematically smoothed rotation command
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -251,6 +251,8 @@ protected:
   rclcpp::Clock::SharedPtr clock_;
 
   double desired_linear_vel_, base_desired_linear_vel_;
+  double max_linear_accel_;
+  double max_linear_decel_;
   double lookahead_dist_;
   double rotate_to_heading_angular_vel_;
   double max_lookahead_dist_;
@@ -273,6 +275,12 @@ protected:
   double rotate_to_heading_min_angle_;
   double goal_dist_tol_;
   bool allow_reversing_;
+  double robot_angle_;
+  bool rotating_;
+  double kp_angle_;
+  double max_linear_jerk_;
+  double max_angular_jerk_;
+  rclcpp::Time system_time_;
 
   nav_msgs::msg::Path global_plan_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> global_path_pub_;
@@ -281,6 +289,16 @@ protected:
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> carrot_arc_pub_;
   std::unique_ptr<nav2_costmap_2d::FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>
   collision_checker_;
+
+  ruckig::Ruckig<1> * distance_profile_;
+  ruckig::InputParameter<1> distance_profile_input_;
+  ruckig::OutputParameter<1> distance_profile_output_;
+  ruckig::Result distance_profile_result_;
+
+  ruckig::Ruckig<1> * angle_profile_;
+  ruckig::InputParameter<1> angle_profile_input_;
+  ruckig::OutputParameter<1> angle_profile_output_;
+  ruckig::Result angle_profile_result_;
 
   // Dynamic parameters handler
   std::mutex mutex_;
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -224,9 +224,11 @@ protected:
    * @brief Get lookahead point
    * @param lookahead_dist Optimal lookahead distance
    * @param path Current global path
+   * @param remaining_path_length path length from lookahead point to the end of pruned path
    * @return Lookahead point
    */
-  geometry_msgs::msg::PoseStamped getLookAheadPoint(const double &, const nav_msgs::msg::Path &);
+  geometry_msgs::msg::PoseStamped getLookAheadPoint(
+    const double &, const nav_msgs::msg::Path &, double & remaining_path_length);
 
   /**
    * @brief checks for the cusp position
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -638,13 +638,12 @@ double RegulatedPurePursuitController::costAtPose(const double & x, const double
 }
 
 void RegulatedPurePursuitController::applyConstraints(
-  const double & dist_error, const double & lookahead_dist,
+  const double & /*dist_error*/, const double & /*lookahead_dist*/,
   const double & curvature, const geometry_msgs::msg::Twist & /*curr_speed*/,
   const double & pose_cost, double & linear_vel, double & sign)
 {
   double curvature_vel = linear_vel;
   double cost_vel = linear_vel;
-  double approach_vel = linear_vel;
 
   // limit the linear velocity by curvature
   const double radius = fabs(1.0 / curvature);
@@ -670,24 +669,6 @@ void RegulatedPurePursuitController::applyConstraints(
   // Use the lowest of the 2 constraint heuristics, but above the minimum translational speed
   linear_vel = std::min(cost_vel, curvature_vel);
   linear_vel = std::max(linear_vel, regulated_linear_scaling_min_speed_);
-
-  // if the actual lookahead distance is shorter than requested, that means we're at the
-  // end of the path. We'll scale linear velocity by error to slow to a smooth stop.
-  // This expression is eq. to (1) holding time to goal, t, constant using the theoretical
-  // lookahead distance and proposed velocity and (2) using t with the actual lookahead
-  // distance to scale the velocity (e.g. t = lookahead / velocity, v = carrot / t).
-  if (dist_error > 2.0 * costmap_->getResolution()) {
-    double velocity_scaling = 1.0 - (dist_error / lookahead_dist);
-    double unbounded_vel = approach_vel * velocity_scaling;
-    if (unbounded_vel < min_approach_linear_velocity_) {
-      approach_vel = min_approach_linear_velocity_;
-    } else {
-      approach_vel *= velocity_scaling;
-    }
-
-    // Use the lowest velocity between approach and other constraints, if all overlapping
-    linear_vel = std::min(linear_vel, approach_vel);
-  }
 
   // Limit linear velocities to be valid
   linear_vel = std::clamp(fabs(linear_vel), 0.0, desired_linear_vel_);
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -387,7 +387,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
   // Make sure we're in compliance with basic constraints
   double angle_to_heading;
-  if (shouldRotateToGoalHeading(carrot_pose) && !rotating_) {
+  if (shouldRotateToGoalHeading(carrot_pose, lookahead_dist) && !rotating_) {
     double angle_to_goal = tf2::getYaw(transformed_plan.poses.back().pose.orientation);
     rotateToHeading(linear_vel, angular_vel, angle_to_goal, speed);
   } else if (shouldRotateToPath(carrot_pose, angle_to_heading) && !rotating_) {
@@ -463,11 +463,12 @@ bool RegulatedPurePursuitController::shouldRotateToPath(
 }
 
 bool RegulatedPurePursuitController::shouldRotateToGoalHeading(
-  const geometry_msgs::msg::PoseStamped & carrot_pose)
+  const geometry_msgs::msg::PoseStamped & carrot_pose, const double & lookahead_dist)
 {
   // Whether we should rotate robot to goal heading
-  double dist_to_goal = std::hypot(carrot_pose.pose.position.x, carrot_pose.pose.position.y);
-  return use_rotate_to_heading_ && dist_to_goal < goal_dist_tol_;
+  double carrot_dist = std::hypot(carrot_pose.pose.position.x, carrot_pose.pose.position.y);
+  return use_rotate_to_heading_ && carrot_dist < goal_dist_tol_ &&
+   fabs(lookahead_dist - carrot_dist) > goal_dist_tol_;
 }
 
 void RegulatedPurePursuitController::rotateToHeading(
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -420,7 +420,7 @@ void RegulatedPurePursuitController::rotateToHeading(
 
 geometry_msgs::msg::PoseStamped RegulatedPurePursuitController::getLookAheadPoint(
   const double & lookahead_dist,
-  const nav_msgs::msg::Path & transformed_plan)
+  const nav_msgs::msg::Path & transformed_plan, double & remaining_path_length)
 {
   // Find the first pose which is at a distance greater than the lookahead distance
   auto goal_pose_it = std::find_if(
@@ -432,6 +432,11 @@ geometry_msgs::msg::PoseStamped RegulatedPurePursuitController::getLookAheadPoin
   if (goal_pose_it == transformed_plan.poses.end()) {
     goal_pose_it = std::prev(transformed_plan.poses.end());
   }
+
+  size_t lookahead_index = std::distance(transformed_plan.poses.begin(), goal_pose_it);
+  remaining_path_length = nav2_util::geometry_utils::calculate_path_length(
+    transformed_plan,
+    lookahead_index);
 
   return *goal_pose_it;
 }
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -192,6 +192,11 @@ void RegulatedPurePursuitController::configure(
   collision_checker_ = std::make_unique<nav2_costmap_2d::
       FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>(costmap_);
   collision_checker_->setCostmap(costmap_);
+
+  distance_profile_ = new ruckig::Ruckig<1>{control_duration_};
+  angle_profile_ = new ruckig::Ruckig<1>{control_duration_};
+
+  system_time_ = clock_->now();
 }
 
 void RegulatedPurePursuitController::cleanup()
@@ -204,6 +209,9 @@ void RegulatedPurePursuitController::cleanup()
   global_path_pub_.reset();
   carrot_pub_.reset();
   carrot_arc_pub_.reset();
+
+  delete distance_profile_;
+  delete angle_profile_;
 }
 
 void RegulatedPurePursuitController::activate()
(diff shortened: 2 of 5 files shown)


doisyg:
Thanks for the PR, it seems to generate nice smoothed cmd vel. 

However I believe there is a bug with the `rotating_`  globlal variable. When rotating at the end of a path, it can be left in the `true` state because the goal checker stops executing `computeVelocityCommands` and then this code path is never executed (i.e. it is never set back to false): https://github.com/memristor/navigation2/blob/7f1e4e5584bed038cb660ac10166cc9a27b65bf7/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L395-L398

This creates an inconsistent/unrequired rotation at the beginning of the next path.

The `rotating_`  globlal variable needs to be reseted before each new execution of a path, for instance by adding `rotating_ = false;` in `setPlan`

angstrem98:
> Thanks for the PR, it seems to generate nice smoothed cmd vel.
> 
> However I believe there is a bug with the `rotating_` globlal variable. When rotating at the end of a path, it can be left in the `true` state because the goal checker stops executing `computeVelocityCommands` and then this code path is never executed (i.e. it is never set back to false): https://github.com/memristor/navigation2/blob/7f1e4e5584bed038cb660ac10166cc9a27b65bf7/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L395-L398
> 
> This creates an inconsistent/unrequired rotation at the beginning of the next path.
> 
> The `rotating_` globlal variable needs to be reseted before each new execution of a path, for instance by adding `rotating_ = false;` in `setPlan`

Thanks for trying it out and pointing an issue. I've pushed a commit that should fix the problem. Besides resetting it in `setPlan` as you proposed, `rotating_` is being reset here too:
https://github.com/memristor/navigation2/blob/4eb8c3920d6e3f772e59f20b67556d4c22a0690d/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L333

I'm looking forward to hearing how it works out!

doisyg:
Your last PR fixes the rotation bug, thx.

doisyg:
You should probably make your new parameters dynamic similarly to the others

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -234,6 +239,13 @@ class RegulatedPurePursuitController : public nav2_core::Controller
    */
   double findDirectionChange(const geometry_msgs::msg::PoseStamped & pose);
 
+  /**
+   * @brief Normalizes angle in range [-pi, pi]
+   * @param angle Angle to normalize
+   * @return Normalized angle
+   */
+  double angleNormalize(double angle);
```
SteveMacenski, line 247: Pass by const reference

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -176,6 +192,11 @@ void RegulatedPurePursuitController::configure(
   collision_checker_ = std::make_unique<nav2_costmap_2d::
       FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>(costmap_);
   collision_checker_->setCostmap(costmap_);
+
+  distance_profile_ = new ruckig::Ruckig<1>{control_duration_};
```
SteveMacenski, line 196: Use unique_ptrs, we've removed any raw memory management  from Nav2

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -188,6 +209,9 @@ void RegulatedPurePursuitController::cleanup()
   global_path_pub_.reset();
   carrot_pub_.reset();
   carrot_arc_pub_.reset();
+
+  delete distance_profile_;
```
SteveMacenski, line 213: Then these may be `.reset()`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -295,6 +370,11 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     curvature = 2.0 * carrot_pose.pose.position.y / carrot_dist2;
   }
 
+  double carrot_dist = sqrt(carrot_dist2);
```
SteveMacenski, line 373: We use the square intentionally to above the sqrt, can't the `remaining_path_length` be `remaining_path_length*remaining_path_length`, that's much cheaper than a sqrt operation

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -700,6 +814,17 @@ bool RegulatedPurePursuitController::transformPose(
   return false;
 }
 
+double RegulatedPurePursuitController::angleNormalize(double angle)
```
SteveMacenski, line 817: `ros/angles` has these types of functions implemented already that can be used, remove from here and use the standard implementations

nav2_regulated_pure_pursuit_controller/test/test_regulated_pp.cpp
```diff
@@ -70,10 +71,10 @@ class BasicAPIRPP : public nav2_regulated_pure_pursuit_controller::RegulatedPure
     return shouldRotateToPath(carrot_pose, angle_to_path);
   }
 
-  bool shouldRotateToGoalHeadingWrapper(const geometry_msgs::msg::PoseStamped & carrot_pose)
-  {
-    return shouldRotateToGoalHeading(carrot_pose);
-  }
+  // bool shouldRotateToGoalHeadingWrapper(const geometry_msgs::msg::PoseStamped & carrot_pose)
```
SteveMacenski, line 74: These tests need to be readded

SteveMacenski (changes requested):
I need to in more detail look into the ruckig stuff, but here are the project / software level changes required.

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -272,6 +286,12 @@ class RegulatedPurePursuitController : public nav2_core::Controller
   double rotate_to_heading_min_angle_;
   double goal_dist_tol_;
   bool allow_reversing_;
+  double robot_angle_;
+  bool rotating_;
+  double kp_angle_;
+  double max_linear_jerk_;
+  double max_angular_jerk_;
+  rclcpp::Time system_time_;
```
SteveMacenski, line 294: This is reliably run at the `controller_frequency` rate, may that be substituted instead of keeping timers?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -303,13 +383,21 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
   linear_vel = desired_linear_vel_;
 
+  distance_profile_input_.max_velocity = {desired_linear_vel_};
```
SteveMacenski, line 386: Isn't this already set?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -568,6 +679,9 @@ void RegulatedPurePursuitController::applyConstraints(
 void RegulatedPurePursuitController::setPlan(const nav_msgs::msg::Path & path)
 {
   global_plan_ = path;
+
+  angle_profile_input_.control_interface = ruckig::ControlInterface::Velocity;
```
SteveMacenski, line 683: Isn't this already set?

SteveMacenski:
Is there another proxy you can use rather than keeping some `rotating_` state? For instance, in rotation, check some value which would indicate as such or be able to reset another variable to indicate that?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -318,19 +406,51 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
     // Apply curvature to angular velocity after constraining linear velocity
     angular_vel = linear_vel * curvature;
+
+    distance_profile_input_.control_interface = ruckig::ControlInterface::Position;
```
SteveMacenski, line 410: Isn't this already set in activate?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -318,19 +406,51 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+    distance_profile_input_.target_position = {
+      distance_profile_input_.current_position[0] + sign * remaining_path_length};
+    distance_profile_input_.max_velocity = {fabs(linear_vel)};
+
+    angle_profile_input_.control_interface = ruckig::ControlInterface::Velocity;
```
SteveMacenski, line 415: Isn't this already set in activate?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -318,19 +406,51 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+  if (angle_profile_result_ == ruckig::Result::Working) {
+    angle_profile_output_.pass_to_input(angle_profile_input_);
+  }
+
+  linear_vel_command = distance_profile_output_.new_velocity[0];
+  if (angle_profile_input_.control_interface == ruckig::ControlInterface::Velocity) {
+    angular_vel_command = angle_profile_output_.new_velocity[0];
+  } else {
+    // proportional controller for robot angle when we are in position mode
+    angular_vel_command = kp_angle_ * angleNormalize(
```
SteveMacenski, line 440: I'm not entirely following this `kp_angle_` thing - can you elaborate on it?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -318,19 +406,51 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
     // Apply curvature to angular velocity after constraining linear velocity
     angular_vel = linear_vel * curvature;
+
+    distance_profile_input_.control_interface = ruckig::ControlInterface::Position;
+    distance_profile_input_.target_position = {
+      distance_profile_input_.current_position[0] + sign * remaining_path_length};
```
SteveMacenski, line 412: Does this strictly need to be the path integrated distance, or can this be the L2 norm? This should only be happening on starting or ending a goal, so either the goal is far away or very close

SteveMacenski:
Another question is whether its strictly required to have the angular velocity be set with profiles. I think it depends on the aims and goals of this. If the angular velocity is smoothed out too much or parameters not set properly, it could deviate from the path and that will cause substantial oscillatory behaviors.

doisyg:
@angstrem98, this PR is promising, any update ?

Aposhian:
I haven't done rigorous testing, but I have tested this in simulation with our robot and found good results!

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
angstrem98: Replaced with `ros/angles`, as pointed out in the other comment.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Switched to smart pointers.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Done.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Sqrt was already used for `carrot_dist2` so I felt free to use it (and I really need sqrt):
https://github.com/ros-planning/navigation2/blob/609342fa27016a1677768f93fb3c44dec98b1f55/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L328

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Thanks, very good suggestion.

nav2_regulated_pure_pursuit_controller/test/test_regulated_pp.cpp
angstrem98: Tests readded.

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
angstrem98: True, it is run at mentioned frequency. I am not using timers, just tracking system time. The problem is, I need to reset some internal states in case goal has been cancelled or goal checker stopped calling `computeVelocityCommands`. There is no such callback that allows me to do that, so my solution is to check if `computeVelocityCommands` has not been called for 4 control loop durations. If that's true, we have to reset internal states since controller has been idle for a while. You can look at the logic here:
https://github.com/memristor/navigation2/blob/03cf93ce88499222ed5e599f1a5dad8e8572c50d/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L324-L344

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Yes, but logic is as follows:

1. Set maximum velocity to `desired_linear_vel_`
2. RPP algorithm modifies maximum velocity in order to folow path

Because maximum velocity limit in motion profile is changed, we must reload it in step 1.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: When new path is passed to controller it could be rotating to some heading at the moment. This allows the main control loop to break that rotation if needed.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Sometimes distance profile goes in velocity mode. This is when we rotate to heading - we set distance profile to velocity mode and command zero velocity which brings us to a smooth stop. So, switching to position mode is needed later.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: It has to be set again since we switch position and velocity modes around and modify velocity limits. Similar to previous answers.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: In case a robot is able to rotate in place I wanted to make it able to rotate to goal heading (or path heading) accurately. Current implementation on main branch doesn't allow fine positioning at high speeds.
We have a goal heading (angle) setpoint which we want to achieve (make out robot angle become equal to setpoint). Difference between setpoint and current angle is error. We can pass this error to a controller such as PID to make the error zero and thus achieve our goal heading (with gains `kp`, `ki`, `kd`). Position tracking controllers are generally designed as P or PD. I made a simple control law that basically does this: `command = proportional_gain * error`.
So, `kp_angle_` is proportional gain. 

We could simply pass it a setpoint and it will eventually reach it, but high gain could cause oscillations and low gain would prevent oscilaltions but it would take pretty long to reach the setpoint. Setting the angle motion profile to position mode makes it generate series of position setpoints which will make us reach the setpoint while respecting kinematic limits. This profile allows us to have a higher `kp` gain.

This approach allowed me to have yaw goal checker at 1 degree and less precision (on a custom robot).

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: It needs to be the path integrated distance, this is the core idea when it comes to stopping on goal point. By passing remaining path length (in meters) to distance profile it can calculate when to start braking.

angstrem98:
> Another question is whether its strictly required to have the angular velocity be set with profiles. I think it depends on the aims and goals of this. If the angular velocity is smoothed out too much or parameters not set properly, it could deviate from the path and that will cause substantial oscillatory behaviors. Linear velocity its more essentially clear the value and need for

I would say that angular velocity still needs profiles (especially when rotating towards heading in position mode.)
I don't think about it only as a smoother, but also as a limiter. Imagine a robot without payload heading to pick up some lab samples. While empty it can go fast towards the goal. After picking up the samples, user sets limits that need to be respected in order to bring them safely. 

@Aposhian Yes, making jerk limits too low will make robot oscillate for sure, that way you are making it really "lazy". Also concerning @SteveMacenski question: My stance is that if kinematic limits make robot oscillate or deviate from path, RPP regulated parameters are simply not properly set (or limits are unreasonable). User has entered the limits that need to be respected, and they are being respected, but regulated parameters need some tuning for proper performance.

angstrem98:
> Is there another proxy you can use rather than keeping some `rotating_` state? For instance, in rotation, check some value which would indicate as such or be able to reset another variable to indicate that?

I got rid of it in the latest push. I am able to check if angle profile is in position or velocity mode.

@doisyg Dynamic parameters should be working now.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Fair enough - replace the call on that line  then with this variable so we don't do it twice

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
SteveMacenski: You should break that logic out into a separate function

SteveMacenski: Would `setPlan` do that? E.g. its a new plan given. The 2 cases is that there's that delay you mention or its a replan during robot execution. 

The state reset if the time is too long seems like a better fit to be executed in `setPlan` than in this function

SteveMacenski:
Architecturally, this PR needs work, there are large chunks of code added to existing functions that should be broken out into semantically meaningful chunks and put into well named functions so the code is better self documenting and easier for users to understand.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: This is another place where semantically meaningful functions and comments are important

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: This should be being handled in the rotateToHeading method, not separately implemented. We can't have multiple rotation implementations using different methods.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: If its on approach though, which is when this matters, wouldn't it be effectively a straight line path? 

This breaks encapsulation of software functions by having the `getLookAheadPoint` return this value and does so in a pretty not-great way as an implicit return type when the main return value is explicit. Architecturally, this is really not great and I think and L2 norm would do effectively the same job since we should only be slowing down when in close proximity to the goal. 

If integrated path distance is required, it should not be being returned from the `getLookAheadPoint` method.

SteveMacenski:
>  I would say that angular velocity still needs profiles (especially when rotating towards heading in position mode.)
I don't think about it only as a smoother, but also as a limiter. Imagine a robot without payload heading to pick up some lab samples. While empty it can go fast towards the goal. After picking up the samples, user sets limits that need to be respected in order to bring them safely.

I agree in pure rotations having some control for smooth behavior is nice. However, for the path tracking PP algorithm part, it should be the one dictating the angular velocity based on the curvature of the path. The 'Regulation' part is only on the linear velocities. If you deviate from the angular velocities, this is no longer a PP and it will not be tracking the path accurately. 

We actually used to have kinematic limits in place until https://github.com/ros-planning/navigation2/issues/2621 which brought up an issue there and we removed it in https://github.com/ros-planning/navigation2/pull/2631. @dpm-seasony @vinnnyr can you comment on this PR / thoughts?

Perhaps some of this work doesn't need to be done in RPP itself but in the velocity smoother https://github.com/ros-planning/navigation2/issues/2633 that interacts at the lower-levels that can be tuned or used as appropriate for a given user. Certain elements of this PR I think could stay in that situation, but it would simplify the constraints that we needed to apply.  

> Yes, making jerk limits too low will make robot oscillate for sure, that way you are making it really "lazy". Also concerning @SteveMacenski question: My stance is that if kinematic limits make robot oscillate or deviate from path, RPP regulated parameters are simply not properly set (or limits are unreasonable). User has entered the limits that need to be respected, and they are being respected, but regulated parameters need some tuning for proper performance.

This needs to be optional to turn off kinematic limiting so that users that don't wish to deeply model their platform can use this without issue.

Aposhian:
>Perhaps some of this work doesn't need to be done in RPP itself but in the velocity smoother https://github.com/ros-planning/navigation2/issues/2633 that interacts at the lower-levels that can be tuned or used as appropriate for a given user.

While I think there is a place for a velocity smoother, a velocity smoother cannot intelligently produce planned velocities to stop at the end of the path.

doisyg:
> > Perhaps some of this work doesn't need to be done in RPP itself but in the velocity smoother #2633 that interacts at the lower-levels that can be tuned or used as appropriate for a given user.
> 
> While I think there is a place for a velocity smoother, a velocity smoother cannot intelligently produce planned velocities to stop at the end of the path.

I agree. This all comes down to not overshooting a goal.
Safe velocities and accelerations/decelerations can be enforced by a velocity smoother (whereas it should be part of nav2 or a lower level node is another debate), BUT the controller needs to be aware of these kinematic constraints otherwise it will generate infeasible `cmd_vel` which will lead to sub-optimal or impossible control.
The example of a robot with a very slow deceleration capability illustrate the problem well: there will be a larger goal overshoot if the controller doesn't start to deccelerate sufficiently before the goal.

vinnnyr:
I don't think this change will reintroduce the problems we encountered in 2621 since as far as I can tell, this change does not account for the robots _actual_ speed, it just provides limits on the _commanded_ speed. (Pending clarification on one line in the PR)

Problems we ran into in the earlier iteration of kinematic limiting in RPP were related to our robot not having great velocity control when the velocities were relatively small. This is a problem I describe in the velocity smoothing ticket here (see Deadband section): https://github.com/ros-planning/navigation2/issues/2633#issuecomment-1030088443 

> We actually used to have kinematic limits in place until https://github.com/ros-planning/navigation2/issues/2621 which brought up an issue there and we removed it in https://github.com/ros-planning/navigation2/pull/2631. @dpm-seasony @vinnnyr can you comment on this PR / thoughts?


As far as the "where should this live" I don't really have a strong preference as it relates to our use case. If there is a desire to put it all in the separate smoother, I can imagine a version of the proposed smoother where the controller calls a smoother service to prevent overshoot.

Not sure if that obfuscates the meaning of a smoother though, so perhaps the "prevent overshoot" portion should remain in RPP / controller and the jerk limit + ang vel limits could remain in the vel smoother?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
angstrem98: Well, the remaining path length for sure must me in meters. Also, we could be approaching the goal on a circular path, like robot driving on a circle?

I was thinking about one solution to lower computational cost. When we are "far away" from the goal we just make the robot speed up. But when we are "near enough" the goal we can start calculating remaining integrated path distance and come to a nice stop at goal. "Near enough" goal could be defined as a parameter (`safe_braking_distance` maybe?), let's say 5 meters default.
In order to detect if we are near enough goal we could simply use L2 distance to goal (L2 dist < `safe_braking_distance^2`).
Other approach is to use your already tried method - find a lookahead point (5 meters ahead) and at the end of path optimal lookahead distance and actual lookahead distance will differ - we can start calculating the remaining path and robot comes to a smooth stop. (Please note this isn't lookahead point used for RPP, it's just to detect if we are at `safe_braking_distance` or closer to the goal )

angstrem98:
> This needs to be optional to turn off kinematic limiting so that users that don't wish to deeply model their platform can use this without issue.
This is a very good idea. 

> Not sure if that obfuscates the meaning of a smoother though, so perhaps the "prevent overshoot" portion should remain in RPP / controller and the jerk limit + ang vel limits could remain in the vel smoother?

I'd definitely keep "prevent overshoot" behavior inside the controller, also the motion profile when turning to heading. When it comes to angular velocity / accel / jerk limits, I guess limiting them inside the controller or inside the velocity smoother leads to same behavior?
I'm leaning a bit towards keeping all the limits inside the controller since user can set them in one location. As far as I know, DWB controller also has all the limits inside it?

If you want to move angular limit to velocity smoother and keep it only inside rotate to heading, sure we can do it if it fits user needs the most.

SteveMacenski:
> I'd definitely keep "prevent overshoot" behavior inside the controller

Agreed, the slowing on the approach is the "meat" of this PR in my eyes .

Maybe it all can / should stay, but I think maybe that's better to reassess after this has a little more design work done on it so its easier to read / understand / broken out into functional chunks.

PR #2818, opened 2022-02-14
Static Layer updates are ignored 

andriimaistruk:
---

## Basic Info

| Info | Static Layer updates are ignored |
| ------ | ----------- |
| Ticket(s) this addresses   | No |
| Robotic platform tested on | Custom Gazebo robot simulation |

---

* removed resizing on update message callback as it blocks updates that are out of bounds of the very first one

---

#### For Maintainers: 


SteveMacenski:
Please describe the reason for this change in more detail

andriimaistruk:
When the callback processes a message of type map_msgs::msg::OccupancyGridUpdate, this message can carry new costmap that's only 3x3 meters when the whole map is of size 100x200 m, and resizing the whole map to dims 3x3 doesn't make sense for me.

PR #2829, opened 2022-02-22
remove unused odometry smoother in bt navigator

alsora:
Signed-off-by: Alberto Soragna <alberto.soragna@gmail.com>

There's an unused `OdomSmoother` object in the BT Navigator code.
Removing it to avoid creation of unnecessary subscription.

#### For Maintainers: 

### DIFF
--- a/nav2_bt_navigator/include/nav2_bt_navigator/bt_navigator.hpp
+++ b/nav2_bt_navigator/include/nav2_bt_navigator/bt_navigator.hpp
@@ -89,6 +89,9 @@ protected:
   poses_navigator_;
   nav2_bt_navigator::NavigatorMuxer plugin_muxer_;
 
+  // Odometry smoother object
+  std::shared_ptr<nav2_util::OdomSmoother> odom_smoother_;
+
   // Metrics for feedback
   std::string robot_frame_;
   std::string global_frame_;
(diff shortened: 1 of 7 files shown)


SteveMacenski:
So this looks perfectly correct, but wouldn't it be better if we instead kept the `odom_smoother_` in the BT navigator and instead removed the 2 instances of it in each navigator type. Then we could pass a shared pointer to the single instance `odom_smoother_` to each of the navigators to share. That would reduce the subscribers from 3 to 1 (vs 3 to 2).

alsora:
Sure that would work, but  I think that it would go in a different direction from how the code is currently structured.
Right now the navigators do not rely on any topic/data coming from the BTNavigatorNode, but rather they create their own entities.
Moreover, there are no requirements for a navigator  to use/care about a smoothed odometry, it's just a simple detail that currently all implementations use it.

I don't have strong opinions/preferences, so if you would still prefer to have only 1 smoother living in the node and shared with the navigators, then I can do that.

SteveMacenski:
I think that makes most sense for now. I wouldn't be opposed to a redesign of BT navigator in general, but that's a topic for another time I think than this PR :smile:

SteveMacenski (approved):

SteveMacenski:
By the way, this did not build, another PR came to fix that https://github.com/ros-planning/navigation2/pull/2836. I know that's the role of CI, but we're currently having trouble with that due to the Rolling 20.04 -> 22.04 transition

Issue #2808, opened 2022-02-08
 [Smac] Crash using footprint instead of radius :bug: 


matijazigic:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20
- ROS2 Version:
  - Rolling
- Version or commit hash:
  - 5a17db5a07e1573d81be33f0b737bcac8c43d2d1
- DDS implementation:
  - default

#### Steps to reproduce issue

```

1. Use smack planner and default turtlebot3 world and map
2. Use footprint instead of radius
3. Send goal

```

#### Additional information
Crash report: 
[component_container_isolated-6] [WARN] [1644321675.406004450] [planner_server]: GridBased plugin failed to plan calculation to (0.43, 1.77): "vector::reserve"
[component_container_isolated-6] [WARN] [1644321675.406055584] [planner_server]: [compute_path_to_pose] [ActionServer] Aborting handle.

    #robot_radius: 0.22
    footprint: "[[0.2, 0.1], [0.2, -0.1], [-0.2, -0.1], [-0.2, 0.1]]"

    planner_server:
      ros__parameters:
        planner_plugins: ["GridBased"]
        use_sim_time: True
        GridBased:
          plugin: "nav2_smac_planner/SmacPlannerLattice"
          tolerance: 0.5                      # tolerance for planning if unable to reach exact pose, in meters, for 2D node
          downsample_costmap: false           # whether or not to downsample the map
          downsampling_factor: 1              # multiplier for the resolution of the costmap layer (e.g. 2 on a 5cm costmap would be 10cm)
          allow_unknown: false                # allow traveling in unknown space
          max_iterations: 1000000             # maximum total iterations to search for before failing (in case unreachable), set to -1 to disable
          max_on_approach_iterations: 1000    # maximum number of iterations to attempt to reach goal once in tolerance, 2D only
          max_planning_time: 3.5              # max time in s for planner to plan, smooth, and upsample. Will scale maximum smoothing and upsampling times based on remaining time after planning.
          motion_model_for_search: "DUBINS"    # 2D Moore, Von Neumann; Hybrid Dubin, Redds-Shepp; State Lattice set internally
[...]

PR #2831, opened 2022-02-25
Fix large angle changes in lattice primitive generator

mattbooker:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2808 |

---

* Removed any large angular changes in the lattice motion primitives
* Added more tests for the lattice generator

None needed.

---

#### For Maintainers: 


nav2_smac_planner/lattice_primitives/trajectory_generator.py
```diff
@@ -566,3 +569,39 @@ def generate_trajectory(
         trajectory_path = self._create_path(trajectory_params, primitive_resolution)
 
         return Trajectory(trajectory_path, trajectory_params)
+
+if __name__ == "__main__":
+    test = TrajectoryGenerator({"turning_radius":0.5})
```
SteveMacenski, line 574: Should this be here?

mattbooker: removed it in latest. Forgot to remove my testing code...

SteveMacenski:
Can you regenerate the existing primitive files so that they have this fix?

SteveMacenski (approved):

SteveMacenski:
https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/test/default_model.json#L1794

it also looks like the default 0.4m one in `test/` needs to be regenerated too, not just the 0.5m/1.0m ones in the generator

nav2_smac_planner/lattice_primitives/helper.py
```diff
@@ -17,7 +17,7 @@
 
 def normalize_angle(angle):
     """
-    Normalize the angle to between [-pi, pi).
+    Normalize the angle to between (-pi, pi].
```
SteveMacenski, line 20: I'll need to test this locally too with a few of these files. I thought we needed 0-2PI not -PI to PI. I don't see how it was working before looking at the parsers https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/node_lattice.cpp but clearly it was so I need to understand what's going on there

nav2_smac_planner/lattice_primitives/helper.py
```diff
@@ -17,7 +17,7 @@
 
 def normalize_angle(angle):
     """
-    Normalize the angle to between [-pi, pi).
+    Normalize the angle to between [0, pi).
```
SteveMacenski, line 20: 0-2PI

PR #2834, opened 2022-02-27
Add all action options

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2415 |

---

* Added optional arguments for action requests

* PR includes docs changes in this package
* Docs in navigation.ros.org also need updating

---

#### For Maintainers: 

### DIFF
--- a/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
+++ b/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
@@ -93,7 +93,7 @@ class BasicNavigator(Node):
         self.initial_pose = initial_pose
         self._setInitialPose()
 
-    def goThroughPoses(self, poses):
+    def goThroughPoses(self, poses, behavior_tree=''):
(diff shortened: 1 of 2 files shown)


nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -274,17 +278,17 @@ def waitUntilNav2Active(self, navigator='bt_navigator', localizer='amcl'):
         self.info('Nav2 is ready for use!')
         return
 
-    def getPath(self, start, goal, planner_id=None):
```
tonynajjar, line 277: We could also use the "None way" everywhere (set the variable if it is not None)

SteveMacenski (approved):
I like this non-None way, makes it more clean to read. I think this is good! Just needs the complementary updates to the docs on navigation.ros.org and we can merge!

PR #2837, opened 2022-03-01
Launch single

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Added a package that allows path metrics to be created and results to be displayed. 

---

#### For Maintainers: 


nav2_planner_metrics/nav2_planner_metrics/metrics.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Wait for navigation to fully activate
+    navigator.waitUntilNav2Active()
+
+    # Get the costmap for start/goal validation
+    costmap_msg = navigator.getGlobalCostmap()
+    costmap = np.asarray(costmap_msg.data)
+    costmap.resize(costmap_msg.metadata.size_y, costmap_msg.metadata.size_x)
+
+    planners = ['NavFn']
```
SteveMacenski, line 119: Add multiple planners here and I think this is good

nav2_planner_metrics/nav2_planner_metrics/metrics.py
```diff
@@ -0,0 +1,157 @@
+#! /usr/bin/env python3
```
SteveMacenski, line 1: I don't want to put this into a new package unless we're planning on doing alot more here in the future, which I don't think either of us are. I think this should go into `/tools` https://github.com/ros-planning/navigation2/tree/main/tools

tools/metrics.py
```diff
@@ -0,0 +1,157 @@
+#! /usr/bin/env python3
```
SteveMacenski, line 1: Make a new directory until `/tools` called `/planner_benchmarking` and place these in there with the 3 10/15/20% occupancy maps

SteveMacenski (changes requested):

SteveMacenski (approved):

PR #2840, opened 2022-03-05
edge fix

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Bug fix for expanding nodes outside the map. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/collision_checker.cpp
+++ b/nav2_smac_planner/src/collision_checker.cpp
@@ -86,6 +86,15 @@ bool GridCollisionChecker::inCollision(
   const float & angle_bin,
   const bool & traverse_unknown)
 {
+  // Check to make sure cell is inside the map
+  if (outsideRange(0, costmap_->getSizeInCellsX(), x) ||
+      outsideRange(0, costmap_->getSizeInCellsY(), y))
+  {
+    return false;
+  }
+
+
+
   // Assumes setFootprint already set
   double wx, wy;
   costmap_->mapToWorld(static_cast<double>(x), static_cast<double>(y), wx, wy);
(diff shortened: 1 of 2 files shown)


nav2_smac_planner/src/collision_checker.cpp
```diff
@@ -86,6 +86,15 @@ bool GridCollisionChecker::inCollision(
...
   const bool & traverse_unknown)
 {
+  // Check to make sure cell is inside the map
+  if (outsideRange(0, costmap_->getSizeInCellsX(), x) ||
+      outsideRange(0, costmap_->getSizeInCellsY(), y))
+  {
+    return false;
+  }
+
+
```
SteveMacenski, line 96: Remove extra lines

SteveMacenski: Actually, run the linters ament_cpplint and ament_uncrustify and fix any issues, there are probably more

nav2_smac_planner/src/collision_checker.cpp
```diff
@@ -86,6 +86,15 @@ bool GridCollisionChecker::inCollision(
   const float & angle_bin,
   const bool & traverse_unknown)
 {
+  // Check to make sure cell is inside the map
+  if (outsideRange(0, costmap_->getSizeInCellsX(), x) ||
```
SteveMacenski, line 90: Since they're both `0` as min, you don't need to specify that as an input, just have `min` be `0` in the function itself.

nav2_smac_planner/src/collision_checker.cpp
```diff
@@ -164,4 +173,9 @@ float GridCollisionChecker::getCost()
   return static_cast<float>(footprint_cost_);
 }
 
+bool GridCollisionChecker::outsideRange(float min, unsigned int max, float value)
```
SteveMacenski, line 176: set as `const float &` and `const T &` so there are no copies

SteveMacenski (changes requested):
LGTM, just some clean up

SteveMacenski (approved):

Issue #2804, opened 2022-02-05
Update BT for only replanning when required + infrequent refinement


SteveMacenski:
CC @jwallace42 

Considerations
- Add the BT! 
- Not just replanning if totally invalid, but sufficiently higher cost
- Figuring out what the right level of replanning is (every 10, 20, 30, 40, ... seconds) 
- Asking users to trial and give us feedback

This should hopefully resolve oscillations problems forever in Nav2 + allow more consistent behavior

PR #2841, opened 2022-03-07
Mixed replan

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2804 |
| Primary OS tested on | Ubutu |

---

To allow mixed replanning, (i.e replan when time has expired or the path is invalid) a global_time_expired node was added. This is because we don't want the node to reset the time if it is idle. 

The node was then added to the BT to only replan if path is invalid or the time threshold has passed. 

---

#### For Maintainers: 


nav2_behavior_tree/plugins/condition/global_time_expired_condition.cpp
```diff
@@ -0,0 +1,68 @@
...
+  getInput("seconds", period_);
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+  start_ = node_->now();
+}
+
+BT::NodeStatus GlobalTimeExpiredCondition::tick()
+{
+  if (first_time) {
+    first_time = false;
+    RCLCPP_INFO(node_->get_logger(), "FIRST TIME");
```
SteveMacenski, line 41: Remove this log

nav2_behavior_tree/plugins/condition/global_time_expired_condition.cpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // Determine how long its been since we've started this iteration
+  auto elapsed = node_->now() - start_;
+
+  // Now, get that in seconds
+  auto seconds = elapsed.seconds();
+
+  if (seconds < period_) {
+    // std::cout << "Time expired: " << "Failure" << std::endl;
```
SteveMacenski, line 53: Remove commented out code

nav2_behavior_tree/plugins/condition/global_time_expired_condition.cpp
```diff
@@ -0,0 +1,68 @@
...
+    return BT::NodeStatus::FAILURE;
+  }
+
+  // std::cout << "Time expired: " << "Success" << std::endl;
```
SteveMacenski, line 57: Remove commented out code

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml
```diff
@@ -10,8 +10,11 @@
       <PipelineSequence name="NavigateWithReplanning">
         <RateController hz="1.0">
           <RecoveryNode number_of_retries="1" name="ComputePathToPose">
-            <Fallback> 
+            <Fallback>
               <ReactiveSequence>
+                <Inverter>
+                  <GlobalTimeExpired/>
```
SteveMacenski, line 16: I don't understand why this is added

SteveMacenski (changes requested):
I'm not entirely sure what this is with respect to

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml
jwallace42: To trigger re-planning at a desired frequency.

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml
```diff
@@ -10,8 +10,11 @@
       <PipelineSequence name="NavigateWithReplanning">
         <RateController hz="1.0">
```
SteveMacenski, line 11: Mhm, there's a rate controller here, I think this should be dropped to 2hz so that its every 500ms we can evaluate things.

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml
```diff
@@ -10,8 +10,11 @@
       <PipelineSequence name="NavigateWithReplanning">
```
SteveMacenski, line 10: Lets make this into another file, and that way we can rename that file. This file says `only` but if we add in the occasionally replanning, this name no longer describes what it is that we have for real. 

`navigate_w_recovery_and_constant_replanning_and_if_path_becomes_invalid.xml`

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml
SteveMacenski: See comment below - make this change only for the new file

SteveMacenski:
I think this work is going to be the new default, so I want to make sure this is solid and then ask some other people to take a look and play with it in their real robots in deployed use. then we can merge / change defaults / work on the documentation updates to support

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_constant_replanning_and_if_path_becomes_invalid.xml
```diff
@@ -0,0 +1,43 @@
...
+  <BehaviorTree ID="MainTree">
+    <RecoveryNode number_of_retries="6" name="NavigateRecovery">
+      <PipelineSequence name="NavigateWithReplanning">
+        <RateController hz="2.0">
+          <RecoveryNode number_of_retries="1" name="ComputePathToPose">
+            <Fallback>
+              <ReactiveSequence>
+                <Inverter>
+                  <GlobalUpdatedGoal/>
+                </Inverter>
```
SteveMacenski, line 16: This doesn't appear to have your new BT node in it :laughing:

SteveMacenski: Or the 15s setting

jwallace42:
Okay, now it resets the timeout if the path has changed. 

Do you have any good ideas for the name of this node? 

PlannerTimerExpiredCondition
PlannerTimeoutCondition
PathExpiredTimer


I like the last one, but let know what you think.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/global_time_expired_condition.hpp
```diff
@@ -55,13 +56,15 @@ class GlobalTimeExpiredCondition : public BT::ConditionNode
...
-      BT::InputPort<double>("seconds", 1.0, "Seconds")
+      BT::InputPort<double>("seconds", 1.0, "Seconds"),
+      BT::InputPort<nav_msgs::msg::Path>("path")
     };
   }
 
 private:
   rclcpp::Node::SharedPtr node_;
   rclcpp::Time start_;
+  nav_msgs::msg::Path prev_path;
```
SteveMacenski, line 67: trailing underscore for a class member

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/global_time_expired_condition.hpp
```diff
@@ -55,13 +56,15 @@ class GlobalTimeExpiredCondition : public BT::ConditionNode
   static BT::PortsList providedPorts()
   {
     return {
-      BT::InputPort<double>("seconds", 1.0, "Seconds")
+      BT::InputPort<double>("seconds", 1.0, "Seconds"),
+      BT::InputPort<nav_msgs::msg::Path>("path")
```
SteveMacenski, line 60: I think this node needs a new name if its now specific to the path. Brainstorm some ideas?

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_constant_replanning_and_if_path_becomes_invalid.xml
```diff
@@ -0,0 +1,46 @@
...
+<root main_tree_to_execute="MainTree">
+  <BehaviorTree ID="MainTree">
+    <RecoveryNode number_of_retries="6" name="NavigateRecovery">
+      <PipelineSequence name="NavigateWithReplanning">
+        <RateController hz="2.0">
+          <RecoveryNode number_of_retries="1" name="ComputePathToPose">
+            <Fallback>
+              <ReactiveSequence>
+                <Inverter>
+                  <GlobalTimeExpired seconds="15"/>
```
SteveMacenski, line 15: You're also missing the `path="{path}"` port -- I don't think this works without that

SteveMacenski:
I'm super bad with naming, unfortunately. 

I think "Global" is needed because of the global nature of it  (unless we can use a non-global one that resets when not currently being ticked?) 

GlobalPath[something]Timer?

GlobalTimerOrPathChange?

GlobalTimerDuringPath?

I tried to think also if there's a way to decouple this into 2 or more nodes and really can't. Getting that flag of a path change is kind of needed unfortunately.

jwallace42:
PathExpiredTimer

SteveMacenski:
PathExpiringTimer?

jwallace42:
Yeah! That sound good!

SteveMacenski:
Awesome, I think that’s a pretty good name!

nav2_behavior_tree/nav2_tree_nodes.xml
```diff
@@ -133,6 +133,11 @@
       <input_port name="seconds">Time to check if expired</input_port>
     </Condition>
 
+
+    <Condition ID="PathExpiringTimer">
+      <input_port name="seconds">Time to check if expired</input_port>
```
SteveMacenski, line 138: Path :-)

nav2_behavior_tree/plugins/condition/path_expiring_timer_condition.cpp
```diff
@@ -0,0 +1,76 @@
...
+PathExpiringTimerCondition::PathExpiringTimerCondition(
+  const std::string & condition_name,
+  const BT::NodeConfiguration & conf)
+: BT::ConditionNode(condition_name, conf),
+  period_(1.0),
+  first_time(true)
+{
+  getInput("seconds", period_);
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+  start_ = node_->now();
```
SteveMacenski, line 34: Remove this, your `first_time` condition will already set `start_` when it first triggers

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/path_expiring_timer_condition.hpp
```diff
@@ -0,0 +1,73 @@
...
+      BT::InputPort<nav_msgs::msg::Path>("path")
+    };
+  }
+
+private:
+  rclcpp::Node::SharedPtr node_;
+  rclcpp::Time start_;
+  nav_msgs::msg::Path prev_path_;
+  double period_;
+  bool first_time;
```
SteveMacenski, line 68: underscore :-)

SteveMacenski:
I think you're ready for the docs PR now though, the name / ports are baked in so when reviewing these final changes, if you include the docs PR I can merge them in as a lot

nav2_bringup/params/nav2_params.yaml
```diff
@@ -85,6 +85,7 @@ bt_navigator:
     - nav2_round_robin_node_bt_node
     - nav2_transform_available_condition_bt_node
     - nav2_time_expired_condition_bt_node
+    - nav2_path_expiring_timer_condition
```
SteveMacenski, line 88: I know its annoying, but there are other YAML files this should be added to in order to make sure they're loaded for all appropriate tests if we use this in the future.

What I would like to suggest is that you actually **delete** this parameter and list from all non `nav2_bringup`-located yaml files. That will stop us from needing to make this update to 20 files every time we add a new node. That way we only need to update the 3 yamls in Nav2 and the default in BT navigator for them all to have access to all of the given nodes

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_constant_replanning_and_if_path_becomes_invalid.xml
```diff
@@ -0,0 +1,46 @@
+<!--
+  This Behavior Tree replans the global path only if the path becomes invalid and it also has
```
SteveMacenski, line 2: This doesn't mention the constant replanning

SteveMacenski (approved):

SteveMacenski:
@doisyg I think this might be worth evaluating on your robots if you're open to doing a couple hours of testing for us. I believe this will resolve some of your behavioral issues around path tracking oscillation.

The tl;dr of this PR is that we added a BT that will replan on a significantly reduced rate (15s), but in exchange for that we also check if the existing path continues to be valid (e.g. not in collision with a newly detected obstacle). This makes it so that the robot won't oscillate between similar path solutions in opposite directions because the robot will "commit" to its course of action for much longer before trying to again refine the path. But it has the ability to ad-hoc replan if strictly required -- which is throttled to 1hz so that if you're in extremely dynamic settings you're not trying to replan every 10ms. 

Any feedback is valuable, but areas I think in particular that would be good to know:
- How do you like 15s? We picked it from a hat more or less knowing we wanted something between 10-30s
- Would it be helpful if we included not only "path invalid" check but also a "path is more expensive by X%" check based on experiential use, not just philosophy of "that sounds like a good idea". 
- Any other thoughts / issues you run into

I'm going to also make an open call on discourse about this. I think this has a strong case to be made the default behavior and I want to get real-user feedback.

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_constant_replanning_and_if_path_becomes_invalid.xml
```diff
@@ -0,0 +1,46 @@
...
+<root main_tree_to_execute="MainTree">
+  <BehaviorTree ID="MainTree">
+    <RecoveryNode number_of_retries="6" name="NavigateRecovery">
+      <PipelineSequence name="NavigateWithReplanning">
+        <RateController hz="2.0">
+          <RecoveryNode number_of_retries="1" name="ComputePathToPose">
+            <Fallback>
+              <ReactiveSequence>
+                <Inverter>
+                  <PathExpiringTimer seconds="15" path="{path}"/>
```
SteveMacenski, line 15: I know its a last minute thing -- but from the feedback on Discourse, can we push this down to 10s (and update any appropriate documentation)? I think 15s might be a little much and its the middle ground between the 5s request made

SteveMacenski (changes requested):

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_constant_replanning_and_if_path_becomes_invalid.xml
SteveMacenski: I marked where in docs the 10s update needs to be made, just 3 places

doisyg:
> @doisyg I think this might be worth evaluating on your robots if you're open to doing a couple hours of testing for us. I believe this will resolve some of your behavioral issues around path tracking oscillation.
> 
> The tl;dr of this PR is that we added a BT that will replan on a significantly reduced rate (15s), but in exchange for that we also check if the existing path continues to be valid (e.g. not in collision with a newly detected obstacle). This makes it so that the robot won't oscillate between similar path solutions in opposite directions because the robot will "commit" to its course of action for much longer before trying to again refine the path. But it has the ability to ad-hoc replan if strictly required -- which is throttled to 1hz so that if you're in extremely dynamic settings you're not trying to replan every 10ms.
> 
> Any feedback is valuable, but areas I think in particular that would be good to know:
> 
>     * How do you like 15s? We picked it from a hat more or less knowing we wanted something between 10-30s
> 
>     * Would it be helpful if we included not only "path invalid" check but also a "path is more expensive by X%" check based on experiential use, not just philosophy of "that sounds like a good idea".
> 
>     * Any other thoughts / issues you run into
> 
> 
> I'm going to also make an open call on discourse about this. I think this has a strong case to be made the default behavior and I want to get real-user feedback.

Sorry for the late reply...
This looks like a similar philosophy of what we've been doing with BTs since ros1 and move_base_flex, though not strictly the same.

My approach is to generate a global plan once (hence there is never any plan oscillation) and then let the controller a lot of freedom to execute it (global plan is used as a cue of which general way to take). In our application, it is the controller job to go around local dynamic obstacles. Then if the controller fails or some specific conditions are met, for instance no progress (by default 10s with less than 0.5m movement), a recovery behavior is triggered, some specific costmap cleaning is done, and finally a new global plan is generated. The whole action is considered failed if more than a configurable number of replan have been done.

This strategy proved to be extremely efficient for human dense environments for our small robot. On different environments or different robot size, we use different strategies.

I cannot comment further on what should be the default behavior for nav2 : I have never used it and at the beginning of the ros2 transition we ported the BTs we were using in ros1. I believe it should be anyway customized for the specific app/environment/behavior the end user wants to get. The fact that nav2 natively uses BTs makes this super easy and  awesome ! And I am still amazed that we independently converged toward a similar architecture that uses the same technical solution (behaviortree cpp). This is one of the reason we adopted nav2 so early.

SteveMacenski (approved):

SteveMacenski:
Got it, thanks for the feedback! I think what you suggest is very reasonable, especially for the situation that you're surrounded by people / tons of other moving objects. Even replanning might not be ideal for you in that situation. I think "in general" though most users would want replanning out of the box as a "smarter" behavior, especially if navigating while building a map or if not many moving people around, but the environment changes alot. 

It sounds like this suggestion is a middle ground between what is happening today in default and what would be happening on your systems, no? The rest of what you described is basically what's happening here minus the occasional replanning. That's basically this BT that Josh added a few weeks ago: https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml

doisyg:
> It sounds like this suggestion is a middle ground between what is happening today in default and what would be happening on your systems, no?

Agreed!



> [](https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml)
> 
> Even if you haven't / aren't using it, given your expertise, do you believe this is a better behavior than the constant 1hz (or some other rate, I suppose) BT we have today? I think from your comment, you probably do, since what you did is similar.

Yes, I've never been a fan of constant global replanning, especially to deal with "close" obstacles (but again, it is also because I ve always relied on "smart" controllers).



> We won't change the defaults right here, right now, to merge this PR but once its in, I'd like to start the transition conversation with relevant stakeholders once the PR is merged so its accessible to all.

:+1:

SteveMacenski:
I'm ready to merge this @jwallace42 but you have a CI test failure of your new BT node:

```
The test did not generate a result file:

[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from PathExpiringTimerConditionTestFixture
Warning: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
         at line 93 in /tmp/binarydeb/ros-rolling-tf2-0.21.0/src/buffer_core.cpp
[INFO] [1648396494.037878325] [test_behavior_tree_fixture]: Transforms are available now!
[ RUN      ] PathExpiringTimerConditionTestFixture.test_behavior
```

Now that we have an approval from an external user and given folks on Discourse a chance to respond, I think this is good to be merged.  I'm not sure what the steps forward for default status are though, I was surprised by the lack of engagement on Discourse on the topic.

 I believe this is the right direction to go, and I would feel more comfortable making it default if I had a fleet myself to test on to get comfortable with it and see the downsides. Its hard to change something this big without either (1) an expert user or (2) several general users testing it and giving the thumbs up.

SteveMacenski:
Awesome! Another big capability in!

Issue #2847, opened 2022-03-16
BT Navigator node segfaults on cleanup, crashes on reconfigure


guilyx:
Hi there, found a few things while playing around with transitioning `bt_navigator` down and up. I'll open a PR shortly, it seems like for the first bug, `haltAllActions` is called and an empty pointer `root_node` is passed to the function and accessed unsafely. This happens when a wrong XML file path is passed to the parameters, but IMO it shouldn't make the whole node crash if it couldn't parse the XML. There should be a warning somewhere at least ?

After this fixed my issue when transitioning to cleanup, I could deactivate and cleanup the node without any problem, but on configure would crash as we're redeclaring parameters without checking if they were already declared.

I have only tested this for Galactic as this didn't change from Galactic to Rolling, but I'll test on Rolling in a few days.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04/21.04
- ROS2 Version:
  - Galactic
- Version or commit hash: 28dbc9472e49accd646f2a063367ce82318f6eb4 (galactic branch latest?)
- DDS implementation:
  - Default for Galactic (didn't touch anything) so I think Cyclone ?

### On Cleanup Segmentation Fault

My XML paths are correct and the tree is populated correctly for both navigators.

#### Steps to reproduce issue

```
$ ros2 launch nav2_bringup bringup_launch.py
$ ros2 lifecycle set /bt_navigator deactivate
Transitioning successful
$ ros2 lifecycle set /bt_navigator cleanup
... (no prompt)
```

#### Stack Trace
```
[bt_navigator-20] Stack trace (most recent call last):
[bt_navigator-20] #15   Object "/home/ros/workspace/cmr_ros_ws/install/nav2_bt_navigator/lib/nav2_bt_navigator/bt_navigator", at 0x5555557783, in 
[bt_navigator-20] #14   Object "/usr/lib/aarch64-linux-gnu/libc.so.6", at 0x7fb7751d4f, in __libc_start_main
[bt_navigator-20] #13   Object "/home/ros/workspace/cmr_ros_ws/install/nav2_bt_navigator/lib/nav2_bt_navigator/bt_navigator", at 0x55555575f7, in 
[bt_navigator-20] #12   Object "/opt/ros/galactic/lib/librclcpp.so", at 0x7fb7dab1c3, in rclcpp::spin(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>)
[...]

PR #2848, opened 2022-03-16
Fix: bt_navigator crashes on lc transitions

guilyx:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #2847 ||
| Robotic platform tested on | Our Robot - but this doesn't require moving. I'll test on Steve's simulation later this week |

---

* Transitioning bt_navigator to the unconfigured state crashed in haltAllActions because the XML file path was wrong. When cleaning up one of the navigators, an empty root node was accessed which crashed the node.
* If the root node is an empty pointer I considered that nothing has to be alted - tell me if you also want the transition to fail?
* When we do manage to clean up the node, if we try to configure it again the node will crash because we're redeclaring an existing parameter without checking if it exists.

* When calling createTreeFromFile, we may want to check if rootNode is empty and send a warning if it is.
* Might be worth it to try and transition the states of each nodes to see if there are some misbehaviors here and there, I could do that.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/src/behavior_tree_engine.cpp
+++ b/nav2_behavior_tree/src/behavior_tree_engine.cpp
@@ -87,6 +87,8 @@ BehaviorTreeEngine::createTreeFromFile(
 void
 BehaviorTreeEngine::haltAllActions(BT::TreeNode * root_node)
 {
+  if (!root_node) return;
+  
   // this halt signal should propagate through the entire tree.
   root_node->halt();
 
--- a/nav2_behavior_tree/src/behavior_tree_engine.cpp
+++ b/nav2_behavior_tree/src/behavior_tree_engine.cpp
@@ -87,7 +87,9 @@ BehaviorTreeEngine::createTreeFromFile(
 void
 BehaviorTreeEngine::haltAllActions(BT::TreeNode * root_node)
 {
-  if (!root_node) return;
+  if (!root_node) {
+    return;
+  }
   
   // this halt signal should propagate through the entire tree.
   root_node->halt();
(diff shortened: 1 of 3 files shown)


nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -48,12 +54,16 @@ NavigateThroughPosesNavigator::getDefaultBTFilepath(
 {
   std::string default_bt_xml_filename;
   auto node = parent_node.lock();
+
+  if (!node->has_parameter("default_nav_through_poses_bt_xml")) {
   std::string pkg_share_dir =
```
SteveMacenski, line 59: This needs indentation for a bunch of lines you added to an if statement

nav2_behavior_tree/src/behavior_tree_engine.cpp
```diff
@@ -87,6 +87,8 @@ BehaviorTreeEngine::createTreeFromFile(
 void
 BehaviorTreeEngine::haltAllActions(BT::TreeNode * root_node)
 {
+  if (!root_node) return;
```
SteveMacenski, line 90: Extend to use brackets / multiple lines for styling of the codebase

SteveMacenski (changes requested):
LGTM, just small  changes

guilyx:
@SteveMacenski should be ready

SteveMacenski (approved):

SteveMacenski:
Thanks for the help!

Issue #2813, opened 2022-02-11
[BUG] Unable to resolve rosdep rolling keys in CI build


Aposhian:
## Bug report

**Required Info:**

- Operating System:
  - Nav2 build environment
- ROS2 Version:
  - rolling
- Version or commit hash:
  - main (`5a17db5a07e1573d81be33f0b737bcac8c43d2d1` right now)
- DDS implementation:
  - N/A

#### Steps to reproduce issue
Open a PR to `main`

#### Expected behavior
Build succeeds, and tests pass.

#### Actual behavior
Build fails on trying to install dependencies:
```ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
nav2_system_tests: No definition of [launch_testing] for OS version [focal]
nav2_costmap_2d: No definition of [launch_testing] for OS version [focal]
nav2_smac_planner: No definition of [ament_cmake_pytest] for OS version [focal]
nav2_util: No definition of [ament_cmake_pytest] for OS version [focal]
costmap_queue: No definition of [ament_cmake_gtest] for OS version [focal]
nav2_waypoint_follower: No definition of [ament_cmake_pytest] for OS version [focal]
nav2_simple_commander: No definition of [ament_pep257] for OS version [focal]
nav2_rotation_shim_controller: No definition of [ament_lint_auto] for OS version [focal]
nav2_recoveries: No definition of [ament_cmake_gtest] for OS version [focal]
nav2_navfn_planner: No definition of [ament_cmake_gtest] for OS version [focal]
nav2_behavior_tree: No definition of [ament_cmake_gtest] for OS version [focal]
nav_2d_msgs: No definition of [rosidl_default_generators] for OS version [focal]
nav2_map_server: No definition of [launch_testing] for OS version [focal]
nav2_smoother: No definition of [ament_cmake_pytest] for OS version [focal]
nav2_controller: No definition of [ament_cmake_pytest] for OS version [focal]
nav_2d_utils: No definition of [ament_cmake_gtest] for OS version [focal]
nav2_amcl: No definition of [ament_lint_auto] for OS version [focal]
dwb_msgs: No definition of [rosidl_default_runtime] for OS version [focal]
nav2_voxel_grid: No definition of [ament_cmake_gtest] for OS version [focal]
nav2_bringup: No definition of [launch_testing] for OS version [focal]
nav2_dwb_controller: No definition of [ament_cmake] for OS version [focal]
[...]

PR #2851, opened 2022-03-17
Ci focal rolling

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2813  |
| Primary OS tested on | NA |
| Robotic platform tested on | NA |

---

This is a hotfix to run CI on `ros:rolling-ros-base-focal` while `ros:rolling` is unstable during the transition to Ubuntu 22.04. This circumvents the rosdep keys being updated by avoiding calls to `rosdep update`.

None

---

Revert this change when everything is ready to build on Ubuntu 22.04.

#### For Maintainers: 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ _commands:
               sha256sum $PWD/lockfile.txt >> lockfile.txt
 
               apt-get update
-              rosdep update --rosdistro $ROS_DISTRO
+              # rosdep update --rosdistro $ROS_DISTRO
               dependencies=$(
                 rosdep install -q -y \
                   --from-paths src \
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 # docker build -t nav2:latest \
 #   --build-arg UNDERLAY_MIXINS \
 #   --build-arg OVERLAY_MIXINS ./
-ARG FROM_IMAGE=ros:rolling
+ARG FROM_IMAGE=ros:rolling-ros-base-focal
 ARG UNDERLAY_WS=/opt/underlay_ws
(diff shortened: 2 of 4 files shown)


Aposhian:
@ruffsl I got this to successfully install dependencies on CircleCI on my fork: https://app.circleci.com/pipelines/github/Aposhian/navigation2/2/workflows/addb9818-071e-406d-89a9-6cd4432d0782/jobs/2

Maybe this is something that can get our CI working while waiting for upstream packages and fixes?

ruffsl:
@Aposhian , clever minimal workaround!

I also thought of just skipping the rosdep update to stick with the cache archived in the old focal image, but figured it'd be better to bite the bullet and at least push through to jammy to keep up with rolling for updates. If we go with this PR, we won't be tracking any further sync updates, as opposed to at least falling back to galactic. But as you mention, such should only be a temporary measure.

As I've recently tried, there are a few API breaks from galactic using tf2 that we've sense adopted, and my attempt to build gazebo_ros from source still hits a snag using jammy: 

 - https://github.com/ros-planning/navigation2/pull/2852#issuecomment-1071957746

I'm fine with this approach, or we could wait to see if the above is easily unblocked. I'll let @SteveMacenski decide though.

SteveMacenski:
@ruffsl I know you don't know the answer to this, but how much longer do you think we'll be waiting for the issues to be resolved upstream of us (minus of course any Gazebo Ignition work we need to do locally to update our stuff to the new simulator)? If we've already sat through the most of it, I'm inclined to just wait it out and communicate our issues to OSRF so they can help unblock.

In the TSC meeting yesterday, Kat mentioned that if we could give her a list of specific problems, she could see what she could do on her side.

ruffsl (approved):

Issue #2833, opened 2022-02-25
Dynamically changing parameters in controller_server while computeControl is running causes deadlock


dkuenster:
## Bug report

**Required Info:**

- Operating System:
  - Debian 10
- ROS2 Version:
  - Rolling Source
- Version or commit hash:
  - 0a961df9cbe50356484877681abfdbeae6ba7b33
- DDS implementation:
  - Cyclone-DDS

#### Steps to reproduce issue
Use ParameterClient for controller_server from another node to dynamically change any parameters of controller_server while computeControl https://github.com/ros-planning/navigation2/blob/0a961df9cbe50356484877681abfdbeae6ba7b33/nav2_controller/src/controller_server.cpp#L344 is currently running.

#### Expected behavior

dynamicParametersCallback in controller_server.cpp https://github.com/ros-planning/navigation2/blob/0a961df9cbe50356484877681abfdbeae6ba7b33/nav2_controller/src/controller_server.cpp#L597 is blocked until we reach the current goal.

#### Actual behavior

![Screenshot from 2022-02-25 13-53-19](https://user-images.githubusercontent.com/14927048/155718512-6a7e9561-4f88-4d69-9b0e-5a17ea4fd96a.png)
I added an output right at the start of https://github.com/ros-planning/navigation2/blob/0a961df9cbe50356484877681abfdbeae6ba7b33/nav2_controller/src/controller_server.cpp#L344

Entering the the dynamicParametersCallback while computeControl is running causes a deadlock, controller_server stops sending a heartbeat and is shutdown.
Despite what was said in #2704 they seem to be running in the same thread, which causes the deadlock, as computeControl holds the mutex until it has reached the goal or it gets aborted.
If you remove the mutex from dynamicParametersCallback changing the parameters while computeControl is running works fine.
I'm however not sure why this causes follow_path to abort the current goal handle.

SteveMacenski:
> Use ParameterClient for controller_server from another node to dynamically change any parameters of controller_server while computeControl 

That is explicitly unsupported. It is dangerous to change controller parameters while an existing plugin is running. Changing of kinematic values or other core-behavioral parameters during execution could cause unstable behavior. This is a unique attribute of the controller server since this server deals with-- well.. the _control_ :wink: . 

[...]

PR #2858, opened 2022-03-22
nav2_controller: stop mutex from deadlocking on parameter change

dkuenster:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2833 |
| Primary OS tested on | Debian |
| Robotic platform tested on | gazebo simulation|

---

* Don't cause a deadlock when we try to dynamically change controller
parameters while the controller is running.
* Also provide a appropriate
error message on failure to change parameters.

Shouldn't be necessary

* Maybe add correct handling of dynamic parameter changes to the plugins

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -344,6 +344,7 @@ bool ControllerServer::findGoalCheckerId(
 void ControllerServer::computeControl()
 {
   std::lock_guard<std::mutex> lock(dynamic_params_lock_);
+
   RCLCPP_INFO(get_logger(), "Received a goal, begin computing control effort.");
 
   try {
@@ -596,12 +597,24 @@ void ControllerServer::speedLimitCallback(const nav2_msgs::msg::SpeedLimit::Shar
 rcl_interfaces::msg::SetParametersResult
 ControllerServer::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
 {
-  std::lock_guard<std::mutex> lock(dynamic_params_lock_);
   rcl_interfaces::msg::SetParametersResult result;
 
   for (auto parameter : parameters) {
     const auto & type = parameter.get_type();
     const auto & name = parameter.get_name();
+
+    // If we are trying to change the parameter of a plugin we can just skip it at this point
+    // as they handle parameter changes themselves and don't need to lock the mutex
+    if (name.find('.') != std::string::npos)
+      continue;
+
+    if (!dynamic_params_lock_.try_lock())
+    {
+      RCLCPP_WARN(get_logger(), "Unable to dynamically change Parameters while the controller is currently running");
+      result.successful = false;
+      result.reason = "Unable to dynamically change Parameters while the controller is currently running";
+      return result;
+    }
 
     if (type == ParameterType::PARAMETER_DOUBLE) {
       if (name == "controller_frequency") {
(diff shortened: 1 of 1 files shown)


nav2_controller/src/controller_server.cpp
```diff
@@ -599,10 +599,24 @@ ControllerServer::dynamicParametersCallback(std::vector<rclcpp::Parameter> param
   std::lock_guard<std::mutex> lock(dynamic_params_lock_);
```
SteveMacenski, line 599: But you're still in the mutex lock here - this is where the block is from

SteveMacenski (changes requested):

nav2_controller/src/controller_server.cpp
dkuenster: I know, but using try_lock for the parameter changes prevents the deadlock and the parameter changes just fail with an error message, while the controller keeps on running, instead of crashing.

dkuenster: I'm not sure how exactly ros2 handles switching between the callbacks, but this is tested in simulation and leads to the behavior I described in my answer above. If I send out parameter changes with the ros2 cli while the controller is currently running I get a failure with the error message I provided and there is no deadlock anymore.

SteveMacenski: It shouldn't though, and I think we're not quite connecting on the issue.  Consider the following 

- In L599 we `lock_guard` the mutex. https://en.cppreference.com/w/cpp/thread/lock_guard This line by itself is waiting for and capturing the mutex in the scope that it exists in.
- This PR wasn't modifying the `lock_guard` so its still being locked immediately once the callback is triggered, for the duration of the callback
-  When you do `dynamic_params_lock_.try_lock()` in L612, you've already locked and have acquired the muted from the `lock_guard` so this will always be successful. In fact the documentation for `try_lock()` states "If try_lock is called by a thread that already owns the mutex, the behavior is undefined. " https://en.cppreference.com/w/cpp/thread/mutex/try_lock so this is probably not correct

I think you need to remove the lock guard and replace it with the `try_lock()` call first thing in the callback. If it fails, exit. If it succeeds, proceed for the callback and then release the lock at the end of the callback.

SteveMacenski:
Any update? Does my concern make sense or is there something I'm overlooking?

nav2_controller/src/controller_server.cpp
dkuenster: Sry, was on vacation. That the lock_guard in L599 was still in there was indeed a mistake.Taking it out however still leads to the same behavior.
I also changed the lock_guard in the computeControl() function to try_lock/unlock with an error message. Still getting the same behavior in the simulation, so that is the current version.

nav2_controller/src/controller_server.cpp
```diff
@@ -343,7 +343,12 @@ bool ControllerServer::findGoalCheckerId(
 
 void ControllerServer::computeControl()
 {
-  std::lock_guard<std::mutex> lock(dynamic_params_lock_);
+  if (!dynamic_params_lock_.try_lock())
+  {
```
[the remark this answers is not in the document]
dkuenster: reverted the change in computeControl()

SteveMacenski (approved):

SteveMacenski:
Awesome! Thanks!  That return warning will be helpful too

Issue #2845, opened 2022-03-10
Removal of Z consideration in nav2_util::geometry_utils::euclidean_distance


vinnnyr:
## Feature request

Requesting we remove consideration of `z` when calculating euclidean distance throughout the repository by changing it in `nav2_util::geometry_utils::euclidean_distance`: https://github.com/ros-planning/navigation2/blob/a9744bd5e854e9e241df596522bc0aa4a02c4557/nav2_util/include/nav2_util/geometry_utils.hpp#L50

#### Feature description
We noticed this when we had an odom estimate that contained a non-zero `z` value in it's pose and was getting large euclidean distances in RPP due to the difference in `z` values.  Instead, we propose that euclidean distances be calculated projected onto the XY plane.

#### Implementation considerations
I can appreciate that removal of Z values _throughout_ the repo may have unintended consequences for users who were somehow depending on this, although I am not sure if there are users of nav2 who could be planning / controlling in 3D space. 

It seems like currently, changing `z` values is at least a case that was tested for:
https://github.com/ros-planning/navigation2/blob/a9744bd5e854e9e241df596522bc0aa4a02c4557/nav2_util/test/test_geometry_utils.cpp#L40

So if this is deemed too big of a change, I would instead propose a parameter be added to ignore Z consideration in distance calculations. 

Currently this change would by proxy affect:
- `removed_passed_goals_action.cpp`
- `truncate_path_action.cpp`
- `distance_traveled_condition.cpp`
- `distance_controller.cpp`
- `navigate_through_poses.cpp`
- `navigate_to_pose.cpp`
- `controller_server.cpp`
- `planner_server.cpp`
- `regulated_pure_pursuit_controller.cpp`
- `backup_recovery_tester.cpp`

vinnnyr:
I have already started to work on this BTW -- but opened feature ticket to continue conversation from Slack / have a ticket to link to the PR if needed

SteveMacenski:
This sounds fine to me, based on the discussion in slack. I suppose we could add a `bool useZ = true` (or `false`) to the API so that it is accessible but not by default and then use that bool for the calculation appropriately.

SteveMacenski:
Any update?

PR #2865, opened 2022-03-24
Removal of Z consideration in nav2_util::geometry_utils::euclidean_distance by default

vinnnyr:
Signed-off-by: Vinny Ruia <vinny.ruia@fireflyautomatix.com>

Throughout nav2, makes 2D euclidean distance (i.e. distance projected onto the XY plane) the default.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2845  |
| Robotic platform tested on | Proprietary Robot Simulation|

---

* Updates the euclidean distance to project onto the XY plane by default

Not sure.

---

N/A?

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/geometry_utils.hpp
+++ b/nav2_util/include/nav2_util/geometry_utils.hpp
@@ -43,48 +43,64 @@ inline geometry_msgs::msg::Quaternion orientationAroundZAxis(double angle)
 }
 
 /**
- * @brief Get the L2 distance between 2 geometry_msgs::Points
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
I very much appreciate the test additions! 

This is a subtle change, so  I think an entry in the migration guide would be valuable https://navigation.ros.org/migration/Galactic.html in case someone has an issue caused by this. If you can write that up, I can merge the pair.

vinnnyr:
ok! here: https://github.com/ros-planning/navigation.ros.org/pull/294

PR #2866, opened 2022-03-24
add the missing ports and description for the XML nodes

padhupradheep:
Adding the missing input ports for the BT-XML nodes given as part of the documentation. I hope, I have added all. 

I'm not sure, if the description for the ``planner_id`` is okay?

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
@@ -68,7 +68,7 @@ public:
           "Destinations to plan through"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
           "start", "Start pose of the path if overriding current robot pose"),
-        BT::InputPort<std::string>("planner_id", ""),
+        BT::InputPort<std::string>("planner_id", "Mapped name to the planner plugin type to use"),
       });
   }
 };
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
@@ -63,7 +63,7 @@ public:
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
           "start", "Start pose of the path if overriding current robot pose"),
-        BT::InputPort<std::string>("planner_id", ""),
+        BT::InputPort<std::string>("planner_id", "Mapped name to the planner plugin type to use"),
       });
   }
 };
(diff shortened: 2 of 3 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
```diff
@@ -68,7 +68,7 @@ class ComputePathThroughPosesAction
           "Destinations to plan through"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
           "start", "Start pose of the path if overriding current robot pose"),
-        BT::InputPort<std::string>("planner_id", ""),
+        BT::InputPort<std::string>("planner_id", "Mapped name to the planner plugin type to use"),
```
SteveMacenski, line 71: The entry string is the default, you would have to do `"", "description"` so that you don't set the default to the description for the string type. I"m not sure how else to get around that since the 2nd field could be cast into a string to be interpreted as a default

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
```diff
@@ -63,7 +63,7 @@ class ComputePathToPoseAction : public BtActionNode<nav2_msgs::action::ComputePa
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
           "start", "Start pose of the path if overriding current robot pose"),
-        BT::InputPort<std::string>("planner_id", ""),
+        BT::InputPort<std::string>("planner_id", "Mapped name to the planner plugin type to use"),
```
SteveMacenski, line 66: Here too

SteveMacenski (changes requested):

SteveMacenski:
Otherwise LGTM, thanks for getting this more complete!

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
padhupradheep: You are right, atleast from the [templates here in this API](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/725eba7e0abbc704284dec393706e5404f1472e3/include/behaviortree_cpp_v3/basic_types.h), I do not see any other option.

padhupradheep: But Input ports has just 2 templates, according to [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/725eba7e0abbc704284dec393706e5404f1472e3/include/behaviortree_cpp_v3/basic_types.h):

You either have this:

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/725eba7e0abbc704284dec393706e5404f1472e3/include/behaviortree_cpp_v3/basic_types.h#L282-L286 

or this:

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/725eba7e0abbc704284dec393706e5404f1472e3/include/behaviortree_cpp_v3/basic_types.h#L300-L306

SteveMacenski: Yes, but the `default` and `description` are both string types as the `dataType` is a `string` as well. I think you can see how that could be an issue that its impossible for the system to distinguish between the default value and the description for the case of string-based ports

SteveMacenski (approved):

Issue #2549, opened 2021-09-07
Migrate Recovery Server to Behavior Server


SteveMacenski:
After some thought around behavior primitives and actions, I think we should rename (and thusly update docs, package names, etc) for the recovery server to become the **behavior server**. At the end of the day, a recovery is simply a primitive behavior to get the robot out of a jam. As we're increasing the capabilities of Nav2 through behaviors and behavior trees, it doesn't make sense to restrict the recovery server to recoveries only. In particular, as we're creating capabilities like assisted teleop and docking, it makes sense to have these located in the same server for mental grouping and computational efficiency reasons.

The recovery server is uniquely designed to have many different action servers hosted within in (e.g. wait, spin, backup, etc) with their own action APIs to allow for customized plugin behaviors. But they are able to share some important information that is somewhat high overhead (e.g. TF buffers, costmap subscribers, footprints, etc) so that each behavior won't need to mimic much of the same information.

- Update docs to remove recovery -> behavior
- Update server and package names to behavior
- Update navigation concepts to have a discussion on behavior primitives 
- Update plugin API to behaviors rather than recoveries
- Think about what additional primitive behaviors might be nice to add (e.g. dock, assisted teleop, make a random move, follow the leader, etc) and file tickets to accomplish
- Remove recovery.hpp and have each recovery implement its own simple action server, this cycle frequency thing doesn't make sense for most behaviors.  

Perhaps provide some basic behaviors like 
- [ ] rotate to sp. heading
- [ ] Drive on heading for X distance (recommission backup plugin)
- [ ] assisted teleop
- [ ] "dance" or some example HRI behavior
- [ ] Back out of recent buffer of commands bc bad situation as recovery, then retry with another algo

MOLOCH-dev:
@SteveMacenski With your guidance, I have tried to do something similar for Assisted  Teleop [PR #2575](https://github.com/ros-planning/navigation2/pull/2575)
Does that seems like a reasonable approach for a behaviour server? I could get started on this

PR #2867, opened 2022-03-24
Recovery to behavior

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2549  |
| Primary OS tested on |  Ubuntu|
| Robotic platform tested on | Gazebo|

---

Renaming recovery server to behavior server

---

#### For Maintainers: 


nav2_behaviors/behavior_plugin.xml
```diff
@@ -0,0 +1,19 @@
+<class_libraries>
```
SteveMacenski, line 1: The readme needs more than just renaming - the content needs updating

nav2_behaviors/include/nav2_behaviors/behavior.hpp
```diff
@@ -48,26 +48,26 @@ enum class Status : int8_t
...
 
 /**
- * @class nav2_recoveries::Recovery
- * @brief An action server recovery base class implementing the action server and basic factory.
+ * @class nav2_behaviors::Behavior
+ * @brief An action server Behavior base class implementing the action server and basic factory.
  */
 template<typename ActionT>
-class Recovery : public nav2_core::Recovery
+class Behavior : public nav2_core::Behavior
```
SteveMacenski, line 55: I like the idea of using this as an opportunity to clarify the "Recovery Plugin" vs "Recovery.hpp"-based Plugin distinction. I think we should add another word to distinguish between "behaviors" that use this time interface which also makes pluguns vs "behaviors" that are plugins

`TimedBehavior` `CycleBehavior` something like that?

SteveMacenski (commented):
This all LGTM, just a couple of small bits. Obviously documentation is still in need

nav2_behaviors/include/nav2_behaviors/behavior.hpp
jwallace42: So, I renamed it for now but the timed behavior it is really a timed recovery because it has the vel_pub and stop robot func. 

We could factor out the vel_pub into the concrete implementation and provide a override for stopping the robot. `onActionTerimation` and `onActionSuccess`. 

However, I don't think that this is really important right now. 

I think we need to just add in some more behaviors and sort of get a feel for how things will look. 

I could not think of other behaviors we need besides assisted teleop but we could use some actions that are not recovery like.

nav2_behaviors/README.md
```diff
@@ -6,10 +6,26 @@ The package defines:
 - A `Recovery` template which is used as a base class to implement specific recovery timed action server - but not required.
 - The `BackUp`, `Spin` and `Wait` recoveries.
```
SteveMacenski: It looks like you forgot to delete the recovery stuff that mirrors it on top :smile:

nav2_behaviors/include/nav2_behaviors/behavior.hpp
SteveMacenski: I think that is still useful for other behaviors, no? It could be a following behavior, that would need the velocity publisher. Or maybe its a "dance"  behavior. 

Do you think `Timed` is the right keyword? I think its better, but I'd be open to suggestions

jwallace42: Yes, but we are restricted to behaviors that produce a velocity. 

Also, I don't even think it is a timed behavior interface. 
The timing is done in the concrete implementation by overriding `onRun` and `onCycleUpdate` which is left up to the implementer. 

Maybe `VelocityBehavior`?

SteveMacenski: > Yes, but we are restricted to behaviors that produce a velocity.

Well that's a good point, that's what I had in mind. What kind of other behaviors are you thinking? I think of "behavior"s being spin, backup, get out of bad situations, assisted teleop, dance, etc which are all velocity-based things. There may be other types of behaviors, but the point of this server is to share assets like the costmap and TF buffer, which are most relevant (I think?) to these types of behaviors.

I was thinking the name more in line with trying to describe what this class is helping to implement: behaviors that run on a fixed cycle or timeline.

jwallace42: Yeah, that makes sense. If we end up having thinking of more behaviors that need those assets but don't produce velocities we can always change it.

nav2_behaviors/README.md
jwallace42: Tarnation! Should be good now.

nav2_behaviors/include/nav2_behaviors/behavior.hpp
SteveMacenski: OK - so `CycleBehavior`? 

What were you thinking though? I'd like to hear what you had in mind

SteveMacenski:
There's also a number of updates to the documentation that should be made (e.g. concepts, plugins, migration, etc)

nav2_behaviors/include/nav2_behaviors/behavior.hpp
jwallace42: I don't think any of my ideas are really practical :)(Yell when the bot is close to an obstacle and some other silly stuff)! Also, sharing the costmap infers that the behaviors will produce a velocity.

I changed it to `TimedBehavior` if that works for you!

nav2_behaviors/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+The value of the centralized behavior server is to **share resources** amongst several behaviors that would otherwise be independent nodes. Subscriptions to TF, costmaps, and more can be quite heavy and add non-trivial compute costs to a robot system. By combining these independent behaviors into a single server, they may share these resources while retaining complete independence in execution and interface.
+
+See its [Configuration Guide Page](https://navigation.ros.org/configuration/packages/configuring-recovery-server.html) for additional parameter descriptions and a [tutorial about writing recovery behaviors](https://navigation.ros.org/plugin_tutorials/docs/writing_new_recovery_plugin.html).
```
SteveMacenski, line 13: The word "recovery" still appears here!

SteveMacenski: URL + "tutorial about writing recovery behaviors"

SteveMacenski: Might be worth going into this directory and `grep`ing for `recovery` `Recovery` `recoveries` `Recoveries` and `RECOVERY` to see if you missed others by accident

SteveMacenski (approved):
I see you found a few more :wink:

SteveMacenski:
You have a conflict, but LGTM

jwallace42:
> I see you found a few more wink

:laughing: Yep!

SteveMacenski (approved):
Just on Docs now

haryngod:
Hi @jwallace42
now I am trying to understand the nav2 source and I have a question for this pr.
I think this behavior server seems to be 'recovery behavior'.
Even though why did you change the name recovery to behavior?

jwallace42:
@haryngod Not all of behaviors in this server are recoveries. I would read the ticket https://github.com/ros-planning/navigation2/issues/2549 to get a better idea of the motivation for changing the name.

PR #2874, opened 2022-03-29
Is path valid test

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Add in a test for is path valid. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_system_tests/src/planning/CMakeLists.txt
+++ b/nav2_system_tests/src/planning/CMakeLists.txt
@@ -59,3 +59,5 @@ target_link_libraries(test_planner_plugins
 ament_add_gtest(test_planner_is_path_valid
               test_planner_is_path_valid.cpp
               planner_tester.cpp)
+
+ament_target_dependencies(test_planner_is_path_valid rclcpp geometry_msgs nav2_msgs ${dependencies})
--- a/nav2_system_tests/src/planning/CMakeLists.txt
+++ b/nav2_system_tests/src/planning/CMakeLists.txt
@@ -55,3 +55,7 @@ ament_target_dependencies(test_planner_plugins rclcpp geometry_msgs nav2_msgs ${
 target_link_libraries(test_planner_plugins
   stdc++fs
 )
+
+ament_add_gtest(test_planner_is_path_valid
+              test_planner_is_path_valid.cpp
+              planner_tester.cpp)
--- a/nav2_system_tests/src/planning/planner_tester.cpp
+++ b/nav2_system_tests/src/planning/planner_tester.cpp
@@ -76,6 +76,7 @@ void PlannerTester::activate()
   planner_tester_->onConfigure(state);
   publishRobotTransform();
   map_pub_ = this->create_publisher<nav_msgs::msg::OccupancyGrid>("map", 1);
+  path_valid_client_ = this->create_client<nav2_msgs::srv::IsPathValid>("is_path_valid");
   rclcpp::Rate r(1);
   r.sleep();
   planner_tester_->onActivate(state);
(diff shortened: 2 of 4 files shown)


jwallace42:
@SteveMacenski 

If you get a chance to look at the test that would be great. 

I am not really sure what is going wrong. The PlannerTester is running on a separate thread while the client is called on another thread. If you have any ideas that would be great!

padhupradheep:
Hey Wallace, I'm poor at explaining things, I pulled your branch and tweaked few things to make it work.. But yeah, I'm also really not sure, if that is the right way to go..

Here is the link of my version of your branch: https://github.com/padhupradheep/navigation2/tree/is_path_valid_test

Hope this helps..!

nav2_system_tests/src/planning/CMakeLists.txt
```diff
@@ -55,3 +55,7 @@ ament_target_dependencies(test_planner_plugins rclcpp geometry_msgs nav2_msgs ${
 target_link_libraries(test_planner_plugins
   stdc++fs
 )
+
+ament_add_gtest(test_planner_is_path_valid
+              test_planner_is_path_valid.cpp
+              planner_tester.cpp)
```
SteveMacenski, line 61: BTW, you're going to want to `ament_target_dependencies`

SteveMacenski:
All of these tests use the planner tester and add wrapped APIs to complete the tests using the `planner_server`'s APIs directly without calling the services / actions manually. You should do that too so you don't need to deal with any of the ROS-y stuff. Create some new functions inside of the planner tester to test the is path valid functions.

You can see examples if you trace back how `createPlan` (or `createPath`, I forget) is called in this tester and the setup done prior. You probably just need to setup a wrapper function for the private is path valid callback so it can be used in the test

nav2_system_tests/src/planning/test_planner_is_path_valid.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  bool is_path_valid = planner_tester->checkPathValid(path);
+  EXPECT_FALSE(is_path_valid);
+
+  for (float i = 0; i < 10; i += 0.05) {
+    for (float j = 0; j < 10; j += 0.05) {
+      geometry_msgs::msg::PoseStamped pose;
+      pose.pose.position.x = i;
+      pose.pose.position.y = j;
+      path.poses.push_back(pose);
```
padhupradheep, line 44: Just a small thought. We have this nice function that creates the plan for you? https://github.com/ros-planning/navigation2/blob/609342fa27016a1677768f93fb3c44dec98b1f55/nav2_system_tests/src/planning/planner_tester.hpp#L77

Maybe it would be handy?

nav2_system_tests/src/planning/planner_tester.cpp
```diff
@@ -396,6 +397,24 @@ TaskStatus PlannerTester::createPlan(
...
 
+bool PlannerTester::checkPathValid(nav_msgs::msg::Path & path)
+{
+  // create a fake service request
+  auto request = std::make_shared<nav2_msgs::srv::IsPathValid::Request>();
+  request->path = path;
+  auto result = path_valid_client_->async_send_request(request);
+
+  RCLCPP_INFO(this->get_logger(), "Waiting for service complete");
+  if (rclcpp::spin_until_future_complete(this->planner_tester_, result, std::chrono::seconds(10)) ==
```
padhupradheep, line 408: Maybe you wanted the timeout time in ms?

nav2_system_tests/src/planning/planner_tester.hpp
```diff
@@ -150,11 +151,14 @@ class PlannerTester : public rclcpp::Node
     ComputePathToPoseResult & path,
     const double deviation_tolerance = 1.0);
 
+
```
padhupradheep, line 154: Linter might be unhappy with the extra white line.

nav2_system_tests/src/planning/test_planner_is_path_valid.cpp
jwallace42: The goal of the first section it to create a invalid path. The `createPath` will create a valid path. 
I think separating them is good. But, honestly either way would probably be fine.

SteveMacenski (approved):
Perfect!

PR #2884, opened 2022-04-01
Better Costmap Error Message

DLu:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Unity simulated Jackal |

---

 * Improve the error messages in the costmap layers to say WHY the origin was out of bounds

 * None

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -571,8 +571,11 @@ ObstacleLayer::raytraceFreespace(
   if (!worldToMap(ox, oy, x0, y0)) {
     RCLCPP_WARN(
       logger_,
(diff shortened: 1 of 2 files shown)


nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -571,8 +571,11 @@ ObstacleLayer::raytraceFreespace(
   if (!worldToMap(ox, oy, x0, y0)) {
     RCLCPP_WARN(
       logger_,
-      "Sensor origin at (%.2f, %.2f) is out of map bounds. The costmap cannot raytrace for it.",
-      ox, oy);
+      "Sensor origin at (%.2f, %.2f) is out of map bounds (%.2f x %.2f at %.2f, %.2f). "
```
SteveMacenski, line 574: Can this be more explicit that the second set of numbers are the origin? I'm not sure I would read this and realize that it was that origin.

SteveMacenski (approved):

Issue #2752, opened 2021-12-21
Lifecycle manager re-transition after a crash of a managed node is respawned


tuxnese:
## Bug report

**Required Info:**

- Operating System:
Ubuntu 20.04
- ROS2 Version:
Galactic binaries
- Version or commit hash:
ros-galactic-navigation2	1.0.7-1focal.20211012.051615

- DDS implementation:
Cyclone

#### Steps to reproduce issue

1) Run a set of managed nodes passing the "respawn=True" parameter

2) Run 'nav2_lifecycle_manager', in my scenario the managed nodes are:
lifecycle_nodes = ['controller_server',
                       'planner_server',
                       'recoveries_server',
                       'bt_navigator',
                       'waypoint_follower']

3) Get the pid of the process of one of the managed nodes

ps -A | grep planner_server

4) Kill the process

#### Expected behavior

The lifecycle manager gets a heartbeat timeout and restarts all the nodes, making the system working again. 

#### Actual behavior

The reset fails and the navigation does not work.

Here is my log

[ERROR] [planner_server-3]: process has died [pid 13642, exit code -15, cmd '/opt/ros/galactic/lib/nav2_planner/planner_server --ros-args -r __node:=planner_server -r __ns:=/virtual_robot_5 --params-file /tmp/tmp0h27fliy --params-file /tmp/tmp6pjfl3ha --params-file /tmp/tmp72cf6i6t --params-file /tmp/tmpbsloxdxh --params-file /tmp/tmp7vpq0t0u --params-file /tmp/tmpl1zy647i --params-file /tmp/tmp33_23par -r cmd_vel:=input/navigation'].
[INFO] [planner_server-3]: process started with pid [14154]
[planner_server-3] [INFO] [1640075883.643358596] [virtual_robot_5.planner_server]: 
[planner_server-3] 	planner_server lifecycle node launched. 
[planner_server-3] 	Waiting on external lifecycle transitions to activate
[planner_server-3] 	See https://design.ros2.org/articles/node_lifecycle.html for more information.
[planner_server-3] [INFO] [1640075883.644622698] [virtual_robot_5.planner_server]: Creating
[planner_server-3] [INFO] [1640075883.651607220] [virtual_robot_5.global_costmap.global_costmap]: 
[planner_server-3] 	global_costmap lifecycle node launched. 
[planner_server-3] 	Waiting on external lifecycle transitions to activate
[...]

PR #2891, opened 2022-04-06
Foxy devel

iampaulidrobo:
---

## Basic Info

Operating System:
Ubuntu 20.04

ROS2 Version:
foxy-devel

| ------ | ----------- |
| Primary OS tested on | (Ubuntu 20.04,ros2 foxy |

---

1)The lifecycle manager gets a heartbeat timeout and restarts all the nodes, making the system working again if any nodes goes down.

*Made all the nodes in localization and navigation launch file to respawn as TRUE.
*Timer check added in Lifecycle_manager package node for checking status of node and activating it henceforth.

---

*To work it without 'respawn true' for each node rather than just lifecycle manager node.
*Adding reverse transition of node for better grip and functionality.

[navigation2-foxy-devel_.zip](https://github.com/ros-planning/navigation2/files/8428082/navigation2-foxy-devel_.zip)

#### For Maintainers: 


iampaulidrobo:
[navigation2-foxy-devel_.zip](https://github.com/ros-planning/navigation2/files/8428008/navigation2-foxy-devel_.zip)

PR #2893, opened 2022-04-06
Foxy devel

iampaulidrobo:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | (gazebo simulation of Tally) |

---

The lifecycle manager gets a heartbeat timeout and restarts all the nodes, making the system working again if any nodes goes down.

1)Made all the nodes in localization and navigation launch file to respawn as TRUE.
2)Timer check added in Lifecycle_manager package node for checking status of node and activating it henceforth.

---

*To work it without 'respawn true' for each node rather than just lifecycle manager node.
*Adding reverse transition of node for better grip and functionality.

#### For Maintainers: 


SteveMacenski:
This is again a `foxy` into `main` PR without your work in it as far as I can tell, it has tens of thousand of lines of code, please stop filing PRs. I gave you some more explicit instructions and questions in Slack to follow up on your git issues, please look over there!

PR #2894, opened 2022-04-07
server_respawn_lifecycleNode

iampaulidrobo:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | (Turtlebot3 simulation) |

---

*The lifecycle manager gets a heartbeat timeout and restarts all the nodes, making the system working again.

1)All node set to respawn true ,to make node running again in case of failure.
2)Lifecycle manager node edited with checking for CONFIGURE and ACTIVATE, and do them in case they are not ACTIVATED.

---

1)Reverse transition for the lifecycle nodes to function with modularity.

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/bringup/launch/localization_launch.py
+++ b/nav2_bringup/bringup/launch/localization_launch.py
@@ -85,6 +85,7 @@ def generate_launch_description():
             executable='map_server',
             name='map_server',
             output='screen',
+            respawn=True,
             parameters=[configured_params],
             remappings=remappings),
 
@@ -93,6 +94,7 @@ def generate_launch_description():
             executable='amcl',
             name='amcl',
             output='screen',
+            respawn=True,
             parameters=[configured_params],
             remappings=remappings),
 
@@ -101,7 +103,8 @@ def generate_launch_description():
             executable='lifecycle_manager',
             name='lifecycle_manager_localization',
             output='screen',
+            respawn=True,
             parameters=[{'use_sim_time': use_sim_time},
-                        {'autostart': autostart},
+                        {'autostart': True},
                         {'node_names': lifecycle_nodes}])
     ])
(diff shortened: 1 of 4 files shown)


nav2_bringup/bringup/launch/localization_launch.py
```diff
@@ -101,7 +103,8 @@ def generate_launch_description():
             executable='lifecycle_manager',
             name='lifecycle_manager_localization',
             output='screen',
+            respawn=True,
             parameters=[{'use_sim_time': use_sim_time},
-                        {'autostart': autostart},
+                        {'autostart': True},
```
SteveMacenski, line 108: Revert, autostart is a parameterization

nav2_bringup/bringup/launch/navigation_launch.py
```diff
@@ -138,8 +143,9 @@ def generate_launch_description():
             executable='lifecycle_manager',
             name='lifecycle_manager_navigation',
             output='screen',
+            respawn=True,
             parameters=[{'use_sim_time': use_sim_time},
-                        {'autostart': autostart},
```
SteveMacenski, line 142: Revert

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -30,130 +30,132 @@
...
-  // Whether to automatically start up the system
-  bool autostart_;
-
-  bool system_active_{false};
-};
-
-}  // namespace nav2_lifecycle_manager
-
-#endif  // NAV2_LIFECYCLE_MANAGER__LIFECYCLE_MANAGER_HPP_
+  class LifecycleManager : public rclcpp::Node
```
SteveMacenski, line 40: Replace original formatting

SteveMacenski (changes requested):

SteveMacenski:
This PR is critically flawed, we should discuss in the ticket. This does not in any way check or aware of the servers failing, its just naively just always trying to bring up, not only when a server was respawned due to a crash. This would cause a critical looping error if we wanted to bring down the lifecycle system safely.

iampaulidrobo:
Yes,sure.
The server comes up whenever any one node goes down.
@SteveMacenski

Issue #2888, opened 2022-04-04
Additional Behavior Server Primitives


SteveMacenski:
Perhaps provide some basic behaviors like

- [ ] rotate to sp. heading
- [ ] Drive on heading for X distance (recommission backup plugin)
- [ ] assisted teleop
- [ ] "dance" or some example HRI behavior
- [ ] Back out of recent buffer of commands bc bad situation as recovery, then retry with another algo

sathak93:
@SteveMacenski  i like to work on rotate to specific heading behaviour. could you elaborate on it.?

SteveMacenski:
It would be 
- Making a behavior to have the robot rotate to a desired relative heading (e.g. +90 or -90 from current orientation)

sathak93:
@SteveMacenski Why cant we use spin behaviour for this? Spin can rotate the robot to particular degree from current pose.

SteveMacenski:
Those are global angles, not relative angles

jwallace42:
I think adding a BT node for spin without increasing the recovery counter would be useful.

SteveMacenski:
Maybe a similar analog to your backup --> drive on heading conversion?

jwallace42:
The changes for the drive on heading added the ability to drive forward and retained the backup action. 

I guess there is no new functionality that we would need to support(I might be missing something though!).

SteveMacenski:
The wiggle behavior can be made by concatenating the spins in the BT with a sequence node. The backout of buffer of commands should be worked on in response to a specific issue. Else, we may prefer to instead have a simple trajectory planner that tries to move a robot into free space instead as a recovery with potential field methods. 

Either way, its something to be readdressed when the need arises. Assisted teleop is covered by another ticket

PR #2898, opened 2022-04-07
Drive on heading

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #2888 |

---

Extended the backup behavior to drive on heading. The behavior can now go forwards or backwards. 

Changes to the behavior server docs and bt nodes docs. 

---

#### For Maintainers: 


SteveMacenski:
What you suggest is definitely not unreasonable, but it is a very large change which will effect every user's configurations since that plugin will cease to exist and all behavior trees will need a different BT node used. I would like to be cautious due to that to try to make this as least disruptive as possible.

Semantically, I think about "back up" versus "drive straight" as 2 different action contexts even though they're representing related things. 

Within the BT, would it be possible to have a derived class from drive on heading for `backup` so that we still have a backup BT node that can be used in the BTs for semantically different tasks? So then we'd have both BT nodes available requiring no changes to end users BTs but then have the additional drive at current heading action (back up just checks its negative). Same with cancel.

I'm debating what I would want for the plugins in the behavior server. From one perceptive, we could do the analog thing here to have both a drive on current heading + backup behavior plugins. But that feels really redundant, and the BT node of `backup` could just call the renamed behavior plugin for the same results -- we just need to enforce the negative value in the backup version of the BT node. I don't think we need to duplicate the plugins code. What do you think?

I've never done this before, but I think if you created 2 instances in here https://github.com/ros-planning/navigation2/blob/a32ebfaa0ca161de9f6a0841fc1a536c3a0e75ba/nav2_behaviors/behavior_plugin.xml#L9 of the name "backup" and of "driveonheading", then it should make it so that you can load this same plugin based on both names if specified in a parameter file. You'd have to check though. If we did that, we could not duplicate any code, but also not require users to update their YAML files based on the change **to load**. We could add a depreciation warning into the description. 

That leaves the only open question is the **action topic ID**. For the high level BT node, both can use the drive on current heading name. Or, for the backup version, we could use the original backup name when it is registered.

 For the behavior plugin, the drive on current heading is fine. If we aliased the backup plugin loading name, I think it would load the plugin correctly, but if we didn't replace the behavior ID in the yaml from `backup` to `drive_on_heading`, then it would use the `backup` action topic which is wrong, I believe. So I think that implies then that the backup BT node should also use the backup moniker to prevent the need of users to change anything for the current behavior.

**but** that would brake the use of `backup` BT node to hook up to the `drive at current heading` server, without manually overriding it in the BT XML `server_name` field. 

Options, options, options...

jwallace42:
It is a lot of changes but the only thing they have to do is switch the BT xml file right? Just change the node names and switch sign of the values. 

But I really don't know much about the how the upgrading from galactic to humble works. It there something else they would have to do?

SteveMacenski:
> It is a lot of changes but the only thing they have to do is switch the BT xml file right? Just change the node names and switch sign of the values.

If you derive from this as a base class and make a derived class for backup, it would be the sign and when registering the plugin, using the backup name / action rather than drive on heading. 

> But I really don't know much about the how the upgrading from galactic to humble works. It there something else they would have to do?

If we didn't make it easy for them, they'd have to modify their configs and BT XMLs because the backup action wouldn't exist and the backup BT node wouldn't exist anymore.

jwallace42:
Alright, so the goal is still support old BT xml. I will just create a derived backup class. Would there be a plan to faze out the backup node or would we always have it?

SteveMacenski:
This is a good question. I think the BT node would be there forever, I think the actual backend separate action server / behavior plugin would be deprecated. The Backup BT node would just be redirected to that new server after the next distribution release. 

I think "DriveOnHeading" vs "BackUp" are semantically different to a autonomy system designer, so having both would make more clear in design context the intent of the move, so both would be nice.

I figure once we have the drive on heading + rotate, we can do a little demo of having the robot drive in a shape or something (e.g. drive 1 meter, turn 90 deg, drive 1 meter, turn 90 deg, ...) which could be helpful for calibration and a POC of its use. Those primitives are common for students in entry-level robotics courses to need to use, so I figure having them around is helpful. Updating the Python API with them would be even more useful. But that takes like 15 minutes, not a big deal

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ list(APPEND plugin_libs nav2_wait_cancel_bt_node)
 add_library(nav2_spin_cancel_bt_node SHARED plugins/action/spin_cancel_node.cpp)
 list(APPEND plugin_libs nav2_spin_cancel_bt_node)
 
-add_library(nav2_back_up_cancel_bt_node SHARED plugins/action/back_up_cancel_node.cpp)
+add_library(nav2_back_up_cancel_bt_node SHARED plugins/action/drive_on_heading_cancel_node.cpp)
```
SteveMacenski, line 69: We still want the backup -- so this should be in addition to not in place of

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/back_up_action.hpp
```diff
@@ -18,15 +18,15 @@
...
 
 namespace nav2_behavior_tree
 {
 
 /**
- * @brief A nav2_behavior_tree::BtActionNode class that wraps nav2_msgs::action::BackUp
+ * @brief A nav2_behavior_tree::BtActionNode class that wraps nav2_msgs::action::DriveOnHeading
  */
-class BackUpAction : public BtActionNode<nav2_msgs::action::BackUp>
+class BackUpAction : public BtActionNode<nav2_msgs::action::DriveOnHeading>
```
SteveMacenski, line 29: Back to backup, etc. All of the backup stuff should be reverted

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/drive_on_heading_action.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing basic ports along with node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return providedBasicPorts(
+      {
+        BT::InputPort<double>("dist_to_travel", -0.15, "Distance to travel"),
```
SteveMacenski, line 56: I think its reasonable to have these be positives (speed + dist)

nav2_system_tests/src/costmap_filters/speed_local_params.yaml
```diff
@@ -293,11 +293,11 @@ behavior_server:
     costmap_topic: local_costmap/costmap_raw
     footprint_topic: local_costmap/published_footprint
     cycle_frequency: 10.0
-    behavior_plugins: ["spin", "backup", "wait"]
+    behavior_plugins: ["spin", "drive_on_heading", "wait"]
     spin:
       plugin: "nav2_behaviors/Spin"
-    backup:
-      plugin: "nav2_behaviors/BackUp"
+    drive_on_heading:
```
SteveMacenski, line 299: Revert / add both

nav2_system_tests/src/costmap_filters/speed_local_params.yaml
```diff
@@ -62,7 +62,7 @@ bt_navigator:
     - nav2_compute_path_through_poses_action_bt_node
     - nav2_smooth_path_action_bt_node
     - nav2_follow_path_action_bt_node
-    - nav2_back_up_action_bt_node
+    - nav2_drive_on_heading_bt_node
```
SteveMacenski, line 65: Add both

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/drive_on_heading_action.hpp
```diff
@@ -0,0 +1,65 @@
...
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing basic ports along with node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return providedBasicPorts(
+      {
+        BT::InputPort<double>("dist_to_travel", 0.15, "Distance to travel"),
+        BT::InputPort<double>("speed", 0.025, "Speed at which to travel"),
+        BT::InputPort<double>("time_allowance", 10.0, "Allowed time for reversing")
```
SteveMacenski, line 58: "Allowed time for driving on heading"

nav2_behavior_tree/plugins/action/drive_on_heading_action.cpp
```diff
@@ -0,0 +1,62 @@
...
+  goal_.target.x = dist;
+  goal_.target.y = 0.0;
+  goal_.target.z = 0.0;
+  goal_.speed = speed;
+  goal_.time_allowance = rclcpp::Duration::from_seconds(time_allowance);
+}
+
+void DriveOnHeadingAction::on_tick()
+{
+  increment_recovery_count();
```
SteveMacenski, line 46: I do not believe this is a recovery, remove

nav2_behaviors/plugins/drive_on_heading.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+using namespace std::chrono_literals;
+
+namespace nav2_behaviors
+{
+
+DriveOnHeading::DriveOnHeading()
+: TimedBehavior<DriveOnHeadingAction>(),
+  feedback_(std::make_shared<DriveOnHeadingAction::Feedback>())
+{
```
SteveMacenski, line 31: I'm going to assume this entire file is exactly the same, not going to look through

SteveMacenski (commented):
Really good -- if I wasn't trying to reduce code duplication, I think we could even call it done right here.

Something I think worth trying, it might not work out due to complexities, but I think it should at least from a architect's 30,000 ft view:
- I looked at what is involved at deriving backup from the drive on heading BT node, and it literally turns into just as much code, just shuffled around. So no need to do that.
- I looked at what's involved at deriving backup behavior from the drive on heading behavior plugins, and I think you can reduce essentially all the code. Then you just need to export the plugin and set the different action definition. I think that might need to make `DriveOnHeading` to be itself templated with a default to `DriveOnHeading`'s message. But that should be a few just method definition  changes, in exchange, we have no duplicate code in the behavior plugins for duplicate behavior.

nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -0,0 +1,89 @@
...
+  double linear_acc_lim_;
+
+  geometry_msgs::msg::PoseStamped initial_pose_;
+  double command_x_;
+  double command_speed_;
+  rclcpp::Duration command_time_allowance_{0, 0};
+  rclcpp::Time end_time_;
+  double simulate_ahead_time_;
+
+  DriveOnHeadingAction::Feedback::SharedPtr feedback_;
```
SteveMacenski, line 84: ActionT::Feedback...

nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace nav2_behaviors
+{
+using DriveOnHeadingAction = nav2_msgs::action::DriveOnHeading;
+
+/**
+ * @class nav2_behaviors::DriveOnHeading
+ * @brief An action server Behavior for spinning in
+ */
+template<typename ActionT>
```
SteveMacenski, line 32: You can set `<typename ActionT = nav2_msgs::action::DriveOnHeading>` and then if no template is defined, it should default to that so you don't need to specify in the plugin registration / xml

SteveMacenski: but test, I might be wrong on the syntax

nav2_behaviors/plugins/drive_on_heading.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "nav2_util/node_utils.hpp"
+
+using namespace std::chrono_literals;
+
+namespace nav2_behaviors
+{
+
+template<typename ActionT>
+DriveOnHeading<ActionT>::DriveOnHeading()
+: TimedBehavior<DriveOnHeadingAction>(),
```
SteveMacenski, line 30: These should all be ActionT's

nav2_behaviors/plugins/drive_on_heading.cpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+    fetch_data = false;
+  }
+  return true;
+}
+
+}  // namespace nav2_behaviors
+
+#include "pluginlib/class_list_macros.hpp"
+PLUGINLIB_EXPORT_CLASS(nav2_behaviors::DriveOnHeading<nav2_behaviors::DriveOnHeadingAction>, nav2_core::Behavior)
```
SteveMacenski, line 171: Use the msg definition here the `nav2_behaviors` prefix is really confusing, that works because you `using`'ed the message type in the namespace, but that makes it seem like there's a separate action class in the package. That would take me an hour as a normal reader to finally decrypt.

SteveMacenski (commented):
Templating mostly looks good, but you missed a few spots. Anywhere where the message is defined should use ActionT so that when you base the backup class on it, you can pass it the backup message type and the rest should just work

jwallace42:
I am having a few issues loading the derived backup class. 

`[component_container_isolated-6] [FATAL] [1649964984.412816516] [behavior_server]: Failed to create behavior backup of type nav2_behaviors/BackUp. Exception: Failed to load library /home/josh/nav2_ws/install/nav2_behaviors/lib/libnav2_backup_behavior.so. Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in the library code, and that names are consistent between this macro and your XML. Error string: Could not load library dlopen error: /home/josh/nav2_ws/install/nav2_behaviors/lib/libnav2_backup_behavior.so: undefined symbol: _ZN14nav2_behaviors14DriveOnHeadingIN9nav2_msgs6action6BackUpEE5onRunESt10shared_ptrIKNS2_12BackUp_Goal_ISaIvEEEE, at /tmp/binarydeb/ros-rolling-rcutils-5.0.1/src/shared_library.c:99`

Not really sure how to go about debugging this

SteveMacenski:
Its hard to say without trying to test and see what is  going on. 

ZN14nav2_behaviors14DriveOnHeadingIN9nav2_msgs6action6BackUpEE5onRunESt10shared_ptrIKNS2_12BackUp_Goal_ISaIvEEEE, at /tmp/binarydeb/ros-rolling-rcutils-5.0.1/src/shared_library.c:99

But it looks like there's a mix up between them where something wasn't abstracted out or non-declared template types

Try adding 

`DriveOnHeading<msg backup>;`
`DriveOnHeading<msg drive on heading>;`

At the end of the CPP file, it might just need to be asked to generate the template branches for each type. That's common and we do that in Smac as well https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/a_star.cpp#L430-L433

jwallace42:
I got the template working. I still have to fix the nav2_system_tests!

jwallace42:
Fixed of test errors. The xml linter complains but the the structure is actually correct.

SteveMacenski:
> In XML < or & are strictly illegal to use as values. There are more which are not strictly illegal but are better avoided. Use an entity reference instead. So your XML file generates an error. The entity reference for < is &lt; and > is &gt;

I think that should help?

nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -80,9 +81,18 @@ class BackUp : public TimedBehavior<BackUpAction>
...
   double simulate_ahead_time_;
 
-  BackUpAction::Feedback::SharedPtr feedback_;
+  typename ActionT::Feedback::SharedPtr feedback_;
+};
+
+
+using BackUpAction = nav2_msgs::action::BackUp;
+
+class BackUp : public DriveOnHeading<nav2_msgs::action::BackUp>
```
SteveMacenski, line 90: Please use as another file for BackUp

nav2_behaviors/plugins/drive_on_heading.cpp
```diff
@@ -151,15 +163,42 @@ bool BackUp::isCollisionFree(
...
+    return Status::FAILED;
+  }
+
+  return Status::SUCCEEDED;
+}
+
 }  // namespace nav2_behaviors
 
 #include "pluginlib/class_list_macros.hpp"
+PLUGINLIB_EXPORT_CLASS(nav2_behaviors::DriveOnHeading<>, nav2_core::Behavior)
```
SteveMacenski, line 203: Ditto on `<>`, I will believe if you try, it should work, but I could be mistaken

nav2_behaviors/plugins/drive_on_heading.cpp
```diff
@@ -17,27 +17,31 @@
...
+DriveOnHeading<ActionT>::~DriveOnHeading()
 {
 }
 
-void BackUp::onConfigure()
+template<typename ActionT>
+void DriveOnHeading<ActionT>::onConfigure()
 {
-  auto node = node_.lock();
+  auto node = this->node_.lock();
```
SteveMacenski, line 44: Why the moves to `this`'s in a few places?

nav2_behaviors/plugins/drive_on_heading.cpp
```diff
@@ -151,15 +163,42 @@ bool BackUp::isCollisionFree(
...
-    if (!collision_checker_->isCollisionFree(pose2d, fetch_data)) {
+    if (!this->collision_checker_->isCollisionFree(pose2d, fetch_data)) {
       return false;
     }
     fetch_data = false;
   }
   return true;
 }
 
+Status BackUp::onRun(const std::shared_ptr<const BackUpAction::Goal> command)
```
SteveMacenski, line 174: Separate file please for backup

nav2_behaviors/behavior_plugin.xml
```diff
@@ -5,12 +5,18 @@
...
 
-	<library path="nav2_backup_behavior">
+	<library path="nav2_drive_on_heading_behavior">
 	  <class name="nav2_behaviors/BackUp" type="nav2_behaviors::BackUp" base_class_type="nav2_core::Behavior">
 	    <description></description>
 	  </class>
 	</library>
 
+	<library path="nav2_drive_on_heading_behavior">
+	  <class name="nav2_behaviors/DriveOnHeading" type="nav2_behaviors::DriveOnHeading<>" base_class_type="nav2_core::Behavior">
```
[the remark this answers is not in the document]
jwallace42: Yeah, C++ 17 removes this.

nav2_behaviors/plugins/drive_on_heading.cpp
jwallace42: https://isocpp.org/wiki/faq/templates
Look at the answer to the question
Why am I getting errors when my template-derived-class uses a member it inherits from its template-base-class?

SteveMacenski: OK - I do like that the best of the options :+1:

jwallace42:
@SteveMacenski should be gtg!

nav2_bringup/launch/bringup_launch.py
```diff
@@ -99,7 +99,7 @@ def generate_launch_description():
         description='Automatically startup the nav2 stack')
 
     declare_use_composition_cmd = DeclareLaunchArgument(
-        'use_composition', default_value='True',
+        'use_composition', default_value='False',
```
SteveMacenski, line 102: Revert

nav2_bringup/launch/tb3_simulation_launch.py
```diff
@@ -103,7 +103,7 @@ def generate_launch_description():
         description='Automatically startup the nav2 stack')
 
     declare_use_composition_cmd = DeclareLaunchArgument(
-        'use_composition', default_value='True',
+        'use_composition', default_value='False',
```
SteveMacenski, line 106: Revert

SteveMacenski (changes requested):
Just some changes on parameterization, but otherwise LGTM!

Did you test manually adding the drive on heading BT XML node in a BT and it worked as you expect?

nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -0,0 +1,217 @@
...
+        current_pose, *this->tf_, this->global_frame_, this->robot_base_frame_,
+        this->transform_tolerance_))
+    {
+      RCLCPP_ERROR(this->logger_, "Current robot pose is not available.");
+      return Status::FAILED;
+    }
+
+    double diff_x = initial_pose_.pose.position.x - current_pose.pose.position.x;
+    double diff_y = initial_pose_.pose.position.y - current_pose.pose.position.y;
+    double distance = sqrt(diff_x * diff_x + diff_y * diff_y);
```
SteveMacenski, line 114:
```suggestion
    double distance = hypot(diff_x * diff_x + diff_y * diff_y);
```

SteveMacenski (approved):

SteveMacenski:
@jwallace42 LGTM last is docs:
- Add to navigation plugins list in docs
- Add parameters page for this

nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -0,0 +1,217 @@
...
+        current_pose, *this->tf_, this->global_frame_, this->robot_base_frame_,
+        this->transform_tolerance_))
+    {
+      RCLCPP_ERROR(this->logger_, "Current robot pose is not available.");
+      return Status::FAILED;
+    }
+
+    double diff_x = initial_pose_.pose.position.x - current_pose.pose.position.x;
+    double diff_y = initial_pose_.pose.position.y - current_pose.pose.position.y;
+    double distance = hypot(diff_x * diff_x + diff_y * diff_y);
```
SteveMacenski, line 114:
```suggestion
    double distance = hypot(diff_x, diff_y);
```

SteveMacenski:
https://codecov.io/gh/ros-planning/navigation2/tree/8f722714e31601911b7bcd8515225a4ac13cb7bf/nav2_behaviors/plugins

A bit more test coverage on these would be good -- the cache breaking fixed the issue. This PR would drop the coverage % below 90% for the package

PR #2904, opened 2022-04-09
standalone assisted teleop

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2888 #2226 |
| Robotic platform tested on | Gazebo|

---

This pr adds a standalone node and core class for assisted teleop. 

Add documentation for a new node

---

- Use the core package to create a new TimedBehavior

#### For Maintainers: 


SteveMacenski:
By the way, CI failed because there is no teleop action in this PR

nav2_assisted_teleop/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+find_package(tf2 REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+
+nav2_package()
+
+include_directories(
+  include
+)
+
+add_executable(${PROJECT_NAME}
```
SteveMacenski, line 27: Missing component Cmake / registration in the library file so that this can be loaded / used as a component, not only a standalone process server. See other packages like nav2_planner for reference of how.

SteveMacenski: Then the library needs to be installed using macro like so: https://github.com/ros-planning/navigation2/blob/main/nav2_planner/CMakeLists.txt#L69-L73

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    input_twist_.linear.x = msg->linear.x;
+    input_twist_.linear.y = msg->linear.y;
+    input_twist_.angular.z = msg->linear.z;
+  }
+
+  void AssistedTeleopServer::joyCallback(const sensor_msgs::msg::Joy msg)
+  {
+    RCLCPP_INFO(logger_, "Hit joy callback");
+    joy_.buttons = msg.buttons;
```
SteveMacenski, line 309: Ditto, why not just store everything `joy_ = *msg` (with shared pointer). Also, `input_joy_` might be better so that its analog to the twist input

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+  geometry_msgs::msg::Twist AssistedTeleop::computeVelocity(geometry_msgs::msg::Twist& twist)
+  {
+    geometry_msgs::msg::PoseStamped current_pose;
+    if (!nav2_util::getCurrentPose(
+        current_pose, *tf_, global_frame_, robot_base_frame_,
+        transform_tolerance_))
+    {
+      RCLCPP_ERROR(logger_, "Failed to get current pose, setting velocity to zero");
+      return geometry_msgs::msg::Twist();
```
SteveMacenski, line 63: I think this should `throw` since that's a critical issue, and in the server, it should catch and 'pass' doing anything in that situation rather than returning an empty message

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    geometry_msgs::msg::Pose2D current_pose_2d;
+    current_pose_2d.x = current_pose.pose.position.x;
+    current_pose_2d.y = current_pose.pose.position.y;
+    current_pose_2d.theta = 0.0; //tf2::getYaw(current_pose.pose.orientation);
```
SteveMacenski, line 69: commented out?

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    const double linear_vel = std::fabs(std::hypot(twist.linear.x, twist.linear.y));
+    const double dt = 0.05 / linear_vel;
```
SteveMacenski, line 72: hardcoded timestep?

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+    {
+      projected_pose = projectPose(projected_pose, twist, dt);
+
+      if(!collision_checker_->isCollisionFree(projected_pose))
+      {
+        adjusted_twist.linear.x *= current_time / projection_time_;
+        adjusted_twist.linear.y *= current_time / projection_time_;
+        adjusted_twist.angular.z *= current_time / projection_time_;
+
+        RCLCPP_WARN_STREAM(logger_, "Collision approaching in " << current_time << ". Reducing velocity");
```
SteveMacenski, line 86: debug logging

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    geometry_msgs::msg::Pose2D projected_pose = current_pose_2d;
+    auto adjusted_twist = twist;
+    for(double current_time = 0; current_time < projection_time_; current_time+=dt)
+    {
+      projected_pose = projectPose(projected_pose, twist, dt);
+
+      if(!collision_checker_->isCollisionFree(projected_pose))
+      {
+        adjusted_twist.linear.x *= current_time / projection_time_;
```
SteveMacenski, line 82: Is `*=` the right thing? Recursively, this would be 0/max, then 0.05/max, then 0.1/max, etc. so each time its not just projecting by the current segment, but all prior as well.

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
SteveMacenski: Probably should be removed if we can't make any general rules about joys

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+  nav2_util::CallbackReturn
+  AssistedTeleopServer::on_configure(const rclcpp_lifecycle::State &)
+  {
+    RCLCPP_INFO(logger_, "Configuring assisted_teleop_server");
+
+    auto node = shared_from_this();
+
+    nav2_util::declare_parameter_if_not_declared(
+      this,
```
SteveMacenski, line 127: node

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    nav2_util::declare_parameter_if_not_declared(
+      node,
+      "output_vel_topic", rclcpp::ParameterValue(std::string("cmd_vel")));
+
+    nav2_util::declare_parameter_if_not_declared(
+      node,
+      "joystick_topic", rclcpp::ParameterValue(std::string("joystick_topic")));
+
+    declare_parameter(
```
SteveMacenski, line 138: declare_parameter_if_not_declared

jwallace42:
Yep, I forgot to add that back in.

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
```diff
@@ -0,0 +1,311 @@
...
+        vel_sub_.reset();
+        joy_sub_.reset();
+        result->total_elapsed_time = steady_clock_.now() -start_time;
+        action_server_->terminate_all(result);
+        return;
+      }
+
+      // preempt request
+      if (action_server_->is_preempt_requested()) {
+        vel_pub_->publish(geometry_msgs::msg::Twist());
```
SteveMacenski, line 280: Shouldn't stop if preempted

SteveMacenski (changes requested):
Overall not a bad go at it. 

Looking at the "core" of what this is doing though, I think its similar to what we're talking about in the collision monitor, one of the reasons I wanted to hold off on this. @AlexeyMerzlyakov is working on adding a collision monitor that checks based on raw sensor data and ensures that we're always N seconds from collision. 

I think this does a similar thing but instead based on the costmaps for general use rather than safety use. I think there's some code that could be shared between them but this looks reasonable.

I think though from looking at this, this belongs as a `behavior` plugin in the new behavior server you renamed rather than as a stand alone entity. This is using a TF buffer + costmap subscriber which are heavy and the primary reason for the behavior server to exist as a central location anyway.

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
SteveMacenski: on stop if canceled, preempt implies that it has a new command so it can be fluidly transitioned to the next goal

jwallace42:
> I think though from looking at this, this belongs as a `behavior` plugin in the new behavior server you renamed rather than as a stand alone entity.

I though we wanted the ability to run a standalone assisted teleop? #2226 The core library would be used to create the behavior.

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
jwallace42: I was missing the break:).

SteveMacenski:
That predates the idea of the recovery server being a generic behavior server. The generic behavior server meets all the intended needs, I believe.

jwallace42:
> That predates the idea of the recovery server being a generic behavior server. The generic behavior server meets all the intended needs, I believe.

So I guess the thing we would be missing is if it was implemented as a `TimedBehavior`. Ideally would want the ability to drive without worrying about a time out. 

So maybe we have a TimedBehavior for recovery calls and a normal behavior for just running?

nav2_assisted_teleop/src/nav2_assisted_teleop.cpp
SteveMacenski: Not sure I understand the comment, but I think we're on the same page :smile:

jwallace42:
> I think it could make sense to be a timed behavior, correct.
> 
> We should probably introduce some timeout parameter though (e.g. if no new cmd vel within N seconds, stop).
> 
> Also, maybe the action request should have a timeout (e.g. only do assisted teleop for 10 seconds, then exit) that way a caller can exit, otherwise, there's not really a clear way to end the assisted teleop session.
> 
> We also need a CancelAssistedTeleop BT XML node, similar to CancelControl / CancelBackup / etc. Its trivial, there's a base cancel node that should do all of it for you
> 
> That would give then users 2 ways of exiting: a timeout and a cancellation after a task is done. That seems good to me

Is there case where we would just want to drive around in assistied teleop and not have a time out?

SteveMacenski:
> Is there a case where we would just want to drive around in assisted teleop and not have a time out?

If the main task of the robot is assisted teleop hah (if you mean timeout for using the behavior; not timeout on velocity commands incoming). I think the default would be no-timeout, but it would be good to have the temporary option that returns after N seconds itself.

jwallace42:
Alright, I will add in the assistedTeleop timed behavior.

SteveMacenski:
There are still a few un-answered comments above that are not resolved, but I'll review again in the plugin form! Thanks, back from PTO sorry for the delay

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+  RCLCPP_INFO(logger_, "Hit input Vel callback");
+  input_twist_.linear.x = msg->linear.x;
+  input_twist_.linear.y = msg->linear.y;
+  input_twist_.angular.z = msg->angular.z;
+}
+
+void AssistedTeleop::joyCallback(const sensor_msgs::msg::Joy msg)
+{
+  RCLCPP_INFO(logger_, "Hit joy callback");
```
SteveMacenski, line 160: Remove logging

Also, should the joy stick be removed? I don't think there is a standard for how joystick layouts should be processed for velocity commands. Something else needs to translate joystick->velocity for us so we know what pressing "A" means

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+Status AssistedTeleop::onCycleUpdate()
+{
+  rclcpp::Duration time_remaining = end_time_ - steady_clock_.now();
+  if (time_remaining.seconds() < 0.0 && command_time_allowance_.seconds() > 0.0) {
+    stopRobot();
+    RCLCPP_WARN(
+      logger_,
+      "Exceeded time allowance before reaching the BackUp goal - Exiting BackUp");
```
SteveMacenski, line 92: Not backup :-)

SteveMacenski: I think this is a success ending point though, no? Its done, its succeeded in teleopping for that period of time

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (!nav2_util::getCurrentPose(
+      current_pose, *tf_, global_frame_, robot_base_frame_,
+      transform_tolerance_))
+  {
+    RCLCPP_ERROR(logger_, "Current robot pose is not available.");
+    return Status::FAILED;
+  }
+
+  // user states that teleop was successful
+  if (joy_.buttons.size() > 0 && joy_.buttons[0] > 0) {
```
SteveMacenski, line 106: If joy is removed, all this can go too

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+  geometry_msgs::msg::Twist cmd_vel;
+  cmd_vel = computeVelocity(input_twist_);
+  vel_pub_->publish(std::move(cmd_vel));
+
+  return Status::RUNNING;
+}
+
+geometry_msgs::msg::Twist AssistedTeleop::computeVelocity(geometry_msgs::msg::Twist & twist)
+{
+  // TODO(jwallace42): Waiting for safety node to be completed.
```
SteveMacenski, line 126: I know you're waiting on the projection math, though this should be easy, look at backup / spin for examples of how we do this. When we find a collision at a given timestep, you can back out the ratio to apply to the velocity command to scale back so always N time from collision with some basic kinematic formula. 

But @AlexeyMerzlyakov can you provide this for him to use? 

Also, I don't see a publisher for velocity to publish out the commands

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+geometry_msgs::msg::Pose2D AssistedTeleop::projectPose(
+  geometry_msgs::msg::Pose2D pose,
+  geometry_msgs::msg::Twist twist,
+  double projection_time)
+{
+  geometry_msgs::msg::Pose2D projected_pose;
+
+  projected_pose.x = projection_time * (
```
SteveMacenski, line 137: This is just the instanteous change for 1 timestep, you need to add to the initial pose to get the full change. This also only works if projection_time is a `dt` and not an absolute time.

SteveMacenski: Again, backup/spin math might help you a bit with projection kinematics -- visualizations can help alot too if you're uncertain

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+      this, std::placeholders::_1));
+
+  return Status::SUCCEEDED;
+}
+
+void AssistedTeleop::onActionCompletion()
+{
+  RCLCPP_INFO(logger_, "Action completed");
+  input_twist_ = geometry_msgs::msg::Twist();
+  vel_sub_.reset();
```
SteveMacenski, line 81: With the way this is structured, you can just keep the cmd_vel topic subscriber. We had the stop/starting before as part of a clever architectural pattern but we're not doing that anymore since we have the behavior timed behavior structure. Since the velocity callback is just storing the most recent command, this is OK to always just let it go.

Before, we had the main "work" in the velocity callback so stopping/starting it on was necessary as the start/stop of the behavior itself called by the action callbacks.

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+  command_time_allowance_ = command->time_allowance;
+  end_time_ = steady_clock_.now() + command_time_allowance_;
+
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+
+  vel_sub_ = node->create_subscription<geometry_msgs::msg::Twist>(
```
SteveMacenski, line 62: So this would go into the configure function

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: How will know if assisted teleop was completed successfully?

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: The timed behavior has the instance of the velocity publisher here [https://github.com/ros-planning/navigation2/blob/main/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp#L161](url).

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: Yeah, I had it all working but I figured it would be replaced by @AlexeyMerzlyakov.

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -70,16 +59,25 @@ Status AssistedTeleop::onRun(const std::shared_ptr<const AssistedTeleopAction::G
...
       this, std::placeholders::_1));
+}
+
+Status AssistedTeleop::onRun(const std::shared_ptr<const AssistedTeleopAction::Goal> command)
+{
+  command_time_allowance_ = command->time_allowance;
+  end_time_ = steady_clock_.now() + command_time_allowance_;
 
+  auto node = node_.lock();
+  if (!node) {
```
SteveMacenski, line 70: You don't need any of this node lock stuff

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: The timeout or it is canceled, I suppose. Either way, we can't make any assumptions about joy commands, there's no standardization on what any particular button means or context, so this doesn't really work. Someone's node that will read in joystick commands and converts into ROS service / action / velocity commands will have to interpret and do the right thing.

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: Okay. It is also in backup and spin.

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: So in the case of a recovery if the teleoperator drives to a new location successfully he/she should be able to give the feedback that assisted teleop action was successful. Over wise, the behavior tree won't know if the assisted teleop action was successful. (Maybe we would always assume it completes successfully? If you can't teleop of the bot out pretty much out of options anyway...)

I was planning on adding parameters for the "success" and "failure" buttons that would be checked. So, it could be up to the user to make that decision.

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: But you don't use it in this function, they use it in those functions, no?

jwallace42: Yep. Removed.

jwallace42:
@SteveMacenski Just finished this up

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: But, let me know what your thinking :)!

SteveMacenski:
There's still a number of comments above that haven't been done yet (maybe you didn't hit the "expand"?) None are huge things, but lots of logging / mathy / callbacky things.

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: > Maybe we would always assume it completes successfully?

I think that's a fair thing to do -- have it always return success in the BT node. A cancel just means to stop and isn't an indication of the successfulness (arguably, if cancelled, means it was successful and ready for navigation again)

nav2_msgs/action/AssistedTeleop.action
```diff
@@ -0,0 +1,8 @@
+#goal definition
+builtin_interfaces/Duration time_allowance
+---
+#result definition
+builtin_interfaces/Duration total_elapsed_time
+---
+#feedback
+builtin_interfaces/Duration time_spent_in_rc
```
SteveMacenski, line 8: `time_spent_in_rc`? how about just `current_teleop_duration` I'm not sure what `rc` means

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  // set up parameters
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    "projection_time", rclcpp::ParameterValue(1.0));
+
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    "input_vel_topic", rclcpp::ParameterValue(std::string("input_vel_topic")));
```
SteveMacenski, line 41: That doesn't seem like a great default :wink: If there isn't a natural default, we can just declare the type and not specify the value so that its a required parameter. But I think something like `teleop_cmd_vel` is natural

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    "joystick_topic", rclcpp::ParameterValue(std::string("joystick_topic")));
```
SteveMacenski, line 45: From other threads - remove all the joystick stuff

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+    twist.linear.y * cos(pose.theta));
+
+  projected_pose.theta += projection_time * twist.angular.z;
+
+  return projected_pose;
+}
+
+void AssistedTeleop::inputVelocityCallback(const geometry_msgs::msg::Twist::SharedPtr msg)
+{
+  input_twist_.linear.x = msg->linear.x;
```
SteveMacenski, line 173: Why not simply `input_twist_ = *msg;`? Down the line, having a timestamp might be helpful.

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+  if (time_remaining.seconds() < 0.0 && command_time_allowance_.seconds() > 0.0) {
+    stopRobot();
+    RCLCPP_WARN(
+      logger_,
+      "Exceeded time allowance before reaching the BackUp goal - Exiting BackUp");
+    return Status::FAILED;
+  }
+
+  geometry_msgs::msg::PoseStamped current_pose;
+  if (!nav2_util::getCurrentPose(
```
SteveMacenski, line 90: This function is not used `current_pose` is not used in this function.

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  geometry_msgs::msg::Twist cmd_vel;
+  cmd_vel = computeVelocity(input_twist_);
+  vel_pub_->publish(std::move(cmd_vel));
+
+  return Status::RUNNING;
+}
+
+geometry_msgs::msg::Twist AssistedTeleop::computeVelocity(geometry_msgs::msg::Twist & twist)
```
SteveMacenski, line 117: Merge this function into `onCycleUpdate`, after you remove the `joy` and the extra `getCurrentPose` calls, `onCycleUpdate` becomes incredibly short to basically just calling this function.

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+  {
+    RCLCPP_ERROR(logger_, "Current robot pose is not available.");
+    return geometry_msgs::msg::Twist();
+  }
+  geometry_msgs::msg::Pose2D projected_pose;
+  projected_pose.x = current_pose.pose.position.x;
+  projected_pose.y = current_pose.pose.position.y;
+  projected_pose.theta = tf2::getYaw(current_pose.pose.orientation);
+
+  const double dt = 0.01;
```
SteveMacenski, line 132: `simulation_time`, it should be a parameter, make it default to `0.1`

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+  geometry_msgs::msg::Twist scaled_twist = twist;
+  for (double time = dt; time < projection_time_; time+=dt)
```
SteveMacenski, line 134:
```suggestion
  for (double time = dt; time < projection_time_; time += dt)
```

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+  {
+    projected_pose = projectPose(projected_pose, twist, dt);
+
+    if (!collision_checker_->isCollisionFree(projected_pose))
+    {
+      RCLCPP_WARN(logger_, "Collision approaching in %.2f seconds", time);
+      double scale_factor = time / projection_time_;
+      scaled_twist.linear.x *= scale_factor;
```
SteveMacenski, line 142: Did you verify this math? I could believe its correct, but I wouldn't immediately assume based on my intuition that it is true that its just a linear ratio scale down. 

E.g. if you ran this projection stuff again on this new velocity, would it detect a collision at exactly the last iteration (or 1 past)?

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+      scaled_twist.linear.y *= scale_factor;
+      scaled_twist.angular.z *= scale_factor;
+      break;
+    }
+  }
+  return scaled_twist;
+}
+
+geometry_msgs::msg::Pose2D AssistedTeleop::projectPose(
+  geometry_msgs::msg::Pose2D pose,
```
SteveMacenski, line 152: If you made these references `&`, then the signature can be `void` and just update the input of `pose` with the new value implicitly.

Also, `twist` and `projection_time` should be `const`

SteveMacenski:
Make sure to check CI for errors

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: So, I thought a bit more about this. 

Even if we have the BT always return success we still don't have any feedback about when the action is done. Otherwise, we can only exit after a timeout. So I still think we need some way to allow the user to stop the action. 

I am not sure how we could implement this without adding another sub...

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: I verified in the simulator and on a piece of paper. 

If I ran this on a new velocity it would detect the collision on a new iteration because the step size is not dependent on the velocity. 

However, the previous method had the timestep as a func of vel. 

`const double dt = collision_checker_->getCostmapResolution() / linear_vel;`

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: Would this just be to remove the coping of geometry_msgs::msg::Pose2D? Or is there any other gains by passing in the output by reference?

I tend to follow https://www.fluentcpp.com/2016/11/22/make-your-functions-functional/ but really don't mind changing it :)

jwallace42:
So, responded to most of your comments. There are two main think left to finish 

- [ ] What do we want to do for a termination condition 
- [ ] Writing the nav2_system_test

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,203 @@
...
+  node->get_parameter("teleop_cmd_vel", teleop_cmd_vel);
+  node->get_parameter("joystick_topic", joystick_topic_);
+
+  vel_sub_ = node->create_subscription<geometry_msgs::msg::Twist>(
+    teleop_cmd_vel, rclcpp::SystemDefaultsQoS(),
+    std::bind(
+      &AssistedTeleop::teleopVelocityCallback,
+      this, std::placeholders::_1));
+
+  joy_sub_ = node->create_subscription<sensor_msgs::msg::Joy>(
```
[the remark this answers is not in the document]
SteveMacenski: We may need to have another sub for thing, but that cannot be a joystick, because again we have no standard mapping for what buttons mean what for any particular user.

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,203 @@
...
+{
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+
+  // set up parameters
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    "projection_time", rclcpp::ParameterValue(1.0));
```
SteveMacenski, line 37: These parameters need to be in the configuration guide where I mention you didn't include parameters from the server

nav2_behavior_tree/plugins/action/assisted_teleop_action.cpp
```diff
@@ -0,0 +1,57 @@
...
+  goal_.time_allowance = rclcpp::Duration::from_seconds(time_allowance);
+}
+
+void AssistedTeleopAction::on_tick()
+{
+  if (recovery_) {
+    increment_recovery_count();
+  }
+}
+
```
SteveMacenski, line 43: Also we never use `is_recovery_` with the result to set the status of the BT node based on if the failure due to timeout means that the BT node should return a failure or a success (if not a recovery; that its just done and always interpret completion as success).

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,203 @@
...
+  geometry_msgs::msg::PoseStamped current_pose;
+  if (!nav2_util::getCurrentPose(
+      current_pose, *tf_, global_frame_, robot_base_frame_,
+      transform_tolerance_))
+  {
+    RCLCPP_ERROR_STREAM(
+      logger_,
+      "Current robot pose is not available for " <<
+        behavior_name_.c_str());
+    return geometry_msgs::msg::Twist();
```
SteveMacenski, line 129: This should cause the action to return status failed

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,203 @@
...
+          logger_,
+          steady_clock_,
+          500,
+          behavior_name_.c_str() << " collided on first time step, setting velocity to zero");
+        scaled_twist.linear.x = 0.0f;
+        scaled_twist.linear.y = 0.0f;
+        scaled_twist.angular.z = 0.0f;
+        break;
+      } else {
+        RCLCPP_WARN_STREAM_THROTTLE(
```
SteveMacenski, line 154: Debug log

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,203 @@
...
+    node,
+    "projection_time", rclcpp::ParameterValue(1.0));
+
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    "simulation_time_step", rclcpp::ParameterValue(0.1));
+
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    "teleop_cmd_vel", rclcpp::ParameterValue(std::string("teleop_cmd_vel")));
```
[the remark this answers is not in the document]
SteveMacenski: You couldn't have known that, just the standard I set in the velocity smoother that all `cmd_vel` topics start with 'cmd_vel' and have their specificity at the end so they all appear together on ros topic lists.

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: I don't believe any of them are in there

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: I believe that's what the other behaviors do. If not, then you can disregard.

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: The point of this is to stop collisions so we don't need to blast them that this behavior is doing its job :-)

nav2_system_tests/CMakeLists.txt
```diff
@@ -54,18 +54,19 @@ if(BUILD_TESTING)
...
-  add_subdirectory(src/system)
-  add_subdirectory(src/system_failure)
-  add_subdirectory(src/updown)
-  add_subdirectory(src/waypoint_follower)
-  add_subdirectory(src/behaviors/spin)
-  add_subdirectory(src/behaviors/wait)
-  add_subdirectory(src/behaviors/backup)
-  add_subdirectory(src/behaviors/drive_on_heading)
-  add_subdirectory(src/costmap_filters)
+  # add_subdirectory(src/behavior_tree)
```
SteveMacenski, line 57: Commented out all other tests should be readded

nav2_system_tests/src/behaviors/assisted_teleop/assisted_teleop_behavior_tester.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+  while (!initial_pose_received_) {
+    RCLCPP_WARN(node_->get_logger(), "Initial pose not received");
+    sendInitialPose();
+    std::this_thread::sleep_for(100ms);
+    rclcpp::spin_some(node_);
+  }
+
+  // Wait for lifecycle_manager_navigation to activate behavior_server
+  std::this_thread::sleep_for(10s);
```
SteveMacenski, line 110: You've got alot of sleeps in this test... is there a way to reduce these times or use other utilities to be event based (e.g. check the lifecycle status of it)? All of these waits _kill_ us in our CI times as they add up

nav2_system_tests/src/behaviors/assisted_teleop/assisted_teleop_behavior_tester.cpp
```diff
@@ -0,0 +1,280 @@
...
+    cmd_vel.linear.x = lin_vel;
+    cmd_vel.angular.z = ang_vel;
+    cmd_vel_pub_->publish(cmd_vel);
+
+    if (counter_ > 1) {
+      break;
+    }
+
+    auto current_time = std::chrono::system_clock::now();
+    if (current_time - start_time > 25s) {
```
SteveMacenski, line 183: Is that how long it takesish?

nav2_behaviors/plugins/assisted_teleop.hpp
```diff
@@ -0,0 +1,106 @@
...
+   * @param msg empty message
+   */
+  void preemptTeleopCallback(const std_msgs::msg::Empty::SharedPtr msg);
+
+  AssistedTeleopAction::Feedback::SharedPtr feedback_;
+
+  // parameters
+  double projection_time_;
+  double simulation_time_step_;
+  std::string joystick_topic_;
```
SteveMacenski, line 92: joystick topic?

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+void AssistedTeleop::teleopVelocityCallback(const geometry_msgs::msg::Twist::SharedPtr msg)
+{
+  teleop_twist_ = *msg;
+}
+
+void AssistedTeleop::preemptTeleopCallback(const std_msgs::msg::Empty::SharedPtr)
+{
+  RCLCPP_INFO(logger_, "HIT PREEMPT CALLBACK");
```
SteveMacenski, line 179: Logging :wink:

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -0,0 +1,186 @@
...
+    "preempt_teleop", rclcpp::SystemDefaultsQoS(),
+    std::bind(
+      &AssistedTeleop::preemptTeleopCallback,
+      this, std::placeholders::_1));
+}
+
+Status AssistedTeleop::onRun(const std::shared_ptr<const AssistedTeleopAction::Goal> command)
+{
+  command_time_allowance_ = command->time_allowance;
+  end_time_ = steady_clock_.now() + command_time_allowance_;
```
SteveMacenski, line 69: Set `preempt_teleop_` to `false` so old stale commands don't take over here

SteveMacenski (changes requested):

nav2_behaviors/plugins/assisted_teleop.cpp
jwallace42: I reset it here https://github.com/jwallace42/navigation2/blob/standalone_assisted_teleop/nav2_behaviors/plugins/assisted_teleop.cpp#L76.

nav2_system_tests/src/behaviors/assisted_teleop/assisted_teleop_behavior_tester.cpp
jwallace42: I can reduce this. We currently have three tests that take place. 

1) collide backward, just lin_vel
2) collide forward, neg lin vel
3) collide while driving will ang and lin vel.

The collide forward takes a while because it drives across the world.

nav2_behaviors/plugins/assisted_teleop.cpp
SteveMacenski: https://github.com/ros-planning/navigation2/blob/da8c297ea54925b042df2ecd1bd9f2a16e018582/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp#L197-L203

Except if `onRun` fails, on action complete isn't called

nav2_system_tests/src/behaviors/assisted_teleop/assisted_teleop_behavior_tester.cpp
SteveMacenski: OK - well try to minimize these or at least time how long it should practically take and set the timeouts to if that some leniency. If we set these to really high values for something that is relatively quick, then that just adds to CI time

nav2_system_tests/src/behaviors/assisted_teleop/assisted_teleop_behavior_tester.hpp
```diff
@@ -46,7 +46,6 @@ class AssistedTeleopBehaviorTester
 {
 public:
   using AssistedTeleop = nav2_msgs::action::AssistedTeleop;
-  // using GoalHandleBackup = rclcpp_action::ClientGoalHandle<BackUp>;
```
SteveMacenski: ? backup

jwallace42: Copied over from a old file and I forgot to remove it earlier :(

nav2_behavior_tree/plugins/action/assisted_teleop_action.cpp
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2022 Joshua Wallace
```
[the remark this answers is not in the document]
jwallace42: Added!

SteveMacenski (approved):

SteveMacenski:
Check CI, there are some errors, particularly `libnav2_assisted_teleop_cancel_bt_nodenav2_wait_cancel_bt_node.so` (somewhere missing a comma)

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -76,6 +77,8 @@ BtNavigator::BtNavigator(const rclcpp::NodeOptions & options)
     "nav2_path_longer_on_approach_bt_node",
     "nav2_wait_cancel_bt_node",
     "nav2_spin_cancel_bt_node",
+    "nav2_assisted_teleop_cancel_bt_node"
```
SteveMacenski, line 80: Comma missing

jwallace42:
Yeah, 

I reduced the sleeps as much as a could a while back. I still need to add a tutorial for the feature as you mentioned:). If the check pass I will take a look at that later in the week.

jwallace42:
Yeah, 

I reduced the sleeps as much as a could a while back. I still need to add a tutorial for the feature as you mentioned:). If the check pass I will take a look at that later in the week.

SteveMacenski:
> Could not load library: libnav2_assisted_teleop_cancel_bt_nodenav2_wait_cancel_bt_node.so: cannot open shared object file

SteveMacenski:
OK CI looks good other than a linting error (which I just fixed in main already)

jwallace42:
I found the docs we merged a while back. https://github.com/ros-planning/navigation.ros.org/pull/318.

PR #2908, opened 2022-04-11
Delete param dumper

SteveMacenski:
ROS 2 CLI has this available now and its not been integrated as part of any workflow within Nav2. It was created in the early days before we had a good accounting of parameters and there was no CLI that would do this for us. These are no longer true so we can deprecate this bit of the code


nav2_bringup/params/nav2_params.yaml
```diff
@@ -135,46 +135,45 @@ controller_server:
       yaw_goal_tolerance: 0.25
     # DWB parameters
     FollowPath:
-      plugin: "dwb_core::DWBLocalPlanner"
```
vinnnyr, line 138: @SteveMacenski  I think I noticed some unrelated code got added in inadvertently :sweat_smile:

PR #2918, opened 2022-04-15
Allow usage of std::string in searchAndGetParam()

### LINKED ISSUE #2916: DWB parameter helper doesn't work with strings
https://github.com/ros-planning/navigation2/blob/21a146a97c0d400eb9bdaeaf31bb9dfb536cf00b/nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/parameters.hpp#L75

Writing new DWB critics and noticed that 'searchAndGetParam<>()' compiles for strings but then fails during execution, since `value` is 0-initialized and there is a [run time error](https://stackoverflow.com/questions/33706112/setting-stdstring-to-0-during-definition-vs-setting-stdstring-to-0).

A solution would be to remove the initialization of the variable, since it is overwritten a few lines later.

roncapat:
Fixes #2916.

Moreover, I double-checked the snipped that was included in the TODO right above, and realized that was referring to old ROS API. For me, it had no more meaning in this context, but maybe @crdelsey can tell if it's right to do this clean up or not.

### DIFF
--- a/nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/parameters.hpp
+++ b/nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/parameters.hpp
@@ -64,15 +64,6 @@ param_t searchAndGetParam(
   const nav2_util::LifecycleNode::SharedPtr & nh, const std::string & param_name,
   const param_t & default_value)
 {
-  // TODO(crdelsey): Handle searchParam
-  // std::string resolved_name;
-  // if (nh->searchParam(param_name, resolved_name))
-  // {
-  //   param_t value = 0;
-  //   nh->param(resolved_name, value, default_value);
-  //   return value;
-  // }
-  param_t value = 0;
   nav2_util::declare_parameter_if_not_declared(
     nh, param_name,
     rclcpp::ParameterValue(default_value));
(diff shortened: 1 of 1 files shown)


nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/parameters.hpp
```diff
@@ -64,15 +64,6 @@ param_t searchAndGetParam(
...
 {
-  // TODO(crdelsey): Handle searchParam
-  // std::string resolved_name;
-  // if (nh->searchParam(param_name, resolved_name))
-  // {
-  //   param_t value = 0;
-  //   nh->param(resolved_name, value, default_value);
-  //   return value;
-  // }
-  param_t value = 0;
```
SteveMacenski, line 75: You deleted the variable? Doesn't that get referenced later?

SteveMacenski (changes requested):

nav2_dwb_controller/nav_2d_utils/include/nav_2d_utils/parameters.hpp
roncapat: Of course, you're right... :) fixing, sorry for that

SteveMacenski (approved):

PR #2922, opened 2022-04-21
Obstacle layer dynamic parameters

AadityaPatanjali:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | gazebo turtlebot simulation |

---

Added dynamic parameter update from this list: https://github.com/ros-planning/navigation/blob/noetic-devel/costmap_2d/cfg/ObstaclePlugin.cfg (enabled, footprint_clearing_enabled, max_obstacle_height, and combination_method)

---

#### For Maintainers: 


jwallace42 (approved):

jwallace42:
LGTM I would wait for @SteveMacenski to take a look and for wait for checks to pass.

SteveMacenski (approved):

SteveMacenski:
@indraneelpatil what are your thoughts on this? You approve? @AadityaPatanjali, @indraneelpatil did the initial one for Obstacle layer, but this looks good to me!

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -272,6 +279,38 @@ void ObstacleLayer::onInitialize()
   }
 }
 
+rcl_interfaces::msg::SetParametersResult
+ObstacleLayer::dynamicParametersCallback(
+  std::vector<rclcpp::Parameter> parameters)
+{
+  std::lock_guard<Costmap2D::mutex_t> guard(*getMutex());
```
SteveMacenski, line 286: I think you need to add `   std::lock_guard<Costmap2D::mutex_t> guard(*getMutex());` at the start of update bounds

SteveMacenski (changes requested):

nav2_costmap_2d/plugins/obstacle_layer.cpp
AadityaPatanjali: I've moved the mutex locking from updateFootprint to updateBounds so that lock_guard isn't called twice.

SteveMacenski (approved):

doisyg:
Hum, what happens here if the update frequency of the layer is very small or equal to zero ?  `combination_method` probably won't have an immediate effect after being set. Same for `max_obstacle_height`, `enabled` and `footprint_clearing_enabled`
I believe costs, footprint and bounds have to be updated in case a parameter impacting them has been changed

SteveMacenski:
Why? It wouldn't take effect until the next time there's a costmap layer update anyway so why preemptively do it? 

Enabled / footprint_clearing_enabled /  combination_method are only used when there's an update coming. Even Max Obstacle Height is only used in update bounds too https://github.com/ros-planning/navigation2/blob/da602f8a19fa6aaa7078e56ff0c3c83517feb533/nav2_costmap_2d/plugins/obstacle_layer.cpp#L452.

You would be correct about `max_obstacle_height` though if we were talking about the sensor's data stream `max_obstacle_height` since the observation buffer will do some work. However, only the global layer-level `max_obstacle_height` is reconfigurable. So far, none of the individual sensor stream parameterizations are dynamic here or in the voxel layer

Issue #2923, opened 2022-04-24
BasicNavigator doesn't clean up ActionClients


m2-farzan:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Galactic
- Version or commit hash:
  - 1.0.8-1focal.20220203.051427
- DDS implementation:
  - Fast-RTPS 

#### Steps to reproduce issue

```
#!/bin/python3

import rclpy
from nav2_simple_commander.robot_navigator import BasicNavigator

rclpy.init()
nav = BasicNavigator()
nav.destroy_node()
```

#### Expected behavior
Clean exit

#### Actual behavior
Exits with 0 code but prints the following warning:
```
Exception ignored in: <function ActionClient.__del__ at 0x7f310cc98f70>
Traceback (most recent call last):
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/action/client.py", line 597, in __del__
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/action/client.py", line 590, in destroy
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/handle.py", line 122, in __enter__
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/handle.py", line 104, in _get_capsule
rclpy._rclpy_pybind11.InvalidHandle: Tried to use a handle that has been destroyed.
Exception ignored in: <function ActionClient.__del__ at 0x7f310cc98f70>
Traceback (most recent call last):
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/action/client.py", line 597, in __del__
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/action/client.py", line 590, in destroy
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/handle.py", line 122, in __enter__
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/handle.py", line 104, in _get_capsule
rclpy._rclpy_pybind11.InvalidHandle: Tried to use a handle that has been destroyed.
Exception ignored in: <function ActionClient.__del__ at 0x7f310cc98f70>
Traceback (most recent call last):
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/action/client.py", line 597, in __del__
  File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/action/client.py", line 590, in destroy
[...]

PR #2924, opened 2022-04-24
Clean up action clients in nav2_simple_commander

m2-farzan:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2923 |

---

- Removes an unwanted warning
- Enables controlled release of resources

---

#### For Maintainers: 

### DIFF
--- a/nav2_simple_commander/README.md
+++ b/nav2_simple_commander/README.md
@@ -39,6 +39,7 @@ The methods provided by the basic navigator are shown below, with inputs and exp
 | waitUntilNav2Active(navigator='bt_navigator, localizer='amcl') | Blocks until Nav2 is completely online and lifecycle nodes are in the active state. To be used in conjunction with autostart or external lifecycle bringup. Custom navigator and localizer nodes can be specified  |
 | lifecycleStartup()                | Sends a request to all lifecycle management servers to bring them into the active state, to be used if autostart is `false` and you want this program to control Nav2's lifecycle. |
 | lifecycleShutdown()               | Sends a request to all lifecycle management servers to shut them down.     |
+| destroyNode()                     | Releases the resources used by the object.                                 |
 
 A general template for building applications is as follows:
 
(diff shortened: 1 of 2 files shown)


jwallace42 (changes requested):
Looks good! I think you still need to call `destroy_node` in the python files using the robot navigator.

SteveMacenski:
I'm not too much of a python ros2 developer, so is `destroy_node` called automatically by the `Node` class to need to be called specifically that with snake_case? Or is this a user's need to call this, in that case it should follow our naming conventions here as camcelCase & have its addition documented on the API page of navigation.ros.org / in this package's readme.

m2-farzan:
@jwallace42 `destroy_node` is only required if we want to create a basic_navigator node right after we are done with another one (e.g. in unit tests). This is not the case in our python demos, so I don't think we need to modify them.

m2-farzan:
@SteveMacenski `destroy_node` is inherited from `Node` base class. I couldn't find in docs if it's called automatically or not but [this little comment](https://github.com/ros2/examples/blob/9ff80f5dcbeba516adff144f974ec92f19266f10/rclpy/topics/minimal_subscriber/examples_rclpy_minimal_subscriber/subscriber_member_function.py#L44) suggests it might be. Either way, it's common practice to use `destroy_node` to clean up after a node. So I found it best to add a camelCase version while keeping the snake_case version there for compatibility & consistency (i.e. to avoid `destroy_node` and `destroyNode` from behaving differently).

In the long term the issue should be addressed in the upstream: https://github.com/ros2/rclpy/issues/524

SteveMacenski:
I'm fine with that -- I prefer defensive programming

SteveMacenski (approved):

SteveMacenski:
Thanks for the contribution!

Issue #2931, opened 2022-04-29
AMCL change map / first_map_only = false


mrmara:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Foxy

#### Steps to reproduce issue
set first_map_only to false
load  a newmap with map server
Ensure that amcl accepted and loaded the new map (log to appear: Received a %d X %d map @ %.3f m/pix) and that it is receiving laserscan data

#### Expected behavior
AMCL should load the new map and start localizing in it
#### Actual behavior
AMCL crashes
#### Additional information
Backtrace:
```
Thread 13 "amcl" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff0920700 (LWP 28300)]
0x00007ffff7413d8c in map_calc_range ()
   from /home/mrmara/ws//install/nav2_amcl/lib/libmap_lib.so
(gdb) backtrace
#0  0x00007ffff7413d8c in map_calc_range ()
   from /home/mrmara/ws//install/nav2_amcl/lib/libmap_lib.so
#1  0x00007ffff74d9ff7 in nav2_amcl::BeamModel::sensorFunctionWithRejection(nav2_amcl::LaserData*, _pf_sample_set_t*, bool*) ()
   from /home/mrmara/ws//install/nav2_amcl/lib/libsensors_lib.so
#2  0x00007ffff740a95d in pf_update_sensor ()
   from /home/mrmara/ws//install/nav2_amcl/lib/libpf_lib.so
#3  0x00007ffff74da465 in nav2_amcl::BeamModel::sensorUpdate(_pf_t*, nav2_amcl::LaserData*, bool, bool*) ()
   from /home/mrmara/ws//install/nav2_amcl/lib/libsensors_lib.so
#4  0x00007ffff7d16889 in nav2_amcl::AmclNode::updateFilter(int const&, std::shared_ptr<sensor_msgs::msg::LaserScan_<std::allocator<void> > const> const&, pf_vector_t const&, pf_vector_t const&) ()
   from /home/mrmara/ws//install/nav2_amcl/lib/libamcl_core.so
#5  0x00007ffff7d18a47 in nav2_amcl::AmclNode::laserReceived(std::shared_ptr<sensor_msgs::msg::LaserScan_<std::allocator<void> > const>) ()
--Type <RET> for more, q to quit, c to continue without paging--ibamcl_core.so
[...]

PR #2932, opened 2022-05-03
Dynamic reconfigure for AMCL

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #956  |
| Robotic platform tested on | TB3-Simulation |

I'm really not sure, if I'm doing justice to the Mutex.. @SteveMacenski 
I gave it a test and seems to work fine though... I will take a look once again, before I make this PR to be a "ready for review"

### DIFF
--- a/nav2_amcl/src/amcl_node.cpp
+++ b/nav2_amcl/src/amcl_node.cpp
@@ -1142,78 +1142,94 @@ AmclNode::dynamicParametersCallback(
     if (param_type == ParameterType::PARAMETER_DOUBLE) {
       if (param_name == "alpha1") {
         alpha1_ = parameter.as_double();
-    } else if (param_name == "alpha2") {
+      } else if (param_name == "alpha2") {
         alpha2_ = parameter.as_double();
-    } else if (param_name == "alpha3") {
+      } else if (param_name == "alpha3") {
         alpha3_ = parameter.as_double();
-    } else if (param_name == "alpha4") {
+      } else if (param_name == "alpha4") {
         alpha4_ = parameter.as_double();
-    } else if (param_name == "alpha5") {
+      } else if (param_name == "alpha5") {
         alpha5_ = parameter.as_double();
-    } else if (param_name == "beam_skip_distance") {
+      } else if (param_name == "beam_skip_distance") {
         beam_skip_distance_ = parameter.as_double();
-    } else if (param_name == "beam_skip_error_threshold") {
+      } else if (param_name == "beam_skip_error_threshold") {
         beam_skip_error_threshold_ = parameter.as_double();
-    } else if (param_name == "beam_skip_threshold") {
+      } else if (param_name == "beam_skip_threshold") {
         beam_skip_threshold_ = parameter.as_double();
-    } else if (param_name == "lambda_short") {
+      } else if (param_name == "lambda_short") {
         lambda_short_ = parameter.as_double();
-    } else if (param_name == "laser_likelihood_max_dist") {
+      } else if (param_name == "laser_likelihood_max_dist") {
         laser_likelihood_max_dist_ = parameter.as_double();
-    } else if (param_name == "laser_max_range") {
+      } else if (param_name == "laser_max_range") {
         laser_max_range_ = parameter.as_double();
-    } else if (param_name == "laser_min_range") {
+      } else if (param_name == "laser_min_range") {
         laser_min_range_ = parameter.as_double();
-    } else if (param_name == "pf_err") {
+      } else if (param_name == "pf_err") {
         pf_err_ = parameter.as_double();
-    } else if (param_name == "pf_z") {
+      } else if (param_name == "pf_z") {
         pf_z_ = parameter.as_double();
-    } else if (param_name == "recovery_alpha_fast") {
+      } else if (param_name == "recovery_alpha_fast") {
         alpha_fast_ = parameter.as_double();
-    } else if (param_name == "recovery_alpha_slow") {
+      } else if (param_name == "recovery_alpha_slow") {
         alpha_slow_ = parameter.as_double();
-    } else if (param_name == "save_pose_rate") {
+      } else if (param_name == "save_pose_rate") {
         save_pose_rate = parameter.as_double();
         save_pose_period_ = tf2::durationFromSec(1.0 / save_pose_rate);
-    } else if (param_name == "sigma_hit") {
+      } else if (param_name == "sigma_hit") {
         sigma_hit_ = parameter.as_double();
-    } else if (param_name == "transform_tolerance") {
+      } else if (param_name == "transform_tolerance") {
         tmp_tol = parameter.as_double();
         transform_tolerance_ = tf2::durationFromSec(tmp_tol);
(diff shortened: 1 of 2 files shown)


nav2_amcl/src/amcl_node.cpp
```diff
@@ -1113,6 +1124,122 @@ AmclNode::initParameters()
...
 
+/**
+  * @brief Callback executed when a parameter change is detected
+  * @param event ParameterEvent message
+  */
+rcl_interfaces::msg::SetParametersResult
+AmclNode::dynamicParametersCallback(
+  std::vector<rclcpp::Parameter> parameters)
+{
+  std::lock_guard<std::mutex> lock(pf_mutex_);
```
SteveMacenski, line 1135: I think there's a purpose-driven configuration mutex.

Also in reviewing this I noticed that that configuration mutex was removed from a number of functions. Please look over the ROS 1 AMCL and readd the mutex to all the appropriate functions to make sure we don't have async issues.

SteveMacenski (changes requested):
In terms of the actual function, it looks good except where noted.

However, AMCL does alot of objects that use these parameters that will need to be reinitialized / reset / new values given to it. Simply setting the variable in the callback doesn't make it used in the actual node, most, if not all, of these parameters are passed to other objects for use in the node which aren't going to be affected by the current PR

nav2_amcl/src/amcl_node.cpp
padhupradheep: #2931 takes care of it..

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1113,6 +1124,195 @@ AmclNode::initParameters()
...
+    initOdometry();
+  }
+
+  // Re-initialize the lasers and it's filters
+  if (reinit_laser) {
+    lasers_.clear();
+    lasers_update_.clear();
+    frame_to_laser_.clear();
+
+    laser_scan_filter_ = std::make_unique<tf2_ros::MessageFilter<sensor_msgs::msg::LaserScan>>(
```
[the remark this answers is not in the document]
padhupradheep: > why not call initMessageFilters like you do with the others?

I did that initially and somehow the node stopped responding.. I'll look into this matter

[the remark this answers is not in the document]
padhupradheep: The issue that I had was seen in the terminal: 

`CRITICAL FAILURE: SERVER amcl IS DOWN after not receiving a heartbeat for 4000 ms. Shutting down related nodes.`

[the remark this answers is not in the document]
padhupradheep: On further investigation, seems that the issue happens, when the laser scanners subscribers are reset...

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1113,6 +1124,203 @@ AmclNode::initParameters()
...
+    } else if (param_type == ParameterType::PARAMETER_BOOL) {
+      if (param_name == "do_beamskip") {
+        do_beamskip_ = parameter.as_bool();
+        reinit_laser = true;
+      } else if (param_name == "tf_broadcast") {
+        tf_broadcast_ = parameter.as_bool();
+      } else if (param_name == "set_initial_pose") {
+        set_initial_pose_ = parameter.as_bool();
+      } else if (param_name == "always_reset_initial_pose") {
+        always_reset_initial_pose_ = parameter.as_bool();
```
doisyg, line 1255: I don't think this is needed

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1113,6 +1124,203 @@ AmclNode::initParameters()
...
+  }
+
+  // Checking if the minimum particles is greater than max_particles_
+  if (min_particles_ > max_particles_) {
+    RCLCPP_WARN(
+      this->get_logger(),
+      "You've set min_particles to be greater than max particles,"
+      "this isn't allowed so they'll be set to be equal.");
+    max_particles_ = min_particles_;
+  }
```
doisyg, line 1282: I am not sure what should be done here in case of an invalid min/max particule set request:
- Set to equal and return success = true
- Set to equal and return success = false
- Not changing the parameter and return success = false ?

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1113,6 +1124,203 @@ AmclNode::initParameters()
...
+    laser_scan_connection_ = laser_scan_filter_->registerCallback(
+      std::bind(
+        &AmclNode::laserReceived,
+        this, std::placeholders::_1));
+  }
+
+  // Re-initialize the map
+  if (reinit_map) {
+    map_sub_.reset();
+    initPubSub();
```
doisyg, line 1317: This is resetting not only the map subscriber but also particle_cloud_pub_, pose_pub_ and initial_pose_sub_ which I believe is not needed

doisyg (commented):
Overall look good, left 2 potential improvements and a question for debate

nav2_amcl/src/amcl_node.cpp
SteveMacenski: Not change and return failure

SteveMacenski:
Toggling to get CI to turn over

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1113,6 +1124,205 @@ AmclNode::initParameters()
...
+    }
+  }
+
+  // Checking if the minimum particles is greater than max_particles_
+  if (min_particles_ > max_particles_) {
+    RCLCPP_ERROR(
+      this->get_logger(),
+      "You've set min_particles to be greater than max particles,"
+      " this isn't allowed.");
+    max_particles_ = min_particles_;
```
SteveMacenski, line 1279: But you still have them set? you should be rejecting the update and using the original parameters for both of these. You may need to store a local copy of these at the start of the function of the particle count changes

nav2_amcl/src/amcl_node.cpp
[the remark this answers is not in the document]
SteveMacenski: OK, its also OK if we set some parameters as non-changable. We could make `scan_topic_` a non-dynamic parameter, although it would be a useful one. 

Though, I don't understand why that's happening. Is that reset process taking on the order of seconds? Did you reset these variables beforehand or just calling init (or try both)? I don't think that'll change anything, but it would be diagnostic if it did.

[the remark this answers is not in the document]
padhupradheep: > Did you reset these variables beforehand or just calling init (or try both)? I don't think that'll change anything,

Yes, you are right. I tried doing both, but somehow both of them yields the same..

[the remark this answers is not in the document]
padhupradheep: Oh, I didn't disconnect the `laser_scan_connection_`. Damn..! Now, I guess it is fine.. let me do a couple of more testing, before the commit..

SteveMacenski (approved):
@doisyg what do you think?

PR #2933, opened 2022-05-04
Added mutex to prevent SEGFAULT on map change in AMCL

mrmara:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (#2931 ) |
| Primary OS tested on | Ubuntu 20.04.3 LTS |

---

Fixed Segmentation Fault happening in map_calc_range() if new laserscans were processed during a map change event.
This was done by adding ```configuration_mutex_``` lock on laserscan callback. Mutex lock has been added also to other methods accordingly to ROS Noetic implementation

none

---

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -132,7 +132,7 @@ protected:
   bool first_map_only_{true};
   std::atomic<bool> first_map_received_{false};
   amcl_hyp_t * initial_pose_hyp_;
-  std::recursive_mutex configuration_mutex_;
+  std::recursive_mutex mutex_;
   rclcpp::Subscription<nav_msgs::msg::OccupancyGrid>::ConstSharedPtr map_sub_;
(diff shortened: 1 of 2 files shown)


SteveMacenski (changes requested):
I think we can rename one of these mutexes to just the `mutex_` and that's used in all of the places that these 2 are used. There's no reason to have both locked in all situations. 

Also, if we already had `pf_mutex_` locked in all these places, how did adding the configuration mutex fix your issue?

mrmara:
> I think we can rename one of these mutexes to just the `mutex_` and that's used in all of the places that these 2 are used. There's no reason to have both locked in all situations.

Yes, you are right! I just pushed the modifications.
Anyway, the automatic tests keep failing but I do not understand what is wrong, is it normal?

padhupradheep:
> Anyway, the automatic tests keep failing but I do not understand what is wrong, is it normal?

In short, you do not have to worry about it.. :-)

SteveMacenski (approved):

PR #2940, opened 2022-05-06
Galactic sync 6

SteveMacenski:
Fixing costmap region clearing logic

FixingTF lookups in controller server

Adding various improvements and new APIs to the Simple Commander Python API

Remove resizing of costmap on incremental update

Removing unused or unncessary methods and dependencies

Adding Theta* to default install list

Fix crashing on LC transition in BT navigator

Fixes crashing issue in AMCL when dynamically setting new maps


lopsided98:
I think https://github.com/ros-planning/navigation2/pull/2886 should have been included here, because otherwise the build fails for me.

Issue #2917, opened 2022-04-15
Nav2 lifecycle manager won't startup nodes


AlexeyMerzlyakov:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu focal 20.04.4
- ROS2 Version:
  - ROS2 rolling built with sources from Apr 13 2022
- Version or commit hash:
  - f1d25e98
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
Run `nav2_lifecycle_manager` with any lifecylce node e.g.:
```
ros2 launch nav2_bringup bringup_launch.py map:=${HOME}/nav2_ws/src/navigation2/nav2_bringup/maps/turtlebot3_world.yaml | grep -e Configuring -e Activating
```

#### Expected behavior
Lifecycle nodes are going to "Configured" and "Active" state:
```
$ ros2 launch nav2_bringup bringup_launch.py map:=${HOME}/nav2_ws/src/navigation2/nav2_bringup/maps/turtlebot3_world.yaml | grep -e Configuring -e Activating

[component_container_isolated-1] [INFO] [1650040180.873389080] [lifecycle_manager_localization]: Configuring map_server
[component_container_isolated-1] [INFO] [1650040180.874164522] [map_server]: Configuring
[component_container_isolated-1] [INFO] [1650040180.919901945] [lifecycle_manager_localization]: Configuring amcl
[component_container_isolated-1] [INFO] [1650040180.920120541] [amcl]: Configuring
[component_container_isolated-1] [INFO] [1650040180.974971983] [lifecycle_manager_localization]: Activating map_server
[component_container_isolated-1] [INFO] [1650040180.975233144] [map_server]: Activating
[component_container_isolated-1] [INFO] [1650040181.081511733] [lifecycle_manager_localization]: Activating amcl
[component_container_isolated-1] [INFO] [1650040181.081762152] [amcl]: Activating
[component_container_isolated-1] [INFO] [1650040181.249707477] [lifecycle_manager_navigation]: Configuring controller_server
[component_container_isolated-1] [INFO] [1650040181.249960165] [controller_server]: Configuring controller interface
[component_container_isolated-1] [INFO] [1650040181.251433253] [local_costmap.local_costmap]: Configuring
[component_container_isolated-1] [INFO] [1650040181.370370000] [lifecycle_manager_navigation]: Configuring smoother_server
[...]

PR #2960, opened 2022-05-20
Fixed lifecycle manager startup (#2917)

robotechvision:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2917 |
| Robotic platform tested on | Tested by @AlexeyMerzlyakov |

---

* Fixed lifecycle manager startup failure with FastRTPS which was caused by an [rclcpp bug](https://github.com/ros2/rclcpp/issues/1611) whose fix hasn't been merged yet

* No updates needed

---

* Can be reverted after ros2/rclcpp#1640 has been merged

#### For Maintainers: 

### DIFF
--- a/nav2_lifecycle_manager/src/lifecycle_manager.cpp
+++ b/nav2_lifecycle_manager/src/lifecycle_manager.cpp
@@ -99,10 +99,10 @@ LifecycleManager::LifecycleManager(const rclcpp::NodeOptions & options)
           },
           callback_group_);
       }
+      auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+      executor->add_callback_group(callback_group_, get_node_base_interface());
+      service_thread_ = std::make_unique<nav2_util::NodeThread>(executor);
     });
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

SteveMacenski:
> Can be reverted after https://github.com/ros2/rclcpp/pull/1640 has been merged

Is there a benefit to having it the other way? To be fair, I more or less skimmed the thread, it seemed like you and Alexey were doing a good job at figuring this out. The impression I got was that it doesn't really matter to us.

AlexeyMerzlyakov:
> Is there a benefit to having it the other way?

Current PR is intended to make `service_thread_` to work correctly. This will make to create/maintain new service thread from the same thread where (and right after) the timer callback for this thread was declared: in the right place where it should be even if https://github.com/ros2/rclcpp/pull/1640 to be fixed. So, in my opinion this is so we should have in Nav2 anyway (without future reverting after rclcpp PR to be fixed).

Issue #2958, opened 2022-05-19
Access action feedback within nav2_behavior_tree::BtActionNode  during node status BT::NodeStatus::RUNNING


gstrommer:
#### Feature description
I am using the nav2_behavior_tree::BtActionNode for executing a self defined ros2 action. When the action is running, action feedback is published by the action server. However, the nav2_behavior_tree::BtActionNode has not defined a feedback callback. I also could not find a way to access the action feedback at all. There was a similar issue that has been closed recently (#2839) with the same feature request. 

#### Implementation considerations
One could define a feedback_callback within the `send_goal_options` in the `send_new_goal()` function or the feedback could be accessible within the `void on_wait_for_result()` function.

SteveMacenski:
Agreed, what to do somewhat depends on how you want to use that information. Can you give a bit more specificity? 

I think something reasonable could be a to have `send_goal_options` have a feedback callback that stores the most recent message and then returns it to the BT node as an argument to `on_wait_for_result()`. It would need to be reset after each termination though so we didn't return stale information. 

Alternatively, have the BTActionNode have an additional public member `getFeedback()` or something that would return the most recent feedback that the BT nodes could use anywhere they desire, but probably in or around `on_wait_for_result()`.

gstrommer:
_Agreed, what to do somewhat depends on how you want to use that information. Can you give a bit more specificity?_    
The BtActionNode we are using triggers a simple movement. The feedback would give us some status information like the remaining distance to the goal position. 

I have added a feedback callback using the `send_goal_options` you mentioned as a quick fix for that problem. I think a public member like `getFeedback()` would be a nice fit here.

SteveMacenski:
> The feedback would give us some status information like the remaining distance to the goal position.

What do you aim to do with that information in the BT node? I figure the Action itself should decide when it stops, so what context does the feedback provide you in the context of a BT node being ticked that's useful? I'm trying to understand the usecase better.

>  I think a public member like getFeedback() would be a nice fit here.

You'd prefer that over having it be given to you as an argument of `on_wait_for_result` (ex `on_wait_for_result(feedback)`)?

gstrommer:
We use the same action for several different behaviors. In some cases, we verify the triggered motion by some additional measurements. Therefore, we compare the action feedback with measures, and if they do not match, we abort the action. This monitoring process should run in a parallel BTree node. Consequently, we need the feedback as an output port of our node.

[...]

PR #2972, opened 2022-05-27
Allow access to action feedback in nav2_behavior_tree::BtActionNode::on_wait_for_result

naiveHobo:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2958 |
| Robotic platform tested on | Gazebo simulation of turtlebot3 |

---

* `nav2_behavior_tree::BtActionNode::on_wait_for_result(feedback)` allows access to a shared_ptr to the latest feedback message

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -17,6 +17,7 @@
 
 #include <memory>
 #include <string>
+#include <mutex>
 
 #include "behaviortree_cpp_v3/action_node.h"
 #include "nav2_util/node_utils.hpp"
@@ -133,8 +134,9 @@ public:
   /**
    * @brief Function to perform some user-defined operation after a timeout
    * waiting for a result that hasn't been received yet
+   * @param feedback shared_ptr to latest feedback message
    */
-  virtual void on_wait_for_result()
+  virtual void on_wait_for_result(std::shared_ptr<const typename ActionT::Feedback> /*feedback*/)
   {
   }
 
@@ -206,7 +208,13 @@ public:
       // The following code corresponds to the "RUNNING" loop
       if (rclcpp::ok() && !goal_result_available_) {
         // user defined callback. May modify the value of "goal_updated_"
-        on_wait_for_result();
+        on_wait_for_result(feedback_);
+
+        // reset feedback to avoid stale information
+        {
+          std::lock_guard<std::mutex> lock_guard(feedback_mutex_);
+          feedback_.reset();
+        }
 
         auto goal_status = goal_handle_->get_status();
         if (goal_updated_ && (goal_status == action_msgs::msg::GoalStatus::STATUS_EXECUTING ||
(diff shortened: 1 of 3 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -340,6 +357,12 @@ class BtActionNode : public BT::ActionNodeBase
           result_ = result;
         }
       };
+    send_goal_options.feedback_callback =
+      [this](typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr,
+             const std::shared_ptr<const typename ActionT::Feedback> feedback) {
```
SteveMacenski, line 362: `spin()` happens during `tick()` and specific function calls, so I don't think there's an issue with async access to `feedback_` to need a mutex. Do you agree?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -206,7 +208,13 @@ class BtActionNode : public BT::ActionNodeBase
       // The following code corresponds to the "RUNNING" loop
       if (rclcpp::ok() && !goal_result_available_) {
         // user defined callback. May modify the value of "goal_updated_"
-        on_wait_for_result();
+        on_wait_for_result(feedback_);
+
+        // reset feedback to avoid stale information
+        {
+          std::lock_guard<std::mutex> lock_guard(feedback_mutex_);
```
SteveMacenski, line 215: What's the work flow you're imagining? Users check if the `feedback_` is nullptr? That should be fine, but would be good at least in the signature of `getFeedback()` and `on_wait_for_result()` that it might be nullptr if no new feedback is received so that there's explicit documentation of that (which most people will miss/ignore, but at least we documented it)

SteveMacenski (changes requested):
LGTM mostly

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: Yup, makes sense.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
naiveHobo: That's exactly it although I'm not seeing too much use of the `getFeedback()` function. `on_wait_for_result()` makes the most sense to give access to feedback to the user.

Another way that would make sense API wise would be to add an `on_feedback()` virtual method like the other `on_***` methods that can be overridden by the user to get access to the latest feedback only when it is available. This would remove the need to check for nullptr, which I agree would be missed by most users. What do you think?

SteveMacenski (approved):

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
SteveMacenski: Agreed, the get feedback was an either/or kind of situation (change `on_wait` or add a `getFeedback()`) but I suppose it doesn't hurt to have both, but you could get rid of `getFeedback` and that would be prefectly fine to me

SteveMacenski (approved):
LGTM waiting on CI

SteveMacenski:
@naiveHobo there are a number of linting errors https://app.circleci.com/pipelines/github/ros-planning/navigation2/7360/workflows/be8d4050-1fd7-4b75-b4f3-771e0b431821/jobs/25981

naiveHobo:
@SteveMacenski  Ah, fixed. Sorry about that, I have not been linting too much. :sweat_smile:

SteveMacenski:
No worries!

PR #2977, opened 2022-05-30
use unique_ptr for map_update_thread

gezp:
replace raw pointer with unique_ptr.

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -239,9 +239,7 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
   stop_updates_ = false;
   map_update_thread_shutdown_ = false;
   map_update_thread_ = std::make_unique<std::thread>(
-    [this]() {
-      mapUpdateLoop(map_update_frequency_);
-    });
+    std::bind(&Costmap2DROS::mapUpdateLoop, this, map_update_frequency_));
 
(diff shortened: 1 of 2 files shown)


nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -238,10 +238,10 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
   stopped_ = true;  // to active plugins
   stop_updates_ = false;
   map_update_thread_shutdown_ = false;
-
-  map_update_thread_ = new std::thread(
-    std::bind(
-      &Costmap2DROS::mapUpdateLoop, this, map_update_frequency_));
+  map_update_thread_ = std::make_unique<std::thread>(
+    [this]() {
```
SteveMacenski, line 242: Why not bind?

SteveMacenski: I'd prefer that, if it works

gezp: it seems that lambda is better than bind, you could see more detail here https://github.com/ros-planning/moveit2/issues/872. but if you still think we should use bind, i can change back to bind.

SteveMacenski: Given its a 1 time setup, I don't think the overhead is a big deal

SteveMacenski (approved):

PR #2979, opened 2022-05-30
fix: correct curvature signal on RPP

afonsofonbraga:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Ackermann robot model|

---

* Testing the RPP with an Ackermann robot moving backwards the robot was moving directions randomly not following the global path: 
error
Taking a closer look into the RPP at this line https://github.com/ros-planning/navigation2/blob/80d7d0a080d7401607fd85216f9c450518e7ed42/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L628 the linear_vel is being multiplied by the signal, changing the direction of the curvature at line 
https://github.com/ros-planning/navigation2/blob/80d7d0a080d7401607fd85216f9c450518e7ed42/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L333
My solution was to also multiply at line 333 by the sign to revert the multiplication and not messing with the applyConstraints function. The result is:
correct

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -330,7 +330,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
       costAtPose(pose.pose.position.x, pose.pose.position.y), linear_vel, sign);
 
     // Apply curvature to angular velocity after constraining linear velocity
-    angular_vel = linear_vel * curvature;
+    angular_vel = linear_vel * sign *curvature;
   }
 
   // Collision checking on this velocity heading


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -330,7 +330,7 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
       costAtPose(pose.pose.position.x, pose.pose.position.y), linear_vel, sign);
 
     // Apply curvature to angular velocity after constraining linear velocity
-    angular_vel = linear_vel * curvature;
+    angular_vel = linear_vel * sign *curvature;
```
SteveMacenski, line 333:
```suggestion
    angular_vel = sign * linear_vel * curvature;
```

Fixes a linting problem and also a little more semantically clear if the `sign` is in front

SteveMacenski:
Toggling to trigger CI, it didn't for some reason :shrug:

SteveMacenski:
LGTM other than what's mentioned above + CI run

afonsofonbraga:
Sorry by the force on the push command, the commit wasn't signed off. All fixed now.

SteveMacenski (approved):

SteveMacenski:
Thanks for the patch!

Issue #2985, opened 2022-05-31
Map saver without 'get_parameter'


MartiBolet:
## Feature request
Hi, I see that the `map_saver.cpp` has the `declare_parameter` for `save_map_timeout`, `free_thresh_default`, `occupied_thresh_default` and `map_subscribe_transient_local` but doesn't get the value of the parameters anywhere.

#### Feature description
Add the `get_parameter(...)` inside `on_configure()` transition.

SteveMacenski:
https://github.com/ros-planning/navigation2/blob/main/nav2_map_server/src/map_saver/map_saver.cpp#L49-L54 They are gotten in the constructor. I'll admit that its not the best place for that kind of thing, but they are gotten.

PR #3000, opened 2022-06-06
Get parameters on configure transition addressing #2985

MartiBolet:
Signed-off-by: MartiBolet <mboletboixeda@gmail.com>

Map saver has getting only the parameters in the constructor of the class.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2985|

---

Added the `get_parameter` in `on_configure` transition for `map_saver`.

None

---

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/src/map_saver/map_saver.cpp
+++ b/nav2_map_server/src/map_saver/map_saver.cpp
@@ -46,12 +46,11 @@ MapSaver::MapSaver(const rclcpp::NodeOptions & options)
 {
   RCLCPP_INFO(get_logger(), "Creating");
 
-  save_map_timeout_ = std::make_shared<rclcpp::Duration>(
-    rclcpp::Duration::from_seconds(declare_parameter("save_map_timeout", 2.0)));
-
-  free_thresh_default_ = declare_parameter("free_thresh_default", 0.25),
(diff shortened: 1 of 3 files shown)


nav2_map_server/src/map_saver/map_saver.cpp
```diff
@@ -66,6 +66,10 @@ MapSaver::on_configure(const rclcpp_lifecycle::State & /*state*/)
   // Make name prefix for services
   const std::string service_prefix = get_name() + std::string("/");
 
+  free_thresh_default_ = get_parameter("free_thresh_default").as_double();
```
SteveMacenski, line 69: In the constructor, we need to remove the setting of values too so the declare lines aren't impacting class members (e.g. remove the `X =` )

SteveMacenski (changes requested):

SteveMacenski (approved):

SteveMacenski:
Waiting on CI

SteveMacenski:
```
/opt/overlay_ws/src/navigation2/nav2_map_server/test/map_saver_cli/test_map_saver_cli.cpp:70
  return_code
    Which is: 62720
  0
```

Map CLI test failed, it looks like it needs to be updated from this change

SteveMacenski:
Thanks!

Issue #3002, opened 2022-06-06
String format specifier error when building on arm32


m2-farzan:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04 on qemu+arm32
- ROS2 Version:
  - Galactic (built from source)
- Version or commit hash:
  - galactic
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
Build navigation2 using colcon

#### Expected behavior
Build successfully

#### Actual behavior
```
/root/ros2_galactic_overlay/src/navigation2/nav2_costmap_2d/plugins/inflation_layer.cpp:164:14: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 5 has type ‘std::vector<geometry_msgs::msg::Point_<std::allocator<void> > >::size_type’ {aka ‘unsigned int’} [-Werror=format=]
  164 |     logger_, "InflationLayer::onFootprintChanged(): num footprint points: %lu,"
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  165 |     " inscribed_radius_ = %.3f, inflation_radius_ = %.3f",
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  166 |     layered_costmap_->getFootprint().size(), inscribed_radius_, inflation_radius_);
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                          |
      |                                          std::vector<geometry_msgs::msg::Point_<std::allocator<void> > >::size_type {aka unsigned int}
/root/ros2_galactic_overlay/src/navigation2/nav2_costmap_2d/plugins/inflation_layer.cpp:164:77: note: format string is defined here
  164 |     logger_, "InflationLayer::onFootprintChanged(): num footprint points: %lu,"
      |                                                                           ~~^
      |                                                                             |
      |                                                                             long unsigned int
      |                                                                           %u
cc1plus: all warnings being treated as errors
make[2]: *** [CMakeFiles/layers.dir/build.make:76: CMakeFiles/layers.dir/plugins/inflation_layer.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:281: CMakeFiles/layers.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[...]

PR #3003, opened 2022-06-06
Fix size_t format specifier

m2-farzan:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3002 |
| Primary OS tested on | Ubuntu on arm32 (qemu) |

---

- This patch uses a portable string format specifier for size_t (%zu) to replace non-generic specifiers (%lu, %li)
- This patch is necessary (and sufficient) for building navigation2 on arm32 machines

---

#### For Maintainers: 


SteveMacenski:
Thanks!

Issue #3014, opened 2022-06-14
Obstacle Layer, Voxel Layer, and Costmap Topic Collision Checker not working in Humble/Rolling due to Fast-DDS Regression


SteveMacenski:
More info https://github.com/ros-planning/navigation2/pull/2489#issuecomment-1154601171

Seems to have broken sine https://github.com/ros-planning/navigation2/pull/2489 for 22.04 / Humble updates. 

CC @jwallace42 @gezp

gezp:
the problem is that Obstacle Layer not working (the scan data isn't subscribed by Obstacle Layer),  becasue `observation_subscribers_->subsciber()` failed when `ObstacleLayer::activate()` is called.

it seems that we can't create subscriber in another callback group at runtime since rollling updated in ubuntu22.04, but i don't know why. and i make a simple PR to sovle this problem by removing `subscribe()`.

SteveMacenski:
I think that makes sense in the interim, but can you file a ticket describing this issue to the message filters repo? Do you see the same behavior if you create a minimal example or just something we see in Nav2's use of it?

I went through the git blame for message filters and don't see any changes in that repo that seem problematic (or really even substantial) in the last year.

SteveMacenski:
https://github.com/ros2/message_filters/blob/master/include/message_filters/subscriber.h#L88-L117

It seems like it might be related to this? It might be using the wrong constructor and even though we pass the options, if the 2nd one is called then its not even sent. These are called in the constructor https://github.com/ros2/message_filters/blob/master/include/message_filters/subscriber.h#L186-L202.

SteveMacenski:
I determined root cause of the costmap collision checker issue. Since the obstacle layer isn't getting its callbacks as @gezp reported, the local costmap is empty so it shows always collision free. We use the local costmap in the behavior server, not the global costmap, which does not have the static map set by default in `nav2_params.yaml`. 

So once we solve the issue described in https://github.com/ros2/rclcpp/issues/1954 / short term patch in https://github.com/ros-planning/navigation2/pull/3018 then this should all be good

SteveMacenski:
https://github.com/ros2/rclcpp/issues/1954 is critically related

ladianchad:
I found same issue. (obstacle_layer doesn't subscribe sensor callback)
So, i tried short term patch in https://github.com/ros-planning/navigation2/pull/3018, and it works good

padhupradheep:
could we pin the issue on the top?

[...]

PR #3018, opened 2022-06-15
fix ObstacleLayer not working due to #2489

gezp:

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #3014) |
| Primary OS tested on | (Ubuntu22.04) |

---

Obstacle Layer not working (the scan data isn't subscribed by Obstacle Layer),  becasue `observation_subscribers_->subsciber()` failed when `ObstacleLayer::activate()` is called since #2489

* use `active_` flag instead of `subscibe()` and `unsubscibe()`  (refer to https://github.com/ros-planning/navigation2/blob/main/nav2_amcl/src/amcl_node.cpp#L614)

i'm not sure the side effect of this change, maybe we could have a better solution.

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/obstacle_layer.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/obstacle_layer.hpp
@@ -252,6 +252,7 @@ protected:
   bool rolling_window_;
   bool was_reset_;
   int combination_method_;
+  bool active_{false};
 };
 
 }  // namespace nav2_costmap_2d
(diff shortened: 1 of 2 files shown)


gezp:
CC @jwallace4

gezp:
CC @jwallace42

SteveMacenski:
If callback groups aren't being respected, how does bypassing the unsubscribe help? Per rclcpp issue thread

gezp:
> If callback groups aren't being respected, how does bypassing the unsubscribe help? Per rclcpp issue thread

maybe I don't get your meaning，the problem is caused by `subscribe()`, not related to `unsubscribe()` .

SteveMacenski:
https://github.com/ros2/rmw_fastrtps/pull/619 supposedly fixes it and is merged!

dcconner:
Humble binaries as of 18-Oct still have this issue.  Switching to Cyclone seems to fix.  Just commenting here for others that might be trying to track down issue

PR #3028, opened 2022-06-21
cleanup warnings

jwallace42:
---

## Basic Info
Cleaning up behaviors

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
@@ -63,18 +63,17 @@ public:
   TimedBehavior()
   : action_server_(nullptr),
     cycle_frequency_(10.0),
-    enabled_(false)
+    enabled_(false),
+    transform_tolerance_(0.0)
   {
   }
 
-  virtual ~TimedBehavior()
-  {
-  }
+  virtual ~TimedBehavior() = default;
 
   // Derived classes can override this method to catch the command and perform some checks
   // before getting into the main loop. The method will only be called
   // once and should return SUCCEEDED otherwise behavior will return FAILED.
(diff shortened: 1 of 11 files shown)


nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
```diff
@@ -63,18 +63,17 @@ class TimedBehavior : public nav2_core::Behavior
...
-  virtual ~TimedBehavior()
-  {
-  }
+  virtual ~TimedBehavior() = default;
 
   // Derived classes can override this method to catch the command and perform some checks
   // before getting into the main loop. The method will only be called
   // once and should return SUCCEEDED otherwise behavior will return FAILED.
-  virtual Status onRun(const std::shared_ptr<const typename ActionT::Goal> command) = 0;
+  virtual Status onRun(const std::shared_ptr<const typename ActionT::Goal> &command) = 0;
```
SteveMacenski, line 76: Remove reference, its a shared pointer, it needs to be given as a shared pointer for the reference counter to be incremented in case its stored long term

jwallace42: None of the behaviors store the command. Instead, values of the command are copied into member variables. We could change the behaviors to store the command instead of multiple member vars.

SteveMacenski: If we make the API that and someone else wants to make their own behavior in the future, that could become a really subtle problem for them. But really, its mostly about best practices, we shouldn't pass references to shared pointers. That's just a C++11 best practices point I learned

SteveMacenski:
Just waiting on CI

Issue #3027, opened 2022-06-21
BT Action Node initialization does not timeout


vinnnyr:
#### Feature description
Add timeout to action client initialization in a BT action client node.

https://github.com/ros-planning/navigation2/blob/08e558fd5a7205142935190ce26d196b8c828911/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L93

Currently this `wait_for_action_server` call does not have a timeout. If the given action server is not healthy, or we are having other RMW/DDS issues, whatever component of the system that initializes the plugin / BT will just get stuck here. 

#### Implementation considerations
Instead, I propose that we use the `server_timeout` parameter to have a timeout for the `wait_for_action_server` and throw an exception if the timeout has been violated so that the upper level system can handle the error / inform the user, or other parts of the system rather than hang out waiting for the action server indefinitely. 

I believe it should be as simple as 
```
auto was_successful = action_client_->wait_for_action_server(server_timeout_)
if (!was_successful) {
  throw std::runtime_error("Server Timeout exceeded");
```

PR #3029, opened 2022-06-21
adding timeout for action client initialization

vinnnyr:
Signed-off-by: Vinny Ruia <vinny.ruia@fireflyautomatix.com>

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3027  |
| Primary OS tested on | N/A* |
| Robotic platform tested on | N/A* |

* I am currently having trouble compiling from `main` on my system. I started this PR so that I can see if CI can do it.
---

Adding timeout to the action server initialization

N/A

---

N/A

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -90,10 +90,12 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    if (!action_client_->wait_for_action_server(server_timeout_)) {
+    static constexpr std::chrono::milliseconds wait_for_server_timeout =
+      std::chrono::milliseconds(1000);
+    if (!action_client_->wait_for_action_server(wait_for_server_timeout)) {
       RCLCPP_ERROR(
         node_->get_logger(), "\"%s\" action server not available after waiting for %li ms",
-        action_name.c_str(), server_timeout_.count());
+        action_name.c_str(), wait_for_server_timeout.count());
       throw std::runtime_error("Action server not available");
     }
   }
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -90,7 +90,12 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    action_client_->wait_for_action_server();
+    if (!action_client_->wait_for_action_server(server_timeout_)) {
+      RCLCPP_ERROR(
+        node_->get_logger(), "\"%s\" action server not available after waiting for %li ms",
+        action_name.c_str(), server_timeout_.count());
+      throw std::runtime_error("Action server not available");
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -90,7 +90,12 @@ class BtActionNode : public BT::ActionNodeBase
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    action_client_->wait_for_action_server();
+    if (!action_client_->wait_for_action_server(server_timeout_)) {
```
vinnnyr, line 93: Ok I think I have to rethink this a small bit. The default `server_timeout` is 20ms, which is not a lot of time to wait by default. 
Should I declare a new parameter for this timeout?

SteveMacenski: I'm hesitant to have a new timeout since that might confuse people, even if well named - but maybe there's a good name that would be clear so its moot? `wait_for_action_server_timeout` `server_initialization_timeout`?

Otherwise, if we can't come up with a good & specific name, I think just hardcoding it at some value would be fine. I'd say 1s just to be nice and round. We have a few like that (only a few) in the stack where it didn't make much sense to parameterize it for various reasons and since they were just initialization operations, it wasn't super important for them to be configurable for 99.8% of users.

vinnnyr:
Good point. 

At least on startup, the exception is caught by the Parent Lifecycle Node


> server_name  | [action_server_1] [ERROR] [1655837597.993815093] [action_name_rclcpp_node]: "follow_path" action server not available after waiting for 2000 ms
server_name  | [action_server_1] [ERROR] [1655837598.076770031] []: Caught exception in callback for transition 13
server_name  | [action_server_1] [ERROR] [1655837598.076800292] []: Original error: Action server not available
server_name  | [action_server_1] [WARN] [1655837598.076856566] []: Error occurred while doing error handling.
server_name  | [ERROR] [launch_ros.actions.lifecycle_node]: Failed to make transition 'TRANSITION_ACTIVATE' for LifecycleNode '/action'

I can look into how it gets caught in steady state too.

SteveMacenski:
Prob just need another try/catch somewhere in the chain

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -90,7 +90,14 @@ class BtActionNode : public BT::ActionNodeBase
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    action_client_->wait_for_action_server();
+    static constexpr std::chrono::milliseconds wait_for_server_timeout =
+      std::chrono::milliseconds(1000);
```
vinnnyr, line 94: hardcoded to a reasonable 1s now

vinnnyr:
@SteveMacenski  Take a look at your convenience. I believe I addressed your concerns, including handling the "switch at runtime" case.

vinnnyr:
As an aside, I am unsure which tests are actually failing here. 
Atleast on my browser / machine, the CircleCI tab for tests seems to be hanging indefinitely
https://app.circleci.com/pipelines/github/ros-planning/navigation2/7517/workflows/12f61cb3-ffeb-414c-8722-b47d4e422afa/jobs/26329/tests

SteveMacenski:
The tests came through, it did have 1 failure that is unusual so retriggering, but I don't think its related to this PR. I'll let it spin though and check back in the morning

PR #3038, opened 2022-06-23
Completed PR 2929

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/2929


SteveMacenski:
@AlexeyMerzlyakov can you review?

nav2_map_server/test/component/test_map_server_node.cpp
```diff
@@ -24,6 +24,8 @@
 #include "nav2_map_server/map_server.hpp"
 #include "nav2_util/lifecycle_service_client.hpp"
 #include "nav2_msgs/srv/load_map.hpp"
+using namespace std::chrono_literals;
+using namespace rclcpp;  // NOLINT
```
AlexeyMerzlyakov, line 28: I do not think, we need to be in `rclcpp` namespace. Everything is seems already addressed starting from rclcpp, e.g.:
```
auto parameters_client = std::make_shared<rclcpp::SyncParametersClient>(node_, "map_server");
```
so this is unnecessary

nav2_map_server/include/nav2_map_server/map_server.hpp
```diff
@@ -138,6 +138,9 @@ class MapServer : public nav2_util::LifecycleNode
 
   // The message to publish on the occupancy grid topic
   nav_msgs::msg::OccupancyGrid msg_;
+
+  // true if msg_ was initialized
+  bool map_available_;
```
AlexeyMerzlyakov, line 143: Just curious: if we will make `msg_` to be a shared pointer `nav_msgs::msg::OccupancyGrid::SharedPtr`, we could initialize it with `nullptr`. And this will be an indicator, whether the map is available or not. Boolean `map_available_` is not needed in this case.

AlexeyMerzlyakov (commented):
@SteveMacenski, the review was done. The update for last version of #2929 items were resolved, except of `rclcpp` namespace. Also, I've got an idea, how to reduce the number of variable in map server class.

SteveMacenski:
Agreed, though I removed the `namespace rclcpp;` and it failed to build so its used somewhere. Its a test file so I'm OK with a little less strict great coding style as its not used / maintained the same as functional code.

PR #3042, opened 2022-06-24
Semantic segmentation plugin

pepisg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | None to my knowledge |
| Robotic platform tested on | Custom skid steer drive robot on gazebo and in real life |

---

* I added a semantic segmentation plugin to the `nav2_costmap_2d` package for adding segmentation data to costmaps. For now this assumes segmentation comes from and RGBD sensor that can produce aligned pointclouds. The configuration for this plugin looks like this: 
```
    semantic_segmentation_layer:
        plugin: nav2_costmap_2d::SemanticSegmentationLayer
        enabled: True
        segmentation_topic: "/semantic_segmentation/segmentation/image_raw"
        pointcloud_topic: "/camera/depth/color/points"
        max_obstacle_distance: 3.5
        min_obstacle_distance: 0.5
        class_types: ["traversable", "warning", "danger"]
        traversable:
          classes: ["Sidewalk"]
          cost: 0
        warning:
          classes: ["Pads"]
          cost: 90
        danger:
          classes: ["Grass", "Other"]
          cost: 254
```
A ray tracing approach was tried to make the plugin work with calibrated RGB cameras but was discarded because projections of obstacles over the ground plane caused obstacles to extend beyond their actual location. @SteveMacenski let me know if you think more information about why this approach was not pursued should be added.

 The plugin is dependant on new [segmentation messages](https://github.com/kiwicampus/vision_msgs/blob/feature/semantic-segmentation/msg/SemanticSegmentation.msg). @SteveMacenski, let me know if you agree with this definition for opening a PR to `vision_msgs`

* Added a new costmap plugin, so a configuration guide needs to be added. The documentation on `nav2_costmap_2d` also needs updating
* Also it would be good to add a tutorial on `navigation2_tutorials`. @SteveMacenski I could provide a demo segmentation node for detecting gazebo's ground plane on an image, let me know what you think would be the best way of doing this.

PD: I will start doing this after the first review in which I expect several changes will be requested

* Extend the plugin to work with any combination of segmentation sources and pointclouds (e.x: 3D lidar + RGB camera) 
* Integrate various handling methods for adjusting cost of pixels according to the confidence of the inference

#### For Maintainers: 

NOTE: This PR will not build due to its dependency to `vision_msgs`'s new messages


nav2_costmap_2d/include/nav2_costmap_2d/semantic_segmentation_layer.hpp
```diff
@@ -0,0 +1,161 @@
...
+  ///< @brief subscriber to guarantee that transforms are available when pointclouds are received
+  std::shared_ptr<tf2_ros::MessageFilter<sensor_msgs::msg::PointCloud2>> pointcloud_tf_sub_;
+  ///< @brief subscriber to sync transforms, pointclouds and segmentations
+  std::shared_ptr<message_filters::TimeSynchronizer<vision_msgs::msg::SemanticSegmentation,
+                                                    sensor_msgs::msg::PointCloud2>>
+    segm_pc_sync_;
+
+  // debug publishers
+  ///< @brief publisher to publish segmentation messages when buffered
+  std::shared_ptr<rclcpp::Publisher<vision_msgs::msg::SemanticSegmentation>> sgm_debug_pub_;
```
SteveMacenski, line 138: Remove `sgm_debug_pub_` `orig_pointcloud_pub_` `proc_pointcloud_pub_`, these are the same as the input topics, there's no reason to repeat them

nav2_costmap_2d/plugins/semantic_segmentation_layer.cpp
```diff
@@ -0,0 +1,304 @@
...
+      break;
+  }
+}
+
+void SemanticSegmentationLayer::syncSegmPointcloudCb(
+  const std::shared_ptr<const vision_msgs::msg::SemanticSegmentation>& segmentation,
+  const std::shared_ptr<const sensor_msgs::msg::PointCloud2>& pointcloud)
+{
+  // check if data has the right dimensions
+  if (segmentation->width * segmentation->height != pointcloud->width * pointcloud->height)
```
SteveMacenski, line 259: Can we not support them if they're not aligned? That would be important to note in documentation that this requires pre-aligned pointclouds.

SteveMacenski: It would be nice to support this in the future, many RGBD sensors will give you different resolution choices for RGB and D information.

nav2_costmap_2d/plugins/semantic_segmentation_layer.cpp
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  global_frame_ = layered_costmap_->getGlobalFrameID();
+  rolling_window_ = layered_costmap_->isRolling();
+
+  matchSize();
+
+  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_sensor_data;
+
+  semantic_segmentation_sub_ =
```
SteveMacenski, line 128: Subs should use the provided callback group https://github.com/ros-planning/navigation2/blob/985fd7e61a9701f130c41fe5e95e009847465ad2/nav2_costmap_2d/plugins/obstacle_layer.cpp#L123 so that it runs on another thread

nav2_costmap_2d/src/segmentation_buffer.cpp
```diff
@@ -0,0 +1,264 @@
...
+    // update the point step and get iterators
+    modifier.resize(cloud_size);
+    sensor_msgs::PointCloud2Iterator<uint8_t> iter_class_obs(segmentation_cloud, "class");
+    sensor_msgs::PointCloud2Iterator<uint8_t> iter_confidence_obs(segmentation_cloud, "confidence");
+    sensor_msgs::PointCloud2Iterator<float> iter_x_obs(segmentation_cloud, "x");
+    sensor_msgs::PointCloud2Iterator<float> iter_y_obs(segmentation_cloud, "y");
+    sensor_msgs::PointCloud2Iterator<float> iter_z_obs(segmentation_cloud, "z");
+
+    // get iterators for the original cloud
+    sensor_msgs::PointCloud2ConstIterator<float> iter_x_global(global_frame_cloud, "x");
```
SteveMacenski, line 137: Same here with use of iterators, see the normal observation buffer as an example

SteveMacenski (changes requested):
Good first go at it! Mostly just smaller stuff to be honest (beyond missing tests)

nav2_costmap_2d/include/nav2_costmap_2d/semantic_segmentation_layer.hpp
pepisg: I think its worth leaving `proc_pointcloud_pub`. This publishes poinclouds with the segmentation `class` and `confidence` fields in each point. It may be useful for visualization. What do you think? This is subject to the `debug_topics` parameter anyway, which is False by default

nav2_costmap_2d/plugins/semantic_segmentation_layer.cpp
pepisg: I think different resolutions can be supported if they are aligned, the mask could be down/up scaled to match the pointcloud resolution (I can do it if you think is reasonable). For not aligned pointclouds things will get more complicated, extra information should be passed to the plugin since it must do the alignment before filling the costmap. I will leave that in the documentation, maybe an issue can be created when the PR is merged to keep it on track

nav2_costmap_2d/include/nav2_costmap_2d/semantic_segmentation_layer.hpp
pepisg: Messages are not published through this topic unless the parameter `debug_topics` is set to `true` in the plugin configuration. This parameter is `false` by default, so users should actually define it to get the messages. Do you still think it must be removed?

SteveMacenski: OK, if you leave that last one that's OK

pepisg:
Hi! 

Sorry for the massive delay. After quite a while I was finally able to pick up this PR. These are the latest changes:

- I removed the custom segmentation message the PR was initially created with. It nows subscribes to `sensor_msgs/msg/Image` as mask and a latched [`vision_msgs/msg/LabelInfo`](https://github.com/ros-perception/vision_msgs/blob/ros2/vision_msgs/msg/LabelInfo.msg) for associating class names to class ids in the mask. 
- I made a simple node for creating segmentation masks in gazebo called [mono_color_segmentation](https://github.com/kiwicampus/mono_color_segmentation) for testing the layer; this node creates a binary mask that separates the ground plane from the other objects in the scene by their color: 

  ezgif com-speed

Here's a video of the layer working in the local costmap:  

https://github.com/ros-planning/navigation2/assets/71234974/1ca670ad-f62c-431b-ba5f-74f7a66ec3ca

It was made with the testing segmentation node mentioned above and the following paramters: 

```.yaml
local_costmap:
  local_costmap:
    ros__parameters:
      update_frequency: 5.0
      publish_frequency: 2.0
      global_frame: odom
      robot_base_frame: base_link
      rolling_window: true
      width: 3
      height: 3
      resolution: 0.05
      robot_radius: 0.22
      plugins: ["semantic_segmentation_layer", "inflation_layer"]
      semantic_segmentation_layer:
        plugin: nav2_costmap_2d::SemanticSegmentationLayer
        enabled: True
        observation_sources: mono_color_segmentation
        mono_color_segmentation:
          segmentation_topic: "/intel_realsense_r200_depth/mask"
          labels_topic: "/intel_realsense_r200_depth/label_info"
          pointcloud_topic: "/intel_realsense_r200_depth/points"
          max_obstacle_distance: 1.5
          min_obstacle_distance: 0.2
          class_types: ["traversable", "untraversable"]
          traversable:
            classes: ["ground_plane"]
            cost: 0
          untraversable:
            classes: ["others"]
            cost: 254
      inflation_layer:
        plugin: "nav2_costmap_2d::InflationLayer"
        cost_scaling_factor: 3.0
        inflation_radius: 0.55

global_costmap:
  global_costmap:
    ros__parameters:
      update_frequency: 1.0
      publish_frequency: 1.0
      global_frame: map
      robot_base_frame: base_link
      robot_radius: 0.22
      resolution: 0.05
      track_unknown_space: true
      plugins: ["static_layer", "semantic_segmentation_layer", "inflation_layer",]
      semantic_segmentation_layer:
        plugin: nav2_costmap_2d::SemanticSegmentationLayer
        enabled: True
        observation_sources: mono_color_segmentation
        mono_color_segmentation:
          segmentation_topic: "/intel_realsense_r200_depth/mask"
          labels_topic: "/intel_realsense_r200_depth/label_info"
          pointcloud_topic: "/intel_realsense_r200_depth/points"
          max_obstacle_distance: 3.0
          min_obstacle_distance: 0.2
          class_types: ["traversable", "untraversable"]
          traversable:
            classes: ["ground_plane"]
            cost: 0
          untraversable:
            classes: ["others"]
            cost: 254
      inflation_layer:
        plugin: "nav2_costmap_2d::InflationLayer"
        cost_scaling_factor: 3.0
        inflation_radius: 0.55
      static_layer:
        plugin: "nav2_costmap_2d::StaticLayer"
        map_subscribe_transient_local: True
```

As a comment, I think the turtlebot world may not be the best simulation to showcase the plugin's capabilities, this layer is more for terrain traversability rather than for geometry-based traversability. It may be worth creating another simple world with a sidewalk-like surface for this purpose. 

I think this layer should also include:

1. A clearing mechanism, right now the only way to clear the layer is to get a new observation on a previously marked tile that now belongs to a traversable class. Something like stvl's decay would be good IMO.
2. Support for taking in confidence masks and a strategy for including that information. From discussions I've had with @SteveMacenski this may be a simple exclusion rule: "pixels with confidence less than `X` do not populate the costmap"; or a more fancy accumulation logic like continously "summing" the confidence on the inference on each costmap tile and having a rule that adjusts cost based on accumulated confidence, that way marking a tile with max cost requires a number of consistent observations for the same class on that tile. 

Those may be included in separate PRs though, maybe this layer can be merged as an MVP with the current "passthrough" logic (where observations are put directly on the costmap regardless of the confidence and never get cleared), and then these and other improvements can me worked on in ther own PRs.

Let me know what you think @SteveMacenski

Issue #3045, opened 2022-06-27
RPP approach velocity scaling should not use lookahead distance as a metric


Aposhian:
#### Feature description
Currently, RPP decides when to trigger approach velocity scaling based on if the actual lookahead distance is different from the requested lookahead distance by a certain margin. This is problematic because lookahead distance is determined from current velocity, so there is a cyclic dependency that can lead to oscillations at the stop point.

```mermaid
graph TD
    velocity --> lookahead_distance --> velocity_scaling --> cmd_vel --> velocity
```
![image (6)](https://user-images.githubusercontent.com/16969110/175988043-4ee8e0bc-fe1d-47e2-bfd7-7e28bc391de0.png)

#### Implementation considerations
Rather than looking to lookahead distance as a proxy measure, I wonder if there is some way to pipe the distance to the goal from a goal checker? Because simple euclidean distance to the goal pose may not be compatible with all goal checkers.

Velocity smoothing and motion profiling has been suggested as a fix to this issue, but I think that we should be able to come up with a simple solution for RPP that fixes this underlying cyclic dependency in the algorithm.

SteveMacenski:
From https://github.com/ros-planning/navigation2/issues/2971

> The first place I'd go is to say that the distance error is the wrong metric to use because of this case. Instead, we could have it be a radius to goal metric with the same other logic. If the last point on the path is within lookahead_dist radius, apply the same logic with the same 'error' calculation. That would be a similar proxy, though inexact.

Perhaps that is the answer? Just a new parameter for proximity to apply it and have the velocity scaling be a function of radial distance from goal rather than lookahead error? I believe this would also solve your other ticket https://github.com/ros-planning/navigation2/issues/2971

> I wonder if there is some way to pipe the distance to the goal from a goal checker? 

The thing about the goal checker is that is to check if its within tolerances to stop navigation entirely. We would need to be scaling velocities to slow down before we got into that tolerance range from what a goal checker would have to say - else we (1) just stop abruptly negating the feature or (2) overshoot the goal because we only then start to slow down.

SteveMacenski:
> What if it looked at the integrated path distance remaining on the transformed plan as the criteria for when to apply approach velocity scaling? That would make it not have to indirectly reference lookahead distance like the current approach does.

Agreed, and solve the other ticket too

SteveMacenski:
Merging imminent

PR #3047, opened 2022-06-27
do not depend on velocity for approach scaling

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3045  |
| Robotic platform tested on | Not yet |

---

Depend only on the path length left on whether to scale velocity on approach, since depending on lookahead distance results in a cyclic dependency

Added a new parameter.

---

Need to test this and build it still.

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -208,17 +208,25 @@ protected:
    */
   double costAtPose(const double & x, const double & y);
 
+  double approachVelocityScalingFactor(
+    const nav_msgs::msg::Path::SharedPtr & path
+  ) const;
+
+  void applyApproachVelocityScaling(
+    const nav_msgs::msg::Path::SharedPtr & path
+    double & linear_vel
+  ) const;
+
   /**
    * @brief apply regulation constraints to the system
    * @param linear_vel robot command linear velocity input
-   * @param dist_error error in the carrot distance and lookahead distance
    * @param lookahead_dist optimal lookahead distance
    * @param curvature curvature of path
    * @param speed Speed of robot
    * @param pose_cost cost at this pose
    */
   void applyConstraints(
-    const double & dist_error, const double & lookahead_dist,
+    const double & lookahead_dist,
     const double & curvature, const geometry_msgs::msg::Twist & speed,
     const double & pose_cost, double & linear_vel, double & sign);
 
@@ -281,6 +289,7 @@ protected:
   bool use_velocity_scaled_lookahead_dist_;
   tf2::Duration transform_tolerance_;
   double min_approach_linear_velocity_;
+  double approach_velocity_scaling_dist_;
   double control_duration_;
   double max_allowed_time_to_collision_up_to_carrot_;
   bool use_regulated_linear_velocity_scaling_;
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -209,11 +209,11 @@ protected:
   double costAtPose(const double & x, const double & y);
 
   double approachVelocityScalingFactor(
-    const nav_msgs::msg::Path::SharedPtr path
+    const nav_msgs::msg::Path & path
   ) const;
 
   void applyApproachVelocityScaling(
-    const nav_msgs::msg::Path::SharedPtr path
+    const nav_msgs::msg::Path & path,
     double & linear_vel
   ) const;
 
@@ -226,9 +226,9 @@ protected:
    * @param pose_cost cost at this pose
    */
   void applyConstraints(
-    const double & lookahead_dist,
     const double & curvature, const geometry_msgs::msg::Twist & speed,
-    const double & pose_cost, double & linear_vel, double & sign);
+    const double & pose_cost, const nav_msgs::msg::Path & path,
+    double & linear_vel, double & sign);
 
   /**
    * @brief Find the intersection a circle and a line segment.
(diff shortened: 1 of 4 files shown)


nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -281,6 +289,7 @@ class RegulatedPurePursuitController : public nav2_core::Controller
   bool use_velocity_scaled_lookahead_dist_;
   tf2::Duration transform_tolerance_;
   double min_approach_linear_velocity_;
+  double approach_velocity_scaling_dist_;
```
SteveMacenski, line 292: The usual parameter guide + migration guide comment applies :smile:

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -325,7 +330,6 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     rotateToHeading(linear_vel, angular_vel, angle_to_heading, speed);
   } else {
     applyConstraints(
```
SteveMacenski, line 332: This call does not contain the `transformed_plan`? Don't believe this compiles

SteveMacenski (changes requested):
Beyond the obvious points, this seems good to me!

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Aposhian: Yes, I haven't fully fleshed this out yet. I will tie up all the loose ends when I get time.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -571,10 +575,39 @@ double RegulatedPurePursuitController::costAtPose(const double & x, const double
   return static_cast<double>(cost);
 }
 
+double RegulatedPurePursuitController::approachVelocityScalingFactor(
+  const nav_msgs::msg::Path::SharedPtr & path
+) const
+{
+  double remaining_distance = nav2_util::geometry_utils::calculate_path_length(path);
+  if (remaining_distance < approach_velocity_scaling_dist_) {
```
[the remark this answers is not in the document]
Aposhian: - [ ] document appropriate bounds for approach velocity scaling dist

SteveMacenski:
@Aposhian any update? It would be great to have this in before the next sync & close out https://github.com/ros-planning/navigation2/issues/3045

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -208,17 +208,25 @@ class RegulatedPurePursuitController : public nav2_core::Controller
    */
   double costAtPose(const double & x, const double & y);
 
+  double approachVelocityScalingFactor(
+    const nav_msgs::msg::Path::SharedPtr & path
```
padhupradheep, line 212: I guess, we do not want reference to the shared pointer. It does not increase the reference count of the shared pointer.

Aposhian:
I may be able to get to this on Saturday, but I have been a little swamped.

Aposhian:
Here is how it ends up performing with the default 1m. I think that is probably not a good default. I will try to figure out how to set this parameter to have roughly the same behavior as before.
image

SteveMacenski:
Sounds good!

Aposhian:
Ok I just updated the default to be one costmap cell less than the forward costmap extent, so that should approximate the old behavior (where scaling was enabled when lookahead distance was shorter than requested). Here is the behavior with that now on my robot sim:
image
One problem with looking at integrated distance is that it jumps by the robot jumping between poses (this is with a sparse path). However, I think this is a far better behavior than what is existing, and this is also an issue that the velocity smoother could help remedy, correct?

SteveMacenski:
> Ok I just updated the default to be one costmap cell less than the forward costmap extent, so that should approximate the old behavior

Uh, wasn't it before that it started to slow down at the lookahead distance away, e.g. by default about 0.6m? This is much higher than that if its 1/2 the costmap size. The added comment in the readme also doesn't say that you should never set it to be higher than that -- I don't think that should be the default. Its a warning not to exceed, not a recommended value. 

Maybe its a better value, but I suppose you didn't mention that you tried the old setting (~0.6) and found that the longer distance was better. 

Yes, the vel smoother will work out the jaggedness.

SteveMacenski:
> Well, I figured there wasn't a good way to recreate the average lookahead distance, since the velocity is unknown. But maybe the default could be calculated based off lookahead time and desired linear velocity?

Why not just use the value of the parameter `lookahead_dist` as the default? Though, if Adaptive behavior is being used, I suppose just `max_lookahead_dist` would be fine too, but if I had to pick one, I'd pick `lookahead_dist` since that should be no matter what type of PP was being used. Just needs to be documented that it defaults to `lookahead_dist` unless otherwise overridden (and should never exceed costmap size / 2)

Aposhian:
Docs PR: https://github.com/ros-planning/navigation.ros.org/pull/336

Aposhian:
I decided to go with just making it an independent default, but going to 0.6 to match previous behavior. Unit tests are updated. I'm just going to do one more test in simulation with this and then I will be happy with it.

Aposhian:
With some tuning this is what the approach velocity scaling ends up looking like (on a sparse path)
image
I think we can probably do better, but this at least removes the oscillating behavior because of the cyclic dependency.

Aposhian:
@SteveMacenski I think all tasks should be addressed. Let me know if there is anything else.

Aposhian:
If you weren't using adaptive lookahead distance before, then this change will be a regression: it will result in a less smooth scaling. I think we can tweak this algorithm to make it scale linearly, regardless of path density. I just need to brainstorm for a bit.

SteveMacenski (approved):

SteveMacenski:
Toggling for CI, Circle's been flaky lately. Will merge once builds

Aposhian:
Here is what it looks like with the smoothed version:
image

PR #3053, opened 2022-06-29
Add duration parameter to the cantransform call in updateCostmap in r…

arslogavitabrevis:
Range_sensor_layer

---

## Basic Info

| Info |  |
| ------ | ----------- |
| Ticket this addresses   | 2975 |

---

* Added a tolerance for the canTransform call in range_sensor_layer. The transformation tolerance value is receivend form the parameter transform_tolerance that was already declared.
---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/range_sensor_layer.cpp
+++ b/nav2_costmap_2d/plugins/range_sensor_layer.cpp
@@ -292,8 +292,11 @@ void RangeSensorLayer::updateCostmap(
   in.header.stamp = range_message.header.stamp;
   in.header.frame_id = range_message.header.frame_id;
 
+  std::string * errstr = NULL;
   if (!tf_->canTransform(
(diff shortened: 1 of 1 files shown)


nav2_costmap_2d/plugins/range_sensor_layer.cpp
```diff
@@ -292,8 +292,11 @@ void RangeSensorLayer::updateCostmap(
   in.header.stamp = range_message.header.stamp;
   in.header.frame_id = range_message.header.frame_id;
 
+  std::string * errstr = NULL;
```
SteveMacenski, line 295: Might want to include that error string in the RCLCPP INFO when it fails to give more context, if you're going to collect it!

SteveMacenski (commented):
LGTM beyond the galactic base

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/3064 to complete

PR #3068, opened 2022-07-11
Navigator becomes plugin.

ladianchad:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | None to my knowledge |
| Primary OS tested on | Ubuntu 22.04 & humble |
| Robotic platform tested on | gazebo simulation of  turtlebot |

---

* I changed navigator to plugin and added parameter  ```naviagtor```. user can choose navigator plugin and customize own navigator. The configuration for this plugin looks like this:
```yaml
bt_navigator:
  ros__parameters:
    use_sim_time: True
    global_frame: map
    robot_base_frame: base_link
    odom_topic: /odom
    bt_loop_duration: 10
    default_server_timeout: 20
    # 'default_nav_through_poses_bt_xml' and 'default_nav_to_pose_bt_xml' are use defaults:
    # nav2_bt_navigator/navigate_to_pose_w_replanning_and_recovery.xml
    # nav2_bt_navigator/navigate_through_poses_w_replanning_and_recovery.xml
    # They can be set here or via a RewrittenYaml remap from a parent launch file to Nav2.
    plugin_lib_names:
    - nav2_compute_path_to_pose_action_bt_node
    - nav2_compute_path_through_poses_action_bt_node
    - nav2_smooth_path_action_bt_node
    - nav2_follow_path_action_bt_node
    #....

    navigator: ["navigate_to_pose", "navigate_through_poses"]
    navigate_to_pose:
      plugin: "nav2_bt_navigator/NavigateToPoseNavigator"
    
    navigate_through_poses:
      plugin: "nav2_bt_navigator/NavigateThroughPosesNavigator"
```
* if user want don't use poses navigator, just remove "navigate_through_poses" in navigator param

* Added a new bt_navigator plugin, so a configuration guide needs to be added. The documentation on bt_navigator also needs updating
*Also it would be good to add a tutorial on navigation2_tutorials. 

---

* I think navigator muxer will be managed by shared_ptr or weak_ptr (now just pointer)

#### For Maintainers: 


SteveMacenski:
Can you clarify this PR as to why it is required and what other things might be useful to add? I think there's some missing motivation behind this and some of the design decisions before getting into the implementation details.

ladianchad:
@SteveMacenski 
Thank you for your comment!!

When developing using navigation2, To make better use of the navigation behavior tree, I often thought it would be good to put more various information on the blackboard at the beginning when the action is called. (customize)
for example,
When checking which controller (or planner) works better in a specific situation, it is possible to create several behavior trees in the past, but if there are many custom controllers to be tested, there will be many behavior trees with the same content.
In this situation, create an action like the one below and a navigator that stores the controller (or planner) to be used initially on the black board, you can easily solve this situation with one behavior tree.

test.action
```
#goal definition
geometry_msgs/PoseStamped pose
string behavior_tree
string controller_id
string planner_id
---
#result definition
std_msgs/Empty result
---
#feedback definition
geometry_msgs/PoseStamped current_pose
builtin_interfaces/Duration navigation_time
builtin_interfaces/Duration estimated_time_remaining
int16 number_of_recoveries
float32 distance_remaining
```


test_behavior_tree.xml
```xml
  <BehaviorTree ID="MainTree">
       <ComputePathToPose goal="{goal}" path="{path}" planner_id="{planner_id}"/>
      <FollowPath path="{path}"  controller_id="{controller_id}"/>
    </PipelineSequence>
  </BehaviorTree>
</root>
```

Also, when not using navigate_through_poses (for example) action, I thought it would be nice to be able to remove it like other costmaps, planner or controller plugin.
(I thought that if one action in the node disappeared, it would be more efficient than having an unused action server exist.)


* use navigate_through_poses action
param
Screenshot from 2022-07-12 11-04-09
action list
image

* not use navigate_through_poses action
param
image
action list
Screenshot from 2022-07-12 10-59-39

ladianchad:
Thanks for the good point

>That way you could change without even changing navigators or BT XMLs. Just mid-request or between requests call the service to change the blackboard ID of the algorithm. Seems almost like a better option actually.

Of course, if data is added to BT while the behavior tree is running, I am aware that creating a service bt node and a service server is a great solution.

However, if additional data is provided once for the first time requesting an action for navigation, the action client node (that requests the navigation action and have additional data) should design a service server for additional data, and process it through thread processing or topic pub sub. I often had to think about it.

In this situation, the simplest solution to solve this complexity was to change the action message and modify the navigator.

In addition, when data is exchanged with a service, if there is information necessary for navigation (suited to the application) as a specific application is designed, developer using navigation requests an action with an action client, and BT It is necessary to check whether the node that calls the service was added , and to check whether the service responds properly. I think this leads to developer's mistake (check service node in XML, servcie server died, service name correct etc.).

If you write the action msg according to this and make sure only the navigator that matches it, the developer puts the data in the request securely when writing the code and there is no additional management for it, I think it is more efficient. (less service server , less management)

>An action server doing nothing has negligible overhead. It would have a slight impact on system initialization, but at steady state during navigation use it would not be noticeable.

I think the concept of removing unused action servers is good in terms of efficiency (if not used, there is no significant effect), but I thought in the same context as the concept of removing unused actions from the navigation behavior server. I thought it fit a bit more with the design of this nav2.

Thanks :)

SteveMacenski:
> Of course, if data is added to BT while the behavior tree is running, I am aware that creating a service bt node and a service server is a great solution.

No, not quite. I meant adding a service to the BT navigator in general that allows you, outside of the behavior tree, to insert arbitrary things into the blackboard. The navigators can expose the service and be called whenever and then it'll be picked up by the running BT on the blackboard. 

Though, adding a BT node that is itself a service server for modifying blackboard variables isn't a bad idea either. Just not what I was specifically suggesting. 

Regardless, I've filed a ticket around that, I think its a good idea either way https://github.com/ros-planning/navigation2/issues/3074

> the simplest solution to solve this complexity was to change the action message and modify the navigator.

I don't disagree, if its always paired with the start of a request. 

OK, if you still think this is the right direction, I can get behind that.

nav2_bt_navigator/include/nav2_bt_navigator/bt_navigator.hpp
```diff
@@ -84,10 +84,20 @@ class BtNavigator : public nav2_util::LifecycleNode
    */
   nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
+  /**
+   * @brief Load Naigator plugins
+   * @param plugin_names bt plugin libaray names
+   * @return SUCCESS or FAILURE 
+   */
+  bool loadNavigatorPlugin(std::vector<std::string> plugin_names);
```
SteveMacenski, line 92:
```suggestion
  bool loadNavigatorPlugins(std::vector<std::string> plugin_names);
```

nav2_bringup/params/nav2_params.yaml
```diff
@@ -96,6 +96,14 @@ bt_navigator:
     - nav2_back_up_cancel_bt_node
     - nav2_drive_on_heading_cancel_bt_node
 
+    navigator: ["navigate_to_pose", "navigate_through_poses"]
```
SteveMacenski, line 99:
```suggestion
    navigators: ["navigate_to_pose", "navigate_through_poses"]
```

Update parameter to be plural

nav2_bt_navigator/include/nav2_bt_navigator/navigator.hpp
```diff
@@ -106,23 +106,121 @@ class NavigatorMuxer
   std::mutex mutex_;
 };
 
+class NavigatorBase
```
SteveMacenski, line 109: Why not have the `Navigator` class that already exists here be the base for pluginlib? There's no requirement that it has to be a pure virtual class.

SteveMacenski: This class looks almost entirely as a pass through anyway.

nav2_bt_navigator/include/nav2_bt_navigator/bt_navigator.hpp
```diff
@@ -84,10 +84,20 @@ class BtNavigator : public nav2_util::LifecycleNode
...
+
   // To handle all the BT related execution
-  std::unique_ptr<nav2_bt_navigator::Navigator<nav2_msgs::action::NavigateToPose>> pose_navigator_;
-  std::unique_ptr<nav2_bt_navigator::Navigator<nav2_msgs::action::NavigateThroughPoses>>
-  poses_navigator_;
+  // Navigator plugin
+  pluginlib::ClassLoader<NavigatorBase> navigator_class_loader_;
+  std::vector<pluginlib::UniquePtr<NavigatorBase>> navigators_;
+  std::vector<std::string> navigator_ids_;
+  std::vector<std::string> navigator_types_;
```
SteveMacenski, line 99: This isn't really used anywhere

SteveMacenski (changes requested):
I think for the `navigator_types_` / `navigator_ids_` you missed some usage of them to establish the default names and types if none are specified. RIght now, this would fail if a `yaml` file didn't contain the navigator plugins.

The IDs would be known from 

```

  const std::vector<std::string> navigators = {
    "navigate_to_pose",
    "navigate_through_poses"
  };
```

but the types that they represent would not, and therefore they would fail to load. See the planner server for examples of how to handle this https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L45-L46. 

I think we just need the class members for the name/type of the defaults, not the currently existing ones (unless you're going to log them, which we probably should once they're loaded the name/type of the loaded navigator plugins)

This is going to need some documentation updates too
- Migration guide update that this is now a plugin
- Plugins page with new table for the navigator plugins
- Adding the new parameters to the respective configuration guides 

For me to do
- Consider moving the navigator base class into `nav2_core` as a core interface
- Probably an update somewhere in the front-level documentation about this being a pluginized thing now  https://navigation.ros.org/concepts/index.html, https://navigation.ros.org/

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -194,6 +194,29 @@ BtNavigator::on_shutdown(const rclcpp_lifecycle::State & /*state*/)
...
+  feedback_utils.robot_frame = robot_frame_;
+  feedback_utils.transform_tolerance = transform_tolerance_;
+
+  for(std::string navigator_id: navigator_ids_){
+    std::string navigator_type = nav2_util::get_plugin_type_param(shared_from_this(), navigator_id);
+    navigators_.push_back(navigator_class_loader_.createUniqueInstance(navigator_type));
+    navigator_types_.push_back(navigator_type);
+    if(!navigators_.back()->on_configure(weak_from_this(), plugin_names, feedback_utils, &plugin_muxer_, odom_smoother_)){
+      return false;
+    }
```
SteveMacenski, line 214: Add a log with the name of the plugin and the type and that it has been loaded

nav2_bt_navigator/include/nav2_bt_navigator/navigator.hpp
ladianchad: Screenshot from 2022-07-14 09-55-53

I know that a base class with a template causes an error in the classloader.

SteveMacenski: Its possible, you just have to be careful. Here's an example of it working: https://github.com/artofnothingness/mppic/tree/b557139c48ae6dc75a06354e9e9001781e4b96ec

In this example, they template both the base function and the output critics. That's not necessary for our case.

SteveMacenski: Eh, but yeah the function for loading the plugins and the class loader object will need to be templated

ladianchad:
@SteveMacenski 
Thanks for thinking this is a good direction.
After seeing your feedback, I modified the code as follows:

* moved navigator to nav2_core (or do I move it back to bt_navigator?)
* Solve default plugin
> Its possible, you just have to be careful. Here's an example of it working: https://github.com/artofnothingness/mppic/tree/b557139c48ae6dc75a06354e9e9001781e4b96ec
> In this example, they template both the base function and the output critics. That's not necessary for our case.

But still the idea of a solution for NavigatorBase doesn't quite come up. Because when the class loader is templated, bt_navigator must also be templated.
I think NavigatorBase is an easier way to solve it. (the class interface used by bt_navigator depends only on NavigatorBase.)

Should this be fixed?

nav2_bringup/params/nav2_multirobot_params_1.yaml
```diff
@@ -95,7 +95,14 @@ bt_navigator:
     - nav2_spin_cancel_bt_node
     - nav2_back_up_cancel_bt_node
     - nav2_drive_on_heading_cancel_bt_node
-
+    navigators: ["navigate_to_pose",
```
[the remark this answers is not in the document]
SteveMacenski: Also put all this stuff at the very top so it can't be missed, probably just before `bt_loop_duration`

nav2_bt_navigator/include/nav2_bt_navigator/bt_navigator.hpp
```diff
@@ -84,11 +84,23 @@ class BtNavigator : public nav2_util::LifecycleNode
...
-  std::unique_ptr<nav2_bt_navigator::Navigator<nav2_msgs::action::NavigateThroughPoses>>
-  poses_navigator_;
-  nav2_bt_navigator::NavigatorMuxer plugin_muxer_;
+  // Navigator plugin
+  pluginlib::ClassLoader<nav2_core::NavigatorBase> navigator_class_loader_;
+  std::vector<pluginlib::UniquePtr<nav2_core::NavigatorBase>> navigators_;
+  std::vector<std::string> default_navigator_ids_;
+  std::vector<std::string> default_navigator_types_;
+  std::vector<std::string> navigator_ids_;
+  std::vector<std::string> navigator_types_;
```
SteveMacenski, line 101: I don't think `navigator_types_` needs to be stored. You only use it for logging, you can just create a `std::string` in that function to hold the current one

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -78,12 +87,23 @@ BtNavigator::BtNavigator(const rclcpp::NodeOptions & options)
     "nav2_back_up_cancel_bt_node"
     "nav2_drive_on_heading_cancel_bt_node"
   };
-
+
```
SteveMacenski, line 90: linting

nav2_core/include/nav2_core/navigator.hpp
```diff
@@ -106,23 +106,121 @@ class NavigatorMuxer
...
+  {
+    plugin_muxer_ = nullptr;
+  }
+
+  virtual bool on_configure(
+    rclcpp_lifecycle::LifecycleNode::WeakPtr parent_node,
+    const std::vector<std::string> &,
+    const FeedbackUtils & feedback_utils,
+    nav2_core::NavigatorMuxer * plugin_muxer,
+    std::shared_ptr<nav2_util::OdomSmoother>)
```
SteveMacenski, line 123: No use of odom smoother?

nav2_core/include/nav2_core/navigator.hpp
```diff
@@ -221,17 +317,9 @@ class Navigator
 
     bt_action_server_.reset();
 
-    return cleanup() && ok;
+    return NavigatorBase::cleanup() && ok;
```
SteveMacenski, line 320: Why calling the base version here? None of the other functions do this.

SteveMacenski (changes requested):
You have quite a few linting errors. Make sure to run `ament_cpplint` and `ament_uncrustify` and fix all issues

SteveMacenski:
I agree, just keep as is, that's fine

nav2_core/include/nav2_core/navigator.hpp
```diff
@@ -106,23 +106,121 @@ class NavigatorMuxer
...
+  {
+    return cleanup();
+  }
+
+protected:
+
+  /**
+   * @param Method to configure resources.
+   */
+  virtual bool configure(
```
SteveMacenski, line 178: Change all these to pure virtuals (e.g. no implementation `= 0`) as required implementations. Configure/cleanup/activate/deactivate

SteveMacenski (changes requested):

nav2_core/include/nav2_core/navigator.hpp
```diff
@@ -198,7 +294,7 @@ class Navigator
    * @brief Deactivation of the navigator's backend BT and actions
    * @return bool If successful
    */
-  bool on_deactivate()
+  bool on_deactivate() override
```
SteveMacenski, line 297: This is kind of a mess, why are we not calling configure / activate / deactivate / cleanup overrides that we established where the `NavigatorBase` does the `on_XYZ` variants? That's super confusing that there are both `configure` and `on_configure` and now 2 of the `on_` variants rather than using the `configure` set up and currently empty

SteveMacenski (changes requested):

nav2_core/include/nav2_core/navigator.hpp
SteveMacenski: E.g. `NavigatorBase` should have all the `on_XYZ` functions, then the `Navigator` should be implementing just the `XYZ` functions that are called as part of `on_XYZ` in the base class to preform work.

nav2_bringup/params/nav2_multirobot_params_1.yaml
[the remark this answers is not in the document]
ladianchad: > Also put all this stuff at the very top so it can't be missed, probably just before `bt_loop_duration`

I tried what you said :)

nav2_bt_navigator/include/nav2_bt_navigator/bt_navigator.hpp
ladianchad: I agree with you. so I removed at this commit.
https://github.com/ladianchad/navigation2/commit/59ced2b08eb0ce4a26f7c7afd7994c16e6c0dec8

nav2_bt_navigator/include/nav2_bt_navigator/navigators/navigate_to_pose.hpp
```diff
@@ -126,11 +123,8 @@ class NavigateToPoseNavigator
 
   std::string goal_blackboard_id_;
   std::string path_blackboard_id_;
-
-  // Odometry smoother object
-  std::shared_ptr<nav2_util::OdomSmoother> odom_smoother_;
```
SteveMacenski, line 131: Revert

nav2_bt_navigator/include/nav2_bt_navigator/navigators/navigate_through_poses.hpp
```diff
@@ -46,16 +46,14 @@ class NavigateThroughPosesNavigator
...
-  : Navigator() {}
+  : PoseNavigator() {}
 
+protected:
   /**
    * @brief A configure state transition to configure navigator's state
    * @param node Weakptr to the lifecycle node
-   * @param odom_smoother Object to get current smoothed robot's speed
    */
-  bool configure(
```
SteveMacenski, line 56: Revert

nav2_bt_navigator/src/navigators/navigate_to_pose.cpp
```diff
@@ -42,9 +40,6 @@ NavigateToPoseNavigator::configure(
 
   path_blackboard_id_ = node->get_parameter("path_blackboard_id").as_string();
 
-  // Odometry smoother object for getting current speed
-  odom_smoother_ = odom_smoother;
```
SteveMacenski, line 46: This is 100% required. Revert.

SteveMacenski (changes requested):

SteveMacenski:
You have a number of linting errors, please make sure to run `ament_cpplint` and `ament_uncrustify` before pushing commits and deal with the issues it mentions. 

I think its mostly good now, just some details. After these changes, I might have a few small things left, but largely this is good now!

> And since it's my first-time pr this to open source, I don't know how to deal with copyright.

There's nothing fancy; we don't have any required paperwork to assign copyright to the project. Technically any changes you make are under your own copyright, but most what you did was shuffling around code so there's no really any copyright headers to assign to you in this PR. In the future, if you create new files of code that your wrote not derivative of another file, then those can be assigned to you in the copyright header. But the header isn't the full story, technically the commit history stores what you changes / did and contains your copyright. In submitting the PR, you're agreeing to its perpetual use in this project.

ladianchad:
>There's nothing fancy; we don't have any required paperwork to assign copyright to the project. Technically any changes you make are under your own copyright, but most what you did was shuffling around code so there's no really any copyright headers to assign to you in this PR. In the future, if you create new files of code that your wrote not derivative of another file, then those can be assigned to you in the copyright header. But the header isn't the full story, technically the commit history stores what you changes / did and contains your copyright. In submitting the PR, you're agreeing to its perpetual use in this project.

Thanks for the detailed information. Thanks, I have studied a lot.

>Just Navigator is fine -- nothing about this strictly requires it to be a "pose" navigator. BTNavigator would be fine too
I changed the name from PoseNavigator to BTNavigator.
The existing BTNavigator class felt like it was managing the navigator like a behavior server, so I renamed it to NavigatorServer and modified its parameters. If these fixes affect anything else, I'll revert. Is it okay to edit like this?


>You have a number of linting errors, please make sure to run ament_cpplint and ament_uncrustify before pushing commits and deal with the issues it mentions.

sorry.
This time I checked that both passed.

Screenshot from 2022-07-19 09-40-02
Screenshot from 2022-07-19 09-48-33

Thank you for your always kind reply.

nav2_bringup/launch/navigation_launch.py
```diff
@@ -42,7 +42,7 @@ def generate_launch_description():
                        'smoother_server',
                        'planner_server',
                        'behavior_server',
-                       'bt_navigator',
+                       'navigator_server',
```
SteveMacenski, line 45: Revert. Please do not make unnecessary changes

SteveMacenski (changes requested):

nav2_bringup/launch/navigation_launch.py
SteveMacenski: You made a bunch of changes related to this; all of which must be reversed before I can review again

nav2_core/include/nav2_core/navigator.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+protected:
+  std::string current_navigator_;
+  std::mutex mutex_;
+};
+
+class Navigator
+{
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS(Navigator)
```
kfabian, line 110: Since this is a pure virtual class `RCLCPP_SMART_PTR_ALIASES_ONLY` has to be used.

ladianchad: I deleted ```RCLCPP_SMART_PTR_DEFINITIONS(Navigator)```  :)

ladianchad:
Is there anything I need to edit??

cmraaron:
This looks like a really useful PR, but is it held up because of documentation requirements?

If so, perhaps it can be split in two.  One PR to refactor so that the navigators inherit from a non-templated base class allowing bt_navigator to hold a list of base class pointers.  And hopefully that could get merged quickly as it would be only internal changes.

And the other moving the navigators to being plugins, and this would then become a much smaller PR, allowing for a more focused discussion.

ladianchad:
> This looks like a really useful PR, but is it held up because of documentation requirements?
> 
> If so, perhaps it can be split in two. One PR to refactor so that the navigators inherit from a non-templated base class allowing bt_navigator to hold a list of base class pointers. And hopefully that could get merged quickly as it would be only internal changes.
> 
> And the other moving the navigators to being plugins, and this would then become a much smaller PR, allowing for a more focused discussion.

Actually I don't know how this is going. your idea looks pretty good,
If my modified code is going to be merged, I think it can be reflected quickly.

@SteveMacenski Is there any action I need to take?

SteveMacenski:
There are a few of these larger PRs I have no had time yet to look over again from my PTO. I only got back from a month of vacation a few work days ago. It might yet be a week before I get to this, sorry. My internal priorities have changed and I can only really get to these in spare minutes.

SteveMacenski:
I had a little time to review this today and it with some fresh eyes after my vacation, its a good overall move, but I need some time to re-evaluate the general changes made to BT navigator in the past year or so. Things have gotten a little out of hand (largely to my own doing) and I need to figure out what the solution is to improve the principled design of the BT navigator before approving any other major changes to it. 

I see a number of still requested changes I would have based on the current state of this PR, so here's what I think is the most reasonable. Rather than frustrating @ladianchad with another few rounds of "change this, not that", I'm going to put this on my queue to personally finish based on @ladianchad's idea and largely the same implementation with some things shifted around / renamed. 

Ultimately, even if I don't get to it for another few weeks, it'll likely get done faster that way because I'll be able to just sit down and complete it in 1 sitting. I appreciate your patience and this action isn't reflective of your work, but rather my own maintenance process that let the BT Navigator get into a bit of a funky state where it needs some architectural attention.

ladianchad:
@SteveMacenski  Thank you for thinking positively about my ideas.
I know that there are many PRs on the navigation.
Rather than simply becoming a plugin as I did, it is very good and I believe that it will be well designed by synthesizing other opinions and making it a better design. If there is anything I can do to help with this, I'd love to help. Please let me know if there is anything like that.
Also, to give one more opinion, I was thinking of making the follow path customizable after the navigator is customized as a plugin. (I know that the follow path calls the action provided by bt navigator...).
I'm sure you're already considering these things, but this is the direction I was planning.

ladianchad:
It seems that this PR can now be closed.

PR #3076, opened 2022-07-14
Added real frequency to log 

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Own gazebo simulation |

---

* I added logging of the real achieved frequency when the target is not reached. Context: we were getting the "Control loop missed" warning but wanted to know at what rate we are running to know over time if this gets worse.

It's unfortunate that rclcpp::WallRate doesn't offer a way to get the achieved period/frequency when sleep returns fals

---

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -405,7 +405,8 @@ void ControllerServer::computeControl()
       if (!loop_rate.sleep()) {
         real_frequency = 1.0e9 / (steady_clock_.now() - begin).nanoseconds();
         RCLCPP_WARN(
-          get_logger(), "Control loop missed its desired rate of %.4fHz. Achieved rate: %.4fHz", controller_frequency_,
+          get_logger(), "Control loop missed its desired rate of %.4fHz. Achieved rate: %.4fHz",
+          controller_frequency_,
           real_frequency);
       }
     }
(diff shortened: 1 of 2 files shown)


tonynajjar:
image

I sometimes get frequencies higher than the target one. Any idea why? Maybe because I'm using a separate clock?

SteveMacenski (approved):

nav2_controller/src/controller_server.cpp
```diff
@@ -400,9 +403,14 @@ void ControllerServer::computeControl()
       }
 
       if (!loop_rate.sleep()) {
-        RCLCPP_WARN(
-          get_logger(), "Control loop missed its desired rate of %.4fHz",
-          controller_frequency_);
+        real_frequency =
```
SteveMacenski, line 406: This does not appear to be following the styling of the project (I see indents with 4 spaces, things being aligned pythonically, etc. 

Please lint this with ament_uncrustify / ament_cpplint for styling

SteveMacenski (changes requested):

tonynajjar:
on it

tonynajjar:
I applied all your suggestions but still same result...
image

I commited the version with the ROS clock. I'm still investigating

SteveMacenski:
It doesn't seem like this is something we should merge if it doesn't strictly speaking... work :frowning:

tonynajjar:
I actually agree, unless the wallrate is actually unstable and this only exposes that. I doubt it though, will investigate a bit more tomorrow

tonynajjar:
I can't find what's wrong and can't spend more time on it. I'm fine to close this PR if you want

SteveMacenski:
OK :frowning_face: I understand.

Issue #3078, opened 2022-07-17
Issue loading saved map when saved under a subdirectory


nakai-omer:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble
- Version or commit hash:
  - ros-humble-navigation2  1.1.0-2jammy.20220607.182532
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
Call the map saver server with a map_url that includes a subdirectory: 
```
ros2 service call /map_saver/save_map nav2_msgs/srv/SaveMap "{map_topic: map, map_url: maps/nakai_world, image_format: pgm, map_mode: trinary, free_thresh: 0.25, occupied_thresh: 0.65}"
```

#### Expected behavior
Output yaml file image file will have a relative path to the yaml file, and not to current working directory. Conversely, when loading the map, map_server should not append the yaml file's path to the image path, rather it should append the currect working directory to the image path. 

#### Actual behavior
Image path inside the map's yaml file, is relative to current working directory and not to the yaml file. 
```
image: maps/nakai_world.pgm 
mode: trinary
resolution: 0.05
origin: [-9.39e-05, -0.68, 0]
negate: 0
occupied_thresh: 0.65
free_thresh: 0.25
```

#### Additional information
When trying to later load the map using map_server, getting the following error: 
```
[ERROR] [map_io]: Failed to load image file ./maps/maps/nakai_world.pgm for reason: Magick: Unable to open file (./maps/maps/nakai_world.pgm) reported by magick/blob.c:3089 (OpenBlob)
```
All commands were ran under the same working directory.

SteveMacenski:
@AlexeyMerzlyakov please add this to your queue.

> Output yaml file image file will have a relative path to the yaml file, and not to current working directory.

The actual command is going to be relative to your requested location. Right, if you give a relative location, it'll be relative to where you called it. So assuming when you call `SaveMap` with a relative path for the yaml of `./my_relative/path`, you're saying that the YAML is correct, yes? Its just the image file that is wrong? 

> Conversely, when loading the map, map_server should not append the yaml file's path to the image path, rather it should append the currect working directory to the image path.

It sounds like the `image: maps/nakai_world.pgm` needs to just me `nakai_world.pgm` and the file itself is stored in the relative directory `maps`. Is that what you're looking for? I agree that that would be sensible. 

[...]

PR #3082, opened 2022-07-20
Fix for #3078, fix image path in YAML

nakai-omer:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3078  |
| Robotic platform tested on | Own robot |

---

* Extraced only the image file name instead of the whole path when saving the meta data YAML file. 

N/A

---

I wanted to avoid having to add another dependency, but it could also be done using std::filesystem, like here: https://en.cppreference.com/w/cpp/filesystem/path/filename

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/src/map_io.cpp
+++ b/nav2_map_server/src/map_io.cpp
@@ -8,7 +8,7 @@
  * modification, are permitted provided that the following conditions are met:
  *
  *     * Redistributions of source code must retain the above copyright
- *       notice, this list of conditions and the following disclaimer.
+ *       notice, this list of conditions and the following disclaimer.151
  *     * Redistributions in binary form must reproduce the above copyright
  *       notice, this list of conditions and the following disclaimer in the
  *       documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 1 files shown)


SteveMacenski (changes requested):

SteveMacenski:
I'm asking @AlexeyMerzlyakov to take charge on this one since I leave for PTO in a short period of time. Feel free to merge once you're happy with it and close the ticket.

nav2_map_server/src/map_io.cpp
```diff
@@ -504,11 +504,14 @@ void tryWriteMapToFile(
     tf2::Matrix3x3 mat(tf2::Quaternion(orientation.x, orientation.y, orientation.z, orientation.w));
     double yaw, pitch, roll;
     mat.getEulerYPR(yaw, pitch, roll);
+    
+    int file_name_index = mapdatafile.find_last_of("/\\");
```
AlexeyMerzlyakov, line 508: Please change `int` -> `const int` to intentionally tell the compiler that this variable could be optimized out

AlexeyMerzlyakov (commented):
The `find_last_of("/\\");` will find the last entrance of both Linux `/` and Windows `\` in file path, so it will work. During the saving map through MapIO library, saved image file name will now always be relative (the same) to the YAML directory, which will also provide the portability of such saved maps.

Please correct Steve's comment and one nit from my side; and this should be OK to go.

AlexeyMerzlyakov (approved):
LGTM!

SteveMacenski (approved):

PR #3084, opened 2022-07-21
build fix - constructor fix for collision monitor

padhupradheep:
small fix


AlexeyMerzlyakov (approved):
Should work according to the latest changes (30048b46bf02c6a60473b1dae6000e6eb2597056)

SteveMacenski (approved):

PR #3087, opened 2022-07-22
Use correct timeout when checking future in bt_servcie_node

samiamlabs:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Custom simulation setup |

---

* I noticed that service calls from the tree using bt_servcie_node blocks in the tick() instead of returning RUNNING
* The reason appears to be a typo, and my understanding is that the unused `timeout` variable was supposed to be used instead of `server_timeout_`.

#### For Maintainers: 


SteveMacenski (approved):
Thanks! Agreed

PR #2982, opened 2022-05-31
Collision monitor

AlexeyMerzlyakov:
Implementation of Collision Monitor node

---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | #1899 |
| Primary OS tested on | Ubuntu 20.04 with ROS2 Rolling built from sources |
| Robotic platform tested on | Gazebo simulation of Turtlebot3 |

---

## Description of contribution in a few bullet points

* New Collision Monitor (formerly Collision Safety) node
* Available 3 modes: Stop, Slowdown and Approach.
  - Stop mode: the robot will be stopped if more than `N` points are inside given polygon
  - Slowdown mode: the robot will be slow down on `S%` if more than `N` points are inside given polygon. Working principle similar to Stop mode
  - Approach mode: the robot reduces its speed to keep constant `M` seconds before a collision with obstacle. In this model, given polygon is acting as robot's footprint.
* Two types of shapes are supported: polygon and circle.
* Two types of inputs are supported: laser scanner and pointcloud
* Each mode is attached/connected to each shape. Although all 3 conditions are mutually exclusive, nobody restricts to set all modes' polygons simultaneously. The priority in this case will be given to the safest mode when two or more activated (The safest mode is considered as Stop. If there is no stop required, the mode producing the lowest speed will be activated.)
* Collision Monitor is acting as independent node, laying "under" Nav2 stack. There are remapped input and output `cmd_vel` required for node operation.

## Description of documentation updates required from your changes

* Node parameters description page
* Collision Monitor usage tutorial
* New feature in Navigation Concepts ?

---

## Future work that may be required in bullet points

* Cover Collision Monitor by tests
* Cover the code by comments / better explaining what is going on
* Performance optimizations for Approach mode
* Increasing the stability of Approach mode when the `odom` speed changes in steps/making noise

#### For Maintainers: 
- [ ] Check that any new parameters added are updated in navigation.ros.org
- [ ] Check that any significant change is added to the migration guide
- [ ] Check that any new features **OR** changes to existing behaviors are reflected in the tuning guide
- [ ] Check that any new functions have Doxygen added
- [ ] Check that any new features have test coverage
- [ ] Check that any new plugins is added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

PR #3099, opened 2022-07-29
Add the support of range sensors to Collision Monitor

AlexeyMerzlyakov:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/pull/2982 |
| Robotic platform tested on | colcon test |

---

Adding the range sensors support to the Collision Monitor, requested in https://github.com/ros-planning/navigation2/pull/2982#issuecomment-1150548172

* Added the support of IR/Sonar `sensor_msgs/Range` sensors to Collision Monitor
* Adjusted testcases for new type of sensors

* Adjust README.md. [No further updates needed](https://navigation.ros.org/configuration/packages/configuring-collision-monitor.html?highlight=range) for official pages documentation.

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/range.cpp
+++ b/nav2_collision_monitor/src/range.cpp
@@ -12,15 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "nav2_collision_monitor/scan.hpp"
+#include "nav2_collision_monitor/range.hpp"
 
+#include <math.h>
 #include <cmath>
 #include <functional>
+
+#include "nav2_util/node_utils.hpp"
 
 namespace nav2_collision_monitor
 {
 
-Scan::Scan(
+Range::Range(
   const nav2_util::LifecycleNode::WeakPtr & node,
   const std::string & source_name,
   const std::shared_ptr<tf2_ros::Buffer> tf_buffer,
@@ -33,16 +36,16 @@ Scan::Scan(
     transform_tolerance, source_timeout),
   data_(nullptr)
 {
-  RCLCPP_INFO(logger_, "[%s]: Creating Scan", source_name_.c_str());
+  RCLCPP_INFO(logger_, "[%s]: Creating Range", source_name_.c_str());
 }
 
-Scan::~Scan()
+Range::~Range()
 {
-  RCLCPP_INFO(logger_, "[%s]: Destroying Scan", source_name_.c_str());
+  RCLCPP_INFO(logger_, "[%s]: Destroying Range", source_name_.c_str());
   data_sub_.reset();
 }
 
-void Scan::configure()
+void Range::configure()
 {
   auto node = node_.lock();
   if (!node) {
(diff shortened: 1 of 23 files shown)


AlexeyMerzlyakov:
This is the first attempt to add range sensors support to the Collision Monitor.
The range arc `{-field_of_view/2  .. field_of_view/2}` from `sensor_msgs/msg/Range` message is being breaked down into equidistant (by angle) points as shown at the illustration below (obstacle angle is being set as ROS-parameter):
range_sensor
Robot simulation testing is required.

nav2_collision_monitor/src/range.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if (data_ == nullptr) {
+    return;
+  }
+  if (!sourceValid(data_->header.stamp, curr_time)) {
+    return;
+  }
+
+  // Ignore data, if its range is out of scope of range sensor abilities
+  if (data_->range < data_->min_range || data_->range > data_->max_range) {
+    RCLCPP_INFO(
```
SteveMacenski, line 80: Warning rather than an info

nav2_collision_monitor/src/range.cpp
```diff
@@ -0,0 +1,131 @@
...
+  tf2::Transform tf_transform;
+  if (!getTransform(data_->header.frame_id, data_->header.stamp, curr_time, tf_transform)) {
+    return;
+  }
+
+  // Calculate poses and refill data array
+  for (
+    float angle = -data_->field_of_view / 2;
+    angle <= data_->field_of_view / 2;
+    angle += obstacles_angle_)
```
SteveMacenski, line 98: I'm not entirely following this obstacles angle parameter, what is this exactly, and why the default that is set?

AlexeyMerzlyakov: Yes, indeed this is an issue. Thanks for noting this. [Added](https://github.com/ros-planning/navigation2/pull/3099/files#diff-8ba7a458af64954e48901b5091ebc91d61f5b592235c046e668b4e2ff82de61eR112-R123) the same code to the end of loop to check upper edge case (and corrected testcase accordingly).

nav2_collision_monitor/src/range.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3099/files#diff-8ba7a458af64954e48901b5091ebc91d61f5b592235c046e668b4e2ff82de61eR80

nav2_collision_monitor/src/range.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <math.h>
+#include <cmath>
+#include <functional>
+
+#include "nav2_util/node_utils.hpp"
+
+namespace nav2_collision_monitor
+{
+
+Range::Range(
```
SteveMacenski, line 26: The tutorial should be updated that it can support Range sensors too. We make a passive remark about the fact that we use Scans but PC2 is also possible -- we should also mention Range/Sonar/IR

SteveMacenski (approved):

SteveMacenski:
Ran the CI jobs in #3154 and all checked out!

PR #3100, opened 2022-08-03
Staticial replanning

### LINKED ISSUE #2880: IsPathValid Improvement: Change in Cost
Have an option to have the paths return their path costs and be able to use that with a replan to compare if the cost have changed "sufficiently" to a parameterized setting to trigger a replan, rather than only when exactly invalid due to collision.

Allows for paths that have become much more expensive to be replanned based on a user-defined metric

Issues to work through:
- If we have the old path and the new path is shorter since it was replanned after some distance was traveled, how to compare these paths "fairly"?  
- Probably involves storing each point-costs in the path and using only the still relevant costs
- That would probably involve a change to the ComputePathToPose API
- Where to store this information? In the planner server? That would make it stateful. If the BT navigator? Probably the best option but where? Probably the blackboard
-  What's the metric we want to use? I would suggest a percentage change potentially.

jwallace42:
---

## Basic Info

Adds additional functionality to determine if a path is invalid using a two sample z test. 

| ------ | ----------- |
| Ticket(s) this addresses   | #2880 |

---

Adds functionality to `is path valid service` to determine if there has been a significant change in cost. This pr closes #2880.

Planner server need to be updated with a new parameter. 

---

## Future work that may be required in bullet points\
We could make this a plugin... 

#### For Maintainers: 


jwallace42:
Currently block by https://github.com/ros2/common_interfaces/pull/187.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
```diff
@@ -58,7 +58,8 @@ class IsPathValidCondition : public BT::ConditionNode
   {
     return {
       BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
-      BT::InputPort<std::chrono::milliseconds>("server_timeout")
+      BT::InputPort<std::chrono::milliseconds>("server_timeout"),
+      BT::InputPort<bool>("consider_cost_change", true, "Consider cost changes")
```
SteveMacenski, line 62: Add to groot XML for people using it to build trees to have the option exposed

nav2_planner/src/planner_server.cpp
```diff
@@ -52,6 +53,12 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
   declare_parameter("planner_plugins", default_ids_);
   declare_parameter("expected_planner_frequency", 1.0);
 
+  nav2_util::declare_parameter_if_not_declared(
+    this, "z_score", rclcpp::ParameterValue(2.55));
+
+
+  get_parameter("z_score", z_score_);
```
SteveMacenski, line 60: In the parameter / variable name, it would be nice to have some more context to what this is used for. While it is a Z-score, that's not really telling a user why its there / where its used. `cost_change_z_score` or `path_difference_z_score` or something would help give some context where/how its used. 

Considering the use of `consider_cost_change` in the action field, I think `cost_change_z_score` or `path_cost_change_score` or something would make sense

SteveMacenski (changes requested):
Otherwise LGTM, I think the biggest part of this is testing / validation of the functionality of the property. 

The only other thing is maybe breaking out the calculation of std / means into lambdas or header utility functions just to make the inline code a bit more straight forward to read.

nav2_planner/src/planner_server.cpp
SteveMacenski: Also parameter guide stuff :-)

jwallace42:
Testing Notes from Steve: 

> Mostly its just some basic test cases to understand how much change will cause a trigger in each situation with the parameter set to see if there's a configuration z_score that would be sensible to work decently well with all of them to know that the feature meets our intent. Ex.
- Have a path that wraps a corner around a shelf from 1 aisle to another. How much increased cost due to sensor noise triggers the update. Basically I want to know if its robust against minor noise when close up to obstacles but reactive to major changes. 
- Path a long path, lets even say long, straight, in free space. If we have some human-sized circle, how/when does it trigger a response?  Basically I want to know for localized changes on a larger path, it will trigger reasonably with the same set of parameters as with small global changes (next experiment)
- Path is a realistic path in some space in a costmap field. How does localization errors / updates impact triggers?
- Path is a realistic path in some space in a costmap field. How does costmap parameterizations of the inflation layer impact things?
- Path is a realistic path in some space in a costmap field. How does small but globalized changes impact replanning triggers?
- Is it sensible to measure the changes based on the cost values or the distance from obstacles (e.g. since we can back out the distance to obstacle if we have the cost with an inflation layer)

If there's some single value of z_score that does well in all of these situations, then we're good to go. If not, then that's a bit problematic since it wouldn't properly trigger replanning in the basic situations where replanning should/shouldn't be triggered

PR #3110, opened 2022-08-12
Log level

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Adds ability to set log level for nodes that are launched. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/bringup_launch.py
+++ b/nav2_bringup/launch/bringup_launch.py
@@ -42,6 +42,7 @@ def generate_launch_description():
     autostart = LaunchConfiguration('autostart')
     use_composition = LaunchConfiguration('use_composition')
     use_respawn = LaunchConfiguration('use_respawn')
+    log_level = LaunchConfiguration('log_level')
 
     # Map fully qualified names to relative ones so the node's namespace can be prepended.
     # In case of the transforms (tf), currently, there doesn't seem to be a better alternative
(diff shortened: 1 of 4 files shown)


SteveMacenski (changes requested):
I think this misses some stuff, namely the composition nodes in localization / navigation / etc

jwallace42:
So, 

you can't set the log level of individual components :( only the container that holds the components. I guess this makes sense but it would be nice to be able to set the log levels of each component.

SteveMacenski (approved):

SteveMacenski:
@jwallace42 rerunning CI, an unusual failure occurred. I don't think its your fault, but want to just check again.

SteveMacenski:
Can you rebase on main? I'm seeing that same error again which I'm not seeing on other PRs in some utils action server unit tests.

SteveMacenski:
The PR got a little messed up, its showing a bunch of other commits.

jwallace42:
So, 

I think this fixed it. I have to add a pr for some linting fixes that didn't get in on https://github.com/ros-planning/navigation2/pull/3114.

tonynajjar:
What's the main reason to implement this? I'd imagine to see the debug logs? Running it in debug mode is unreadable because of all the "sub-loggers" like rcl

SteveMacenski:
I didn’t ask, it seemed like a reasonable request to expose that as a CLI option

tonynajjar:
Alright, to have useful readable debug logs only from Nav2, I'd recommend looking into changing the severity from individual loggers. It's available since galactic

https://docs.ros.org/en/galactic/Tutorials/Demos/Logging-and-logger-configuration.html

PR #3113, opened 2022-08-14
add bt log file

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Added bt logger file to log transitions in a separate file from ros debug logs. Currently, the transitions can be see by setting the log level to debug. However, this spams the logs making it unreadable. Instead, adding the transitions to a separate file will allow users to provide useful log files to developers. For example, if a bug come up the user attaches the ros logs and the bt transition logs. The pr uses the file logger specified here https://www.behaviortree.dev/tutorial_05_subtrees/. 

ex: 

[1660495514.917637]: NavigateRecovery          IDLE    -> RUNNING
[1660495514.917656]:       RateController      IDLE    -> RUNNING
[1660495514.917658]:          ComputePathToPoseRecovery IDLE    -> RUNNING
[1660495514.917662]:             ComputePathToPose IDLE    -> RUNNING
[1660495514.917852]:    NavigateWithReplanning IDLE    -> RUNNING

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -179,6 +179,7 @@ bool BtActionServer<ActionT>::loadBehaviorTree(const std::string & bt_xml_filena
     return false;
   }
 
+  logger_file_ = std::make_unique<BT::FileLogger>(tree_, "bt_trace.fbl");
   topic_logger_ = std::make_unique<RosTopicLogger>(client_node_, tree_);
 
   current_bt_xml_filename_ = filename;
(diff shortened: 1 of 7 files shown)


padhupradheep (approved):

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -179,6 +180,7 @@ bool BtActionServer<ActionT>::loadBehaviorTree(const std::string & bt_xml_filena
     return false;
   }
 
+  logger_file_ = std::make_unique<BT::FileLogger>(tree_, "bt_trace.fbl");
```
SteveMacenski, line 183: A few questions:
- Should this be timestamped or somehow associated with a request? E.g. should we append to the end of the base name the timestamp so we don't override previous ones or maybe use the goal ID hash?
- Should this be optional so that the comment above doesn't generate a ton of files in the users' running directory, or allow the user to specify a directory to save these in?

SteveMacenski:
It seems generally like a good idea!

jwallace42:
So,

I did a bit of thinking and we can trash this pr. The user can launch everything with the log levels set to debug and we can use rqt console to filter out results to just look at the bt transition logs to find any bugs :).

Issue #3090, opened 2022-07-24
behavior tree "..only_if_path_becomes_invalid.xml" leads to navigation on cancelled goal


automech-rb:
## Bug report

**Required Info:**

- Operating System:
  - Docker 20.04 Ubuntu
- ROS2 Version:
  - rolling
- Version or commit hash:
  - latest 104bdc51fb72e05af4ccb4fb96a2fd421239f016
- DDS implementation:
  - rmw_cyclonedds_cpp

#### Steps to reproduce issue

```
Use SMAC lattice planner and RPP
Send a goal which is easily (quickly) calculable by SMAC lattice
Cancel the above goal
Send a goal which is difficult to calculate (takes time)
```

#### Expected behavior
Local planner (controller) waits until new path is calculated.

#### Actual behavior
Controller navigates on old path which is already canceled.

#### Additional information
![bug](https://user-images.githubusercontent.com/24780015/180637139-99cc2b07-d8e8-4819-a9c4-651b33513da9.gif)

jwallace42:
How often does this error occur for you?

automech-rb:
HI @jwallace42 

This error is very repeatable and occurs every time when a current navigation is cancelled and next goal is difficult or non-navigable (very close to obstacle). It is also very easy to emulate on any map (small or big). For a rectangular footprint robot, set goal pose very near to obstacle facing into it.

jwallace42:
Does this only happen with SMAC and RRP? I ran with the default planners for about 10 minutes and couldn't reproduce your error.

automech-rb:
Yeah, that is what I think. I didn't saw this error with other planners. 
I will try again tomorrow to check with others (smac Hybrid-A* Planner and Navfn)

jwallace42:
Okay, 

That is pretty odd. I would assume this issue would be for any combination of planner and controller. Let me know how the testing goes.

automech-rb:
Hi @jwallace42 

This time I used publicly available [Amazon warehouse](https://github.com/aws-robotics/aws-robomaker-small-warehouse-world) with gazebo. Here are the results:

- SMAC Lattice + RPP :: `Leads to error`
- SMAC Lattice + DWB :: ` Leads to error`
- Navfn + RPP ::  `Leads to error`

[...]

PR #3114, opened 2022-08-14
ComputePathToPose Sets empty path on blackboard if action is aborted or cancelled

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3090 |

---

This pr sets the path to be empty on the blackboard if compute pose to path fails. It addresses the issue with the behavior tree navigating along a old path using https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_only_if_path_becomes_invalid.xml. 

The new behavior will be
    Tries to compute path, fails clears costmap
    Recovery node ticks again, compute path does get ticked again because isPathValid == FAILURE due to the path being empty.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
+++ b/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
@@ -43,6 +43,20 @@ BT::NodeStatus ComputePathToPoseAction::on_success()
   return BT::NodeStatus::SUCCESS;
 }
 
+BT::NodeStatus ComputePathToPoseAction::on_aborted()
+{
+  geometry_msgs::msg::PoseStamped empty_path;
+  setOutput("path", empty_path);
+  return BT::NodeStatus::FAILURE;
+}
+
+BT::NodeStatus ComputePathToPoseAction::on_cancelled()
+{
+  geometry_msgs::msg::PoseStamped empty_path;
+  setOutput("path", empty_path);
+  return BT::NodeStatus::SUCCESS;
+}
+
 }  // namespace nav2_behavior_tree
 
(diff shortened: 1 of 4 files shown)


jwallace42:
@automech-rb this should fix the issue you posted. If you get a chance to run it that would be great :).

SteveMacenski (approved):
Oh good idea, can you also do the `ComputePathThroughPoses` too?

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -43,6 +43,20 @@ BT::NodeStatus ComputePathToPoseAction::on_success()
   return BT::NodeStatus::SUCCESS;
 }
 
+BT::NodeStatus ComputePathToPoseAction::on_aborted()
+{
+  geometry_msgs::msg::PoseStamped empty_path;
```
SteveMacenski, line 48: Wrong type :wink: This should be a Path not a pose stamped. 

https://app.circleci.com/pipelines/github/ros-planning/navigation2/7718/workflows/ff4e7675-61ac-482f-baad-bc0d3fde29ec/jobs/26788

CI failed on this (also @facontidavide great user protective error! That saved us a bunch of time and energy)

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
facontidavide: you are welcome :)

SteveMacenski (approved):

PR #3118, opened 2022-08-18
Fix Costmap Filters system tests

AlexeyMerzlyakov:
Continue PR of #3098 (fixes failing in CI `test_keepout_filter`, `test_speed_filter_global`, `test_speed_filter_local tests`)

#### For Maintainers: 


SteveMacenski:
Ah same thing

Issue #3119, opened 2022-08-18
Replace config use_sim_time with SetUseSimTime in launch file


tonynajjar:
Small enhancement: you could replace all the use_sim_time in the config with the [SetUseSimTime](https://github.com/ros2/launch_ros/blob/humble/launch_ros/launch_ros/actions/set_use_sim_time.py) introduced in Humble. That would remove the RewrittenYaml logic from the launch file

One thing to note is that bt_navigator_navigate_through_poses_rclcpp_node won't get this parameter because it's created internally. Here is a workaround in bt_action_server_impl.h: 

```
  auto options = rclcpp::NodeOptions().arguments(
    {"--ros-args",
      "-r",
      std::string("__node:=") + std::string(node->get_name()) + client_node_name + "_rclcpp_node",
      "-p",
      "use_sim_time:=" + std::string(node->get_parameter("use_sim_time").as_bool() ? "true": "false"),
      "--"});
```

SteveMacenski:
Can you submit a PR :heart: ?

tonynajjar:
Here is a draft https://github.com/ros-planning/navigation2/pull/3131

PR #3131, opened 2022-08-21
use_sim_time refactoring

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3119) |

---

* I used SetParameter to set the use_sim_time parameter and removed the parameter from launch files

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -83,7 +83,8 @@ bool BtActionServer<ActionT>::on_configure()
       std::string("__node:=") +
       std::string(node->get_name()) + "_" + client_node_name + "_rclcpp_node",
       "-p",
-      "use_sim_time:=" + std::string(node->get_parameter("use_sim_time").as_bool() ? "true": "false"),
+      "use_sim_time:=" +
+      std::string(node->get_parameter("use_sim_time").as_bool() ? "true" : "false"),
       "--"});
 
   // Support for handling the topic-based goal pose from rviz
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -82,6 +82,8 @@ bool BtActionServer<ActionT>::on_configure()
       "-r",
       std::string("__node:=") +
       std::string(node->get_name()) + "_" + client_node_name + "_rclcpp_node",
+      "-p",
+      "use_sim_time:=" + std::string(node->get_parameter("use_sim_time").as_bool() ? "true": "false"),
       "--"});
 
   // Support for handling the topic-based goal pose from rviz
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -20,7 +20,7 @@ from launch import LaunchDescription
 from launch.actions import DeclareLaunchArgument, GroupAction, SetEnvironmentVariable
 from launch.conditions import IfCondition
 from launch.substitutions import LaunchConfiguration, PythonExpression
-from launch_ros.actions import LoadComposableNodes
+from launch_ros.actions import LoadComposableNodes, SetParameter
 from launch_ros.actions import Node
 from launch_ros.descriptions import ComposableNode
 from nav2_common.launch import RewrittenYaml
@@ -101,6 +101,7 @@ def generate_launch_description():
     load_nodes = GroupAction(
         condition=IfCondition(PythonExpression(['not ', use_composition])),
         actions=[
+            SetParameter("use_sim_time", use_sim_time),
             Node(
                 package='nav2_map_server',
                 executable='map_server',
@@ -129,8 +130,10 @@ def generate_launch_description():
         ]
     )
 
-    load_composable_nodes = LoadComposableNodes(
+    load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
+        actions=[SetParameter("use_sim_time", use_sim_time),
+            LoadComposableNodes(
         target_container=container_name,
         composable_node_descriptions=[
             ComposableNode(
@@ -152,7 +155,7 @@ def generate_launch_description():
                 parameters=[{'autostart': autostart,
                              'node_names': lifecycle_nodes}]),
         ],
-    )
+    )])
 
     # Create the launch description and populate
     ld = LaunchDescription()
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -52,7 +52,6 @@ def generate_launch_description():
 
     # Create our own temporary YAML files that include substitutions
     param_substitutions = {
-        'use_sim_time': use_sim_time,
         'yaml_filename': map_yaml_file}
 
     configured_params = RewrittenYaml(
@@ -125,8 +124,7 @@ def generate_launch_description():
                 executable='lifecycle_manager',
                 name='lifecycle_manager_localization',
                 output='screen',
-                parameters=[{'use_sim_time': use_sim_time},
-                            {'autostart': autostart},
+                parameters=[{'autostart': autostart},
                             {'node_names': lifecycle_nodes}])
         ]
     )
@@ -151,8 +149,7 @@ def generate_launch_description():
                 package='nav2_lifecycle_manager',
                 plugin='nav2_lifecycle_manager::LifecycleManager',
                 name='lifecycle_manager_localization',
-                parameters=[{'use_sim_time': use_sim_time,
-                             'autostart': autostart,
+                parameters=[{'autostart': autostart,
                              'node_names': lifecycle_nodes}]),
         ],
     )
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -157,8 +157,8 @@ def generate_launch_description():
                                     'node_names': lifecycle_nodes}]),
                 ],
             )
-            ]
-        )
+        ]
+    )
 
     # Create the launch description and populate
     ld = LaunchDescription()
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -132,30 +132,33 @@ def generate_launch_description():
 
     load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
-        actions=[SetParameter("use_sim_time", use_sim_time),
+        actions=[
+            SetParameter("use_sim_time", use_sim_time),
             LoadComposableNodes(
-        target_container=container_name,
-        composable_node_descriptions=[
-            ComposableNode(
-                package='nav2_map_server',
-                plugin='nav2_map_server::MapServer',
-                name='map_server',
-                parameters=[configured_params],
-                remappings=remappings),
-            ComposableNode(
-                package='nav2_amcl',
-                plugin='nav2_amcl::AmclNode',
-                name='amcl',
-                parameters=[configured_params],
-                remappings=remappings),
-            ComposableNode(
-                package='nav2_lifecycle_manager',
-                plugin='nav2_lifecycle_manager::LifecycleManager',
-                name='lifecycle_manager_localization',
-                parameters=[{'autostart': autostart,
-                             'node_names': lifecycle_nodes}]),
-        ],
-    )])
+                target_container=container_name,
+                composable_node_descriptions=[
+                    ComposableNode(
+                        package='nav2_map_server',
+                        plugin='nav2_map_server::MapServer',
+                        name='map_server',
+                        parameters=[configured_params],
+                        remappings=remappings),
+                    ComposableNode(
+                        package='nav2_amcl',
+                        plugin='nav2_amcl::AmclNode',
+                        name='amcl',
+                        parameters=[configured_params],
+                        remappings=remappings),
+                    ComposableNode(
+                        package='nav2_lifecycle_manager',
+                        plugin='nav2_lifecycle_manager::LifecycleManager',
+                        name='lifecycle_manager_localization',
+                        parameters=[{'autostart': autostart,
+                                    'node_names': lifecycle_nodes}]),
+                ],
+            )
+            ]
+        )
 
     # Create the launch description and populate
     ld = LaunchDescription()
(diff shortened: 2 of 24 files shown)


nav2_bringup/launch/localization_launch.py
```diff
@@ -102,6 +101,7 @@ def generate_launch_description():
     load_nodes = GroupAction(
         condition=IfCondition(PythonExpression(['not ', use_composition])),
         actions=[
+            SetParameter("use_sim_time", use_sim_time),
```
tonynajjar, line 104: I first expected to use `SetUseSimTime(True)` but that didn't work. Upon further inspection of the [Action](https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/actions/set_use_sim_time.py#L28), I don't think it's meant to set the parameter for all the nodes in the scope. I'm not really sure what it's for.

nav2_bringup/launch/localization_launch.py
```diff
@@ -125,37 +125,40 @@ def generate_launch_description():
...
-                parameters=[{'use_sim_time': use_sim_time},
-                            {'autostart': autostart},
+                parameters=[{'autostart': autostart},
                             {'node_names': lifecycle_nodes}])
         ]
     )
 
-    load_composable_nodes = LoadComposableNodes(
+    load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
```
tonynajjar, line 134: I would think that using the `condition` of LoadComposableNodes or GroupAction is the same

nav2_bringup/launch/slam_launch.py
```diff
@@ -42,13 +42,10 @@ def generate_launch_description():
...
 
     # Create our own temporary YAML files that include substitutions
-    param_substitutions = {
-        'use_sim_time': use_sim_time}
-
     configured_params = RewrittenYaml(
         source_file=params_file,
         root_key=namespace,
-        param_rewrites=param_substitutions,
+        param_rewrites={},
```
tonynajjar, line 48: I think we still need the RewrittenYaml here for setting the namespace?

nav2_bringup/launch/localization_launch.py
SteveMacenski, line 104: does SetParameter work globally or can it be used for a single node? Also, does that work if the parameter in question is inside of the yaml file also (clearly it works if its not inside the yaml file)? I ask because this could be a really nice solution for the `yaml_filename` thing too if we can set it for only the map server if there is a non-trivial value set in the launch configuration.

nav2_bringup/launch/slam_launch.py
SteveMacenski: yes

nav2_controller/src/controller_server.cpp
```diff
@@ -63,7 +63,10 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
     "local_costmap", std::string{get_namespace()}, "local_costmap");
-
+  costmap_ros_->set_parameter(
```
SteveMacenski, line 66: ?

nav2_planner/src/planner_server.cpp
```diff
@@ -62,7 +62,10 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
   // Setup the global costmap
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
     "global_costmap", std::string{get_namespace()}, "global_costmap");
-
+  costmap_ros_->set_parameter(
```
SteveMacenski, line 65: ?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -82,6 +82,9 @@ bool BtActionServer<ActionT>::on_configure()
       "-r",
       std::string("__node:=") +
       std::string(node->get_name()) + "_" + client_node_name + "_rclcpp_node",
+      "-p",
+      "use_sim_time:=" +
+      std::string(node->get_parameter("use_sim_time").as_bool() ? "true" : "false"),
```
SteveMacenski, line 87: What if its not set in the yaml + they used different launch files without the set parameter?

nav2_bringup/launch/slam_launch.py
```diff
@@ -77,22 +74,22 @@ def generate_launch_description():
 
     # Nodes launching commands
 
-    start_map_saver_server_cmd = Node(
+    start_map_server = GroupAction(
+        actions=[SetParameter("use_sim_time", use_sim_time), Node(
```
SteveMacenski, line 78: Set new line for the `Node` start

SteveMacenski: The formatting here is a little funky for the `start_map_server` action

SteveMacenski (changes requested):
Generally LGTM! This is worth an update to the migration guide for - mentioning you can remove the `use_sim_time` from your param files.

nav2_bringup/launch/localization_launch.py
SteveMacenski: Why not use this for each node then?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
tonynajjar: image

in case you're wondering, doesn't apply for any other param

nav2_bringup/launch/localization_launch.py
tonynajjar: > does SetParameter work globally or can it be used for a single node?

It applies the parameter for all nodes within the GroupAction. See [here](https://github.com/ros2/launch_ros/blob/5434428a30a981aeee1dbadfd5b43622cef12d67/launch_ros/launch_ros/actions/set_parameter.py#L37)

> Also, does that work if the parameter in question is inside of the yaml file also

yes, the parameter in the yaml would override

For the refactoring of `yaml_filename`, I'd prefer if we can open a separate issue. I might take it but can't commit right now

nav2_bringup/launch/slam_launch.py
tonynajjar: Yeah I know it's funky but it passes the tests, I still don't understand why we don't have an automatic python formatter in ament_lint, like we have ament_uncrustify. I found this [discussion](https://discourse.ros.org/t/python-black-formatting/13277/27?page=2) about it.

Do you use any automatic python formatter?

nav2_bringup/launch/localization_launch.py
SteveMacenski: Totally understood, just asking the question, it sounds like this would be a solution for it too. Duly noted.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: Huh. OK, good to know

nav2_bringup/launch/slam_launch.py
tonynajjar: Do you usually fix those manually? Does ROS not have an auto-formatter in ament_lint?

nav2_controller/src/controller_server.cpp
tonynajjar: Ah yes forgot to comment here. This is similar to what is implemented in` bt_action_server_impl.hpp`. These are nodes created internally, as opposed to in the launch file, so the SetParameter doesn't seem to have any power here.

Unfortunately I did not find a way to set the `use_sim_time` parameter during the creation of the Node here because the parent node (of the controller server) is not passed `Costmap2DROS`. The next best thing I found was to set it after its creation.

I thought it wouldn't work, I assumed the Node would need to know this parameter at creation to know which kind of clock to create/subscribe to my tests proved otherwise :shrug:

nav2_planner/src/planner_server.cpp
tonynajjar: same as above

tonynajjar (commented):
In general I think this approach to using sim time should be made more public/popular in the ROS2 community. I can't find the discussions anymore but I remember many complaints from people switching from ROS1 on having to set this parameter manually for every node.

nav2_controller/src/controller_server.cpp
SteveMacenski: Could you add it as a constructor argument for `Costmap2DROS` and do it in there?

nav2_bringup/launch/localization_launch.py
SteveMacenski: I suppose just using the macro as Launch provides is nice, but if there's a strong reason to set it manually this way, that's OK too.

SteveMacenski:
> In general I think this approach to using sim time should be made more public/popular in the ROS2 community. I can't find the discussions anymore but I remember many complaints from people switching from ROS1 on having to set this parameter manually for every node.


Once this is merged, you should post about it with a link!

nav2_controller/src/controller_server.cpp
tonynajjar: Feasible I guess but can you clarify why you like this option better? It feels a bit intrusive to me

tonynajjar: Also, to make sure I understand: you mean to pass only the use_sim_time parameter right? not the whole parent node? Should it default to False?

SteveMacenski:
Other commits are showing up in this PR now - please rebase on main with only these comments / conflict resolutions

nav2_controller/src/controller_server.cpp
SteveMacenski: Yeah, just add the constructor argument `bool use_sim_time` which can just be passed via `get_parameter("use_sim_time").as_bool()`. Then within the costmap we can set it when the internal node is constructed as you did in the BT action. 

Setting a parameter outside like this seems a little hacky and I think just like the name / namespace of the costmap, which clock to use seems like a pretty fundamental input. Plus, something might initialize using a wrong clock then if changes are made later if `use_sim_time` is set after construction so its protecting us from a really subtle timing bug. It seems like a bad idea to have the wrong time source initialized then being changed in the application after construction. For example, other users of Costmap2DROS outside of the application servers won't know about this change and could cause weeks of headaches trying to track down why the wrong clock is used now that the example param files don't contain that argument

Less importantly, but other yaml files in the system tests are going to have the `use_sim_time` expectation to be set that this didn't update in launch, so if that's gone / deprecated, we need to update them to have it set in code - but its easier to find if you literally change the signature to find where it needs to be done.

tonynajjar:
Not sure I see what you mean, I merged master into my branch to fix the conflicts. The changes in "Files changed" are only my changes.
Should I have rebased instead and why?

SteveMacenski:
Oh sorry, I was a little quick on the draw there, this is fine

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -88,7 +88,8 @@ class Costmap2DROS : public nav2_util::LifecycleNode
   explicit Costmap2DROS(
     const std::string & name,
     const std::string & parent_namespace,
-    const std::string & local_namespace);
+    const std::string & local_namespace,
+    const bool & use_sim_time = false);
```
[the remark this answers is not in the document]
tonynajjar: meaning that I need to change all creations Costmap2DROS to pass an additional use_sim_time parameter?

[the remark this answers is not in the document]
SteveMacenski: Yes, but to be fair that's only like 3 places and you already did 2 of them - only missing `costmap_2d_node.cpp`. 

Also the `name` constructor on L80 above needs the time too. That's used in more places, but setting a default there would be OK since that's mostly there for generating tests. 

> Wouldn't that be a breaking change for anyone using that class standalone (I don't know how likely that is)?

Yes, but that is what `main` is for, we need to be able to make breaking changes. The fact that we set the `use_sim_time` in launch rather than in the yaml files was already a breaking change enough that this couldn't be backported.

nav2_bringup/launch/localization_launch.py
SteveMacenski: Leaving this un-resolved as a reminder to myself

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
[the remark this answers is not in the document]
tonynajjar: Yeah I was talking more about the name-only constructor. Thanks for clarifying that this one can take a default.
Great then, I will try to wrap this up tomorrow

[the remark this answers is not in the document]
tonynajjar: > only missing costmap_2d_node.cpp

are you sure? I see the name-constructor being used there

tonynajjar:
>  This is worth an update to the migration guide

In which migration guide should that be? Do you plan to backport this to Humble? If not, will you create the next migration guide for Iron?

tonynajjar:
> This is worth an update to the migration guide fo

https://github.com/ros-planning/navigation.ros.org/pull/339

nav2_controller/src/controller_server.cpp
```diff
@@ -62,8 +62,8 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
 
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
-    "local_costmap", std::string{get_namespace()}, "local_costmap");
-
+    "local_costmap", std::string{get_namespace()}, "local_costmap", get_parameter(
```
SteveMacenski, line 65: Last nit-pick and I'll merge this: Put the `get_parameter(` on the new line, its weird to have the line extension break there. I suspect that was a linter auto-fix. You can place `get_parameter("use_sim_time").as_bool());` all on the second line. As well for planner server, this is just a readability thing.

SteveMacenski (approved):

nav2_controller/src/controller_server.cpp
tonynajjar: Done. Yeah it was just automatic uncrustify. I'm pretty sure this can be avoided with some custom uncrustify config

SteveMacenski (approved):

Issue #3134, opened 2022-08-22
Unable to trigger minimum speed condition in xy_theta_iterator due to always False theta speed condition (min_speed_theta = 0.0)


smkyle90:
https://github.com/ros-planning/navigation2/blob/1421f17bd033b7c76affb655885a9f44904e1ecf/nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp#L102

I have been trying to configure the `dwb_core::DWBLocalPlanner` to generate valid twists with a linear speed range of [0.5, 1.0], however, the minimum speed condition in [xy_theta_iterator.cpp](https://github.com/ros-planning/navigation2/blob/1421f17bd033b7c76affb655885a9f44904e1ecf/nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp#L102) has an `&&` condition that is always false for my application.

Thus, even if my minimum speed condition is true and renders the speed invalid (desired behaviour), the `&&` condition on the theta velocity is always false (since `minimum_theta_speed: 0.0`) and the logic of `fabs(theta) + EPSILON < kinematics.getMinSpeedTheta()` is necessarily `false`, returning false for the whole statement, and therefore, all speeds in the range [0.0, 1.0] are deemed to be valid.

Is there a reason the condition that checks the `min_speed_xy` configuration is coupled with the `min_speed_theta` logic? 

If this is a one-off for my application, I am happy to fork and carry on. Otherwise, I can submit a PR de-coupling these checks. See code snippet below that is now giving the expected outcome. 

```C++
bool XYThetaIterator::isValidSpeed(double x, double y, double theta)
{
  KinematicParameters kinematics = kinematics_handler_->getKinematics();
  double vmag_sq = x * x + y * y;
  if (kinematics.getMaxSpeedXY() >= 0.0 && vmag_sq > kinematics.getMaxSpeedXY_SQ() + EPSILON) {
    return false;
  }
  // separate linear and rotational min speeds as they are not really dependent on each other.
  if (kinematics.getMinSpeedXY() >= 0.0 && vmag_sq + EPSILON < kinematics.getMinSpeedXY_SQ()) {
    return false;
  }
  if (kinematics.getMinSpeedTheta() >= 0.0 && fabs(theta) + EPSILON < kinematics.getMinSpeedTheta()) {
    return false;
  }
  if (vmag_sq == 0.0 && th_it_->getVelocity() == 0.0) {
    return false;
  }
  return true;
}
```

[...]

PR #3138, opened 2022-08-22
Decouple minimum xy speed check from the minimum theta speed check

smkyle90:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [3134](https://github.com/ros-planning/navigation2/issues/3134) |
| Robotic platform tested on | None. In simulation only. |

---

* This PR decouples the logic check in the `dwb_plugins::XYThetaIterator` to determine if a twist's speed is valid and as per the user-defined configuration.
* For a speed combination to be invalid the logic checks on lines 99, 102, 105 need to be true. However, for a system configured with `min_speed_xy > 0` and `min_speed_theta = 0` the logic defined in xy_theta_iterator.cpp [L102](https://github.com/ros-planning/navigation2/blob/d49e4053b7d9de52b42aaeb041f6450800dedb9d/nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp#L102) will always resolve to false, meaning a speed combination (specifically a lower than configured minimum xy speed) can incorrectly be deemed valid.
* The logic checks that are currently using an `&&`  are split into two independent checks. If either of the checks are true (essentially a `||` condition) the speed combination is invalid and returns the method returns `false`.
* There are no tests for the `XYThetaIterator` class, so no tests have been modified.

* None
---

* None

#### For Maintainers: 

### DIFF
--- a/nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp
+++ b/nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp
@@ -99,9 +99,10 @@ bool XYThetaIterator::isValidSpeed(double x, double y, double theta)
   if (kinematics.getMaxSpeedXY() >= 0.0 && vmag_sq > kinematics.getMaxSpeedXY_SQ() + EPSILON) {
     return false;
   }
-  if (kinematics.getMinSpeedXY() >= 0.0 && vmag_sq + EPSILON < kinematics.getMinSpeedXY_SQ() &&
-    kinematics.getMinSpeedTheta() >= 0.0 && fabs(theta) + EPSILON < kinematics.getMinSpeedTheta())
-  {
+  if (kinematics.getMinSpeedXY() >= 0.0 && vmag_sq + EPSILON < kinematics.getMinSpeedXY_SQ()) {
+    return false;
+  }
+  if (kinematics.getMinSpeedTheta() >= 0.0 && fabs(theta) + EPSILON < kinematics.getMinSpeedTheta()) {
     return false;
   }
   if (vmag_sq == 0.0 && th_it_->getVelocity() == 0.0) {


SteveMacenski (approved):

jwallace42 (approved):

SteveMacenski:
> The lengths are now less than main, indicating invalid twists were passing through the checks, and to me the numbers make sense -- as the bounds are increased in the tests, the number of valid twists also increases.

From a first look, I would agree with that statement. 

In concept, we know the accelerations / decelerations / min/max speeds, so we should be able to look at the matrix of values and determine how many "should" pass for each test condition & verify the number coming out is that value. 

I agree, I have no idea how more than 2000 could be generated, from the quick glance I did through this code. This was a port from ROS 1's `robot_navigation` so I'm actually not the most familiar with DWB.

@DLu would be the best person I know to judge this and let us know.

ayrton04:
This logic comes from the original `dwa_local_planner` in the `navigation` repo:

https://github.com/ros-planning/navigation/blob/47c9c629fc93e6c6bbcd3109eb2d5b55efce400d/base_local_planner/src/simple_trajectory_generator.cpp#L191-L196

I think the idea here was to make sure that at least _one_ of the dimensions (linear or angular speed) would be greater than its minimum. So the parameters would be read as "any trajectory with a value greater than this dimension's minimum parameter is valid", rather than "any trajectory with a value less than this dimension's minimum parameter is invalid." I think the parameter name is confusing, but given that this would be a break from every major DWA implementation since the inception of ROS, we might want to tread lightly.

smkyle90:
Thanks for the information.

Before I go review the details to determine the "correct" values for the test cases, I think it is prudent to ask, based on @ayrton04's insight, if the modified file is the right place to resolve the issue at hand.

At the end of the day, as a user, I want my planner to only score velocity combinations in the range that I have configured. One step further, I want to only generate twists in the configured range. Currently, "invalid" twists are being generated **and** the DWB planner is critiquing these twist combinations. (By "invalid" I really mean unexpected values outside the bounds of my configuration.)

Happy to change course as necessary.

SteveMacenski:
I think what Tom is saying is that those are valid trajectories, which I suppose makes some sense after stopping to think about it for a bit. For example, it would be pretty non-sensible for you to set a minimum theta velocity of non-zero value and reject any below that point -- that would mean your robot could literally never move in a straight line because it would be required to have some non-zero angular component.

I think its a different topic entirely about why more than 2000 trajectories are being generated - I agree that's odd. 

> we might want to tread lightly

I agree

smkyle90:
I understand there are use cases that don't make sense, but if the user (let's assume they know what they are doing) decides they want a specific configuration then I would expect that to be enforced. I think that is where I am getting hung up because it is seemingly (and as per @ayrton04's principal above) a softly enforced contract.

So I am not sure if there is an issue to be addressed here. There is for my specific application's parameters, but not for the masses who are using this code (or the earlier ROS version). I am intrigued by that.

Alas, I don't think the changes in this PR are appropriate. 

If we want to think about "strict" enforcement of configuration or the number of twists in the test cases, I am happy to keep the conversation going.

Thanks for all your time and input.

SteveMacenski:
I think this is OK to be closed - I think this could make sense w.r.t. translational velocities, but it is not very sensible for rotational velocities and unless an outcome here would be to remove the minimums on the rotational velocities (which I wouldn't support because there could be some small threshold for it that you'd like to have above for overcoming friction but not limiting to higher values), so I think this is a `wontfix` situation and perhaps its sensible for you to have a forked version of DWB.

Issue #3085, opened 2022-07-22
Start not included in SmacPlanner path


mangoschorle:
The path returned by smac planner doesn't explicitly include the start.

`bool NodeHybrid::backtracePath(CoordinateVector & path` will not add the start node (who's parrent I would assume is a nullptr). Whats the reasoning not to include the start. Is this on purpose or should there be a do while loop.

https://github.com/ros-planning/navigation2/blob/97068787fb35894a098f9c4600b5a06c7dddf43b/nav2_smac_planner/src/node_hybrid.cpp#L700

SteveMacenski:
That was likely an oversight. Feel free to submit a PR and I'd be happy to merge it. Its likely that all 3 planners have this issue, so fixing them across the board would be appreciated.

PR #3140, opened 2022-08-22
start is now added to the path

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3085 |

---

Start point is now added to the path

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/node_2d.cpp
+++ b/nav2_smac_planner/src/node_2d.cpp
@@ -154,11 +154,12 @@ bool Node2D::backtracePath(CoordinateVector & path)
 
   NodePtr current_node = this;
 
-  while (current_node->parent) {
+  do {
     path.push_back(
       Node2D::getCoords(current_node->getIndex()));
     current_node = current_node->parent;
   }
+  while (current_node->parent);
 
   return path.size() > 0;
 }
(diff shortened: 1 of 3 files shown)


nav2_smac_planner/src/node_2d.cpp
```diff
@@ -154,11 +154,12 @@ bool Node2D::backtracePath(CoordinateVector & path)
 
   NodePtr current_node = this;
 
-  while (current_node->parent) {
+  do {
     path.push_back(
       Node2D::getCoords(current_node->getIndex()));
     current_node = current_node->parent;
   }
+  while (current_node->parent);
```
SteveMacenski, line 162: Will have to wait for CI, but I think the while should be on the same line as the end bracket by styling, but LGTM if linting is happy!

SteveMacenski (approved):

mangoschorle:
@SteveMacenski  @jwallace42 

This is still incorrect
You need a final `path.push_back(
      Node2D::getCoords(current_node->getIndex()));` after the do while loop

jwallace42:
https://github.com/ros-planning/navigation2/pull/3168

Whoops, 

here is the fix. For some reason I only though we needed the do while loop to solve the issue. 

Thanks for pointing that out. :)

Issue #3109, opened 2022-08-12
Deadlock when a static layer gets map updated


daisukes:
I found a deadlock bug when I tested our new prototype, which frequently updates the static costmap.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu20.04
- ROS2 Version:
  - Galactic (ros:galactic docker image)
- Version or commit hash:
  - Galactic sync 2-based (a25cd438aa4cb08735ba9a6e9213c78b8ad7f528) custom build
- DDS implementation:
  - default (cyclone dds)

#### Steps to reproduce issue
- frequently update the static costmap (2hz)

#### Expected behavior
- no deadlock

#### Actual behavior
- deadlock
- gdb traceback
  - https://gist.github.com/daisukes/5e6e1d3758daa0c34996d60822b41170

#### Additional information
- main thread locks (Thread 15 in the gdb log)
  1. the mutex of the layered costmap
    https://github.com/ros-planning/navigation2/blob/6a9508b772eb6f07cbfd350b3578aa9de2567d59/nav2_costmap_2d/src/layered_costmap.cpp#L130
  2. then, the mutex of the static layer
    https://github.com/ros-planning/navigation2/blob/6a9508b772eb6f07cbfd350b3578aa9de2567d59/nav2_costmap_2d/plugins/static_layer.cpp#L336
- updating map thread locks (Thread 13 in the gdb log)
  1. the mutex of the static layer
    https://github.com/ros-planning/navigation2/blob/6a9508b772eb6f07cbfd350b3578aa9de2567d59/nav2_costmap_2d/plugins/static_layer.cpp#L272
  2. then, the mutex of the layered costmap, when the layered costmap needs to be resized
    https://github.com/ros-planning/navigation2/blob/6a9508b772eb6f07cbfd350b3578aa9de2567d59/nav2_costmap_2d/src/layered_costmap.cpp#L104

#### Implementation considerations
The mutex of the static layer should  be unlocked before calling resizeMap of the layered costmap
https://github.com/ros-planning/navigation2/blob/6a9508b772eb6f07cbfd350b3578aa9de2567d59/nav2_costmap_2d/plugins/static_layer.cpp#L194-L198

I think this lock can be removed. What do you think?
https://github.com/ros-planning/navigation2/blob/6a9508b772eb6f07cbfd350b3578aa9de2567d59/nav2_costmap_2d/plugins/static_layer.cpp#L272

SteveMacenski:
[...]

Issue #1625, opened 2020-04-06
nav2_costmap_2d StaticLayer locking problem


dwisse:
When the StaticLayer receives an incomingMap or incomingUpdate that changes the dimensions of the costmap during a updateMap in the LayeredCostmap, the updateCosts of the StaticLayer can cause getCost of the StaticLayer to do an out of bounds read of the underlying costmap.

The problem is that the dimension of the costmaps should not change between the updateBounds and the updateCosts calls in the updateMap function of the LayeredCostmap.

To resolve this issue, the incomingMap and incomingUpdate should lock the mutex of the layeredCostmap instead of it's local mutex. Then also the local locks in updateBounds and processMap can be removed

SteveMacenski:
> incomingMap and incomingUpdate should lock the mutex of the layeredCostmap instead of it's local mutex

That doesn't seem right to lock the entire costmap any time a single layer gets some data sources. Shouldn't you just lock the local mutex at the start of updateCosts and unlock at the end of updateMap?

dwisse:
I've found a problem using the locking in updateBounds and unlocking in
updateCosts approach.
When updateBounds doesn't detect a change in the costmap, updateCosts is
never called, and the mutex is never going to be unlock.
Could make a explicit method unlock in the costmap interface, but that
doesn't look nice.

SteveMacenski:
Perhaps in the layered costmap we can `layer->getMutex()` and lock them externally? We'd then have to remove all the mutexes from the costmap layers itself to ensure no lock out except by external sources (e.g. sensor, map, data updates).

I don't think there's going to be a single clean solution here since we separated these tasks. Maybe @dlu has a suggestion.

DLu:
I might be misunderstanding things, but I also wrote this code many moons ago and am by no means a threading/locking expert. 
 * When the static layers gets a new map of a different size, it [calls `LayeredCostmap::resizeMap`](https://github.com/ros-planning/navigation2/blob/26d776b4b1e3b39c01bd2f3a32fc924fff04ddf5/nav2_costmap_2d/plugins/static_layer.cpp#L169)
 * [`LayeredCostmap::resizeMap` locks the Layered Costmap's mutex](https://github.com/ros-planning/navigation2/blob/26d776b4b1e3b39c01bd2f3a32fc924fff04ddf5/nav2_costmap_2d/src/layered_costmap.cpp#L92)
[...]

PR #3145, opened 2022-08-23
bugfix (#3109) deadlock when costmap receives new map

daisukes:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3109 |
| Primary OS tested on | Ubuntu22.04 on docker |
| Robotic platform tested on | test code |

---

* fix the deadlock issue when costmap receives a new map
  * tested with a test code https://github.com/daisukes/nav2-costmap-deadlock-test
* as suggested by @SteveMacenski  https://github.com/ros-planning/navigation2/issues/3109#issuecomment-1222654667, `processMap` is called in `incomingMap` only for the first map
  * if the first map is buffered, some will fail
* since map is processed sequentially, `update_in_progress_` is removed

N/A
---

* backporting

#### For Maintainers: 


SteveMacenski (approved):
@gimait can you take a look at this since you added that logic to make sure you think this is sensible?

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -277,17 +276,13 @@ StaticLayer::interpretValue(unsigned char value)
...
 {
-  std::lock_guard<Costmap2D::mutex_t> guard(*getMutex());
   if (!map_received_) {
-    map_received_ = true;
-    processMap(*new_map);
-  }
-  if (update_in_progress_.load()) {
-    map_buffer_ = new_map;
-  } else {
     processMap(*new_map);
```
gimait, line 280: This will probably never be a problem, but please note that part of the `processMap` method is not thread-safe if we don't lock the Costmap2D mutex here as well. There is also a second call to this mutex within `processMap`, which might be able to produce the same deadlock that we are fixing.
The best way to go would probably find a way to move `processMap` to the main thread completely.
Again, this will probably never be a problem, but just in case.

gimait (approved):
I think this is great! We should have moved the processing of the map to the main loop from the beginning, all the threading hell we had here disappeared.

My only comment is that it would be great to remove that last call to `processMap` from `incomingMap`.
I also think that we should retest #1625 just in case we are missing something in the review, but I don't see how this might have been broken.

nav2_costmap_2d/plugins/static_layer.cpp
daisukes: Actually, I removed it once, but I got some errors in the test.

```
[133.395s] The following tests FAILED:
[133.395s] 	 15 - test_collision_checker (Timeout)
[133.395s] 	 17 - inflation_tests (Timeout)
```

These tests use `LayeredCostmap` and `StaticLayer`, but do not call `updateMap` and wait for the costmap to be ready. So they got a timeout.

So I needed to keep this call here and keep it mutex free because the mutex can cause a deadlock at the very beginning.

Do you want me to remove the call and fix the tests too?

gimait: I don't think you can simply remove it. If we want to remove it from here, we will need to do some other changes. There might be several places where the main loop might attempt to make use of the first map when was received, but not yet processed. We need to ensure that all places where the processed map is used are aware of the possibility of the map not being yet initialised.

Then, in my opinion, it is good that that test fails, it should, because you are changing the behaviour of this plugin. However, the tests failing are not a reason to not do the full change to processing in the main loop. If we want to move that last call to `processMap`, then we should fix these tests and/or write new ones that fit the current implementation.

>So I needed to keep this call here and keep it mutex free because the mutex can cause a deadlock at the very beginning.

I understand that you are focused on fixing the deadlock, but that doesn't mean you can simply leave unprotected code behind.
There are three points I want to do here, which I'm not sure have been considered here yet:

First, in [line 225 of the static layer](https://github.com/ros-planning/navigation2/blob/1d9d3c6208debf13b8ee38145bc25eef65a6b662/nav2_costmap_2d/plugins/static_layer.cpp#L225) we are locking the same mutex. This means you might not have fixed the lock entirely, only you don't see it anymore because it only happens with subsequent calls to `incomingMap`.

Second, [all this code](https://github.com/ros-planning/navigation2/blob/1d9d3c6208debf13b8ee38145bc25eef65a6b662/nav2_costmap_2d/plugins/static_layer.cpp#L189-L220) is retrieving data and performing operations on the layered costmap without protection, which to me, sounds like a very bad idea.

Lastly, this one I'm not sure about, but I think should be considered: setting `map_received_` without protection might bring undesired behaviour when checking it from the main loop. For example, in the main thread, you could have just called `updateBounds` with `map_received_` set to false, followed by a `updateCosts` with `map_received_` set to false.

As a rule of thumb, never leave code shared between threads unprotected.

SteveMacenski:
From https://github.com/ros-planning/navigation2/issues/1625

> The problem is that the dimension of the costmaps should not change between the updateBounds and the updateCosts calls in the updateMap function of the LayeredCostmap.

Now the updates are sequential in updateBounds so that should work fine. The exception is that first map coming in which I suppose could cause a problem in odd situations where `map_received_` is set to `true` so in updateCosts, its not passed through. 

I suppose we could re-introduce that atomic bool for if there's processing currently happening, though I don't love that. 

Another option would be to introduce a new class member boolean `map_received_in_update_bounds_` which is set to true/false based on the value of `map_updated_` in `updateBounds` so that we use the global state for the first update function, and then use whatever status was there for `updateCosts` instead of the global state. That would deal with the 1st map problem without blocking the threads anymore than necessary. 

Or maybe even `has_updated_data_` would do that already if that's set to `true`?

nav2_costmap_2d/plugins/static_layer.cpp
SteveMacenski: I don't quite understand the issue. In steady state, `processMap` is only called in the `updateBounds` function, so we're really only talking about the very first map when the current costmap is not set to any particular valid size/content.

In the situation we're talking about with the first map, the `map_received_` boolean at the start of both Update functions will make them immediately return without trying to do anything (if false) - which is before any map is processed. 

So, if no map is processed before updateBounds its OK. If its processed before that point in the map callback, also OK. 

If no map is processed before updateCosts and updateBounds, OK. If its processed after updateBounds but before updateCosts, that's where it could potentially be not OK. I think my comment just below lays out a solution to that (maintain the state of the "updatedness" from updateBounds to also use in updateCosts.

daisukes:
@gimait, @SteveMacenski, Thank you for the comments!

Let me summarize based on my understanding.
- The current main code has two issues; (1) deadlock in `incomingMap` and (2) `map_received_` can be changed between `updateBounds` and `updateCosts` at the first `incomingMap` call
- My PR focused on issue (1) and does not fix issue (2)
- To fix issue (2), introducing `map_received_in_update_bounds_` would be good

> Then it’s up to you to remove the last processMap call here and fix the tests.
I like removing the last `processMap` call in `incomingMap` to make it as simple as possible, but as @gimait mentioned, it will change the behavior of the system.

How about

1. I will make a change to fix (2) in this PR
2. I will also make another PR that will be a version having no `processMap` in `incomingMap` and fixed tests

and then, you will decide on which version you will merge. What do you think?

SteveMacenski (approved):
LGTM - can you confirm the deadlock is gone still (just sanity checking)?

Can you rebase / pull in `main`? There are some test failures that should be green b/c its outdated. I want to see CI green on all PRs from now on :-)

daisukes:
I've merged the latest main branch and checked the deadlock is gone with my test code (15 mins).

SteveMacenski:
In your view, is this ready to merge or are there some updates to tests you want to make?

daisukes:
I think it is ready.

SteveMacenski:
Awesome, thanks!

Issue #3043, opened 2022-06-25
Feature Request: Feedback for Navigation Failures


karl-schulz:
## Summary

When reacting to navigation failures, it is essential to know *why* it failed to be able to react to it in a reasonable way. 

## Explanation

* Even inside the stack, planner's `ComputePathToPose` and controller's `FollowPath` provide no additional information than success/failure and a `std_msgs/Empty`.
* From the outside of navigation, there is also no way of knowing the overall reason for a navigation failure (e.g. Planner or Controller failure, no progress, ...). Calling the `NavigateToPose` action, the resulting response is just a `std_msgs/Empty`. 

Example where feedback could be useful: In the behavior trees, we often see recovery actions for planner/controller failures. This makes sense if the computation failed because the robot is stuck somehow and turning/waiting can resolve the issue. But if the planner failed because the goal is occupied, it makes no sense that the robot backs up or turns around as soon as it sees it. Different reactions depending on the cause for failure are needed.

## What can we do about it?

First of all: Do you agree that this is problematic?

I am not sure what the best way of dealing with this issue is and wanted to openly ask what you think about it. 

Some ideas: 

* Planner and controllers could loosely publish failure reasons on separate topics for anyone to react to it
* Alternatively, they could use the blackboard for this
* The action interfaces could have `error` fields

Still, I don't see a clean way of dealing with failure reasons in the behavior trees, the architecture doesn't seem to allow for it.

vinnnyr:
In terms of implementation, I wonder if some of this is a good application of diagnostics

mikeferguson:
Just a note, ROS diagnostics are usually meant to be "human readable" - the format is a bunch of strings, so it's not usually the best system if you want to autonomously make decisions based on the data.

mikeferguson:
I will also note - this feature (different recovery behaviors depending on the actual cause of failure) was a frequent request in ROS1 as well - we never actually got there.

SteveMacenski:
>  But if the planner failed because the goal is occupied, it makes no sense that the robot backs up or turns around as soon as it sees it. Different reactions depending on the cause for failure are needed.

[...]

PR #3146, opened 2022-08-23
added result codes for global planner

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3043 |

---

---

#### For Maintainers: 


nav2_msgs/action/ComputePathToPose.action
```diff
@@ -7,5 +7,6 @@ bool use_start # If false, use current robot pose as path start, if true, use st
 #result definition
 nav_msgs/Path path
 builtin_interfaces/Duration planning_time
+GlobalPlannerResultCode result_code
```
SteveMacenski, line 10: You can define the codes in this message and have the type here just be the int16 (and 16 is probably fine for this).

Start the enumeration at 2 though, leave `0` for invalid and `1` for unknown. 

Also, how did you come up with that list of result codes? Did you go through all the planners and find all the reasons why they throw exceptions / fail? Any others we should add? Might be worth brain storming some ideas for other ones that would be of use

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
```diff
@@ -70,6 +70,8 @@ class ComputePathToPoseAction : public BtActionNode<nav2_msgs::action::ComputePa
     return providedBasicPorts(
       {
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathToPose node"),
+        BT::OutputPort<nav2_msgs::msg::GlobalPlannerResultCode>("global_planner_result_code",
```
SteveMacenski, line 73: Just an int32 perhaps? Or I think you can even do `nav2_msgs::msg::ComputePathToGoal::Response::ErrorCode` (or similar)

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -40,20 +40,29 @@ void ComputePathToPoseAction::on_tick()
...
+  setOutput("global_planner_result_code", result_.result);
   return BT::NodeStatus::SUCCESS;
 }
 
 BT::NodeStatus ComputePathToPoseAction::on_aborted()
 {
   nav_msgs::msg::Path empty_path;
   setOutput("path", empty_path);
+
+  nav2_msgs::msg::GlobalPlannerResultCode result_code;
```
SteveMacenski, line 52: A strong reason for `0` to be invalid

nav2_planner/src/planner_server.cpp
```diff
@@ -461,26 +461,52 @@ PlannerServer::computePlan()
       return;
     }
 
-    result->path = getPlan(start, goal_pose, goal->planner_id);
+    try
```
SteveMacenski, line 464: There's already a try wrapping this full thing, I think you just needed to add the new catches before the default one

nav2_planner/src/planner_server.cpp
```diff
@@ -461,26 +461,52 @@ PlannerServer::computePlan()
...
+            get_logger(),
+            "Planner loop missed its desired rate of %.4f Hz. Current loop rate is %.4f Hz",
+            1 / max_planner_duration_, 1 / cycle_duration.seconds());
+      }
 
-    action_server_pose_->succeeded_current(result);
+      action_server_pose_->succeeded_current(result);
+    }
+    catch (nav2_core::GlobalPlannerStartOccupiedException &e)
+    {
```
SteveMacenski, line 488: Some logging for each would be nice (like the default exception catching already in the server)

SteveMacenski (changes requested):
Generally speaking, LGTM with some refinement

jwallace42:
Alright, 

I cleaned it up a quite a bit. I haven't explicitly made a list for all the runtime errors that planners could through. I can try to work on that next. I was trying to get a rough structure out there :).

jwallace42:
As for errors, 

smac is currently the only planner that throws anything. 

1. No costmap given
2. No valid start or goal 
3. goal occupied with no tolerance 
4. Start in lethal space

The other planners don't throw any runtime errors.

SteveMacenski:
Sure, but there are other error conditions that might be helpful to know about, even if they don't exist as exceptions right now. For instance, NavFn has [off costmap checks](https://github.com/ros-planning/navigation2/blob/main/nav2_navfn_planner/src/navfn_planner.cpp#L225). 

Also, lets do 1 subsystem at a time, so lets keep this with planning, open a different PR for controllers

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -40,20 +40,23 @@ void ComputePathToPoseAction::on_tick()
...
   setOutput("path", empty_path);
+  setOutput("global_planner_error_code", result_.result->error_code);
   return BT::NodeStatus::FAILURE;
 }
 
 BT::NodeStatus ComputePathToPoseAction::on_cancelled()
 {
   nav_msgs::msg::Path empty_path;
   setOutput("path", empty_path);
+  setOutput("global_planner_error_code", result_.result->error_code);
```
SteveMacenski, line 59: If it was aborted, there would be no error code (or I suppose the error code would be that we canceled it, not really an error though) so set to 0?

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -40,20 +40,23 @@ void ComputePathToPoseAction::on_tick()
 BT::NodeStatus ComputePathToPoseAction::on_success()
 {
   setOutput("path", result_.result->path);
+  setOutput("global_planner_result_code", result_.result->error_code);
```
SteveMacenski, line 43: If its successful, it shouldn't have an error code - so set to 0?

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -40,20 +40,23 @@ void ComputePathToPoseAction::on_tick()
...
   return BT::NodeStatus::SUCCESS;
 }
 
 BT::NodeStatus ComputePathToPoseAction::on_aborted()
 {
   nav_msgs::msg::Path empty_path;
   setOutput("path", empty_path);
+  setOutput("global_planner_error_code", result_.result->error_code);
```
SteveMacenski, line 51: Aborted I think is the only error state here

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
jwallace42: Sure, 

but shouldn't that just be passed though on the action server side? 

`action_server_->on_success(result)`

I guess it never hurts to specify.

jwallace42:
So, 

the list is currently 

int16 START_OCCUPIED=2
int16 GOAL_OCCUPIED=3
int16 NO_VALID_PATH=4
int16 TIMEOUT=5
int16 START_OUTSIDE_MAP=6
int16 GOAL_OUTSIDE_MAP=7
int16 START_IS_EQUAL_TO_GOAL=8
int16 TF_ERROR=9

I assume we will also want to add this to the compute path through poses action?

jwallace42:
@SteveMacenski let me know if there is anything else you might want me to do here(what do we want to do for code cov). I can also create a separate pr for the ComputePathThroughPoses error codes.

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
SteveMacenski: No, I think `aborted` is the only state whre the error code is meaningfully populated. Cancel means we called it so there's not really an error. Success means it came back with a result completed. 

The current status of this PR actually sets aborted to 0 and cancel to value, when it should be the opposite.

Also, please add inline comments when you set to zero why we do that for each of the cases (success is.. successful; cancel is user cancelled so algorithm didn't fail)

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -250,28 +250,15 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
   // Compute plan
   Node2D::CoordinateVector path;
   int num_iterations = 0;
-  std::string error;
```
SteveMacenski, line 253: A note in all 3 planners as a comment that the exceptions thrown by the planners will be captured by the planner servers and states returned to the action requester

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -250,28 +250,15 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
...
-  try {
-    if (!_a_star->createPath(
```
SteveMacenski, line 255: No updates to Theta* or NavFn?

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -250,28 +250,15 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
...
-        error = std::string("exceeded maximum iterations");
-      }
+  if (!_a_star->createPath(
+      path, num_iterations,
+      _tolerance / static_cast<float>(costmap->getResolution())))
+  {
+    if (num_iterations < _a_star->getMaxIterations()) {
+      throw nav2_core::NoValidPathCouldBeFound("no valid path found");
+    } else {
+      throw nav2_core::NoValidPathCouldBeFound("exceeded maximum iterations");
```
SteveMacenski, line 260: This seems like `TIMEOUT` might be the right thing -- its not "time" out, but its out of valid iterations from a search-based planner

nav2_msgs/action/ComputePathToPose.action
```diff
@@ -1,3 +1,15 @@
+# Error codes
+int16 NONE=0
+int16 UNKNOWN=1
+int16 START_OCCUPIED=2
+int16 GOAL_OCCUPIED=3
+int16 NO_VALID_PATH=4
+int16 TIMEOUT=5
+int16 START_OUTSIDE_MAP=6
```
SteveMacenski, line 8: Seems like 6,7,8 could be put into use in the planners? I don't see them applied.

nav2_planner/src/planner_server.cpp
```diff
@@ -481,12 +481,73 @@ PlannerServer::computePlan()
...
+  } catch (nav2_core::StartIsEqualToGoal & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Start equal to goal: \"%s\"",
+      goal->planner_id.c_str(),
+      ex.what());
+    result->error_code = nav2_msgs::action::ComputePathToPose::Goal::START_IS_EQUAL_TO_GOAL;
+    action_server_pose_->terminate_current(result);
+  } catch (nav2_core::PlannerTFError & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Start equal to goal: \"%s\"",
```
SteveMacenski, line 540: TF error message should be updated

nav2_planner/src/planner_server.cpp
```diff
@@ -481,12 +481,73 @@ PlannerServer::computePlan()
...
+      ex.what());
+    result->error_code = nav2_msgs::action::ComputePathToPose::Goal::TF_ERROR;
+    action_server_pose_->terminate_current(result);
   } catch (std::exception & ex) {
     RCLCPP_WARN(
-      get_logger(), "%s plugin failed to plan calculation to (%.2f, %.2f): \"%s\"",
-      goal->planner_id.c_str(), goal->goal.pose.position.x,
-      goal->goal.pose.position.y, ex.what());
-    action_server_pose_->terminate_current();
+      get_logger(), "%s plugin failed to plan: \"%s\"",
```
SteveMacenski, line 547: Still add the goal pose here, it helps understand which request specifically failed

nav2_planner/src/planner_server.cpp
```diff
@@ -481,12 +481,73 @@ PlannerServer::computePlan()
...
+  } catch (nav2_core::GoalOutsideMapBounds & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Goal Outside map: \"%s\"",
+      goal->planner_id.c_str(),
+      ex.what());
+    result->error_code = nav2_msgs::action::ComputePathToPose::Goal::GOAL_OUTSIDE_MAP;
+    action_server_pose_->terminate_current(result);
+  } catch (nav2_core::StartIsEqualToGoal & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Start equal to goal: \"%s\"",
```
SteveMacenski, line 533: Add in that pose

nav2_planner/src/planner_server.cpp
```diff
@@ -481,12 +481,73 @@ PlannerServer::computePlan()
...
+  } catch (nav2_core::StartOutsideMapBounds & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Start Outside map: \"%s\"",
+      goal->planner_id.c_str(),
+      ex.what());
+    result->error_code = nav2_msgs::action::ComputePathToPose::Goal::START_OUTSIDE_MAP;
+    action_server_pose_->terminate_current(result);
+  } catch (nav2_core::GoalOutsideMapBounds & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Goal Outside map: \"%s\"",
```
SteveMacenski, line 526: Add in that pose

nav2_planner/src/planner_server.cpp
```diff
@@ -481,12 +481,73 @@ PlannerServer::computePlan()
...
+  } catch (nav2_core::PlannerTimedOut & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Timeout occurred: \"%s\"",
+      goal->planner_id.c_str(),
+      ex.what());
+    result->error_code = nav2_msgs::action::ComputePathToPose::Goal::TIMEOUT;
+    action_server_pose_->terminate_current(result);
+  } catch (nav2_core::StartOutsideMapBounds & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. Start Outside map: \"%s\"",
```
SteveMacenski, line 519: Add in that pose

nav2_planner/src/planner_server.cpp
```diff
@@ -481,12 +481,73 @@ PlannerServer::computePlan()
...
+      get_logger(), "%s plugin failed to plan path. Goal Pose (%.2f, %.2f) was occupied: \"%s\"",
+      goal->planner_id.c_str(),
+      goal->goal.pose.position.x,
+      goal->goal.pose.position.y,
+      ex.what());
+    result->error_code = nav2_msgs::action::ComputePathToPose::Goal::GOAL_OCCUPIED;
+    action_server_pose_->terminate_current(result);
+  } catch (nav2_core::NoValidPathCouldBeFound & ex) {
+    RCLCPP_WARN(
+      get_logger(), "%s plugin failed to plan path. No Valid Path found: \"%s\"",
```
SteveMacenski, line 505: Add in the goal pose, it helps understand which goal request this is in reference to

SteveMacenski (changes requested):
Generally speaking LGTM!

SteveMacenski:
@jwallace42 see comments from older reviews above - still some things wrong :smile:

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
jwallace42: Ah, 

I misunderstood you the first time :).

jwallace42:
I think this should be all set :).

jwallace42:
The other thing we should look at(separate pr) is passing the accumulation of the errors up through the navigate to pose action.

nav2_msgs/action/ComputePathToPose.action
jwallace42: No valid path and timeout gets throw by smac and hybrid a star. Start outside map gets throw by navfn.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
```diff
@@ -75,6 +75,8 @@ class ComputePathToPoseAction : public BtActionNode<nav2_msgs::action::ComputePa
     return providedBasicPorts(
       {
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathToPose node"),
+        BT::OutputPort<nav2_msgs::action::ComputePathToPose::Result::_error_code_type>(
+          "compute_path_to_pose_error_code", "The global planner error code"),
```
SteveMacenski, line 79: "The compute path to pose error code"

nav2_theta_star_planner/src/theta_star_planner.cpp
```diff
@@ -154,13 +153,13 @@ void ThetaStarPlanner::getPlan(nav_msgs::msg::Path & global_path)
 {
   std::vector<coordsW> path;
   if (planner_->isUnsafeToPlan()) {
-    RCLCPP_ERROR(logger_, "Either of the start or goal pose are an obstacle! ");
     global_path.poses.clear();
+    throw nav2_core::PlannerException("Either of the start or goal pose are an obstacle! ");
```
SteveMacenski, line 157: Check test coverage - exceptions are being thrown in unit tests that need to be caught so they continue for error condition testing.

nav2_theta_star_planner/src/theta_star_planner.cpp
```diff
@@ -93,8 +93,7 @@ nav_msgs::msg::Path ThetaStarPlanner::createPlan(
   planner_->costmap_->worldToMap(goal.pose.position.x, goal.pose.position.y, mx_goal, my_goal);
   if (mx_start == mx_goal && my_start == my_goal) {
```
SteveMacenski, line 94: Isn't this start=goal a planner error type?

I think you should carefully review the smac/navfn/theta* planners and make sure that for the 8 types of errors you have enumerated, each (where relevant) is thrown in the planner algorithms

nav2_smac_planner/include/nav2_smac_planner/smac_planner_hybrid.hpp
```diff
@@ -36,6 +36,11 @@
 namespace nav2_smac_planner
 {
 
+/**
+ * Note: All exceptions thrown are handled by the planner server and returned to the action
```
SteveMacenski, line 40: This note should be in each of the 3 smac planners when `_a_star->createPath` is called, not here :smile:

nav2_theta_star_planner/src/theta_star_planner.cpp
SteveMacenski: If there are some checks missing, maybe also worth adding them

SteveMacenski (changes requested):

nav2_smac_planner/src/smac_planner_2d.cpp
SteveMacenski: @jwallace42 still relevant

nav2_theta_star_planner/src/theta_star_planner.cpp
jwallace42: So, 

the angle could be different that is just check the x y position. 

Those eight exceptions don't necessary need to be throw by the all the planners. I went through and substituted out warning for throwing exceptions. If the planners doesn't care about a particular error I ignored it. (thought it was outside the scope :)). 

If we want to keep consistent across planners generic checks should be done up front in the planner server rather than the plugin.

The planner server would throw:
1. goal/start occupied 
2. start/goal outside map 
3. start==goal

The planners would throw: 
1. NoValidPathCouldBeFound
2. PlannerTimedOut.

jwallace42: Out of scope but something I was just thinking about. 

If a goal is blocked it be quite useful to come up with some way of adjusting the goal to a valid cell.

SteveMacenski: > the angle could be different that is just checking the x y position.

That's only important for feasible planners (e.g. Hybrid-A* and State Lattice). The angle otherwise isn't considered with the rest since they're holonomic planners. That's kind of the main reason those checks can't happen in the planner server itself - it has no knowledge about which planners care about orientation and which don't. 

I agree that the 8 exceptions do not need to be thrown by all planners. I guess what I'm asking for is some consistency over all of the algorithms that they throw the "core" ones (goal/start occupied; off map; or equal) since we're now actually changing the behavior of the server to catch and send these up to the autonomy application. If a user has a BT that relies on these codes for recovery actions, then it stands to reason that if they change to any of the "supported" algorithms, they should be interchangable. 

The only way to get that is if each of the plugins then throws the all of the exceptions that are relevant to them, even if right now they didn't previously log a complaint. For example, the Theta* planner doesn't have a time limit, so it wouldn't be sensible for it to throw a `TIMEDOUT` error. But all of the planners can check if `start=goal` and have the identical behavior. For the feasible planners, that check should include orientation, else, it shouldn't. 

> If a goal is blocked it would quite useful to come up with some way of adjusting the goal to a valid cell.

NavFn and Smac2D do this. Navfn will backtrack the path to the goal until it is outside of the set `tolerance` for something valid. Smac2D has the `iterations_on_approach` to check for the number of iterations within tolerance, once exceeded, it returns the best result. We don't have this for the feasible planners (yet). I'm not sure why we don't have it for Theta*.

nav2_theta_star_planner/test/test_theta_star.cpp
```diff
@@ -174,8 +174,12 @@ TEST(ThetaStarPlanner, test_theta_star_planner) {
   }
   goal.pose.position.x = 1.0;
   goal.pose.position.y = 1.0;
-  path = planner_2d->createPlan(start, goal);
-  EXPECT_EQ(static_cast<int>(path.poses.size()), 0);
+
+  try {
+    path = planner_2d->createPlan(start, goal);
```
SteveMacenski, line 179: You can use `EXPECT_THROW` with the type of exception here

nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -134,6 +134,22 @@ nav_msgs::msg::Path NavfnPlanner::createPlan(
...
+
+  if (!costmap_->worldToMap(goal.pose.position.x, goal.pose.position.y, mx_goal, my_goal)) {
+    throw nav2_core::GoalOutsideMapBounds("Goal outside map bounds");
+  }
+
+  if (costmap_->getCost(mx_start, my_goal) == nav2_costmap_2d::LETHAL_OBSTACLE) {
+    throw nav2_core::StartOccupied("Failed to create a unique pose path because of obstacles");
+  }
+
+  if (costmap_->getCost(mx_goal, my_goal) == nav2_costmap_2d::LETHAL_OBSTACLE) {
```
SteveMacenski, line 150: Navfn will handle occupied goals actually, if tolerance is set to non-0. This should check if tolerance is set to 0, else this shouldn't be thrown

nav2_planner/src/planner_server.cpp
```diff
@@ -282,65 +282,32 @@ void PlannerServer::getPreemptedGoalIfRequested(
 
 template<typename T>
 bool PlannerServer::getStartPose(
-  std::unique_ptr<nav2_util::SimpleActionServer<T>> & action_server,
```
SteveMacenski, line 285: Why remove the action server? This is strictly necessary to terminate in the places you removed them from.

nav2_theta_star_planner/src/theta_star_planner.cpp
```diff
@@ -89,13 +89,27 @@ nav_msgs::msg::Path ThetaStarPlanner::createPlan(
...
+  }
+
+  if (!planner_->costmap_->worldToMap(
+      goal.pose.position.x, goal.pose.position.y, mx_goal, my_goal))
+  {
+    throw nav2_core::GoalOutsideMapBounds("Goal outside map bounds");
+  }
+
+  if (planner_->costmap_->getCost(mx_start, my_start) == nav2_costmap_2d::LETHAL_OBSTACLE) {
+    throw nav2_core::StartOccupied("Failed to create a unique pose path because of obstacles");
```
SteveMacenski, line 105: Mention start/goal (depending which is being thrown) for logging in all of the `XYZOccupied` messages please

nav2_planner/src/planner_server.cpp
jwallace42: Yeah, the action server still gets terminated. If `getStartPose` returns false, we throw a plannerTFError which gets caught and handled.

jwallace42:
> NavFn and Smac2D do this. Navfn will backtrack the path to the goal until it is outside of the set tolerance for something valid. Smac2D has the iterations_on_approach to check for the number of iterations within tolerance, once exceeded, it returns the best result. We don't have this for the feasible planners (yet). I'm not sure why we don't have it for Theta*.
> 
> The easiest implementation would be to check isGoal() as within planning tolerance and just return true if so, that way even if the goal is invalid, it'll return a valid path as long as there's some point on approach during planning that is within tolerance. The downside is that then it will usually not give you the exact goal point, even if possible. It'll always be something within a tolerance circle.
> 
> Another way might be to check if goal is occupied and then only do that behavior if that is true, e.g. isGoal() has a bool condition if the goal is occupied and either wants exact if not, or within tolerance if it is. That would be applicable to both the 2D-A* in Smac as well as the Theta*. Then we could expose the iterations_on_approach for only the Hybrid/Lattice planners to give it the ability to do approximate goals.

nav2_planner/src/planner_server.cpp
SteveMacenski: Ah makes sense! What about removing `validatePath`? That's a useful abstraction and could have a similar thing (if returns false, then throws and catches to terminate)

SteveMacenski:
Talked with @doisyg over email - I think having a single-point trivial path where start=goal is sensible. This is definitely something that is interpretable to different people different ways, but I think that's the most sensible default at this time. This would allow those requests to succeed so that the task can auto-succeed in the BT navigator. If we throw, then we would fail the task which could mess with an application, even though we are in fact at the same location. 

> What about moving this outside the planning algorithm?

This is algorithm specific knowledge / methodology (2D search will need to have separate checks then SE2 search; or sampling, optimization approaches, etc). Also even if we did it in the planner server, the plugins still need to validate their inputs for good software design so it wouldn't really change much.

SteveMacenski:
I had some time to think about this today and this is what I  came up with:
- I think that the Smac2D method is pretty good for the method because it'll return the closest possible path if we have something within tolerances and we cannot get better (and still trying to search)
- However, if we run out of things to search for (like in confined settings), we don't actually return the closest path since we have to exceed the number of attempted iterations within the tolerance
- So, I added a check on the best heuristic cost **after** the loop as well, and turn a path if within tolerances & ran out of search iterations
- Then exposing the same tolerance and max_iterations_on_approach parameters as Ndoe2D
- Poof, Hybrid-A* is really quite good now at on approach estimations

In this example - I request the middle of the obstacle, and it gives me out a pretty decent approximate path 
Screenshot from 2022-09-22 14-32-46

PR to follow, I'll tag you

nav2_planner/src/planner_server.cpp
jwallace42: Well that func only checks to see if the path was non empty. So that got changed to look like

` 
if (result->path.poses.empty()) {
     throw nav2_core::PlannerException(goal->planner_id + "generated a empty path");
}`

SteveMacenski: But that was abstracting out logic from both the plan to pose and plan through poses APIs, and likely is to be expanded over time for additional checks. Please readd the function even if for now it just logs / checks for the path being empty to reduce code duplication

nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -148,12 +172,6 @@ nav_msgs::msg::Path NavfnPlanner::createPlan(
   if (start.pose.position.x == goal.pose.position.x &&
     start.pose.position.y == goal.pose.position.y)
   {
-    unsigned int mx, my;
-    costmap_->worldToMap(start.pose.position.x, start.pose.position.y, mx, my);
```
SteveMacenski, line 152: Readd this - if in collision we should still fail, the trivial single-point paths should only be given if start is valid (because might need to rotate in place or something, and if already in collision, that's not good)!

nav2_planner/src/planner_server.cpp
```diff
@@ -450,21 +418,20 @@ PlannerServer::computePlan()
     getPreemptedGoalIfRequested(action_server_pose_, goal);
 
     // Use start pose if provided otherwise use current robot pose
-    geometry_msgs::msg::PoseStamped start;
-    if (!getStartPose(action_server_pose_, goal, start)) {
-      return;
+    if (!getStartPose<ActionToPose>(goal, start)) {
```
SteveMacenski, line 421: Is the explicit template needed?

SteveMacenski (changes requested):
I do really like the exceptions you added in not having to pass through the action server to `terminate` to a number of the internal functions, that is a really nice quality-of-reading improvement to have that all handled in the `catches`.

nav2_navfn_planner/src/navfn_planner.cpp
SteveMacenski: There's also an eq. check in Smac2D that is still there, both of these should now throw exceptions for path invalid due to collision (but not due to single valued path)

SteveMacenski:
Other than that, I think this is good to go!

doisyg:
Still away from a programming and robotics test setup for a few days but I really like the spirit of this PR. This will help formalize what several robotics applications are doing closed source about handling planner failures. On top of my head, all the use cases related to occupied start or goal. Before: checks had to be done about the validity of the requested start and goal poses before a planner call; with this PR, it will be a matter of checking the failure code.
One comment however, it may be useful to clearly specify the priority of the error codes and to check their consistency. If both the start and goal are occupied, which code should be returned ? The order at which they are defined in the action message is probably a good implicit definition of their priority though.

nav2_navfn_planner/src/navfn_planner.cpp
jwallace42: For navfn, that issue should be caught above before it even gets to the equality check in line 150. 

I updated the check for smac2D.

nav2_planner/src/planner_server.cpp
jwallace42: Yes, 

once you remove the action server parameter it is unable to infer the template arg.

jwallace42:
I think the order of throwing order should be 
```
# Error codes
int16 TF_ERROR
int16 START_OCCUPIED
int16 START_OUTSIDE_MAP
int16 GOAL_OCCUPIED
int16 GOAL_OUTSIDE_MAP
int16 NO_VALID_PATH
int16 TIMEOUT
```
@doisyg any thoughts?

nav2_planner/src/planner_server.cpp
jwallace42: added it back in :)

jwallace42:
> For what analogs exist here that require changes, if you make them to the controller PR, I can start reviewing that critically next week!

Once this gets I will go though and clean it up :).

doisyg:
> I think the order of throwing order should be
> 
> ```
> # Error codes
> int16 TF_ERROR
> int16 START_OCCUPIED
> int16 START_OUTSIDE_MAP
> int16 GOAL_OCCUPIED
> int16 GOAL_OUTSIDE_MAP
> int16 NO_VALID_PATH
> int16 TIMEOUT
> ```
> 
> @doisyg any thoughts?

hm, I don't have a strong opinion this. Maybe it is more logical to have the outside map test before the occupied one ? dk

jwallace42:
Well, if you are outside the map, it can't be occupied and vis versa.  So, the ordering of occupied vs outside doesn't matter because they mutually exclusive.

nav2_navfn_planner/src/navfn_planner.cpp
SteveMacenski: Good point!

SteveMacenski:
> One comment however, it may be useful to clearly specify the priority of the error codes and to check their consistency. If both the start and goal are occupied, which code should be returned ? The order at which they are defined in the action message is probably a good implicit definition of their priority though.

Good question! 

I propose:

```
# Error codes
int16 TF_ERROR #system failure
int16 START_OUTSIDE_MAP #request failure
int16 GOAL_OUTSIDE_MAP #request failure
int16 START_OCCUPIED #request failure
int16 GOAL_OCCUPIED #request failure if tol=0 
int16 TIMEOUT # too difficult or impossible
int16 NO_VALID_PATH #impossible
```

As critical failures are highest priority (system cannot operate or request is totally invalid via occupation). Then we have the codes that relate to real run-time failures unable to find paths. Between invalid path and timeout, I don't have a strong opinion, but timeout seems "less serious" than total failure to me.

nav2_planner/include/nav2_planner/planner_server.hpp
```diff
@@ -152,24 +153,24 @@ class PlannerServer : public nav2_util::LifecycleNode
...
    * @return bool If successful in transforming poses
    */
-  template<typename T>
   bool transformPosesToGlobalFrame(
-    std::unique_ptr<nav2_util::SimpleActionServer<T>> & action_server,
     geometry_msgs::msg::PoseStamped & curr_start,
     geometry_msgs::msg::PoseStamped & curr_goal);
 
+  // Our action server implements the ComputePathToPose action
+  std::unique_ptr<ActionServerToPose> action_server_pose_;
```
SteveMacenski, line 171: Move back down - all member objects after all member functions

nav2_planner/src/planner_server.cpp
```diff
@@ -316,18 +309,16 @@ bool PlannerServer::transformPosesToGlobalFrame(
...
 {
-  if (path.poses.size() == 0) {
+  if (path.poses.empty()) {
     RCLCPP_WARN(
       get_logger(), "Planning algorithm %s failed to generate a valid"
       " path to (%.2f, %.2f)", planner_id.c_str(),
       goal.pose.position.x, goal.pose.position.y);
-    action_server->terminate_current();
-    return false;
+    throw nav2_core::PlannerException(planner_id + "generated a empty path");
```
SteveMacenski, line 321: Isn't this a coded failure type, `NO_VALID_PATH`?

nav2_planner/src/planner_server.cpp
```diff
@@ -386,17 +376,14 @@ PlannerServer::computePlanThroughPoses()
...
       }
 
       // Get plan from start -> goal
       nav_msgs::msg::Path curr_path = getPlan(curr_start, curr_goal, goal->planner_id);
 
-      // check path for validity
-      if (!validatePath(action_server_poses_, curr_goal, curr_path, goal->planner_id)) {
-        return;
-      }
+      validatePath<ActionThroughPoses>(curr_goal, curr_path, goal->planner_id);
```
SteveMacenski, line 386: We have a few functions here of the style `if (!function(args)) {throw ...}`, we should do that here too. Have validate return a bool that throws here, not in the function body, to be consistent

jwallace42:
> But the larger problem of how to do we take the "most serious one" isn't easy. A TF failure could happen in the planner after some start/goal outside map checks, but if we through from map checks, then we'd never know. I'd be interested to know what you would have in mind for this. Though, many of these are easier (e.g. do outside map checks before occupation checks).

The planners now throw errors in the order stated above. 

I think that order above should be a _rough_ guideline. In the particular instance you mentioned I think even if we were able to collect all the exception at once and choose we would still need the handle the first exception thrown.

SteveMacenski:
Got it - well if they're all in that order, then I think the last thing is to switch that order in the message definition + make a note in the message that that is the expected priority order

nav2_msgs/action/ComputePathToPose.action
```diff
@@ -1,13 +1,14 @@
...
 int16 NONE=0
 int16 UNKNOWN=1
-int16 START_OCCUPIED=2
-int16 GOAL_OCCUPIED=3
-int16 START_OUTSIDE_MAP=4
-int16 GOAL_OUTSIDE_MAP=5
-int16 NO_VALID_PATH=6
-int16 TIMEOUT=7
-int16 TF_ERROR=8
+int16 TF_ERROR=2                #system failure
```
SteveMacenski, line 5: Oh, don't actually put those remarks!

SteveMacenski (approved):

nav2_behavior_tree/nav2_tree_nodes.xml
```diff
@@ -77,6 +77,7 @@
       <input_port name="service_name">Service name</input_port>
       <input_port name="server_timeout">Server timeout</input_port>
       <output_port name="path">Path created by ComputePathToPose node</output_port>
+      <output_port name="compute_path_to_pose_error_code"> "The compute path to pose error code</output_port>"
```
SteveMacenski, line 80:
```suggestion
      <output_port name="compute_path_to_pose_error_code">"The compute path to pose error code"</output_port>
```

PR #3158, opened 2022-08-30
Change deprecated ceres function

Tobias-Fischer:
error: 'bool ceres::IsInfinite(double)' is deprecated: ceres::IsInfinite will be removed in a future Ceres Solver release. Please use ceres::isinf.

### DIFF
--- a/nav2_constrained_smoother/include/nav2_constrained_smoother/smoother_cost_function.hpp
+++ b/nav2_constrained_smoother/include/nav2_constrained_smoother/smoother_cost_function.hpp
@@ -159,7 +159,7 @@ protected:
     Eigen::Matrix<T, 2, 1> center = arcCenter(
       pt_prev, pt, pt_next,
       next_to_last_length_ratio_ < 0);
-    if (ceres::IsInfinite(center[0])) {
+    if (ceres::isinf(center[0])) {
       return;
     }
     T turning_rad = (pt - center).norm();
--- a/nav2_constrained_smoother/include/nav2_constrained_smoother/utils.hpp
+++ b/nav2_constrained_smoother/include/nav2_constrained_smoother/utils.hpp
@@ -86,7 +86,7 @@ inline Eigen::Matrix<T, 2, 1> tangentDir(
   bool is_cusp)
 {
   Eigen::Matrix<T, 2, 1> center = arcCenter(pt_prev, pt, pt_next, is_cusp);
-  if (ceres::IsInfinite(center[0])) {  // straight line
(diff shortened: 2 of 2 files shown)


AlexeyMerzlyakov (approved):
LGTM: This API is [being updated](https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/jet.h#L1080) in ceres. In Nav2, there are no other functions from the deprecated ceres API anymore.

ozzdemir:
While building nav2_constrained_smoother (1.1.6-1) from source on a ubuntu 20.04 docker (which comes with libceres-dev 1.14.0), had the same error mentioned on the https://github.com/ros/rosdistro/pull/35270#issuecomment-1311057275 and to fix it, reverted the changes mentioned here (i.e. ceres::isinf -> ceres::IsInfinite).

While checking the [backward incompatible API change](http://ceres-solver.org/version_history.html#backward-incompatible-api-changes,) for the root cause, I observed that it proposes use of std counterparts. 

> Classification functions like IsFinite are deprecated. Use the **C++11 functions (isfinite, isnan etc)** going forward. However to maintain consistent behaviour with comparison operators, these functions only inspect the scalar part of the Jet.

I think ceres::isnan introduced with ceres 2.0 or later, which is only for internal uses. Wouldn't it be better to use the std counterparts in order to remain compatible with the both versions? What do you think @Tobias-Fischer @AlexeyMerzlyakov ?

AlexeyMerzlyakov:
@ozzdemir, I am not sure about whether the `ceres::isinf()` is an internal Ceres function, because for this Ceres has [internal](https://github.com/ceres-solver/ceres-solver/tree/master/include/ceres/internal) includes not intended to be used in public. However, since Ubuntu 20.04 is in [officially supported](https://docs.ros.org/en/rolling/Installation/Alternatives/Ubuntu-Development-Setup.html#system-requirements) platforms list for ROS2, the fix for it is welcoming.

Please note, that `nav2_constrained_smoother` is using generic indefinite templates methods, so we may need to define both instantiations of `std::isinf(x)`: for `T = Scalar (a.k.a. double)`  and for `T = ceres::Jet(D, N)` types. Something like simply:
```
template <typename D, int N> 
bool MyIsInfinite(const ceres::Jet<D, N>& j) {
  return std::isinf(j.a);
}

bool MyIsInfinite(const double x) {
  return std::isinf(x);
}
```

So, you are welcoming to open a new PR for this. Maybe, there are more elegant solution, I've lost sight of, pls let us know there.

Issue #2826, opened 2022-02-18
Feature request: Waypoint Publisher in Rviz Panel


girvenavery2022:
#### Feature description
The Waypoint Publisher feature would allow developers to give nav2 a set of waypoints taken in from a file and then allow Waypoint Follower to follow them in order. This will solve the issue of only being able to send a set of waypoints through the Rviz2 plugin

#### Implementation considerations
Right now, I have a working proof of theory package that needs some refactoring/clean up but has been tested in simulation on 2 separate robots in foxy/galactic as well as tested on a physcial robot in foxy. The way I implemented it was, the node takes in a CSV file of waypoints described in x, y, z, roll, pitch, and yaw. After it formats the points and pushes them into a vector, it then waits until an initial pose is set in the map frame before it sends the waypoints to the WayPoint follower.

padhupradheep:
This is the same exact feature, Me and Steve were speaking during the last weekend. Your idea seems to be good. But the way I had implemented was to directly store the information from the pre-existing navigation plugin in RViz (have some buttons on it "Load" and "save"). Then you can easily load and store the waypoints, wherever you want to. 

On top of this feature, I will also like to see a loop functionality with an count option, where the robot repeats the specified waypoints for the given count. 

All these features on RViz, would enhance the user friendliness and makes it much more easier for non-domain specific people who wants to use Nav2. Also, this really paves a way for semantic navigation as Steve said.

I'm going +1 for this feature.

girvenavery2022:
I actually really like the idea of the loop functionality with a counter (why didn't I think of that before?) and I am all for enhancing user friendless, so hooking into the RViz plugin would also be a great feature.

Since you implemented by directly storing the waypoints through the RViz plugin, and then allowing the user to load and store wherever. Could we combine both our ideas and allow the user to choose which one they would like? either through the navigation parameter file or through like the RViz plugin?

Anyways, Please let me know and I will get right onto this feature!

padhupradheep:
> Since you implemented by directly storing the waypoints through the RViz plugin, and then allowing the user to load and store wherever. Could we combine both our ideas and allow the user to choose which one they would like? 

Sure, we can do that. 

> either through the navigation parameter file or through like the RViz plugin?

[...]

PR #3165, opened 2022-09-01
Feature - load, save and loop waypoints

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #2826 ) |
| Robotic platform tested on | Gazebo simulation of Neobotix |

---

* Added new feature to load and save waypoints in a yaml file for waypoint following
* Added the loop functionality to revisit the waypoints
* Pause and resume button for stopping and continuing through the waypoints. 

- Add an example yaml file
- Description about the functionality

---

* Update the pause status in the plugin, currently the status reads cancelled when the robot is paused in between the waypoints (will be done as part of this PR). 

#### For Maintainers: 


padhupradheep:
Still have some work left such as updating the status for pausing in the RViz plugin and renaming the variables. Hope to complete it on the weekend.

nav2_msgs/action/FollowWaypoints.action
```diff
@@ -1,4 +1,6 @@
 #goal definition
+uint32 number_of_loops
+uint32 goal_index
```
SteveMacenski, line 3: What's `goal_index`? This doesn't seem to me like information to have sent

padhupradheep: It denotes the waypoint that the robot is currently in. Especially when you pause the waypoint navigation from the nav2_panel, you need to somehow let the waypoint follower know about the last waypoint that the robot was navigating to. Because, when you pause the navigation, it typically means you cancel the navigation. So, we store the information about the goal index through the feedback and we send it again to the waypoint follower. If you see this [line](https://github.com/ros-planning/navigation2/blob/634d2e3d9b6bde5558c90fe6583d52b3ed66cf55/nav2_waypoint_follower/src/waypoint_follower.cpp#L173) from the original code, the goal index is always set to 0. Now, we just make it a bit flexible by adding an option for the FollowWaypoints action, so that we get an option to define which waypoint the robot should navigate to and of course gives us the the ability to pause and resume the waypoint navigation.

This may also help a scenario, where an user has 100 waypoints. He just does not want his robot to start from the first waypoint, but from 30th waypoint and then navigate to the rest of the 70 waypoints.

nav2_msgs/action/FollowWaypoints.action
```diff
@@ -1,4 +1,6 @@
 #goal definition
+uint32 number_of_loops
```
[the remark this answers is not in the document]
padhupradheep: Initially I handled them in the panel level. I just appended the waypoints depending on the number of loops. So say for example you have three waypoints and 3 loops, we totally have 9 waypoints. When you start the navigation and the robot reaches the 1st waypoint again, then in the terminal you see the logs, stating that you are reaching 4th waypoint, which totally did not make sense, because it typically is the 1st waypoint and might cause some sort of confusion if you have large number of waypoints.

nav2_msgs/action/FollowWaypoints.action
SteveMacenski: That should not be in the request field - the first waypoint in the request is what it should do. I think this is a reason to not have the pause option

nav2_msgs/action/FollowWaypoints.action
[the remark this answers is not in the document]
SteveMacenski: > stating that the robot is reaching 4th waypoint

I think that's fine in context, but let me take a look at this again with a fresh mind that this is going to involve some changes in the WP follower. It just wasn't my initial expectation

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -261,14 +272,18 @@ WaypointFollower::followWaypoints()
...
+        if (current_loop_no == no_of_loops) {
+          RCLCPP_INFO(
+            get_logger(), "Completed all %zu waypoints requested.",
+            goal->poses.size());
+          result->missed_waypoints = failed_ids_;
+          action_server_->succeeded_current(result);
+          failed_ids_.clear();
+          return;
+        }
+        goal_index = 0;
```
SteveMacenski, line 285: Log that we're starting a new loop

SteveMacenski:
any update?

padhupradheep:
Not yet, still away on PTO, hopefully this should be up by mid October. Sorry for the delay.

SteveMacenski:
OK! Wow, what do I need to do to get that kind of PTO :wink: 

Hey, no worries, enjoy vacation! Sorry to bother.

padhupradheep:
FYI, I'm back from my PTO. Hopefully would finish it ASAP.

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -381,6 +508,119 @@ Nav2Panel::~Nav2Panel()
 {
 }
 
+void Nav2Panel::loophandler()
+{
+  loop = nr_of_loops->displayText().toStdString();
```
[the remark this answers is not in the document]
padhupradheep: This function connects with the text box where the user specify the number of loops. Just in case if the user leaves it empty, then the loop button is not enabled, thus the state remains as before.

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -170,10 +174,17 @@ WaypointFollower::followWaypoints()
   }
 
   rclcpp::WallRate r(loop_rate_);
-  uint32_t goal_index = 0;
+  uint32_t goal_index = goal->goal_index;
   bool new_goal = true;
 
   while (rclcpp::ok()) {
+    if (no_of_loops > 0) {
+      RCLCPP_INFO_EXPRESSION(
```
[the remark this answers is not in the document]
padhupradheep: sure will take a look

nav2_msgs/action/FollowWaypoints.action
padhupradheep: Alright, let me see if I can find a workaround.

padhupradheep:
Okay now handling everything in the panel level. Now in the below video you can see that I'm able to select waypoints, set number of loops and start waypoint following. 

Additionally you can also see, there is a status saying at which waypoint and at which loop the current navigation is happening, also when you press the pause button, you also get a note saying that the last cancelled waypoint is stored, which could be resumed. 

https://user-images.githubusercontent.com/20242192/195167389-8e4a3cd7-6af1-42fb-b56c-1cdeb7a34bb0.mp4

It is not still yet perfect (need to solve edge cases), but just wanted to know, whether will this additional status stuff be interesting for you / users ? what do you think ? 

(sorry for the poor quality video, for some reason my graphics card is behaving weird in the 22.04, which slows down RViz a lot.)

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -381,6 +508,119 @@ Nav2Panel::~Nav2Panel()
...
+
+void Nav2Panel::handleGoalSaver()
+{
+// Check if the waypoints are accumulated
+
+  if (acummulated_poses_.empty()) {
+    std::cout << "No accumulated Points to Save!" << std::endl;
+    return;
+  } else {
+    std::cout << "Clearing Waypoints!" << std::endl;
```
SteveMacenski, line 583: This log seems out of place?

SteveMacenski:
Yeah! Super useful! Question: if you have looping of waypoint following where you set the poses via rviz (not a file), does the initial robot point count as a starting point? That way if you loop you have the initial segment included as well? Should it? I'm not sure, but seems reasonable. 

Still not a fan of the `goal_index` as an entry in the waypoint follower - but I can live with it if we set it to `0` by default in the `msg` definition so if no one fills it in, it explicitly goes to use the full list

padhupradheep:
> if you have looping of waypoint following where you set the poses via rviz (not a file), does the initial robot point count as a starting point? 

No, If the user wants to use the initial starting point as well, then he has to select the initial starting pose as like the rest of the poses. I initially thought about adding the initial pose as part of the waypoint list, then anyways if the user wants it, then he's just a drag away. 

> That way if you loop you have the initial segment included as well? Should it? I'm not sure, but seems reasonable.

I feel, this depends more on the use-case that the user wants to handle. Say if the user is starting from a docking area, then he definitely does not want the starting point included. 

Maybe we can just handle both the cases by having a check box for selecting if the start pose needs to be included or not. Leaving the choice to the users. 

> Still not a fan of the goal_index as an entry in the waypoint follower - but I can live with it if we set it to 0 by default in the msg definition so if no one fills it in, it explicitly goes to use the full list

`goal_index` has been removed from the waypoint follower, I'm now handling everything in the panel level. I just haven't pushed the changes. :sweat_smile:

SteveMacenski:
For now, we can ignore the initial point

> goal_index has been removed from the waypoint follower, I'm now handling everything in the panel level. I just haven't pushed the changes. sweat_smile

Oh? It was starting to grow on me a bit :wink:

padhupradheep:
You can avoid looking at the recent commit and providing a review for it, still need to clean out few things. 

I'll let you know, once I'm done. 

Meanwhile, I added the check for including the `initial_pose`

Screenshot from 2022-10-12 19-40-08

SteveMacenski:
Nice!

padhupradheep:
> I do think the looping feature of the WP follower is valuable, I would actually like that logic back into the WPF server if possible.

No problem! 

Only thing is that, we also need the ‘goal_index’ back as well, that is very much essential for the waypoint follower to know about the pausing. As you said earlier in this thread I can set it’s default to 0.

SteveMacenski:
Ok!

nav2_waypoint_follower/src/waypoint_follower.cpp
padhupradheep: Do we need to log, when we can see the status of the loop on the RViz plugin? 

I feel we can safely skip it,

SteveMacenski: A log would be nice because this feature could be used outside of the rviz context.

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -423,6 +709,14 @@ Nav2Panel::onInitialize()
         navigation_feedback_indicator_->setText(getNavThroughPosesFeedbackLabel());
       }
     });
+
+  // create map_pose subscription for initial_pose
+  map_pose_sub_ = node->create_subscription<geometry_msgs::msg::PoseWithCovarianceStamped>(
+    "map_pose",
```
SteveMacenski, line 715: Who's publishing that?

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -170,7 +174,9 @@ WaypointFollower::followWaypoints()
   }
 
   rclcpp::WallRate r(loop_rate_);
-  uint32_t goal_index = 0;
+
+  // get the goal index
```
SteveMacenski, line 178:
```suggestion
  // get the goal index, by default, the first in the list of waypoints given.
```

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Oops my bad, I need to use `amcl_pose`. 

In neo_localization, we have a topic called `map_pose` publishing the pose of the robot on the map. 

Totally forgot about amcl!

SteveMacenski: In fact, SLAM Toolbox also publishes to the AMCL pose topic as well (just so it has drop in replacement functionality with rviz and other tools), you may consider that as well, even if its not "technically" amcl

padhupradheep:
Would be nice to have some user feedback, before marking it ready for review. 

Our nice friend @jwallace42 has agreed to test it and provide some feedback tomorrow. Meanwhile, I'll also test it a bit more, making sure, there are no edge cases.

jwallace42:
Looks pretty good :). Here is what I found. 

1. I can crash the waypoint server by trying to follow zero waypoints. We should protect against this and just fail the action right away. 
2. If I hit the pause waypoint button then cancel it stays stuck and I cannot start a new way point following. If I don't pause and just cancel, things work as expected. 
3. Before the robot is localized at the start the save wappoints, load wappoints and pause waypoints boxes have no text in them. 
4. Server crashes if I put a huge number for loops or something that is not a number :) I think you need to bound the inputs to only positive numbers within a range. 
5. I am also able to change the loop number in the text box during nav. if you can gray that out that would be good. 
6. Way points can be put in invalid places. (outside map, LETHAL area)
7. If I pause the waypoint, should be able to save the waypoints.

jwallace42:
Also, 

pull in main :).

padhupradheep:
Once again, thanks for the nice feedback. @jwallace42 

> I am also able to change the loop number in the text box during nav. if you can gray that out that would be good.

do you mean during nav_through_poses?

nav2_waypoint_follower/src/waypoint_follower.cpp
padhupradheep: BTW, this has been added few lines below

padhupradheep:
I have added couple of additional checks to verify the specified waypoints are valid and there are no obstacle in that particular point in the map and also checked if the waypoints are specified inside the map bounds. 

@jwallace42 I think I have handled all your points and most of the other edge cases, do you mind testing it once again and giving again your feedback if you have time? Your insights were really helpful last time around. Nothing urgent though, anyways Steve is away for ROSCon.

jwallace42:
@padhupradheep 
Yeah, 

I will test this today and if all goes well I will review the code :). 

Do you have a docs pr up?

padhupradheep:
> Can't check the store initial pose box.

Can you tell me at which state, that is happening. Usually you will be able to select it only when you have pressed the _waypoint / nav through poses mode_ 


> When start way point is given zero poses, it fails but the screen jumps back to the previous prompts screen. It should stay on the start nav through poses cancel accumulation and start waypoint following.

That is a nice point. It's something like, you are in the _accumulation_ state, where you accumulate waypoints and then when you press the **start waypoint following** you move to the _accumulated_wp_ state, this is the possible state where you can check if there are waypoints stored or not.  I might have an idea for the workaround, let me comeback. 

> I can put negative numbers in for loops and it will just take the abs value. I think you should fail and prompt like you did for inputs that are letters or other symbols.

Thinking of it now, this is actually a good point.

padhupradheep:
Coming back to this point 

> When start way point is given zero poses, it fails but the screen jumps back to the previous prompts screen. It should stay on the start nav through poses cancel accumulation and start waypoint following.

The easiest way to deal with this issue is that, we need to transition to `accumulating` after the `accumated_wp_` state becomes inactive. 

https://github.com/ros-planning/navigation2/blob/a50d45f1c275fdb557a8b0e4f89754887800f67a/nav2_rviz_plugins/src/nav2_panel.cpp#L426

Currently, this state usually goes to inactive, when the waypoint following has been completed, or if the goal status from the waypoint follower is anything other than accepted or executing. Also, in this PR, I make it inactive, when there are no waypoints accumulated or when there is an invalid loop value. 

`accumulating` state is the state where you accumulate goals for either waypoint following or nav through poses.

jwallace42:
> Can you tell me at which state, that is happening? Usually you will be able to select it only when you have pressed the waypoint / nav through poses mode

Played around with it a bit more. It is working as expected :). 

> That is a nice point. It's something like, you are in the accumulation state, where you accumulate way points and then when you press the start waypoint following you move to the accumulated_wp state, this is the possible state where you can check if there are waypoints stored or not. I might have an idea for the workaround, let me comeback.

This is a nice to have, but it is out of the scope of your pr :)(my bad). Rather than block this pr for that I can open a separate 
[issue](https://github.com/ros-planning/navigation2/issues/3249) for you.

jwallace42 (changes requested):
LGTM! 

Once you address the costmap changes we can wait for steve to review and then get this in. :)

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -113,6 +132,14 @@ private Q_SLOTS:
...
     nav_through_poses_goal_status_sub_;
 
+  // map_pose subscriber for initial pose
+  rclcpp::Subscription<geometry_msgs::msg::PoseWithCovarianceStamped>::SharedPtr
+    map_pose_sub_;
+
+  // global_costmap subscriber
+  nav2_costmap_2d::CostmapSubscriber * costmap_sub_;
+  std::shared_ptr<nav2_costmap_2d::Costmap2D> costmap_;
+
```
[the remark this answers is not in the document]
jwallace42: The `collision_checker_'` uses the costmap and the footprint polygon to determine if a pose is valid. 

Your current implementation does this but does not check the footprint. 

To reduce code, and prevent creating another instance of the costmap you can just use the `collision_checker_` by `costmap_sub_ = std::make_unique<nav2_costmap_2d::CostmapSubscriber>(
    shared_from_this(), costmap_topic);
  footprint_sub_ = std::make_unique<nav2_costmap_2d::FootprintSubscriber>(
    shared_from_this(), footprint_topic, *tf_, robot_base_frame, transform_tolerance);

  collision_checker_ = std::make_shared<nav2_costmap_2d::CostmapTopicCollisionChecker>(
    *costmap_sub_, *footprint_sub_, this->get_name())`
    
   I should have been a bit more careful with my line placement :).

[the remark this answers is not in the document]
padhupradheep: Gottcha!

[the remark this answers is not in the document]
padhupradheep: The only concern in adding the collision_checker_ for me is that, we then need to maintain configs for the rav2_rviz_plugins. This is mainly because some of the parameters in the collision_checker_ might be robot specific such as the robot_base_frame, footprint topic and transform_tolerance. This would actually add an extra configuration step for the user while setting up the robot. 

Maybe that is the way to go forward, if we want to make RViz a bit more user-friendly.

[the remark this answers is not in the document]
padhupradheep: We might also want to add the collision checker to the waypoint follower, for the people who use wp_follower standalone. Would be good for a new issue right @jwallace42 ?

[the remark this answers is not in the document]
jwallace42: Yeah that sounds good :).

padhupradheep:
This is gtg for a review.  @SteveMacenski 

Thanks to @jwallace42, he had found some super nice edge cases, which has been all fixed.

SteveMacenski:
Its my intention to play around with this and review it tomorrow, sorry for the delay. I've been getting a bunch of small to medium sized features/bugs done after ROSCon that were pending to get some half-done stuff off of my plate to start fresh before the holidays

SteveMacenski:
The plugin is alot wider than before that I cannot shrink. Maybe split things onto 2 lines?

Screenshot from 2022-11-09 17-14-56

Store initial pose doesn't seem to be working properly It appears to do some action at the initial pose right when we request the WPF task, rather than doing it on the loop (it shouldn't do anything on the initial pose when we're there already and that's just the starting pose we have for requesting a loop). The first iteration when we start at that pose, we shouldn't do anything. Its only on future iteration we do so. 

Also, rviz crashed on me due to the changes here when I requested a number of goals / task to test different settings.  It seems like the state of the application UX and the action server got out of sync. When I was sending a waypoint task, the terminal showed a cancelation request. 

On saving, if the user doesn't specify yaml, you should add a `.yaml` extension on the end. When  I type "hi" it should come out as "hi.yaml"

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -261,20 +267,20 @@ WaypointFollower::followWaypoints()
...
+      if (goal_index == goal->poses.size()) {
+        if (current_loop_no == no_of_loops) {
+          RCLCPP_INFO(
+            get_logger(), "Completed all %zu waypoints requested.",
+            goal->poses.size());
+          result->missed_waypoints = failed_ids_;
+          action_server_->succeeded_current(result);
+          failed_ids_.clear();
+          return;
+        }
```
SteveMacenski, line 279: Use actual RCLCPP_INFO logs please so they appear in the log files

padhupradheep:
Thanks for the feedback!

image

I added a group box for all the waypoint following tools. Later, when the necessity comes, we could make a collapsible group box.

Only issue now is, we cannot adjust the height of the panel :frowning_face:

SteveMacenski:
Cant the buttons be put on 1 line and the load/save/initial_pose on another?

padhupradheep:
> Cant the buttons be put on 1 line and the load/save/initial_pose on another?

Sure, we can do that. We need to be aware that, these buttons are multi-purpose. For example, the pause button, expands to "Start Waypoint following", which automatically expands, when this state is reached.  One immediate option would be to have short forms like "Start WpF", "Nav to Pos", "Cancel Accu" then it would make sense to bring those into a single line. 

I'll get a bit deep into QT and explore some more options.. 

> It only happened 1 time, but I was only playing with it for like 10 minutes

Hmm..!!

> Also, please don't force push, it deletes the github memory and makes it very difficult for me to review since I can't see diffs of just what you changed.

Oops, I never pushed my changes yet, I only rebased and force pushed.. I think, except for the panel customization, I'll push the rest now.. 

I think, this feature would be for Iron.. So, I feel, we can take some more time and provide a better interface..

SteveMacenski:
> For example, the pause button, expands to "Start Waypoint following", which automatically expands, when this state is reached. One immediate option would be to have short forms like "Start WpF", "Nav to Pos", "Cancel Accu" then it would make sense to bring those into a single line.

I'm confused - why would pause button even be non-greyed out until you're in the waypoint following/nav through poses menu? 

I'm proposing leaving what's in the plugin before this PR the same but add 2 lines: (1) Save/Load/Pause buttons (2) Loops / Initial Pose radio

padhupradheep:
Got it! It’s clear now, sorry I got you all wrong 😅

padhupradheep:
Does this workout? This is the smallest size that we can have. 

image

SteveMacenski:
Yup! Though the `WP's` should not have the apostrophe + `Number of Loops` can be `Num of Loops`

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -381,6 +567,189 @@ Nav2Panel::~Nav2Panel()
...
+  const YAML::Node & waypoint_iter = available_waypoints["waypoints"];
+  for (YAML::const_iterator it = waypoint_iter.begin(); it != waypoint_iter.end(); ++it) {
+    auto waypoint = waypoint_iter[it->first.as<std::string>()];
+    auto pose = waypoint["pose"].as<std::vector<double>>();
+    auto orientation = waypoint["orientation"].as<std::vector<double>>();
+    acummulated_poses_.push_back(convert_to_msg(pose, orientation));
+  }
+
+  // Check for goal poses, if they are not in an obstacle
+  for (unsigned int i = 0; i < acummulated_poses_.size(); ++i) {
```
SteveMacenski, line 669: Remove this - lets just make this application "dumb"

SteveMacenski: + remove the collision checker / costmap / footprint / etc subscribers.

Do we need TF?

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -566,15 +1053,57 @@ Nav2Panel::onCancelButtonPressed()
...
+  navigation_mode_button_->setEnabled(false);
+  pause_resume_button_->setEnabled(false);
+
   std::cout << "Start waypoint" << std::endl;
+
+  // storing and removing the initial pose based on checkbox state
+  if (store_poses_.size() == 0) {
+    // Setting the final loop value on the text box for sanity
+    nr_of_loops_->setText(QString::fromStdString(loop_no_));
+    if (store_initial_pose_) {
```
SteveMacenski, line 1088: This is not a robust way to get the initial pose. Use TF

SteveMacenski (changes requested):

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Sure, we can scale it down. The reason I assume that you didn't want to have it is maybe because collision_checker might make the plugin a bit heavy ?

padhupradheep:
Hopefully one final thing, the logo that is used in readme is bit bigger, and while rescaling it using the QT, it gets a bit pixelated.. Actually the one you see on the panel above is a rescaled one from another tool. Is it okay, if we can have another small sized logo included in the repo? I'll add it to the doc folder.

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: Yeah as well as the case that it may be likely pre-defined paths could have blocked waypoints, but I think that's for the WPF to handle, not the visualization / command-control application's role

SteveMacenski:
> Is it okay, if we can have another small sized logo included in the repo?

Yeah I think that's fine. I have the original files but yeah making it too high resolution is just wasteful in memory. 

Anything else here I should wait for before trying it out again?

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -113,6 +131,10 @@ private Q_SLOTS:
   rclcpp::Subscription<nav2_msgs::action::NavigateThroughPoses::Impl::GoalStatusMessage>::SharedPtr
     nav_through_poses_goal_status_sub_;
 
+  // Tf's for initial pose
+  std::shared_ptr<tf2_ros::Buffer> tf2_buffer;
```
SteveMacenski, line 135: tf2_buffer_

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -143,6 +174,12 @@ private Q_SLOTS:
   QState * reset_{nullptr};
   QState * paused_{nullptr};
   QState * resumed_{nullptr};
+  QState * paused_wp_{nullptr};
+  QState * resumed_wp_{nullptr};
+
+  QImage * image_{nullptr};
+  QLabel * imgDisplayLabel{nullptr};
```
SteveMacenski, line 181: Missing underscore trailing

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -261,20 +267,22 @@ WaypointFollower::followWaypoints()
       // Update server state
       goal_index++;
       new_goal = true;
-      if (goal_index >= goal->poses.size()) {
+      if (goal_index == goal->poses.size()) {
```
SteveMacenski, line 270: >= in case there's some issues where it increments too fast, the stuff involving action servers and async results are stable right now, but doesn't mean there can't be regressions in rclcpp that this could help protect us from

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -566,15 +994,82 @@ Nav2Panel::onCancelButtonPressed()
...
+    nr_of_loops_->setText(QString::fromStdString(loop_no_));
+
+    // Variable to store initial pose
+    geometry_msgs::msg::TransformStamped init_transform;
+
+    // Looking up transform to get initial pose
+    if (store_initial_pose_) {
+      try {
+        init_transform = tf2_buffer->lookupTransform(
+          "map", "base_footprint",
```
SteveMacenski, line 1037: I don't suppose there are other things in this plugin we can base these off of instead of hard coding them? Maybe somewhere something coming in has a map and/or base_footprint frame that can be stored for use here?

padhupradheep: We can definitely get the `map` frame.. But I'm not sure if there is a way to get the base frame..

padhupradheep:
> Anything else here I should wait for before trying it out again? 

Just trying to figure out an automated way for bring in the `base_footprint` for lookup..

> Will look at the state machine updates in more detail while sitting in front of the application.

That would be real nice, I just can't able to locate a mistake, since I'm so use to this code..

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: OK - lets at least do map, I see other functions for navigation before this PR gets map frame to send navigation requests. 

Maybe expose the base_frame as a parameter so that it can be overrided if someone likes, but yeah I doubt anyone will functionally do that. More about making it possible.

SteveMacenski:
Yeah, please just go over it in full one time to make sure you give it the thumbs up and then I'll take a look as well.

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Sure, I'll do that.. 

-----------

Maybe somewhere in the future, we might want to have a single RViz for multi-robot system, then we might need a huge overhaul of the complete plugin with the namespaces.. 

We already have a [plugin](https://github.com/neobotix/neo_fleet_rviz2_plugin/tree/cleanup) for selecting a robot and sending a goal for that particular robot.. Of course, all done in a single RViz.. but it certainly lacks the features that are present right now in the nav2_plugin..

SteveMacenski: > we might want to have a single RViz for multi-robot system

That's been suggested before and I've generally been opposed to it being a requirement / task for maintainers. If someone made a new rviz plugin for multi-robot, I'd be happy to review/merge it, but its not the kind of thing that the maintainers are going to work on and its a pipe-dream feature request in the issue tracker unless the person proposing it is going to do it themselves. I try to keep the issue tracker to actionable features (and obviously all bugs) which will realistically be accomplished eventually

padhupradheep:
Okay, I'm now giving a thumbs up after going through the code few more times..

Please give it a spin, I have tried all the possible things that we could do with it, I cannot localize any issues..

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -153,11 +223,24 @@ Nav2Panel::Nav2Panel(QWidget * parent)
...
+
+  accumulated_wp_->assignProperty(nr_of_loops_, "enabled", false);
+  accumulated_wp_->assignProperty(store_initial_pose_checkbox_, "enabled", false);
+
   accumulated_nav_through_poses_ = new QState();
   accumulated_nav_through_poses_->setObjectName("accumulated_nav_through_poses");
   accumulated_nav_through_poses_->assignProperty(start_reset_button_, "text", "Cancel");
   accumulated_nav_through_poses_->assignProperty(start_reset_button_, "toolTip", cancel_msg);
   accumulated_nav_through_poses_->assignProperty(start_reset_button_, "enabled", true);
+  accumulated_nav_through_poses_->assignProperty(store_initial_pose_checkbox_, "enabled", false);
```
SteveMacenski, line 243: Should this state have some specification for the Save/Load waypoints buttons? I would think they should be disabled.

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -207,20 +309,75 @@ Nav2Panel::Nav2Panel(QWidget * parent)
...
+  paused_->assignProperty(save_waypoints_button_, "text", "Save WPs");
+  paused_->assignProperty(save_waypoints_button_, "enabled", false);
+
+  paused_->assignProperty(load_waypoints_button_, "text", "Load WPs");
+  paused_->assignProperty(load_waypoints_button_, "enabled", false);
+
+  paused_->assignProperty(pause_waypoint_button_, "text", "Pause WP");
+  paused_->assignProperty(pause_waypoint_button_, "enabled", false);
+
+  paused_->assignProperty(store_initial_pose_checkbox_, "enabled", false);
```
SteveMacenski, line 321: Should the paused state have an explicit policy for the number of loops text field?

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -409,8 +808,16 @@ Nav2Panel::onInitialize()
     [this](const action_msgs::msg::GoalStatusArray::SharedPtr msg) {
       navigation_goal_status_indicator_->setText(
         getGoalStatusLabel(msg->status_list.back().status));
-      if (msg->status_list.back().status != action_msgs::msg::GoalStatus::STATUS_EXECUTING) {
-        navigation_feedback_indicator_->setText(getNavToPoseFeedbackLabel());
```
SteveMacenski, line 413: Shouldn't we still reset this to be blank at the end of the task?

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -515,13 +940,20 @@ Nav2Panel::onNewGoal(double x, double y, double theta, QString frame)
   pose.pose.position.z = 0.0;
   pose.pose.orientation = orientationAroundZAxis(theta);
 
-  if (state_machine_.configuration().contains(accumulating_)) {
-    acummulated_poses_.push_back(pose);
+  if (store_poses_.empty()) {
```
SteveMacenski, line 943: Why the check on store poses here?

SteveMacenski: Is there a less fragile way to check what you need to do here?

SteveMacenski (changes requested):

SteveMacenski:
Otherwise, I played with the functionality and very happy with it!

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Yes "enabled" but false .. which is disabled.. seems bit funny though..

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Oops! Yes! Nice catch!

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Yes, I guess that's just the same confusion as the above..

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: This is kind of essential. We do not want to have any accidentally selected waypoints making it's way into the waypoint list. 

I'm bit unsure why you say it fragile?

padhupradheep:
just rebasing, please ignore..

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -170,7 +174,9 @@ WaypointFollower::followWaypoints()
   }
 
   rclcpp::WallRate r(loop_rate_);
-  uint32_t goal_index = 0;
+
+  // get the goal index, by default, the first in the list of waypoints given.
+  uint32_t goal_index = goal->goal_index;
```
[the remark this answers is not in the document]
padhupradheep: That was the discussion which we apparently had few months ago. We do need this to recover the waypoint where the goal was paused..

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Currently, we do not show how long did the robot took to complete the entire waypoint following.. We just show the time taken for a specific waypoint and reset the time and move to the next. Maybe we need to have "total time taken" for showing how long the robot took to complete... 

Also I think it would be a nice idea for logging the time taken for each waypoint, which might be used for some sort of analysis later.. 

I think this should be handled as a separate issue..

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Because we go the pause state and return back to the accumulated state.. In between, we should make sure that we don't update the pose array with the mistakenly given goals from the RViz..

nav2_waypoint_follower/src/waypoint_follower.cpp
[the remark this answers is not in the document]
SteveMacenski: That's preempted with a new request while the current request is active. It could be from an entirely new set of goals - why would we ignore one of the fields for a new request? This isn't just for the RViz plugin tool if its in this server

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: I don't see any reference to `save_waypoints_button_` or `load_waypoints_button_` under the `accumulated_nav_through_poses_` state

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: I don't see any policy for `nr_of_loops_` under `paused_`

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: > Currently, we do not show how long did the robot took to complete the entire waypoint following.. We just show the time taken for a specific waypoint and reset the time and move to the next. Maybe we need to have "total time taken" for showing how long the robot took to complete...

That would be a good idea, but I think that should be done in a followup PR since that would involve editing the waypoint follower message definition further to provide that information 

> Also I think it would be a nice idea for logging the time taken for each waypoint, which might be used for some sort of analysis later..

I'm not sure that's the role of this plugin, but we could have the waypoint follower's results provide that information! Also another follow up PR thing so that we don't bloat this too much (and so this can get in to break apart the changes for easier reviewing)

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: I thought you may have been using that as a proxy for the state, this is fine

nav2_waypoint_follower/src/waypoint_follower.cpp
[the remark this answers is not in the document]
padhupradheep: Yes you are right, we just send a completely new request after we pause and restart the waypoint following. During pausing we just store the index of the goal to which the robot was navigating.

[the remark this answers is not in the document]
SteveMacenski: But when you pause, I think you actually cancel and then reissue later the WPF command? That's fine with how things work right now. The change I'm requesting is for preemption, which is not currently exercised by that workflow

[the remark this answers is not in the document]
padhupradheep: Got it! You meant L197! My bad, I some was stuck in L179!

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Ah! Got what you meant! 

The logic is somewhere here:

https://github.com/ros-planning/navigation2/blob/3d612b2af42daebdc6099541acde85124c0b3c50/nav2_rviz_plugins/src/nav2_panel.cpp#L651-L656

padhupradheep: I get your point, 

seems that it is taken care here:

https://github.com/ros-planning/navigation2/blob/3d612b2af42daebdc6099541acde85124c0b3c50/nav2_rviz_plugins/src/nav2_panel.cpp#L235

when you go into the `accumulated_wp_` state it is already disabled. Since pausing comes after that, the `nr_of_loops_` never gets enabled.

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Sorry got confused! 

This needs to be implemented..

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Sure! I can add it to the issue tracker! If that's okay!

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: I know, but being explicit about all of the states of all the buttons in all of the states in the machine is useful to make sure no later things create odd conditions

SteveMacenski (approved):
I'm happy with this! Anything else before merging?

You should post on discourse about this!

padhupradheep:
> I'm happy with this! Anything else before merging?

I was actually looking for something to update in docs. But I think the plugin is easy to understand, there is not much of a trick there. 

> You should post on discourse about this!

Sure, I'll take it to our community..

SteveMacenski:
I think a migration guide entry just to acknowledge the changes would be good! 

I'm also going to re-run CI, I see the WPF test failed and given there are substantive changes there, its important that passes

SteveMacenski:
@padhupradheep definitely post on discourse with a nice video / gif of this is action!

padhupradheep:
> definitely post on discourse with a nice video / gif of this is action!

Sure! 

Thanks for the consistent motivation and bearing with me through the feature design. :smiley:

SteveMacenski:
No worries at all, glad to have this in! Only took... 3 months :laughing:

PR #3168, opened 2022-09-02
fixed start

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3085 |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/node_2d.cpp
+++ b/nav2_smac_planner/src/node_2d.cpp
@@ -154,13 +154,16 @@ bool Node2D::backtracePath(CoordinateVector & path)
 
   NodePtr current_node = this;
 
-  do {
+  while (current_node->parent) {
     path.push_back(
       Node2D::getCoords(current_node->getIndex()));
     current_node = current_node->parent;
-  } while (current_node->parent);
+  }
 
-  return path.size() > 0;
+  // add the start pose
+  path.push_back(Node2D::getCoords(current_node->getIndex()));
+
+  return true;
 }
 
 }  // namespace nav2_smac_planner
--- a/nav2_smac_planner/src/node_hybrid.cpp
+++ b/nav2_smac_planner/src/node_hybrid.cpp
@@ -705,14 +705,19 @@ bool NodeHybrid::backtracePath(CoordinateVector & path)
 
   NodePtr current_node = this;
 
-  do {
+  while (current_node->parent) {
     path.push_back(current_node->pose);
     // Convert angle to radians
     path.back().theta = NodeHybrid::motion_table.getAngleFromBin(path.back().theta);
     current_node = current_node->parent;
-  } while (current_node->parent);
+  }
 
-  return path.size() > 0;
+  // add the start pose
+  path.push_back(current_node->pose);
+  // Convert angle to radians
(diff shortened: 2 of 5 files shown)


nav2_smac_planner/src/node_2d.cpp
```diff
@@ -154,13 +154,16 @@ bool Node2D::backtracePath(CoordinateVector & path)
...
       Node2D::getCoords(current_node->getIndex()));
     current_node = current_node->parent;
-  } while (current_node->parent);
+  }
+
+  // add the start pose
+  path.push_back(Node2D::getCoords(current_node->getIndex()));
 
-  return path.size() > 0;
+  return true;
```
SteveMacenski, line 166: Rever to path.size() > 0

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -705,14 +705,19 @@ bool NodeHybrid::backtracePath(CoordinateVector & path)
...
-  } while (current_node->parent);
+  }
+
+  // add the start pose
+  path.push_back(current_node->pose);
+  // Convert angle to radians
+  path.back().theta = NodeHybrid::motion_table.getAngleFromBin(path.back().theta);
 
-  return path.size() > 0;
+  return true;
```
SteveMacenski, line 720: Rever to path.size() > 0

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -541,44 +541,53 @@ bool NodeLattice::backtracePath(CoordinateVector & path)
...
-      // For analytic expansion nodes where there is no valid motion primitive
-      path.push_back(current_node->pose);
-      path.back().theta = NodeLattice::motion_table.getAngleFromBin(path.back().theta);
+      path.push_back(prim_pose);
     }
-
-    current_node = current_node->parent;
-  } while (current_node->parent);
-
-  return path.size() > 0;
```
SteveMacenski, line 581: Rever to path.size() > 0 -- also looks like a return here generally was dropped?

Why is this change so much more involves than the others?

nav2_smac_planner/src/node_2d.cpp
jwallace42: So, the path size will always be greater than 0 because we check 
` if (!this->parent) {
    return false;
  }`
We will always at least two points. The first one from the while and second from the start.

nav2_smac_planner/src/node_lattice.cpp
jwallace42: This got broken out into a seperate func to avoid coping 20ish lines of code.

nav2_smac_planner/src/node_2d.cpp
SteveMacenski: Oh - then > 1?

SteveMacenski: Very good point :smile: 

Apparently my defensive programming is defensive about defense :laughing:

SteveMacenski (approved):

SteveMacenski:
Thanks! Looks great

nav2_smac_planner/src/node_2d.cpp
jwallace42: :)

Issue #3149, opened 2022-08-25
Use SetParameter Launch API to set the yaml filename for map_server


SteveMacenski:
Would let us remove the RewrittenYaml remappings entirely and let us remove the requirement of a bogus `yaml_filename` entry in the yaml file to be overwritten. Now, it could just not be present at all and that would be fine. Just need to make sure that this is set only for the `map_server` node, which should be straight forward. 

See https://github.com/ros-planning/navigation2/pull/3131 for a motivating example and fixes https://github.com/ros-planning/navigation2/issues/3111 to be less quirky @superseppl. Thanks @tonynajjar once again for a timely PR that helps fix another problem

Note that we should also remove the `note` I added to the navigation.ros.org configuration for map server explaining that `yaml_filename` is strictly required.

stevedanomodolor:
@SteveMacenski I can try it this weekend

stevedanomodolor:
This might sound like a silly question but just to confirm. When you mean by

> Just need to make sure that this is set only for the `map_server` node, which should be straight forward.

If the set parameter is set for only the map_server node, what about other node that takes uses the params. For example the following launch file from the localization launch file 

```
load_nodes = GroupAction(
        condition=IfCondition(PythonExpression(['not ', use_composition])),
        actions=[
            SetParameter("use_sim_time", use_sim_time),
Node(
                package='nav2_map_server',
                executable='map_server',
                name='map_server',
                output='screen',
                respawn=use_respawn,
                respawn_delay=2.0,
                parameters=[configured_params],
                arguments=['--ros-args', '--log-level', log_level],
                remappings=remappings),
            Node(
                package='nav2_amcl',
                executable='amcl',
                name='amcl',
                output='screen',
                respawn=use_respawn,
                respawn_delay=2.0,
                parameters=[configured_params],
                arguments=['--ros-args', '--log-level', log_level],
                remappings=remappings),
            Node(
                package='nav2_lifecycle_manager',
                executable='lifecycle_manager',
                name='lifecycle_manager_localization',
                output='screen',
                arguments=['--ros-args', '--log-level', log_level],
                parameters=[{'autostart': autostart},
                            {'node_names': lifecycle_nodes}])
```
Similar to https://github.com/ros-planning/navigation2/pull/3131, would it not be also effective to set the yaml filename similarly as the  use_sim_time is set?

SteveMacenski:
Awesome!

PR #3174, opened 2022-09-06
Use SetParameter Launch API to set the yaml filename for map_server

stevedanomodolor:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #3149 ) |
| Robotic platform tested on | (gazebo)|

---

* Use SetParameter Launch API to set the yaml filename for map_server 

* Remove the note added to the navigation.ros.org configuration for map server explaining that yaml_filename is strictly required.

---

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/bringup_launch.py
+++ b/nav2_bringup/launch/bringup_launch.py
@@ -53,14 +53,10 @@ def generate_launch_description():
     remappings = [('/tf', 'tf'),
                   ('/tf_static', 'tf_static')]
 
-    # Create our own temporary YAML files that include substitutions
-    param_substitutions = {
-        'yaml_filename': map_yaml_file}
-
     configured_params = RewrittenYaml(
         source_file=params_file,
         root_key=namespace,
-        param_rewrites=param_substitutions,
+        param_rewrites={},
         convert_types=True)
 
     stdout_linebuf_envvar = SetEnvironmentVariable(
@@ -83,6 +79,7 @@ def generate_launch_description():
 
     declare_map_yaml_cmd = DeclareLaunchArgument(
         'map',
+        default_value='',
         description='Full path to map yaml file to load')
 
     declare_use_sim_time_cmd = DeclareLaunchArgument(
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -155,7 +155,12 @@ def generate_launch_description():
     # yaml configuration file. They are separated since the conditions
     # currently only work on the LoadComposableNodes commands and not on the
     # ComposableNode node function itself
-    # See https://github.com/ros2/launch_ros/issues/328
+    # EqualsSubstitution and NotEqualsSubstitution susbsitutions was recently
+    # added to solve this problem but it has not been ported yet to
+    # ros-rolling. See https://github.com/ros2/launch_ros/issues/328.
+    # LaunchConfigurationEquals and LaunchConfigurationNotEquals are scheduled
+    # for deprecation once a Rolling sync is conducted. Switching to this new
+    # would be required for both ComposableNode and normal nodes.
     load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
         actions=[
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -17,8 +17,11 @@ import os
 from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, GroupAction, SetEnvironmentVariable
+from launch.actions import DeclareLaunchArgument, GroupAction
+from launch.actions import SetEnvironmentVariable
 from launch.conditions import IfCondition
+from launch.conditions import LaunchConfigurationEquals
+from launch.conditions import LaunchConfigurationNotEquals
 from launch.substitutions import LaunchConfiguration, PythonExpression
 from launch_ros.actions import LoadComposableNodes, SetParameter
 from launch_ros.actions import Node
@@ -51,14 +54,10 @@ def generate_launch_description():
     remappings = [('/tf', 'tf'),
                   ('/tf_static', 'tf_static')]
 
-    # Create our own temporary YAML files that include substitutions
-    param_substitutions = {
-        'yaml_filename': map_yaml_file}
-
     configured_params = RewrittenYaml(
         source_file=params_file,
         root_key=namespace,
-        param_rewrites=param_substitutions,
+        param_rewrites={},
         convert_types=True)
 
     stdout_linebuf_envvar = SetEnvironmentVariable(
@@ -71,6 +70,7 @@ def generate_launch_description():
 
     declare_map_yaml_cmd = DeclareLaunchArgument(
         'map',
+        default_value='',
         description='Full path to map yaml file to load')
 
     declare_use_sim_time_cmd = DeclareLaunchArgument(
@@ -108,6 +108,7 @@ def generate_launch_description():
         actions=[
             SetParameter("use_sim_time", use_sim_time),
             Node(
+                condition=LaunchConfigurationEquals('map', ''),
                 package='nav2_map_server',
                 executable='map_server',
                 name='map_server',
@@ -115,6 +116,18 @@ def generate_launch_description():
                 respawn=use_respawn,
                 respawn_delay=2.0,
                 parameters=[configured_params],
+                arguments=['--ros-args', '--log-level', log_level],
+                remappings=remappings),
+            Node(
+                condition=LaunchConfigurationNotEquals('map', ''),
+                package='nav2_map_server',
+                executable='map_server',
+                name='map_server',
+                output='screen',
+                respawn=use_respawn,
+                respawn_delay=2.0,
+                parameters=[configured_params,
+                            {'yaml_filename': map_yaml_file}],
                 arguments=['--ros-args', '--log-level', log_level],
                 remappings=remappings),
             Node(
@@ -137,13 +150,19 @@ def generate_launch_description():
                             {'node_names': lifecycle_nodes}])
         ]
     )
-
+    # LoadComposableNode for map server twice depending if we should use the
+    # value of map from a CLI or launch default or user defined value in the
+    # yaml configuration file. They are separated since the conditions
+    # currently only work on the LoadComposableNodes commands and not on the
+    # ComposableNode node function itself
+    # See https://github.com/ros2/launch_ros/issues/328
     load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
         actions=[
             SetParameter("use_sim_time", use_sim_time),
             LoadComposableNodes(
                 target_container=container_name,
+                condition=LaunchConfigurationEquals('map', ''),
                 composable_node_descriptions=[
                     ComposableNode(
                         package='nav2_map_server',
@@ -151,6 +170,24 @@ def generate_launch_description():
                         name='map_server',
                         parameters=[configured_params],
                         remappings=remappings),
+                ],
+            ),
+            LoadComposableNodes(
+                target_container=container_name,
+                condition=LaunchConfigurationNotEquals('map', ''),
+                composable_node_descriptions=[
+                    ComposableNode(
+                        package='nav2_map_server',
+                        plugin='nav2_map_server::MapServer',
+                        name='map_server',
+                        parameters=[configured_params,
+                                    {'yaml_filename': map_yaml_file}],
+                        remappings=remappings),
+                ],
+            ),
+            LoadComposableNodes(
+                target_container=container_name,
+                composable_node_descriptions=[
                     ComposableNode(
                         package='nav2_amcl',
                         plugin='nav2_amcl::AmclNode',
--- a/nav2_bringup/params/nav2_params.yaml
+++ b/nav2_bringup/params/nav2_params.yaml
@@ -243,11 +243,12 @@ global_costmap:
         inflation_radius: 0.55
       always_send_full_costmap: True
 
-map_server:
-  ros__parameters:
-    # Overridden in launch by the "map" launch configuration or provided default value.
-    # To use in yaml, remove the default "map" value in the tb3_simulation_launch.py file & provide full path to map below.
-    yaml_filename: ""
+# The yaml_filename does not need to be specified since it going to be set by defaults in launch.
+# If you'd rather set it in the yaml, remove the default "map" value in the tb3_simulation_launch.py
+# file & provide full path to map below. If CLI map configuration or launch default is provided, that will be used.
+# map_server:
+#   ros__parameters:
+#     yaml_filename: ""
 
 map_saver:
   ros__parameters:


stevedanomodolor:
@SteveMacenski I am still not sure if a group action is required in order to use the setparameter or it can be used under the Node action directly, there is not that much documentation.  I have to test it to see if it is necessary. I have not tested this commit thats why its on draft.

nav2_bringup/launch/localization_launch.py
```diff
@@ -51,16 +51,6 @@ def generate_launch_description():
     remappings = [('/tf', 'tf'),
                   ('/tf_static', 'tf_static')]
 
-    # Create our own temporary YAML files that include substitutions
-    param_substitutions = {
-        'yaml_filename': map_yaml_file}
-
-    configured_params = RewrittenYaml(
```
SteveMacenski, line 58: Still need this for injecting namespaces

nav2_bringup/launch/bringup_launch.py
```diff
@@ -55,7 +55,7 @@ def generate_launch_description():
 
     # Create our own temporary YAML files that include substitutions
     param_substitutions = {
-        'yaml_filename': map_yaml_file}
+        }
```
SteveMacenski, line 58: I think you can just remove the `param_substitutions` object

nav2_bringup/launch/slam_launch.py
```diff
@@ -42,13 +42,6 @@ def generate_launch_description():
     slam_toolbox_dir = get_package_share_directory('slam_toolbox')
     slam_launch_file = os.path.join(slam_toolbox_dir, 'launch', 'online_sync_launch.py')
 
-    # Create our own temporary YAML files that include substitutions
-    configured_params = RewrittenYaml(
```
SteveMacenski, line 46: Same - still needed :-)

nav2_bringup/launch/slam_launch.py
```diff
@@ -89,7 +82,7 @@ def generate_launch_description():
                 respawn=use_respawn,
                 respawn_delay=2.0,
                 arguments=['--ros-args', '--log-level', log_level],
-                parameters=[configured_params]),
+                parameters=[params_file]),
```
SteveMacenski, line 85: So this can be reverted too!

SteveMacenski:
if this all works out - you can also remove the `yaml_filename` from the yaml files. Leave it there, but just comment it out and update the comment above it explaining that it doesn't need to be specified since going to be set by defaults in launch. If you'd rather set it in the yaml, remove the defaults in launch and set the full filepath here. 

And subsequent updates in docs

Thanks so much for the help! And @tonynajjar for the idea!

nav2_bringup/launch/localization_launch.py
tonynajjar: @SteveMacenski  take a look at https://github.com/ros2/rclcpp/pull/1839, I think with this you could remove the RewrittenYaml (for injecting a namespace at least)

stevedanomodolor: I am not then required to leave it then?, also it is worth mentioning that the autostart parameter for the nav2_container still uses the RewrittenYaml.

tonynajjar: This would need some additional work, it's not for this PR imo. So you still need it for now.

stevedanomodolor:
@SteveMacenski @tonynajjar. i have tried setting the parameters but still get the same error 

`[ERROR] [launch]: Caught exception in launch (see debug for traceback): 'NoneType' object is not iterable
`
I am supposed to declare the yaml_filename in a specific way. There is not too much documentation on it.

SteveMacenski:
What if yaml_filename is non-empty string? Then does it work? I'm curious if its failing because its just not set (and then we could use a launch action condition to only set the parameter if its set to something).

stevedanomodolor:
You mean the yaml_filename in the launch file or in the parameter file?

stevedanomodolor:
> What if yaml_filename is non-empty string? Then does it work? I'm curious if its failing because its just not set (and then we could use a launch action condition to only set the parameter if its set to something).
> 
> It doesn't seem to be that there needs to be a declaration: https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/actions/set_parameter.py#L38-L55 but it does seem to me like it might need to exist

I am performing all test with the tb3_simulation_launch.py with ros rolling(used for development)
What if yaml_filename is non-empty string?  ->  I am assuming you are referring to the yaml_filename in the parameter file.
When the yaml_filename parameter in the parameter list file is set to an empty string , the launch file execute with no map independent of if the setparameter was set or not. In this case the param substitution is not used. And when the yaml_filename parameters is not set in the parameter file and the param substitution not used, it gives the error nonetype as said  previously. I also tried setting the default parameters in the launch file to non empty string but still gives the nonetype error. The set parameter does not seem to influence at all the behavior.

SteveMacenski:
>  What if yaml_filename is non-empty string? -> I am assuming you are referring to the yaml_filename in the parameter file.

I meant in the launch file declaration, so that when you call `SetParameter()` there's a non-empty value to it

I don't think the type has anything to do with it. @tonynajjar any thoughts?

stevedanomodolor:
@SteveMacenski The issue was from my side. I had an issue with the docker, I have fixed it now.

tonynajjar:
Great, I was about to have a look. If you have another issue, please make sure to push your code so I can try to reproduce (can't guarantee a timely response though)

SteveMacenski:
@stevedanomodolor so this works now?

stevedanomodolor:
> @stevedanomodolor so this works?
> 
> * With map file in the .yaml file -- uses that (if default in launch configuration is empty string)
> * Without any map defined in the .yaml file -- uses default in launch configuration
> 
> I was thinking we might need an `IfCondition` to apply the `SetParameter` only if the value of it in the launch configuration was non-empty string. I wasn't sure if an empty string launch configuration would override a real valued parameter definition in the .yaml file.

Scenario 1 -> works 
Scenario 2 -> works -> current commit
The yaml_filename when defined in the .yaml always overrides the launch file. 
Sorry for this taking longer than it should, I had some issue with the docker version

SteveMacenski:
Hey, no worries. Its all a process :smile: 

>  The yaml_filename when defined in the .yaml always overrides the launch file.

Oh really? So if we have `yaml_filename: ""` in .yaml, it'll override the `map` launch configuration that has some real-existing default value location? I would have thought that the launch configuration would override the yaml file value for any parameter when you call `SetParameter()`. 

I'm trying to game out what I would expect to happen in the major situations:

- No launch default + in yaml -> use yaml
- Launch default or set on commandline + in yaml -> use launch default or value on commandline
- No launch default + not in yaml -> completely unset
- Launch default or set on commandline + not in yaml -> use launch default or value on commandline

I think you're saying that the 2nd case is the opposite? I suppose that is fine if that's the way this acts, but that seems suspicious to me that a launch-set parameter would not take priority over a yaml set one.

stevedanomodolor:
Launch default or set on commandline + in yaml -> use launch default or value on commandline
launch default + yaml -> it pays attention to yaml. I am not if it does that because of the way configured_params loads the yaml after the setparameter has been called. This might be a better suggestion I think, I  am not sure. 

_I was thinking we might need an IfCondition to apply the SetParameter only if the value of it in the launch configuration was non-empty string. I wasn't sure if an empty string launch configuration would override a real valued parameter definition in the .yaml file._

SteveMacenski:
> Launch default or set on commandline + in yaml ...
> launch default + yaml ...

Aren't these the same?

If both in yaml and launch configuration, I would expect it to use the launch override, is there a way to accomplish that? If not, maybe we can document the fact that yaml takes over.

stevedanomodolor:
The main reason why I think the yaml file takes over is because it is just a redeclaration of the variable. Basically, calling the setparameter before the node is a requirement which initially declares the variable but the yaml file is added to the node causing the redeclaring I was referring too. Probably using an ifcondition to see the state of the variable and based on that redeclare with the yaml might be an alternative. i could try that

SteveMacenski:
It would be nice to experiment to see if that's possible! But ultimately, if its not, I think we can live with it. We just need to make sure we clearly document it in a few places to ward off confusion. Ex. in the .yaml file above the commented out `yaml_filename` comment block, in the Nav2 documentation for that parameter, and as a comment in `localization.yaml`. Basically, mention it anywhere where someone would try to investigate.

clalancette:
> @clalancette Is this expected behavior? When we use `SetParameter()` in launch and the same parameter exists in the `params.yaml` file, the value in `params.yaml` is used rather than the launch's `SetParameter()` value which was set via CLI. My intuition would say that launch-time overrides should take precedence, but that's not the behavior we experience. If that is expected behavior, please let us know (and I'll open a PR to document that somewhere).

I'm not really an expert on what launch does.  But here's what I found when playing around with the command-line.

For instance, if I have the following parameters in a file called `cam2image.yaml`:
```
/cam2image:
  ros__parameters:
    burger_mode: true
    depth: 10
    device_id: 0
    frame_id: bar
    frequency: 30.0
    height: 240
    history: keep_all
    qos_overrides:
      /parameter_events:
        publisher:
          depth: 1000
          durability: volatile
          history: keep_last
          reliability: reliable
    reliability: reliable
    show_camera: false
    use_sim_time: false
    width: 320
```

And I run:
```
ros2 run image_tools cam2image --ros-args --params-file cam2image.yaml
```

Then the `frame_id` is set to `bar`, like we expect.

If I then run:
```
ros2 run image_tools cam2image --ros-args --params-file cam2image.yaml -p frame_id:=foo
```

Then the `frame_id` is overridden to `foo`, also like we expect.  So the command-line seems to be working properly.

To move forward here, I think it would be worthwhile to see what the command-line and parameters file that launch generates look like.  That may give us a better idea of where the precedence is being set.

SteveMacenski:
Related, but not quite the same thing as what we're handling. We have a launch system with a non-trivial number of nodes, so there's no `-p <a_specific_parameter>:=val`, this is via `ros2 launch` not `ros2 run`. For launch, we set specific `LaunchConfiguration`'s which have `val` set, and when we `SetParameter()` of those launch configurations, the `yaml` is taking priority. That seems like the opposite behavior than when running a single node, which is not good that we have inconsistent behavior between 2 analog processes

---

> ros2 run image_tools cam2image --ros-args

Here nor there, but is the reverse true, if ` -p frame_id:=foo  --params-file cam2image.yaml` instead?

stevedanomodolor:
@SteveMacenski Already testing what @clalancette said. 

Assuming we used the following yaml file
```
map_server:
  ros__parameters:
    yaml_filename: "bar"
```
I replaced all the launch variable name 'map' in the launch file to 'yaml_filename' to make it easier to understand
The normal situation without commenting the map_server in the yaml file
```
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False 
ros2 param get /map_server yaml_filename
```
String value is: bar

For the second situation, I modified the yaml_filename parameters from the launch argument
```
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False yaml_filename:=test
ros2 param get /map_server yaml_filename
```
String value is: bar

The result makes sense because in the end we are just modifying the launch parameter, nothing to do with the node 

Now, equivalent to the second example you mentioned @clalancette  
`parameters=[configured_params, {'yaml_filename': map_yaml_file}],
`
```
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False yaml_filename:=test
ros2 param get /map_server yaml_filename
```
String value is: test

This still means that the SetParameter declares but the fact that configured_params is added to the parameters of the map server node, the value in the yaml file overrides that declaration, and {'yaml_filename': map_yaml_file} overrides the yaml file. 
In the last case, the yaml file is always overridden and redundant and the yaml_filename must be set in the CLI or launch file

SteveMacenski:
Does that make sense @clalancette? `SetParameter()` in Launch only works if never otherwise set, which is very not intuitive and not consistent with the way that `ros2 run` (as you showed) or manually setting in a launch file's conflicting `parameters=[]` field handles things. 

Should I file a ticket about this?

---

@stevedanomodolor So for a path forward here, it sounds like we should: 
- Add ` parameters=[configured_params, {'yaml_filename': map_yaml_file}],` if `map_yaml_file` is non-empty string, rather than use `SetParameter`
-  Its not super clean, but that should be possible if we have 2 instances of the `Node()` for map server, one with a `condition=IfCondition(PythonExpression(['not ', map_yaml_file])),` to not set the map, and one with `condition=IfCondition(PythonExpression([map_yaml_file, ' == ""'])),` to set the map (representing that the map is non-empty)

That way, we get:
- If not set on launch commandline or by a launch configuration default, we _don't_ override a yaml default
- If we set it on launch commandline / launch configuration, we _do_ override a yaml default

I think that gets us what we want, yes? I don't see a cleaner way to have a condition for including a parameter field to the `parameter=[]` list. I'd just add in a 1-line comment about why this was necessary for future users.

stevedanomodolor:
@SteveMacenski  It seems that comparison with variables that are not either boolean or numbers does not work

`condition=IfCondition(PythonExpression([map_yaml_file, ' != ""']))
condition=IfCondition(PythonExpression(['not ', map_yaml_file]))
`

It constantly gives the error
` [ERROR] [launch]: Caught exception in launch (see debug for traceback): invalid syntax (<string>, line 1)
`

SteveMacenski:
Huh, in googling looks like there's new `LaunchConfigurationEquals` and `LaunchConfigurationNotEquals` that should do the trick!

https://answers.ros.org/question/360264/ifconditionpythonexpression-am-i-doing-it-right/

nav2_bringup/launch/localization_launch.py
```diff
@@ -146,11 +162,21 @@ def generate_launch_description():
                 target_container=container_name,
                 composable_node_descriptions=[
                     ComposableNode(
+                        condition=LaunchConfigurationEquals('map', ''),
```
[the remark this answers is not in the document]
SteveMacenski: Not sure, perhaps at the `LoadComposableNodes` level instead of the `ComposableNode` level? So there would be 3 separate loads: 1 with map reconfigured, 1 without, and 1 for everything else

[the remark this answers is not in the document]
SteveMacenski: It looks like the composable node function is a ["description"](https://github.com/ros2/launch_ros/blob/7b0aa892662cf1e2d2a6a5ffc4fa67ed9f834410/launch_ros/launch_ros/descriptions/composable_node.py#L36) whereas the node function is an ["action"](https://github.com/ros2/launch_ros/blob/7b0aa892662cf1e2d2a6a5ffc4fa67ed9f834410/launch_ros/launch_ros/actions/node.py#L113). Load Composable nodes is an ["action"](https://github.com/ros2/launch_ros/blob/7b0aa892662cf1e2d2a6a5ffc4fa67ed9f834410/launch_ros/launch_ros/actions/load_composable_nodes.py#L55) so maybe that has the `condition` field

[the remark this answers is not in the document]
stevedanomodolor: The [descriptor](https://github.com/ros2/launch_ros/blob/7b0aa892662cf1e2d2a6a5ffc4fa67ed9f834410/launch_ros/launch_ros/descriptions/composable_node.py#L60,) also has a condition param, I am guessing it should to be possible to use our condition but in its specific format of type [SomeSubstitutionsType ](https://github.com/ros2/launch/blob/rolling/launch/launch/some_substitutions_type.py). I will look into it, if not try the load composable nodes solution.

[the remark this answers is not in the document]
stevedanomodolor: Related issue, bug in composablenode launch: https://github.com/ros2/launch_ros/issues/114

stevedanomodolor:
Related issue https://github.com/ros2/launch_ros/issues/114

nav2_bringup/launch/localization_launch.py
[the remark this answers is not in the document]
stevedanomodolor: > Not sure, perhaps at the `LoadComposableNodes` level instead of the `ComposableNode` level? So there would be 3 separate loads: 1 with map reconfigured, 1 without, and 1 for everything else

The current solution uses this proposal and it works. The issue with the condition param of the composablenode not working should have been fixed in https://github.com/ros2/launch_ros/issues/114 but for some reason it does not work. I am not sure if it has something to do with the ros2 version.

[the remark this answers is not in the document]
SteveMacenski: Or it was broken again by something, mhm.

nav2_bringup/params/nav2_params.yaml
```diff
@@ -243,10 +243,11 @@ global_costmap:
         inflation_radius: 0.55
       always_send_full_costmap: True
 
+# The yaml_filename does not need to be specified since it going to be set by defaults in launch.
+# If you'd rather set it in the yaml, remove the default "map" value in the tb3_simulation_launch.py
+#file & provide full path to map below.
```
SteveMacenski, line 248:
```suggestion
# file & provide full path to map below. If CLI map configuration or launch default is provided, that will be used.
```

SteveMacenski:
OK - overall this looks good now - I think there's just the action items in the open comments on the reviews. Mostly, just cleaning up what you have and adding some comments in the localization launch file about "why" we did this this way for future generations (e.g. LoadComposableNode for map server twice depending if we should use the value of map from a CLI or launch default or user defined value in the yaml configuration file. They are separated since the conditions currently only work on Load commands and not on the composition node function itself).

nav2_bringup/launch/localization_launch.py
```diff
@@ -102,12 +104,14 @@ def generate_launch_description():
     declare_log_level_cmd = DeclareLaunchArgument(
         'log_level', default_value='info',
         description='log level')
+    declare_flag_cmd = DeclareLaunchArgument(name='flag', default_value=['False'])
```
SteveMacenski, line 107: What's this?

stevedanomodolor: Mistake from my part

nav2_bringup/params/nav2_params.yaml
```diff
@@ -243,11 +243,12 @@ global_costmap:
...
       always_send_full_costmap: True
 
-map_server:
-  ros__parameters:
-    # Overridden in launch by the "map" launch configuration or provided default value.
-    # To use in yaml, remove the default "map" value in the tb3_simulation_launch.py file & provide full path to map below.
-    yaml_filename: ""
+# The yaml_filename does not need to be specified since it going to be set by defaults in launch.
+# If you'd rather set it in the yaml, remove the default "map" value in the tb3_simulation_launch.py
+#file & provide full path to map below.If CLI map configuration or launch default is provided, that will be used.
```
SteveMacenski, line 248:
```suggestion
# file & provide full path to map below. If CLI map configuration or launch default is provided, that will be used.
```

SteveMacenski (changes requested):
Small nit and ready! Please also open a PR against our documentation to add to the migration guide this change!

stevedanomodolor:
> Reference in

In which of the migration rule?, in the four of them?

SteveMacenski:
Hi,

Sorry for the delay (roscon/iros), I meant adding to the migration guide this behavior has changed https://github.com/ros-planning/navigation.ros.org/blob/master/migration/Humble.rst. Your updates to the parameter guide though are very helpful and merged those!

SteveMacenski (approved):

SteveMacenski:
https://github.com/ros2/launch_ros/issues/328 The folks at OR responded  to your launch issue - can you try out @methylDragon 's suggestion?

stevedanomodolor:
> [ros2/launch_ros#328](https://github.com/ros2/launch_ros/issues/328) The folks at OR responded to your launch issue - can you try out @methylDragon 's suggestion?

I have added the changes to the migration guide. Wrt to [ros2/launch_ros#328](https://github.com/ros2/launch_ros/issues/328), it seems that the feature was added not long ago, when i try to update the launch package, I do not get that update, currently using the docker form of nav2, I an not sure if that is the reason why it does not get those changes. If i get that resolved, I will try it, if you know a way to solve this, let me know, always happy to hear your feedback.

SteveMacenski:
OK, might be something good to revisit then once things sync up to Rolling. 

Can you make a note about that in that comment block so next time I stumble into it in the future, I can test it out?

Reviewed the migration guide PR, just small stuff and then we can merge the pair of PRs.

Thanks for your patience and help with this!

stevedanomodolor:
When you mean comment block, are you referring to the migration guide ?, btw I already made a code to test this new feature: https://github.com/stevedanomodolor/navigation2/tree/feat/EqualsSubstitution, if it might be useful if you want to test it too.

SteveMacenski:
Thanks! Its not yet released yet from MenthylDragon's response. So, that's a future thing in a few weeks, but I don't want to block this PR for that. I can apply your patch later on. 

I mean here: https://github.com/ros-planning/navigation2/pull/3174/files#diff-a76c17e1db2110fbb385675216aa5a94bba6f06a2c5744b38621ae6b5e526c27R158 add to the comment block in the launch file also a note about switching to use the new method because the current one is scheduled for deprecation once a Rolling sync is conducted.

stevedanomodolor:
> Thanks! Its not yet released yet from MenthylDragon's response. So, that's a future thing in a few weeks, but I don't want to block this PR for that. I can apply your patch later on.
> 
> I mean here: https://github.com/ros-planning/navigation2/pull/3174/files#diff-a76c17e1db2110fbb385675216aa5a94bba6f06a2c5744b38621ae6b5e526c27R158 add to the comment block in the launch file also a note about switching to use the new method because the current one is scheduled for deprecation once a Rolling sync is conducted.

Changes added

SteveMacenski (approved):

SteveMacenski:
Just waiting on CI

Issue #3171, opened 2022-09-05
Collision monitor not detecting future collision with pure rotation in approach model


automech-rb:
## Bug report

**Required Info:**

- Operating System:
  - Docker 22.04 Ubuntu
- ROS2 Version:
  - rolling
- Version or commit hash:
  - 7813feefd4ad4d80064af82b4ac659f52a68283e
- DDS implementation:
  - rmw_cyclonedds_cpp

#### Steps to reproduce issue

I am using a rectangular robot with two laser scanner. Config as follows:
```yaml
collision_monitor:
  ros__parameters:
    base_frame_id: "base_footprint"
    odom_frame_id: "odom"
    cmd_vel_in_topic: "cmd_vel_raw"
    cmd_vel_out_topic: "cmd_vel"
    transform_tolerance: 0.5
    source_timeout: 5.0
    stop_pub_timeout: 2.0
    polygons: ["FootprintApproach"]
    FootprintApproach:
      type: "polygon"
      action_type: "approach"
      footprint_topic: "/local_costmap/published_footprint"
      time_before_collision: 2.0
      simulation_time_step: 0.02
      max_points: 1
      visualize: True
    observation_sources: ["scan_front", "scan_back"]
    scan_front:
      type: "scan"
      topic: "/scan/front"
    scan_back:
      type: "scan"
      topic: "/scan/back"
```
I am publishing 0.3 rad/sec twist to cmd_vel_raw topic in both gif below and footprint is from `FootprintApproach`

#### Expected behavior
As `time_before_collision` is 2.0 sec, it should detect collision for obstacles under 34 degree in rotation wrt current pose. It should slow down to avoid collsion.
#### Actual behavior
Collision is not detected (in the log), and robot continued to rotate and collide with obstacle (In gif, it is a medium sized block)
In first gif below, first I try to collide head on, but the collision is detected and robot slows down. As next I try to rotate in place, but robot collides with block (log don't show any collision warning).
![2_finas](https://user-images.githubusercontent.com/24780015/188447440-011f6716-7ffa-4fd8-9326-aa696dafccfa.gif)

In this gif, same behavior but with different angle of approach.
![1_fina](https://user-images.githubusercontent.com/24780015/188447702-2576e64c-e81f-41c5-9259-1cc910be22ed.gif)

#### Additional information
[...]

PR #3177, opened 2022-09-07
Fix velocities comparison for rotation at place case

AlexeyMerzlyakov:
Fixes incorrect velocities comparison appeared when robot is rotating on the place. This was leading to failure of `safe_vel < robot_action.req_vel` checks for APPROACH and SLOWDOWN behavioral models, that causes robot to ignoring its intended behavior.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3171 |
| Robotic platform tested on | TB3 simulation, colcon testing & coverage verification |

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -15,6 +15,8 @@
 #ifndef NAV2_COLLISION_MONITOR__TYPES_HPP_
 #define NAV2_COLLISION_MONITOR__TYPES_HPP_
 
+#include <cmath>
+
 namespace nav2_collision_monitor
 {
 
@@ -29,7 +31,15 @@ struct Velocity
   {
     const double first_vel = x * x + y * y;
     const double second_vel = second.x * second.x + second.y * second.y;
-    return first_vel < second_vel;
+    if (second_vel > 0.0) {
+      // Robot moves: comparing velocities. In Collision Monitor
+      // twists will change proportionally to linear velocities.
+      // We do not need to compare them in this case.
+      return first_vel < second_vel;
+    } else {
+      // Robot is already stays in place. We need to compare twists.
+      return std::fabs(tw) < std::fabs(second.tw);
+    }
   }
 
   inline Velocity operator*(const double & mul) const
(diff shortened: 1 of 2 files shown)


nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
```diff
@@ -29,7 +31,15 @@ struct Velocity
   {
     const double first_vel = x * x + y * y;
     const double second_vel = second.x * second.x + second.y * second.y;
-    return first_vel < second_vel;
+    if (second_vel > 0.0) {
```
SteveMacenski, line 34: I'd make this some small epsilon like 1e-3 for small odometric error for not quite pure rotations but largely pure rotations

SteveMacenski:
The other option would be to compare the total components, e.g. not just translation and rotation separately but together

```
x * x + y * y + theta * theta < second.x * second.x + second.y * second.y + second.theta * second.theta
```

AlexeyMerzlyakov:
I thought about it: the latest comparison will work fine for Collision Monitor (only), since it is being used in two places: in "approach" and "slowdown" models comparisons during decision making of selected behavior. Here the velocity is being multiplied on some slowdown coefficient, all of components including twists that will be changed proportionally. So, adding this term in the velocities comparison won't brake anything, and moreover, this is better solution. Thank you for the suggestion! I'll update the patch and re-run the testing to verify updated solution works fine.

nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
AlexeyMerzlyakov: No need, since latest [comment](https://github.com/ros-planning/navigation2/pull/3177#issuecomment-1239644594) was met.

nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
```diff
@@ -27,8 +27,10 @@ struct Velocity
 
   inline bool operator<(const Velocity & second) const
   {
-    const double first_vel = x * x + y * y;
-    const double second_vel = second.x * second.x + second.y * second.y;
+    const double first_vel = x * x + y * y + tw * tw;
+    const double second_vel = second.x * second.x + second.y * second.y + second.tw * second.tw;
+    // In Collision Monitor twists will change proportionally to linear velocities,
```
SteveMacenski, line 32: I'm not sure you need the comment to justify it - if you add a comment at all it should just be explaining what its doing / the intention (and the fact that this resolves rotation in place issues)

AlexeyMerzlyakov: Okay, I've [updated](https://github.com/ros-planning/navigation2/pull/3177/commits/32bcf9e603293b973c44664266bfb4b35c339a04) the comment intention

SteveMacenski (approved):

PR #3188, opened 2022-09-14
Upgrade resource class for CI workers

ruffsl:
in an attempt to fix CI Build timed outs.
- https://github.com/ros-planning/navigation2/issues/3189


SteveMacenski:
Good to go or?

PR #3190, opened 2022-09-14
Control exceptions

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Add error codes for local planning failures. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_dwb_controller/dwb_core/include/dwb_core/illegal_trajectory_tracker.hpp
+++ b/nav2_dwb_controller/dwb_core/include/dwb_core/illegal_trajectory_tracker.hpp
@@ -39,7 +39,7 @@
 #include <utility>
 #include <string>
 #include "dwb_core/exceptions.hpp"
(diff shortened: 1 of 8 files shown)


SteveMacenski:
FYI - I'm mostly ignoring this one since I think there will be meta-conversations in the planning PR that will apply here too. So waiting for that to be largely done before moving on here so there isn't a huge amount of redundant related topics.

nav2_dwb_controller/dwb_core/include/dwb_core/illegal_trajectory_tracker.hpp
```diff
@@ -66,11 +66,11 @@ class IllegalTrajectoryTracker
  * @brief Thrown when all the trajectories explored are illegal
  */
 class NoLegalTrajectoriesException
```
SteveMacenski, line 68: Should these be entirely removed and replaced with nav2_msgs versions?

Issue #3191, opened 2022-09-15
Galactic branch's Spin.Action missing duration definition


madgrizzle:
## Bug report
- Operating System:
Ubuntu 20.04
- ROS2 Version:
Galactic
- Version or commit hash:
ros-galactic-navigation2  1.08-1focal.20220203.051427
- DDS implementation:
FastRTPS

#### Steps to reproduce issue
When using nav2_simple_commander spin command, a crash occurs and error in line 187 of robot_navigator.py is reported that the "Spin Goal" object has no attribute 'time_allowance'.   

#### Additional information
robot_navigator.py imports Spin from nav2_msgs.action and attempts to set the time_allowance, but Spin.action's goal definition in the galactic branch only defines float32 target_yaw and does not include builtin_interfaces/Duration time_allowance found in the definition of the main branch. 

https://github.com/ros-planning/navigation2/blob/3df00a74b79b4581599a65d3dbf60a9568cc5b5b/nav2_simple_commander/nav2_simple_commander/robot_navigator.py#L186

**galactic branch of Spin.action:**
https://github.com/ros-planning/navigation2/blob/3df00a74b79b4581599a65d3dbf60a9568cc5b5b/nav2_msgs/action/Spin.action#L1-L7

**main branch of Spin.action:**
https://github.com/ros-planning/navigation2/blob/34ecee20a1913a71bc345b0438a0081b73159458/nav2_msgs/action/Spin.action#L1-L9

SteveMacenski:
We cannot change the message definition after the release of a distribution, since that is not ABI compatible for existing users.

It is an oversight that that was backported, so instead of changing the message that portion of the simple API should just be removed

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/3193 Can you verify this resolves your issue?

madgrizzle:
understood.  thanks.  I've been working from the release package and not building from source, but will give it a go and let you know.

SteveMacenski:
Understandable - sorry about that. We don't have unit tests set up in Python3 (yet) so that isn't covered. Interpreted languages :eyeroll: amiright?!

madgrizzle:
hmmm.. is this related?  I got this error message right after the robot physically spun. 
![image](https://user-images.githubusercontent.com/5055291/190298462-6a5b5bc1-6aa1-438f-91a6-c0b5212317b2.png)

madgrizzle:
[...]

PR #3192, opened 2022-09-15
Update Spin.Action in Galactic branch

madgrizzle:
Update to include time_allowance in the goal definition.  robot_navigator.py attempts to set this in the spin command and crashes because its not defined.  Reference https://github.com/ros-planning/navigation2/issues/3191#issue-1373760857

#### For Maintainers: 


SteveMacenski:
Will follow up in the ticket, but this is a wontfix in changing the message definition

Issue #3151, opened 2022-08-26
Simple Commander Costmap API


SteveMacenski:
Add a costmap API for `OccupancyGrid` messages coming into the Simple Commander API

Basically have the basic API from https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp + adding global frame ID and timestamp with a constructor that takes in the `OccupancyGrid`.

Potentially even a python collision checker too when given a footprint from the footprint topic + this new `CostmapPy` class

padhupradheep:
For someone who is going to implement it, we had something something similar, which can be found under https://github.com/neobotix/neo_nav2_py_costmap2D

(wish it was me, but unfortunately I don't have time until october to get into this)

stevedanomodolor:
I would love to try it, is t possible?

SteveMacenski:
Absolutely!

afifswaidan:
I would like to contribute as well, let me know what I can do!

igrak34:
Same here, let me know if you need some help with it :)

manishsaini0901:
I would love to contribute. Let me know if we can do it as a team. @SteveMacenski @afifswaidan @igrak34 @stevedanomodolor

afifswaidan:
> I would love to contribute. Let me know if we can do it as a team. @SteveMacenski @afifswaidan @igrak34 @stevedanomodolor 

I have no problem from my side, count me in.

stevedanomodolor:
> I would love to contribute. Let me know if we can do it as a team. @SteveMacenski @afifswaidan @igrak34 @stevedanomodolor

Sure me too, I don't mind

JacksonK9:
I would like to contribute. But, I am not so familiar with ROS2, but familiar with ROS1.
Can I try to contribute with as a team? @SteveMacenski @afifswaidan @igrak34 @stevedanomodolor

SteveMacenski:
Awesome! 

So the general things that need to be implemented are those in the files I linked to above, namely a class `PyCostmap2D` 

- A: Constructor for OccupancyGrid which stores the relevant information 
- A: getSizeInCellsX / getSizeInCellsY
- A: getSizeInMetersX / getSizeInMetersY
- A: getOriginX / getOriginY
- A: getResolution
- A: getGlobalFrameID
- A: getCostmapTimestamp
- B: getCost at map pixels and index
- B: setCost at map pixels
[...]

PR #3197, opened 2022-09-17
Added Line Iterator

afifswaidan:
---

## Basic Info

| Info | Added Line Iterator |
| ------ | ----------- |
| Ticket(s) this addresses   | Issue number #3151|
| Primary OS tested on | Ubuntu 20.04.5 LTS |
| Robotic platform tested on | ROS2 Foxy|

---

- Added Line Iterator Python API

---

- I have few questions before I start implementing the python version of the footprint_collision_checker:
- The Cpp script uses footprint.hpp and several other modules like "nav2_costmap_2d/exceptions.hpp", should I implement all of these modules first in python (So I use them in the python version of footprint_collision_checker) or are they already implemented somewhere else?

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
How did you test this? A set of unit tests here would be good.

afifswaidan:
Hello,

As the main purpose of this class is to do some basic mathematical computations, I tested it manually by giving some valid and invalid inputs and I print the outputs for validation.

If required, let me know what exact tests you have in mind so I can implement and commit them.

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,119 @@
...
+            self.numadd_ = self.delta_x_
+            self.numpixels_ = self.delta_y_
+
+    def isValid(self):
+        """Checks if the given line is valid"""
+        return self.curpixel_ <= self.numpixels_
+
+    def advance(self):
+        """Advances to the next point in the line."""
+        self.num_ = self.numadd_
```
SteveMacenski, line 88: Should this be += ?

SteveMacenski:
add some unit tests in `/tests` that can be run in our CI suite that check that this works properly. Give it a few different line segments at different angles and check that it visits every cell increment correctly

afifswaidan:
Okay sure, I'll implement and push the changes today or tomorrow.
Thanks!

afifswaidan:
Hello @SteveMacenski 
I implemented a new way for the line iterator that contains more features and more precision.

Can you please have a look at it before I start implementing the unit tests for it?
Maybe you want me to add/remove anything.
Thanks!

SteveMacenski:
Can you explain the method / benefits over the old? 

CI shows this has linting errors

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,141 @@
...
+        self.x_ = x0
+        self.y_ = y0
+
+        if x1 != x0 and y1 != y0:
+            self.valid_ = True
+            self.distance_ = abs(x1-x0)
+            self.m_ = (y1-y0)/(x1-x0)
+            self.b_ = y1 - (self.m_*x1)
+            if (self.b_ < 0):
+                self.equation_ = "y = " + str(self.m_) + "*x " + str(self.b_)
```
SteveMacenski, line 55: Missing a plus or minus or something for interaction between x and b?

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,141 @@
...
+            self.distance_ = abs(y1-y0)
+            self.equation_ = "x = " + str(x1)
+        elif y1 == y1 and x1 != x0:
+            self.valid_ = True
+            self.distance_ = abs(x1-x0)
+            self.equation_ = "y = " + str(y1)
+        else:
+            self.valid_ = False
+            self.equation_ = "Invalid"
+        self.step_ = self.distance_/1000
```
SteveMacenski, line 69: Why 1000?

afifswaidan: Yes, I agree, my plan was to add another argument in the class constructor where the users input the resolution (number of steps) that they want to go through his line, this way the user can decide the resolution of the iterations.
Please let me know what you think about that.

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: In this case no since b is already negative so it will be displayed with a minus.
In the other case where b is positive I added the + sign to the string.

afifswaidan:
> 

As for the linting errors, I am developing on a new machine and apparently, no linter was installed, I'll move back to my old machine tomorrow.

As for the advantages:

- The user can input the resolution of the calculations  (mainly the iterations)
- The user can get the equation of the line
- The user can get the characteristics of the line (slope m and y-intercept b) to solve for any x or y points. (I will add two methods called solve_for_x and solve_for_y where the user inputs x or y and the function solves it using the equation of the line.
- This method takes into consideration the cases where the equation includes only x (like, x=5)

afifswaidan:
> I think the density of resolution is a good point - but that would need to be parameterized and by default just every 1-cell jumps. I think though someone could want to try 2x or 3x cell lengths to get cells that are barely touching but might technically be in collision. This comes as a high cost for longer range queries, but for footprints I think that's a useful feature. 
> 
> If you run `ament_pep257` and `ament_flake8` you should be able to see the outputs locally to iterate


Okay I will make the resolution as a parameter that the user can define, and by default it'll be 1.

As for the linting I will do it as well, and then I will commit the changes for review, once everything is good I will implement a few test cases in the /test directory and commit it for the final review and merge.

Thank you so much @SteveMacenski

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,201 @@
...
+            self.equation_ = "Invalid"
+
+    def isValid(self):
+        """Returns True if the line is valid
+        Returns:
+            Bool: Flag if live is valid (true) or not (false)
+        """
+        return self.valid_
+
+    def advance(self):
```
SteveMacenski, line 73: What about the very last point? If we have a terminus of `(x,y)`, we march towards it in `step_size` increments, but it might not _exactly_ hit `(x,y)`. Right now, we'd blow past it, then have `valid` be `false`, but never actual check the terminus point

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,201 @@
...
+            if self.x_ > self.x1_:
+                self.x_ = round(self.clamp(
+                    self.x_ - self.step_size_, self.x1_, self.x0_), 5)
+                self.y_ = round(self.m_ * self.x_ + self.b_, 5)
+            else:
+                self.valid_ = False
+        else:
+            if self.y1_ > self.y0_:
+                if self.y_ < self.y1_:
+                    self.y_ = round(self.clamp(
```
SteveMacenski, line 92: And X?

SteveMacenski (changes requested):

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: This if statement handles the case where the line is vertical (X1 == X0), so the equation will be for example x = 3 and only Y is changing. Thus I'm only incrementing Y here as X is fixed in the vertical line

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: In the case of the very last point, if the step_size will increment in a way past the final coordinates (x,y) the "clamp" function that I'm using will make sure not to increment past the terminus of (x,y). This way we will always march on the line using step_size increments until the last increment that will be constrained not to cross the terminus of (x,y).

I tested it in a test script and it never crosses past the final point in the line while it increments according to step_size along the line.

nav2_simple_commander/nav2_simple_commander/line_iterator.py
SteveMacenski: Then why compute / store `m` and `b`?

afifswaidan: In the case of X1 == X0 (Line 54), m and b are not computed nor stored, and they are not used in the advance function in this case (Lines 90 --> 93)

afifswaidan: m and b are only computed/stored only in the cases where the line equation contains m and b (all cases other than the vertical line case)

SteveMacenski (approved):

SteveMacenski:
Make sure the tests execute during CI!

afifswaidan:
> Make sure the tests execute during CI! You may need to add something to the setup.py/cfg files

Thanks for the info, I will try to do that asap this week. I'm sorry if the process is going slow but it's a busy period with my current full-time job.

I'll handle the CI tests and wait for your feedback.
Thanks!

SteveMacenski:
Totally understand! Thanks for talking me through the changes on my questions.

afifswaidan:
Hello @SteveMacenski , I added all the unit tests and modified line_iterator.py to handle more exceptions that can be caused by the user.
But I am facing an issue while importing the line_iterator.py module to the test script in /test directory. I am not facing this issue when the files are in the same directory.
Can you please check it and give me your opinion?

Thanks!

SteveMacenski:
https://github.com/ros2/geometry2/tree/rolling/tf2_ros_py Try using this Python3 package as a guide.

afifswaidan:
> https://github.com/ros2/geometry2/tree/rolling/tf2_ros_py Try using this Python3 package as a guide.

Hello,
Thanks for the reference. But I noticed in the tests script they are using the "unittest" package.
While in the setup.py it contains "tests_require=['pytest']"
This is why I got confused and I implemented the tests using pytest instead. (Also the tf2_ros_py repo contains pytest.ini file but they didn't use the pytest package in their test scripts)

I will implement the using "unittest" as you suggested based on tf2_ros_repo.
But I just wanted to ask you about this because it got me a bit confused.
Thanks @SteveMacenski

afifswaidan:
@SteveMacenski Linting and Unit Testing are done, let me know if you need anything else for this commit.
Thanks! 🙂

afifswaidan:
@SteveMacenski anything else required from my side?

SteveMacenski:
Sorry, was gone at ROSCon/IROS, just getting back to this now

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def getY1(self):
+        """Get the ordinate of the final point."""
+        return self.y1_
+
+    def get_line_length(self):
+        """Get the length of the line."""
+        return sqrt(pow(self.x1_ - self.x0_, 2) + pow(self.y1_ - self.y0_, 2))
+
+    def get_line_equation(self):
```
SteveMacenski, line 164: Can this be removed? I'm just not sure when this would be useful in practice. Also removes some of the stuff to populate `self.equation_` (+ is missing unit tests)

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def getX1(self):
+        """Get the abscissa of the final point."""
+        return self.x1_
+
+    def getY1(self):
+        """Get the ordinate of the final point."""
+        return self.y1_
+
+    def get_line_length(self):
```
SteveMacenski, line 160: This method doesn't appear to be used

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,196 @@
...
+        """Get the length of the line."""
+        return sqrt(pow(self.x1_ - self.x0_, 2) + pow(self.y1_ - self.y0_, 2))
+
+    def get_line_equation(self):
+        """Get the equation of the line as a string."""
+        return self.equation_
+
+    def get_curr_point_str(self):
+        """Get the coordinates of the current point as string."""
+        return "X: " + str(self.x_) + "   Y: " + str(self.y_)
```
SteveMacenski, line 170: This method doesn't appear to be used

nav2_simple_commander/nav2_simple_commander/line_iterator.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def get_curr_point(self):
+        """Get the coordinates of the current point as list [X,Y]."""
+        return [self.x_, self.y_]
```
SteveMacenski, line 174: This method doesn't appear to be used

SteveMacenski (changes requested):
Otherise LGTM, just remove some methods that don't really appear like that need to be in the file. Dead code is one common way how bugs are made (if its not tested and not used, then if you want to use in the future, you're going to assume it works). Plus some of it seems like printing convenience functions that can be made if/when it becomes an issue. Nice for early development, but I don't think need to be here anymore

nav2_simple_commander/nav2_simple_commander/line_iterator.py
SteveMacenski: I'd argue its useful, but needs to be used somewhere, or at the very least have its own unit test

afifswaidan:
> Otherise LGTM, just remove some methods that don't really appear like that need to be in the file. Dead code is one common way how bugs are made (if its not tested and not used, then if you want to use in the future, you're going to assume it works). Plus some of it seems like printing convenience functions that can be made if/when it becomes an issue. Nice for early development, but I don't think need to be here anymore

I hope you enjoyed the conference and welcome back!
All these unused methods that you mentioned are basically extra methods that the user can later on use if they need more information about the line that they fed to the class (the line equation, the current point they are at after calling advance(), etc...)
I can either add unit tests for these methods or remove them, whatever you think is better. I just added them to provide extra features for the user later on to access more information about the given line.
Let me know what you think :)

SteveMacenski:
I updated the comments for each about what to remove / keep and add tests. The things that return debugging strings lets remove. The line length can be a helpful gut check, lets add unit tests for that. `get_curr_point` is already possible via `getX` and `getY`.

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: Method removed.

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: Unit test added

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: Method removed.

nav2_simple_commander/nav2_simple_commander/line_iterator.py
afifswaidan: Method removed.

afifswaidan:
@SteveMacenski I was checking why the circleci release_test failed, I found it was a failure in another package (nav2_controller).
Is anything required from my side?
Thanks!
image

SteveMacenski:
Probably not, I just retriggered CI to see if it was just a fluke

SteveMacenski (approved):

SteveMacenski:
Please push a bogus commit (remove a white space in a doc or change a word), CI's built for detecting C++ changes not Python changes so CI isn't happy with caching with getting it to re-run

SteveMacenski:
FYI @ruffsl the colcon `cache` keyword did not detect the changes to the python files for rebuilding

SteveMacenski:
Huh, just an odd fluke. If it happens again I'll look into it.

Thanks for this contribution!

afifswaidan:
> Huh, just an odd fluke. If it happens again I'll look into it.
> 
> Thanks for this contribution!

Most welcome, anything for ROS and Robots you know! 😎👌

Glad everything is merged now, thank you for your proper guidance and assistance as well.

PR #3199, opened 2022-09-18
Add nav2_fake_localization

jihoonl:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu 22.04
| Robotic platform tested on | motor-only simulated robot based on [kobuki_soft](https://github.com/yujinrobot/kobuki_soft)  |

---

* I have ported a subset of [fake_localization](https://github.com/ros-planning/navigation/tree/noetic-devel/fake_localization) to nav2 format with a few changes. 
* Instead of `base_pose_ground_truth`, it listens to `odom`.
* It doesn't provide `amcl_pose` and `particlecloud`

---

#### For Maintainers: 

### DIFF
--- a/nav2_fake_localization/CMakeLists.txt
+++ b/nav2_fake_localization/CMakeLists.txt
@@ -25,7 +25,7 @@ set(DEPENDENCIES
   geometry_msgs
   nav_msgs
   message_filters
-  )
+)
 
 include_directories(
   include
--- a/nav2_fake_localization/CMakeLists.txt
+++ b/nav2_fake_localization/CMakeLists.txt
@@ -1,0 +1,60 @@
+cmake_minimum_required(VERSION 3.8)
+project(nav2_fake_localization)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# find dependencies
+find_package(ament_cmake_ros REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(angles REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(nav_msgs REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(message_filters REQUIRED)
+
+set(DEPENDENCIES
+  angles
+  tf2
+  tf2_ros
+  tf2_geometry_msgs
+  rclcpp
+  geometry_msgs
+  nav_msgs
+  message_filters
+  )
+
+include_directories(
+  include
+)
+
+add_executable(fake_localization src/fake_localization.cpp)
+ament_target_dependencies(fake_localization ${DEPENDENCIES})
+
+install(
+  DIRECTORY include/
+  DESTINATION include
+)
+install(
+  TARGETS fake_localization
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION lib/${PROJECT_NAME}
+)
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  # the following line skips the linter which checks for copyrights
+  # comment the line when a copyright and license is added to all source files
+  set(ament_cmake_copyright_FOUND TRUE)
+  # the following line skips cpplint (only works in a git repo)
+  # comment the line when this package is in a git repo and when
+  # a copyright and license is added to all source files
+  set(ament_cmake_cpplint_FOUND TRUE)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
+ament_package()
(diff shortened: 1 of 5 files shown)


jihoonl:
It seems to fail due to the package I never touched. Would I need to address the issue?

SteveMacenski:
Yeah that's fine - understood, that won't block this PR.

A question:  What's the motivation to add this package? This was more helpful in the earlier days of Navigation development when we lacked a great deal of compute power, but these days a full robot stack with simulation only takes about 50% of a consumer laptop. I think its best to have simulation test the full systems together to model the error that is actually present in localization algorithms being used. But maybe there's a use-case here I'm not thinking of. 

Besides that, for packages to come into Nav2, we require test coverage > 90% with documentation on navigation.ros.org (configuration guide page for parameter descriptions, perhaps a tutorial for something that isn't default-used so that users know how to use it - but can be relatively simple).

nav2_fake_localization/include/nav2_fake_localization/fake_localization.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <geometry_msgs/msg/transform_stamped.hpp>
+#include <memory>
+#include <nav_msgs/msg/odometry.hpp>
+#include <rclcpp/rclcpp.hpp>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
+
+namespace nav2_fake_localization
+{
+
+class FakeOdomNode : public rclcpp::Node
```
SteveMacenski, line 70: Class and member functions need doxygen added

SteveMacenski: Why not a lifecycle node?

nav2_fake_localization/src/fake_localization.cpp
```diff
@@ -0,0 +1,176 @@
...
+  auto node = std::make_shared<nav2_fake_localization::FakeOdomNode>();
+  rclcpp::Rate loop(20);
+
+  while (rclcpp::ok()) {
+    rclcpp::spin_some(node);
+    loop.sleep();
+  }
+  rclcpp::shutdown();
+  return 0;
+}
```
SteveMacenski, line 176: Remove main, make this a component. You can have a `main` in another file, but that way it can be used as a component or as a standalone node

nav2_fake_localization/src/fake_localization.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+} // namespace nav2_fake_localization
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  auto node = std::make_shared<nav2_fake_localization::FakeOdomNode>();
+  rclcpp::Rate loop(20);
+
+  while (rclcpp::ok()) {
```
SteveMacenski, line 170: Why not just `spin()`?

nav2_fake_localization/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+cmake_minimum_required(VERSION 3.8)
+project(nav2_fake_localization)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# find dependencies
```
SteveMacenski, line 8: Check out other Nav2 package CMakeLists, your missing some exports + `nav2_package()` macro calls.

SteveMacenski (changes requested):

SteveMacenski:
Any followup?

jihoonl:
> A question: What's the motivation to add this package? This was more helpful in the earlier days of Navigation development when we lacked a great deal of compute power, but these days a full robot stack with simulation only takes about 50% of a consumer laptop. I think its best to have simulation test the full systems together to model the error that is actually present in localization algorithms being used. But maybe there's a use-case here I'm not thinking of.

The motivation is to test navigation planner and controller parameters in various environments with minimal effort. Although the great improvements in simulation, it still requires high computation and effort to design a testing environment. With this node, I can purely test planner and controller parameters in various environments and situations with only map images.

I agree that localization matters for navigation but also I believe that many things can be tested without localization concerns. Such as application-level systems that are above the single robot navigation stack. 

In addition, I also believe configuring multi-robot tests with simulation would still require quite strong computation and enough experience. :)

> Besides that, for packages to come into Nav2, we require test coverage > 90% with documentation on navigation.ros.org (configuration guide page for parameter descriptions, perhaps a tutorial for something that isn't default-used so that users know how to use it - but can be relatively simple). Unlike in ROS 1 Navigation, I've set some pretty high quality standards to make sure maintaining this long term is possible and provides enough tools for users to know how to use the things that exist here so there aren't any "hidden" features. Something like this though could be tested via a nav2_system_test and probably hit that 90% number in some trivial A->B navigation task using this for localization and just checking that its working within reasonable bounds.

I agree with having high-quality standards for long term maintenance. I will try to meet the standards as well as address comments soon.

jihoonl:
I will try my best to follow up before roscon.

SteveMacenski:
> With this node, I can purely test planner and controller parameters in various environments and situations with only map images.

Why do you need localization for either of those? Planner doesn't rely on localization - this is an example benchmarking framework showing that https://github.com/ros-planning/navigation2/tree/main/tools/planner_benchmarking. The controller uses odometry and doesn't actually rely on a map->odom transform.

jihoonl:
I couldn't find time to follow up on the PR. I will reopen it when I get sufficient time to work on it.

Issue #3167, opened 2022-09-02
time wasted in _graph.emplace


mangoschorle:
https://github.com/ros-planning/navigation2/blob/634d2e3d9b6bde5558c90fe6583d52b3ed66cf55/nav2_smac_planner/src/a_star.cpp#L122
can be improved. 
emplace_back will always construct the value even if it is already in the graph. So if it is in the graph, it is created and immediately afterwards destroyed. (see https://en.cppreference.com/w/cpp/container/unordered_map/emplace)

We could spare unnecessary constructions doing sth like this

 
```
  auto iter = graph_.find(index);
  if (iter == graph_.end()) {
    return &(std::get<0>(graph_.emplace(std::piecewise_construct, std::forward_as_tuple(index),
                                        std::forward_as_tuple(index /*initializer list of node t or NodeT(index)*/)))
                 ->second);
  }
  return &(iter->second);
```

Another thing I don't understand is why  AStar reserves _graph.reserve(100000);
Seems a bit random to me. Why not reserving max_iterations?

SteveMacenski:
Please submit a PR, sounds good to me!

> Why not reserving max_iterations?

Frankly, the `max_iterations` didn't exist at the time that was written to optimize performance. But also `max_iterations` isn't the right thing because the graph will contain far more than just the number of iterations of the planner. The iterations of the planner are the number of **visited** nodes in the graph, but theoretically up to `num_prims` number of nodes are **queued** into the graph at each visitation.

mangoschorle:
Will do.

SteveMacenski:
I’d also make sure in a test that it actually is faster. It wouldn’t surprise me if the immediate destruction due to a pre-existing node was faster than the find implementation- even though either way the hash table’s bucket would need to be found. I suppose it depends on how find is implemented and how the compiler optimizes it. I wouldn’t necessarily assume this is faster so it should be an empirical decision based on benchmarking. 

On the max iterations- something to mention is that when the graph is cleared, the memory block is still allocated. So if there’s a much larger graph size after the first plan, it’ll retain that capacity. This initial reserve is mostly just for the 1st planning instance to give it a reasonable starting size rather than incurring multiple copying while the container is resized

mangoschorle:
On a side node before digging into this. Each of the nodes (hybrid, lattice, ..) has a reset() method. This isn't used currently or is it?

mangoschorle:
[...]

PR #3201, opened 2022-09-19
[Smac] Robin hood data structure improves performance by 10-15%!

SteveMacenski:
https://github.com/ros-planning/navigation2/issues/3167


SteveMacenski:
Check that robin hood header is skipped by coverage metrics

PR #3203, opened 2022-09-20
Add a min_obstacle_height param to the nav2_costmap_2d's ObstacleLayer plugin

milidam:
## Basic Info

| ------ | ----------- |

---

When using a PointCloud2 source, nav2_costmap_2d's ObstacleLayer plugin won't clear obstacles detected on the ground when they move away, as the `<source>.min_obstacle_height` parameter either will filter out the ground for the raytracing (if set to > 0.0) or detect it as an obstacle (if set to 0.0).
This problem is similar to [#662](https://github.com/ros-planning/navigation/issues/662).

This PR adds a new global `min_obstacle_height` parameter, allowing the per-source one to be extended to include the range to be taken into account for the retracing, while still filtering obstacles down to that new limit.

/

---

/

#### For Maintainers:

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/obstacle_layer.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/obstacle_layer.hpp
@@ -226,6 +226,7 @@ protected:
     double * max_y);
 
   std::string global_frame_;  ///< @brief The global frame for the costmap
+  double min_obstacle_height_;  ///< @brief Max Obstacle Height
   double max_obstacle_height_;  ///< @brief Max Obstacle Height
 
   /// @brief Used to project laser scans into point clouds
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
Please submit to `main`, we do not accept pull requests directly to released distributions.

By the way, its worth mentioning that we already have the minimum obstacle height, settable in the sources https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/plugins/obstacle_layer.cpp#L140. The global parameter you add makes that work as a backup if the source values are not set, but they are handled themselves within the observation buffer when measurements are received.

milidam:
> Please submit to `main`, we do not accept pull requests directly to released distributions. Please open a PR to `main` and I'll merge the pair.

Thansk for the approval!
The new PR done in #3211

PR #3211, opened 2022-09-22
Add a min_obstacle_height param to the nav2_costmap_2d's ObstacleLayer plugin

milidam:
## Basic Info

| ------ | ----------- |

---

When using a PointCloud2 source, nav2_costmap_2d's ObstacleLayer plugin won't clear obstacles detected on the ground when they move away, as the `<source>.min_obstacle_height` parameter either will filter out the ground for the raytracing (if set to > 0.0) or detect it as an obstacle (if set to 0.0).
This problem is similar to [#662](https://github.com/ros-planning/navigation/issues/662).

This PR adds a new global `min_obstacle_height` parameter, allowing the per-source one to be extended to include the range to be taken into account for the retracing, while still filtering obstacles down to that new limit.

/

---

/

#### For Maintainers:

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/obstacle_layer.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/obstacle_layer.hpp
@@ -226,6 +226,7 @@ protected:
     double * max_y);
 
   std::string global_frame_;  ///< @brief The global frame for the costmap
+  double min_obstacle_height_;  ///< @brief Max Obstacle Height
   double max_obstacle_height_;  ///< @brief Max Obstacle Height
 
   /// @brief Used to project laser scans into point clouds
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

mini-1235:
Given this PR, is it reasonable to add this parameter to voxel layer as well?

SteveMacenski:
We get the parameter, yes, but do we use it anywhere for the voxel grid? The Obstacle Layer's Update Bounds isn't used in the voxel layer I don't think. But we could make use of it easily in the same way in the Voxel Layer (and then add the parameter, since it would _do something_)

```
      // if the obstacle is too low, we won't add it
      if (pz < min_obstacle_height_) {
        RCLCPP_DEBUG(logger_, "The point is too low");
        continue;
      }
```

mini-1235:
Yes, I meant to add this part
```cpp
      // if the obstacle is too low, we won't add it
      if (pz < min_obstacle_height_) {
        RCLCPP_DEBUG(logger_, "The point is too low");
        continue;
      }
```
to voxel layer since this PR do so

SteveMacenski:
OK! Please open the PR and add to the docs :-)

PR #3217, opened 2022-09-22
removing hypotf from smac planner heuristic computation

SteveMacenski:
The impact is small, less than 2% or so, but it definitely doesn't hurt performance and I still see deterministically the same path metrics from experiments of 100 randomly generated paths.


SteveMacenski:
Do in 2D node as well!

PR #3219, opened 2022-09-22
Adding Smac Planner Tolerances for Feasible Planners 

SteveMacenski:
@jwallace42

### DIFF
--- a/nav2_smac_planner/README.md
+++ b/nav2_smac_planner/README.md
@@ -103,12 +103,12 @@ planner_server:
 
     GridBased:
       plugin: "nav2_smac_planner/SmacPlannerHybrid"
-      tolerance: 0.5                      # tolerance for planning if unable to reach exact pose, in meters, for 2D node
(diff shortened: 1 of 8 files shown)


jwallace42:
Cool! 

I think it might be useful to have the ability to constrain the viable region with other geometries like a rectangle. 

The other thing a think is important is the threshold. I think it something that should be determined by the user. 

I think it is reasonable to assume that different goal could have different thresholds.

SteveMacenski:
> I think it might be useful to have the ability to constrain the viable region with other geometries like a rectangle.

That would be... quite strange. I've never heard that request before 

> The other thing a think is important is the threshold. I think it something that should be determined by the user.

 It is, via `tolerance`, so it only returns valid options inside of a configurable tolerance radius around the goal, not just "best effort". 

> I think it is reasonable to assume that different goal could have different thresholds.

It is a dynamically reconfigurable parameter, but the idea of having different tolerance navigation requests has been actually a topic of conversation every few years since ROS 1 navigation was released

Issue #3189, opened 2022-09-14
CI build times out


ruffsl:
Looks like the current CI build is timing out, again. Even with previous countermeasures in place, it appears that these mitigations are no longer sufficient to cut the overall build time to under the maximum one hour limit when building from scratch, or when caching is intentionally busted due to dependency and system package upgrades. For example:

https://app.circleci.com/pipelines/github/ros-planning/navigation2/7903/workflows/d5454092-e664-417b-ab70-b1c05d5c42e7/jobs/27157

The `release_build` job linked above demonstrates a `context deadline exceeded` while attempting to finish build the final workspace package. When caching is used, including ccache and colcon-cache, building the nav2 workspace normally has no such issue. However, when caching is deliberately busted to insure clean hygienic builds upon upstream dependency updates, it seems the current codebase exceeds the deadline limit with respect to CircleCI's free tier jobs.

For reference, resource utilization for the medium docker resource class is almost fully maxed out in terms of core count; exceptions include the bottlenecks in building the first few and last packages in the workspace, such as `nav2_utils` (?) and `nav2_system_tests`.

![image](https://user-images.githubusercontent.com/2293573/190198700-9d289b8c-1f46-41f9-8293-aff94641a66e.png)
 
### Upgrade resource class for CI workers

By default , the currently resource class used for CI docker workers is `medium`, with 2 vCPUs and 4GB of RAM. Aside from the occasional spike once or twice a month when the caches are busted, this hasn't been much of an issue since caching was added:

![image](https://user-images.githubusercontent.com/2293573/190194787-2893bbee-69c2-4621-8533-a7f14b93072d.png)

Given the current issue, one solution would be to simply upgrade the resource class for workers to something more powerful such as `large`, with 4 vCPUs and 8GB of RAM:

https://circleci.com/docs/configuration-reference#resourceclass

![image](https://user-images.githubusercontent.com/2293573/190191333-88f2cb1a-662e-4af9-ab14-b59588158222.png)

See here for such an example:

- https://github.com/ros-planning/navigation2/pull/3188

However, in practice, switching to a `large`er resource class is not enough, given the bottlenecks described above. Despite the principle of diminishing returns, I am also surprised in how unaffected the build time is; perhaps the windows of time where make jobs exceed the former limit of 2 is countered by the greater degree of context switching between make processes? Note: at the 47min mark, the step in the CPU plot remains the same as before with the `medium` resource class.

[...]

Issue #1281, opened 2019-10-23
Travis CI times out when building gazebo_plugins


mkhansen-intel:
# Bug report

Travis CI is timing out at 10 minutes due to no output on the console when building gazebo_plugins.

See issue filed here for details:
https://travis-ci.community/t/help-please-how-to-deal-with-a-travis-timeout-in-the-middle-of-a-multi-package-build/5647

crdelsey:
We could try passing through the console output when we build that package. Presumably, individual files build in less time than 10 minutes.

eg. After line 65, we can add `--event-handlers console_direct+ \`
https://github.com/ros-planning/navigation2/blob/67a92be87f8a00470313c232b1462d45379273e8/Dockerfile#L61-L65

mkhansen-intel:
This is now fixed by #1287

PR #3220, opened 2022-09-23
Improve Colcon Output 

ruffsl:
in an attempt to fix CI Build timed outs.
- https://github.com/ros-planning/navigation2/issues/3189

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -447,6 +447,7 @@ _environments:
     RCUTILS_LOGGING_BUFFERED_STREAM: "0"
     RCUTILS_LOGGING_USE_STDOUT: "0"
     DEBIAN_FRONTEND: "noninteractive"
+    PYTHONUNBUFFERED: "1"
 
 executors:
   release_exec:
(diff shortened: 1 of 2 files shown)


ruffsl:
Disable Output Buffering alone should also fix our context timeout issues, as it should ensure the wait asyncio in colcon does in fact print its `Processing` status output every 30sec during the build:

- https://github.com/colcon/colcon-parallel-executor/blob/09e805d47ac22e38c139251fb7cc83f91402f81e/colcon_parallel_executor/executor/parallel.py#L149-L155

I remeber seeing `console_direct` being used before back from https://github.com/ros-planning/navigation2/issues/1281#issuecomment-545658852 , and decided to re-include it here. As mentioned in the commit message, it adds a bit a noise to the output, but the more detailed progress status could be more helpful.

ruffsl:
Looks building from scratch is no longer timing out.

https://app.circleci.com/pipelines/github/ros-planning/navigation2/7980/workflows/d35040e8-fb34-4871-b7f8-fd96c5eacc13/jobs/27317

I've reverted 92d41f76625d35728ed91f0979c0a960c392a309 as I ony busted the cache for testing. @SteveMacenski I'll let you chose if you want to merge with fe764f6502173d7e0b4a85480a83416a22b31212 , but as I said 0406bd44b3ef2685e06f457059eba53aa9d1f42f alone should also fix things.

SteveMacenski (approved):

SteveMacenski:
Merge as you see fit!

Issue #3213, opened 2022-09-22
costmap_2d_node: enable setting namespace from launch files


tonynajjar:
#### Feature description

I want to run a standalone costmap node and set its (parent) namespace from the launch file because it is dynamically generated.
Currently this is not possible as stated in this [comment](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_ros.cpp#L70) 

This is a feature request to enable this all while retaining the current functionality of setting the namespace via NodeOptions

If you agree, I can contribute here with some guidance what you'd like the design to look like

SteveMacenski:
Seems sensible. You may be able to find the namespaces / remaps in the `main.cpp` file that you can just populate into the constructor so that it uses that code to perform what you want. Its a round-a-bout way of getting the same behavior, but is the lowest technical effort I believe. Though, if you can find another way to messing with node options that doesn't break current behavior, have at it!

PR #3222, opened 2022-09-23
Add new constructor to cosmtap2DROS

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [(3213)](https://github.com/ros-planning/navigation2/issues/3213) |

---

* Added a new constructor in cosmtap2DROS to allow setting namespace and other options from launch file

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -61,6 +61,20 @@ namespace nav2_costmap_2d
 Costmap2DROS::Costmap2DROS(const std::string & name, const bool & use_sim_time)
 : Costmap2DROS(name, "/", name, use_sim_time) {}
 
+Costmap2DROS::Costmap2DROS()
+: nav2_util::LifecycleNode("costmap", ""),
+  name_("costmap"),
+  default_plugins_{"static_layer", "obstacle_layer", "inflation_layer"},
+  default_types_{
+    "nav2_costmap_2d::StaticLayer",
+    "nav2_costmap_2d::ObstacleLayer",
+    "nav2_costmap_2d::InflationLayer"}
+{
+  declare_parameter("map_topic", rclcpp::ParameterValue(std::string("map")));
+  declare_parameter("plugins", rclcpp::ParameterValue(default_plugins_));
+  init();
+}
+
 Costmap2DROS::Costmap2DROS(
   const std::string & name,
   const std::string & parent_namespace,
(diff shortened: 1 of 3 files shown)


tonynajjar:
@SteveMacenski I'm not entirely sure I understood what you meant in the [comment](https://github.com/ros-planning/navigation2/issues/3213#issuecomment-1255323699). Please let me know if I'm far off

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -84,6 +98,15 @@ Costmap2DROS::Costmap2DROS(
     "nav2_costmap_2d::StaticLayer",
     "nav2_costmap_2d::ObstacleLayer",
     "nav2_costmap_2d::InflationLayer"}
+{
+  declare_parameter(
```
SteveMacenski, line 102: Why is `plugins` singled out? That appears the same either way.

SteveMacenski (changes requested):

nav2_costmap_2d/src/costmap_2d_ros.cpp
tonynajjar: Just didn't want to pass it to init() but I can

tonynajjar: Oops, you're right, fixed

tonynajjar (commented):
2 questions:

Can we now remove the (3-years old) TODO from @orduno? I'm not sure what is meant with `Pass a sub-node instead of creating a new node` 

Should this constructor be the default for the standalone node?

SteveMacenski:
Should this new constructor be used in the node main file?

tonynajjar:
> Should this new constructor be used in the node main file?

Jinx :grin: I would think so yes, it can be confusing when the name and namespaces are not being set from the launch file

SteveMacenski:
Just waiting on the main updates and I'll merge. Please prepare a navigation.ros.org PR to detail the change in the migration guide

tonynajjar:
> Just waiting on the main updates and I'll merge

Ah ok was waiting for confirmation. main updated

> Please prepare a navigation.ros.org PR to detail the change in the migration guide

You mean to detail the changes needed in the launch files for people using the standalone node?

SteveMacenski (approved):

SteveMacenski:
Detail that the costmap standalone node has changed, and yes, if any changes are needed to get the previous behavior, what those are

PR #3227, opened 2022-10-02
Controller exceptions

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Add error codes for local controller failures.

---

#### For Maintainers: 


nav2_dwb_controller/dwb_core/include/dwb_core/illegal_trajectory_tracker.hpp
```diff
@@ -66,11 +66,11 @@ class IllegalTrajectoryTracker
  * @brief Thrown when all the trajectories explored are illegal
  */
 class NoLegalTrajectoriesException
```
jwallace42, line 68: > Should these be entirely removed and replaced with nav2_msgs versions?

I don't think so. If you look below it has a specific illegal trajectory tracker for DWB. Once you have MPPI in we could re-evaluate and see if this is something we would like to pull out. Right now it is tied to the critics.

jwallace42:
| First Header  | Second Header | h | g|
| ------------- | ------------- |-----------|----------|---------|
| Content Cell  | Content Cell  |
| Content Cell  | Content Cell  |

nav2_constrained_smoother/src/constrained_smoother.cpp
```diff
@@ -21,7 +21,7 @@
 #include <vector>
 
 #include "nav2_constrained_smoother/constrained_smoother.hpp"
-#include "nav2_core/exceptions.hpp"
+#include "nav2_core/planner_exceptions.hpp"
```
SteveMacenski, line 24: Why not using normal exceptions for the smoothers? Seems odd to use the planner-type exceptions?

nav2_navfn_planner/include/nav2_navfn_planner/navfn_planner.hpp
```diff
@@ -25,7 +25,7 @@
 #include "geometry_msgs/msg/point.hpp"
 #include "geometry_msgs/msg/pose_stamped.hpp"
 #include "nav2_core/global_planner.hpp"
-#include "nav2_core/exceptions.hpp"
+#include "nav2_core/planner_exceptions.hpp"
```
SteveMacenski, line 28: Why are there planner changes here?

nav2_planner/src/planner_server.cpp
```diff
@@ -36,16 +36,14 @@ using namespace std::chrono_literals;
 using rcl_interfaces::msg::ParameterType;
 using std::placeholders::_1;
 
-namespace nav2_planner
```
SteveMacenski, line 39: Why are there planner server changes here? This is like 200 line code diff!

nav2_msgs/action/FollowPath.action
```diff
@@ -1,10 +1,18 @@
+# Error codes
+int16 INVALID=0
+int16 UNKNOWN=1
+int16 FAILED_TO_MAKE_PROGRESS=2
+int16 PATIENCE_EXCEEDED=3
+int16 INVALID_PATH=4
```
SteveMacenski, line 6: For discussion on exception codes:

These all make sense to me, but should there also be ones for:
- TF
- Could not generate
- Trajectories are in collision
- any others?

Also, `0` should be `NONE`, no?

SteveMacenski (changes requested):

nav2_navfn_planner/include/nav2_navfn_planner/navfn_planner.hpp
jwallace42: I renamed the exceptions to planner_exceptions to keep with the naming convention that I imposed. 

planner_exceptions
controller_exceptions

nav2_msgs/action/FollowPath.action
jwallace42: So, 

NoValidTrajectory (I can see a place to add that for dwb) -> Trajectories are in collision. 

I don't see a place for _could not generate trajectory_ in dwb. It seems like trajectory generation shouldn't cause any issues since it is just a projection. But maybe you have something in mind for MPPI? 

Fixed the message :). 

@doisyg are there any controller errors that you think should be added? 

```
int16 NONE=0
int16 UNKNOWN=1
int16 TF_ERROR=2
int16 NO_VALID_TRAJECTORIES=3
int16 FAILED_TO_MAKE_PROGRESS=4
int16 PATIENCE_EXCEEDED=5
int16 INVALID_PATH=6
```

nav2_core/include/nav2_core/controller_exceptions.hpp
```diff
@@ -0,0 +1,67 @@
...
+};
+
+class InvalidPath : public ControllerException
+{
+public:
+  explicit InvalidPath(const std::string & description)
+  : ControllerException(description) {}
+};
+
+class NoValidTrajectories : public ControllerException
```
SteveMacenski, line 58: `NoValidControl` perhaps? Its semantics, but technically the plugins here don't _need_ to compute a trajectory (even though by defacto standard they do). I think this language should be kept consistent with the controller server's name

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -352,8 +352,8 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
   // Collision checking on this velocity heading
   const double & carrot_dist = hypot(carrot_pose.pose.position.x, carrot_pose.pose.position.y);
-  if (use_collision_detection_ && isCollisionImminent(pose, linear_vel, angular_vel, carrot_dist)) {
```
SteveMacenski, line 355: Deleted the configuration added here - maybe a merge conflict resolution mistake?

nav2_msgs/action/FollowPath.action
```diff
@@ -1,10 +1,21 @@
+# Error codes
+# Note: The expected priority order of the errors should match the message order
+int16 NONE=0
+int16 UNKNOWN=1
+int16 TF_ERROR=2
+int16 NO_VALID_TRAJECTORIES=3
```
SteveMacenski, line 6: NO_VALID_CONTROL

nav2_controller/src/controller_server.cpp
```diff
@@ -446,11 +481,11 @@ void ControllerServer::computeAndPublishVelocity()
   geometry_msgs::msg::PoseStamped pose;
 
   if (!getRobotPose(pose)) {
-    throw nav2_core::PlannerException("Failed to obtain robot pose");
+    throw nav2_core::ControllerException("Failed to obtain robot pose");
```
SteveMacenski, line 484: TF error?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -333,7 +343,9 @@ DWBLocalPlanner::computeVelocityCommands(
     pub_->publishLocalPlan(pose.header, empty_traj);
     pub_->publishCostGrid(costmap_ros_, critics_);
 
-    throw;
+    throw nav2_core::ControllerException(
```
SteveMacenski, line 346: NO_VALID_CONTROL ?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -450,8 +462,8 @@ DWBLocalPlanner::transformGlobalPlan(
       tf_, global_plan_.header.frame_id, pose,
       robot_pose, transform_tolerance_))
   {
-    throw dwb_core::
-          PlannerTFException("Unable to transform robot pose into global plan's frame");
+    throw nav2_core::
+          ControllerTFError("Unable to transform robot pose into global plan's frame");
```
SteveMacenski, line 466: TF error?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -535,7 +547,7 @@ DWBLocalPlanner::transformGlobalPlan(
   }
 
   if (transformed_plan.poses.empty()) {
-    throw nav2_core::PlannerException("Resulting plan has 0 poses in it.");
+    throw nav2_core::ControllerTFError("Resulting plan has 0 poses in it.");
```
SteveMacenski, line 550: Invalid path?

nav2_dwb_controller/dwb_core/src/trajectory_utils.cpp
```diff
@@ -49,7 +49,7 @@ const geometry_msgs::msg::Pose2D & getClosestPose(
   rclcpp::Duration goal_time = rclcpp::Duration::from_seconds(time_offset);
   const unsigned int num_poses = trajectory.poses.size();
   if (num_poses == 0) {
-    throw nav2_core::PlannerException("Cannot call getClosestPose on empty trajectory.");
+    throw nav2_core::ControllerException("Cannot call getClosestPose on empty trajectory.");
```
SteveMacenski, line 52: Invalid path?

nav2_dwb_controller/dwb_core/src/trajectory_utils.cpp
```diff
@@ -73,7 +73,7 @@ geometry_msgs::msg::Pose2D projectPose(
   rclcpp::Duration goal_time = rclcpp::Duration::from_seconds(time_offset);
   const unsigned int num_poses = trajectory.poses.size();
   if (num_poses == 0) {
-    throw nav2_core::PlannerException("Cannot call projectPose on empty trajectory.");
+    throw nav2_core::ControllerException("Cannot call projectPose on empty trajectory.");
```
SteveMacenski, line 76: Invalid path?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -352,8 +352,8 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
-    throw nav2_core::PlannerException("RegulatedPurePursuitController detected collision ahead!");
+  if (isCollisionImminent(pose, linear_vel, angular_vel, carrot_dist)) {
+    throw nav2_core::ControllerException("RegulatedPurePursuitController detected collision ahead!");
```
SteveMacenski, line 356: NO VALID CONTROL?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -749,7 +751,7 @@ nav_msgs::msg::Path RegulatedPurePursuitController::transformGlobalPlan(
   global_path_pub_->publish(transformed_plan);
 
   if (transformed_plan.poses.empty()) {
-    throw nav2_core::PlannerException("Resulting plan has 0 poses in it.");
+    throw nav2_core::ControllerTFError("Resulting plan has 0 poses in it.");
```
SteveMacenski, line 754: Invalid path?

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -259,11 +258,11 @@ void RotationShimController::isCollisionFree(
     if (footprint_cost == static_cast<double>(NO_INFORMATION) &&
       costmap_ros_->getLayeredCostmap()->isTrackingUnknown())
     {
-      throw std::runtime_error("RotationShimController detected a potential collision ahead!");
+      throw nav2_core::ControllerException("RotationShimController detected a potential collision ahead!");
```
SteveMacenski, line 261: NO VALID CONTROL?

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -259,11 +258,11 @@ void RotationShimController::isCollisionFree(
...
     }
 
     if (footprint_cost >= static_cast<double>(LETHAL_OBSTACLE)) {
-      throw std::runtime_error("RotationShimController detected collision ahead!");
+      throw nav2_core::ControllerException("RotationShimController detected collision ahead!");
```
SteveMacenski, line 265: NO VALID CONTROL?

SteveMacenski (changes requested):
Generally LGTM, but seems like some of the control exceptions could be more specific than those they were changed to generically as `ControlException`

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -253,9 +257,15 @@ DWBLocalPlanner::computeVelocityCommands(
...
     return cmd_vel;
-  } catch (const nav2_core::PlannerException & e) {
+  } catch (const nav2_core::ControllerTFError & e) {
     pub_->publishEvaluation(results);
-    throw;
+    throw e;
+  } catch (const nav2_core::InvalidPath & e) {
+    pub_->publishEvaluation(results);
+    throw e;
+  } catch (const nav2_core::ControllerException & e) {
```
SteveMacenski, line 266: Since they all derive from `ControllerException` and they are all handled the same, I think only this is required - though I actually don't know when its re-thrown if it gets rethrown by the base `ControllerException` type or the derived type

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -204,7 +203,7 @@ RotationShimController::transformPoseToBaseFrame(const geometry_msgs::msg::PoseS
 {
   geometry_msgs::msg::PoseStamped pt_base;
   if (!nav2_util::transformPoseInTargetFrame(pt, pt_base, *tf_, costmap_ros_->getBaseFrameID())) {
-    throw nav2_core::PlannerException("Failed to transform pose to base frame!");
+    throw nav2_core::ControllerException("Failed to transform pose to base frame!");
```
SteveMacenski, line 206: TF exception?

SteveMacenski (changes requested):
LGTM -- though I had to revert the planner server eq. to this due to a bunch of false detections of collisions in the NavFn planner plugin (maybe others, but that's what i was using at the time). I think its worth looking into that to resolve and then merging these as pairs -- since something an issue there may also be an issue here

nav2_msgs/action/FollowPath.action
```diff
@@ -1,10 +1,21 @@
+# Error codes
+# Note: The expected priority order of the errors should match the message order
+int16 NONE=0
```
SteveMacenski, line 3: For ordering,

None, Unknown, TF, invalid path, patience, progress, no valid control?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
jwallace42: I just tested, 

It strips the class back to `ControllerException` so I think we have to keep it for now :(.

nav2_controller/src/controller_server.cpp
```diff
@@ -405,10 +403,47 @@ void ControllerServer::computeControl()
           controller_frequency_);
       }
     }
-  } catch (nav2_core::PlannerException & e) {
-    RCLCPP_ERROR(this->get_logger(), e.what());
+  } catch (nav2_core::ControllerTFError & e) {
+    RCLCPP_ERROR(this->get_logger(), "%s", e.what());
+    publishZeroVelocity();
+    std::shared_ptr<nav2_msgs::action::FollowPath::Result> result = std::make_shared<nav2_msgs::action::FollowPath::Result>();
```
SteveMacenski, line 409: There's almost no way this meets the 100 characters line limit ;-)

nav2_controller/src/controller_server.cpp
```diff
@@ -405,10 +403,47 @@ void ControllerServer::computeControl()
...
+    result->error_code = nav2_msgs::action::FollowPath::Goal::PATIENCE_EXCEEDED;
+    action_server_->terminate_current(result);
+    return;
+  } catch (nav2_core::InvalidPath & e) {
+    RCLCPP_ERROR(this->get_logger(), "%s", e.what());
+    publishZeroVelocity();
+   std::shared_ptr<nav2_msgs::action::FollowPath::Result> result = std::make_shared<nav2_msgs::action::FollowPath::Result>();
```
SteveMacenski, line 437: Missing space

SteveMacenski:
There are a bunch of conflicts that need to be resolved from the planner execeptions PR

jwallace42:
I was just adding in the last few final touches. I have been testing with DWB in sim and it has gone well.

jwallace42:
If you could take a close look at the `computeAndPublishVelocity` method that would be great. 

Before, 

All exceptions would be caught here. Now it is only NoValidControl is caught. So, the behavior has changed very slightly but I think this was the original intention.

nav2_controller/src/controller_server.cpp
```diff
@@ -464,7 +499,7 @@ void ControllerServer::computeAndPublishVelocity()
       nav_2d_utils::twist2Dto3D(twist),
       goal_checkers_[current_goal_checker_].get());
     last_valid_cmd_time_ = now();
-  } catch (nav2_core::PlannerException & e) {
+  } catch (nav2_core::NoValidControl & e) {
```
SteveMacenski, line 502: I think this is sensible. The only other 2 conditions are TF and path invalid. For TF, that's usually from critically misconfigured robots that would be good to tell a user to "fix now" versus try to mask it. Path invalid is more... questionable, but I suppose the boundary conditions of the request won't change if the path is the reason for failure until a new path is given, so why even bother trying. 

Good call! 

I do wonder though about more generic throws due to "other" issues. We don't catch generic exceptions here - but if we did, we'd get all those other TF/plan/etc ones too, annoyingly. I suppose we could create a `GenericControllerError` from the base `ControllerException` and then capture the generic one, but maybe I'm over-optimizing for edge cases that don't exist yet.

SteveMacenski, line 502: This is subtle - add a comment here: "Only no valid control exception types are valid to attempt to have control patience, as other types will not be resolved with more attempts."

SteveMacenski (approved):

Issue #3206, opened 2022-09-21
Binary costmap filter


SteveMacenski:
For binary on / off service of costmap filters 

Parameters:
- Threshold "on" / "off"
- Service name

Example uses:
- turn off camera in sensitive areas
- turn on headlights / leds / other safety things when its outside 
- switch operating mode

PR #3228, opened 2022-10-04
Add binary flip costmap filter

AlexeyMerzlyakov:
Binary flip costmap filter

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3206 |
| Robotic platform tested on | TB3 Gazebo verification, unit testing, stability testing ~20 minutes in a row, codecov 95.9% of new code covered |

---

* Added binary flip filter
* Added bool topics publishing instead of services: Otherwise, if we will add service client on BinaryFilter side, we also have to implement additional service server for requests handling, which is architecturally more complicated than message broadcasting.
* New BinaryState.msg containing bool flip state

---

#### For Maintainers: 


nav2_costmap_2d/include/nav2_costmap_2d/costmap_filters/binary_filter.hpp
```diff
@@ -0,0 +1,152 @@
+/*********************************************************************
+ *
+ * Software License Agreement (BSD License)
```
SteveMacenski, line 3: Apache?

nav2_msgs/msg/BinaryState.msg
```diff
@@ -0,0 +1,2 @@
+# Binary state
+bool state
```
SteveMacenski, line 2: `std_msgs/Bool`?

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,310 @@
...
+      "BinaryFilter: New filter mask arrived from %s topic. Updating old filter mask.",
+      mask_topic_.c_str());
+    filter_mask_.reset();
+  }
+
+  filter_mask_ = msg;
+  mask_frame_ = msg->header.frame_id;
+}
+
+bool BinaryFilter::transformPose(
```
SteveMacenski, line 158: Is this function copy pasted from other filters? We could add this to the filter base class. Seems generally useful.

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,310 @@
...
+  // Get global frame required for binary state publisher
+  global_frame_ = layered_costmap_->getGlobalFrameID();
+
+  // Create new binary state publisher
+  binary_state_pub_ = node->create_publisher<nav2_msgs::msg::BinaryState>(
+    binary_state_topic, rclcpp::QoS(10));
+  binary_state_pub_->on_activate();
+
+  // Initialize state as "false" by-default
+  changeState(false);
```
SteveMacenski, line 94: I think there should be a param for the default value so that we can initialize / reset by user requested behavior.

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,310 @@
...
+  } else {
+    // Filter mask and current layer are in the same frame:
+    // Just use pose coordinates
+    mask_pose = pose;
+  }
+
+  return true;
+}
+
+bool BinaryFilter::worldToMask(double wx, double wy, unsigned int & mx, unsigned int & my) const
```
SteveMacenski, line 195: Same here -- if there are any of these functions that are reimplemented in multiple filters that might be helpful for future implementation, we might want to abstract them a layer down. This seems like another good candidate

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,310 @@
...
+  // Transforming robot pose from current layer frame to mask frame
+  if (!transformPose(pose, mask_pose)) {
+    return;
+  }
+
+  // Converting mask_pose robot position to filter_mask_ indexes (mask_robot_i, mask_robot_j)
+  unsigned int mask_robot_i, mask_robot_j;
+  if (!worldToMask(mask_pose.x, mask_pose.y, mask_robot_i, mask_robot_j)) {
+    // Robot went out of mask range. Set "false" state by-default
+    changeState(false);
```
SteveMacenski, line 248: Use default value -- also log a warning here that we're outside of the mask.

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,310 @@
...
+      changeState(true);
+    }
+  } else {
+    if (binary_state_) {
+      changeState(false);
+    }
+  }
+}
+
+void BinaryFilter::resetFilter()
```
SteveMacenski, line 269: Reset to default value

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,310 @@
...
+  if (!worldToMask(mask_pose.x, mask_pose.y, mask_robot_i, mask_robot_j)) {
+    // Robot went out of mask range. Set "false" state by-default
+    changeState(false);
+    return;
+  }
+
+  // Getting filter_mask data from cell where the robot placed
+  int8_t mask_data = getMaskData(mask_robot_i, mask_robot_j);
+  // If filter_mask data is not empty, change binary state (if necessary)
+  if (
```
SteveMacenski, line 255: Shouldn't there be some checking of the costmap filter info about how to interpret these values?

SteveMacenski (changes requested):

nav2_costmap_2d/include/nav2_costmap_2d/costmap_filters/binary_filter.hpp
AlexeyMerzlyakov: Binary filter source was taken from [speed_filter.cpp](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/plugins/costmap_filters/speed_filter.cpp) and [keepout_filter.cpp](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp). Both of these sources are being a part of `Costmap2D` and having a BSD-3 license. Therefore, I think we should distribute all derivative works under the same license.

The opposite situation for the testcases, that were developed "from the scratch", so they having more actual Apache-2.0 license.

nav2_msgs/msg/BinaryState.msg
AlexeyMerzlyakov: They are officially recommend to have a new message for that and seems deprecated since Foxy:
```
cat ros2/common_interfaces/std_msgs/msg/Bool.msg
# This was originally provided as an example message.
# It is deprecated as of Foxy
# It is recommended to create your own semantically meaningful message.
# However if you would like to continue using this please use the equivalent in example_msgs.

bool data
```
We have also `ros2/example_interfaces/msg/Bool.msg` which was not deprecated, but still recommending to make your own service. I'd like to agree, that utilizing an existing service is better to reinvent the wheel. But will it be OK, if we will use `example_interfaces` as a messages utilized by Nav2?

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: I am not sure about the intention of this:
* If you are talking about the mask thresholds for "on" and "off", this already exists in input mask YAML-files: `trinary` mode could be used, and necessary threshold is being set for input map.
* Someone wants to set some specific value for flipping only. In this case, binary filter is easy thing, which flips some state to `true` or `false`, and I do not believe that we need to complicate the architecture (add new fields to the `nav2_msgs/msg/CostmapFilterInfo.msg` messages only because of this use-case). In my intention, the binary filter should work simply, as Keepout Filter: for "non-0" values it will flip the filter to `true`, otherwise - `false`.
* If someone want to set e.g. `UNKNOWN` values to turn-on the filter, I am not sure that this is real use-case. Where could we apply it? Even if so, the map mask could be easily changed by moving `occupied_thresh` in YAML-file to have former unknowns to be recognized by map server as occupied values.

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: Okay, done in https://github.com/ros-planning/navigation2/pull/3228/commits/9fba2a9bff21d33455daf83e24142179552d03ec

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3228/commits/9fba2a9bff21d33455daf83e24142179552d03ec

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: Added default (and parametrized by user) binary filter state in https://github.com/ros-planning/navigation2/pull/3228/commits/056ed5e09f0092855591bed2223511fc8d0cac71

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3228/commits/056ed5e09f0092855591bed2223511fc8d0cac71

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3228/commits/056ed5e09f0092855591bed2223511fc8d0cac71

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
SteveMacenski: I meant that we have this whole standard about how to intepret the costmap 0-255 data to represent different things like keepout/speed/etc data. It seems odd that this would not use that information from the costmap filter info given to it. The unknown/free space special numbers are for cost values, but that's not what the filter mask represents by the documentation we've generated.

There should be some threshold value for on/off and using the costmap info base/multiplier to get it. By default, it could literally just be any non-0 or non-255 value so that it, in effect, works that way. But it should be setable by users.

nav2_msgs/msg/BinaryState.msg
SteveMacenski: https://github.com/ros-planning/navigation2/search?p=2&q=std_msgs

We have hundreds of uses of std_msgs, I doubt that'll be actually deprecated in another 5 years. What's 1 more?

AlexeyMerzlyakov: Hm... Despite the `std_msgs/msg/Bool.msg` is claimed to be deprecated in ROS2, I see no build or other issues after I've changed it. So, the https://github.com/ros-planning/navigation2/pull/3228/commits/07ada1f2212291a86592bd0b4399bd5d1e1998f0 does this.

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: Okay, looks like I've understood your intention. Added arbitrary filter space values and `flip_threshold` parameter to BinaryFilter: https://github.com/ros-planning/navigation2/pull/3228/commits/8f58d33e3640e8a239a324c8baab1b8c358eb399

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,268 @@
...
+    RCLCPP_WARN(
+      logger_,
+      "BinaryFilter: Robot is outside of filter mask. Resetting binary state to default.");
+    changeState(default_state_);
+    return;
+  }
+
+  // Getting filter_mask data from cell where the robot placed
+  int8_t mask_data = getMaskData(filter_mask_, mask_robot_i, mask_robot_j);
+  if (mask_data == nav2_util::OCC_GRID_UNKNOWN) {
```
SteveMacenski, line 203: Remove, the unknown costs are only relevant for cost values, this is not that :smile:

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
```diff
@@ -0,0 +1,268 @@
...
+    // Warn and do nothing.
+    RCLCPP_WARN_THROTTLE(
+      logger_, *(clock_), 2000,
+      "BinaryFilter: Filter mask [%i, %i] data is unknown. Do nothing.",
+      mask_robot_i, mask_robot_j);
+    return;
+  }
+  // Check and flip binary state, if necessary
+  if (base_ + mask_data * multiplier_ > flip_threshold_) {
+    if (binary_state_ == default_state_) {
```
SteveMacenski, line 214: I don't understand the comparison with `default_state_`. I don't think there should be any checking happening here beyond just that binary_state_ is true or false in order to either change the state or not based on if the data is above the threshold e.g.

```
if (!binary_state_) {changeState(false);}
```

SteveMacenski: Or is your intention that above the threshold moves from the default state to the non-default state? I suppose that would be more general.

nav2_costmap_2d/plugins/costmap_filters/binary_filter.cpp
AlexeyMerzlyakov: The `mask_data` does not belong to Costmap2D space, it is `OccupancyGrid` value.
OccGrid values  are in the range from `[0,100]`, or unknown which is `-1`. This is exception value we do not want to decode back to the filter space value, therefore this situation should be pulled-out and handled separately.

AlexeyMerzlyakov: The similar case is already [handled](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/plugins/costmap_filters/speed_filter.cpp#L276) in `SpeedFilter`, but for some reasons (I can not recall why), we are using separate const value names for it.

SteveMacenski: OK. I didn't recall that.

SteveMacenski (approved):

SteveMacenski:
Have a conflict, but otherwise LGTM

AlexeyMerzlyakov:
The changes were rebased over latest `main` branch, so it now looks ready to merge. @SteveMacenski, thank you for the review!

PR #3236, opened 2022-10-10
Add Path Smoothers Benchmarking suite

AlexeyMerzlyakov:
The task - is a continue of #2837, adopted for Smoother Benchmarking.

* Took as the basis the suite prepared for path planning, made by @jwallace42 
* Adopted the suite to run smoothers benchmarking

---

* For this change it is required to have all latest changes to be back-ported (in a separate PR) from Smac Simple Smoother back to `nav2_smoother`

#### For Maintainers: 


tools/smoother_benchmarking/smoother_benchmark_bringup.py
```diff
@@ -0,0 +1,106 @@
...
+from ament_index_python.packages import get_package_share_directory
+
+def generate_launch_description():
+    nav2_bringup_dir = get_package_share_directory('nav2_bringup')
+    python_commander_dir = get_package_share_directory('nav2_simple_commander')
+
+    map_yaml_file = os.path.join(os.getcwd(), 'maps', 'smoothers_world.yaml')
+
+    # start the simulation
+    start_gazebo_server_cmd = ExecuteProcess(
```
SteveMacenski, line 30: Shouldn't you just need the planner/smoother servers? I don't think you should need gazebo, no?

tools/smoother_benchmarking/metrics.py
```diff
@@ -0,0 +1,173 @@
...
+    res = costmap_msg.metadata.resolution
+    while i < random_pairs:
+        print("Cycle: ", i, "out of: ", random_pairs)
+        start = getRandomStart(costmap, max_cost, side_buffer, time_stamp, res)
+        goal = getRandomGoal(costmap, start, max_cost, side_buffer, time_stamp, res)
+        print("Start", start)
+        print("Goal", goal)
+        result = getPlannerResults(navigator, start, goal, planner)
+        if result is not None:
+            print("MAO: result = ", result)
```
SteveMacenski, line 146: MAO?

tools/smoother_benchmarking/README.md
```diff
@@ -0,0 +1,121 @@
...
+```
+diff --git a/nav2_planner/src/planner_server.cpp b/nav2_planner/src/planner_server.cpp
+index c7a90bcb..510bb6ea 100644
+--- a/nav2_planner/src/planner_server.cpp
++++ b/nav2_planner/src/planner_server.cpp
+@@ -285,7 +285,7 @@ bool PlannerServer::getStartPose(
+   typename std::shared_ptr<const typename T::Goal> goal,
+   geometry_msgs::msg::PoseStamped & start)
+ {
+-  if (goal->use_start) {
```
SteveMacenski, line 24: I don't understand this diff, you can just set `use_start = True` in your `getPlan` request and it'll do this without modifying the source code

tools/smoother_benchmarking/README.md
```diff
@@ -0,0 +1,121 @@
...
+   } else if (!costmap_ros_->getRobotPose(start)) {
+     return false;
+@@ -381,7 +381,10 @@ void PlannerServer::computePlanThroughPoses()
+       }
+ 
+       // Get plan from start -> goal
++      auto planning_start = steady_clock_.now();
+       nav_msgs::msg::Path curr_path = getPlan(curr_start, curr_goal, goal->planner_id);
++      auto planning_duration = steady_clock_.now() - planning_start;
++      result->planning_time = planning_duration;
```
SteveMacenski, line 36: Why move this?

tools/smoother_benchmarking/README.md
```diff
@@ -0,0 +1,121 @@
...
+       RCLCPP_WARN(
+diff --git a/nav2_smoother/src/nav2_smoother.cpp b/nav2_smoother/src/nav2_smoother.cpp
+index ada1f664..610e9512 100644
+--- a/nav2_smoother/src/nav2_smoother.cpp
++++ b/nav2_smoother/src/nav2_smoother.cpp
+@@ -253,8 +253,6 @@ bool SmootherServer::findSmootherId(
+ 
+ void SmootherServer::smoothPlan()
+ {
+-  auto start_time = steady_clock_.now();
```
SteveMacenski, line 57: Why move this?

tools/smoother_benchmarking/README.md
```diff
@@ -0,0 +1,121 @@
...
+      w_smooth: 100000.0 # tuned
+```
+
+Then enable SmoothPath sequence for selected path planner, as shown for default `navigate_to_pose_w_replanning_and_recovery.xml` behavior tree:
+
+```
+       <PipelineSequence name="NavigateWithReplanning">
+         <RateController hz="1.0">
+           <RecoveryNode number_of_retries="1" name="ComputePathToPose">
+-            <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased"/>
```
SteveMacenski, line 108: Why modify the BT? You're just calling the task servers to get the path / smoothing results in the data collection script

SteveMacenski (changes requested):

tools/smoother_benchmarking/smoother_benchmark_bringup.py
SteveMacenski: If you look at the planner benchmark launch file,  I think this would be the same + the smoother server, no?

AlexeyMerzlyakov: [Fixed](https://github.com/ros-planning/navigation2/pull/3236/files#diff-6858250f29552fae42d8964a159abe07361c54244c44a0f848dec64193818c7a). The latest `smoother_benchmark_bringup.py` file was derived from `smoother_benchmark_bringup.py` with `smoother_server` added.

tools/smoother_benchmarking/metrics.py
AlexeyMerzlyakov: That was the initials of my second, first and middle name, I've used for debugging :smiley: Removed in the latest update.

tools/smoother_benchmarking/README.md
AlexeyMerzlyakov: Yep. Missed it. Set `use_start = True` for `getPath` [instead](https://github.com/ros-planning/navigation2/pull/3236/files#diff-65e8732e1465516bc6d72d450f91fb172336c4e37ece025895c3e197ef6432a6R36).

tools/smoother_benchmarking/README.md
AlexeyMerzlyakov: Indeed, this is only required when we are using standard operation flow. When calling actions from palth planner / smoother servers, we do not need it at all. Removed from README.

tools/smoother_benchmarking/README.md
AlexeyMerzlyakov: Initially, the time measured for planner and smoother actions - is a whole period of time required for the selected action. This time includes a lot of of supplementary cycles, like TF to global frame or waiting for costmaps. During the benchmarking, we need to know exact time of `getPlan()` for the selected path planner and `smooth()` routine for smoothers case. That is why for the most accurate results, start and end time measurements should be moved right before and after measured routines, to exclude all unnecessary wrapping/transformation/converting/preparation/etc. stuff

tools/smoother_benchmarking/README.md
AlexeyMerzlyakov: This was moved for the same [reasons](https://github.com/ros-planning/navigation2/pull/3236#discussion_r992016301).

jwallace42:
It might be worth it to start thinking about adding a bench-marking package to facilitate future bench-marking tasks.

tools/smoother_benchmarking/metrics.py
SteveMacenski: Ok! I use "STEVE" for similar so I know what you mean.  Just something easy your brain can pick up the pattern in easily in a dense log file

SteveMacenski:
Agreed. I think we need a controller one and then we're sufficient I think to do that. Probably the same random environment with some static path we use and then take metrics like time to goal, distance traversed, "smoothness" of velocities and behaviors in some metrics, closeness to obstacles, maybe maximum possible run-time. Maybe add option to have random dynamic obstacles in the way too

tools/smoother_benchmarking/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+Bechmarking scripts require the following python packages to be installed:
+
+```
+pip install transforms3d
+pip install seaborn
+pip install tabulate
+```
+
+For the benchmarking purposes, the clarification of execution time should be made for planner and smoother servers:
```
SteveMacenski, line 13:
```suggestion
For the benchmarking purposes, the clarification of execution time may be made for planner and smoother servers, to reduce impacts caused by other system actions outside of the planning / smoothing algorithm (optional):
```

SteveMacenski, line 13: Move this stuff to be last, after the config file modifications

SteveMacenski (changes requested):

tools/smoother_benchmarking/README.md
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3236/commits/5b8c3a23baf9fa68b7ff38c18dbaf5b07fac8bab

tools/smoother_benchmarking/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+... and path smoothers for benchmark:
+
+```
+ smoother_server:
+   ros__parameters:
+    smoother_plugins: ["simple_smoother", "constrained_smoother"]
+    simple_smoother:
+      plugin: "nav2_smoother::SimpleSmoother"
+      holonomic: False # should be disabled to utilize motion/rotation model
```
SteveMacenski, line 36: Should this be here if we didn't add the Smac stuff here?

SteveMacenski (commented):
Otherwise LGTM to merge

tools/smoother_benchmarking/README.md
AlexeyMerzlyakov: Yes, thanks for noting this. Removed in latest commit.

SteveMacenski (approved):

SteveMacenski:
Awesome, thanks! This is such an unexpected but really nice thing that we've been able to add from the nav2 algorithms paper. Having some benchmarking scripts -- even if simple -- is going to pay dividends in the long-term future.

PR #3240, opened 2022-10-11
Update controller server goal checker

NicolasRochaPacheco:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Custom skid steer drive robot under gazebo  |

---

* We identified that when trying to change the controller and the goal checker on a FollowPath action, only the controller was updated.
* We tracked down this to a function in the controller server in which only the controller was updated. 
* After doing the changes included in this PR, we confirmed that the goal checker change was actually done.

* No documentation is required.

---

* No future work is required.

#### For Maintainers: 


SteveMacenski (approved):
Good call, thanks!

SteveMacenski:
Pull in `main` to retrigger CI, the system should turn over properly

Issue #3235, opened 2022-10-06
[Controller] invalid ptr during mapping and navigating possibly caused by bugs in nav2_costmap_2d


Cryst4L9527:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - foxy
- Version or commit hash:
  - a7955d891bcee5ad3f8acecc6ea3bdb1ddc73ce9
- DDS implementation:
  - Fast-RTPS（default）

#### Steps to reproduce issue
just like the issue #3231 ,except for the configall.yaml:
```
/loc2d_ros:
  ros__parameters:
    base_frame_id: base_footprint
    d_thresh: 0.02
    global_frame_id: map
    initial_pos_a: 0.0
    initial_pos_x: -2.0
    initial_pos_y: -0.5
    odom_frame_id: odom
    scan_topic: /scan
    transform_tolerance: 0.2
    use_sim_time: true
amcl:
  ros__parameters:
    alpha1: 0.2
    alpha2: 0.2
    alpha3: 0.2
    alpha4: 0.2
    alpha5: 0.2
    base_frame_id: base_footprint
    beam_skip_distance: 0.5
    beam_skip_error_threshold: 0.9
    beam_skip_threshold: 0.3
    do_beamskip: false
    global_frame_id: map
    initial_pose:
      x: -2.0
      y: -0.5
      yaw: 0.0
      z: 0.0
    lambda_short: 2.3000000000000003
    laser_likelihood_max_dist: 2.0
    laser_max_range: 100.0
    laser_min_range: -4.4
    laser_model_type: likelihood_field
    max_beams: 59
    max_particles: 2000
    min_particles: 521
    odom_frame_id: odom
    pf_err: 0.05
    pf_z: 0.99
    recovery_alpha_fast: 0.0
    recovery_alpha_slow: 0.0
    resample_interval: 1
    robot_model_type: differential
    save_pose_rate: 0.5
    set_initial_pose: true
    sigma_hit: 0.2
    tf_broadcast: true
    transform_tolerance: 1.0
    update_min_a: 0.2
    update_min_d: 0.25
    use_sim_time: true
    z_hit: 10.4
    z_max: -10.55
    z_rand: 5.2
    z_short: -5.3500000000000005
amcl_map_client:
  ros__parameters:
    use_sim_time: true
amcl_rclcpp_node:
  ros__parameters:
    use_sim_time: true
bt_navigator:
  ros__parameters:
[...]

Issue #3233, opened 2022-10-06
[nav2_amcl] heap-buffer-overflow during mapping and navigating possibly caused by bugs in map_cspace.cpp


Cryst4L9527:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - foxy
- Version or commit hash:
  - a7955d891bcee5ad3f8acecc6ea3bdb1ddc73ce9
- DDS implementation:
  - Fast-RTPS（default）

#### Steps to reproduce issue
just like the issue #3231 ,except for the configall.yaml:
```
/loc2d_ros:
  ros__parameters:
    base_frame_id: base_footprint
    d_thresh: 0.02
    global_frame_id: map
    initial_pos_a: 0.0
    initial_pos_x: -2.0
    initial_pos_y: -0.5
    odom_frame_id: odom
    scan_topic: /scan
    transform_tolerance: 0.2
    use_sim_time: true
amcl:
  ros__parameters:
    alpha1: 0.2
    alpha2: 0.2
    alpha3: 0.2
    alpha4: 0.2
    alpha5: 0.2
    base_frame_id: base_footprint
    beam_skip_distance: 0.5
    beam_skip_error_threshold: 0.9
    beam_skip_threshold: 0.3
    do_beamskip: false
    global_frame_id: map
    initial_pose:
      x: -2.0
      y: -0.5
      yaw: 0.0
      z: 0.0
    lambda_short: 0.1
    laser_likelihood_max_dist: 2.0
    laser_max_range: 100.0
    laser_min_range: -1.0
    laser_model_type: likelihood_field
    max_beams: 60
    max_particles: 2000
    min_particles: 500
    odom_frame_id: odom
    pf_err: 0.05
    pf_z: 0.99
    recovery_alpha_fast: 0.0
    recovery_alpha_slow: 0.0
    resample_interval: 1
    robot_model_type: differential
    save_pose_rate: 0.5
    set_initial_pose: true
    sigma_hit: 0.2
    tf_broadcast: true
    transform_tolerance: 1.0
    update_min_a: 0.2
    update_min_d: 0.25
    use_sim_time: true
    z_hit: 0.5
    z_max: 0.05
    z_rand: 0.5
    z_short: 0.05
amcl_map_client:
  ros__parameters:
    use_sim_time: true
amcl_rclcpp_node:
  ros__parameters:
    use_sim_time: true
bt_navigator:
  ros__parameters:
[...]

Issue #3232, opened 2022-10-06
[Controller] invalid ptr during mapping and navigating possibly caused by bugs in nav2_costmap_2d


Cryst4L9527:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - foxy
- Version or commit hash:
  - a7955d891bcee5ad3f8acecc6ea3bdb1ddc73ce9
- DDS implementation:
  - Fast-RTPS（default）

#### Steps to reproduce issue
just like the issue #3231 ,except for the configall.yaml:
```
/loc2d_ros:
  ros__parameters:
    base_frame_id: base_footprint
    d_thresh: 0.02
    global_frame_id: map
    initial_pos_a: 0.0
    initial_pos_x: -2.0
    initial_pos_y: -0.5
    odom_frame_id: odom
    scan_topic: /scan
    transform_tolerance: -0.7
    use_sim_time: true
amcl:
  ros__parameters:
    alpha1: 0.2
    alpha2: 0.2
    alpha3: 0.2
    alpha4: 0.2
    alpha5: 0.2
    base_frame_id: base_footprint
    beam_skip_distance: 0.4
    beam_skip_error_threshold: -0.5000000000000001
    beam_skip_threshold: 0.3
    do_beamskip: false
    global_frame_id: map
    initial_pose:
      x: -2.0
      y: -0.5
      yaw: -0.0
      z: 0.0
    lambda_short: 0.1
    laser_likelihood_max_dist: 0.19999999999999996
    laser_max_range: 87.2
    laser_min_range: -1.4
    laser_model_type: likelihood_field
    max_beams: 60
    max_particles: 1999
    min_particles: 499
    odom_frame_id: odom
    pf_err: -0.05
    pf_z: 13.690000000000001
    recovery_alpha_fast: 0.0
    recovery_alpha_slow: 0.0
    resample_interval: 1
    robot_model_type: differential
    save_pose_rate: 0.5
    set_initial_pose: true
    sigma_hit: 0.2
    tf_broadcast: true
    transform_tolerance: 1.0
    update_min_a: 0.2
    update_min_d: 0.25
    use_sim_time: true
    z_hit: 0.5
    z_max: 3.25
    z_rand: 0.5
    z_short: 0.05
amcl_map_client:
  ros__parameters:
    use_sim_time: true
amcl_rclcpp_node:
  ros__parameters:
    use_sim_time: true
bt_navigator:
[...]

Issue #3231, opened 2022-10-06
Randomly Fuzzing Params Can Cause Server Crashes 


Cryst4L9527:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Foxy
- Version or commit hash:
  - a7955d891bcee5ad3f8acecc6ea3bdb1ddc73ce9
- DDS implementation:
  -  Fast-RTPS（default）

#### Steps to reproduce issue
launch the following command together in processes:
 
```
ros2 launch fuzz mygazebo.py 
ros2 run nav2_controller controller_server  --ros-args --params-fole configall.yaml -r __node:=controller_server
ros2 run nav2_planner planner_server  --ros-args --params-fole configall.yaml -r __node:=planner_server
ros2 run nav2_recoveries recoveries_server  --ros-args --params-fole configall.yaml -r __node:=recoveries_server
ros2 run nav2_bt_navigator bt_navigator  --ros-args --params-fole configall.yaml -r __node:=bt_navigator
ros2 run nav2_waypoint_follower waypoint_follower --ros-args --params-fole configall.yaml -r __node:=waypoint_follower
ros2 run nav2_lifecycle_manager lifecycle_manager  --ros-args --params-fole configall.yaml -r __node:=lifecycle_manager_navigation
ros2 run nav2_map_server map_server  --ros-args --params-fole configall.yaml -r __node:=map_server
ros2 run nav2_amcl amcl  --ros-args --params-fole configall.yaml -r __node:=amcl
ros2 run nav2_lifecycle_manager lifecycle_manager  --ros-args --params-fole configall.yaml -r __node:=lifecycle_manager_localization
```
and add another node to publish /goal_pose per 4s:
```
        self.goal_pub = self.node.create_publisher(geometry_msgs.msg.PoseStamped, '/goal_pose', 10)
        …
def _publish_goal(self, px=2.0, py=0.5, oz=1.0):
        msg = geometry_msgs.msg.PoseStamped()
        msg.header.frame_id = 'map'
        msg.pose.position.x = float(px)
        msg.pose.position.y = float(py)
        msg.pose.orientation.z = float(oz)
        print('[!] change_goal:'+str(px)+" "+str(py)+" "+str(oz)+" ")
        self.goal_pub.publish(msg)

4s: px = -2.0 py = 0.5 oz = 1.0
8s: px = 0.0 py = 0.5 oz = 1.0
[...]

PR #3242, opened 2022-10-12
map-size restriction to avoid overflow and nullptr caused by user-misconfiguration

Cryst4L9527:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3231 |

---

solve #3231, #3232, #3234, #3235.
With a negative width of map, 
getindex() function like my*size_x_+mx in many place will get a wrong negative index and cause overflow when visiting the array with this index.
getSizeInCellsX() will return a negative value , makes cached_distances_ has a size of zero, then throw a nullptr visiting problem when running distanceLookup().

#3233 is caused by  laser_likelihood_max_dist < 0, which seems to be solved by some commits before.
So now the issue #3231 can be closed.

---

Maybe more checks when visit the arrays and resize queues.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -661,7 +661,7 @@ Costmap2DROS::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameter
           resize_map = true;
           map_width_meters_ = parameter.as_int();
         } else {
-            RCLCPP_ERROR(
+          RCLCPP_ERROR(
             get_logger(), "You try to set width of map to be negative or zero,"
             " this isn't allowed, please give a positive value.");
         }
@@ -670,7 +670,7 @@ Costmap2DROS::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameter
           resize_map = true;
           map_height_meters_ = parameter.as_int();
         } else {
-            RCLCPP_ERROR(
+          RCLCPP_ERROR(
             get_logger(), "You try to set height of map to be negative or zero,"
             " this isn't allowed, please give a positive value.");
         }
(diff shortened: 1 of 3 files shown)


nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -659,9 +659,21 @@ Costmap2DROS::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameter
       if (name == "width") {
         resize_map = true;
         map_width_meters_ = parameter.as_int();
+        if (map_width_meters_ <= 0) {
```
SteveMacenski, line 662: I think these should be rejected -- e.g. not changed and the `result` is returned as false indicating the parameter requested change is invalid and rejected. You can elevate the log to an error message and remove setting default to 5 stuff since we'll just reject it.

Cryst4L9527: You are right! I've changed the logic as: if the value of the request is not sensible, just give an error message and do nothing else; if it's sensible ,run as before.

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -657,11 +657,23 @@ Costmap2DROS::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameter
...
       if (name == "width") {
-        resize_map = true;
-        map_width_meters_ = parameter.as_int();
+        if (parameter.as_int() > 0) {
+          resize_map = true;
+          map_width_meters_ = parameter.as_int();
+        } else {
+          RCLCPP_ERROR(
+            get_logger(), "You try to set width of map to be negative or zero,"
+            " this isn't allowed, please give a positive value.");
```
SteveMacenski, line 666: This should also set `result.successful = false;` and return to indicate the update failed

SteveMacenski (changes requested):

SteveMacenski (approved):
Thanks!

Issue #3196, opened 2022-09-15
Feature: Dynamically change polygons in collision monitor


tonynajjar:
Since the collision monitor is supposed to replace a safety scanner, I think it would make sense to emulate field set switching by subscribing to polygons.

Typically fieldsets would change according to speed/steering angle. one could write a separate node that subscribes to speed/steering angle and publishes which fieldset/polygon should be active.

Question here is: does it undermine the safety aspect of the collision monitor?

SteveMacenski:
> is supposed to replace a safety scanner

It is **not** supposed to replace a safety scanner, to be clear. It is a collision monitor which is not certified as a replacement for any safety equipment. 

I'm not entirely clear what you're asking for. The polygons used are published that can be subscribed to. Are you asking for the ability to dynamically change the polygons?

tonynajjar:
Gotcha, wrong wording, I did see the discussion about that on Discourse.

Anyway, yes I mean to dynamically change polygons

SteveMacenski:
Sure, yeah, I think that sounds reasonable. Your choices would be via dynamic parameters, topics, or services. What do you think there?

tonynajjar:
> Edit: or do you mean disabling/enabling several different pre-established polygons? Services there would make alot of sense.

Mmm, no I was actually thinking new unconfigured polygons for greater flexibility.

> Your choices would be via dynamic parameters, topics, or services

Yeah good question. I would think topics because of the potentially high frequency of change (e.g if the switching is dependent on speed and steering angle). As far as I know topics would offer the least latency.

vinnnyr:
+1 for topics. We too have a custom variant of this feature living elsewhere in the system, and like stated @tonynajjar  stated here:

> Typically fieldsets would change according to speed/steering angle.

we change the "zones" based on the current speed and current steering angle of our robot, maybe at a rate near 10hz. I don't know if it makes sense hitting a service at quite that rate.

SteveMacenski:
Topics seems reasonable to me :+1:

AlexeyMerzlyakov:
Good. The major thing here was agreed:
The current functionality of polygons in the Collision Monitor to be enhanced by subscribing to the topics:
the similar way as `APPROACH` model utilizes. Current functionality with strictly-defined polygon will be intact.
This could be reached e.g. by adding a new `polygon_topic` parameter. If this parameter is not set, using `points` array with polygon vertexes. Otherwise - subscribing to the polygon topic.

[...]

PR #3245, opened 2022-10-13
Dynamically changing polygons support

AlexeyMerzlyakov:
Adding an ability to dynamically read polygon vertices from input topic for "STOP" and "SLOWDOWN" behavior models.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3196 |
| Primary OS tested on | Ubuntu 20.04 with ROS2 Rolling onboard |
| Robotic platform tested on | TB3 simulation, unit & codecov testing |

---

* New ROS-parameter `polygon_sub_topic`. If it is set, subscribe to the polygon from input `PolygonStamped` messages. If not - use classical static user-defined polygon vertices.
* Allowing to use polygon shape published in a different from `robot_base` frame by reading `PolygonStamped->header.frame_id`.
* The change applied to "STOP" and "SLOWDOWN" behavior models, since for "APPROACH" one we have already utilized footprint subscriber.

After the change will be reviewed and agreed, new `polygon_sub_topic` and its related logic will be added to the documentation pages.

---

* Update official Nav2 docs

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/collision_monitor_node.cpp
+++ b/nav2_collision_monitor/src/collision_monitor_node.cpp
@@ -389,6 +389,10 @@ bool CollisionMonitor::processStopSlowdown(
   const Velocity & velocity,
   Action & robot_action) const
 {
+  if (!polygon->isShapeSet()) {
+    return false;
+  }
+
   if (polygon->getPointsInside(collision_points) > polygon->getMaxPoints()) {
     if (polygon->getActionType() == STOP) {
       // Setting up zero velocity for STOP model
@@ -419,6 +423,10 @@ bool CollisionMonitor::processApproach(
   Action & robot_action) const
 {
   polygon->updatePolygon();
+
+  if (!polygon->isShapeSet()) {
+    return false;
+  }
 
   // Obtain time before a collision
   const double collision_time = polygon->getCollisionTime(collision_points, velocity);
--- a/nav2_collision_monitor/src/polygon.cpp
+++ b/nav2_collision_monitor/src/polygon.cpp
@@ -20,10 +20,8 @@
 #include "geometry_msgs/msg/point.hpp"
 #include "geometry_msgs/msg/point32.hpp"
 
-#include "tf2/convert.h"
-#include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
-
 #include "nav2_util/node_utils.hpp"
+#include "nav2_util/robot_utils.hpp"
 
 #include "nav2_collision_monitor/kinematics.hpp"
 
@@ -379,36 +377,6 @@ bool Polygon::getParameters(
   return true;
 }
 
-bool Polygon::getTransform(
-  const std::string & source_frame_id,
-  tf2::Transform & tf2_transform) const
-{
-  geometry_msgs::msg::TransformStamped transform;
-  tf2_transform.setIdentity();  // initialize by identical transform
-
-  if (source_frame_id == base_frame_id_) {
-    // We are already in required frame
-    return true;
-  }
-
-  try {
-    // Obtaining the transform to get data from source to base frame
-    transform = tf_buffer_->lookupTransform(
-      base_frame_id_, source_frame_id,
-      tf2::TimePointZero, transform_tolerance_);
-  } catch (tf2::TransformException & e) {
-    RCLCPP_ERROR(
-      logger_,
-      "[%s]: Failed to get \"%s\"->\"%s\" frame transform: %s",
-      polygon_name_.c_str(), source_frame_id.c_str(), base_frame_id_.c_str(), e.what());
-    return false;
-  }
-
-  // Convert TransformStamped to TF2 transform
-  tf2::fromMsg(transform.transform, tf2_transform);
-  return true;
-}
-
 void Polygon::updatePolygon(geometry_msgs::msg::PolygonStamped::ConstSharedPtr msg)
 {
   std::size_t new_size = msg->polygon.points.size();
@@ -423,7 +391,11 @@ void Polygon::updatePolygon(geometry_msgs::msg::PolygonStamped::ConstSharedPtr m
 
   // Get the transform from PolygonStamped frame to base_frame_id_
   tf2::Transform tf_transform;
-  if (!getTransform(msg->header.frame_id, tf_transform)) {
+  if (
+    !nav2_util::getTransform(
+      msg->header.frame_id, base_frame_id_,
+      transform_tolerance_, tf_buffer_, tf_transform))
+  {
     return;
   }
 
--- a/nav2_collision_monitor/src/polygon.cpp
+++ b/nav2_collision_monitor/src/polygon.cpp
@@ -86,7 +86,8 @@ bool Polygon::configure()
   }
 
   if (visualize_) {
-    // Fill polygon_ points for future usage
+    // Fill polygon_ for future usage
+    polygon_.header.frame_id = base_frame_id_;
     std::vector<Point> poly;
     getPolygon(poly);
     for (const Point & p : poly) {
@@ -94,7 +95,7 @@ bool Polygon::configure()
       p_s.x = p.x;
       p_s.y = p.y;
       // p_s.z will remain 0.0
-      polygon_.points.push_back(p_s);
+      polygon_.polygon.points.push_back(p_s);
     }
 
     rclcpp::QoS polygon_qos = rclcpp::SystemDefaultsQoS();  // set to default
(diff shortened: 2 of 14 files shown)


nav2_collision_monitor/src/polygon.cpp
```diff
@@ -300,14 +319,29 @@ bool Polygon::getParameters(std::string & polygon_pub_topic, std::string & footp
...
+      // This is robot footprint: do not need to get polygon points from ROS parameters or topic.
       // It will be set dynamically later.
+      polygon_sub_topic.clear();
       return true;
     } else {
       // Make it empty otherwise
       footprint_topic.clear();
     }
 
+    try {
```
SteveMacenski, line 331: There should be an explicit policy that if the polygon points parameter has a polygon in it- we use that and not dynamic from a topic. This seems like it will do both - if the topic exists, it'll set it but also still grab the parameters from the yaml just below. This should be mutually exclusive. So then the polygon class here needs to wait on the first polygon to come in from the topic to be able to meaningfully process data over its region

AlexeyMerzlyakov: Yep, these parameters already made to be mutually exclusive. The try-catch block handling `polygon_sub_topic` parameter already has a return if `polygon_sub_topic` parameter was set:
```
    try {
      nav2_util::declare_parameter_if_not_declared(".polygon_sub_topic");
      polygon_sub_topic = node->get_parameter(".polygon_sub_topic");
      // Do not need to proceed further, if polygon_sub_topic parameter is defined.
      // Dynamic polygon will be used.
      return true; <- "points" parameter to be ignored in this case.
    } catch (const rclcpp::exceptions::ParameterUninitializedException & ex) {
      // polygon_sub_topic parameter is not defined: initializing polygon with points.
      // Static polygon will be used.
      polygon_sub_topic.clear();
    }

    Handling "points" parameter otherwise -> ...
```


So, we have the following logic:
- If `polygon_sub_topic` was set, polygon subscription will be made and `points` parameter to be ignored.
- If `polygon_sub_topic` was not set, we are reading `points` parameter, which will be mandatory in this case.
- If none (`polygon_sub_topic` and `points`) parameters were not set, getting an error.

The same logic we have for footprint subscriber and its `footprint_topic` parameter for "APPROACH" case, so I doubt, that we should flip the priority here.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -345,6 +379,80 @@ bool Polygon::getParameters(std::string & polygon_pub_topic, std::string & footp
   return true;
 }
 
+bool Polygon::getTransform(
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: The similar function was used in data sources `Source::getTransform()`, but it is interpolating frames using time shifts between them. [Moved](https://github.com/ros-planning/navigation2/pull/3245/commits/b94bcc48e3e24d697cf04903c4d5c1e46ad29236) both functions to `nav2_util/robot_utils.hpp`, since it already has similar transformation routine.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -203,15 +225,9 @@ void Polygon::publish() const
...
-    std::make_unique<geometry_msgs::msg::PolygonStamped>();
-  poly_s->header.stamp = node->now();
-  poly_s->header.frame_id = base_frame_id_;
-  poly_s->polygon = polygon_;
-
-  // Publish polygon
-  polygon_pub_->publish(std::move(poly_s));
+  // Actualize the time to current and publish the polygon
+  polygon_.header.stamp = node->now();
+  polygon_pub_->publish(polygon_);
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: Yes, de-facto standard of publishing in Nav2 - it is a transferring resources to published message to another object.
However, the `polygon_` structure in our case is being used for two purposes: 1. storing polygon vertices and polygon frame data for publishing and 2. messages publishing. The storing some data into polygon (1) do much less frequently than publishing (2) polygon with updated time.
Since we need some structure for polygon points and frame storing (be it `PolygonStamped polygon_` or something else), we can not transfer the ownership to another object to it, since the stored object will be re-used in next cycles. Therefore, the way to `publish(std::move(unique_object))` - is to make a new `unique_object` and copy all the data into it. Something like:
```
auto unique_object = std::make_unique<PolygonStamped>(polygon_);
polygon_pub_->publish(std::move(unique_object));
```
However, what is the difference between the construction from above and just publishing the `polygon_` directly?
```
polygon_pub_->publish(polygon_);
```

During the publishing without std::move, ROS2 will allocate the memory for new object and make the copy of `polygon_` into it. The same as in the first example. If so, is there any performance reason to make these changes?

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: Got it - the `return` line in L338 I missed, that all makes sense. I think though the opposite policy is most reasonable for default behavior (do you agree?) where if `points` are set, use them. Else, check if there's a topic. If not, error not. My thinking is go from most specific to most general. If something specific is set (points), always use that. 

Also what protections exist to check if polygon is populated from a topic when an iteration is run? Do we have a check if `poly_.size()` is empty and log a warning that it hasn't been set yet by the topic?

nav2_collision_monitor/src/polygon.cpp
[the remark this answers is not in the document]
SteveMacenski: I forget the exact rationale, but when publishing pointclouds in that fashion in developing the Ouster Drivers, Tom from Box Robotics pointed out that by giving it a smart pointer and moving its ownership to the publisher, the jitter the publication rate possible was significantly was reduced. But the idea is we're creating an object containing the data as a smart pointer and then handing it over without storing it for future iterations. I want to also say that you have to give it a smart pointer for IPC optimizations to work properly - which we don't use right now due to restrictions on QoS settings possible, but may in the future.

nav2_collision_monitor/src/polygon.cpp
AlexeyMerzlyakov: Okay, I would like to agree and will change the priority of handling for all models in polygons:
- If points set, set polygon shape by them. If them were set incorrectly - fail node. If not set, move to next check
- Subscribe for polygon (or footprint for `APPROACH` model) topic. When receiving, check that message is correct (>= 2 points in there).
- While polygon/footprint was not received - warn a message back to user

> Also what protections exist to check if polygon is populated from a topic when an iteration is run? Do we have a check if poly_.size() is empty and log a warning that it hasn't been set yet by the topic?

In current model not. I will also add it in next commit.

SteveMacenski: Thanks, sounds great!

SteveMacenski:
@AlexeyMerzlyakov lets try to get this done this week please!

nav2_collision_monitor/src/polygon.cpp
AlexeyMerzlyakov: Switched to new logic and added warnings in https://github.com/ros-planning/navigation2/pull/3245/commits/a6939d94c7fbf2f18e0b4be33ea42525e12fd0b6

nav2_collision_monitor/src/polygon.cpp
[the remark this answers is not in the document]
AlexeyMerzlyakov: Okay, added new object and ownership movement from it in the latest commit:
https://github.com/ros-planning/navigation2/pull/3245/files#diff-e7fd1328f82b3191cac8b30062c6eee0ea702a0be381dde8572120fa21a87595R237-R238

AlexeyMerzlyakov:
Sure, I've pushed the latest update meeting remaining review items. Please refer to it (unroftunately CircleCI is not able to do verification, but I've done local codecov verification with `98.1%` CM code line coverage during `colcon test`

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -420,6 +424,10 @@ bool CollisionMonitor::processApproach(
 {
   polygon->updatePolygon();
 
+  if (!polygon->isShapeSet()) {
```
SteveMacenski, line 427: It would seems sensible to me to have this before updating since we shouldn't need to update if its not set.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -129,6 +148,15 @@ void Polygon::getPolygon(std::vector<Point> & poly) const
   poly = poly_;
 }
 
+bool Polygon::isShapeSet()
+{
+  if (poly_.empty()) {
+    RCLCPP_WARN(logger_, "[%s]: Polygon shape is not set yet", polygon_name_.c_str());
```
SteveMacenski, line 154: Make this a throttled warning, since this would blast the entire terminal at a high rate until set

SteveMacenski:
Otherwise LGTM

SteveMacenski:
Also docs updates for the new parameters

nav2_collision_monitor/src/collision_monitor_node.cpp
AlexeyMerzlyakov: The main thing here is that `updatePolygon()` optionally updating polygon by request by calling `getFootprintInRobotFrame()` FootprintSubscriber API. However, this might not update polygon, e.g. if FootprintSubscriber has not received footprint yet, or if we are using `Circle` object instead of `Polygon`.
For cases when `updatePolygon()` will set polygon shape, we need to check that shape is set after this call, in order to not miss the cycle when we already have footprint published, and thus it could be handled.
So, this check should be after `updatePolygon()` call for approach model.

SteveMacenski (approved):
Just need the docs updates and this is good to go 

Alt text

mbed92:
Does that functionality work on Humble? I cannot get it to work - `polygon_topic` or `polygon_sub_topic` doesn't exist.

PR #3248, opened 2022-10-17
exceptions for compute path through poses

jwallace42:
---

## Basic Info

| ------ | -----  |

---

Added error codes for compute path through poses. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_msgs/action/ComputePathThroughPoses.action
+++ b/nav2_msgs/action/ComputePathThroughPoses.action
@@ -1,3 +1,16 @@
+# Error codes
+# Note: The expected priority order of the errors should match the message order
+int16 NONE=0
+int16 UNKNOWN=1
+int16 TF_ERROR=2
+int16 START_OUTSIDE_MAP=3
+int16 GOAL_OUTSIDE_MAP=4
+int16 START_OCCUPIED=5
+int16 GOAL_OCCUPIED=6
+int16 TIMEOUT=7
+int16 NO_VALID_PATH=8
+int16 NO_WAYPOINTS_GIVEN=9
+
 #goal definition
 geometry_msgs/PoseStamped[] goals
 geometry_msgs/PoseStamped start
@@ -7,5 +20,6 @@ bool use_start # If false, use current robot pose as path start, if true, use st
 #result definition
 nav_msgs/Path path
 builtin_interfaces/Duration planning_time
+int16 error_code
 ---
 #feedback definition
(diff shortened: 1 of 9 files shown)


nav2_planner/src/planner_server.cpp
```diff
@@ -336,11 +336,13 @@ void PlannerServer::computePlanThroughPoses()
 
   auto start_time = steady_clock_.now();
 
-  // Initialize the ComputePathToPose goal and result
+  // Initialize the ComputePathThroughPoses goal and result
   auto goal = action_server_poses_->get_current_goal();
   auto result = std::make_shared<ActionThroughPoses::Result>();
   nav_msgs::msg::Path concat_path;
 
+  geometry_msgs::msg::PoseStamped curr_start, curr_goal;
```
SteveMacenski, line 344: Move back to original location near where its first use is

nav2_planner/src/planner_server.cpp
```diff
@@ -350,11 +352,12 @@ void PlannerServer::computePlanThroughPoses()
 
     getPreemptedGoalIfRequested(action_server_poses_, goal);
 
-    if (goal->goals.size() == 0) {
+    if (goal->goals.empty()) {
       RCLCPP_WARN(
         get_logger(),
-        "Compute path through poses requested a plan with no viapoint poses, returning.");
+        "Compute path through poses requested a plan with no waypoint poses, returning.");
```
SteveMacenski, line 358: Would this be what throws NoWaypointsGiven?

nav2_msgs/action/ComputePathThroughPoses.action
```diff
@@ -1,3 +1,16 @@
...
+int16 NONE=0
+int16 UNKNOWN=1
+int16 TF_ERROR=2
+int16 START_OUTSIDE_MAP=3
+int16 GOAL_OUTSIDE_MAP=4
+int16 START_OCCUPIED=5
+int16 GOAL_OCCUPIED=6
+int16 TIMEOUT=7
+int16 NO_VALID_PATH=8
+int16 NO_WAYPOINTS_GIVEN=9
```
SteveMacenski, line 12: I prefer the via point language when talking about points going through states continuously. I think "waypoints" is a loaded term that people usually think of as things to stop at. Given we have the waypoint follower that stops at poses, I think having another name is helpful to separate these ideas

nav2_planner/src/planner_server.cpp
SteveMacenski: See viapoint comment above too

SteveMacenski (changes requested):

nav2_planner/src/planner_server.cpp
jwallace42: The curr_start and goal need to be outside of the try catch in order to print out the offending start/goal if an exception is thrown.

SteveMacenski (changes requested):

SteveMacenski (approved):
LGTM, anything else before merge? I retriggered CI, doesn't seem like it went the last time

jwallace42:
Nope! I think this is good to go.

PR #3251, opened 2022-10-20
Pipe error codes

### LINKED ISSUE #3043: Feature Request: Feedback for Navigation Failures

jwallace42:
---

## Basic Info

| ------ | ----- |

---

This pr pipes error codes through the behavior tree to the `compute path to pose` and `compute path through poses` to help close #3043 

---

#### For Maintainers: 


jwallace42:
blocked by https://github.com/ros-planning/navigation2/pull/3248

nav2_bt_navigator/src/navigators/navigate_to_pose.cpp
```diff
@@ -212,6 +211,8 @@ NavigateToPoseNavigator::initializeGoalPose(ActionT::Goal::ConstSharedPtr goal)
   start_time_ = clock_->now();
   auto blackboard = bt_action_server_->getBlackboard();
   blackboard->set<int>("number_recoveries", 0);  // NOLINT
+  blackboard->set<int>("follow_path_error_code", 0);
```
SteveMacenski, line 214: Set based on message NONE types. I know they're `0` but that removes a magic number from the code + if it changes in the future, we don't have a subtle breakage

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +251,28 @@ void BtActionServer<ActionT>::executeCallback()
...
 
+template<class ActionT>
+void BtActionServer<ActionT>::populateErrorCode(typename std::shared_ptr<typename
+    ActionT::Result> result)
+{
+    // Grab error codes from blackboard
+    std::vector<int> error_codes;
+
+    // 0-99 error codes for follow path
+    error_codes.push_back(blackboard_->get<int>("follow_path_error_code"));
```
SteveMacenski, line 262: You need to check if that's even possible to get it like you do below. This may not be set (in which case this throws an exception)

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +251,28 @@ void BtActionServer<ActionT>::executeCallback()
...
+
+  // 100-199 error codes for compute path to pose or compute path through poses
+  int compute_path_error_code;
+  if (blackboard_->get<int>("compute_path_to_pose_error_code", compute_path_error_code)){
```
SteveMacenski, line 266: Its not necessarily true that just because this doesn't exist, the other does. There could be a misconfiguration.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -226,6 +228,9 @@ void BtActionServer<ActionT>::executeCallback()
   // Give server an opportunity to populate the result message or simple give
   // an indication that the action is complete.
   auto result = std::make_shared<typename ActionT::Result>();
+
+  populateErrorCode(result);
```
SteveMacenski, line 232: Wouldn't this only be required if the goal failed (e.g. add to L244)?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +251,28 @@ void BtActionServer<ActionT>::executeCallback()
   }
 }
 
+template<class ActionT>
+void BtActionServer<ActionT>::populateErrorCode(typename std::shared_ptr<typename
```
SteveMacenski, line 255: This feels pretty hacky and doesn't exactly scale when we start talking about smoother server and other 3rd party server failure codes. 

Some other options
- Add in a flag for which sever this belongs to in the message definition with an enumeration like we did the error messages in the others. That would be extensible so that more servers could be added by the user in closed source and just need to push their ID next (or more to leave room to grow). 
- We change the message enumerations so that they themselves are separated by 100s (or 1000's why not, we have the room in 16 bit) so its immediately clear when trying to look at the error code match where it came from.  That would be extensible so that more servers could be added by the users in closed source and just need to tick all their error codes up another set of 1,000 (or more for room to grow in the main project). Basically, what you did here but in the message definitions themselves so there are no conflicts. Would we want also to have an enum here for the ranges to help translate? I'm not sure. 
- Bitshift stuff around depending on the type of server so a single 16 bit int can represent it, but can separate the server it came from by location 
- (probably not good) Duplicate all of the error code IDs in the NavToPose/similar messages with a prefix on server name and convert here

SteveMacenski: @ayrton04 thoughts? These are all kind of off the cuff.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: For example as I mention in a comment below if the BT port name doesn't match the param name

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: Also use the parameterized name

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: Also use the parameterized names

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -42,15 +42,15 @@ BT::NodeStatus ComputePathToPoseAction::on_success()
...
+  config().blackboard->set<int>("compute_path_to_pose_error_code", result_.result->error_code);
   return BT::NodeStatus::SUCCESS;
 }
 
 BT::NodeStatus ComputePathToPoseAction::on_aborted()
 {
   nav_msgs::msg::Path empty_path;
   setOutput("path", empty_path);
-  setOutput("compute_path_to_pose_error_code", result_.result->error_code);
+  config().blackboard->set<int>("compute_path_to_pose_error_code", result_.result->error_code);
```
SteveMacenski, line 53: The output port was the right place, not the blackboard. The output port can be set to a blackboard value using `{my_var_name}` (so `{compute_path_to_pose_error_code}`), but then allows someone to remap this name if they like, since its conceivable that there are multiple instances of these action nodes in a BT so you might want to differentiate them. 

So the use of this BT node would be `<ComputePathToPose error_code_id={compute_path_to_pose_error_code}/>`

Note that I simplified `error_code`, I think it would actually be wise to set the output port for all of these actions with failure codes to a generic name like that for brevity of the BT. As a result, then all of the actual blackboard key names must be variables/parameters (e.g. `compute_path_to_pose_error_code`). 

For example, for the other "special" output/input values of the BT, we have parameters for the node that have to match the BT in use (https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/src/navigators/navigate_to_pose.cpp#L33-L43). That way you can change the goal ID name in the BT `{goal}` to `{my_goal}` for whatever reason and then can tell BT navigator what key it should look for for the goal without naively assuming its hardcoded to `goal`. 

So 
- There needs to be a parameterization of BT navigator for the error code blackboard names to use to look for error codes, and that used to set it like you do in the file below this one 
- The use of the error code must be set in the BT at the output port so that they match up, for all the BTs
- Rename the `setOutput` / `outputPort` (after adding back in) to `error_code_id` for shortening, since that's not the actual name of the variable being used, we can create a common name foundation for this in all the BT nodes

The question kind of remains though in the BT navigator how to handle if we have multiple key names for a compute path action to get them for populating the final message. I suppose we could make the set of follow / compute path / etc error code parameter field vectors? 

---

Or actually, maybe we can make a global const name for the names of these error code port names so they can still be modified but will all be the same from the BT navigator + BT nodes. `const std::string path_error_code_name = "my_code_name"`, and use that variable everywhere. It wouldn't be run-time configurable, only compile-time.

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
SteveMacenski: So in line with the other comment, there would be no need to specify which error code IDs belong to which server, we could just take in the vector of `error_code_ids` and populate them to `0` (which we should def to a variable of name `NONE`) instead. I actually don't see anywhere in the code of BT navigator where we specifically need to know the name of a given servers error code id. Just as long as they're all in `error_code_ids` and specified in the BT output ports have matching sets.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: But see commentary below - this changes a bit

SteveMacenski:
Bc new task servers need to be offset by 100,000 (arbitrarily large number so it never conflicts with new Nav2 servers), we should make sure to document that clearly. 

Maybe we can update https://navigation.ros.org/setup_guides/lifecycle_composition/setup_lifecycle_composition.html to be targeting "Add New Nav2 Task Server" and have a subsection on the error code stuff (if its required, add to msg with offset + explain why in increments of 1000 and priority, add to BT navigator `error_code_ids` param to query). I think actually with that, that would complete that page to literally just _be_ the tl;dr on adding a new task server. Though actually maybe add a subsection on adding a behavior tree node to interact with it. Else, this handles bringup (lifecycle) composition, and error handling.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
jwallace42: The result goes into this `on_completion_callback_(result, rc);`. I assumed that we should populate the error before this function call.

SteveMacenski:
Ping me back when you want me to take a look again!

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: Good idea

nav2_msgs/action/ComputePathToPose.action
```diff
@@ -1,14 +1,14 @@
...
-int16 NONE=0
-int16 UNKNOWN=1
-int16 TF_ERROR=2
-int16 START_OUTSIDE_MAP=3
-int16 GOAL_OUTSIDE_MAP=4
-int16 START_OCCUPIED=5
-int16 GOAL_OCCUPIED=6
-int16 TIMEOUT=7
-int16 NO_VALID_PATH=8
+int16 NONE=1000
```
SteveMacenski, line 3: Lets give ourselves room to grow: start the controller server at 1000 and then offset the rest accordingly. For all we know, something serious could be added between the controller and the base in the future (right now we actually already have a couple: velocity smoother + collision monitor - though neither are in the nav2 bt framework)

nav2_msgs/action/NavigateThroughPoses.action
```diff
@@ -4,6 +4,7 @@ string behavior_tree
 ---
 #result definition
 std_msgs/Empty result
+int16 error_code
```
SteveMacenski, line 7: waypoint follower too? Also, you need to extract that from the WPF result if there's a failure

SteveMacenski: (basically just taking the result from NavToPose and passing it through to the WPF client, its a 1-liner)

SteveMacenski (changes requested):
Very good and involves surprisingly few changes beyond changes required for the BTs!

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +261,29 @@ void BtActionServer<ActionT>::executeCallback()
...
+    int current_error_code = 0;
+    try {
+      current_error_code = blackboard_->get<int>(error_code_id_name);
+    } catch (...) {
+      RCLCPP_ERROR(
+        logger_,
+        "Failed to get %s from blackboard",
+        error_code_id_name.c_str());
+    }
+    if (current_error_code > highest_priority_error_code) {
```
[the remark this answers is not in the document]
SteveMacenski: Can you update nav2_system_test test(s) that are testing failure cases to check that this error code is non-NONE?  If we know exactly which error code should trigger for a failure, check that specific code.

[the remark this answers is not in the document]
jwallace42: I am not quite sure which tests you are referring to?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -59,6 +59,15 @@ BtActionServer<ActionT>::BtActionServer(
...
     node->declare_parameter("default_server_timeout", 20);
   }
+
+  std::vector<std::string> error_code_id_names = {
+    "follow_path_error_code_id",
+    "compute_path_error_code_id"
+  };
+
+  if (!node->has_parameter("error_code_id_names")) {
+    node->declare_parameter("error_code_id_names", error_code_id_names);
```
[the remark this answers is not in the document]
jwallace42: I want to do something like this.

`#include<iostream>
#include<vector>
template <typename T>
std::ostream & operator << (std::ostream & os, const std::vector<T> & vec)
{
    for(auto elem : vec)
    {
        os<<elem<< " ";
    }
    return os;
}`

I am not really sure where it would go... Could I add it to the nav2_utils?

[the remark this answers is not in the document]
SteveMacenski: Sure thing

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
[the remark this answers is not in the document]
SteveMacenski: In nav2_system_tests, there are tests for WPF / Nav2Pose / NavThroughPoses / etc. Some of those tests specifically are designed to fail (e.g. giving it bogus goals or starting off the map, etc) so after this PR, they should have error codes that you can read in the test and make sure not only did they fail in the action response, but that the error code is populated and correct.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +268,34 @@ void BtActionServer<ActionT>::executeCallback()
...
+        logger_,
+        "Failed to get %s from blackboard",
+        error_code_id_name.c_str());
+    }
+  }
+
+  if (highest_priority_error_code != std::numeric_limits<int>::max()) {
+    result->error_code = highest_priority_error_code;
+  } else {
+    result->error_code = 0;
```
SteveMacenski, line 295: This should be initialized to `0` as an int type

nav2_msgs/action/ComputePathThroughPoses.action
```diff
@@ -1,15 +1,15 @@
...
-int16 UNKNOWN=1
-int16 TF_ERROR=2
-int16 START_OUTSIDE_MAP=3
-int16 GOAL_OUTSIDE_MAP=4
-int16 START_OCCUPIED=5
-int16 GOAL_OCCUPIED=6
-int16 TIMEOUT=7
-int16 NO_VALID_PATH=8
-int16 NO_VIAPOINTS_GIVEN=9
+int16 NONE=2000
```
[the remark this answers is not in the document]
SteveMacenski: Still needs to be done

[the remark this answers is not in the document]
SteveMacenski: Still needs to be done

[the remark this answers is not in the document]
jwallace42: No sure why git is showing that. It disagrees with the actual change I have [here](https://github.com/jwallace42/navigation2/blob/pipe_error_codes/nav2_msgs/action/ComputePathThroughPoses.action).

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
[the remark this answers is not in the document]
jwallace42: So, 

inside the nav2 system tests the way point follower has one test that causes global planning to fail.  

In the _planning_ dir random path are being generated so there isn't any error codes we can check. 

I don't really see anywhere else in nav2_system_test that actually fail the planner and controller.

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -268,8 +270,14 @@ WaypointFollower::followWaypoints()
           get_logger(), "Completed all %zu waypoints requested.",
           goal->poses.size());
         result->missed_waypoints = failed_ids_;
-        action_server_->succeeded_current(result);
+        if (error_code_ > 0) {
```
SteveMacenski, line 273: Woof, I forgot about this. WPF can have multiple failed requests. I think similar to how we handle `missed_waypoints` we need to handle the error codes

jwallace42: Do we want the error codes for each of the failures or only when the action fails? 

I assumed that the old error codes do not provide any useful information to higher level autonomy systems and thus we only should return the last error. However, switching it should be quite simple.

jwallace42:
Not really sure why this is failing. I ran it about 100 times locally with no issues :(.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -21,10 +21,12 @@
 #include <set>
 #include <exception>
 #include <vector>
+#include <limits>
 
 #include "nav2_msgs/action/navigate_to_pose.hpp"
 #include "nav2_behavior_tree/bt_action_server.hpp"
 #include "ament_index_cpp/get_package_share_directory.hpp"
+//#include "nav2_util/print_utils.hpp"
```
SteveMacenski, line 29: commented out?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +272,32 @@ void BtActionServer<ActionT>::executeCallback()
...
 
+template<class ActionT>
+void BtActionServer<ActionT>::populateErrorCode(
+  typename std::shared_ptr<typename ActionT::Result> result)
+{
+  int highest_priority_error_code = std::numeric_limits<int>::max();
+  for (const auto & error_code_id_name : error_code_ids_) {
+    int current_error_code = std::numeric_limits<int>::max();
+    try {
+      current_error_code = blackboard_->get<int>(error_code_id_name);
```
SteveMacenski, line 283: You can do `blackboard_->get<int>("name", var_to_store)` and it return a bool, so you don't need to catch and you don't need to initialize the var_to_store as numeric limits since you only need to check `if (blackboard->...)` to enter the section of code to check the priority level

SteveMacenski: Exceptions are expensive

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: That's a really great question. I think it would be helpful context to a user to know not only why the last one failed, but for any that failed, why that occurred. It could be useful context for analysis of failure modes.

SteveMacenski: It might actually be sensible then to make the Missed WP stuff a new struct to store in the WPF to keep that data associated in case we want to add more fields to that in the future `struct MissedWaypoints {int index; int error_code;}`.

Perhaps also updating the message as well for that kind of organization? If we did that, it might be sensible to also add the index's actual goal pose too.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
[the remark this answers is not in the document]
SteveMacenski: I don't understand, the WPF test uses the full stack and has a failure mode of an off-map request, that should be pretty clear-cut. There's also the system_failure tests too to choose from. For the system-level, it doesn't need to be both planner/controller (later unit tests for the planner/controller server that we've discussed will do that), it needs to just be a test to evaluate how once any of the servers fail, its being properly communicated back to the BT navigator or waypoint follower or what have you. The WPF is, to me, the best choice, because that will exercise not only the BT Nav stuff, but also the WPF stuff simultaneously 

We need some kind of coverage for this feature and those seem like really easy / obvious places for it.

nav2_waypoint_follower/src/waypoint_follower.cpp
jwallace42: Could we create a new message type `MissedWaypoint.msg` and have the `FollowWaypoint.action` return a array of `MissedWaypoint.msg`?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +271,32 @@ void BtActionServer<ActionT>::executeCallback()
...
+template<class ActionT>
+void BtActionServer<ActionT>::populateErrorCode(
+  typename std::shared_ptr<typename ActionT::Result> result)
+{
+  int highest_priority_error_code = std::numeric_limits<int>::max();
+  for (const auto & error_code_id_name : error_code_ids_) {
+    int current_error_code;
+    bool found_error_code = blackboard_->get<int>(error_code_id_name, current_error_code);
+
+    if (found_error_code) {
```
SteveMacenski, line 283:
```suggestion
    if (blackboard_->get<int>(error_code_id_name, current_error_code)) {
```

Then you can delete L281

nav2_msgs/action/FollowWaypoints.action
```diff
@@ -3,6 +3,7 @@ geometry_msgs/PoseStamped[] poses
 ---
 #result definition
 int32[] missed_waypoints
+int16[] error_code
```
SteveMacenski, line 6:
```suggestion
int16[] error_codes
```

nav2_waypoint_follower/include/nav2_waypoint_follower/waypoint_follower.hpp
```diff
@@ -133,10 +139,13 @@ class WaypointFollower : public nav2_util::LifecycleNode
   rclcpp::CallbackGroup::SharedPtr callback_group_;
   rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
   std::shared_future<rclcpp_action::ClientGoalHandle<ClientT>::SharedPtr> future_goal_handle_;
+  int error_code_;
```
SteveMacenski, line 142: remove?

SteveMacenski: Oh this is used to store the current one, maybe `current_error_code_`?

SteveMacenski: Better yet, a struct containing the values of `current_goal_status_` and `error_code_` since they're both populated in the `resultCallback`, which itself would be named `current_goal_status_` with the int error code and the ActionStatus enum.

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -237,16 +243,24 @@ WaypointFollower::followWaypoints()
       RCLCPP_INFO(
         get_logger(), "Task execution at waypoint %i %s", goal_index,
         is_task_executed ? "succeeded" : "failed!");
+      if (!is_task_executed) {
```
SteveMacenski, line 246: Remove this - there would be no error code because this is for the waypoint executor plugin, not from the constituent navigation server failures. We haven't defined a set of exceptions/error codes for the waypoint task executors. Arguably, we cannot since its so broad. 

Unless we added an error code for WPF for failed task executor just generically. That would be smart.

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: Yup!

SteveMacenski:
> Not really sure why this is failing. I ran it about 100 times locally with no issues :(.

Not sure, that change seemed legit to me. How about printing the value in CI that it seems to be receiving that isn't the value of GOAL OUT OF MAP you expect? That may be telling.

nav2_waypoint_follower/src/waypoint_follower.cpp
jwallace42: Sounds like a future project :).

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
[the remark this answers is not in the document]
jwallace42: Okay, 

I wasn't quite sure _what_ we wanted to test. This could be added to the error_code test and would be quicker (don't have to run through recoveries on failure by running with a similar BT). Once we update the main behavior trees to handle the errors this should be a lot faster and we could update the follow waypoint test.

nav2_system_tests/src/error_codes/error_code_param.yaml
```diff
@@ -0,0 +1,356 @@
+amcl:
```
SteveMacenski, line 1: Why does this file have anything other than planner/controller server params in it?

nav2_system_tests/src/error_codes/test_error_codes_launch.py
```diff
@@ -0,0 +1,59 @@
...
+from ament_index_python.packages import get_package_share_directory
+
+from launch import LaunchDescription
+from launch import LaunchService
+from launch.actions import ExecuteProcess, IncludeLaunchDescription
+from launch_testing.legacy import LaunchTestService
+from launch.launch_description_sources import PythonLaunchDescriptionSource
+
+
+def generate_launch_description():
```
SteveMacenski, line 28: Why does this not only launch the planner/controller servers?

nav2_system_tests/CMakeLists.txt
```diff
@@ -47,6 +47,40 @@ set(dependencies
   behaviortree_cpp_v3
 )
 
+set(local_controller_plugin_lib local_controller)
```
SteveMacenski, line 50: This can be placed into the CMakeLists file in the src/error_codes directory that we can include `add_subdirectory(src/error_codes)`

nav2_system_tests/src/waypoint_follower/tester.py
```diff
@@ -208,7 +208,7 @@ def main(argv=sys.argv[1:]):
     time.sleep(2)
     test.cancel_goal()
 
-    # a failure case
+    # Set a waypoint outside the bounds of the map
```
SteveMacenski, line 211: You need to get the error code and make sure its correct for the type of error this represents

nav2_msgs/action/ComputePathToPose.action
```diff
@@ -21,4 +21,3 @@ nav_msgs/Path path
 builtin_interfaces/Duration planning_time
 int16 error_code
 ---
-#feedback definition
```
SteveMacenski, line 24: Readd

SteveMacenski (changes requested):
These + you still have a number of open comments from previous reviews

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: Its just adding an enum to the waypoint follower for the next server up (6000's?) for `TASK_EXECUTOR_FAILED`

nav2_system_tests/src/error_codes/error_code_param.yaml
jwallace42: It uses the nav2_simple_commander.

nav2_system_tests/CMakeLists.txt
jwallace42: I tried this, and the servers were unable to find the plugins.

SteveMacenski: Was this all specified before the tests themselves?

jwallace42: It can't be inside the `if(BUILD_TESTING)`.

nav2_system_tests/src/waypoint_follower/tester.py
jwallace42: A added a test inside the error_codes to verify the piping.

nav2_system_tests/src/error_codes/test_error_codes_launch.py
jwallace42: Same as above

nav2_system_tests/src/error_codes/error_code_param.yaml
SteveMacenski: You don't need to bring up everything using the simple navigator // wait for things to initilize since you shouldn't launch the other stuff

nav2_system_tests/src/error_codes/test_error_codes_launch.py
SteveMacenski: Please fix.

nav2_system_tests/CMakeLists.txt
SteveMacenski: That's probably why.

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: Is the next number in line actually 6000?

nav2_system_tests/CMakeLists.txt
jwallace42: oh, sorry I misunderstood. 

Yes, it was placed before the tests. I though you were referring to the first cmake file.

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -345,49 +345,61 @@ def _getPathImpl(self, start, goal, planner_id='', use_start=False):
...
+    def _getPathThroughPosesImpl(self, start, goals, planner_id='', use_start=False):
+        """
+        Send a `ComputePathThroughPoses` action request.
+
+        Internal implementation to get the full result, not just the path.
+        """
         self.debug("Waiting for 'ComputePathThroughPoses' action server")
         while not self.compute_path_through_poses_client.wait_for_server(timeout_sec=1.0):
             self.info("'ComputePathThroughPoses' action server not available, waiting...")
-
```
SteveMacenski, line 367: Please readd the code breaks deleted

nav2_system_tests/CMakeLists.txt
```diff
@@ -47,28 +47,65 @@ set(dependencies
...
-  add_subdirectory(src/system_failure)
-  add_subdirectory(src/updown)
-  add_subdirectory(src/waypoint_follower)
-  add_subdirectory(src/behaviors/spin)
-  add_subdirectory(src/behaviors/wait)
-  add_subdirectory(src/behaviors/backup)
-  add_subdirectory(src/behaviors/drive_on_heading)
-  add_subdirectory(src/behaviors/assisted_teleop)
-  add_subdirectory(src/costmap_filters)
+#  add_subdirectory(src/behavior_tree)
```
SteveMacenski, line 91: comment back in?

nav2_system_tests/src/waypoint_follower/tester.py
SteveMacenski: I don't see anywhere in that which replaces the full failure -> application like what the WPF or Nav2Pose test would provide. I really think we need to have a test here as a strict requirement.

nav2_waypoint_follower/src/waypoint_follower.cpp
jwallace42: 100-199 controller
200-299 compute path to pose 
300-399 compute path through poses 
400-499 smoother server? 

So, it depends on if there is anything else we want to add.

SteveMacenski: OK - leave it as 600, I have something in mind between them for early next year

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -246,6 +271,32 @@ void BtActionServer<ActionT>::executeCallback()
   }
 }
 
+template<class ActionT>
+void BtActionServer<ActionT>::populateErrorCode(
+  typename std::shared_ptr<typename ActionT::Result> result)
+{
+  int highest_priority_error_code = std::numeric_limits<int>::max();
+  for (const auto & error_code_id : error_code_ids_) {
+    // TODO(jwallace42): Using try catch due to
```
SteveMacenski, line 280: TODO?

nav2_system_tests/src/waypoint_follower/tester.py
```diff
@@ -208,12 +208,14 @@ def main(argv=sys.argv[1:]):
...
     test.cancel_goal()
 
-    # a failure case
+    # set waypoint outside of map
     time.sleep(2)
     test.setWaypoints([[100.0, 100.0]])
     result = test.run(True)
     assert not result
     result = not result
+    print("ERROR CODES: ", test.result.error_codes)
```
SteveMacenski, line 217: Remove print if done?

SteveMacenski:
Test coverage looks good! The only thing I see missing is https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L384 where I would expect a plugin to just return true but empty and get a return code of 0-points in the path due to an empty path return from a dummy plugin that does nothing

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -237,16 +243,25 @@ WaypointFollower::followWaypoints()
       RCLCPP_INFO(
         get_logger(), "Task execution at waypoint %i %s", goal_index,
         is_task_executed ? "succeeded" : "failed!");
+
+      if (!is_task_executed) {
+        missed_waypoints_.emplace_back(MissedWaypoint{goal_index, current_goal_status_.error_code});
```
SteveMacenski, line 248: How is this status code set to the missed waypoint code from the message enum? Shouldn't this be manually set to the WPF error code? Or is this changed somewhere I missed?

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: @jwallace42 was this dropped for some technical reason?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
jwallace42: Once // https://github.com/BehaviorTree/BehaviorTree.CPP/issues/271 gets fixed the try catch can be removed.

SteveMacenski: OK - but you have a TODO, I don't allow TODOs to be merged. If its to-do, it should be do-ed :laughing:

jwallace42: Makes sense. 

I was following the old pattern :). 

https://github.com/ros-planning/navigation2/blob/main/nav2_navfn_planner/src/navfn_planner.cpp#L231

https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/test/deprecated/upsampler_cost_function.hpp#L251

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -277,7 +277,7 @@ void BtActionServer<ActionT>::populateErrorCode(
 {
   int highest_priority_error_code = std::numeric_limits<int>::max();
   for (const auto & error_code_id : error_code_ids_) {
-    // TODO(jwallace42): Using try catch due to
+    // Note: Using try catch due to
```
SteveMacenski, line 280: Why not just delete all this comment? I don't see that the context is super important

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: Well I hope 97 of them are before my time. Those in the constrained smoother are in a depreciated file (see the directory) and not meant for production. That's there for historical reference and was never completed

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -59,6 +60,24 @@ BtActionServer<ActionT>::BtActionServer(
...
+    "follow_path_error_code_id",
+    "compute_path_error_code_id"
+  };
+
+  if (!node->has_parameter("error_code_id_names")) {
+    std::string error_code_ids_str;
+    for (const auto & error_code_id : error_code_id_names) {
+      error_code_ids_str += error_code_id + "\n";
+    }
+    RCLCPP_INFO_STREAM(
```
SteveMacenski, line 74:
```suggestion
    RCLCPP_WARN_STREAM(
```

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -59,6 +60,24 @@ BtActionServer<ActionT>::BtActionServer(
   if (!node->has_parameter("default_server_timeout")) {
     node->declare_parameter("default_server_timeout", 20);
   }
+
+  std::vector<std::string> error_code_id_names = {
+    "follow_path_error_code_id",
```
SteveMacenski, line 65: I think these should be `follow_path_error_code` without the ID. The ID is the identifier for the code (e.g. `error_code_id` port to be populated about what ID the error code is stored in). The actual error code storage variable shouldn't have ID.

Please fix here + BTs + configuration file

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
```diff
@@ -86,7 +86,7 @@ class ComputePathThroughPosesAction
           "Mapped name to the planner plugin type to use"),
         BT::OutputPort<nav_msgs::msg::Path>("path", "Path created by ComputePathThroughPoses node"),
         BT::OutputPort<ActionResult::_error_code_type>(
-          "compute_path_through_poses_error_code", "The compute path through poses error code"),
```
SteveMacenski, line 89: Shouldn't the NavigateToPose and NavigateThroughPoses BT Nodes also have the output port for the error code to set if there are autonomy applications using them (since the piping now exposes those return codes through to the action's caller)?

nav2_waypoint_follower/include/nav2_waypoint_follower/waypoint_follower.hpp
```diff
@@ -43,6 +43,18 @@ enum class ActionStatus
   SUCCEEDED = 3
 };
 
+struct MissedWaypoint
```
SteveMacenski, line 46: We had a discussion about adding a new nav2_msgs/MissedWaypoint message for this so that it can be directly correlated more easily in the message. Was that dropped for a technical reason?

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
SteveMacenski: I don't think you need to do anything with them other than exposing the output port and setting the error code in those 2 BT nodes. We don't have any integration tests using these nodes that would need updating and we already show that the Nav2Pose action pipes correctly  from your WPF test updates so this can just be done in code and we're good

SteveMacenski: Named navigate_to_pose_error_code + navigate_through_poses_error_code

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/navigate_through_poses_action.hpp
```diff
@@ -57,6 +76,8 @@ class NavigateThroughPosesAction : public BtActionNode<nav2_msgs::action::Naviga
       {
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goals", "Destinations to plan through"),
         BT::InputPort<std::string>("behavior_tree", "Behavior tree to run"),
+        BT::OutputPort<ActionResult::_error_code_type>(
+            "navigate_through_poses_error_code", "The navigate through poses error code"),
```
SteveMacenski, line 80: Didn't we decide all these to be `error_code_id`?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/navigate_to_pose_action.hpp
```diff
@@ -57,6 +76,8 @@ class NavigateToPoseAction : public BtActionNode<nav2_msgs::action::NavigateToPo
       {
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
         BT::InputPort<std::string>("behavior_tree", "Behavior tree to run"),
+        BT::OutputPort<ActionResult::_error_code_type>(
+            "navigate_to_pose_error_code", "Navigate to pose error code"),
```
SteveMacenski, line 80: Ditto

SteveMacenski:
Got some linting errors

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/navigate_through_poses_action.hpp
SteveMacenski: I think there's also going to need to be documentation updates for the new ports + migration guide

SteveMacenski:
> 89.33%

Ah shi-zit, we're getting mighty close

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/navigate_through_poses_action.hpp
jwallace42: > Named navigate_to_pose_error_code + navigate_through_poses_error_code

I just copied this.

SteveMacenski (approved):

PR #3255, opened 2022-10-24
Give behavior server access to both costmaps

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

This pr allows recoveries to have access to a local and global collision checker. 

https://github.com/ros-planning/navigation.ros.org/pull/360

---

#### For Maintainers: 


SteveMacenski:
Question: 

Is it that you want to have **both** or that you want the option to instead use the global costmap? We should be able to make the behavior server work with either or pretty trivially (might even just be only renaming variable names to make it clear it could be either depending on what the costmap / footprint topics / frames are given to it)

nav2_behaviors/test/test_behaviors.cpp
```diff
@@ -103,35 +103,68 @@ class BehaviorTest : public ::testing::Test
...
-    std::shared_ptr<nav2_costmap_2d::CostmapSubscriber> costmap_sub_ =
+    std::string local_costmap_topic, global_costmap_topic;
+    std::string local_footprint_topic, global_footprint_topic;
+    node_lifecycle_->get_parameter("local_costmap_topic", local_costmap_topic);
+    node_lifecycle_->get_parameter("global_costmap_topic", global_costmap_topic);
+    node_lifecycle_->get_parameter("local_footprint_topic", local_footprint_topic);
+    node_lifecycle_->get_parameter("global_footprint_topic", global_footprint_topic);
+
+
+    std::cout << "Starting " << std::endl;
```
SteveMacenski, line 134: debug logging

SteveMacenski:
Just a linting error on test failure now.

But the global frame stuff is still not there. 

May I ask, is this for a single, specific behavior or for a series of behaviors? If this is a single behavior that needs both costmaps, I might say its better to just implement that as a separate server of its ownright rather than complicating this one. If its a number of behaviors that need both, then maybe its worth actually taking a bit of time for us to think about how we pass and work with parameter data for this server if its going to increase to multiple collision checkers / frames / footprints. 

Potentially, we could ask the plugins what it wants (new method) for global/local so that we only subscribe to the ones required. Adding in a global costmap subscriber is **really, non-trivial** in cost, especially if no one is actually currently using it. There's a reason this server centralizes behaviors, even just streaming the local costmap at 5hz given its size is non-trivial.

jwallace42:
> Just a linting error on test failure now.
> 
> But the global frame stuff is still not there.
Ahh!

> May I ask, is this for a single, specific behavior or for a series of behaviors? If this is a single behavior that needs both costmaps, I might say its better to just implement that as a separate server of its ownright rather than complicating this one. If its a number of behaviors that need both, then maybe its worth actually taking a bit of time for us to think about how we pass and work with parameter data for this server if its going to increase to multiple collision checkers / frames / footprints. Maybe having a behavior information struct containing a local / global struct of values, since they exactly mirror each other. E.g. my_info containing global_info and local_info each are instances of a struct info since they both involve costmap subs, footprint subs, and global frames (obviously not using those terrible names).

It is for a series of behaviors that need both costmaps. 

> Potentially, we could ask the plugins what it wants (new method) for global/local so that we only subscribe to the ones required. Adding in a global costmap subscriber is really, non-trivial in cost, especially if no one is actually currently using it. There's a reason this server centralizes behaviors, even just streaming the local costmap at 5hz given its size is non-trivial. The server can have the logic for both, but only does the minimum required. based on if any of its behaviors desire it.

I think this sounds great since it will minimize changes to the current behaviors(and tests) while adding in the desired flexibility. If this is our path forward I can close this pr.

SteveMacenski:
Why close it? Isn't it just adding in if condition and an API in the behaviors as to what they want?

jwallace42:
That seems like a way we could go but I don't see that scaling well if we end up sharing more things across the behavior servers (maybe we don't). I was thinking about using https://en.cppreference.com/w/cpp/language/parameter_pack to feasibility pass different inputs to the concrete implementation. I have just starting looking into this so maybe it won't work :).

SteveMacenski:
I'm not entirely sure what you mean, can you explain what you're thinking? I only suggested that rather than passing the behaviors a ton of objects, we just create a struct to manage them all + an API for each behavior to tell the server which, or both or neither, of the cost information they care about so we only make subscriptions to things strictly necessary.

jwallace42:
Sure, 

I was hoping we that parameters packs would allow us to configure the different servers to support a api as such:  

`configure(local_costmap, global_costmap, tf)` needs 3 things
`configure(local_costmap, tf)` only needs two 
`configure(tf)`  just needs one :).  

So, 

you suggest that we have a struct that contains the necessary info that gets passed into the behavior though the configure func and the user selects what info to keep?

Or do we have multiple configure functions that support passing in only the necessary objects defined by some parameters? Similar to what I have above?

SteveMacenski:
> configure(local_costmap, tf)

The problem with that is how to differentiate between the local and global costmap if only one is given. How do they differ? I'll admit I'm not super familiar with parameter packs. It also doesn't entirely solve the problem because you still need to know which constructor to evoke for each given plugin, which would still require and API to query about what it wants. 

I'm just suggesting the struct to manage the data better. But If you had a different design direction in mind that's better, I'm all ears. For how it is right now with just the local/global collision checker its fine without that. But I see this potentially getting more complicated once you have to add in the global/local reference frames (or it doesn't).

jwallace42:
Let me know what you think of this solution. If want to go with the struct it is a easy switch.

nav2_bringup/params/nav2_params.yaml
```diff
@@ -296,7 +298,8 @@ behavior_server:
       plugin: "nav2_behaviors/Wait"
     assisted_teleop:
       plugin: "nav2_behaviors/AssistedTeleop"
-    global_frame: odom
+    odom_frame: odom
```
SteveMacenski, line 301: I'd prefer if these more 1-1 aligned with the global/local nomenclature established for their counterpart information. How about `local_reference_frame` and `global_reference_frame`?

nav2_behaviors/src/behavior_server.cpp
```diff
@@ -111,7 +136,12 @@ BehaviorServer::loadBehaviorPlugins()
         get_logger(), "Creating behavior plugin %s of type %s",
         behavior_ids_[i].c_str(), behavior_types_[i].c_str());
       behaviors_.push_back(plugin_loader_.createUniqueInstance(behavior_types_[i]));
-      behaviors_.back()->configure(node, behavior_ids_[i], tf_, collision_checker_);
+      behaviors_.back()->configure(
```
SteveMacenski, line 139: I'd like a new Behavior plugin API to return an enum 

```
enum class CostmapInfoType {
  NONE = 0,
  LOCAL = 1,
  GLOBAL = 2,
  BOTH = 3
};

...

CostmapInfoType getRequiredCostmapInfoTypes() {return CostmapInfoType::LOCAL;}
```

From that, we can query all the plugins as to which they need. If there are any (global or local) that are not required by any plugins, then that collision checker, footprint sub, and costmap sub can be reset. Especially the costmap sub for the global costmap is non-trivial in its overhead if no one requires it.

SteveMacenski (changes requested):
A couple of things but otherwise LGTM. Will obviously require docs updates for new params + migration guide

nav2_bringup/params/nav2_params.yaml
SteveMacenski: Or maybe just `local_frame` and `global_frame`?

nav2_behaviors/src/behavior_server.cpp
SteveMacenski: You don't need to use those exact class/function names, I was just trying to be illustrative

jwallace42: Alright, 

That sounds good.

nav2_core/include/nav2_core/behavior.hpp
```diff
@@ -65,6 +73,11 @@ class Behavior
    * @brief Method to deactive Behavior and any threads involved in execution.
    */
   virtual void deactivate() = 0;
+
+  /**
+   * @brief Method to deterimen the required costmap info
+   */
+  CostmapInfoType getResourceInfo() {return CostmapInfoType::LOCAL;}
```
SteveMacenski, line 80: Shouldn't this be a virtual function so that it can be overridden for each plugin? I also don't think this is reasonable to have a default, this should be a forced implementation `= 0` for each plugin to setup

nav2_behaviors/src/behavior_server.cpp
```diff
@@ -124,6 +151,74 @@ BehaviorServer::loadBehaviorPlugins()
...
+  std::string robot_base_frame;
+  double transform_tolerance;
+  get_parameter("local_costmap_topic", local_costmap_topic);
+  get_parameter("global_costmap_topic", global_costmap_topic);
+  get_parameter("local_footprint_topic", local_footprint_topic);
+  get_parameter("global_footprint_topic", global_footprint_topic);
+  get_parameter("transform_tolerance", transform_tolerance);
+  get_parameter("robot_base_frame", robot_base_frame);
+
+  for (const auto & behavior : behaviors_) {
```
SteveMacenski, line 183: Suggestion: This obviously works too, but maybe easier to read would be something like 

```
bool need_local_costmap = false; need_global_costmap = false;
for (const auto & behavior : behaviors_) {
  if (behavior->getResourceInfo() == both) {// set both to true and break, no reason to continue}
  if (behavior->getResourceInfo() == local) {local = true}
  if (behavior->getResourceInfo() == global) {global = true}
}
```

Then you can just use the  bools in the statements to create the objects which might be easier to read the `if` statement :-)

SteveMacenski: The multi-line multi-scope if statement is a bit much potentially

nav2_behaviors/src/behavior_server.cpp
```diff
@@ -124,6 +123,72 @@ BehaviorServer::loadBehaviorPlugins()
...
+  get_parameter("global_costmap_topic", global_costmap_topic);
+  get_parameter("local_footprint_topic", local_footprint_topic);
+  get_parameter("global_footprint_topic", global_footprint_topic);
+  get_parameter("transform_tolerance", transform_tolerance);
+  get_parameter("robot_base_frame", robot_base_frame);
+
+  bool need_local_costmap = false;
+  bool need_global_costmap = false;
+  for (const auto & behavior : behaviors_) {
+    if (behavior->getResourceInfo() == nav2_core::CostmapInfoType::BOTH) {
```
SteveMacenski, line 156: `auto & type = behavior->getResourceInfo()` then `type` can be used in the if statements `if (type == BOTH)`, etc.

SteveMacenski (changes requested):
Last change here, then just need the docs PR for the new parameters and parameter name updates. Needs migration guide about this as well since we change parameters without backward compatibility - just to save you time by mentioning up front - please make sure that the Migration Guide changes are **explicit** about which parameters they need to change the names to to get the same behavior.

SteveMacenski (approved):

PR #3259, opened 2022-10-26
Increase BT action node's action server timeout

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu Jammy |
| Robotic platform tested on | own simulation |

---

Sometimes the BT nodes are launched before the action servers (as part of a launch file) causing the BT nodes initialization to fail, so I increased the timeout from 1 to 10 seconds

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -93,9 +93,9 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    if (!action_client_->wait_for_action_server(1s)) {
+    if (!action_client_->wait_for_action_server(10s)) {
       RCLCPP_ERROR(
-        node_->get_logger(), "\"%s\" action server not available after waiting for 1 s",
+        node_->get_logger(), "\"%s\" action server not available after waiting for 10 s",
         action_name.c_str());
       throw std::runtime_error(std::string("Action server %s not available", action_name.c_str()));
     }


jwallace42:
I assume you are seeing this behavior in a custom launch file? I am not able to reproduce with the default launch files. 

You would also have to extend this timeout for the other Bt base classes. (service, cancel ....). 

I think the best approach is to pull this out into a parameter and allow the user to set the max wait time.

SteveMacenski:
They shouldn’t be an issue when using lifecycle nodes. the lifecycle manager only brings up constituent nodes after the previous others  are already initialized. Maybe you just need to select a more reasonable order for node initialization of the navigation system? The order of nodes in the launch file given to the lifecycle manager is not arbitrary.

SteveMacenski:
any update?

tonynajjar:
not yet but not forgotten!

tonynajjar:
> I assume you are seeing this behavior in a custom launch file? I am not able to reproduce with the default launch files.

yes custom launch files

> This shouldn’t be an issue when using lifecycle nodes. the lifecycle manager only brings up constituent nodes after the previous others are already initialized. Maybe you just need to select a more reasonable order for node initialization of the navigation system?

I see what you mean and I think I see the issue now. I have custom BT Action nodes as well as custom behaviors/action servers. These action servers are launched non-deterministically in the launch file; they are python nodes so unfortunately I can't use lifecycle nodes.

Do you have other recommendations for my case other than increasing the timeout?

SteveMacenski:
> They are python nodes so unfortunately I can't use lifecycle nodes

Can't you? https://github.com/ros2/rclpy/tree/rolling/rclpy/rclpy/lifecycle they appear to be in Humble and Rolling

But otherwise, I believe that Launch is deterministic in execution order provided, so you can bring up all the servers in the launch file first before the BT navigator. Maybe that'll be enough? 

Hackier: use the launch prefix field to basically ask for a sleep as this answer points out: https://answers.ros.org/question/233353/set-delay-between-starting-nodes-within-launch-file/ for the BT navigator node to give the other nodes times to initialize. 

Better yet: https://www.reddit.com/r/ROS/comments/qxbi3z/ros2_is_it_possible_to_have_something_delayed_at/ there's a timed delay action launch node it appears. 

Best: use the rclpy lifecycle nodes

tonynajjar:
> Can't you? https://github.com/ros2/rclpy/tree/rolling/rclpy/rclpy/lifecycle they appear to be in Humble and Rolling

> Best: use the rclpy lifecycle nodes

Oh I thought these would never see the light, thanks for the info. But it's too much work for the moment

> But otherwise, I believe that Launch is deterministic in execution order provided, so you can bring up all the servers in the launch file first before the BT navigator. Maybe that'll be enough?

really? But even then I don't think launching the action servers a couple of seconds earlier would fix the issue

> there's a timed delay action launch node it appears.

Yes I know about this one, might be a somewhat hacky solution.


I still think increasing the timeout would be the easiest/quickest solution, do you see any drawbacks to it (apart from users having to wait a bit longer to see an error message)

SteveMacenski:
> But it's too much work for the moment

:cry: 

> do you see any drawbacks to it 

That's a long delay to fail and would give users a bad user-experience if they launched BTs with task servers legit just not existing. Waiting 10 seconds to throw and then fail the navigation task is not very "reactive". As a user of this, I'd be pretty frustrated to have to wait 10 seconds to be told of a trivial typo error while prototyping.

tonynajjar:
Alright then, if there is not solution to contribute here, feel free to close the PR. Thanks for the alternatives, will look into them more.

SteveMacenski:
OK - sorry for not being able to get this one in. But I think there are some optimal (lifecycle py) options and sketchier but suitable solutions (timed action) at launch time.

PR #3265, opened 2022-11-03
refactoring RPP a bit for cleanliness on way to ROSCon

SteveMacenski:
TODO 
- [x] Test path handling works properly
- [x] Test parameter read-in and dynamic parameters still work
- [x] Test Speed limits still work
- [x] Test collision checking works
- [x] Test collision arc publication works
- [x] Test regulation heuristics work
- [x] Test slow on approach works

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/CMakeLists.txt
+++ b/nav2_regulated_pure_pursuit_controller/CMakeLists.txt
@@ -35,7 +35,10 @@ set(dependencies
 set(library_name nav2_regulated_pure_pursuit_controller)
 
 add_library(${library_name} SHARED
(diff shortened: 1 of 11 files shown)


SteveMacenski:
FYI @Aposhian I did a refactor on RPP. I did some testing that I'm reasonably confident nothing's broken but just letting you know you may want to scrutinize performance once pulling these changes in. I  don't think I missed anything, but worth a note of caution whenever changing this much code at once.

Aposhian:
I like this change, organizationally. Nice job! Our attention is on other things right now, so we probably won't get to testing the new version for a while, but that is something to watch when we jump to the next Nav2 release. Thanks for the heads up!

PR #3266, opened 2022-11-03
Reclaim Our CI Coverage from the Lords of Painful Subtle Regressions ⚔️⚔️⚔️

SteveMacenski:


SteveMacenski:
@ruffsl and I are the shields that defend the land of navigation from woeful regressions

Our swords cut through the mightiest of enemies to save the realm from [20 robot pileups](https://www.sfgate.com/local/article/driverless-cruise-cars-block-SF-traffic-17467985.php) 

We are the watchers on the repo

PR #3278, opened 2022-11-11
Error code tests

### LINKED ISSUE #3043: Feature Request: Feedback for Navigation Failures

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

This pr creates integration tests for servers producing error codes to resolve #3043  

---

#### For Maintainers: 


jwallace42:
This is a rough pr to show my attempt to test error code production of the different nav2 servers. 

There is currently a few things wrong that I plan to fix. 
1. The package needs to be moved into the nav2 system tests. It currently exists in the nav2_error_code_test and nav2_simple_commander. 
2. Support a way to inject and modify the nav2_param.yaml file to insert the plugins.

SteveMacenski:
You can declare / set the parameters in the node before you configure the server instance if you have the server + testing stuff in 1 process. Otherwise, in the launch file, you can give the node some parameters to set the plugins instead of passing a configuration file. Or, you could just have a `tester_params.yaml` in the test directory that your launch file uses. All works

nav2_error_code_test/include/nav2_error_code_test/global_planners/unknown_error_planner.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "nav_msgs/msg/path.hpp"
+#include "nav2_util/robot_utils.hpp"
+#include "nav2_util/lifecycle_node.hpp"
+#include "nav2_costmap_2d/costmap_2d_ros.hpp"
+#include "nav2_core/planner_exceptions.hpp"
+
+namespace nav2_error_code_test
+{
+
+class UnknownErrorPlanner : public nav2_core::GlobalPlanner
```
SteveMacenski, line 35: I'd put this in the same file as the other planner plugins (ditto with controllers all in 1 file)

nav2_simple_commander/launch/test_error_codes_launch.py
```diff
@@ -0,0 +1,45 @@
...
+        launch_arguments={'headless': 'True',
+                          'use_rviz': 'False'}.items())
+
+    test_error_code = Node(
+        package='nav2_simple_commander',
+        executable='test_error_codes',
+        emulate_tty=True,
+        output='screen')
+
+    ld = LaunchDescription()
```
SteveMacenski, line 42: Look at the system tests for Nav2 for examples of how to do a testing script with launch. There's a testing API that will launch some nodes + also launch a "testing script". It works such that once the testing script exits, it kills everything since the test is over

Issue #3285, opened 2022-11-15
Theta star planner fails to get path when goals falls in unknown space, where tile value = NO_INFORMATION


pepisg:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 22
- ROS2 Version:
  - Humble
- Version or commit hash:
  - 0d86469 (latest humble branch)
  - cyclone DDS

#### Steps to reproduce issue
Launch the turtlebot3 simulation with these paramters:

```.yaml

global_costmap:
  global_costmap:
    ros__parameters:
      update_frequency: 1.0
      publish_frequency: 1.0
      global_frame: map
      robot_base_frame: base_link
      rolling_window: true
      width: 20
      height: 20
      robot_radius: 0.22
      resolution: 0.05
      track_unknown_space: true
      plugins: ["obstacle_layer", "inflation_layer"]
      obstacle_layer:
        plugin: "nav2_costmap_2d::ObstacleLayer"
        enabled: True
        observation_sources: scan
        scan:
          topic: /scan
          max_obstacle_height: 2.0
          clearing: True
          marking: True
          data_type: "LaserScan"
          raytrace_max_range: 3.0
          raytrace_min_range: 0.0
          obstacle_max_range: 2.5
          obstacle_min_range: 0.0
      static_layer:
        plugin: "nav2_costmap_2d::StaticLayer"
        map_subscribe_transient_local: True
      inflation_layer:
        plugin: "nav2_costmap_2d::InflationLayer"
        cost_scaling_factor: 3.0
        inflation_radius: 0.55
      always_send_full_costmap: True

planner_server:
  ros__parameters:
    expected_planner_frequency: 20.0
    planner_plugins: ["GridBased"]
    GridBased:
      plugin: "nav2_theta_star_planner/ThetaStarPlanner"
      how_many_corners: 8
      w_euc_cost: 1.0
      w_traversal_cost: 2.0
      w_heuristic_cost: 1.0

```

They are basically the default parameters but they use a rolling global costmap without a static layer (so we can easily send the robot to an unknown tile) and the defaults for the Theta Star planner.

Then try to send the robot to a place marked as unknown in the global costmap.

#### Expected behavior

The theta star planner generates a path to the requested goal even if it lies on a tile whose value is NO_INFORMATION,

#### Actual behavior

The theta star planner fails to compute a path and shows the following error message: `Either of the start or goal pose are an obstacle! `

#### Additional information

[...]

PR #3286, opened 2022-11-15
Add allow_unknown parameter to theta star planner

pepisg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3285 |
| Primary OS tested on | Ubuntu 22 |
| Robotic platform tested on | Custom gazebo robot |

---

Added the allow_unknown parameter to the theta star planner to make it possible to plan through unknown space

I need to add the new parameter to the configuration guide in navigation.ros.org

---

#### For Maintainers: 

### DIFF
--- a/nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
+++ b/nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
@@ -95,7 +95,9 @@ public:
    */
   inline bool isSafe(const int & cx, const int & cy) const
   {
-    return (costmap_->getCost(cx, cy) == UNKNOWN_COST && allow_unknown_) || costmap_->getCost(cx, cy) < LETHAL_COST;
+    return (costmap_->getCost(
+             cx,
+             cy) == UNKNOWN_COST && allow_unknown_) || costmap_->getCost(cx, cy) < LETHAL_COST;
   }
 
   /**
@@ -190,7 +192,7 @@ protected:
   {
     double curr_cost = getCost(cx, cy);
     if ((costmap_->getCost(cx, cy) == UNKNOWN_COST && allow_unknown_) || curr_cost < LETHAL_COST) {
-      if(costmap_->getCost(cx, cy) == UNKNOWN_COST) {
+      if (costmap_->getCost(cx, cy) == UNKNOWN_COST) {
         curr_cost = OBS_COST - 1;
       }
       cost += w_traversal_cost_ * curr_cost * curr_cost / LETHAL_COST / LETHAL_COST;
--- a/nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
+++ b/nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
@@ -24,8 +24,9 @@
 #include "nav2_costmap_2d/costmap_2d_ros.hpp"
 
 const double INF_COST = DBL_MAX;
-const int LETHAL_COST = 252;
 const int UNKNOWN_COST = 255;
+const int OBS_COST = 254;
+const int LETHAL_COST = 252;
 
 struct coordsM
 {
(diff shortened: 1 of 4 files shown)


nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
```diff
@@ -185,7 +188,7 @@ class ThetaStar
   bool isSafe(const int & cx, const int & cy, double & cost) const
   {
     double curr_cost = getCost(cx, cy);
-    if (curr_cost < LETHAL_COST) {
+    if ((costmap_->getCost(cx, cy) == UNKNOWN_COST && allow_unknown_) || curr_cost < LETHAL_COST) {
```
pepisg, line 191: not sure if a different cost should be considered for the planning itself when the cell is unknown rather than using the maximum (which if I understand correctly would penalize unknown tiles the same way as obstacles)

nav2_theta_star_planner/src/theta_star_planner.cpp
```diff
@@ -237,6 +241,8 @@ ThetaStarPlanner::dynamicParametersCallback(std::vector<rclcpp::Parameter> param
     } else if (type == ParameterType::PARAMETER_BOOL) {
       if (name == name_ + ".use_final_approach_orientation") {
         use_final_approach_orientation_ = parameter.as_bool();
+      } else if (name == name_ + ".allow_unknown") {
+        planner_->allow_unknown_ = parameter.as_bool();
```
SteveMacenski, line 245: Add test coverage for this changing in the unit tests (should be very clear when you open the file)

nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
SteveMacenski: Look at what NavFn does with this value and you can copy that methodology here

nav2_theta_star_planner/src/theta_star_planner.cpp
```diff
@@ -45,6 +45,10 @@ void ThetaStarPlanner::configure(
     RCLCPP_WARN(logger_, "Your value for - .how_many_corners  was overridden, and is now set to 8");
   }
 
+  nav2_util::declare_parameter_if_not_declared(
+    node, name_ + ".allow_unknown", rclcpp::ParameterValue(true));
```
SteveMacenski, line 49: Need to add to the configuration guide to Theta* in docs new param

SteveMacenski (changes requested):
LGTM

nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
pepisg: just to confirm, that is [setting it to the obstacle cost - 1](https://github.com/ros-planning/navigation2/blob/main/nav2_navfn_planner/src/navfn.cpp#L266)?

nav2_theta_star_planner/src/theta_star_planner.cpp
pepisg: done here: https://github.com/ros-planning/navigation.ros.org/pull/371

SteveMacenski:
What conclusions did you come to about costs for UNKNOWN?

SteveMacenski:
I think doing the same thing as NavFn is reasonable - no one's complained about that in years. Good enough for that, good enough for us here :laughing:

nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
```diff
@@ -185,7 +189,10 @@ class ThetaStar
...
+      if(costmap_->getCost(cx, cy) == UNKNOWN_COST) {
```
SteveMacenski, line 193: Linting issue, which I'm sure CI picked up on

SteveMacenski (changes requested):

nav2_theta_star_planner/include/nav2_theta_star_planner/theta_star.hpp
SteveMacenski: CI actually shows 2 linting issues

pepisg: Can you recommend a tool for autostyling the code so it complies with the project's standards? or do you have a clang-format file I can use?

SteveMacenski: ament_cpplint and ament_uncrustify or just run colcon test locally which will invoke them

SteveMacenski (approved):

PR #3296, opened 2022-11-21
Smoother error codes

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3043 |

---

- added error codes to smoother server 
- added error codes to smoother bt action node 
- added tests smoother error codes

---

#### For Maintainers: 


jwallace42:
Hey Steve!

The one thing that I would like your close attention for is the timeout error. From reading the server this weekend it seems like we still would like the server to succeed if it times out. (I.e partial smoothing has occurred) and just set a blackboard var to let anything else know that the smoothing was not completed. If this is the intention, I don't think timeout is a necessary error code since we handle it in the server. 

My current implementation is the rough structure assuming that timeout is indeed a error. If not it can be easily changed.

SteveMacenski:
On PTO this week, but I took a quick glance on the plane and this looks good to me assuming CI turns over and the details align. 

Only potential question I have right now is the error code setting for BT navigator / documentation updates.

SteveMacenski:
CI shows some linting issues + comments on missing stuff above stands. Otherwise LGTM now in front of a computer

nav2_smoother/src/nav2_smoother.cpp
```diff
@@ -274,15 +273,6 @@ void SmootherServer::smoothPlan()
       result->path, goal->max_smoothing_duration);
     result->smoothing_duration = steady_clock_.now() - start_time;
 
-    if (!result->was_completed) {
```
SteveMacenski, line 277: Why not still log this?

nav2_smoother/src/simple_smoother.cpp
```diff
@@ -174,7 +176,7 @@ bool SimpleSmoother::smoothImpl(
           "Returning the last path before the infeasibility was introduced.");
         path = last_path;
         updateApproximatePathOrientations(path, reversing_segment);
-        return false;
+        throw nav2_core::SmoothedPathInCollision("Smoothed Path in collision");
```
SteveMacenski, line 179: One thing I kind of liked about this was that we'd still try to smooth the other path segments. Just because one is in collision and we stop doesn't mean others couldn't be refined too.

As the warning mentions, this actually returns a still-valid path, just an iteration prior to the collision so its still valid. I'm not sure that throwing here would be the right move.

nav2_msgs/action/SmoothPath.action
```diff
@@ -1,3 +1,11 @@
+# Error codes
+# Note: The expected priority order of the errors should match the message order
+uint16 NONE=0
+uint16 UNKNOWN=500
+uint16 INVALID_SMOOTHER=501
```
SteveMacenski, line 5: constrained smoother updates?

nav2_smoother/src/simple_smoother.cpp
jwallace42: Yeah, I agree :). That was thing I mentioned here: 

> From reading the server this weekend it seems like we still would like the server to succeed if it times out. (I.e partial smoothing has occurred) and just set a blackboard var to let anything else know that the smoothing was not completed.

I assumed partial smoothing is better than none at all :) but I wasn't sure.

jwallace42:
This is looking like a Wednesday completion :). I got class tomorrow.

SteveMacenski:
> Only potential question I have right now is the error code setting for BT navigator / documentation updates. Also missing TF and invalid inputs (path invalid or empty) error code

Still valid :-)

nav2_constrained_smoother/src/constrained_smoother.cpp
```diff
@@ -137,8 +137,8 @@ bool ConstrainedSmoother::smooth(nav_msgs::msg::Path & path, const rclcpp::Durat
       logger_,
       "%s: failed to smooth plan, Ceres could not find a usable solution to optimize.",
       plugin_name_.c_str());
-    throw std::runtime_error(
-            "Failed to smooth plan, Ceres could not find a usable solution.");
+    throw nav2_core::FailedToSmoothPath("Failed to smooth plan, Ceres could not find a usable "
```
SteveMacenski, line 140: There aren't any other places where we can use the exceptions in this smoother?

SteveMacenski: Also, no updates to the simple smoother anymore?

SteveMacenski: Or savitsky-golay? These exceptions aren't used anywhere then!

SteveMacenski (changes requested):

nav2_constrained_smoother/src/constrained_smoother.cpp
jwallace42: We would still have all the exceptions except timeout. 

How about we do this for the savitsky-golay and the simple smoother if and only if all segments fail to be smoothed we throw the FailedToSmoothPath error.

jwallace42:
> > missing TF and invalid inputs (path invalid or empty) error code
It appears that the current set of smoothers and the smoother server do not transform anything. So I think we can ignore the TF error. I will add the path invalid error in :).

nav2_core/include/nav2_core/smoother_exceptions.hpp
```diff
@@ -0,0 +1,70 @@
...
+};
+
+class InvalidSmoother : public SmootherException
+{
+public:
+  explicit InvalidSmoother(const std::string & description)
+  : SmootherException(description) {}
+};
+
+class InvalidPath : public SmootherException
```
SteveMacenski, line 38: Needs test plugin now that added

SteveMacenski:
> It appears that the current set of smoothers and the smoother server do not transform anything. So I think we can ignore the TF error. I will add the path invalid error in :).

OK! You do have some linting errors now

nav2_constrained_smoother/src/constrained_smoother.cpp
SteveMacenski: Where? In the diff of this PR, there are no modifications to the simple smoother or SG smoother files. The only change to the constrained smoother is the error `FailedToSmoothPath`. Maybe you didnt push something?

jwallace42: Nope, 

I just wanted to see if this sounded good to you before proceeding. :).

jwallace42: Sounds good!

jwallace42: So, 

for 
- https://github.com/ros-planning/navigation2/blob/main/nav2_smoother/src/simple_smoother.cpp#L137
- https://github.com/ros-planning/navigation2/blob/main/nav2_smoother/src/savitzky_golay_smoother.cpp#L70

If we throw a timeout error here, there could be other segments of the path that were smoothed successfully. I thought we wanted to return a partially smoothed path?

SteveMacenski: I think that's true for some things but timeouts are god- when a timeout is exceeded, we return immediately regardless (if you see the code a timeout edits the Impl function. As new segments are called, this is immediately triggered again without any work or very little). So behavior today doesn't actually do anything for additional segments (or nothing of note) when the timeout is exceeded.

nav2_core/include/nav2_core/smoother_exceptions.hpp
jwallace42: Added this in :).

nav2_smac_planner/src/smoother.cpp
```diff
@@ -136,7 +137,7 @@ bool Smoother::smoothImpl(
         "Smoothing time exceeded allowed duration of %0.2f.", max_time);
       path = last_path;
       updateApproximatePathOrientations(path, reversing_segment);
-      return false;
+      throw nav2_core::SmootherTimedOut("Smoothing time exceed allowed duration");
```
SteveMacenski, line 140: This is internal to the path planner not a separate smoother plugin - this should be handled by the planner exceptions / failure codes. Revert.

nav2_smoother/src/savitzky_golay_smoother.cpp
```diff
@@ -49,11 +50,16 @@ bool SavitzkyGolaySmoother::smooth(
   double time_remaining = max_time.seconds();
 
   bool success = true, reversing_segment;
+  unsigned int segments_smoothed = 0;
   nav_msgs::msg::Path curr_path_segment;
   curr_path_segment.header = path.header;
 
   std::vector<PathSegment> path_segments = findDirectionalPathSegments(path);
 
+  if (path_segments.empty()) {
```
SteveMacenski, line 59: I can see what you're thinking here, but this function always returns at least 1 segment for the 1 segment that could exist for a non-empty path which would be checked by your validation function in the server itself. This may be removed

nav2_smoother/src/savitzky_golay_smoother.cpp
```diff
@@ -71,11 +77,16 @@ bool SavitzkyGolaySmoother::smooth(
         RCLCPP_WARN(
           logger_,
           "Smoothing time exceeded allowed duration of %0.2f.", max_time.seconds());
-        return false;
+        throw nav2_core::SmootherTimedOut("Smoothing time exceed allowed duration");
+      }
+
+      bool segment_was_smoothed = smoothImpl(curr_path_segment, reversing_segment);
```
SteveMacenski, line 83: smoothImpl always returns true, these changes are a bit redundant

SteveMacenski (changes requested):
Last little bits and can be merged in this week!

Make sure to update the docs PR and if there's anything I asked for there to merge in

nav2_smoother/src/savitzky_golay_smoother.cpp
SteveMacenski: But are useful in the Simple smoother, I think just here in the SG-Smoother its unnecessary.

jwallace42: Oh?

Why even return a bool if it is always true?

nav2_smoother/test/test_simple_smoother.cpp
```diff
@@ -248,7 +244,7 @@ TEST(SmootherTest, test_simple_smoother)
   straight_irregular_path.poses[6].pose.position.x = 0.5;
   straight_irregular_path.poses[6].pose.position.y = 0.5;
   straight_irregular_path.poses[6].pose.orientation = tf2::toMsg(quat2);
-  EXPECT_TRUE(smoother->smooth(straight_irregular_path, max_time));
+  EXPECT_THROW(smoother->smooth(straight_irregular_path, max_time), nav2_core::FailedToSmoothPath);
```
jwallace42, line 247: Adding the failed to smooth exception exposes that this test currently doesn't do anything. 

The path segments are less than 9 thus nothing gets smoothed. 
https://github.com/ros-planning/navigation2/blob/main/nav2_smoother/src/simple_smoother.cpp#L74

The reason it returns true is because success is initialized to true.

nav2_smoother/test/test_simple_smoother.cpp
```diff
@@ -276,5 +272,5 @@ TEST(SmootherTest, test_simple_smoother)
   max_its_path.poses[9].pose.position.y = 0.9;
   max_its_path.poses[10].pose.position.x = 0.5;
   max_its_path.poses[10].pose.position.y = 1.0;
-  EXPECT_FALSE(smoother->smooth(max_its_path, max_time));
+  EXPECT_THROW(smoother->smooth(max_its_path, max_time), nav2_core::FailedToSmoothPath);
```
jwallace42, line 275: Adding the failed to smooth exception exposes that this test currently doesn't do anything. 

The path segments are less than 9 thus nothing gets smoothed. 
https://github.com/ros-planning/navigation2/blob/main/nav2_smoother/src/simple_smoother.cpp#L74

The reason it returns true is because success is initialized to true.

nav2_smoother/src/savitzky_golay_smoother.cpp
SteveMacenski: The SG-Smoother was a 2 day project and I started with the simple Smoother as a base, so just an overlooked thing

nav2_smoother/test/test_simple_smoother.cpp
SteveMacenski: Got it - well I think this update makes sense, we can keep it :+1:

SteveMacenski (approved):
Rerunning CI since something odd happened but I don't think its due to this work. I think this is good to go now, anything else to add before merging? The docs are the only thing that come to mind, though I see in my notifications tray you updated something there I'll look at shortly.

SteveMacenski:
Sounds good! I think the only thing blocking this being merged is the issues with the documentation PR. 

Next steps seem like: updating BTs to use this feature + more in documentation
- https://navigation.ros.org/ list in bullets of features on main page + description in paragraph 2 when we talk about action servers
- https://navigation.ros.org/concepts/index.html when we talk about task servers 
- https://navigation.ros.org/tuning/index.html perhaps an entry in the tuning guide around using these to communicate contextual failures for people looking to design behavior trees
- https://navigation.ros.org/behavior_trees/index.html Something in the BT section, naturally

Other than the BT section, the other 3 bullets look like: 1 bullet, 2 sentences, and a paragraph, so not too bad at all.

rayferric:
With those changes, simple smoother aborts navigation when the robot approaches the end of path.

This is the cause:
```cpp
  if (segments_smoothed == 0) {
    throw nav2_core::FailedToSmoothPath("No segments were smoothed");
  }
```

I can't seem to find a way to handle this error in any online resources.
  
nav2 params:
```yaml
smoother_server:
ros__parameters:
  costmap_topic: global_costmap/costmap_raw
  footprint_topic: global_costmap/published_footprint
  robot_base_frame: base_footprint
  transform_timeout: 0.1
  smoother_plugins: ["simple_smoother"]
  simple_smoother:
    # The maximum number of iterations the smoother has to smooth the path, to bound potential computation.
    max_its: 1000
    # Whether to smooth the smoothed path recursively to refine the quality further
    do_refinement: True
    # Weight for smoother to apply to smooth out the data points
    w_smooth: 0.3
    # Weight for smoother to apply to retain original data information
    w_data: 0.2
    # Parameter tolerance change amount to terminate smoothing session
    tolerance: 1e-10
```

bt:
```xml
<Sequence name="ComputeAndSmoothPath">
  <SmoothPath unsmoothed_path="{path}" smoothed_path="{smooth_path}"/>
</Sequence>
```

PR #3308, opened 2022-12-02
Make mapUpdateLoop() indicator variables to be thread-safe

AlexeyMerzlyakov:
This is a elaboration of https://github.com/ros-planning/navigation2/pull/3307/files#r1038041946 comment for #3307.
Costmap2DROS contains one `std::thread` parallel execution, which calls `mapUpdateLoop()` routine.

As stated in C++11 standard [docs](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf) in a "5.1.2.4 Multi-threaded executions and data races"

> The execution of a program contains a data race if it contains two conflicting actions in
> different threads, at least one of which is not atomic, and neither happens before the
> other. Any such data race results in undefined behavior.

So, all write-write and write-read parallel operations should be protected by mutex, or `std::atomic` in case if it is the variable.

The `mapUpdateLoop()` thread contains the following boolean indicator variables appeared to be thread-unsafe:
* `stop_updates_`: being read in `Costmap2DROS::updateMap()` loop called from std::thread; and written in parallel in the `Costmap2DROS::start()` routine called from main thread.
* `initialized_`: being read in `Costmap2DROS::start()`, which belongs to main thread; and written in parallel by `Costmap2DROS::updateMap()` from std::thread.
* `stopped_`: to be thread-unsafe since #3307, according to the https://github.com/ros-planning/navigation2/pull/3307/files#r1038041946 comment

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3307 |
| Robotic platform tested on | `colcon test --packages-select nav2_costmap_2d` |

---

* Protected all bool indicator variables in `mapUpdateLoop()` thread from unsafe usage.

---

* It would be cool to check all places with the `std::thread` utilization in Nav2, and find all shared variables (bool, int, string, vectors, etc...) and make them to be thread-safe.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
@@ -38,6 +38,7 @@
 #ifndef NAV2_COSTMAP_2D__COSTMAP_2D_ROS_HPP_
 #define NAV2_COSTMAP_2D__COSTMAP_2D_ROS_HPP_
 
+#include <atomic>
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
God I hate all the stuff going on in this file. I welcome the day I can delete all this and do a rewrite

Issue #3311, opened 2022-12-05
AMCL Crashes Randomly When Using Recovery Parameters


borongyuan:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Foxy
- Version or commit hash:
  - 0.4.7
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
It's easy to reproduce this issue. Just use the getting started example [here](https://navigation.ros.org/getting_started/index.html#running-the-example). Only modify the following 3 parameters in nav2_params.yaml.
```
laser_model_type: "likelihood_field_prob"
recovery_alpha_fast: 0.1
recovery_alpha_slow: 0.001
```
Start the simulation and keep setting target points to keep TurtleBot moving (in order for AMCL to keep updating).
```
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False
```

#### Expected behavior
AMCL should never crash.

#### Actual behavior
AMCL will randomly crash after a few minutes of continuous running (usually within 10 minutes).
```
[ERROR] [amcl-6]: process has died [pid 29525, exit code -11, cmd '/opt/ros/foxy/lib/nav2_amcl/amcl --ros-args -r __node:=amcl --params-file /tmp/tmpmfwsb4zo -r /tf:=tf -r /tf_static:=tf_static'].
```

#### Additional information
The same problem exists if `laser_model_type` is set to the default `likelihood_field`. But setting it to `likelihood_field_prob` seems to have a higher probability of crashing. I'm still looking for the exact reason, but I open this issue first, maybe some of you already know it. It really bothers me when this problem occurs at the same time as the communication problem of DDS in WiFi scenarios. Fortunately, we can easily reproduce this in the basic simulation environment now. I don't know if the same problem exists in ROS1, I haven't tested this before. It stands to reason that this part of the code should have been heavily used and tested in ROS1. Since I'm using precompiled version 0.4.7, it's not clear to me if it has been fixed in subsequent updates. It might be related to #3242, because these few parameters should be covered by the search space of #3231. I will test the updated version later.

SteveMacenski:
Please recompile with debug flags and run GDB so we can see specifically where its crashing. See the tutorial https://navigation.ros.org/tutorials/docs/get_backtrace.html

borongyuan:
Thanks for your advice. I got the traceback information as below.
```
[...]

PR #3315, opened 2022-12-06
Fix Wrong Map Pointer (#3311)

borongyuan:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3311 |
| Robotic platform tested on | Gazebo Simulation of TurtleBot3, Our Own Robot) |

---

Not assigning fixed map pointer to particle filter, using latest when resample.

---

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/include/nav2_amcl/pf/pf.hpp
+++ b/nav2_amcl/include/nav2_amcl/pf/pf.hpp
@@ -129,7 +129,6 @@ typedef struct _pf_t
 
   // Function used to draw random pose samples
   pf_init_model_fn_t random_pose_fn;
-  void * random_pose_data;
 
   double dist_threshold;  // distance threshold in each axis over which the pf is considered to not
                           // be converged
@@ -141,7 +140,7 @@ typedef struct _pf_t
 pf_t * pf_alloc(
   int min_samples, int max_samples,
   double alpha_slow, double alpha_fast,
-  pf_init_model_fn_t random_pose_fn, void * random_pose_data);
+  pf_init_model_fn_t random_pose_fn);
 
 // Free an existing filter
 void pf_free(pf_t * pf);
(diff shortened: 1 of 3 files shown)


mikeferguson:
So, comparing this against ROS1, this looks like it will work - but there may be other issues lurking:

 * In ROS1 - the function handleMapMessage() resets the particle filter instance entirely after loading the map
 * In ROS2 - the function handleMapMessage() just updates the map, and does nothing else....

SteveMacenski:
Ugh. The person that migrated AMCL to ROS 2 had the best intentions of cleaning things up, but ended up creating more issues than solving problems. I'll merge in this fix for @borongyuan but this may be something that needs to be revisited in more detail if any more issues come up.

hidmic:
@SteveMacenski would you consider backporting this patch to Humble?

SteveMacenski:
It changes function signatures -- which is an ABI breaking change. Generally I wouldn't backport something but if this resolves a crash I'm more open to it. Can you confirm that you can reproduce a crash that this fully resolves (and understand why it resolves it)?

hidmic:
I can confirm, under the exact same conditions described in https://github.com/ros-planning/navigation2/issues/3311, BUT I don't have an MRE to share at hand. I can try and craft one. 

That said, the problematic execution path is more or less evident once you know where to look. The underlying particle filter is initialized [on configure](https://github.com/ros-planning/navigation2/blob/95b23227d49782ecf8ec445beffcf4995b148fe5/nav2_amcl/src/amcl_node.cpp#L243) and [upon dynamic re-parametrization](https://github.com/ros-planning/navigation2/blob/95b23227d49782ecf8ec445beffcf4995b148fe5/nav2_amcl/src/amcl_node.cpp#L1336), but the map gets clobbered and recreated every time [a new occupancy grid](https://github.com/ros-planning/navigation2/blob/95b23227d49782ecf8ec445beffcf4995b148fe5/nav2_amcl/src/amcl_node.cpp#L1396-L1397) gets published. It doesn't take much for the underlying particle filter to wind up with a non-owning reference to nothing, that doesn't become a problem unless adaptive resampling is enabled (and the particle filter tries to use that reference).

SteveMacenski:
OK. Backport in progress.

PR #3320, opened 2022-12-08
Publish layers

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


nav2_system_tests/src/error_codes/test_error_codes.py
```diff
@@ -26,6 +26,7 @@
 
 def main(argv=sys.argv[1:]):
     rclpy.init()
+    time.sleep(10)
```
SteveMacenski, line 29: ? 10 seconds is alot to add to a test. What is this waiting for? Can this be reduced or based on an event?

jwallace42: It is just waiting for the whole stack to come up. The same thing is done for the nav though poses test. 

https://github.com/ros-planning/navigation2/blob/main/nav2_system_tests/src/system/nav_through_poses_tester_node.py#L321

This should be resolved with the life cycle changes for python.

BriceRenaudeau:
There is a crash due to this PR :boom: .

on cleanup, the layer_pub are reset but not removed from the layer_publishers vector. On activate, the activation of the layer_pub crashes.

Easy fix: add vector cleaning (`layer_publishers_.clear();`) after the following lines.

https://github.com/ros-planning/navigation2/blob/778c0faee41788c9e3323a420ff2c10122ba4aae/nav2_costmap_2d/src/costmap_2d_ros.cpp#L341-L343

SteveMacenski:
https://github.com/ros-planning/navigation2/pull/3687 to resolve

PR #3332, opened 2022-12-19
[bt_navigator] plugin_lib_names default: Remove duplicate of nav2_back_up_cancel_bt_node

svenlr:
---

## Basic Info

The duplicate led to a failure during configuration stage of bt_navigator, when the parameter fallbacks to its default.   
(A ticket was not created in order to avoid unnecessary noise.)

---

- Removed the duplicate of nav2_back_up_cancel_bt_node

Nice to have:
- One could remove duplicates in the list programmatically, i.e. turn the list into a set and back to a list to avoid this kind of problems forever, and for custom configurations, too.

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #3341, opened 2022-12-27
Constrained Smoother tuning

AlexeyMerzlyakov:
During the benchmarking of Path Smoothers, it was discovered abnormal behavior of Constrained Smoother. This behavior appeared on some experiments with SmacHybridA* planner and Constrained Smoother, where the path smoothed by Constrained Smoother could have some “wobblings” or “oscillations” like depicted below:
Screenshot_2022-11-20_15-01-46

The analysis is being performed and attached to current PR in a [CS_path_oscillation_analysis_v7.pdf](https://github.com/ros-planning/navigation2/files/10309245/CS_path_oscillation_analysis_v7.pdf) file.
As the results of analysis and during the numerous tests and benchmarks under different environments (on small and large maps, narrow- and wide-spaced) it was found and proved the optimal set of the residual coefficients causing Constrained Smoother to produce the rational path without oscillation effects:

| Residual coefficient | Standard value change |
| ------ | ----------- |
| w_smooth | 15,000 → 2,000,000 |
| w_cost | 0.015 |
| w_curve | 30.0 |
| w_dist | 0.0 |

Thus, for best Constrained Smoother production experience is suggested to update default `w_smooth` value from 15K to 2M.

---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on |TB3 Gazebo simulation, Path Smoothers [benchmarking](https://github.com/ros-planning/navigation2/tree/main/tools/smoother_benchmarking), colcon tests, lcov|

---

* Updated defauly `w_smooth` value from 15K -> to 2M
* Tuned `test_constrained_smoother` for new `w_smooth` value:
  - `testingObstacleAvoidance`: cost avoidance improvement value was decreased, as expected from params ratio change
  - `testingObstacleAvoidanceNearCusps`: test seems to be tuned for the 15K w_smooth value - leaved untouched
  - `testingDownsamplingUpsampling`: slight smoothness improvement change
  - `testingStartGoalOrientations`: slight change of smoothness improvement in normal orientation, ~2x change in smoothness improvement and angles in reversed path orientation
* Fixed `dist` calculation formula in `test_constrained_smoother` causing incorrect exponent factor calculation, which causes overflows in costs making incorrect default costmap for the test (although, having no effect on final test results)
* Fixed incorrect integer ROS-parameter setting while it is required floating-point one for some cases in `test_constrained_smoother`  test. This issue leaded to `w_smooth` coefficient was not being updated in these cases.
 
* `w_smooth` default value should be updated at the CS configuration guide page, if this change to be agreed and accepted
---

* (optionally and might be not necessary) Re-work `test_constrained_smoother -> testingObstacleAvoidanceNearCusps` testcase for new `w_smooth = 2M` value

#### For Maintainers: 

### DIFF
--- a/nav2_constrained_smoother/include/nav2_constrained_smoother/options.hpp
+++ b/nav2_constrained_smoother/include/nav2_constrained_smoother/options.hpp
@@ -72,7 +72,7 @@ struct SmootherParams
       node, local_name + "w_dist", rclcpp::ParameterValue(0.0));
     node->get_parameter(local_name + "w_dist", distance_weight);
     nav2_util::declare_parameter_if_not_declared(
-      node, local_name + "w_smooth", rclcpp::ParameterValue(15000.0));
+      node, local_name + "w_smooth", rclcpp::ParameterValue(2000000.0));
     node->get_parameter(local_name + "w_smooth", smooth_weight);
     nav2_util::declare_parameter_if_not_declared(
       node, local_name + "cost_check_points", rclcpp::ParameterValue(std::vector<double>()));
--- a/nav2_constrained_smoother/README.md
+++ b/nav2_constrained_smoother/README.md
@@ -22,7 +22,7 @@ smoother_server:
       minimum_turning_radius: 0.40  # minimum turning radius the robot can perform. Can be set to 0.0 (or w_curve can be set to 0.0 with the same effect) for diff-drive/holonomic robots
       w_curve: 30.0                 # weight to enforce minimum_turning_radius
       w_dist: 0.0                   # weight to bind path to original as optional replacement for cost weight
-      w_smooth: 15000.0             # weight to maximize smoothness of path
+      w_smooth: 2000000.0           # weight to maximize smoothness of path
       w_cost: 0.015                 # weight to steer robot away from collision and cost
 
       # Parameters used to improve obstacle avoidance near cusps (forward/reverse movement changes)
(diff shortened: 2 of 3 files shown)


nav2_constrained_smoother/include/nav2_constrained_smoother/options.hpp
```diff
@@ -72,7 +72,7 @@ struct SmootherParams
       node, local_name + "w_dist", rclcpp::ParameterValue(0.0));
     node->get_parameter(local_name + "w_dist", distance_weight);
     nav2_util::declare_parameter_if_not_declared(
-      node, local_name + "w_smooth", rclcpp::ParameterValue(15000.0));
+      node, local_name + "w_smooth", rclcpp::ParameterValue(2000000.0));
```
SteveMacenski, line 75: Docs update to match?

SteveMacenski (changes requested):

nav2_constrained_smoother/include/nav2_constrained_smoother/options.hpp
AlexeyMerzlyakov: Sure, if you are OK with `w_smooth` 2M new value (and also with this change in general), I'll submit a new doc update PR.

SteveMacenski: Please do

AlexeyMerzlyakov: Done: https://github.com/ros-planning/navigation.ros.org/pull/385. Seems to be mentioned only on two places, there is no tutorials or something else needed to be updated.

Issue #3335, opened 2022-12-21
Pluginize nav2_bt_navigator


redvinaa:
#### Feature description
I'm planning to extend the bt_navigator package to be plugin based, after the [discussion on Discourse](https://discourse.ros.org/t/make-bt-navigator-extendible/28570). The navigators would be plugins, loaded into the bt_navigator node at runtime. This would make it easier to write new navigators in the case of custom behavior trees needing parameters different than the NavigateToPose or NavigateThroughPoses actions.

#### Implementation considerations
As the other plugin-based packages, the implementation would use the pluginlib library. As for the base class, it's a question to be answered. The navigator base class currently used for the navigators is actually a class template (the template type is the type of the action server that the navigator implements), which cannot be used as a base class as is.

A possible solution would be to separate the action server from the behavior tree engine, and only include the latter in the plugins' base class, although I understand the best would be to include both. @SteveMacenski do you have a better idea how to implement this?

mradams-bastian:
There might be some overlap here with [this](https://github.com/ros-planning/navigation2/pull/3068) PR

SteveMacenski:
This is the [base navigator class](https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/include/nav2_bt_navigator/navigator.hpp) 

The biggest question I would propose from looking at this is if every navigator type would be necessarily going to use our BT Action Server (https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/include/nav2_bt_navigator/navigator.hpp#L159-L167). Everything else here looks sufficiently general for use as a plugin base class. The `bt_action_server_` is the root of the BT behavior, but conceptually someone could modify that (but I would hope they'd contribute it back). It also is the main operation that we play with in the `on_*` functions to remove the boilerplate of calling the object's methods so that the implemented plugin can focus only on what it needs to configure/activate/deactivate/cleanup. 

I think this is good as-is, but that's a point for discussion.

redvinaa:
> There might be some overlap here with https://github.com/ros-planning/navigation2/pull/3068 PR

That PR solved the above problem by creating a virtually empty `nav2_core::Navigator` class as a base class and that doesn't enforce the plugins to be BT related at all.

> The biggest question I would propose from looking at this is if every navigator type would be necessarily going to use our BT Action Server

[...]

PR #3345, opened 2023-01-03
Pluginizing BT Navigators

SteveMacenski:
Closing https://github.com/ros-planning/navigation2/issues/3335

Creates a new NavigatorBase class for which is used ONLY to create navigator templated plugin vectors for the BT Navigator server. Its methods are implemented as `final` in the `BehaviorTreeNavigator` plugin class, which is the user-facing class to implement their BT Action Navigator types. The virtual methods within that class are the core to creating new navigator behaviors.

APIs to implement for plugins:
- getName
- getDefaultBTFilepath
- goalReceived
- onLoop
- onPreempt
- goalCompleted
- configure
- cleanup
- activate
- deactivate

### DIFF
--- a/nav2_bringup/params/nav2_multirobot_params_1.yaml
+++ b/nav2_bringup/params/nav2_multirobot_params_1.yaml
@@ -45,6 +45,11 @@ bt_navigator:
     odom_topic: /odom
     bt_loop_duration: 10
     default_server_timeout: 20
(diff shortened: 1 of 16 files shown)


SteveMacenski:
https://github.com/ros-planning/navigation.ros.org/pull/380 docs PR with tutorial

PR #3350, opened 2023-01-05
Adding new MPPI controller to Nav2

SteveMacenski:
Adds in the New MPPI Controller that was the result of a long-term collaboration between @artofnothingness and myself in 2022 in https://github.com/artofnothingness/mppic. 

Remaining action items are described in https://github.com/artofnothingness/mppic/issues/107 as:

- [x] Rename for Nav2 naming conventions (`nav2_mppi_controller`)
- [x] Add to metapackage list
- [x] Add documentation page on website https://github.com/ros-planning/navigation.ros.org/pull/378
- [x] Add to plugins list https://github.com/ros-planning/navigation.ros.org/pull/378
- [x] Add to migration guide https://github.com/ros-planning/navigation.ros.org/pull/378
- [x] Transfer tickets: Log-Normal distribution + Zero-mean + SMPPI + Acceleration characteristics https://github.com/ros-planning/navigation2/issues/3351
- [x] CI turning over with timeouts
- [x] Feedback from 2-3 users happy with performance on hardware  thanks to Neobotix @padhupradheep and Dexory @doisyg !
- [x] Update exceptions to use controller exception set
- [x] Add readme note that the path critics are not necessary if you want more "free" behavior to do whatever for just a goal pose

MPPI docs merged already and live: https://github.com/ros-planning/navigation.ros.org/pull/378


SteveMacenski:
@ruffsl as you're looking into this I think the following packages should be leaves:
- RPP Controller
- MPPI Controller (this)
- Smac Planner
- Collision Monitor
- Constrained Smoother
- NavFn Planner
- Rotation Shim Controller
- Theta* planner
- Velocity Smoother

Basically all of the plugins + the things that sit below the stack. Of these, the MPPI & Smac are the heaviest, at 8 minutes each (followed by Collision Monitor at 3 minutes, Rotation Shim / Theta* / RPP / Constrained Smoother / NavFn at a minute each)

ruffsl:
I also updated the makeflags to increase the number of allowed build and linker jobs to match the core count. Didn't seem to make much of a difference (saved about a minute) as compared to just increasing the CI resource instance class from medium to large. But this looks to be a simple enough stopgap for now.

I'll revisit GitHub actions again while I experimenting with GitHub codespaces. Perhaps some of caching ergonomics have since improved.

SteveMacenski:
> I also updated the makeflags to increase the number of allowed build and linker jobs to match the core count. Didn't seem to make much of a difference (saved about a minute) as compared to just increasing the CI resource instance class from medium to large. But this looks to be a simple enough stopgap for now.

It seems like the `parallel-workers = 2` in the defaults is what controlled the number of threads for building. I did look at the MAKEFLAGS but I don't think anything uses makefile so I didn't bother :shrug: 

> I'll revisit GitHub actions again while I experimenting with GitHub codespaces. Perhaps some of caching ergonomics have since improved.

:+1: We also have the additional space from the upped tier as well

SteveMacenski:
> 89.94%

Project is so close :cry:

ruffsl:
You could probably revert my change, as it only really seems to thrash the max CPU with context switching between cores:

chrome_screenshot_1673201852411
chrome_screenshot_1673201807517

Although, I wonder why the system tests package build is being pegged to only one CPU core? Is that package's internal compilation really so sequentially linear that it isn't being allocated any multiple parallel jobs/cores for its entirety?

Is this a similar build resource usage pattern you can observe when building locally on your workstation?

SteveMacenski:
> I wonder why the system tests package build is being pegged to only one CPU core? Is this a similar build resource usage pattern you can observe when building locally on your workstation?

I do see that locally too with `colcon build --packages-select nav2_system_tests` on my local machine. Doesn't appear to be a Circle/VM issue, that's a colcon or the cmake being invoked

tonynajjar:
https://github.com/ros-planning/navigation.ros.org/pull/397

PR #3356, opened 2023-01-10
Do only consider enabled costmap plugins/filters in the isCurrent() method of a LayeredCostmap

### LINKED ISSUE #3353: LayeredCostmap isCurrent() does not consider whether the plugins are enabled or not
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble (apt binaries)
- Version or commit hash:
  - 1.1.3-1jammy.20221207.081347
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
 * configure nav2 with the `InflationLayer` and `ObstacleLayer` plugins configured for the local costmap; the `InflationLayer` is enabled, but the `ObstacleLayer` is disabled
 * implement a "go to pose" behavior tree, executed so that a `follow_path` is executed after having cleared the local costmap (e.g. after the execution of a recovery)

#### Expected behavior
nav2's controller server properly follows the computed path.

#### Actual behavior
nav2's controller server gets stuck in a endless loop after `Received a goal, begin computing control effort.`.

#### Additional information
The endless loop is [here](https://github.com/ros-planning/navigation2/blob/humble/nav2_controller/src/controller_server.cpp#L389).

After clearing the local costmap, calling the `reset()` of the different layered_costmap plugins, the `current_` attribute of the `ObstacleLayer` plugin is set to `False`,  so that the `isCurrent()` of the plugins returns `False`, and the `LayeredCostmap`'s `isCurrent()` returns `False` as well, even is the plugin is disabled.
`ObstacleLayer`'s `current_` is never set back to `True`, as normally done in the the `updateCosts()`, that is ["bypassed"](https://github.com/ros-planning/navigation2/blob/humble/nav2_costmap_2d/plugins/obstacle_layer.cpp#L529) when the plugin is not enabled.

#### Feature description
The `isCurrent()` of the `LayeredCostmap` class should take into account whether the plugins are enabled or not.

Add a `isEnabled()` method to the `Layer` class, and modify
```
current_ = current_ && (*plugin)->isCurrent();
```
by
```
current_ = current_ && ((*plugin)->isCurrent() || !(*plugin)->isEnabled();
```
in [`LayeredCostmap`'s `isCurrent()` implementation](https://github.com/ros-planning/navigation2/blob/humble/nav2_costmap_2d/src/layered_costmap.cpp#L253).

Or simply replace [`Layer`'s `isCurrent()` implementation](https://github.com/ros-planning/navigation2/blob/humble/nav2_costmap_2d/include/nav2_costmap_2d/layer.hpp#L137) by
```
    return current_ || !enabled_;
```

milidam:
When clearing entirely a costmap (see the `clear_entirely_*` service), its layers are reset. This sets their `current_` attribute to False, until they are updated (`updateCosts()`). However, for disabled plugins/filters, the `updateCosts()` is bypassed so that the `current_` attribute is never reset to True, and the costmap never becomes "current" again; because of that the controller_server used to get stuck in an endless loop (see `while (!costmap_ros_->isCurrent())` in `ControllerServer::computeControl()`).

This patch fixes that by adding a condition for not considering disabled plugins/filters in the `LayeredCostmap::isCurrent()`.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3353  |
| Primary OS tested on |  |

---

- fix #3353

N/A

---

N/A

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/layer.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/layer.hpp
@@ -139,6 +139,12 @@ public:
     return current_;
   }
 
+  /**@brief Gets whether the layer is enabled. */
+  bool isEnabled() const
+  {
+    return enabled_;
+  }
+
   /** @brief Convenience function for layered_costmap_->getFootprint(). */
   const std::vector<geometry_msgs::msg::Point> & getFootprint() const;
 
--- a/nav2_costmap_2d/src/layered_costmap.cpp
+++ b/nav2_costmap_2d/src/layered_costmap.cpp
@@ -256,12 +256,12 @@ bool LayeredCostmap::isCurrent()
   for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin();
     plugin != plugins_.end(); ++plugin)
   {
-    current_ = current_ && (*plugin)->isCurrent();
+    current_ = current_ && ((*plugin)->isCurrent() || !(*plugin)->isEnabled());
   }
   for (vector<std::shared_ptr<Layer>>::iterator filter = filters_.begin();
     filter != filters_.end(); ++filter)
   {
-    current_ = current_ && (*filter)->isCurrent();
+    current_ = current_ && ((*filter)->isCurrent() || !(*filter)->isEnabled());
   }
   return current_;
 }


SteveMacenski (approved):

SteveMacenski:
> /opt/overlay_ws/src/navigation2/nav2_costmap_2d/src/layered_costmap.cpp:259:67: error: ‘using element_type = class nav2_costmap_2d::Layer’ {aka ‘class nav2_costmap_2d::Layer’} has no member named ‘isEnabled’

This does not build

milidam:
@SteveMacenski thanks for merging the commit.
Would it be possible to backport it to the humble branch as well?

SteveMacenski:
Yes, this will be backported in a few weeks when I run a new sync

PR #3360, opened 2023-01-12
Fix nav2_velocity_smoother for sim time

HovorunB:
---

## Basic Info

| Info | Fix nav2_velocity_smoother for sim time |
| ------ | ----------- |
| Primary OS tested on | Ubuntu  |

---

If sim time is used: `now()` might be close to 0 on startup and `last_command_time_` is defined 0 making the condition `now() - last_command_time_ > velocity_timeout_` False, which assumes that velocity command was received recently and then leads to an error when trying to access command_'s properties ( `command_->linear.x` )
So i added check for command_ being null before proceeding with smoothing

---

#### For Maintainers: 

### DIFF
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -210,7 +210,7 @@ double VelocitySmoother::applyConstraints(
 void VelocitySmoother::smootherTimer()
 {
   // Wait until the first command is received
-  if (command_==nullptr) {
+  if (!command_) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


nav2_velocity_smoother/src/velocity_smoother.cpp
```diff
@@ -209,6 +209,11 @@ double VelocitySmoother::applyConstraints(
 
 void VelocitySmoother::smootherTimer()
 {
+  // Wait until the first command is received
+  if (command_==nullptr) {
```
SteveMacenski, line 213:
```suggestion
  if (!command_) {
```

SteveMacenski: Fixes both a linting issue and a style issue

HovorunB: Thank you!

SteveMacenski (approved):

AlexeyMerzlyakov (approved):
The change LGTM: the main logic is comes from `smootherTimer()` callback, so we need to wait until first message to be received.

PR #3372, opened 2023-01-17
export test files

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

In order to have users write tests for bt nodes in order modules, the test utility files must be exported. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/CMakeLists.txt
+++ b/nav2_behavior_tree/CMakeLists.txt
@@ -220,6 +220,14 @@ install(DIRECTORY include/
   DESTINATION include/
 )
 
+install(FILES test/test_action_server.hpp
+              test/test_behavior_tree_fixture.hpp
+              test/test_dummy_tree_node.hpp
+              test/test_service.hpp
+              test/test_transform_handler.hpp
+  DESTINATION test
+)
+
 install(FILES nav2_tree_nodes.xml DESTINATION share/${PROJECT_NAME})
 
 if(BUILD_TESTING)
@@ -231,7 +239,9 @@ endif()
 
 ament_export_include_directories(
   include
+  test
 )
+
 
 ament_export_libraries(
   ${library_name}


nav2_behavior_tree/CMakeLists.txt
```diff
@@ -222,6 +222,14 @@ install(DIRECTORY include/
 
 install(FILES nav2_tree_nodes.xml DESTINATION share/${PROJECT_NAME})
 
+install(FILES test/test_action_server.hpp
```
SteveMacenski, line 225: Why is this not in the test directory's cmakelist? Also, you could just use `include_directories()` for the `test/` to just include it all

SteveMacenski (changes requested):

nav2_behavior_tree/test/plugins/decorator/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+include_directories(../../)
```
SteveMacenski, line 1: ... this is not good

nav2_behavior_tree/test/plugins/control/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+include_directories(../../)
```
SteveMacenski, line 1: Why not in the root test cmakelists file before the others here are called so they're included?

nav2_behavior_tree/test/CMakeLists.txt
```diff
@@ -5,3 +5,8 @@ add_subdirectory(plugins/condition)
 add_subdirectory(plugins/decorator)
 add_subdirectory(plugins/control)
 add_subdirectory(plugins/action)
+
```
SteveMacenski, line 8: Would not including the directories here be better before adding the subdirectories?

SteveMacenski (approved):

SteveMacenski:
Woof. Lots of linting issues this created

PR #3374, opened 2023-01-19
nav2_dwb_controller: add forward_prune_distance parameter

DylanDeCoeyer-Quimesis:
Until now, the prune_distance was used as distance threshold to shorten the upcoming path when shorten_transformed_plan was enabled. However, the prune and shortening mechanisms are de-correlated mechanisms. One could wish to use a different shortening distance for upcoming points, than the prune distance used for passed points. For this reason, a new parameter "forward_prune_distance" was added.

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_dwb_controller/dwb_core/include/dwb_core/dwb_local_planner.hpp
+++ b/nav2_dwb_controller/dwb_core/include/dwb_core/dwb_local_planner.hpp
@@ -201,6 +201,7 @@ protected:
   bool debug_trajectory_details_;
   rclcpp::Duration transform_tolerance_{0, 0};
   bool shorten_transformed_plan_;
+  double forward_prune_distance_;
 
   /**
    * @brief try to resolve a possibly shortened critic name with the default namespaces and the suffix "Critic"
--- a/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
+++ b/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
@@ -91,6 +91,9 @@ void DWBLocalPlanner::configure(
     node, dwb_plugin_name_ + ".prune_distance",
     rclcpp::ParameterValue(2.0));
   declare_parameter_if_not_declared(
+    node, dwb_plugin_name_ + ".forward_prune_distance",
+    rclcpp::ParameterValue(2.0));
+  declare_parameter_if_not_declared(
     node, dwb_plugin_name_ + ".debug_trajectory_details",
     rclcpp::ParameterValue(false));
   declare_parameter_if_not_declared(
@@ -115,6 +118,7 @@ void DWBLocalPlanner::configure(
 
   node->get_parameter(dwb_plugin_name_ + ".prune_plan", prune_plan_);
   node->get_parameter(dwb_plugin_name_ + ".prune_distance", prune_distance_);
+  node->get_parameter(dwb_plugin_name_ + ".forward_prune_distance", forward_prune_distance_);
   node->get_parameter(dwb_plugin_name_ + ".debug_trajectory_details", debug_trajectory_details_);
   node->get_parameter(dwb_plugin_name_ + ".trajectory_generator_name", traj_generator_name);
   node->get_parameter(
@@ -474,7 +478,6 @@ DWBLocalPlanner::transformGlobalPlan(
   double dist_threshold = std::max(costmap->getSizeInCellsX(), costmap->getSizeInCellsY()) *
     costmap->getResolution() / 2.0;
 
-
   // If prune_plan is enabled (it is by default) then we want to restrict the
   // plan to distances within that range as well.
   double prune_dist = prune_distance_;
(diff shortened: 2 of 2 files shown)


SteveMacenski:
To make sure I understand, this is basically just separating the prune in reverse parameter separate from the prune in the future?

nav2_dwb_controller/dwb_core/include/dwb_core/dwb_local_planner.hpp
```diff
@@ -201,6 +201,7 @@ class DWBLocalPlanner : public nav2_core::Controller
   bool debug_trajectory_details_;
   rclcpp::Duration transform_tolerance_{0, 0};
   bool shorten_transformed_plan_;
+  double shorten_distance_;
```
SteveMacenski, line 204: New parameters must be added to our docs in the configuration page with a description and added to the example XML at the bottom (https://github.com/ros-planning/navigation.ros.org). Additionally, need a migration guide entry about the change for the new distributions

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -115,6 +118,7 @@ void DWBLocalPlanner::configure(
 
   node->get_parameter(dwb_plugin_name_ + ".prune_plan", prune_plan_);
   node->get_parameter(dwb_plugin_name_ + ".prune_distance", prune_distance_);
+  node->get_parameter(dwb_plugin_name_ + ".shorten_distance", shorten_distance_);
```
SteveMacenski, line 121: While I don't have an issue with the concept, the name is a little non-descript. Any better options?

nav2_system_tests/src/costmap_filters/keepout_params.yaml
```diff
@@ -119,6 +119,7 @@ controller_server:
       plugin: "dwb_core::DWBLocalPlanner"
       debug_trajectory_details: True
       prune_distance: 1.0
+      shorten_distance: 2.0
```
SteveMacenski, line 122: This param needs to be added to our default param files in the nav2_bringup/params directory

SteveMacenski (approved):

SteveMacenski:
Any updates?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
DylanDeCoeyer-Quimesis: I took inspiration from the related "shorten_transformed_plan" and "prune_distance" parameters. Any suggestion? future_plan_shortening_distance?

nav2_system_tests/src/costmap_filters/keepout_params.yaml
DylanDeCoeyer-Quimesis: Ok, note that `prune_plan`, `prune_distance`, and `shorten_transformed_plan` are currently not set in those configuration files. Do you confirm that I should add this new parameter?

SteveMacenski: Why was it added here?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
SteveMacenski: `forward_prune_distance` perhaps

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
```diff
@@ -89,6 +89,9 @@ void DWBLocalPlanner::configure(
     rclcpp::ParameterValue(true));
   declare_parameter_if_not_declared(
     node, dwb_plugin_name_ + ".prune_distance",
+    rclcpp::ParameterValue(1.0));
```
SteveMacenski, line 92: Make sure the behavior is the same as before, so the prune / forward prune distance default parameters are the same as before

DylanDeCoeyer-Quimesis: Ok, done.

nav2_system_tests/src/costmap_filters/keepout_params.yaml
DylanDeCoeyer-Quimesis: Because `prune_distance` is set here to a non-default value. So should be `forward_prune_distance` to keep the same behavior. Am I wrong?

nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp
DylanDeCoeyer-Quimesis: Ok, done.

nav2_system_tests/src/costmap_filters/keepout_params.yaml
SteveMacenski: Looking at the diff, it doesn't look like there was ever a line in that file for `prune_distance`, it appears on my reading of this that you added a parameter that wasn't previously there.

SteveMacenski (approved):

SteveMacenski:
The last thing are described in the 2 open things:
- Documentation updates for parameters in DWB so that people know the parameter exists + migration guide quick entry about the change  
- Remove the excess parameter (and matching default) from param file

SteveMacenski:
I'd love to get this in this week if possible!

nav2_system_tests/src/costmap_filters/keepout_params.yaml
DylanDeCoeyer-Quimesis: Ok as you wish. Then the default value of `forward_prune_distance` will be used and it won't match the value of `prune_dist`.

nav2_dwb_controller/dwb_core/include/dwb_core/dwb_local_planner.hpp
DylanDeCoeyer-Quimesis: Done in https://github.com/ros-planning/navigation.ros.org/pull/389

Issue #3271, opened 2022-11-07
Codecov metrics for Assisted Teleop / Drive on Heading behaviors not being picked up 


SteveMacenski:
CC @jwallace42 

This is especially odd since the other plugins are getting their metrics reported correctly (wait, spin, backup) and backup uses alot of drive on heading as part of its code. I can't seem to find any reason for this after a couple of hours of initial debugging.

padhupradheep:
Just a small hunch, I think by default `assisted_teleop` is not included as part of the default behavior tree nor in any of the BT's that we have. Could that be a cause ? :thinking:

SteveMacenski:
No, the system tests call the behavior directly

SteveMacenski:
I was playing around this afternoon in https://github.com/ros-planning/navigation2/pull/3305. Forward declaring template class did not help (I suspected something odd was happening with the template type patterns), nor did matching the public/private functions + destructor file location (not that I thought that would help, but it was one of the very things different about the 2 failing plugins from the others). 

The files look to me to be exactly the same otherwise. As well with the system tests, only difference I can find is drive on heading uses `goal.time_allowance.sec = test_params.time_allowance;` in test cases. I'm stumped at the moment. I notice the tests aren't failing cleanly which could potentially be part of it? But doesn't explain why some work and some don't, they all appear to have that issue. Next up to diagnose the ghost in the system is to start renaming stuff and moving files around to see if I can isolate it to something in cmake / linking potentially (e.g. swap the test code or plugin code to see which it follows - issue is in system tests or in the code itself // swap order of build / link to see if its related to something in CI with build order and caching).

mmattb:
Thanks to Steve for pointing this issue out to me on the discussion forums. I'd love to help out, since I've touched code coverage mechanics before in other contexts. Can somebody point me to some materials I can use to get bootstrapped on this? e.g. what package and what tests are involved, how do I install and run them, what does your workflow for gathering the code cov info look like?

SteveMacenski:
Hi,

Oh my god, what a savior! Let me start with some context and resources, then I'll answer your specifics.

[...]

Issue #3364, opened 2023-01-14
Build error: undefined reference to `rclcpp::PublisherBase::PublisherBase


AndyZe:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22
- ROS2 Version:
  - Rolling, building from source
- Version or commit hash:
  - latest commit on main (c841417)
- DDS implementation:
  - Cyclone

When building from source, I get these errors:

```
#0 71.22 --- stderr: nav2_util
#0 71.22 /usr/bin/ld: /opt/ros/rolling/lib/libbondcpp.so: undefined reference to `rclcpp::PublisherBase::PublisherBase(rclcpp::node_interfaces::NodeBaseInterface*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rosidl_message_type_support_t const&, rcl_publisher_options_s const&, rclcpp::PublisherEventCallbacks const&, bool)'
#0 71.22 /usr/bin/ld: /opt/ros/rolling/lib/libbondcpp.so: undefined reference to `rclcpp::SubscriptionBase::SubscriptionBase(rclcpp::node_interfaces::NodeBaseInterface*, rosidl_message_type_support_t const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rcl_subscription_options_s const&, rclcpp::SubscriptionEventCallbacks const&, bool, bool)'
#0 71.22 /usr/bin/ld: /opt/ros/rolling/lib/libbondcpp.so: undefined reference to `rclcpp::Clock::now() const'
#0 71.22 collect2: error: ld returned 1 exit status
```

I think the docker commands I'm using to build are pretty straight forward:

```
FROM ros:rolling

SHELL [ "/bin/bash", "-c" ]

# Install dependencies
# Note the ros-rolling-gazebo-ros* packages are needed here as they are 
# not captured by the package dependencies when running rosdep install.
RUN apt-get update && \
    apt-get install --no-install-recommends -y \
        bash-completion \
        openssh-client \
        python3-pip \
        ros-rolling-gazebo-ros*

# Set up CycloneDDS as ROS 2 middleware
RUN apt-get update && \
    apt-get install --no-install-recommends -y \
        ros-rolling-rmw-cyclonedds-cpp
ENV RMW_IMPLEMENTATION=rmw_cyclonedds_cpp

# Create Underlay workspace
ARG UNDERLAY_WS=/opt/underlay_ws
[...]

PR #3377, opened 2023-01-20
Address code cov loss due to unclean shutdown (issue #3271)

mmattb:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | 3271 |
| Robotic platform tested on | nav2_system_tests (i.e. turtlebot3_waffle?) |

---

In brief: unclean shutdown of several processes resulted in code coverage data being dropped on the floor, rather than flushed to the .gcda files. So to get the code coverage data, we fix the unclean shutdowns. There are a few instances fixed in this PR, but likely not all (see Future work below).

The gist of all these fixes is that rclcpp::shutdown() invalidates a bunch of state, and if something attempts to touch that state again after the shutdown(), exceptions and seg faults can occur, resulting in unclean shutdown. rclcpp provides a preshutdown callback mechanism for use cases like these. The fixes are:

* Add a preshutdown callback to `lifecycle_node` which tosses `bond_`. If a lifecycle_node is destructed after `shutdown()` while the `bond_` still exists, the `Bond` destructor will fire, which touches invalidated state. The result is an RCLError in the destructor code path. The fix is to drop the `bond_` prior to shutdown, i.e. in the preshutdown callback.
* Subclass the preshutdown callback in planner_server in order to drop additional state, for the same reason.
* Address a segfault in `nav2_costmap_2d_cloud main()` by manual destructing some global state prior to shutdown(). This one could not be reproduced on @SteveMacenski's machine, so it may be version dependent? Not sure.

The first two of these appeared in the circleCI output during code cov runs, as well as in my local sandbox, and fixing them drastically improved code coverage on the Nav2 plugins. I have not spotted the 3rd of these in the circleCI logs (though I haven't tried very hard), but it's low enough risk to justify, reasoning that this is probably version dependent somehow.

None, unless we want to give an example of a preshutdown callback?

---

* Additional shutdown bugs are likely lurking, but they seem hard to pin down, and it's unclear what is due to this developer's local setup.
* We should probably change our nav2_system_tests to fail if unexpected exit codes are encountered, in order to catch the issues in the prior bullet point.
* Because of the shutdown issues or other reasons, we may still be missing code coverage somewhere. Additional work could be done to find where. This present fix addresses Assisted Teleop and Drive-on-Heading coverage issues, and likely many others, but no effort was made to be systematic about finding all such issues.

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -19,6 +19,7 @@
 #include <utility>
 #include <limits>
 
+#include "lifecycle_msgs/msg/state.hpp"
 #include "nav2_core/controller_exceptions.hpp"
 #include "nav_2d_utils/conversions.hpp"
 #include "nav_2d_utils/tf_help.hpp"
@@ -246,7 +247,19 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
   for (it = controllers_.begin(); it != controllers_.end(); ++it) {
     it->second->deactivate();
   }
-  costmap_ros_->deactivate();
+
+  /*
+   * The costmap is also a lifecycle node, so it may have already fired on_deactivate
+   * via rcl preshutdown cb. Despite the rclcpp docs saying on_shutdown callbacks fire
+   * in the order added, the preshutdown callbacks clearly don't per se, due to using an
+   * unordered_set iteration. Once this issue is resolved, we can maybe make a stronger
+   * ordering assumption: https://github.com/ros2/rclcpp/issues/2096
+   */
+  if (costmap_ros_->get_current_state().id() ==
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+  {
+    costmap_ros_->deactivate();
+  }
 
   publishZeroVelocity();
   vel_publisher_->on_deactivate();
@@ -271,11 +284,16 @@ ControllerServer::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
   controllers_.clear();
 
   goal_checkers_.clear();
-  costmap_ros_->cleanup();
+  if (costmap_ros_->get_current_state().id() ==
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+  {
+    costmap_ros_->cleanup();
+  }
 
   // Release any allocated resources
   action_server_.reset();
   odom_sub_.reset();
+  costmap_thread_.reset();
   vel_publisher_.reset();
   speed_limit_sub_.reset();
 
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -299,18 +299,21 @@ Costmap2DROS::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
 
   dyn_params_handler.reset();
 
+  stop();
+
+  // Map thread stuff
+  map_update_thread_shutdown_ = true;
+
+  if (map_update_thread_->joinable()) {
+    map_update_thread_->join();
+  }
+
   footprint_pub_->on_deactivate();
   costmap_publisher_->on_deactivate();
 
   for (auto & layer_pub : layer_publishers_) {
     layer_pub->on_deactivate();
   }
-
-  stop();
-
-  // Map thread stuff
-  map_update_thread_shutdown_ = true;
-  map_update_thread_->join();
 
   return nav2_util::CallbackReturn::SUCCESS;
 }
@@ -575,18 +578,23 @@ void
 Costmap2DROS::stop()
 {
   stop_updates_ = true;
-  std::vector<std::shared_ptr<Layer>> * plugins = layered_costmap_->getPlugins();
-  std::vector<std::shared_ptr<Layer>> * filters = layered_costmap_->getFilters();
-  // unsubscribe from topics
-  for (std::vector<std::shared_ptr<Layer>>::iterator plugin = plugins->begin();
-    plugin != plugins->end(); ++plugin)
-  {
-    (*plugin)->deactivate();
-  }
-  for (std::vector<std::shared_ptr<Layer>>::iterator filter = filters->begin();
-    filter != filters->end(); ++filter)
-  {
-    (*filter)->deactivate();
+
+  // layered_costmap_ is set only if on_configure has been called
+  if (layered_costmap_) {
+    std::vector<std::shared_ptr<Layer>> * plugins = layered_costmap_->getPlugins();
+    std::vector<std::shared_ptr<Layer>> * filters = layered_costmap_->getFilters();
+
+    // unsubscribe from topics
+    for (std::vector<std::shared_ptr<Layer>>::iterator plugin = plugins->begin();
+      plugin != plugins->end(); ++plugin)
+    {
+      (*plugin)->deactivate();
+    }
+    for (std::vector<std::shared_ptr<Layer>>::iterator filter = filters->begin();
+      filter != filters->end(); ++filter)
+    {
+      (*filter)->deactivate();
+    }
   }
   initialized_ = false;
   stopped_ = true;
--- a/nav2_util/src/lifecycle_node.cpp
+++ b/nav2_util/src/lifecycle_node.cpp
@@ -36,17 +36,20 @@ LifecycleNode::LifecycleNode(
       bond::msg::Constants::DISABLE_HEARTBEAT_TIMEOUT_PARAM, true));
 
   printLifecycleNodeNotification();
+
+  register_rcl_preshutdown_callback();
 }
 
 LifecycleNode::~LifecycleNode()
 {
   RCLCPP_INFO(get_logger(), "Destroying");
-  // In case this lifecycle node wasn't properly shut down, do it here
-  if (get_current_state().id() ==
-    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
-  {
-    on_deactivate(get_current_state());
-    on_cleanup(get_current_state());
+
+  runCleanups();
+
+  if (rcl_preshutdown_cb_handle_) {
+    rclcpp::Context::SharedPtr context = get_node_base_interface()->get_context();
+    context->remove_pre_shutdown_callback(*(rcl_preshutdown_cb_handle_.get()));
+    rcl_preshutdown_cb_handle_.reset();
   }
 }
 
@@ -62,6 +65,47 @@ void LifecycleNode::createBond()
   bond_->setHeartbeatPeriod(0.10);
   bond_->setHeartbeatTimeout(4.0);
   bond_->start();
+}
+
+void LifecycleNode::runCleanups()
+{
+  /*
+   * In case this lifecycle node wasn't properly shut down, do it here.
+   * We will give the user some ability to clean up properly here, but it's
+   * best effort; i.e. we aren't trying to account for all possible states.
+   */
+  if (get_current_state().id() ==
+    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+  {
+    this->deactivate();
+  }
+
+  if (get_current_state().id() ==
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+  {
+    this->cleanup();
+  }
+}
+
+void LifecycleNode::on_rcl_preshutdown()
+{
+  RCLCPP_INFO(
+    get_logger(), "Running Nav2 LifecycleNode rcl preshutdown (%s)",
+    this->get_name());
+
+  runCleanups();
+
+  destroyBond();
+}
+
+void LifecycleNode::register_rcl_preshutdown_callback()
+{
+  rclcpp::Context::SharedPtr context = get_node_base_interface()->get_context();
+
+  rcl_preshutdown_cb_handle_ = std::make_unique<rclcpp::PreShutdownCallbackHandle>(
+    context->add_pre_shutdown_callback(
+      std::bind(&LifecycleNode::on_rcl_preshutdown, this))
+  );
 }
 
 void LifecycleNode::destroyBond()
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -107,7 +107,7 @@ def generate_launch_description():
     load_nodes = GroupAction(
         condition=IfCondition(PythonExpression(['not ', use_composition])),
         actions=[
-            SetParameter("use_sim_time", use_sim_time),
+            SetParameter('use_sim_time', use_sim_time),
             Node(
                 condition=LaunchConfigurationEquals('map', ''),
                 package='nav2_map_server',
@@ -165,7 +165,7 @@ def generate_launch_description():
     load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
         actions=[
-            SetParameter("use_sim_time", use_sim_time),
+            SetParameter('use_sim_time', use_sim_time),
             LoadComposableNodes(
                 target_container=container_name_full,
                 condition=LaunchConfigurationEquals('map', ''),
--- a/nav2_bringup/launch/navigation_launch.py
+++ b/nav2_bringup/launch/navigation_launch.py
@@ -108,7 +108,7 @@ def generate_launch_description():
     load_nodes = GroupAction(
         condition=IfCondition(PythonExpression(['not ', use_composition])),
         actions=[
-            SetParameter("use_sim_time", use_sim_time),
+            SetParameter('use_sim_time', use_sim_time),
             Node(
                 package='nav2_controller',
                 executable='controller_server',
@@ -193,7 +193,7 @@ def generate_launch_description():
     load_composable_nodes = GroupAction(
         condition=IfCondition(use_composition),
         actions=[
-            SetParameter("use_sim_time", use_sim_time),
+            SetParameter('use_sim_time', use_sim_time),
             LoadComposableNodes(
                 target_container=container_name_full,
                 composable_node_descriptions=[
--- a/nav2_bringup/launch/slam_launch.py
+++ b/nav2_bringup/launch/slam_launch.py
@@ -17,7 +17,7 @@ import os
 from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription, GroupAction
+from launch.actions import DeclareLaunchArgument, GroupAction, IncludeLaunchDescription
 from launch.conditions import IfCondition, UnlessCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch.substitutions import LaunchConfiguration
@@ -81,7 +81,7 @@ def generate_launch_description():
 
     start_map_server = GroupAction(
         actions=[
-            SetParameter("use_sim_time", use_sim_time),
+            SetParameter('use_sim_time', use_sim_time),
             Node(
                 package='nav2_map_server',
                 executable='map_saver_server',
--- a/nav2_costmap_2d/src/costmap_2d_cloud.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_cloud.cpp
@@ -225,6 +225,11 @@ int main(int argc, char ** argv)
     "voxel_grid", rclcpp::SystemDefaultsQoS(), voxelCallback);
 
   rclcpp::spin(g_node->get_node_base_interface());
+
+  g_node.reset();
+  pub_marked.reset();
+  pub_unknown.reset();
+
   rclcpp::shutdown();
 
   return 0;
--- a/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
+++ b/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
@@ -114,6 +114,13 @@ protected:
    */
   bool resume();
 
+  /**
+   * @brief Perform preshutdown activities before our Context is shutdown.
+   * Note that this is related to our Context's shutdown sequence, not the
+   * lifecycle node state machine or shutdown().
+   */
+  void onRclPreshutdown();
+
   // Support function for creating service clients
   /**
    * @brief Support function for creating service clients
@@ -185,6 +192,14 @@ protected:
    * @brief function to check if the Nav2 system is active
    */
   void CreateActiveDiagnostic(diagnostic_updater::DiagnosticStatusWrapper & stat);
+
+  /**
+   * Register our preshutdown callback for this Node's rcl Context.
+   * The callback fires before this Node's Context is shutdown.
+   * Note this is not directly related to the lifecycle state machine or the
+   * shutdown() instance function.
+   */
+  void registerRclPreshutdownCallback();
 
   // Timer thread to look at bond connections
   rclcpp::TimerBase::SharedPtr init_timer_;
--- a/nav2_lifecycle_manager/src/lifecycle_manager.cpp
+++ b/nav2_lifecycle_manager/src/lifecycle_manager.cpp
@@ -45,6 +45,8 @@ LifecycleManager::LifecycleManager(const rclcpp::NodeOptions & options)
   declare_parameter("bond_respawn_max_duration", 10.0);
   declare_parameter("attempt_respawn_reconnection", true);
 
+  registerRclPreshutdownCallback();
+
   node_names_ = get_parameter("node_names").as_string_array();
   get_parameter("autostart", autostart_);
   double bond_timeout_s;
@@ -376,6 +378,35 @@ LifecycleManager::destroyBondTimer()
     bond_timer_->cancel();
     bond_timer_.reset();
   }
+}
+
+void
+LifecycleManager::onRclPreshutdown()
+{
+  RCLCPP_INFO(
+    get_logger(), "Running Nav2 LifecycleManager rcl preshutdown (%s)",
+    this->get_name());
+
+  destroyBondTimer();
+
+  /*
+   * Dropping the bond map is what we really need here, but we drop the others
+   * to prevent the bond map being used. Likewise, squash the service thread.
+   */
+  service_thread_.reset();
+  node_names_.clear();
+  node_map_.clear();
+  bond_map_.clear();
+}
+
+void
+LifecycleManager::registerRclPreshutdownCallback()
+{
+  rclcpp::Context::SharedPtr context = get_node_base_interface()->get_context();
+
+  context->add_pre_shutdown_callback(
+    std::bind(&LifecycleManager::onRclPreshutdown, this)
+  );
 }
 
 void
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -25,6 +25,7 @@
 #include <utility>
 
 #include "builtin_interfaces/msg/duration.hpp"
+#include "lifecycle_msgs/msg/state.hpp"
 #include "nav2_util/costmap.hpp"
 #include "nav2_util/node_utils.hpp"
 #include "nav2_util/geometry_utils.hpp"
@@ -69,6 +70,10 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
 
 PlannerServer::~PlannerServer()
 {
+  /*
+   * Backstop ensuring this state is destroyed, even if deactivate/cleanup are
+   * never called.
+   */
   planners_.clear();
   costmap_thread_.reset();
 }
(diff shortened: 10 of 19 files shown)


mmattb:
I don't have permissions to set labels on this PR, but anyway please consider this one a WiP. I'm working out the uncrustification, and I'll take a stab at one more shutdown issue that I see in a blue moon.

mmattb:
Mixed results. Some tests failed. Are these the intermittent failures you noted in the Issue @SteveMacenski?  If not, I have some chasing to do...

SteveMacenski:
> This one could not be reproduced on @SteveMacenski's machine, so it may be version dependent? Not sure.

What version are you on? I'm using Rolling on 22.04.

> Are these the intermittent failures you noted in the Issue @SteveMacenski? If not, I have some chasing to do...

The Theta* test failure is something that popped up very recently this year. I'll get to that today hopefully to fix. The `SingleTriggerTestFixture` is caused by a regression in BT.CPP who's fix is propagating through releases already.

Looking at the codecov results from this, there are a ton of code now no longer covered - which is unusual. Its not unusual for our coverage to iterate a few tenths of a percent from run to run, but this is pretty significant with files losing ~60%+ of their coverage metrics.

mmattb:
Good to know. I think that waypoint follower blip stands out. It comes and goes. I'll investigate.

mmattb:
It looks like the waypoint follower miss does happen sometimes outside of my branch. It's not consistent, even on my branch:

https://app.codecov.io/gh/ros-planning/navigation2/pull/3374

Likewise with oscillation.cpp:

https://app.codecov.io/gh/ros-planning/navigation2/pull/3372

Clearly the fixes I have here help, but are not complete. Shall I open issues for these other ones?

mmattb:
re: costmap2d: with a fresh clone of both ros2_rolling and navigation2:main today (i.e. without my fix applied) I still hit the segfault. My best guess it that my apt-get installed libbondcpp is different? And maybe other external deps? But I'm grasping at straws since I didn't try alternate versions of anything else. I'm okay to back that one out if it seems strange to you, but luckily I think it's low risk.

SteveMacenski:
> It looks like the waypoint follower miss does happen sometimes outside of my branch. It's not consistent, even on my branch

If you're saying that run is a fluke, I'll retrigger CI and see if it goes away. I haven't noted that blip before, but if its irregular (e.g. not common, but happens on rare occasions) then that's something I can live with, at least for now. I just retriggered CI too see to.  

I'm not sure about the costmap part, I use bare metal (since Nav2 development is just my day-to-day) with binaries provided by the Rolling distribution that have been updated. Maybe as part of an apt update there was something that didn't quite sit right with your workspace cache? I know the issue reported in #3364 is caused by a workspace cache becoming invalidated after updating binaries with an ABI breaking change. 

> I'm okay to back that one out if it seems strange to you, but luckily I think it's low risk.

I don't think it hurts anyone, I'd just remove the comment

nav2_util/src/lifecycle_node.cpp
```diff
@@ -64,6 +73,24 @@ void LifecycleNode::createBond()
...
+{
+  RCLCPP_INFO(
+    get_logger(), "Running Nav2 LifecycleNode rcl preshutdown (%s)",
+    this->get_name());
+  destroyBond();
+}
+
+void LifecycleNode::register_rcl_preshutdown_callback()
+{
+  rclcpp::Context::SharedPtr context = get_node_base_interface()->get_context();
```
SteveMacenski, line 86: If something like this is truly necessary, I think we need to escalate this to rclcpp that we cannot get clean exits from component containers - this seems like a serious issue. 

Do you see this also with non-composed bringups (e.g. launching a server itself in its own process)? We have a `nav2_util` util that can help with transitioning up the nodes: https://github.com/ros-planning/navigation2/blob/main/nav2_util/src/lifecycle_bringup_commandline.cpp

Note also that https://github.com/ros2/rclcpp/pull/2085 was just merged, which should help with specifically the isolated component container shutdown. Though curious if you see this issue with any other component container type.

SteveMacenski: Also, I love the debugging and learning about new APIs I hadn't yet even played with myself in ROS 2! The list of the lower-level APIs are so massive, I've just never bothered to even look at many of them. 

Edit: As I looked for it, or here's an example the autogenerated docs don't even have :sweat:

mmattb:
> If you're saying that run is a fluke, I'll retrigger CI and see if it goes away. I haven't noted that blip before, but if its irregular (e.g. not common, but happens on rare occasions) then that's something I can live with, at least for now. I just retriggered CI too see to.

I think the waypoint follower coverage regression not a fluke, but not related to my change afaict.  Do we always run the same set of tests, and always the same way?

The new run on my PR doesn't show the issue: https://app.codecov.io/gh/ros-planning/navigation2/pull/3377

The runs on some recent unrelated CRs do:
https://app.codecov.io/gh/ros-planning/navigation2/pull/3374
https://app.codecov.io/gh/ros-planning/navigation2/pull/3371

Looks like with this merge the waypoint follower coverage drastically *increased*, from a level similar to the one seen in my PR. Seems like a moving target for some reason.
https://app.codecov.io/gh/ros-planning/navigation2/pull/3372

SteveMacenski:
For Theta*, I think those PRs need to be rebased. I fixed a bug in the Theta* tests and merged at the end of last week.

Looking at the other parts of that run, we're talking about changes of +/- 2 lines of code, which is within the margin of error for the results of coverage due to the system tests. DWB and Costmap2D are legacy packages that did not have very good unit test coverage when they were adopted and their coverage is largely on system-level testing rather than unit level function testing (unlike the rest of the packages). 

So those diffs you see aren't jitter, those are PRs that need to be rebased for the Theta* coverage to be returned. 

If you rebase your PR, you should see that go away too. Whatever that jitter is is another problem if it was touching Waypoint Follower (which hasn't substantively changed in some time now)

mmattb:
> If something like this is truly necessary, I think we need to escalate this to rclcpp that we cannot get clean exits from lifecycle nodes at all (which guarantees normal nodes can't either) - this seems like a serious issue. Something seems fishy to me though that this would be necessary
>Something doesn't seem quite right to me that this is a ROS 2-wide issue.

I think it's actually specific to Nav2's use of Bonds in its LifecycleNode child class. I don't think it affects all of ROS2. The bond I'm cleaning up in my PR is a Nav2 thing. But I agree the shutdown sequence seems needlessly difficult to get right. Know any ROS2 people we can tag or email to ask about this? I can present the stacks/etc.

I'll rebase and see if the issue repros both with/without composition.

mmattb:
> I know the issue reported in https://github.com/ros-planning/navigation2/issues/3364 is caused by a workspace cache becoming invalidated after updating binaries with an ABI breaking change.

Could be! I wish libbondcpp was brought into my ros2_rolling clone when I did a vcs import!  Maybe I need to manually build that guy and update it. That's the only stuff from the apt install I see in my stacks.

Anyway...comment removed!

mmattb:
Running with composition and without my fix we see the same RCLError that the preshutdown callback addresses. At least the tests don't time out anymore!  Running with composition and with my fix, the errors are indeed gone, and I don't see any more.

At any rate, I don't know why ROS2 raises exceptions for these various cases, so @SteveMacenski if you want to pursue a conversation with the ROS2 devs, I'd love a pointer to the right people! Maybe you can tag them here?

SteveMacenski:
> I wish libbondcpp was brought into my ros2_rolling clone when I did a vcs import!

Sorry about that, our vcs file for nav2 has all of the dependencies commented out (if you uncomment them, all should work). We have it that way since our CI uses that file to check for dependencies to build in an underlay workspace if we have temporarily un-released dependencies. That's a legacy of before ROS 2 had a build farm and binaries were available for anything (and rolling did not exist). I left the file intact with the contents commented out in case of a "break glass" moment and to keep tracking what we depend on above a base install. 

> Maybe you can tag them here?

I do, but first, did you test using the rclcpp new container updates after https://github.com/ros2/rclcpp/pull/2085? I don't want to poke them about getting involved if the issue was actually resolved ahead of us. If the issue is composition isolated, then this could be it (unless you showed it happens with other non-isolated container types). 

---

I know in this PR you went looking for the composition testing issues, but curious if there are the behavior server fixes here too? I don't see anything particular about the behavior server and the missing overage we had there - this seems more general for composition -- unless its all the same cause?

mmattb:
> Are there changes we can add to bond instead potentially? It seems odd that our destroying bond would be the cause of an issue - just resetting a shared pointer - unless bond's destruction sequence was really the problem.

I haven't spelunked the Bond code to find out. It certainly seems odd that one would be constructing something in a destructor path (a ROS2 TimerBase)! Yet that is clearly what is happening. You can see the stack I pasted in the Issue.

Having said that - the planner_server issue is *not* Bond related AFAIK, but is similar in form: invalidated ROS state is still being used after ROS shutdown. That is due to a 'planner_' thread still running. That issue looks like:

```
[1384.354782] (nav2_system_tests) StdoutLine: {'line': b"31: [planner_server-12] terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'\n"}
[1384.355242] (nav2_system_tests) StdoutLine: {'line': b'31: [planner_server-12]   what():  failed to send response: client w    ill not receive response, at /home/mbryan/Projects/ros/ros2_rolling/src/ros2/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/rmw_response.cpp:149, at /home/mbryan/Projects/ros/ros2_rolling/src/ros2/rcl/rcl/src/rcl/service.c:314\n'}
```

Instead of taking my fix, another approach is to always wrap these node types in code blocks which will exit prior to shutdown(). Example:

```
int run_planner_server()
{
  auto ps = std::make_shared<nav2_planner::PlannerServer>();
  
  ...
  
}  // ps dtor will fire here, before the return, and therefore before shutdown

int main()
{
  int exitcode {0};
  
  rclcpp::init(blah blah);
  
  exitcode = run_planner_server();
  
  rclcpp::shutdown();
  return exitcode;
}
```

The downside to that approach is that it reduces the flexibility in how these nodes are used. Example: the user will need to copy any information out of the instance which they want to keep prior to shutdown() being called, since the dtor will have fired by then. That's also a bit of a landmine waiting to be stepped on.

>I know in this PR you went looking for the composition testing issues, but curious if there are the behavior server fixes here too? I don't see anything particular about the behavior server and the missing overage we had there - this seems more general for composition -- unless its all the same cause?

Indeed it's the same cause: this issue can in theory affect everything which inherits from `nav2_util::LifecycleNode`, which `BehaviorServer` does; that's the connection. Having said that: if an instance of the inherited class is destructed prior to shutdown() you won't see the exception (see alternate solution above).

> vcs file for nav2 has all of the dependencies commented out
Good to know!

SteveMacenski:
>  It certainly seems odd that one would be constructing something in a destructor path (a ROS2 TimerBase)! Yet that is clearly what is happening.

But ideally if we cycle the lifecycle down to deactivated before shutdown, the `bond` object is reset so there shouldn't be any poitner to anything bond related when we shutdown? I understand that's an issue if we don't properly transition down the lifecycle first, but I think you said that it happens either way which I'm not clear on how that's possible. 

`planner_` thread? Edit: Oh, that's in the system tests, not the planner server itself, got it. 

> Instead of taking my fix, another approach is to always wrap these node types in code blocks which will exit prior to shutdown().

We don't get that choice in components :frowning_face: 

> Indeed it's the same cause

Got it!

mmattb:
> But ideally if we cycle the lifecycle down to deactivated before shutdown, the bond object is reset so there shouldn't be any pointer to anything bond related when we shutdown?

Maybe there is a better way to shut the node down than I'm doing, but we still need something to trigger that shutdown. Right now it's being triggered by the signal handler. ROS controls the signal handler, and so we need to figure out how to trigger our shutdown sequence in the path of that signal handler. ROS provides the preshutdown callback for that. That is: the order of events, which are all controlled by ROS, are: signal received -> preshutdown callbacks -> rclcpp::shutdown() -> shutdown callbacks -> exit.

So: we can ask "What should we do in the preshutdown callback?" The PR takes the quickest exit to the door, which is just to tear down the aspects of the instance's state which we trip on in shutdown.

But! I'm not sure what would be cleanest here, due to lack of experience. Should we be going through the lifecycle transition instead, e.g. calling `on_deactivate()` and `on_cleanup()`?  That's possible to do instead. That solution could maybe look like:

* Keep the preshutdown callback
* Change the preshutdown callback to call on_deactivate() and on_cleanup()    (or whatever the right way to transition is)
* Preshutdown callback continues to call destroyBond() just in case the child class didn't; that's to avoid this becoming a land mine for users.
* PlannerServer's on_cleanup already resets `planner_`, so we can drop the preshutdown callback from PlannerServer if we call on_cleanup from the super class.

Would that be preferable? If so, what is the right way to trigger those transitions? Just calling the callbacks? It would maybe give a cleaner way for the child classes to clean up nicely, if they need to?

> planner_ thread? 

Sorry I misremembered/typo'd. I mean the `planners_` and `costmap_thread_` here, which now I'm destroying in the callback: https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L72

mmattb:
Let me know if a quick video chat helps :-D  I'm in Indochina / Thailand time right now :-D

SteveMacenski:
> Maybe there is a better way to shut the node down than I'm doing, but we still need something to trigger that shutdown. Right now it's being triggered by the signal handler

Ah, I see. Some explanation of lifecycle might be good.  Lifecycle nodes have the state transitions for various things (e.g. configuring, activating, deactivating, cleaning up). When you use the autostart feature of Nav2, it will automatically transition the nodes up from the unconfigured state to the active state to process data in deterministic order. There's a complimentary shutdown you can do as well which will bring the system from active -> deactivate -> inactive -> cleanup -> unconfigured state. The deactive function you can see in the lifecycle node implementations have the destroy bond function which resets the pointer. If you use the lifecycle manager to do this, then the manager also has its bond timers destroyed in the complementary function in reset/shutdown/etc. 

So when you're using the signal handler to kill the program, you're not using any of that, you're just killing it in the active state and no attempt is made to cycle the system down. That's a suboptimal way to kill the servers, but honestly super common so I'm all on board with making that exit cleanly as well. In this case, `on_deactivate` is not called thus the bond stuff is still around, thus the non-clean problems we've noted. If you shut things down through the lifecycle transition before hitting control+C, then `on_deactivate` should be called and then the bond stuff should all be reset to null pointers. This is where I would expect you to find no issue at all -- similiar to if you were to hit control+C on a server that you executed on the commandline without cycling the lifecycle up for the bond stuff to be setup in `on_activate` in the first place.

So... that leads us to then the question to make sure that you do indeed see on your side that when you transition down the lifecycle nodes **or** don't try to activate the server(s) at all, you don't see those bond issues (since in both situations, `bond_` should be a `nullptr`). If you see that all working out well and its only on the un-lifecycle-shutdown program exits we see the problem, that is what I would expect given the rest of the conversation we've been having. 

---

So on top of that understanding, if we find that what I say above is true and our goal is to have clean shutdowns from the active state of the lifecycle nodes (for you know... the 80% of users that I suspect do that because its easy) that's where this pre-shutdown stuff can be helpful -- especially if we can essentially hide it from the users in our `nav2_util::LifecycleNode` class without impacting the server's host code. You mention we can just call `on_deactivate` and `on_cleanup` there, which are both already virtual methods of the lifecycle node. That sounds like a perfect solution, but maybe also check the state of the lifecycle node to see if that's even necessary (e.g. if the user did properly shutdown the lifecycle nodes, then this wouldn't need to do anything). We can do that via [`get_current_state`](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp#L67-L68) to check that we're in the `finalized` or `unconfigured` states. 

If we did that, we also wouldn't need to inject any code in the planner server since the `planners_.clear()` is in the cleanup stage as well https://github.com/ros-planning/navigation2/blob/562d9ac7c25256731a937adb17dffe278b3682db/nav2_planner/src/planner_server.cpp#L240. Though, it looks like we missed the costmap thread one, so we'll have to add 1 line to `on_cleanup` to do that as well :sweat_smile: 

Myself and the Intel guys that wrote Nav2 initially with me apparently did not know that the destructors wouldn't properly cleanup resources when control+C was called as you have found. With this understanding and use of the pre-shutdown callbacks, we can just make all the destructors default since not matter what, we'll end up lifecycling the nodes down before the context is deleted. 

---

Lots of words:
- I think we have a misunderstanding on the nature of  lifecycle, try that test out and make sure you see what I think you should see
- If it works, then we can use your proposed pre-shutdown callbacks to properly handle the lifecycle transitions for the user if they call control+C without bringing the lifecycle state down. If we check the current state, we can also make sure we only call the deactivation / cleanup as properly needed in case the server isn't active yet anyway.
- That should (?) resolve the issues

mmattb:
That's great info @SteveMacenski ! Helps a bunch.  So I tried a deactivate then cleanup via the preshutdown callbacks today, and hit some deadlock deep in the publisher on_deactivate. It's been pretty tough to debug.  Specifically we hang here:

https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_ros.cpp#L302

That is, in PlannerServer (also ControllerServer), we call on_deactivate on a costmap handle, which then calls on_deactivate on the publisher at the line above. At that point we hang.

And if you comment out that line, we hang on the one after it. So it smells like there is a deadlock involving a pub on_deactivate while in the preshutdown context.  It's possible that it's just slow, not a deadlock, and we eventually hit a timeout. Anyway, so it's smelling like this approach won't work without some more work... The ROS shutdown sequence has turned out to have quite a few land mines to step on.  I'll spend more time to see if I can find the specific deadlock / slow thing, but it smells like it will be a poor use of time.

SteveMacenski:
I'm going to make an educated guess that its because the footpritn publisher is used in this thread https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_ros.cpp#L450 which isn't shutdown before the publisher is deactivated - though you didn't show me the trace so I'm just guessing. Though I don't know why that would by special otherwise.  

Seems like https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_ros.cpp#L309-L313 should be before the publisher shutdowns since that's a shared resource with the thread that needs to be shutdown first. 

The next line you mention also is used in that thread, so that would make sense to see the same problem

mmattb:
Experimenting with a fix along those lines. There are some additional checks I need to do in a few places so that the ROS2 lifecycle_node code isn't so noisy about deactivations that we shouldn't be taking. This solution leveraging deactivate and cleanup is likely to be a lot more complex, and will require some special casing around state that we aren't doing today. You can see some examples I've just added: extra 'if' statements in some places. There will be more before I'm done.

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -295,23 +295,24 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
 nav2_util::CallbackReturn
 Costmap2DROS::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
 {
-  RCLCPP_INFO(get_logger(), "Deactivating");
```
mmattb, line 298: Note to self: this was removed on accident.

mmattb:
Seems okay now? So this is my second attempt that leverages the lifecycle transitions. It seems to work. There is still an intermittent failure involving publishers doing something during the shutdown process. I haven't chased after that one because it seems we are getting good coverage reliably. How does it look?

nav2_controller/src/controller_server.cpp
```diff
@@ -246,7 +247,16 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
   for (it = controllers_.begin(); it != controllers_.end(); ++it) {
     it->second->deactivate();
   }
-  costmap_ros_->deactivate();
+
+  /*
+   * The costmap is also a lifecycle node, so it may have already fired on_deactivate
+   * via rcl preshutdown cb. Let's not make an assumption about the order these fire...
+   */
+  if (costmap_ros_->get_current_state().id() ==
```
SteveMacenski, line 255: In what situation would the costmap state differ from the server state? If none, then you can use the `state` given in this function to make that decision, no?

Also, the lifecycle_node.cpp entries are checking if the state is active to run deactivation on shutdown. So if the lifecycle manager already cycled it down, then it wouldn't trigger this -- so why the duplicate check here at all?

nav2_util/src/lifecycle_node.cpp
```diff
@@ -36,17 +36,32 @@ LifecycleNode::LifecycleNode(
...
+    this->deactivate();
+  }
+
+  if (get_current_state().id() ==
+    lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+  {
+    this->cleanup();
+  }
+
+  if (rcl_preshutdown_cb_handle_) {
```
SteveMacenski, line 60: I was curious about this line about why its required to remove the callback?

nav2_util/src/lifecycle_node.cpp
```diff
@@ -36,17 +36,32 @@ LifecycleNode::LifecycleNode(
...
   // In case this lifecycle node wasn't properly shut down, do it here
   if (get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
   {
-    on_deactivate(get_current_state());
-    on_cleanup(get_current_state());
+    this->deactivate();
+  }
+
+  if (get_current_state().id() ==
```
SteveMacenski, line 54: Shouldn't this all be handled by the pre-shutdown callback? I'm not sure why duplicated

SteveMacenski (commented):
Generally no concerns beyond duplication of some elements that I'd like to have only a single copy of for clarity if its not required

nav2_controller/src/controller_server.cpp
mmattb: Well, for a few reasons...   The first is that the rcl preshutdown callback may have fired on the costmap already, and in that case we don't want to call deactivate() on it here, since we needlessly get a log line if we do:

```
[187.382828] (nav2_costmap_2d) StdoutLine: {'line': b'22: [costmap_tests-5] [WARN] [1674969001.522910754] [rcl_lifecycle]: No transition matching 4 found for current state inactive\n'}
```

In that case we may ask why we call deactivate() on the costmap here at all, and the reason is that deactivate() can be called on the controller_server at any time, not just for shutdown. The current logic (prior to my change) is to also deactivate the costmap. My change ensures we still call it in the case that we aren't shutting down.

w.r.t. the lifecycle_node logic: same thing. We may not be shutting down here at all. This may be a deactivate transition that the user triggered themselves for some other reason.

nav2_util/src/lifecycle_node.cpp
mmattb: Same logic as above: we don't know the code path we took to get here or the order that things are happening. The node can be destructed long before ros shutdown is called. When it's destructed we need to unregister its callback since otherwise we would have a use-after-free when the callback was called later.

nav2_util/src/lifecycle_node.cpp
mmattb: See prior response. This is to prevent a use after free in the case that the callback is called after the destructor fires. At that point the instance has evaporated, so there is no 'this' to bind the call to -> seg fault. I found this out the hard way :)

nav2_controller/src/controller_server.cpp
SteveMacenski: > rcl preshutdown callback may have fired on the costmap already

Is that because we register the preshutdown callback in the constructor? I would have assumed given the focus on determinism in ROS 2 that the execution of the preshutdown callbacks would be too. It looks like the costmap would always be before the controller/planner server since it completes its constructor while the planner/controller server's constructors are still crunching along. 

If that is true which would cause this issue, if we moved the costmap construction and node thread to the controller/planner server's `on_configure`, that would make it so that deterministically, those server's pre-shutdown callbacks should be called first (and thus we know preshutdown hasn't messed with the costmap state separately from the server's callbacks). Arguably, that's where those should be anyway since memory allocation should be in configuration. 

I think from the rest of what you mention, my suggestion is just an inversion of ordering which should resolve the warning you mention + not require checking the state in the server and push all that down to the already implemented logic in the lifecycle node base class.

nav2_util/src/lifecycle_node.cpp
SteveMacenski: Got it, create a new function that does the deactivation / cleanup / bond stuff so that it can be reused both places. I guess most of my issue is just duplication (in case we need to add more later and we forget its doubled up).

nav2_util/src/lifecycle_node.cpp
mmattb: Right; somebody would need to add callback handle ptr members to the base Node class, and unregister the callbacks in the Node dtor. That would be a ROS2 request :)

nav2_util/src/lifecycle_node.cpp
mmattb: Sure.

nav2_util/src/lifecycle_node.cpp
SteveMacenski: Can you open that ticket on rclcpp?

SteveMacenski: (that someone probably ultimately to be me in like 6 months)

nav2_controller/src/controller_server.cpp
mmattb: >  I would have assumed given the focus on determinism in ROS 2 that the execution of the preshutdown callbacks would be too.

Somewhat shockingly the callbacks are stored in an unordered set, which I believe means they will be executed in some arbitrary hash order.  That means they may even run in a different order between different platforms!

https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/context.hpp#L382

https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/context.cpp#L315

The reason it's shocking is that the on_shutdown (i.e. post-shutdown) callbacks are implemented the same way, and their docs specifically say the callbacks will be executed in the order added:

https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Context.html#a64bce7cbf8b92f3e0283077f26facf9b

The preshutdown callbacks don't have docs that I have found, so technically nothing is in violation there, but it's certainly a surprise.  

In-fact I think the current code would work correctly (with the 'if' removed) if the callbacks *were* called in the order added, since super constructors run before child class constructors.

Seems like the docs are wrong? I'll float an issue to ROS2. But either way, my change needs to be this way for now...

mmattb: Filed: https://github.com/ros2/rclcpp/issues/2096

mmattb: I wish somebody would hire me to work on ROS2 + Nav2 so I could go whack a mole all this stuff!

nav2_util/src/lifecycle_node.cpp
mmattb: Sure! To me that smells like a libbondcpp bug as a first pass; i.e. why the heck are we constructing things in the destructor path? I wasn't able to CC you for some reason, but here it is:  https://github.com/ros/bond_core/issues/92

I additionally filed a bug for us to catch process deaths in the system tests, so we don't need to rely on declining code coverage to find future issues:  https://github.com/ros-planning/navigation2/issues/3391

nav2_controller/src/controller_server.cpp
SteveMacenski: That's ridiculous, but thank you. I look at that and don't really see why that isn't a vector or even just a `set` anyway for how its called. 

 If you want to work on ROS 2 core, let me know, I can connect you :wink:

nav2_util/src/lifecycle_node.cpp
SteveMacenski: Can you file the `there is no 'this' to bind the call to -> seg fault` ticket as well for the shutdown callbacks not checking if the node is even still alive? I can escalate the Bond + 2 rclcpp issues in the weekly ROS 2 core libs meeting (though looks like Michael's already seen the 2 you filed).

SteveMacenski (approved):

SteveMacenski:
Awesome work, I really appreciate your detailed look and attention here! If you can file that last ticket on the shutdown callbacks not checking the node is still in existence that would be appreciated to button up any additional issues to expose back to the core libraries.

If you're open to working on another project or fishing for some other issues you ran into during this excursion, let me know! I'm happy for any assistance you're open to and would enjoy the puzzle to contribute.

mmattb:
> Awesome work, I really appreciate your detailed look and attention here! If you can file that last ticket on the shutdown callbacks not checking the node is still in existence that would be appreciated to button up any additional issues to expose back to the core libraries.
> 
> If you're open to working on another project or fishing for some other issues you ran into during this excursion, let me know! I'm happy for any assistance you're open to and would enjoy the puzzle to contribute.

Thanks a bunch Steve. It was great to jump in on it. I have a paper to finish editing in the next month, then maybe I can jump back in (and start my job search again). Any leads on Dev positions that don't require robotics expertise, let me know:-)

SteveMacenski:
What kind of dev positions? I can see if I know anyone. They're all going to be robotics companies but most jobs at a robotics company aren't straight up robotics algorithm development.

mmattb:
> What kind of dev positions? I can see if I know anyone. They're all going to be robotics companies but most jobs at a robotics company aren't straight up robotics algorithm development.

Shot you an email @SteveMacenski !

PR #3378, opened 2023-01-20
Fix SpeedController

borongyuan:
* Using blackboard get() method that does not throw error
* Share odom_smoother on the blackboard

Signed-off-by: Borong Yuan <yuanborong@hotmail.com>

---

## Basic Info

| ------ | ----------- |

---

The SpeedController is broken since (#2271).
There are two get() methods in blackboard. One of them throws an error if `val` doesn't exist. Obviously `goal` and `goals` cannot exist on blackboard at the same time. So we need to use another get() method like in GoalUpdatedController.
Now the BT Navigator already has an instance of OdomSmoother, we can't create another one in SpeedController, otherwise we won't get the correct data. A simple fix is to share odom_smoother on blackboard.

Since odom_smoother is shared from BT Navigator, we cannot set `filter_duration` in SpeedController. So I removed this parameter. The default value of 0.3 works well in most cases, so this should not affect usage. But after these changes are confirmed, we need to update the documentation.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/speed_controller.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/decorator/speed_controller.hpp
@@ -57,7 +57,6 @@ public:
       BT::InputPort<double>("max_rate", 1.0, "Maximum rate"),
       BT::InputPort<double>("min_speed", 0.0, "Minimum speed"),
       BT::InputPort<double>("max_speed", 0.5, "Maximum speed"),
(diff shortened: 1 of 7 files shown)


SteveMacenski:
Check CI. This PR has linting, test coverage, and XML invalidation errors.

SteveMacenski (approved):
Otherwise after CI is happy with your changes, this all looks good to me

SteveMacenski (approved):

SteveMacenski:
Thanks for the fix!

Issue #3182, opened 2022-09-09
Running TB3 simulation with some namespace won't work with composition nodes


AlexeyMerzlyakov:
When I running Multi-TB3 simulation with `use_composition:=True` (default), Nav2 stack does not start:
```
[leha@leha-PC nav2_ws]$ ros2 node list
/gazebo
/launch_ros_4468
/robot1/intel_realsense_r200_depth_driver
/robot1/nav2_container
/robot1/robot_state_publisher
/robot1/rviz
/robot1/transform_listener_impl_7fe25c0109c0
/robot1/turtlebot3_diff_drive
/robot1/turtlebot3_imu
/robot1/turtlebot3_joint_state
/robot1/turtlebot3_laserscan
/robot2/intel_realsense_r200_depth_driver
/robot2/nav2_container
/robot2/robot_state_publisher
/robot2/rviz
/robot2/transform_listener_impl_7fc7e4001800
/robot2/turtlebot3_diff_drive
/robot2/turtlebot3_imu
/robot2/turtlebot3_joint_state
/robot2/turtlebot3_laserscan
```

When disabling it, everything works fine.

## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 20.04
- ROS2 Version:
  - ROS2 Rolling built from sources
- Version or commit hash:
  - 42d0bbd026f84fb6447982f4448ea5c51f6fdebb
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
```
ros2 launch nav2_bringup multi_tb3_simulation_launch.py autostart:=True
```
#### Expected behavior
Nav2 stack starts

#### Actual behavior
Nav2 main nodes do not start

#### Additional information

It turns out, that running `tb3_simulation_launch.py` with some non-empty root namespace works only with disabled composition nodes. If enable `use_composition` parameter for that case, no nodes from `localization_launch.py` and `navigation_launch.py` (map_server, amcl, controller, smoother, etc...) are starting. This appears without any error or warning messages in the console: [fail.log](https://github.com/ros-planning/navigation2/files/9536799/fail.log)

SteveMacenski:
It would have to be run in different containers that the containers have the namespaces in, I believe. Or, we can remove the option for composition for multiple robot bringup and hardcode it in the launch file to be non-composed.

austin-InDro:
I've run into this issue and wanted to try Compositions for my robot. I've made a temporary fix for the problem (not sure if it's proper).

[...]

PR #3383, opened 2023-01-25
Fix multi-robot launch issues

AlexeyMerzlyakov:
* Use full name of composable nodes container
* Correct and update BT plugins and error codes list
* Fix and update local and global costmap used plugins parameters

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3182 |
| Primary OS tested on | Ubuntu 20.04, 22.04 |
| Robotic platform tested on | Single-bot and multi-robot TB3 standard simulation |

---

This fixes Issues No.1, No.3 and No.4 reported in #3182:

**Issue1**
Nav2 stack is not starting for multi-robot simulation in composable nodes mode.

Root cause:
If we have the push namespace, container for composable nodes won't work in the same `PushNamespace`.

Workaround fix:
Use full container name in all configurations.

**Issue3**
BT node loading failed due to unrecognized `WouldAPlannerRecoveryHelp` node:

Root cause:
Outdated BT plugins list in the parameter YAML-s.

Fix:
Correct and update BT plugins and error codes list.

**Issue4**
Robots are being started, but the local costmap is empty.

Root cause:
Local costmap is not being updated due to absence of any source for obstacle layer,
due to incorrect and outdated parameter YAML-s.

Fix:
Fix and update local and global costmap used plugins parameters.

---

As result, multi-robot launch is working fine for non-composition nodes as presented in the table below:
| Problem 1, 3-4 fixed | Separate nodes | Composite nodes |
| ------------------------------ | --------------------- | ------------------------- |
| Single bot simulation test | PASS | PASS |
| Multi-robot simulation test | PASS | FAIL due to Issue2 |

No documentation updates required

---

Fix remaining **Issue2**:
YAML-parameters are not delivered for all Nav2 composable nodes, what is causing Nav2 stack startup to fail.

Cause:
Composition nodes are not parsing multiple child trees in a non-empty root YAML namespace.

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/localization_launch.py
+++ b/nav2_bringup/launch/localization_launch.py
@@ -40,6 +40,7 @@ def generate_launch_description():
     params_file = LaunchConfiguration('params_file')
     use_composition = LaunchConfiguration('use_composition')
     container_name = LaunchConfiguration('container_name')
+    container_name_full = (namespace, '/', container_name)
     use_respawn = LaunchConfiguration('use_respawn')
     log_level = LaunchConfiguration('log_level')
 
@@ -166,7 +167,7 @@ def generate_launch_description():
         actions=[
             SetParameter("use_sim_time", use_sim_time),
             LoadComposableNodes(
-                target_container=container_name,
+                target_container=container_name_full,
                 condition=LaunchConfigurationEquals('map', ''),
                 composable_node_descriptions=[
                     ComposableNode(
@@ -178,7 +179,7 @@ def generate_launch_description():
                 ],
             ),
             LoadComposableNodes(
-                target_container=container_name,
+                target_container=container_name_full,
                 condition=LaunchConfigurationNotEquals('map', ''),
                 composable_node_descriptions=[
                     ComposableNode(
@@ -191,7 +192,7 @@ def generate_launch_description():
                 ],
             ),
             LoadComposableNodes(
-                target_container=container_name,
+                target_container=container_name_full,
                 composable_node_descriptions=[
                     ComposableNode(
                         package='nav2_amcl',
--- a/nav2_bringup/params/nav2_multirobot_params_1.yaml
+++ b/nav2_bringup/params/nav2_multirobot_params_1.yaml
@@ -187,7 +187,7 @@ local_costmap:
       voxel_layer:
         plugin: "nav2_costmap_2d::VoxelLayer"
         enabled: True
-        publish_voxel_map: True
+        publish_voxel_map: False
         origin_z: 0.0
         z_resolution: 0.05
         z_voxels: 16
--- a/nav2_bringup/params/nav2_multirobot_params_1.yaml
+++ b/nav2_bringup/params/nav2_multirobot_params_1.yaml
@@ -55,52 +55,59 @@ bt_navigator:
     # nav2_bt_navigator/navigate_through_poses_w_replanning_and_recovery.xml
     # They can be set here or via a RewrittenYaml remap from a parent launch file to Nav2.
     plugin_lib_names:
-    - nav2_compute_path_to_pose_action_bt_node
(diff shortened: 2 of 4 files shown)


nav2_bringup/launch/localization_launch.py
```diff
@@ -40,6 +40,7 @@ def generate_launch_description():
     params_file = LaunchConfiguration('params_file')
     use_composition = LaunchConfiguration('use_composition')
     container_name = LaunchConfiguration('container_name')
+    container_name_full = (namespace, '/', container_name)
```
SteveMacenski, line 43: Why not have the bringup file have the `namespace/container_name` as the field for `namespace=` so that the children launch files (localization / navigation) don't have to think about it?

nav2_bringup/params/nav2_multirobot_params_1.yaml
```diff
@@ -172,47 +179,62 @@ local_costmap:
...
       inflation_layer:
         plugin: "nav2_costmap_2d::InflationLayer"
         cost_scaling_factor: 3.0
         inflation_radius: 0.55
-      obstacle_layer:
-        plugin: "nav2_costmap_2d::ObstacleLayer"
+      voxel_layer:
+        plugin: "nav2_costmap_2d::VoxelLayer"
         enabled: True
+        publish_voxel_map: True
```
SteveMacenski, line 190: set to false

nav2_bringup/params/nav2_multirobot_params_2.yaml
```diff
@@ -172,14 +179,21 @@ local_costmap:
...
       inflation_layer:
         plugin: "nav2_costmap_2d::InflationLayer"
         cost_scaling_factor: 3.0
         inflation_radius: 0.55
-      obstacle_layer:
-        plugin: "nav2_costmap_2d::ObstacleLayer"
+      voxel_layer:
+        plugin: "nav2_costmap_2d::VoxelLayer"
         enabled: True
+        publish_voxel_map: True
```
SteveMacenski, line 190: ditto

SteveMacenski (changes requested):

nav2_bringup/launch/localization_launch.py
AlexeyMerzlyakov: That was initial solution, but I've decided to follow the [comment](https://github.com/ros-planning/navigation2/issues/3182#issuecomment-1397543983):
> Or, what if we set the namespace [here instead](https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/launch/navigation_launch.py#L197)?

because of input parameter `constainer_name` will preserve its essence: relative container name. In this case, we do not need to worry about setting full container name relative to namespace as input parameter when launching `localization_launch.py`/`navigation_launch.py` separately without `bringup_launch.py`, which is more suitable for future utilization, I think.

Otherwise, if you like first approach, I will use it instead.

nav2_bringup/params/nav2_multirobot_params_1.yaml
AlexeyMerzlyakov: Done in the https://github.com/ros-planning/navigation2/pull/3383/commits/8039ff25d68f2f349660c426f62c8cdeb5479d20

nav2_bringup/params/nav2_multirobot_params_2.yaml
AlexeyMerzlyakov: Done in the https://github.com/ros-planning/navigation2/pull/3383/commits/8039ff25d68f2f349660c426f62c8cdeb5479d20

SteveMacenski (approved):

LucasHaug:
Just asking, wiill this fix be in the next humble release?

Issue #3347, opened 2023-01-04
logging and displaying time taken for completing the waypoint following in nav2 rviz panel


padhupradheep:
## Feature request

Follow up from the discussion in https://github.com/ros-planning/navigation2/pull/3165#discussion_r1060055035

#### Feature description

- Currently, we do not show how long did the robot took to complete the entire waypoint following. We just show the time taken for a specific waypoint and reset the time and move to the next. Maybe we need to have "total time taken" for showing how long the robot took to complete. 
- Additionally have an info stating about the "time taken for the last waypoint".
- Also I think it would be a nice idea for logging the time taken for each waypoint, which might be used for some sort of analysis later.

SteveMacenski:
Yes, so the waypoint follower result should store time to completion and return that on the final waypoint following result. 

The waypoint follower feedback can also contain the current total execution time alongside the current waypoint

Should be simple enough :smile:

SteveMacenski:
@padhupradheep was this on your radar to do?

padhupradheep:
Yes, it is! I need some time to get in here.. 

But if someone is already interested in getting their hands dirty, I'll let them take it.

Ryanf55:
Hey there, I'd like to learn a bit about waypoints in NAV2. Since this seems like the most reasonably scoped issue related to waypoints, mind assigning it to me?

SteveMacenski:
Sure! We'd love the help here and I think its a great well-defined scope project!

Ryanf55:
> Sure! We'd love the help here and I think its a great well-defined scope project!

Awesome I appreciate it. I'll need a bit of time to get up to speed. First off, it appears there to be support for pausing the navigation, or [waiting](https://github.com/ros-planning/navigation2/blob/main/nav2_msgs/action/Wait.action). 

When computing the total time to complete a waypoint set, should the time the robot is "waiting" or "paused" be excluded from the elapsed time? 

IE, if robot pauses, elapsed timer temporarily is paused.

SteveMacenski:
No, I think the full time it actually takes, even when completing some task or recovery behavior operations should count

Ryanf55:
It took me a bit of time to figure out how to use the waypoint follower and lots of squinting at the discourse Gif's, but I figured out how to use it. 

On my branch, it's printing the time to completion successfully. 
[...]

PR #3384, opened 2023-01-25
Log display time taken waypoints

Ryanf55:
---

## Basic Info

| ------ | ----------- |
|  Display elapsed time in RVIZ panel and add time logging  |  #3347 |

---

* I added two new fields in the waypoint follower action for time logging
* I fixed a typo for a random field that was defined wrong
* I added a field to the RVIZ panel to display total elapsed time
-->

* Added description of additional time reporting fields in the nav2_waypoint_follower

---

* Display current waypoint number and current loop count in RVIZ
* Display of the result.wp_elapsed_time as a table in RVIZ 

## TODO list before merge
- [ ] Propose message contents have comments explaining the usage of the fields
- [ ] Add some unit tests in thenav2_waypoint_follower package
- [x] Properly handle end of waypoint mission setting time to zero
- [x] Decide if the total elapsed time field should not appear if not in waypoint follow mode
- [x] Convert to QTableWidget and make the UI look nearly the same as before
- [x] Handle hiding/showing UI items dynamically like before, but with QTableWidget
- [x] Investigate if calling `new` every render is a bad idea for the table elements 

#### For Maintainers: 


padhupradheep:
You have to rebase from main.. so that you have an updated copy of mergify.yaml in your branch..

Ryanf55:
Ok, so mostly it's working well, however there are times it seems the waypoint follower never terminates despite meeting the goal. 

```
[component_container_isolated-6] [INFO] [1674671838.601573493] [controller_server]: Passing new path to controller.
[component_container_isolated-6] [INFO] [1674671839.055998472] [controller_server]: Reached the goal!
[component_container_isolated-6] [INFO] [1674671839.090739459] [bt_navigator]: Goal succeeded
```

That is what's printed, however the feedback for follow waypoints is still incrementing elapsed time, and the robot is not moving.  Any suggestions?

Ryanf55:
I've just implemented the final part of setting total time taken back to 0 on completion using the follow waypoints goal status message. 

The one bug in the UI right now is there is a spurious Feedback indicator, but only when the waypoints are executing. No idea what's causing it yet. 
image

Ryanf55:
> You have to rebase from main.. so that you have an updated copy of mergify.yaml in your branch..

Just rebased, still failing.

SteveMacenski:
Yeah, the extra feedback status needs to be removed.  Why is there a larger space between the time taken and the recovery counter? I'd reduce the displayed time elapsed for waypoint following to 1 decimal point. I don't think anyone needs the millisecond level accuracy in this tool

nav2_waypoint_follower/include/nav2_waypoint_follower/waypoint_follower.hpp
```diff
@@ -141,6 +141,9 @@ class WaypointFollower : public nav2_util::LifecycleNode
   rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
   std::shared_future<rclcpp_action::ClientGoalHandle<ClientT>::SharedPtr> future_goal_handle_;
 
+  // Clock
+  rclcpp::Clock steady_clock_{RCL_STEADY_TIME};
```
SteveMacenski, line 145: Remove this clock and use `this->now()` for the node's clock which will use simulation or real time when they are different

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -215,6 +221,7 @@ WaypointFollower::followWaypoints()
     }
 
     feedback->current_waypoint = goal_index;
+    feedback->elapsed_time = elapsed_time;
```
SteveMacenski, line 224: Compute elapsed time here, eg `current_time - start_time` and remove L188.

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -277,11 +283,14 @@ WaypointFollower::followWaypoints()
...
+      result->wp_elapsed_time.push_back(now_time - last_wp_time);
+      last_wp_time = now_time;
       if (goal_index >= goal->poses.size()) {
         if (current_loop_no == no_of_loops) {
           RCLCPP_INFO(
-            get_logger(), "Completed all %zu waypoints requested.",
-            goal->poses.size());
+            get_logger(), "Completed all %zu waypoints requested in a time of %.2f seconds.",
+            goal->poses.size(),
+            elapsed_time.seconds());
```
SteveMacenski, line 293: You can reference the feedback message here

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -175,18 +175,24 @@ WaypointFollower::followWaypoints()
...
+  auto start_time = steady_clock_.now();
+  auto last_wp_time = start_time;
+  rclcpp::Duration elapsed_time{0, 0};
+
   // get the goal index, by default, the first in the list of waypoints given.
   uint32_t goal_index = goal->goal_index;
   bool new_goal = true;
 
   while (rclcpp::ok()) {
+    const auto now_time = steady_clock_.now();
```
SteveMacenski, line 187: Rename `current_time`

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -175,18 +175,24 @@ WaypointFollower::followWaypoints()
 
   rclcpp::WallRate r(loop_rate_);
 
+  auto start_time = steady_clock_.now();
+  auto last_wp_time = start_time;
+  rclcpp::Duration elapsed_time{0, 0};
```
SteveMacenski, line 180: This isn't used, you reinstantiate a `auto elapsed_time = now_time - start_time;` in L188.

nav2_bringup/waypoints/tb3_simulation_waypoint_example.yaml
```diff
@@ -0,0 +1,52 @@
+waypoints:
```
SteveMacenski, line 1: Remove this file

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -845,6 +859,14 @@ Nav2Panel::onInitialize()
         navigation_feedback_indicator_->setText(getNavThroughPosesFeedbackLabel());
       }
     });
+  follow_waypoints_goal_status_sub_ = node->create_subscription<action_msgs::msg::GoalStatusArray>(
```
SteveMacenski, line 862: This might be where you duplication lies? I'm not sure why this is necessary for the status of the WPF action

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -1476,38 +1498,48 @@ Nav2Panel::getGoalStatusLabel(int8_t status)
...
 {
   return std::string(
-    "<tr><td width=150>ETA:</td><td>" +
+    "<tr><td width=200>ETA:</td><td>" +
     toString(rclcpp::Duration(msg.estimated_time_remaining).seconds(), 0) + " s"
-    "</td></tr><tr><td width=150>Distance remaining:</td><td>" +
+    "</td></tr><tr><td width=200>Distance remaining:</td><td>" +
     toString(msg.distance_remaining, 2) + " m"
-    "</td></tr><tr><td width=150>Time taken:</td><td>" +
+    "</td></tr><tr><td width=200>Time Taken for Current WP:</td><td>" +
```
SteveMacenski, line 1534: Is this just for WPF or applies to other things?

SteveMacenski (changes requested):

nav2_rviz_plugins/src/nav2_panel.cpp
Ryanf55: It applies to all since this code is used for all. Do you want me to make it conditionally display `for Current WP` and `for current Pose` or something instead?

nav2_rviz_plugins/src/nav2_panel.cpp
Ryanf55: We still need this so that when the WPF action finishes, the elapsed time field gets cleaned up being reset to 0. Without this section, we would need to infer that it is finished from another action, which I have seen is NOT always true. Perhaps it's a bug, but I didn't want to assume it was the same. 

Also, I'm trying to find the duplication for why the status is rendered twice, but nothing I am seeing is calling that part of the code again in my diff.

nav2_waypoint_follower/include/nav2_waypoint_follower/waypoint_follower.hpp
Ryanf55: > Yeah, the extra feedback status needs to be removed. Why is there a larger space between the time taken and the recovery counter? I'd reduce the displayed time elapsed for waypoint following to 1 decimal point. I don't think anyone needs the millisecond level accuracy in this tool

I'm no UI expert, but I believe it's because there are two tables. Changing it to a single table I think would require some more extensive modifications to the formatters and data structures. I'll poke around and see what I can do to fix it. 

 Can change to 1 decimal.

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: I don't want any changes to the normal behavior, only for the WP follower :smile:

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: OK that's fine but you should remove the UX element of it beyond just resetting to zero (if this even is where that duplication is created)

Ryanf55: I'll split these up. Because the data comes from two different callbacks, and you also want it displayed uniformly (a single table), I think this means I'd have to cache the data from the two feedbacks as a member variable, then refactor the formatter to conditionally output parts of the stored data depending on the state.

Ryanf55: @padhupradheep Can you help me by sharing the motivation for keeping the Nav through poses and waypoint following as the same state? Any strong opposition to making these separate states in `nav2_panel`? 

Steve has asked, reasonably, that the UI information for waypoints only shows up when navigating through waypoints. I think that makes sense. Aside from adding some flags to keep track of if the waypoint navigation was started, cancelled, or finished, I was thinking splitting that stuff into another state would be a good approach. Your input is appreciated.

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: The message has a field called "sec" https://docs.ros2.org/foxy/api/builtin_interfaces/msg/Duration.html :wink:

nav2_rviz_plugins/src/nav2_panel.cpp
SteveMacenski: They should be the same for gathering the waypoints, but not the same during processing - they should diverge in the state machine as there are 2 different buttons (NavThroughPose vs WPF) which trigger that transition.

SteveMacenski:
Any updates? :smile:

Ryanf55:
Hey there, 

1) I got my head wrapped around the axle on the state machine and couldn't figure out exactly how to transition the state machine to the new "running" state; I'm brand new to QT. 
2) I got very sidetracked with the table generation and realized that it would be a whole ton cleaner probably to generate the tables dynamically with the [QTableWidget ](https://doc.qt.io/qt-6/qtablewidget.html) class; that would be for a different PR. For this one, there just might be some more code duplication because the strings are hard coded XML and that's how it's already done
3)  I am struggling to test my changes repeatably and don't know how to trigger certain failure conditions (these can be seen by the lack of code coverage for failure modes in waypoint navigation looking at the codecov report). This[ excellent video ](https://www.youtube.com/watch?v=J27HoZM_PTI&ab_channel=CppCon) shows how one might automate testing of a QT application to give more confidence in making changes to something like a state machine. This also would be a separate issue. As much as it would be great to use TDD for the UI in some sense, I'd rather spend some time contributing to some camera control waypoint following algorithm type stuff in NAV2 than a UI that I can't use on my robots. 

I'll be joining the NAV2 working group meeting to formally introduce myself. See ya there.

Ryanf55:
Ok for problem 2, I just manually copied the state machine through inspection to a state-machine capable renderer for markdown called mermaid. See below. Now that I've actually mapped out the state machine, it's clearer how the new state will fit in.

If you like, this diagram could be tracked in source code of the `nav2_rviz_plugins` package in the spirit of a design doc? It ideally would need to maintained in parallel with source.


```mermaid
stateDiagram-v2
    state "Pre Initial" as pre_initial
    state "Initial" as initial
    state "Idle" as idle
    state "Reset" as reset
    state "Paused" as paused
    state "Resumed" as resumed
    state "Paused WP" as paused_wp
    state "Resumed WP" as resumed_wp
    state "Running Nav through Poses" as running_nav_through_poses
    state "Running Follow Waypoints" as running_follow_waypoints
    state "Canceled" as canceled
    state "Accumulating" as accumulating
    state "Accumulated WP" as accumulated_wp
    state "Accumulated Nav through Poses" as accumulated_nav_through_poses

    [*] --> pre_initial

    %% Start/Reset button click transitions
    pre_initial --> idle: InitialThread navigationActive()
    pre_initial --> initial: InitialThread navigationInactive()
    initial --> idle: start reset button clicked()
    idle --> reset: start reset button clicked()
    running_nav_through_poses --> canceled: start reset button clicked()
    paused --> reset: start reset button clicked()
    idle --> accumulating: navigation mode button clicked()
    accumulating --> accumulated_wp: navigation mode button clicked()
    accumulating --> accumulated_nav_through_poses: pause resume button clicked()
    accumulating --> idle: start reset button clicked()
    accumulated_wp --> canceled: start reset button clicked()
    accumulated_nav_through_poses --> canceled: start reset button clicked()

    %% Internal state transitions
    canceled --> idle: Internal Transition
    reset --> initial: Internal Transition
    resumed --> idle: Internal Transition

    %% Pause/Resume button click transitions
    idle --> paused: pause resume button clicked()
    paused --> resumed: pause resume button clicked()

    %% Pause/Resume button waypoint transition
    accumulated_wp --> resumed_wp: pause waypoint button clicked()
    resumed_wp --> accumulated_wp: pause waypoint button clicked()
    resumed_wp --> canceled: start_reset_button clicked()
    

    %% ROSAction Transitions
    idle --> running_nav_through_poses: ROSActionQTransition IdleTransition
    running_nav_through_poses --> idle: ROSActionQTransition runningNavThroughPosesTransition
    running_follow_waypoints --> idle: ROSActionQTransition runningFollowWaypointsTransition
    idle --> accumulated_wp_: ROSActionQTransition idleAccumulatedWpTransition
    accumulated_wp --> accumulating: ROSActionQTransition accumulatedWpTransition
    idle --> accumulated_nav_through_poses: ROSActionQTransition idleAccumulatedNTPTransition
    accumulated_nav_through_poses --> idle: ROSActionQTransition accumulatedNTPTransition
```

SteveMacenski:
I'm not sure to the extent that you should have to play with the state machine. You should (?) be able to make a function for populating that new table when the state for WPF is called (e.g. your `getFollowWaypointsFeedbackLabel`). 

For example, if you don't even make a new feedback indicator object, just use the `navigation_feedback_indicator_`, you could still set the text on those feedback messages and add in the new `Total Time Taken` but also call the other `FeedbackLabel()` function(s) to populate the table within the `getFollowWaypointsFeedbackLabel` function so its all 1 table assembed together. I haven't done the work to know how to store/sync the navigation information with the WPF results, but that's perhaps a direction to go rather than adding a new field that you have to integrate with a new state (?). We already use the same `navigation_feedback_indicator_` to display different information via `getNavToPoseFeedbackLabel` and `getNavThroughPosesFeedbackLabel`. I don't see how the WPF would be manifestly different. 

Pradheep shows an example of that for the WPF already with the WP and Loop numbers: https://github.com/ros-planning/navigation2/blob/ef2c9fc725ccd8fe53cd46165cbf6c70c266087e/nav2_rviz_plugins/src/nav2_panel.cpp#L804-L813. I'm sure you could just add in your additional fields here for only a couple of lines of changes, no?

I'll say that the RVIZ plugin is by far the least well designed thing in the stack. Unlike the rest of it, this is just for demos and for folks to play with when getting started or testing. So I'm less strict to have it be super clean and well tested (plus, when I inherited it, it was already kind of difficult to read, but new features have only made it worse) as its not a "runtime" tool for any serious use.

padhupradheep:
Hey @Ryanf55, Steve is giving you all the necessary help.. 

just in case if you need an extra hand, let me know(ping me on Slack) 😃 I’ll also try to give you an hand.. 🤙🏼

Ryanf55:
> getFollowWaypointsFeedbackLabel

After trying this approach, I've run into the snag that there is mixed information between nav through poses and follow waypoints. The `toLabel` function used in both `getNavThroughPosesFeedbackLabel` and `getNavToPoseFeedbackLabel` templated out the message, assuming it has `estimated_time_remaining`, `distance_remaining`, `navigation_time` and `number_of_recoveries`, which is not present in the feedback for `NavigateWaypoints`. 

Unless the panel starts to cache feedback data between the three actions, `getFollowWaypointsFeedbackLabel` does not have access to all the data needed. The table needs to be populated async from two different feedback subs. Thoughts on this: Use a [QTableWidget](https://doc.qt.io/qt-5/qtablewidget.html) to be populated by the three different feedback indicators. This table would be stored as a member variable of the class. Instead of `getNavThroughPosesFeedbackLabel` returning  a `Qstring` to populate `navigation_feedback_indicator_`, `QLabel`, change `navigation_feedback_indicator_` to a `QTableWidget`. The `Get*Feedback` signature is modified to populate a passed in table. 
```
  static inline void PopulateNavThroughPosesFeedback(
    const nav2_msgs::action::NavigateThroughPoses::Feedback =
    nav2_msgs::action::NavigateThroughPoses::Feedback(),
    QTableWidget* table);
```

An enum is built up for all the table widget data like so:
```
enum class NavigationFeedbackRows {
    waypoint, 
   loop:
   poses_remaining,
   total_time_taken,
   eta,
   time_taken_for_current_wp,
   distance_remaining,
   recoveries
   ...
}
```

This enum is used to populate the rows in each table. Depending on the mode, at each transition, the rows that need to be displayed are either hidden or shown with [hideRow](https://doc.qt.io/qt-5/qtableview.html#hideRow) during the transitions between states.

SteveMacenski:
> the panel starts to cache feedback data

Sure, why not? There's a bunch of stuff already stored: 
- https://github.com/ros-planning/navigation2/blob/ef2c9fc725ccd8fe53cd46165cbf6c70c266087e/nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp#L86-L92
- https://github.com/ros-planning/navigation2/blob/ef2c9fc725ccd8fe53cd46165cbf6c70c266087e/nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp#L144-L149

A table could work too!

SteveMacenski:
LMK when you want me to take a look again

Ryanf55:
> LMK when you want me to take a look again

Will do. QTableWidget is the best (no more hard coded XML strings), but I'm spending time mucking with the appearance to get it to be fixed size. I'll ping you here when it's ready for some eyes. No blocks.

Ryanf55:
At this point, I have done an attempt at replacing the get*Label() functions with  the following: 
```
RenderNavThroughWPFeedback(
    QTableWidget* table,
    const nav2_msgs::action::FollowWaypoints::Feedback msg =
    nav2_msgs::action::FollowWaypoints::Feedback()
);
```

There is a single `QTableWidget` that stores all the previous data. It has allocated enough rows to view all the previous data. In the above commit, I implemented the waypoint feedback and result rendering, and it works. The table is not formatted at all, so that will need to be done (coloring, sizing, etc). I think it will work good. 

Other than styling of the table,  and implementing the other feedback/result channels for the table, I'll need to implement state change behavior. At each state change, the table rows may need to be hidden/shown. 

That should cover the rest of the work.

Ryanf55:
I could use some clarification on the "cleanup" behavior after actions complete. There are some spots where data is left as-is, and others where it's immediately reset to zero at the state change. Is this inconsistency intentional depending on which data member it is, or is there a general process I should follow? This applies to rows in the feedback table that stick around. 

For example, you do a simple Nav2 goal. While the robot is going to goal, data updates. Then, the action completes. Should `Distance Remaining`, and `Time taken for Current WP` be reset to 0, or do users want to see the data for a bit?

To me, it makes sense to preserve it until you go to the Reset state, or obviously, start another action.

SteveMacenski:
>  There are some spots where data is left as-is

As far as I can tell from a quick test, its always reset back to 0 after being cancelled or completed (the only difference is the `Feedback:` field being "canceled" in yellow or "reached" in green). What's not cleaned up? 

I'd actually argue thinking about it that keeping `time taken` and `recoveries` would be nice to keep up until the next run. The ETA/distance remaining should both be exactly `0` if we have achieved the goal, so those should be reset in case there's an small error.

Ryanf55:
Yea, overall that agrees with my expectations. Pushing what I have now in 68cfb2f838afa37d818b72e7110d271f55bfe9d6, which is everything is always visible. I'll go through and figure out how to use `assignProperty` on the row visibility, and also setting the things to zero as appropriate, when I have more time. Thanks!

Ryanf55:
Good news - UI is looking less look a data table and more like the original. Rows are dynamically added fine now too. 
image

I think found some bugs in behavior, not sure if they are in scope to fix. I'll need to confirm they are present on rolling. 

1. In Follow WP mode, Waypoint cound does not change UNLESS there is more than 0 loops. To me, this does not seem correct, and the user would like to know what waypoint number they are on, even if there are no loops. Fix: TBD. 
2. In Nav to Pose mode, the result condition is never met because `goal_index_ == static_cast<int>(store_poses_.size()) - 1` is never true. This is preventing from using that code block to reset the data members to 0 in that goal handle. Fix: Remove the `-1`, it doesn't seem right, and I can't seem to find any reason why it's like that. The other conditions there look fine.

Ryanf55:
PR is ready for review. I've met all the acceptance criteria. Of note, I couldn't ever figure out how to get the state machine into `running_follow_waypoints`. So either I can remove everything referencing that state, or we can figure out how to enter it. The logic for hiding/showing navigation feedback isn't the clearest because I couldn't link it to the states with `assignProperty` (it's not a public attribute of the table).

nav2_waypoint_follower/src/waypoint_follower.cpp
Ryanf55: Done in 41d480d4a85169076efe88cd9ff85b67cd5493e3

SteveMacenski:
Awesome! What about the big gap at the end of the table your last image shows? 

> In Follow WP mode, Waypoint cound does not change UNLESS there is more than 0 loops. To me, this does not seem correct, and the user would like to know what waypoint number they are on, even if there are no loops.

@padhupradheep ? 

For your second issue - I'm just reporting what I saw when I tested out the resetting behavior myself on a test environment. It seemed to work fine for me - which is odd if that condition as you say is never met! That doesn't feel like something that would be flaky as in working sometimes but not others.

padhupradheep:
> In Follow WP mode, Waypoint cound does not change UNLESS there is more than 0 loops. To me, this does not seem correct, and the user would like to know what waypoint number they are on, even if there are no loops.

I understand, the reason I left it because, the waypoint number can be seen on the waypoint marker. But thinking of it, I guess it'll be a good idea to add it for "no loop" case as well. I'd suggest a separate PR for that.

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -15,7 +15,9 @@
 #ifndef NAV2_RVIZ_PLUGINS__NAV2_PANEL_HPP_
 #define NAV2_RVIZ_PLUGINS__NAV2_PANEL_HPP_
 
+#include <QString>
```
padhupradheep, line 18: Please remove, not required

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -15,7 +15,9 @@
...
 #include <QtWidgets>
+#include <QTableWidget>
```
padhupradheep, line 20: Same here, not required

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -17,9 +17,12 @@
 #include <QtConcurrent/QtConcurrent>
 #include <QVBoxLayout>
 #include <QHBoxLayout>
+#include <QTableWidgetItem>
```
padhupradheep, line 20: not required

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -17,9 +17,12 @@
...
 #include <QTextEdit>
 #include <QCheckBox>
 
+
+#include <array>
```
padhupradheep, line 25: this too

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -262,7 +312,7 @@ Nav2Panel::Nav2Panel(QWidget * parent)
 
   accumulated_nav_through_poses_->assignProperty(
     navigation_mode_button_, "text",
-    "Start Waypoint Following");
+    "antp: Start Waypoint Following");
```
padhupradheep, line 315: what does "antp" mean? I would leave it unturned..

padhupradheep (changes requested):
Good job.. 

First of satisfy linters... run `ament_cpplint` and `ament_uncrustify` to see what all things that need to be changed for satisfying for it. 

I have requested some minor changes after doing a quick glance. I'll go through it once again and give you another review, I also have to test it in my side.

nav2_rviz_plugins/src/nav2_panel.cpp
Ryanf55: This was for debugging. As far as I can tell, I can not figure out how to trigger this line of code.

Ryanf55:
> Awesome! What about the big gap at the end of the table your last image shows?
> 
> > In Follow WP mode, Waypoint cound does not change UNLESS there is more than 0 loops. To me, this does not seem correct, and the user would like to know what waypoint number they are on, even if there are no loops.
> 
> @padhupradheep ?
> 
> For your second issue - I'm just reporting what I saw when I tested out the resetting behavior myself on a test environment. It seemed to work fine for me - which is odd if that condition as you say is never met! That doesn't feel like something that would be flaky as in working sometimes but not others.
> 
> If you can't preserve things - just reset everything to 0 then at the end of a run as current behavior does.
> 
> Please rebase onto `main` since alot has changed since this was opened. @padhupradheep can you review?

Will do. 


Rebased on `main` to now branch from 6d12726efb936797b527edfa2f3191ae52f9ae77

nav2_rviz_plugins/src/nav2_panel.cpp
Ryanf55: Does NAV2 intend to use transitive includes? I am using array in the header now.

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -182,10 +190,12 @@ private Q_SLOTS:
...
 
   // The following states are added to allow for the state of the button to only expose reset
-  // while the NavigateToPoses action is not active. While running, the user will be allowed to
-  // cancel the action. The ROSActionTransition allows for the state of the action to be detected
+  // while neither the NavigateToPoses action nor FollowWaypoints are active. While running,
+  // the user will be allowed to cancel the action.
+  // The ROSActionTransition allows for the state of the action to be detected
   // and the button state to change automatically.
-  QState * running_{nullptr};
+  QState * running_nav_through_poses{nullptr};
```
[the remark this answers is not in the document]
Ryanf55: done in 3facdbbbdf590307983297551b756b8cab501f69

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -175,6 +175,9 @@ WaypointFollower::followWaypoints()
 
   rclcpp::WallRate r(loop_rate_);
 
+  auto start_time = this->now();
```
SteveMacenski, line 178: start_time and last_wp_time isn't reset when the action is preempted with a new task

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -215,6 +218,8 @@ WaypointFollower::followWaypoints()
     }
 
     feedback->current_waypoint = goal_index;
+    const auto current_time = this->now();
+    feedback->elapsed_time = current_time - start_time;;
```
SteveMacenski, line 222: double `;`

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: Same with `result = std::make_shared<ActionT::Result>();` to reset the wp states if we have a new task. This isn't a bug you introduced, but just finding it now

SteveMacenski (commented):
I didn't have the opportunity today to review the rviz parts, but I did the WPF parts

nav2_rviz_plugins/src/nav2_panel.cpp
padhupradheep: Interesting.. I think array is already part of the standard library and there in no necessity to add it here, since I guess the std library is already added. It should have been the case with vector too. We need memory because it is not part of the standard library. 

Maybe @SteveMacenski can shed some light here..

padhupradheep:
@Ryanf55 could you please explain why do you want to go into the `running_follow_waypoints` state? exactly at which scenario?

padhupradheep:
>  Of note, I couldn't ever figure out how to get the state machine into running_follow_waypoints

@Ryanf55 could you please explain why do you want to go into the `running_follow_waypoints` state? exactly at which scenario?

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -39,6 +42,36 @@ namespace nav2_rviz_plugins
 {
 using nav2_util::geometry_utils::orientationAroundZAxis;
 
+// Define the row numbers for navigation_data_table_
+enum class NavigationDataTableRows {
+  // Waypoint number
+  kWaypoint,
```
padhupradheep, line 48: why do you have "k" at the beginning, could you maybe remove them and probably rename them?

nav2_rviz_plugins/include/nav2_rviz_plugins/nav2_panel.hpp
```diff
@@ -182,10 +190,12 @@ private Q_SLOTS:
   QLabel * imgDisplayLabel_{nullptr};
 
   // The following states are added to allow for the state of the button to only expose reset
-  // while the NavigateToPoses action is not active. While running, the user will be allowed to
-  // cancel the action. The ROSActionTransition allows for the state of the action to be detected
+  // while neither the NavigateToPoses action nor FollowWaypoints are active. While running,
+  // the user will be allowed to cancel the action.
+  // The ROSActionTransition allows for the state of the action to be detected
   // and the button state to change automatically.
-  QState * running_{nullptr};
```
padhupradheep, line 188: on first place, why is this being split into two different states? `running_` is kind of the state that the state machine transition to, when you give a single goal. This is no ways related to the waypoint navigation.

The states `accumulated_wp_` and `accumulated_nav_through_poses_` are the equivalent to the `running_` state for the navigation in the waypoint mode.

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -281,31 +331,58 @@ Nav2Panel::Nav2Panel(QWidget * parent)
...
+  running_nav_through_poses_ = new QState();
+  running_nav_through_poses_->setObjectName("running_nav_through_poses");
+  running_nav_through_poses_->assignProperty(start_reset_button_, "text", "Cancel");
+  running_nav_through_poses_->assignProperty(start_reset_button_, "toolTip", cancel_msg);
+
+  running_nav_through_poses_->assignProperty(pause_resume_button_, "text", "Pause");
+  running_nav_through_poses_->assignProperty(pause_resume_button_, "enabled", false);
+
+  running_nav_through_poses_->assignProperty(navigation_mode_button_, "text", "Waypoint mode");
+  running_nav_through_poses_->assignProperty(navigation_mode_button_, "enabled", false);
```
padhupradheep, line 345: This needs to go back to the original state.. `running_` state should be untouched..

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -425,12 +504,17 @@ Nav2Panel::Nav2Panel(QWidget * parent)
...
   // states and then goes inactive, move back to the idle state. Vise versa as well.
   ROSActionQTransition * idleTransition = new ROSActionQTransition(QActionState::INACTIVE);
-  idleTransition->setTargetState(running_);
+  idleTransition->setTargetState(running_nav_through_poses_);
   idle_->addTransition(idleTransition);
+  //! @todo add a transition for follow WP here?
+
+  ROSActionQTransition * runningNavThroughPosesTransition = new ROSActionQTransition(QActionState::ACTIVE);
+  runningNavThroughPosesTransition->setTargetState(idle_);
+  running_nav_through_poses_->addTransition(runningNavThroughPosesTransition);
```
padhupradheep, line 513: No the idle accumulated_state_transition is taken care somewhere in the below, you do not have to do it here..

padhupradheep (changes requested):
I think there seems to be a confusion in the understanding of `running_` , `accumulated_nav_through_poses_` and `accumulated_wp_`. 

As explained in the comments, `running_` state is for just a single goal. There is no waypoint following happening there.. So you do not have to split them.

Ryanf55:
> > Of note, I couldn't ever figure out how to get the state machine into running_follow_waypoints
> 
> @Ryanf55 could you please explain why do you want to go into the `running_follow_waypoints` state? exactly at which scenario?

I wanted a different state than `running_` which I incorrectly thought was used by all the nav through poses and follow WP. Based on your other comments about this, I will revert those specific changes of adding a new state since it's not needed. 

Would you be open to a rename of the three states though? IMHO, `accumulated_wp_` does not mean robot is actually doing anything, more that it has waypoints available to run, but isn't actively going through them.

My proposal: 
| Original State Name | New State Name |
| - | - |
| running_ | running_nav_to_pose_ |
| accumulated_wp_ | running_wp_ |
| accumulated_nav_through_poses_ | running_nav_through_poses |

padhupradheep:
Yeah, I have to agree with you, the naming sounds bit confusing.  Sure, I would guess Steve will also agree with your proposal. 

As Steve pointed out sometime ago, the RViz plugin is getting bit crowded and messy, there must be at least a partial refactor sometime soon in the future. 

-------------------------------------------------------------
Also make sure that you remove all the debuggers. It would be easy for us to review, if you can show a working GUI. Use `ffmpeg` for creating the GiF and `simplescreenrecorder` for creating the video. 

Once you think you are done with the feature, try to satisfy the linters and then we are good to go. 

I've to really appreciate the time and effort you are putting @Ryanf55 .. There will be many people benefiting out on what you are working on. So keep it up.

Ryanf55:
> Yeah, I have to agree with you, the naming sounds bit confusing. Sure, I would guess Steve will also agree with your proposal.
> 
> As Steve pointed out sometime ago, the RViz plugin is getting bit crowded and messy, there must be at least a partial refactor sometime soon in the future.
> 
> Before the next request for review, make sure that you remove all the debuggers. It would be easy for us to review, if you can show a working GIF. Use `ffmpeg` for creating the GiF and `simplescreenrecorder` for creating the video.
> 
> Once you think you are done with the feature, try to satisfy the linters and then let us know once again.
> 
> I've to really appreciate the time and effort you are putting @Ryanf55 .. There will be many people benefiting out on what you are working on. So keep it up.

Will do on touching up those last things. I'll add a video once it's done. 

Thanks, sounds good. Hopefully some of the changes I've done here using QTableWidget, using more pure functions rather than global state, and building more of the UI at compile time rather than through manual HTML will make that easier.

nav2_rviz_plugins/src/nav2_panel.cpp
Ryanf55: Because ROS tier one packages use `ament_lint_common` which uses google style guide. Google style guide says enums follow that format. 

Reference: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

nav2_waypoint_follower/src/waypoint_follower.cpp
Ryanf55: Do you want me to fix it in this PR? Or can I split that into its own?

nav2_waypoint_follower/src/waypoint_follower.cpp
Ryanf55: Did you have a linter that told you that, or are you just a ninja?

Ryanf55:
> > > Of note, I couldn't ever figure out how to get the state machine into running_follow_waypoints
> > 
> > 
> > @Ryanf55 could you please explain why do you want to go into the `running_follow_waypoints` state? exactly at which scenario?
> 
> I wanted a different state than `running_` which I incorrectly thought was used by all the nav through poses and follow WP. Based on your other comments about this, I will revert those specific changes of adding a new state since it's not needed.
> 
> Would you be open to a rename of the three states though? IMHO, `accumulated_wp_` does not mean robot is actually doing anything, more that it has waypoints available to run, but isn't actively going through them.
> 
> My proposal:
> 
> Original State Name	New State Name
> running_	running_nav_to_pose_
> accumulated_wp_	running_wp_
> accumulated_nav_through_poses_	running_nav_through_poses

This is now changed on 4fb895daf33203ea84d52a61daebd2ab9de7f6ec

Ryanf55:
> ament_cpplint

Done in 4381b27be282e80d9e97fe9171fbca388c8c550b

padhupradheep:
Probably try this.. https://superuser.com/questions/556029/how-do-i-convert-a-video-to-gif-using-ffmpeg-with-reasonable-quality

nav2_waypoint_follower/src/waypoint_follower.cpp
SteveMacenski: It should be fixed here

SteveMacenski:
The renaming sounds good to me - i think that's the only thing from the discussion above I needed to respond to?

padhupradheep:
yes!

And the transitive includes..

nav2_rviz_plugins/src/nav2_panel.cpp
Ryanf55: @SteveMacenski Please advise on the use of transitive includes in NAV2? 

Google style guide says not to reply on them, but @padhupradheep says my includes are not necessary. 
https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

SteveMacenski: You need to have it if you use it.

SteveMacenski:
I went through the code and overall looks good to be beyond:
- I didn't do any functional testing at this point, so I can't say if its works or whatnot
- There were alot of inlined TODOs that need to be addressed
- I didn't pick up during the scan if things like "loop number" are always present on the table even when in non-looping operating modes. I did see some hiding stuff going on, but it wasn't at least immediately clear to me that was what was going on there

Ryanf55:
> I went through the code and overall looks good to be beyond:
> 
> * I didn't do any functional testing at this point, so I can't say if its works or whatnot
> * There were alot of inlined TODOs that need to be addressed
> * I didn't pick up during the scan if things like "loop number" are always present on the table even when in non-looping operating modes. I did see some hiding stuff going on, but it wasn't at least immediately clear to me that was what was going on there
> * It sounds like from Pradheep's comments that there are missing transitions

Thanks. Yes, the showing and hiding is annoyingly hard to reason about because I couldn't tie it to the state object directly, rather the transition logic. Overall, the logic does not make any assumptions about what was shown or hidden before, it always goes through the whole table. It's only a few more cycles. 

The missing transitions are no longer there since that state was removed. It was unnecessary. The state machine has renamed states, but no logic has been changed regarding transitions in this PR.

padhupradheep:
image

I still see a lot of space... :thinking:

Ryanf55:
It looks like before the elements were equally spaced out. 
image

I had to allocate the size for the table (it's there the whole time at max size, you just don't see it). If the size allocated was too small, QT added a scroll bar, which we don't want. To solve this, I could implement dynamic sizing, however I can't figure out how. Any tips?

Ryanf55:
This is the closest reference I found. Pretty wild this isn't built into the widget. 
https://stackoverflow.com/questions/8766633/how-to-determine-the-correct-size-of-a-qtablewidget

SteveMacenski:
I'm not entirely sure - the QT Table is a different tool than I've used before to be able to give you concrete advise. I think using only the minimum amount of space could help. Obviously that last screen shot was more space than the table would use even if fully populated.

Ryanf55:
> 

Sadly, that screenshot was pre-QTable. I think QTable is the right thing, so I'll just need some more time to play with it. If you happen to know any competent QT devs in the ROS2 community, please let them know I could use a hand.

SteveMacenski:
Hi @Ryanf55 any updates? :smile:

padhupradheep:
I'll probably take it over if @Ryanf55 is busy with something else.

SteveMacenski:
What's the status of this? Something you're planning on finishing up?

padhupradheep:
First of all really sorry for the long silence. 

My schedule is clearing up and I was planning to come back to this by mid of November. 

It’s definitely in my plans.

SteveMacenski:
Closing as `wontfix`, this doesn't seem to be of much interest and the technical challenges are not worth the superficial payoff. If someone would like to implement / needs it, I'd welcome a functional PR to review and merge, but this can be safely closed.

PR #3388, opened 2023-01-30
Fix mergify

ruffsl:
Related to

From the mergify documentation on attributes, it looks like `body-raw` was still what you wanted, however it seems that the yaml parser was tripping up on the complete regex pattern of the entire line. For example, I couldn't get it to escape the colon punctuation, so I just truncated the pattern.  

https://docs.mergify.com/conditions/#attributes 

Before:
https://github.com/ruffsl/navigation2/pull/10

After:
https://github.com/ruffsl/navigation2/pull/13

#### For Maintainers: 

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -31,7 +31,7 @@ pull_request_rules:
       - merged
     actions:
       delete_head_branch:
- 
+
   - name: ask to resolve conflict
     conditions:
       - conflict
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Still triggered… good thought though! I didn’t even think about the colon. I had tried to match the full line both body and body-raw, with and without the section header / comment about do not remove- but never the colon!

ruffsl:
> Still triggered…

Just like most repo level CI/CD configs, they don't often apply until they merged into the base branch. I often a safety or security mesture so that randos can't just open a public PR with the config rules changed to get there commits to pass CI/CD. For example, you can see this notice when I PR'ed a config change to my own fork.

https://github.com/ruffsl/navigation2/runs/10980446644

PR #3398, opened 2023-02-06
Nav2 route server parser

jwallace42:
---

## Basic Info
Addes geojson parser

| ------ | ----------- |
| Robotic platform tested on | NONE |

---

- Adds geojson parser
- pulls out graph file loader into a interface

---

#### For Maintainers: 


nav2_route/graphs/geojson/aws_graph.geojson
```diff
@@ -0,0 +1,131 @@
+{
+"type": "FeatureCollection",
+"name": "nodes",
```
SteveMacenski, line 3: I think we mentioned having a timestamp associated with the header of last edited?

nav2_route/graphs/geojson/aws_graph.geojson
```diff
@@ -0,0 +1,131 @@
...
+"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::3857" } },
+"features": [
+{ "type": "Feature", "properties": { "id": 0 }, "geometry": { "type": "Point", "coordinates": [ -4.0, -9.35 ] } },
```
SteveMacenski, line 6: None of the coordinates have the frame. Generally isn't required but for the best practices example it should have them

nav2_route/include/nav2_route/interfaces/graph_file_loader.hpp
```diff
@@ -0,0 +1,61 @@
...
+   * @brief Constructor
+   */
+  GraphFileLoader() = default;
+
+  /**
+   * @brief Virtual destructor
+   */
+  virtual ~GraphFileLoader() = default;
+
+  /**
```
SteveMacenski, line 45: This should have a validate file API to check if this file exists and is valid. Different specifications might have different ways of wanting to do that (e.g. xml can be verified via STL, as we do in BT Navigator before attempting to load a file to know if its not valid)

nav2_route/src/route_server.cpp
```diff
@@ -65,14 +65,20 @@ RouteServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
   declare_parameter_if_not_declared(
     node, "max_planning_time", rclcpp::ParameterValue(2.0));
 
+  nav2_util::declare_parameter_if_not_declared(
```
SteveMacenski, line 68: Move back into the GraphLoader object to declare and get the parameter value. I don't think we need it here in the main route server file. We can let that object grab it itself.

SteveMacenski: (and remove the class member added to the server too)

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  // Convert all coordinates to `frame` (in a new method) for standardization
+  // Including conversion of GPS coordinates, so we can populate it in some
+  // cartesian frame necessary for traversal cost estimation and densifying
+  // (and so we don't need to propogate it through our structures)
+}
+
+bool GraphLoader::fileExists(const std::string & filepath)
+{
+  return std::filesystem::exists(filepath);
```
SteveMacenski, line 103: Yeah, this should be part of that validate function in the plugins

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+
+bool GraphLoader::loadGraphFromFile(
+  Graph & graph,
+  GraphToIDMap & graph_to_id_map,
+  const std::string & filepath,
+  std::string parser_id)
+{
+  if (!fileExists(filepath)) {
```
SteveMacenski, line 71: Add a general log at the start here that we're loading a file of path XYZ via plugin XYZ for debugging

nav2_route/src/route_server.cpp
SteveMacenski: I think this also makes sense to have no default value, so if it throws and exception while getting it, it should be caught as a failure.

nav2_route/src/graph_loader.cpp
SteveMacenski, line 71: Yeah, this `fileExist` should be more like `plugin->isFileValid(filepath)`

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,106 @@
...
+    RCLCPP_WARN(logger_, "Parser id was unset, setting to %s", default_plugin_id_[0].c_str());
+    parser_id = default_plugin_id_[0];
+  }
+
+  bool result = false;
+  if (graph_parsers_.find(parser_id) != graph_parsers_.end()) {
+    try {
+      result = graph_parsers_[parser_id]->loadGraphFromFile(graph, graph_to_id_map, filepath);
+    } catch (std::exception & ex) {
+      throw ex;
```
SteveMacenski, line 86: If you're just going to throw, you don't need to catch or wrap in the try in the first place

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,106 @@
...
+    } catch (pluginlib::PluginlibException & ex) {
+      RCLCPP_FATAL(
+        logger_,
+        "Failed to create graph parser. Exception: %s", ex.what());
+      throw ex;
+    }
+  }
+}
+
+bool GraphLoader::loadGraphFromFile(
```
SteveMacenski, line 65: See my branch - we need this to be signature to check if the file to use is given in the parameter or not to use the configuration default. This lets us use this to both not give it a file or to give it a file for use on a service (or unit test). If we don't provide it, use the parameterized default. 

I suppose this makes sense to "always use" since you moved the graph filename from parameter stuff into the server, but I want to keep that in the server for encapsulation of like. Its a subjective choice, but its the choice I made all across this server to have each object grab their own data so this should follow that lead.

SteveMacenski (changes requested):

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,106 @@
...
+  return result;
+
+  // Validate file is legit using a plugin API TODO(sm)
+  // Load file using a plugin API
+
+  // Convert all coordinates to `frame` (in a new method) for standardization
+  // Including conversion of GPS coordinates, so we can populate it in some
+  // cartesian frame necessary for traversal cost estimation and densifying
+  // (and so we don't need to propogate it through our structures)
+}
```
SteveMacenski, line 99: TODO on metadata

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+bool GeoJsonGraphFileLoader::loadGraphFromFile(
+  Graph & graph, GraphToIDMap & graph_to_id_map, std::string filepath)
+{
+  std::ifstream graph_file(filepath);
+
+  Json json;
+  graph_file >> json;
+
+  auto features = json.at("features");
```
SteveMacenski, line 33: Is features a universal thing or do we need to make that a parameter to search or and/or grab this from the file's root itself?

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+void GeoJsonGraphFileLoader::addNodesToGraph(
+  Graph & graph, GraphToIDMap & graph_to_id_map, std::vector<Json> & nodes)
+{
+  int idx = 0;
+  for (const auto & node : nodes) {
+    // Required data
+    unsigned int id = node["properties"]["id"];
+    float x = node["geometry"]["coordinates"][0];
```
SteveMacenski, line 71: Not parsing frame

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,107 @@
...
+  getEdges(features, edges);
+
+  graph.resize(nodes.size());
+  addNodesToGraph(graph, graph_to_id_map, nodes);
+  addEdgesToGraph(graph, edges);
+
+  return true;
+}
+
+void GeoJsonGraphFileLoader::getNodes(const Json & features, std::vector<Json> & nodes)
```
SteveMacenski, line 46: Combining the getnodes and get edges functions would make this faster since we're not iterating over the potentially massive file 2x

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Recommended data
+    if (edge_properties.contains("cost")) {
+      edge_cost.cost = edge_properties["cost"];
+    }
+
+    if (edge_properties.contains("overridable")) {
+      edge_cost.overridable = edge_properties["overridable"];
+    }
+
+    graph[start_id].addEdge(edge_cost, &graph[end_id], id);
```
SteveMacenski, line 100: start_id is the nodeid of the node, but as you know, that doesn't line up with the graph's indexing! You have to use the `graph_to_id_map` to know what the graph's index belongs to the nodeID. 

Its a little convoluted, but trust me, the speed up is worth it. Storing the graph in an unordered_map would incur far higher lookup times than direct memory access in a vector at runtime.

SteveMacenski:
And unit tests, of course :-)

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
jwallace42: This is universal for geojson

nav2_route/graphs/geojson/aws_graph.geojson
SteveMacenski: If that's possible!

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,132 @@
...
+  graph.resize(nodes.size());
+  addNodesToGraph(graph, graph_to_id_map, nodes);
+  addEdgesToGraph(graph, graph_to_id_map, edges);
+
+  return true;
+}
+
+bool GeoJsonGraphFileLoader::fileExists(const std::string & filepath)
+{
+  return std::filesystem::exists(filepath);
```
SteveMacenski, line 70: Also check `std::ifstream xml_file(filename); if (!xml_file.good()) {return false;}` to check that the file is well formed in XML Format standards. This catches if there's syntactical errors. 

You could actually store that `ifstream` object for use later in the main function. I don't know how heavy that is to reconstruct if its meaningful.

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+
+void GeoJsonGraphFileLoader::getGraphElements(
+  const Json & features, std::vector<Json> & nodes, std::vector<Json> & edges)
+{
+  for (const auto & feature : features) {
+    if (feature["geometry"]["type"] == "Point") {
+      nodes.emplace_back(feature);
+    }
+    if (feature["geometry"]["type"] == "MultiLineString") {
```
SteveMacenski, line 81: else if this so we don't check after we found a node if its also an edge

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,132 @@
...
+  for (const auto & edge : edges) {
+    // Required data
+    const auto edge_properties = edge["properties"];
+    unsigned int id = edge_properties["id"];
+    unsigned int start_id = edge_properties["startid"];
+    unsigned int end_id = edge_properties["endid"];
+
+    // Recommended data
+    if (edge_properties.contains("cost")) {
+      edge_cost.cost = edge_properties["cost"];
```
SteveMacenski, line 118: EdgeCost isn't reset between iterations of the for loop, so if 1 populates it and others don't, that information is propogated. YOu might just want to allocate the `edge_cost` at the start of the for loop instead since its a trivial struct of a float and a bool (proportional to the things already being allocated for the unsigned int's).

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+void GeoJsonGraphFileLoader::addNodesToGraph(
+  Graph & graph, GraphToIDMap & graph_to_id_map, std::vector<Json> & nodes)
+{
+  int idx = 0;
+  for (const auto & node : nodes) {
+    // Required data
+    unsigned int id = node["properties"]["id"];
+    std::string frame = node["properties"]["frame"];
```
SteveMacenski, line 94: What if frame isn't given? Should this be handled similarly to the edge cost/override fields?

nav2_route/test/test_geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,252 @@
...
+  bool result = graph_file_loader.loadGraphFromFile(graph, graph_to_id_map, file_path);
+  EXPECT_TRUE(result);
+
+  EXPECT_EQ(graph.size(), 2u);
+
+  EXPECT_EQ(graph[0].nodeid, 0u);
+  EXPECT_EQ(graph[0].coords.x, 0.0);
+  EXPECT_EQ(graph[0].coords.x, 0.0);
+  EXPECT_EQ(graph[0].coords.frame_id, "map");
+  EXPECT_EQ(graph[0].neighbors.size(), 1u);
```
SteveMacenski, line 241: Check the edge is populated correctly

nav2_route/test/test_geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  Graph graph;
+  GraphToIDMap graph_to_id_map;
+  GeoJsonGraphFileLoader graph_file_loader;
+  bool result = graph_file_loader.loadGraphFromFile(graph, graph_to_id_map, file_path);
+  EXPECT_FALSE(result);
+  std::filesystem::remove(file_path);
+}
+
+TEST(GeoJsonGraphFileLoader, missing_node_id)
```
SteveMacenski, line 116: 1 more test: what is a node which the edge is referring to doesn't exist anymore (e.g. due to someone manipulating a graph and messing it up)

SteveMacenski (changes requested):

SteveMacenski:
The last thing is having test coverage for the file loader object itself, not only the parser. Just makes sure that the factory is picking up the configurations properly (e.g. run a test where you set it to bogus plugins so it causes it to fail to load -- see the edge scoring test file) and setting a logical default that works on some basic graph example. 

The tests I've written are complete, but admittedly a little odd. I wrote the tests for the factory objects (e.g. edge scorer, operations manager) but then also wrote the tests for the algorithms loading them through that interface one by one in different tests. It was necessary for the operations manager to test the different operation processing modes (which the plugins have a spanning set of) but lazily did it for the edge scorer as well. Its not great tactical design choice, but its still completely functional and I was able to do what I needed to do through that second level of indirection easily.

nav2_route/test/CMakeLists.txt
```diff
@@ -53,6 +53,17 @@ target_link_libraries(test_operations
   ${library_name} route_operations
 )
 
+# Test geojson parser
+ament_add_gtest(test_geojson_graph_file_loader
+    test_geojson_graph_file_loader.cpp
+  ../src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```
SteveMacenski, line 59: huh? 

Remove that. INstead in the target link libraries, just add the geojson library. See `test_collision_operation` below where it relies on a `route_operation` plugin in that library name. That way there are no fragile filepaths

nav2_route/include/nav2_route/interfaces/graph_file_loader.hpp
```diff
@@ -0,0 +1,78 @@
...
+    Graph & graph,
+    GraphToIDMap & graph_to_id_map,
+    std::string filepath) = 0;
+
+  /**
+   * @brief Checks if a file even exists on the filesystem
+   * @param filepath file to check
+   * @return bool If the file exists
+   */
+  virtual inline bool fileExists(const std::string & filepath) = 0;
```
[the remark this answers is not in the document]
jwallace42: I added a try catch to the load graph from file method. Otherwise we have redudent parsing.

jwallace42:
Things left to do: 
- [ ] Test for graph_loader server 
- [ ] Parse meta data 
- [ ] Frame transformations

nav2_route/test/CMakeLists.txt
jwallace42: Yeah, I just did this quick a while ago to get the test up and running :).

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,148 @@
...
+    unsigned int end_id = edge_properties["endid"];
+
+    if (graph_to_id_map.find(start_id) == graph_to_id_map.end()) {
+      RCLCPP_ERROR(logger_, "Start id %u does not exist for edge id %u", start_id, id);
+      throw std::runtime_error("Start id does not exist");
+    }
+
+    if (graph_to_id_map.find(end_id) == graph_to_id_map.end()) {
+      RCLCPP_ERROR(logger_, "End id of %u does not exist for edge id %u", end_id, id);
+      throw std::runtime_error("End id does not exist");
```
SteveMacenski, line 126: nav2_core::NoValidGraph

SteveMacenski: (and above)

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+bool GraphLoader::loadGraphFromFile(
+  Graph & graph,
+  GraphToIDMap & graph_to_id_map,
+  std::string filepath)
+{
+  if (filepath.empty()) {
+    RCLCPP_INFO(
```
SteveMacenski, line 77: Debug log - since this is what happens on bringup and isn't an error as much as it is just a different way of using the function. I don't think an end users needs to see this on screen

SteveMacenski (commented):
There's a couple of them from before you missed (probably didn't expand)

SteveMacenski:
Generally looks good now except for your listed TODOs and the merge conflicts

nav2_route/include/nav2_route/graph_loader.hpp
```diff
@@ -69,9 +70,18 @@ class GraphLoader
...
+   * @brief Transform the graph coordinates to the route frame
+   * @param[in/out] graph The graph to be transformed
+   * @return True if transform was successful
+   */
+  bool transformGraph(Graph &graph);
+
   std::string route_frame_, graph_filepath_;
   std::shared_ptr<tf2_ros::Buffer> tf_;
   rclcpp::Logger logger_{rclcpp::get_logger("GraphLoader")};
+  nav2_util::LifecycleNode::SharedPtr node_;
```
SteveMacenski, line 84: Don't store this - it doesn't look like you actually use it outside of the constructor anyway. If you must, store it as a weakptr, but I don't think you do

nav2_route/src/graph_loader.cpp
```diff
@@ -80,24 +81,63 @@ bool GraphLoader::loadGraphFromFile(
...
 
   // Convert all coordinates to `frame` (in a new method) for standardization
   // Including conversion of GPS coordinates, so we can populate it in some
   // cartesian frame necessary for traversal cost estimation and densifying
   // (and so we don't need to propogate it through our structures)
 }
+
+bool GraphLoader::transformGraph(Graph &graph)
+{
+  //TODO(jw): Can we assume one frame for all?
```
SteveMacenski, line 110: Nope :frowning: 

What you can do though is store a `std::unordered_map<std::string, tf2::Transform>` so that you can lookup a transform from `my_frame` to `route_frame` and store it so we don't lookup for each node, only each **unique** node. Otherwise, we can use the cached transform from a previous node to transform the point

SteveMacenski: So basically everything you have here is right, it just needs to be reshuffled and add that map caching. You don't need to store that in the class though - just locally in the function which is deleted out of scope is fine

nav2_route/include/nav2_route/types.hpp
```diff
@@ -184,6 +184,13 @@ struct Node
...
   SearchState search_state;  // State maintained by route search algorithm
 
+  void addEdge(
+    EdgeCost & cost, NodePtr node, unsigned int edgeid, Metadata & Metadata,
+    Operations & Operations)
+  {
+    neighbors.push_back({edgeid, this, node, cost, Metadata, Operations});
+  }
+
   void addEdge(EdgeCost & cost, NodePtr node, unsigned int edgeid)
```
SteveMacenski, line 194: You can remove this implementation if you make your other one have default values for the metadata / operations arguments (eg `{}`) if not provided. Then both ways would work.

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  graph.resize(nodes.size());
+  addNodesToGraph(graph, graph_to_id_map, nodes);
+  addEdgesToGraph(graph, graph_to_id_map, edges);
+
+  return true;
+}
+
+bool GeoJsonGraphFileLoader::fileExists(const std::string & filepath)
```
SteveMacenski, line 66: I suppose if we're not going to validate it, we can remove this from the pluginlib API and just have it in the loader object itself. I thought we'd do more with this with validating the file, but you mentioned that it would throw if fails to load something properly at parse-time so I think its OK just to validate the file exists (which doesn't need to be reimplemented by the plugins)

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,138 @@
...
+  // Convert all coordinates to `frame` (in a new method) for standardization
+  // Including conversion of GPS coordinates, so we can populate it in some
+  // cartesian frame necessary for traversal cost estimation and densifying
+  // (and so we don't need to propogate it through our structures)
+}
+
+bool GraphLoader::transformGraph(Graph & graph)
+{
+  for (auto & node : graph) {
+    std::string node_frame = node.coords.frame_id;
```
SteveMacenski, line 110: Check if its empty string too to skip in the `continue` line

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  const auto & coordinates = node["geometry"]["coordinates"];
+  coords.x = coordinates[0];
+  coords.y = coordinates[1];
+
+  return coords;
+}
+
+Metadata GeoJsonGraphFileLoader::getMetaData(const Json & properties, const std::string & key)
```
SteveMacenski, line 144: The key is never set to anything else. Why not just have `properties.contains("metadata")` like we have hardcoded for other things like frames / cost / etc?

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,235 @@
...
+      if (data.value().is_number_integer()) {
+        int value = data.value();
+        metadata.setValue(data.key(), value);
+        continue;
+      } else if (data.value().is_number_float()) {
+        float value = data.value();
+        metadata.setValue(data.key(), value);
+        continue;
+      } else {
+        double value = data.value();
```
SteveMacenski, line 168: Missing unsigned https://json.nlohmann.me/api/basic_json/is_number_unsigned/

Also, double doesn't appear to be a type (its going to be a float) in the library, so if you rearrange this if/else so that the final catch-all `else` processes `float` it will handle both float and doubles

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,235 @@
...
+  if (!properties.contains(key)) {return metadata;}
+
+
+  for (const auto & data : properties[key].items()) {
+    if (data.value().is_object() ) {
+      Metadata new_metadata = getMetaData(properties[key], data.key());
+      metadata.setValue(data.key(), new_metadata);
+    }
+
+    if (!data.value().is_primitive()) {continue;}
```
SteveMacenski, line 156: https://json.nlohmann.me/api/basic_json/is_array/ is_array can deal with vector types for zones / polygons / other vector-based things!

SteveMacenski: If you check if its an array after `is_object` and before `is_primitive`, you can catch that in the natural place to load up a vector

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
jwallace42: It is needed for the recursive calls.

SteveMacenski: Oh I see, I missed one of the calls. Got it!

nav2_route/graphs/geojson/aws_graph.geojson
jwallace42: I added this to the aws_graph.geojson file.

SteveMacenski: What about the timestamp? Not just the date

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,127 @@
...
+  std::shared_ptr<tf2_ros::Buffer> tf,
+  const std::string frame)
+: plugin_loader_("nav2_route", "nav2_route::GraphFileLoader"),
+  default_plugin_id_("GeoJsonGraphFileLoader")
+{
+  logger_ = node->get_logger();
+  tf_ = tf;
+  route_frame_ = frame;
+
+  if (!node->has_parameter("graph_filepath")) {
```
SteveMacenski, line 34: Where was this declared?

SteveMacenski: This should be declared and grabbed here. 

```
    node, "graph_filepath", rclcpp::ParameterType::PARAMETER_STRING);
  graph_filepath_ = node->get_parameter("graph_filepath").as_string();
```

SteveMacenski: The `get` will throw if it wasn't declared so you don't need to manually throw

nav2_route/src/graph_loader.cpp
```diff
@@ -0,0 +1,127 @@
...
+  return true;
+}
+
+bool GraphLoader::transformGraph(Graph & graph)
+{
+  for (auto & node : graph) {
+    std::string node_frame = node.coords.frame_id;
+    if (node_frame.empty() || node_frame == route_frame_) {continue;}
+
+    if (cashed_transforms_.find(node_frame) == cashed_transforms_.end()) {
```
SteveMacenski, line 102: `cached` is misspelled. Also, this doesn't need to be a class member, it can be constructed here and gone out of scope in case the transformations are different on different file loadings. We except loading a new file to take a little bit of time since its not a steady-state change so we can reconstruct it each time OK

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,242 @@
...
+  addNodesToGraph(graph, graph_to_id_map, nodes);
+  addEdgesToGraph(graph, graph_to_id_map, edges);
+  return true;
+}
+
+bool GeoJsonGraphFileLoader::doesFileExist(const std::string & filepath)
+{
+  return std::filesystem::exists(filepath);
+}
+
```
SteveMacenski, line 71: Extra space

SteveMacenski: Generally: make sure to run the linters

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,242 @@
...
+Coordinates GeoJsonGraphFileLoader::convertCoordinatesFromJson(const Json & node)
+{
+  Coordinates coords;
+  const auto & properties = node["properties"];
+  if (properties.contains("frame")) {
+    coords.frame_id = properties["frame"];
+  }
+
+  const auto & coordinates = node["geometry"]["coordinates"];
```
SteveMacenski, line 138: Will the json throw or crash if there was, for instance, only a single point set in this vector and we try to access `[1]`?

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  return operation;
+}
+
+Operations GeoJsonGraphFileLoader::convertOperationsFromJson(const Json & properties)
+{
+  Operations operations;
+  if (properties.contains("operations")) {
+    for (const auto & json_operation : properties["operations"]) {
+      std::cout << "getting operations" << std::endl;
```
SteveMacenski, line 217: print :-)

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
```diff
@@ -0,0 +1,242 @@
...
+      Operation operation;
```
SteveMacenski, line 218:
```suggestion
      operations.push_back(convertOperationFromJson(json_operation));
```

All in 1 line, no need for 3 lines

nav2_route/graphs/geojson/aws_graph.geojson
SteveMacenski: Or is `urn:ogc:def:crs:EPSG::3857` some unique hash?

SteveMacenski (changes requested):

nav2_route/src/plugins/graph_file_loaders/geojson_graph_file_loader.cpp
jwallace42: Yep.

nav2_route/graphs/geojson/aws_graph.geojson
jwallace42: I was following the same pattern as the lattice primitive files. I updated it to include the timestamp. 

That is the type of coordinate reference system so it is not a unique hash.

SteveMacenski (approved):

PR #3401, opened 2023-02-09
Added missing conditions to nav2_tree_nodes

muttistefano:
Added the following conditions to nav2_tree_nodes:

WouldAControllerRecoveryHelp
WouldAPlannerRecoveryHelp
WouldASmootherRecoveryHelp
AreErrorCodesPresent

### DIFF
--- a/nav2_behavior_tree/nav2_tree_nodes.xml
+++ b/nav2_behavior_tree/nav2_tree_nodes.xml
@@ -261,6 +261,7 @@
 
     <Condition ID="AreErrorCodesPresent">
       <input_port name="error_code">Error code</input_port>
+      <input_port name="error_codes_to_check">Error codes to check, user defined</input_port>
     </Condition>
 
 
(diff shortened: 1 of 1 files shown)


nav2_behavior_tree/nav2_tree_nodes.xml
```diff
@@ -247,6 +247,23 @@
...
+
+    <Condition ID="WouldAPlannerRecoveryHelp">
+      <input_port name="error_code">Error code</input_port>
+    </Condition>
+
+    <Condition ID="WouldASmootherRecoveryHelp">
+      <input_port name="error_code">Error code</input_port>
+    </Condition>
+
+    <Condition ID="AreErrorCodesPresent">
```
SteveMacenski, line 262: Missing the input port of `error_codes_to_check`

SteveMacenski (changes requested):

nav2_behavior_tree/nav2_tree_nodes.xml
SteveMacenski: This is not used by the other 3, but this is the base class that allows users to use it on its own so they can specify arbitrary numbers

SteveMacenski (approved):

Issue #3325, opened 2022-12-14
Mixing Wall time with simulation time in Costmap2DROS::mapUpdateLoop


AzaelCicero:
## Bug report

**Required Info:**

- Operating System: 
  - All
- ROS2 Version: 
  - All
- Version or commit hash:
  - All 
- DDS implementation:
  - All

#### Steps to reproduce issue

Just use costmap_2d node with simulation time, which has very different rate than wall time. You will observe that HZ of map update and map publish is not reaching the expected value and are far off the target. 

#### Expected behavior
Configured rates of update and publish should be achieved in respect to simulation time.

#### Actual behavior
Map update is reaching expected rate in respect to wall time. Publish rate is bounded to the update rate, as it can only occur during the update, yet it is triggered in respect to simulation time. Depending on the mismatch of the rate between simulation and wall time you will observe different HZ of update and publish. 

#### Additional information
Code should be written in respect to one time source, in this case I think that simulation time is better suited. 

https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_ros.cpp#L461-L498

SteveMacenski:
@AzaelCicero A PR is always appreciated. 

> For local costmap, updateMap() execution time took ~1.0 sec, for global costmap - ~1.5 sec.

If the issue is that the actual updating of the costmap is too long, that's not a problem with clocks / timers. That's trying to do too much "stuff" to be faster than real-time algorithmically. 

> sleep_time = 1/frequency - (cycle_end - cycle_start)

The ROS rate should be doing that for us to keep a regular frequency - unless there's something wrong with that.

AlexeyMerzlyakov:
> The ROS rate should be doing that for us to keep a regular frequency - unless there's something wrong with that.

Indeed the `WallRate` already does the same logic I've supposed in previous message. From [rate.hpp](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/rate.hpp):
```
  virtual bool
  sleep()
  {
    // Time coming into sleep
    auto now = Clock::now();
    // Time of next interval
    auto next_interval = last_interval_ + period_;
    ...
    // Calculate the time to sleep
    auto time_to_sleep = next_interval - now;
    ...
    // Sleep (will get interrupted by ctrl-c, may not sleep full time)
    rclcpp::sleep_for(time_to_sleep);
    return true;
  }
```
So, that is my oversight.

[...]

PR #3404, opened 2023-02-10
Costmap update in accordance to ROS time

AlexeyMerzlyakov:
This is continue of #3327 ticket, supplemented by review items resolved.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3325 |
| Robotic platform tested on | TB3 Gazebo simulation; `colcon test --packages-select nav2_costmap_2d nav2_system_tests` |

---

 - Changed update map frequency control mechanism from WallRate to ROS Timer, so now it respects simulation time if `use_sim_time` parameters is true.
 - Calling `resume()` after `stop()`: the behavior was changed. In default case, we will hang in this routine; while after this PR the execution just returns back from `resume()` when node is stopped. This seems to cause no issue: `resume()` should be called after `pause()`, while `stop()` is chaind with `start()`. So, behavior should be even more correct causing no hangs at this case.
 - Uncommented the "Map update loop missed its desired rate" warning

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -43,6 +43,7 @@
 #include <string>
 #include <vector>
 #include <utility>
+#include <limits>
 
 #include "nav2_costmap_2d/layered_costmap.hpp"
 #include "nav2_util/execution_timer.hpp"
@@ -279,9 +280,22 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
   // Create a thread to handle updating the map
   stopped_ = true;  // to active plugins
   stop_updates_ = false;
-  map_update_thread_shutdown_ = false;
-  map_update_thread_ = std::make_unique<std::thread>(
-    std::bind(&Costmap2DROS::mapUpdateLoop, this, map_update_frequency_));
+
+  if (map_update_timer_ == nullptr) {
+    RCLCPP_DEBUG(get_logger(), "mapUpdateLoop frequency: %lf", map_update_frequency_);
+    if (map_update_frequency_ > 0.) {
+      map_update_timer_ = rclcpp::create_timer(
+        this, get_clock(), rclcpp::Duration::from_seconds(1 / map_update_frequency_),
+        [this]() {mapUpdateLoop();});
+    } else {
+      RCLCPP_WARN(
+        get_logger(),
+        "Costmap won't be periodically updated as frequency  %lf should be greater than zero.",
+        map_update_frequency_);
+    }
+  } else {
+    map_update_timer_->reset();
+  }
 
   start();
 
@@ -302,10 +316,8 @@ Costmap2DROS::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
   stop();
 
   // Map thread stuff
-  map_update_thread_shutdown_ = true;
-
-  if (map_update_thread_->joinable()) {
-    map_update_thread_->join();
+  if (map_update_timer_ != nullptr) {
+    map_update_timer_->cancel();
   }
 
   footprint_pub_->on_deactivate();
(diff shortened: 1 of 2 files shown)


AlexeyMerzlyakov:
> kpawelczyk and others added 2 commits

@SteveMacenski, will this PR preserve original kpawelczyk authorship after merge, or I done something wrong?

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -279,9 +280,22 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
...
+        this, get_clock(), rclcpp::Duration::from_seconds(1 / map_update_frequency_),
+        [this]() {mapUpdateLoop();});
+    } else {
+      RCLCPP_WARN(
+        get_logger(),
+        "Costmap won't be periodically updated as frequency %lf should be greater than zero.",
+        map_update_frequency_);
+    }
+  } else {
+    map_update_timer_->reset();
```
SteveMacenski, line 297: if `map_update_timer_` its a nullptr, why reset it?

SteveMacenski: I think that may actually be dereferencing a nullptr

SteveMacenski:
I thought we discussed that having the timer had other issues and that we were just going to update the rates away from `wall` to use simulation time? A `rclcpp::Rate` object instead of `rclcpp::WallRate` would do the job. Unless you figured out the Timer / callback starvation issue?

AlexeyMerzlyakov:
> A `rclcpp::Rate` object instead of `rclcpp::WallRate` would do the job. Unless you figured out the Timer / callback starvation issue?

It seems like `rclcpp::Rate` is another wrapper for `GenericRate` clock, but instead of `WallRate`, it uses steady clock:
https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/rate.hpp#L115
So, this won't be a solution in this case, since we need to respect simulation ROS clock.

---------------------

However, this idea made me a think about the accuracy of solution I've [presented](https://github.com/ros-planning/navigation2/issues/3325#issuecomment-1352552264) before. After close double-check, I've discovered that this solution was using periodic granular `std::this_thread::sleep_for(10ms);` to wait, which gave us inaccuracies reported [here](https://github.com/ros-planning/navigation2/issues/3325#issuecomment-1424379453). Instead of this, it is better to copy the RCLCPP's [solution](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/rate.hpp#L59), but instead of using `rclcpp::sleep_for(chrono time_to_sleep)` it should be used `[rclcpp::Clock::SharedPtr]clock_->sleep_for(time_to_sleep);`. This is an implementation of our own `nav2_utils::ROSTimer`, as latest experiments shown, has approximately the same time accuracy as proposed in this ticket `rclcpp::timer` approach.

> I thought we discussed that having the timer had other issues and that we were just going to update the rates away from `wall` to use simulation time? 

As of experiments, it seems that accuracy of this approach somewhat similar to having our own `nav2_utils::ROSTimer`.  However, if we will load the main thread with heavy PCL-calculation algorithms, your words might be confirmed, I've just never checked it.

From the said above, it looks like the solution to have a correct copy of a `rclcpp::WallRate` implementation in a `nav2_utils::ROSTime` still have to take place. @SteveMacenski, if you will do a green light, I'll change this PR to `nav2_utils::ROSTime` implementation.

SteveMacenski:
Interesting. `rclcpp::Timer` uses the clock though https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/timer.hpp which can be set to ROS time. I'm surprised that Rate doesn't have that option. It might be worth filing a ticket with rclcpp to have the Rate be able to respect ROS time, which I looked into and I think could be possible if we made the clock a parameter option for construction and use. My guess is it wasn't done for two reasons: (1) this was an early feature of rclcpp and ROS time may not have been fully baked yet and (2) ROS times can stop or reverse so it complicates things if you're trying to retain a regularized rate of execution. `Rate` will block a thread versus the `Timer` which adds things to the callback queue.

> Instead of this, it is better to copy the RCLCPP's [solution](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/rate.hpp#L59), but instead of using rclcpp::sleep_for(chrono time_to_sleep) it should use rclcpp::Clock->sleep_for(time_to_sleep);

Maybe its worth actually just changing the Rate in rclcpp to support what we need instead? I have a call with the ROS Maintainers later today, I'll inquire about the interest of this.

> nav2_utils::ROSTimer

Can you show me this "nav2_utils::ROSTimer"? I don't see it in the stack.

AlexeyMerzlyakov:
> Can you show me this "nav2_utils::ROSTimer"? I don't see it in the stack.

I've just copied the implementation of `rclcpp::GenericTimer` and moved it on ROS time rails. It is a prototype yet, but here we are:
```
class ROSRate
{
public:
  ROSRate(double rate, rclcpp::Clock::SharedPtr clock)
  : ROSRate(rclcpp::Duration::from_seconds(1.0 / rate), clock)
  {}
  ROSRate(rclcpp::Duration period, rclcpp::Clock::SharedPtr clock)
  : period_(period), last_interval_(clock->now()), clock_(clock)
  {}

  bool sleep()
  {
    // Time coming into sleep
    rclcpp::Time now = clock_->now();
    // Time of next interval
    rclcpp::Time next_interval = last_interval_ + period_;
    // Detect backwards time flow
    if (now < last_interval_) {
      next_interval = now + period_;
    }

    // Update the interval
    last_interval_ += period_;

    // If time of next interval is earlier than now, don't sleep
    if (next_interval <= now) {
      // If an entire cycle was missed then reset next interval.
      // This might happen if the loop took more than a cycle.
      // Or if time jumps forward.
      if (now > next_interval + period_) {
        last_interval_ = now + period_;
      }
      // Either way do not sleep and return false
      return false;
    }

    // Calculate the time to sleep
    const rclcpp::Duration time_to_sleep = next_interval - now;
    // Sleep
    clock_->sleep_for(time_to_sleep);

    return true;
  }

  void reset()
  {
    last_interval_ = clock_->now();
  }

  rclcpp::Duration period() const
  {
    return period_;
  }

protected:
  rclcpp::Duration period_;
  rclcpp::Time last_interval_;
  rclcpp::Clock::SharedPtr clock_;
};
```

> I have a call with the ROS Maintainers later today, I'll inquire about the interest of this.

Any news from RCLCPP team? If so, I am ready to submit a new feature request there.

nav2_costmap_2d/src/costmap_2d_ros.cpp
AlexeyMerzlyakov: If `map_update_timer_` is a nullptr, we are trying to create a new timer, else - resetting existing timer:
```
if (!map_update_timer_) {
  // map_update_timer_ == nullptr
  try to create a new map_update_timer_ object
} else {
  // map_update_timer_ != nullptr
  // which means that timer is already created. We just need to reset it:
  map_update_timer_->reset();
}
```
I looks like there are no cases where we could dereference a null pointer.

SteveMacenski:
See Slack - I got some feedback and resources from Tully and Chris about how to contribute a ROS Time version of Rate into rclcpp. There's interest in it so we don't have to maintain it here

tonynajjar:
Could you please link to the continuation of this? Is there a PR for rclcpp?

AlexeyMerzlyakov:
Yes, the proper way - is to add ROS-respective rate to RCLCPP API. The according PR is: https://github.com/ros2/rclcpp/pull/2123

SteveMacenski:
What's the story here? Is the change being looked at for Rolling now?

AlexeyMerzlyakov:
I've just created a new PR https://github.com/ros2/system_tests/pull/516 in `system_tests`, fixing the test regressions I've noted in logs from CI verificaiton that was failed in https://github.com/ros2/rclcpp/pull/2123. After PR516 will be merged, I expect no more test regressions in PR2123 appeared with new `rclcpp::Rate` API.

SteveMacenski:
OK! Let me know if I can help

SteveMacenski:
Yay! rclcpp is merged!

SteveMacenski:
@AlexeyMerzlyakov from your analysis in https://github.com/ros-planning/navigation2/issues/3325#issuecomment-1424379453 with the higher variation of executions, is this ready to go (e.g. not suffering from this problem)? 

I'm looking to address the timing issues myself this week, it looks like you covered this one but can you verify this doesn't suffer from that problem you previously mentioned and that this is good to go?

AlexeyMerzlyakov:
This PR is used as an alternative for updated `rclcpp::Rate` API, since costmaps are using `rclcpp::WallRate`, as you've mentioned. So, ` rclcpp::WallRate r(frequency);` could be replaced to `rclcpp::Rate r(this->get_clock(), frequency);` and this PR is not needed.

Regarding the https://github.com/ros-planning/navigation2/issues/3325#issuecomment-1424379453 variations accuracy loosing - this was related to incorrect waiting grain approach in proposed by https://github.com/ros-planning/navigation2/issues/3325#issuecomment-1352552264 solution
```
std::this_thread::sleep_for(10ms);  // Sleep grain is 10ms
```
which was finally refused. So this is not an our case for now.

SteveMacenski:
Ok got it! Closing in favor of the new API then, thanks for the clarification and your help on this issue!

Issue #3402, opened 2023-02-09
[Collision Monitor] Performance issue


HovorunB:
Hello! Recently when using the collision monitor I discovered that it adds too much latency to the cmd_vel path. The problem to that was waiting for the transform from data source frame to base frame on every callback. I saw that you are using an [Advanced API of lookup_transform](https://docs.ros.org/en/foxy/Tutorials/Intermediate/Tf2/Time-Travel-With-Tf2-Cpp.html#advanced-api-for-lookuptransform) which relies not only on getting the transform data_source_frame->base_frame but also at fixed_frame->base_frame. This in our case adds latency (we publish fixed_frame->base_frame at 30Hz which adds at most ~33ms to the latency) Publishing fixed_frame->base_frame faster will reduce the latency but i feel like this is more of a workaround rather than a solution. 

I wanted to ask whether this is expected and collision monitor could provide such default latency. Maybe you have some thoughts on how to make it run faster?

HovorunB:
Thank you for the reply!

> Please provide some reference to your claim that `relies not only on getting the transform data_source_frame->base_frame but also at fixed_frame->base_frame` and what specific lines you're pointing to. 

I'm pointing to [This implimentation](https://github.com/ros2/geometry2/blob/599fa7c5dd050fd0978d7b1b4a9fca0ebb60410b/tf2/src/buffer_core.cpp#L598)
which is called from [nav2_collision_monitor/src/pointcloud.cpp](https://github.com/ros-planning/navigation2/blob/4aaefad1162d5edcba26878cdb4a2f11322ada99/nav2_collision_monitor/src/pointcloud.cpp#L83)   and then [robot_utils.cpp](https://github.com/ros-planning/navigation2/blob/50f0e0ec36a04f7f018abe02d846a10b56eb15e3/nav2_util/src/robot_utils.cpp#L112)

The way I understand is that it was made like this in order to publish relevant information about the obstacle:
Imagine robot that has a camera on the front side is driving with high speed towards the obstacle. Then it receives sensor data that the obstacle is 2m away. By the time e.g. pointcloud is published into robot's base frame the obstacle would be already closer than 2m away. Therefore we pass fixed_frame to lookup_transform in order to publish the pointcloud relative to the base frame of robot not at the current time, but at the time the pointcloud was captured. 

> I know this isn't precisely what you said, but are you sure that the issue here isn't that your static transforms from base_frame -> sensor_frame isn't actually throttling the system to 30 hz? 

[...]

PR #3412, opened 2023-02-16
Option allowing to use simple lookupTransform API

AlexeyMerzlyakov:
ignoring time shifts between source and base frame during the movement
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3402 |
| Robotic platform tested on | TB3 Gazebo simulation, `colcon test --packages-select nav2_collision_monitor`, code coverage |

---

* Added an new option allowing to control which `lookupTransform()` API to use: [simple](https://docs.ros2.org/latest/api/tf2/classtf2_1_1BufferCore.html#a5d578ecbea2ec0fdb92de1616251e175) or [advanced](https://docs.ros2.org/latest/api/tf2/classtf2_1_1BufferCore.html#a779c28b06b41f9ebc63e8c0d0da77e9b), considering considering the difference between current time and latest source time and allowing to correct source data towards to base frame movement.
* Using advanced `lookupTransform()` API will lead to better accuracy in cost of performance (strongly recommended for fast moving robots)
* Using simple `lookupTransform()` API will lower the accuracy but will improve the performance for `~1/(2*odom_rate)` per each `cmd_vel` calculation cycle, which might be suitable for slow robots.
* Using of simple `lookupTransform()` API is being covered by `testIgnoreTimeShift` testcase (advanced `lookupTransform()` API is already covered by previous tests)

* Added new parameter on `Collision Monitor` configuration page

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/pointcloud.cpp
+++ b/nav2_collision_monitor/src/pointcloud.cpp
@@ -77,7 +77,6 @@ void PointCloud::getData(
     return;
   }
 
-
   tf2::Transform tf_transform;
   if (base_shift_correction_) {
     // Obtaining the transform to get data from source frame and time where it was received
@@ -92,7 +91,8 @@ void PointCloud::getData(
     }
   } else {
     // Obtaining the transform to get data from source frame to base frame without time shift
-    // considered. Less accurate but much more faster option.
+    // considered. Less accurate but much more faster option not dependent on state estimation
+    // frames.
     if (
       !nav2_util::getTransform(
         data_->header.frame_id, base_frame_id_,
--- a/nav2_collision_monitor/src/pointcloud.cpp
+++ b/nav2_collision_monitor/src/pointcloud.cpp
@@ -31,10 +31,11 @@ PointCloud::PointCloud(
   const std::string & base_frame_id,
   const std::string & global_frame_id,
   const tf2::Duration & transform_tolerance,
-  const rclcpp::Duration & source_timeout)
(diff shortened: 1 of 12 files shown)


AlexeyMerzlyakov:
Local code coverage results:
Screenshot 2023-02-16 at 12-01-45 LCOV - cm info

nav2_collision_monitor/src/pointcloud.cpp
```diff
@@ -76,16 +77,29 @@ void PointCloud::getData(
...
+      !nav2_util::getTransform(
+        data_->header.frame_id, data_->header.stamp,
+        base_frame_id_, curr_time, global_frame_id_,
+        transform_tolerance_, tf_buffer_, tf_transform))
+    {
+      return;
+    }
+  } else {
+    // Obtaining the transform to get data from source frame to base frame without time shift
+    // considered. Less accurate but much more faster option.
```
SteveMacenski, line 95:
```suggestion
    // considered. Less accurate but much more faster option not dependent on state estimation frames.
```

Ditto in other locations - this is mostly a note for myself in the future

SteveMacenski (approved):
LGTM otherwise

nav2_collision_monitor/src/pointcloud.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3412/commits/7e432ecf90f8982cf38f88a527406becf5e9090c

PR #3414, opened 2023-02-17
Rpp curve regulation reference cycle

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Custom: Sim and Real |

---

Introduce option to set lookahead for curvature detection to a fixed distance. This breaks a reference cycle that becomes especially problematic with long lookaheads.

Added two new parameters.

---

Determine if there is a way to derive this metric so that it can be enabled by default for all users

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/README.md
+++ b/nav2_regulated_pure_pursuit_controller/README.md
@@ -119,7 +119,7 @@ controller_server:
       use_cost_regulated_linear_velocity_scaling: false
       regulated_linear_scaling_min_radius: 0.9
       regulated_linear_scaling_min_speed: 0.25
-      use_fixed_curvature_lookahead: true
+      use_fixed_curvature_lookahead: false
       curvature_lookahead_dist: 1.0
       use_rotate_to_heading: true
       rotate_to_heading_min_angle: 0.785
--- a/nav2_regulated_pure_pursuit_controller/README.md
+++ b/nav2_regulated_pure_pursuit_controller/README.md
@@ -71,6 +71,8 @@ Note: The maximum allowed time to collision is thresholded by the lookahead poin
 | `inflation_cost_scaling_factor` | The value of `cost_scaling_factor` set for the inflation layer in the local costmap. The value should be exactly the same for accurately computing distance from obstacles using the inflated cell values | 
 | `regulated_linear_scaling_min_radius` | The turning radius for which the regulation features are triggered. Remember, sharper turns have smaller radii | 
 | `regulated_linear_scaling_min_speed` | The minimum speed for which the regulated features can send, to ensure process is still achievable even in high cost spaces with high curvature. | 
+| `use_fixed_curvature_lookahead` | Enable fixed lookahead for curvature detection. Useful for systems with long lookahead. | 
+| `curvature_lookahead_dist` | Distance to lookahead to determine curvature for velocity regulation purposes. Only used if `use_fixed_curvature_lookahead` is enabled. | 
 | `use_rotate_to_heading` | Whether to enable rotating to rough heading and goal orientation when using holonomic planners. Recommended on for all robot types except ackermann, which cannot rotate in place. | 
 | `rotate_to_heading_min_angle` | The difference in the path orientation and the starting robot orientation to trigger a rotate in place, if `use_rotate_to_heading` is enabled. | 
 | `max_angular_accel` | Maximum allowable angular acceleration while rotating to heading, if enabled | 
@@ -117,6 +119,8 @@ controller_server:
       use_cost_regulated_linear_velocity_scaling: false
       regulated_linear_scaling_min_radius: 0.9
       regulated_linear_scaling_min_speed: 0.25
+      use_fixed_curvature_lookahead: true
+      curvature_lookahead_dist: 1.0
       use_rotate_to_heading: true
       rotate_to_heading_min_angle: 0.785
       max_angular_accel: 3.2
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -193,14 +193,12 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
   double desired_curvature = calculateCurvature(carrot_pose.pose.position);
 
-  double regulation_curvature;
+  double regulation_curvature = desired_curvature;
   if (params_->use_fixed_curvature_lookahead) {
     auto curvature_lookahead_pose = getLookAheadPoint(
       params_->curvature_lookahead_dist,
       transformed_plan);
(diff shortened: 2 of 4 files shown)


nav2_regulated_pure_pursuit_controller/README.md
```diff
@@ -71,6 +71,8 @@ Note: The maximum allowed time to collision is thresholded by the lookahead poin
 | `inflation_cost_scaling_factor` | The value of `cost_scaling_factor` set for the inflation layer in the local costmap. The value should be exactly the same for accurately computing distance from obstacles using the inflated cell values | 
 | `regulated_linear_scaling_min_radius` | The turning radius for which the regulation features are triggered. Remember, sharper turns have smaller radii | 
 | `regulated_linear_scaling_min_speed` | The minimum speed for which the regulated features can send, to ensure process is still achievable even in high cost spaces with high curvature. | 
+| `use_fixed_curvature_lookahead` | Enable fixed lookahead for curvature detection. Useful for systems with long lookahead. |
```
SteveMacenski, line 74: Add to the configuration guide page for RPP + migration guide that this option is now available (and why necessary)

nav2_regulated_pure_pursuit_controller/README.md
```diff
@@ -117,6 +119,8 @@ controller_server:
       use_cost_regulated_linear_velocity_scaling: false
       regulated_linear_scaling_min_radius: 0.9
       regulated_linear_scaling_min_speed: 0.25
+      use_fixed_curvature_lookahead: true
```
SteveMacenski, line 122: Default to false -- mostly since the "usual" user of Nav2 isn't going to be running a tractor :laughing:

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -175,16 +191,16 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
 
   double linear_vel, angular_vel;
 
-  // Find distance^2 to look ahead point (carrot) in robot base frame
-  // This is the chord length of the circle
-  const double carrot_dist2 =
-    (carrot_pose.pose.position.x * carrot_pose.pose.position.x) +
-    (carrot_pose.pose.position.y * carrot_pose.pose.position.y);
+  double desired_curvature = calculateCurvature(carrot_pose.pose.position);
```
SteveMacenski, line 194: I think this is just "path_curvature" or simply "curvature". Desired seems like an odd word choice

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -175,16 +191,16 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
-  // Find curvature of circle (k = 1 / R)
-  double curvature = 0.0;
-  if (carrot_dist2 > 0.001) {
-    curvature = 2.0 * carrot_pose.pose.position.y / carrot_dist2;
+  double regulation_curvature;
+  if (params_->use_fixed_curvature_lookahead) {
+    auto curvature_lookahead_pose = getLookAheadPoint(
+      params_->curvature_lookahead_dist,
+      transformed_plan);
+    regulation_curvature = calculateCurvature(curvature_lookahead_pose.pose.position);
```
SteveMacenski, line 201: if you initialize `regulation_curvature` with `curvature`, then you can remove the `else` statement.

SteveMacenski:
No real complaints -- though CI running is odd. Try rebasing?

nav2_regulated_pure_pursuit_controller/README.md
Aposhian: Yes. Are you ok on this naming of the parameter?

nav2_regulated_pure_pursuit_controller/README.md
Aposhian: Ok will change (the default is actually false, but I will match here)

Aposhian: Done

Aposhian:
Ok, the fork for our main branch must have been polluted when I was trying to run CI. I just rebased on upstream main.

nav2_regulated_pure_pursuit_controller/README.md
SteveMacenski: Yup!

Aposhian:
param guide here: https://github.com/ros-planning/navigation.ros.org/pull/390

SteveMacenski:
Still a few items above, but just details

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Aposhian: I want to disambiguate between the curvature that is used for regulating velocity, and the curvature that is used for calculating the angular velocity ("desired curvature")

Aposhian: Maybe I could rename this to `command_curvature`?

SteveMacenski: Lookahead curvature perhaps

Aposhian: Well, but the curvature used for computing the commanded angular vel is also a lookahead curvature.

SteveMacenski: On L228 that is the current variable named `desired_curvature` which we're looking at renaming, so that seems to line up perfectly :wink:

SteveMacenski: `lookahead_curvature`

Aposhian: For which one? There are two different curvatures at play here and they are both lookahead curvatures.

SteveMacenski: `regulation_curvature` is used by the regulation term for curvature, so I think this is great.

`desired_curvature` is the actual lookahead-point's curvature that is being used to generate velocities and perform collision checking, so I think that should be `lookahead_curvature`. 

The regulation one's parameter implies "lookahead distance" but it is not itself the actual Pure Pursuit lookahead point used by the algorithm. 

Does that make sense? The only change I'm asking for is `desired_curvature` -> `lookahead_curvature`. I know I'll read  `desired` in the future and always wonder what that means in particular: Desired by _whom_, given there are multiple consumers.

Aposhian: Ok that makes sense. I will change it to that!

SteveMacenski (approved):

Issue #3419, opened 2023-02-20
Filter mask receiving time is too high for KeepoutFilter


AlexeyMerzlyakov:
When working with large maps and filter masks (`100m x 150m`), filter mask receiving time takes a long time `~30ms` for `KeepoutFilter`, while for other filters (`SpeedFilter` and `BinaryFilter`), it is imperceptibly small.

This appears because in the mask callback, KeepoutFilter is creating a new `Costmap2D` object from whole filter mask:
```
  // Making a new mask_costmap_
  mask_costmap_ = std::make_unique<Costmap2D>(*msg);
```
The constructor [converts by-item](https://github.com/ros-planning/navigation2/blob/118fb8f1f6e0eb724304b904fa56deb2cd20f31c/nav2_costmap_2d/src/costmap_2d.cpp#L79-L90) each OccupancyGrid cell to Costmap2D cell, which takes long time for large masks.

Instead of it, we could just store the pointer to the `OccupancyGrid` message in the mask callback, and then operate directly with `OccupancyGrid` cells inside dynamic window when it is necessary.

Preliminary experiments show the mask receiving time will drop from `30ms` down to `0.12ms`, while `KeepoutFilter::process()` time will be increased from `0.13ms` to `0.28ms` which is insignificant, as follows from the table below:
| Routine  | current time, us | time after optimization applied, us |
| ------------- | ------------- | --- |
| maskCallback()  | 29869.53 | 116.13 |
| process() | 130.72 | 275.34 |

So, at the expense of small map processing time increase, we may hugely optimize mask receiving callback.
It is very notable for the systems with large maps and filter masks, especially when filter masks to be updated dynamically with moving objects (allowing to use there high update rates).

## Feature request

**Required Info:**

- Operating System:
  - Ubuntu 20.04, 22.04
- ROS2 Version:
  - ROS2 rolling built from sources (source date: 2023.01.10)
- Version or commit hash:
  - 4aaefad1162d5edcba26878cdb4a2f11322ada99
- DDS implementation:
  - Cyclone DDS

#### Steps to reproduce measurements
* Prepare `150x100m` OccupancyGrid map (`3000x2000` cells with `resultion = 0.05`)
* Enable KeepoutFilter in the local costmap and publish this map as input mask for the filter
* Run standard TB3 simulation with enabled KeepoutFilter
[...]

PR #3420, opened 2023-02-21
Improve KeepoutFilter mask receiving performance

AlexeyMerzlyakov:
Do not copy `OccupancyGrid` mask -> to `Costmap2D` in mask receiving callback. Directly store and use `OccupancyGrid` mask instead.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3419 |
| Robotic platform tested on | TB3 simulation, colcon test, local code coverage |

---

* Do not copy OccupancyGrid mask -> to Costmap2D. Directly store and use OccupancyGrid mask instead:
```
-  std::unique_ptr<Costmap2D> mask_costmap_;
+  nav_msgs::msg::OccupancyGrid::SharedPtr filter_mask_;
```
* Add new unsigned char `CostmapFilter::getMaskCost(OccupancyGrid::SharedPtr filter_mask, & mx, & my)` routine to CostmapFilter, which will convert OccupancyGrid values to Costmap2D ones; and will be called each time, when we need to update master costmap cell in the rolling window during `KeepoutFilter::process()`
* Other code to be changed to use OccupancyGrid instead of Costmap

* No documentation changed required

---

## Additional notes

* This PR should be updated and merged after #3418 PR completion.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/costmap_filters/costmap_filter.cpp
+++ b/nav2_costmap_2d/plugins/costmap_filters/costmap_filter.cpp
@@ -42,6 +42,9 @@
 #include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
 #include "geometry_msgs/msg/point_stamped.hpp"
 
+#include "nav2_costmap_2d/cost_values.hpp"
+#include "nav2_util/occ_grid_values.hpp"
+
 namespace nav2_costmap_2d
 {
 
@@ -204,6 +207,24 @@ bool CostmapFilter::worldToMask(
   }
 
   return true;
+}
+
+unsigned char CostmapFilter::getMaskCost(
+  nav_msgs::msg::OccupancyGrid::ConstSharedPtr filter_mask,
+  const unsigned int mx, const unsigned int & my) const
+{
+  const unsigned int index = my * filter_mask->info.width + mx;
+
+  const char data = filter_mask->data[index];
+  if (data == nav2_util::OCC_GRID_UNKNOWN) {
+    return NO_INFORMATION;
+  } else {
+    // Linear conversion from OccupancyGrid data range [OCC_GRID_FREE..OCC_GRID_OCCUPIED]
+    // to costmap data range [FREE_SPACE..LETHAL_OBSTACLE]
+    return std::round(
+      static_cast<double>(data) * (LETHAL_OBSTACLE - FREE_SPACE) /
+      (nav2_util::OCC_GRID_OCCUPIED - nav2_util::OCC_GRID_FREE));
+  }
 }
 
 }  // namespace nav2_costmap_2d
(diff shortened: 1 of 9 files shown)


nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
```diff
@@ -276,9 +277,9 @@ void KeepoutFilter::process(
         msk_wx = gl_wx;
         msk_wy = gl_wy;
       }
-      // Get mask coordinates corresponding to (i, j) point at mask_costmap_
-      if (mask_costmap_->worldToMap(msk_wx, msk_wy, mx, my)) {
-        data = mask_costmap_->getCost(mx, my);
+      // Get mask coordinates corresponding to (i, j) point at filter_mask_
+      if (worldToMask(filter_mask_, msk_wx, msk_wy, mx, my)) {
+        data = getMaskCost(filter_mask_, mx, my);
```
SteveMacenski, line 282: Was the previous object already having its values transformed both in coordinates (worldToMask) and in value (getMaskCost)? Just making sure this is a 1:1 correct reinterpretation.

AlexeyMerzlyakov: Yes, previous `mask_costmap_` was an `Costmap2D` object made from input `OccupancyGrid` mask message. This constructor [copies](https://github.com/ros-planning/navigation2/blob/63b690a27a4ba1bde24ebd8e6c592c9894eec169/nav2_costmap_2d/src/costmap_2d.cpp#L68-L72) all origin, size, frame, etc... information from occupancy grid, what now is `filter_mask_`, so `worldToMask()` will return the same coordinates for `filter_mask_` as `worldToMap()` for `filter_mask_`. The same story for `getMaskCost()` v.s. `getCost()` return value.

So, this should be 1:1 conversion. This also confirms by keepout test which keeps going after patch, which has a [case](https://github.com/ros-planning/navigation2/blob/63b690a27a4ba1bde24ebd8e6c592c9894eec169/nav2_costmap_2d/test/unit/keepout_filter_test.cpp#L452) for mask and map are in different frames and have different size.

nav2_costmap_2d/plugins/costmap_filters/costmap_filter.cpp
```diff
@@ -206,4 +209,22 @@ bool CostmapFilter::worldToMask(
...
+{
+  const unsigned int index = my * filter_mask->info.width + mx;
+
+  const char data = filter_mask->data[index];
+  if (data == nav2_util::OCC_GRID_UNKNOWN) {
+    return NO_INFORMATION;
+  } else {
+    // Linear conversion from OccupancyGrid data range [OCC_GRID_FREE..OCC_GRID_OCCUPIED]
+    // to costmap data range [FREE_SPACE..LETHAL_OBSTACLE]
+    return std::round(
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: Unlike world->occupancy_grid coordinates conversion, this rounding is designed to find the _nearest_ integer value in `[0..254]` Costmap value space when making a linear conversion from `[0..100]` OccupancyGrid values. So, this should be OK.

SteveMacenski (approved):

PR #3425, opened 2023-02-23
Fix using different frame for global and local costmap

doisyg:
It is currently impossible to use a different global frame for the global and local costmap.
The use of `worldToMap` to check if the pose is inside the costmap works only if the pose to check is referenced relative to the costmap frame: https://github.com/ros-planning/navigation2/blob/9dca1cfa60af9cdc4f6111db960bae72fd85b17c/nav2_mppi_controller/src/path_handler.cpp#L66

Proposed fix is to replace `transformPlanPosesToCostmapFrame` and `getGlobalPlanConsideringBounds` by `getGlobalPlanConsideringBoundsInCostmapFrame` in order not to have to transform the pose from the global plan/global costmap frame to the controller/local costmap frame twice.

### DIFF
--- a/nav2_mppi_controller/src/path_handler.cpp
+++ b/nav2_mppi_controller/src/path_handler.cpp
@@ -57,9 +57,9 @@ PathHandler::getGlobalPlanConsideringBoundsInCostmapFrame(
       return euclidean_distance(global_pose, ps);
     });
 
-  nav_msgs::msg::Path transformedPlan;
-  transformedPlan.header.frame_id = costmap_->getGlobalFrameID();
-  transformedPlan.header.stamp = global_pose.header.stamp;
+  nav_msgs::msg::Path transformed_plan;
+  transformed_plan.header.frame_id = costmap_->getGlobalFrameID();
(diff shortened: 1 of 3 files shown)


nav2_mppi_controller/src/path_handler.cpp
```diff
@@ -55,19 +57,38 @@ PathRange PathHandler::getGlobalPlanConsideringBounds(
       return euclidean_distance(global_pose, ps);
     });
 
-  // Find the furthest relevent point on the path to consider within costmap
-  // bounds
-  const auto * costmap = costmap_->getCostmap();
+  nav_msgs::msg::Path transformedPlan;
```
SteveMacenski, line 60:
```suggestion
  nav_msgs::msg::Path transformed_plan;
```

that's out of style for the project

SteveMacenski (approved):
Otherwise LGTM functionally!

SteveMacenski (approved):

SteveMacenski (approved):

PR #3431, opened 2023-02-28
Add flag to not send request in BTServiceNode

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -45,7 +45,7 @@ public:
   BtServiceNode(
     const std::string & service_node_name,
     const BT::NodeConfiguration & conf)
-  : BT::ActionNodeBase(service_node_name, conf), service_node_name_(service_node_name)
+  : BT::ActionNodeBase(service_node_name, conf), service_node_name_(service_node_name), should_send_request_(true)
   {
     node_ = config().blackboard->template get<rclcpp::Node::SharedPtr>("node");
     callback_group_ = node_->create_callback_group(
@@ -128,10 +128,17 @@ public:
   BT::NodeStatus tick() override
   {
     if (!request_sent_) {
+      // reset the flag to send the goal or not, allowing the user the option to set it in on_tick
+      should_send_request_ = true;
+
+      // user defined callback, may modify "should_send_request_".
       on_tick();
-      future_result_ = service_client_->async_send_request(request_).share();
-      sent_time_ = node_->now();
-      request_sent_ = true;
+
+      if (should_send_request_) {
+        future_result_ = service_client_->async_send_request(request_).share();
+        sent_time_ = node_->now();
+        request_sent_ = true;
+      }
     }
     return check_future();
   }
@@ -240,6 +247,9 @@ protected:
   std::shared_future<typename ServiceT::Response::SharedPtr> future_result_;
   bool request_sent_{false};
   rclcpp::Time sent_time_;
+
+  // Can be set in on_tick or on_wait_for_result to indicate if a goal should be sent.
+  bool should_send_request_;
 };
 
 }  // namespace nav2_behavior_tree


nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -240,6 +247,9 @@ class BtServiceNode : public BT::ActionNodeBase
   std::shared_future<typename ServiceT::Response::SharedPtr> future_result_;
   bool request_sent_{false};
   rclcpp::Time sent_time_;
+
+  // Can be set in on_tick or on_wait_for_result to indicate if a request should be sent.
+  bool should_send_request_;
```
tonynajjar, line 252: Would you rather this be called similarly as in the BTActionNode (should_send_goal_) or better it be semantically more correct?

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -128,10 +128,17 @@ class BtServiceNode : public BT::ActionNodeBase
...
-      sent_time_ = node_->now();
-      request_sent_ = true;
+
+      if (should_send_request_) {
+        future_result_ = service_client_->async_send_request(request_).share();
+        sent_time_ = node_->now();
+        request_sent_ = true;
+      }
     }
     return check_future();
```
tonynajjar, line 143: I have a small doubt what to do here. if `should_send_request_` then `future_result_` is null and we can't `check_future`. What should we do in this case? Should the Node return Failure?

SteveMacenski: I'd log that the node set that it was not valid to send request and then return `BT::NodeStatus::FAILURE`

SteveMacenski:
any update?

SteveMacenski:
any update?

tonynajjar:
Internal ticket is still on hold (while I jumped on MPPI :smile:), I'll have a quick look at it tomorrow again if it's a quick thing

tonynajjar:
I went with you proposal of directly failing

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -131,10 +131,20 @@ class BtServiceNode : public BT::ActionNodeBase
...
       on_tick();
+      
+      if(!should_send_request_){
+        return BT::NodeStatus::FAILURE;
+      }
+
       future_result_ = service_client_->async_send_request(request_).share();
       sent_time_ = node_->now();
       request_sent_ = true;
+
```
SteveMacenski, line 147: remove extra line, linter won't like that

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -131,10 +131,20 @@ class BtServiceNode : public BT::ActionNodeBase
   BT::NodeStatus tick() override
   {
     if (!request_sent_) {
+      // reset the flag to send the request or not, allowing the user the option to set it in on_tick
+      should_send_request_ = true;
+
+      // user defined callback, may modify "should_send_request_".
       on_tick();
+      
+      if(!should_send_request_){
```
SteveMacenski, line 140:
```suggestion
      if (!should_send_request_) {
```

Linter no likey

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -48,7 +48,7 @@ class BtServiceNode : public BT::ActionNodeBase
     const BT::NodeConfiguration & conf,
     const std::string & service_name = "")
   : BT::ActionNodeBase(service_node_name, conf), service_name_(service_name), service_node_name_(
-      service_node_name)
+      service_node_name), should_send_request_(true)
```
SteveMacenski, line 51:
```suggestion
      service_node_name)
```

We set it in the start of tick, this is not necessary

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -196,9 +196,11 @@ class BtActionNode : public BT::ActionNodeBase
       // user defined callback, may modify "should_send_goal_".
       on_tick();
 
-      if (should_send_goal_) {
-        send_new_goal();
+      if(!should_send_goal_){
```
SteveMacenski, line 199: Fix linting.

SteveMacenski (changes requested):

SteveMacenski:
Do your changes to the action node also fix https://github.com/ros-planning/navigation2/issues/3434?

tonynajjar:
> Do your changes to the action node also fix #3434?

yes

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -131,7 +131,17 @@ class BtServiceNode : public BT::ActionNodeBase
...
+      // reset the flag to send the request or not,
+      // allowing the user the option to set it in on_tick
+      should_send_request_ = true;
+
+      // user defined callback, may modify "should_send_request_".
       on_tick();
+
+      if (!should_send_request_) {
+        return BT::NodeStatus::FAILURE;
+      }
```
tonynajjar, line 143: Looking back at this @SteveMacenski, we now have the use case where we want to not send a request and succeed. E.g. the BT node sees that we are already in the target state, so no need to send a request and succeed.

Would is make sense to implement something like:

```
      // user defined callback, may modify "should_fail_without_sending_goal_" and "should_succeed_without_sending_goal_".
      on_tick();

      if (should_fail_without_sending_goal_) {
      }
      if (should_succeed_without_sending_goal_) {
      }     
```

SteveMacenski: I think that should be handled in a different BT node, if you read the BTs for AI and Robotics book, the common design pattern is to have a conditional node if a particular objective is complete to skip doing the actual task (e.g. a branch of the tree with a "if ball not in hand, --> then pick up ball" so that if the conditional node checking if the ball is in the hand will trigger or not trigger picking up the ball action)

If that's the case - which it also sounds like you may have (?) the return code for this doesn't matter because that BT node wouldn't be triggered if under the appropriate control flow node.

PR #3432, opened 2023-02-28
Allow undeclared parameters in BT node

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | N/A) |

---

I'm using ROS parameters in my custom BT nodes and in some cases it makes sense to allow for undeclared parameters. For example I have a BT node that can have a variable number of parameters. While this might be specific to my BT nodes, I don't see the harm allowing undeclared parameters.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -96,7 +96,8 @@ bool BtActionServer<ActionT>::on_configure()
   std::string client_node_name = action_name_;
   std::replace(client_node_name.begin(), client_node_name.end(), '/', '_');
   // Use suffix '_rclcpp_node' to keep parameter file consistency #1773
-  auto options = rclcpp::NodeOptions().arguments(
+  auto options = rclcpp::NodeOptions().allow_undeclared_parameters(true
+    ).automatically_declare_parameters_from_overrides(true).arguments(
     {"--ros-args",
       "-r",
       std::string("__node:=") +


SteveMacenski:
> While this might be specific to my BT nodes, I don't see the harm allowing undeclared parameters.

This breaks the required static type setting of parameters (e.g. we declare that parameter `ABC` has value of type `double`, not `int`, `string`, or `vector<int>`). This is something ROS 1 did not directly enforce, but we do in ROS 2 for type safety and I wouldn't support removing it as a best practice. There are many reasons not to do this and its been a discussion point over the years to actually remove that option to even make it possible to do this (so its probably not a good solution to rest on, it might go away)

But..

> I'm using ROS parameters in my custom BT nodes and in some cases it makes sense to allow for undeclared parameters. For example I have a BT node that can have a variable number of parameters. 

Variable number of parameters is totally OK, see how the `obstacle layer` handles multiple sources and thusly many different parameters and volumes of parameters. Also check Robot Localization, which has to do a similar thing. Just because you have to declare the type, doesn't mean you don't have all the tools in your toolbox to programatically declare them in loops or check for the first declared parameter which has no value set (e.g. if you're doing an `odom0` `odom1` odomN` thing, you know when to stop looking)

But overall, unless there are specific and compelling reasons to allow for this, I highly suggest you don't. I would really like to not merge this into Nav2 since its a described anti-pattern of ROS 2 to do this

tonynajjar:
Thanks for the detailled explanation and for pointing me to alternatives. I'm fine for closing this PR then!

PR #3439, opened 2023-03-02
Humble backport of MPPI controller

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | N/a |

---

* Backported MPPI controller package and made necessary changes to comply with Humble

---

#### For Maintainers: 


SteveMacenski:
There appear to me to be a few non-backport-able changes here which need some discussion about how to handle:
- We cannot add the `reset()` method to the controller interface, that would break ABI policy. I'm not sure what the solution here is. We could remove it, but then new requests to the controller will be "hot started" with the previous state. I suppose we could have a timer that checks if > N seconds has passed to do a reset each time compute trajectories are called
- The controller exceptions can't be used by the controller server since that also breaks ABI policy, but since Humble doesn't have the contextual exception handling in the interfaces, it doesn't really matter. That can be removed and all of the exceptions can just be `std::runtime_error`s. 
- The 2 inflation layer API changes also break ABI. It would be annoying, but I suppose we can make those parameters for the Humble version instead of grabbing them directly

tonynajjar:
I think it's finally time for me to learn what ABI actually is (I know API). Do you have a quick explanation? How do you recognize when something breaks ABI?

SteveMacenski:
Ugh, please don't force push. If at all possible, please revert that. Otherwise, I will have to line-by-line recheck 10,000 lines of code because I can't see exactly what you changed relative to what I already line-by-line checked before release on Rolling. That will take me hours. 

Also CI is failing to build, a service API changed from Humble to Rolling which would need to be changed to build in Humble-land

tonynajjar:
I can't undo the force push I did earlier so I started a new history to make it clearer (had to force push once more for that :see_no_evil:). The [first commit](https://github.com/ros-planning/navigation2/pull/3439/commits/d701c1543160a5b784c7859d1c1197a821d87412) is the cherrypick of the PR from main **as-is**. Afterwards are my changes.

I'm still working on the reset but you can take a look at the rest. I hope it's clearer now.

> Have you played around with it yet? Thoughts?

nope but it's planned!

nav2_mppi_controller/src/controller.cpp
```diff
@@ -0,0 +1,132 @@
...
+  nav2_core::GoalChecker * goal_checker)
+{
+#ifdef BENCHMARK_TESTING
+  auto start = std::chrono::system_clock::now();
+#endif
+
+  if ((clock_->now() - last_time_called_).seconds() > reset_period_) {
+    reset();
+  }
+  last_time_called_ = clock_->now();
```
tonynajjar, line 91: You meant something like this or would you rather the check to be in an external function called by a ROS timer? I think you meant with the timer but I figured that's potentially simpler, provided that it's acceptable to reset only once `computeVelocityCommands` is called again, as opposed to during inactivity.

SteveMacenski:
Also this should be backported alongside https://github.com/ros-planning/navigation2/pull/3444 It makes things more smooth

SteveMacenski:
> [remove changes unrelated to mppi](https://github.com/ros-planning/navigation2/pull/3439/commits/55fec35fdb82356e7952c9c9c3ee7fd7195422f4)

This commit needs to be reverted those are critical for MMPI

SteveMacenski:
LMK when you want me to take a look at the timer stuff - that if you could squash into a single commit would be nice so I can see the total change diffs. Everything else I looked at and looks fine to me

nav2_mppi_controller/src/controller.cpp
SteveMacenski: Nope, this is exactly what I meant

tonynajjar:
> Also this should be backported alongside https://github.com/ros-planning/navigation2/pull/3444 It makes things more smooth

added

> LMK when you want me to take a look at the timer stuff - that if you could squash into a single commit would be nice so I can see the total change diffs. Everything else I looked at and looks fine to me

I think the whole PR is ready. Can't really squash anymore, it's only two commits related to the timer (+1 for the readme)

tonynajjar:
I don't know which editor you use but in vscode, with the git plugins you can easily compare the whole package across branches

image

tonynajjar:
BTW I did test the timer and the new parameters retrieval manually

tonynajjar:
https://github.com/ros-planning/navigation2/pull/3439/commits/8c54aba89fd3926827cecc735f51ec4fece25962

added this commit that fixes the tests (because https://github.com/ros-planning/navigation2/pull/3131 is not in Humble)

nav2_mppi_controller/README.md
```diff
@@ -0,0 +1,270 @@
...
+ | gamma                      | double | Default: 0.015. A trade-off between smoothness (high) and low energy (low). This is a complex parameter that likely won't need to be changed from the default of `0.1` which works well for a broad range of cases. See Section 3D-2 in "Information Theoretic Model Predictive Control: Theory and Applications to Autonomous Driving" for detailed information.       |
+ | visualize                  | bool   | Default: false. Publish visualization of trajectories, which can slow down the controller significantly. Use only for debugging.                                                                                                                                       |
+ | retry_attempt_limit        | int    | Default 1. Number of attempts to find feasible trajectory on failure for soft-resets before reporting failure.                                                                                                                                                                                                       |
+| reset_period            | double    | Default 1.0. required time of inactivity to reset optimizer  (only in Humble)            |
```
SteveMacenski, line 55:
```suggestion
| reset_period            | double    | Default 1.0. required time of inactivity to reset optimizer  (only in Humble due to backport ABI policies)            |
```

nav2_mppi_controller/src/controller.cpp
```diff
@@ -0,0 +1,133 @@
...
+geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
+  const geometry_msgs::msg::PoseStamped & robot_pose,
+  const geometry_msgs::msg::Twist & robot_speed,
+  nav2_core::GoalChecker * goal_checker)
+{
+#ifdef BENCHMARK_TESTING
+  auto start = std::chrono::system_clock::now();
+#endif
+
+  if ((clock_->now() - last_time_called_).seconds() > reset_period_) {
```
SteveMacenski, line 89: If you convert the reset period `double` into a rclcpp::Duration using the `from_seconds` static method, then you can just do 

`if (now() - last > reset_period_)` without converting to seconds. This in only something I'm concerned about because the behavior of time seconds() and nanoseconds() have changed a few times over the years and I'd be concerned if this was backported even further in time it would break alot of mechanics. I think humble does the right thing and returns a floating point of `seconds()` but it used to just give a whole number rounded, which most the time would render to `0.0`.

SteveMacenski:
Otherwise LGTM

SteveMacenski:
Build still failing for the aforementioned service API reasons :smile: It never even gets to MPPI to make sure _it_ compiles. It looks like 1-2 1 line changes.

nav2_mppi_controller/src/controller.cpp
tonynajjar: okay, i'll change. FYI I copied the approach from [here](https://github.com/ros-planning/navigation2/blob/9b91662eacba3cc73719cdebc7ee1d409f5a3a47/nav2_controller/src/controller_server.cpp#L484).

SteveMacenski: Yeah, that's not great either :frowning:

SteveMacenski (approved):

SteveMacenski:
Thanks for the rapid iteration, glad to have this in so quick

PR #3443, opened 2023-03-02
[MPPI] Fix transformed path oscillations

doisyg:
---

## Basic Info

| ------ | ----------- |

---

This PR fixes one of the issue with looping paths in MPPI. See example of oscillation below:

https://user-images.githubusercontent.com/15727892/222525415-181b87a8-7732-4eb4-9605-eb8c04e7622e.mp4

(global path in green, transformed path in purple arrows)

- For searching the upper bound , we search up to the pruning distance along the path, instead of searching up the euclidean distance.
- For the the lower bound, the closest pose from the robot on the path is still used. But the closest pose search is modified: we look up to the min between `max_robot_pose_search_dist` and `prune_distance`

To discuss: is `max_robot_pose_search_dist` still needed ? Also, its definition is a bit misleading:

> Distance ahead of nearest point on path to robot to prune path to. 

I believe it is actually the max integrated distance/path distance to look for the closest point to the robot, from the first pose of the pruned global path

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/path_handler.cpp
+++ b/nav2_mppi_controller/src/path_handler.cpp
@@ -44,11 +44,12 @@ PathHandler::getGlobalPlanConsideringBoundsInCostmapFrame(
   using nav2_util::geometry_utils::euclidean_distance;
 
   auto begin = global_plan_.poses.begin();
-  auto end = global_plan_.poses.end();
 
+  // Don't search further away than the prune distance
   auto closest_pose_upper_bound =
     nav2_util::geometry_utils::first_after_integrated_distance(
-    global_plan_.poses.begin(), global_plan_.poses.end(), max_robot_pose_search_dist_);
+    global_plan_.poses.begin(), global_plan_.poses.end(),
+        std::min(max_robot_pose_search_dist_, prune_distance_));
 
   // Find closest point to the robot
   auto closest_point = nav2_util::geometry_utils::min_by(
@@ -61,17 +62,17 @@ PathHandler::getGlobalPlanConsideringBoundsInCostmapFrame(
   transformed_plan.header.frame_id = costmap_->getGlobalFrameID();
   transformed_plan.header.stamp = global_pose.header.stamp;
 
+  auto pose_above_prune_distance =
+    nav2_util::geometry_utils::first_after_integrated_distance(
+    closest_point, global_plan_.poses.end(), prune_distance_);
+
   unsigned int mx, my;
   // Find the furthest relevent pose on the path to consider within costmap
   // bounds
   // Transforming it to the costmap frame in the same loop
-  for (auto global_plan_pose = closest_point; global_plan_pose != end; ++global_plan_pose) {
-    // Distance relative to robot pose check
-    auto distance = euclidean_distance(global_pose, *global_plan_pose);
-    if (distance >= prune_distance_) {
-      return {transformed_plan, closest_point};
-    }
-
+  for (auto global_plan_pose = closest_point; global_plan_pose != pose_above_prune_distance;
+    ++global_plan_pose)
+  {
     // Transform from global plan frame to costmap frame
     geometry_msgs::msg::PoseStamped costmap_plan_pose;
     global_plan_pose->header.stamp = global_pose.header.stamp;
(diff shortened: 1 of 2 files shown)


nav2_mppi_controller/src/path_handler.cpp
```diff
@@ -61,17 +62,17 @@ PathHandler::getGlobalPlanConsideringBoundsInCostmapFrame(
   transformed_plan.header.frame_id = costmap_->getGlobalFrameID();
   transformed_plan.header.stamp = global_pose.header.stamp;
 
+  auto pose_above_prune_distance =
```
SteveMacenski, line 65: Change to `pruned_plan_end`

nav2_mppi_controller/src/path_handler.cpp
```diff
@@ -44,11 +44,12 @@ PathHandler::getGlobalPlanConsideringBoundsInCostmapFrame(
...
 
   auto begin = global_plan_.poses.begin();
-  auto end = global_plan_.poses.end();
 
+  // Don't search further away than the prune distance
   auto closest_pose_upper_bound =
     nav2_util::geometry_utils::first_after_integrated_distance(
-    global_plan_.poses.begin(), global_plan_.poses.end(), max_robot_pose_search_dist_);
+    global_plan_.poses.begin(), global_plan_.poses.end(),
+        std::min(max_robot_pose_search_dist_, prune_distance_));
```
SteveMacenski, line 52: I don't think prune distance has a role here in deciding where to look for the closest point on the path. It is not related to the pruning task, it is trying to find the closest point on the path if there is or is not regular replanning.

> >  Distance ahead of nearest point on path to robot to prune path to.

> I believe it is actually the max integrated distance/path distance to look for the closest point to the robot, from the first pose of the pruned global path

I agree this is a wrong description, where is that from? I write most of the docs quickly in one sitting so I probably copy pasted wrong.

SteveMacenski (changes requested):
There are some legit test failures to be addressed from this PR

nav2_mppi_controller/src/path_handler.cpp
doisyg: > I don't think prune distance has a role here in deciding where to look for the closest point on the path. It is not related to the pruning task, it is trying to find the closest point on the path if there is or is not regular replanning. If someone hasn't replanned recently, I'm not sure by the prune distance would impact the search distance.

We need to limit our search for the closest point up to the prune_distance. Otherwise, because the closest point is used as the lower bound in the search for the upper bound of the plan to transform, we could have the closest point further along the path that the `prunned_plan_end` and hence the lower bound higher than the upper bound.

```
  // Find the furthest relevent pose on the path to consider within costmap
  // bounds
  // Transforming it to the costmap frame in the same loop
  for (auto global_plan_pose = closest_point; global_plan_pose != pruned_plan_end;
    ++global_plan_pose)
  {
```


> I agree this is a wrong description, where is that from? I write most of the docs quickly in one sitting so I probably copy pasted wrong.
Description has been updated compared to what I was looking at, but still doesn't reflect that it is the distance from the first point of the global plan (which is pruned at each iteration)

https://github.com/ros-planning/navigation2/blob/29ad43b96097383e90b7b5c24594476b6990b09b/nav2_mppi_controller/README.md?plain=1#L64

doisyg:
> There are some legit test failures to be addressed from this PR

Fixed

nav2_mppi_controller/src/path_handler.cpp
SteveMacenski: > We need to limit our search for the closest point up to the prune_distance. Otherwise, because the closest point is used as the lower bound in the search for the upper bound of the plan to transform, we could have the closest point further along the path that the prunned_plan_end and hence the lower bound higher than the upper bound.

I don't think that's accurate? If we have a path that we're just starting (so path marker 0 is the closest point) then we are searching forward from there by the `prune_distance` to find the last point of interest for us forward.  If we instead had path marker 22 as the closest point, the same logic applies. No matter what path marker is chosen as the closest point, we're integrating **from that pose** by the `prune_distance`. If we selected a path pose further from marker 0 than the `prune_distance`, that's unimportant (since we then prune forward **from** that pose by the distance). If you wanted to limit how far we look ahead from the last point, that's what `max_robot_pose_search_dist` is for.

From this line

```
  auto pruned_plan_end =
    nav2_util::geometry_utils::first_after_integrated_distance(
    closest_point, global_plan_.poses.end(), prune_distance_);
 ```

I don't see how its possible that `closest_point` could ever be further along the path than `pruned_plan_end`. By definition here, its at least `closest_point` or further.

doisyg: You are totally right, I got confused

SteveMacenski (approved):

PR #3450, opened 2023-03-07
Remove back quote

alek5k:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Real hardware |

---

* There seems to be a back quote character in the default pose navigation file. I'm not sure how it's considered valid XML, but in the case that XML parsers are switched, this should be a little more resilient.

---

#### For Maintainers: 


SteveMacenski (approved):
Huh, thanks!

Issue #3351, opened 2023-01-06
Potential MPPI Controller Improvements


SteveMacenski:
General improvements
- [ ] Add in some hand-crafted trajectories into the mix via https://arxiv.org/pdf/2401.09241 (i.e. drive to goal, stop, slow, fast, even other algorithms perhaps, etc with samples)
- [ ] Allowing the std of the velocities to be weighted by the speed of the robot (and/or speed of the control sample) or by the prediction horizon (less noise further out; Ornstein-Uhlenbeck processes)

Improved smoothness options
- [ ] Test using a Log-Normal sampling distribution as established in Log-MPPI paper. Handles less noise for similar exploration of space, reducing sampling noise needing to process.
- [ ] Add in the derivative SMPPI term (mutually exclusive with 1% sampling). Attempts to smooth using differential changes in control 
- [ ] Use the described Tsallis VI variation described below to weight only the best samples in averaging
- [ ] Mostly fine, just needs minorly better path post processing - increase the SG window size or order, recursively call it, or consider other options
- [ ] MPPI: Low-frequency distribution sampling to create smoother trajectories https://github.com/ros-navigation/navigation2/issues/5973
- [ ] Adding a low-pass filter on each iteration update
- [ ] Add smoothness penalty function the penalizes large changes between iterations 

Improvements in speed potentially
- [ ] GPU for trajectory rollouts or critic analysis
- [ ] Precomputation and otherwise optimization in a specialized collision checker object

Improved cost critic
- [ ] Using the full cost of the footprint not just the center cost did not work out well due to the plateau of cost from inflated-inscribed 253 space. However, maybe we can integrate the cost across the footprint's border rather than using the single highest cost so that footprints that are "more" near obstacles are penalized more than those "sort of" near obstacles.  This would involve using the Cost Critic to add the costs up. We may want to normalize this by the number of points or something & need to handle such cases when we're far from obstacles so we only need to use the center-point cost. This may address https://github.com/ros-navigation/navigation2/issues/5668
- [ ] Using a full distance field rather than inflation layer so we know the instance to obstacle for a continuous potential field rather than having a plateau. Side effect of being able to score by any function you like (exponential, linear, etc) as well as set minimum distance constraints as well
- [ ]  Remove the Obstacle Critic as deprecated 

New critics
[...]

PR #3454, opened 2023-03-07
Fix Path Angle Critic for bi-directionality

doisyg:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/3351 |
| Robotic platform tested on | Dexory simulation |

---

Draft PR for addressing the asymmetric behavior of the Path Angle Critic between forward and backward motion. 
Currently, as the angle calculation for scoring is done between the trajectory poses direction to the goal (on path) and the trajectories yaws, there is an asymmetric scoring between the forward and backward movement as backward yaws are inverted relative to the direction of motion:
https://github.com/ros-planning/navigation2/blob/34f18d9222e33a2fddb2e45653cd5c0ef7b3bb11/nav2_mppi_controller/src/critics/path_angle_critic.cpp#L65-L69

Also, the filter for discarding scoring if the difference between the robot pose and the robot direction to the goal (on path) is below `max_angle_to_furthest`, is using the yaw of the robot pose, hence never acting if the robot is moving backward:
https://github.com/ros-planning/navigation2/blob/34f18d9222e33a2fddb2e45653cd5c0ef7b3bb11/nav2_mppi_controller/src/critics/path_angle_critic.cpp#L61-L63

This draft PR makes the Path Angle Critics symmetric  by using the minimum angle of the +X and -X orientation.

(on a system with over-weighted `PathAngleCritic`)

Without this PR:

https://user-images.githubusercontent.com/15727892/223572581-467edbea-0907-4546-b63c-a895c361a453.mp4

With this PR:

https://user-images.githubusercontent.com/15727892/223573089-cd24e9a5-0025-45c3-aec5-54ea0d7c5075.mp4

Maybe we should also discuss if the symmetric bi-directional behavior of the Path Angle Critics should be optional.

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/tools/utils.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/tools/utils.hpp
@@ -418,13 +418,20 @@ inline void setPathCostsIfNotSet(
  * @param point_y Point to find angle relative to Y axis
  * @return Angle between two points
  */
-inline double posePointAngle(const geometry_msgs::msg::Pose & pose, double point_x, double point_y)
+inline double posePointAngle(
+  const geometry_msgs::msg::Pose & pose, double point_x, double point_y,
+  double point_yaw)
 {
   double pose_x = pose.position.x;
   double pose_y = pose.position.y;
   double pose_yaw = tf2::getYaw(pose.orientation);
 
   double yaw = atan2(point_y - pose_y, point_x - pose_x);
+
+  if (abs(angles::shortest_angular_distance(yaw, point_yaw)) > M_PI_2) {
+    yaw = angles::normalize_angle(yaw + M_PI);
+  }
+
   return abs(angles::shortest_angular_distance(yaw, pose_yaw));
 }
 
--- a/nav2_mppi_controller/src/critics/path_angle_critic.cpp
+++ b/nav2_mppi_controller/src/critics/path_angle_critic.cpp
@@ -58,17 +58,26 @@ void PathAngleCritic::score(CriticData & data)
   const float goal_x = xt::view(data.path.x, offseted_idx);
   const float goal_y = xt::view(data.path.y, offseted_idx);
 
-  if (utils::posePointAngle(data.state.pose.pose, goal_x, goal_y) < max_angle_to_furthest_) {
+  double forward_angle = utils::posePointAngle(data.state.pose.pose, goal_x, goal_y);
+  double backward_angle = angles::normalize_angle(forward_angle + M_PI);
+
+  if (forward_angle < max_angle_to_furthest_ ||
+    backward_angle < max_angle_to_furthest_)
+  {
     return;
   }
 
   const auto yaws_between_points = xt::atan2(
     goal_y - data.trajectories.y,
     goal_x - data.trajectories.x);
+  const auto yaws_plus_pi_between_points = utils::normalize_angles(yaws_between_points + M_PI);
   const auto yaws =
     xt::abs(utils::shortest_angular_distance(data.trajectories.yaws, yaws_between_points));
+  const auto yaws_plus_pi =
+    xt::abs(utils::shortest_angular_distance(data.trajectories.yaws, yaws_plus_pi_between_points));
+  const auto min_yaws = xt::minimum(yaws, yaws_plus_pi);
 
-  data.costs += xt::pow(xt::mean(yaws, {1}, immediate) * weight_, power_);
+  data.costs += xt::pow(xt::mean(min_yaws, {1}, immediate) * weight_, power_);
 }
 
 }  // namespace mppi::critics
--- a/nav2_mppi_controller/src/critics/path_angle_critic.cpp
+++ b/nav2_mppi_controller/src/critics/path_angle_critic.cpp
@@ -59,7 +59,7 @@ void PathAngleCritic::score(CriticData & data)
   const float goal_y = xt::view(data.path.y, offseted_idx);
 
   double forward_angle = utils::posePointAngle(data.state.pose.pose, goal_x, goal_y);
-  double backward_angle = angles::normalize_angle(forward_angle + M_PI);
+  double backward_angle = abs(angles::normalize_angle(forward_angle + M_PI));
 
   if (forward_angle < max_angle_to_furthest_ ||
     backward_angle < max_angle_to_furthest_)
--- a/nav2_mppi_controller/src/critics/path_angle_critic.cpp
+++ b/nav2_mppi_controller/src/critics/path_angle_critic.cpp
@@ -57,12 +57,11 @@ void PathAngleCritic::score(CriticData & data)
 
   const float goal_x = xt::view(data.path.x, offseted_idx);
   const float goal_y = xt::view(data.path.y, offseted_idx);
+  const float goal_yaw = xt::view(data.path.yaws, offseted_idx);
 
-  double forward_angle = utils::posePointAngle(data.state.pose.pose, goal_x, goal_y);
-  double backward_angle = abs(angles::normalize_angle(forward_angle + M_PI));
-
-  if (forward_angle < max_angle_to_furthest_ ||
-    backward_angle < max_angle_to_furthest_)
+  if (utils::posePointAngle(
+      data.state.pose.pose, goal_x, goal_y,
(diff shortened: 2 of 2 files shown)


nav2_mppi_controller/src/critics/path_angle_critic.cpp
```diff
@@ -58,17 +58,26 @@ void PathAngleCritic::score(CriticData & data)
...
 
   const auto yaws_between_points = xt::atan2(
     goal_y - data.trajectories.y,
     goal_x - data.trajectories.x);
+  const auto yaws_plus_pi_between_points = utils::normalize_angles(yaws_between_points + M_PI);
   const auto yaws =
     xt::abs(utils::shortest_angular_distance(data.trajectories.yaws, yaws_between_points));
+  const auto yaws_plus_pi =
+    xt::abs(utils::shortest_angular_distance(data.trajectories.yaws, yaws_plus_pi_between_points));
+  const auto min_yaws = xt::minimum(yaws, yaws_plus_pi);
```
SteveMacenski, line 78: This provides point-by-point minimums between the 2 sets, not deciding which set to use. The trajectories could be all over the place, some backward, some forward, some jaggedy. We need to decide which to use analytically and not guess

nav2_mppi_controller/src/critics/path_angle_critic.cpp
```diff
@@ -58,17 +58,26 @@ void PathAngleCritic::score(CriticData & data)
   const float goal_x = xt::view(data.path.x, offseted_idx);
   const float goal_y = xt::view(data.path.y, offseted_idx);
 
-  if (utils::posePointAngle(data.state.pose.pose, goal_x, goal_y) < max_angle_to_furthest_) {
+  double forward_angle = utils::posePointAngle(data.state.pose.pose, goal_x, goal_y);
+  double backward_angle = abs(angles::normalize_angle(forward_angle + M_PI));
```
SteveMacenski, line 62: I think we need to make a parameter if multidirectionality is valid. If we have preferences for forward, we should not be skipping based on the path going in the opposite direction of the current orientation. We need to apply it in that case (that's actually the most extreme case of why this critic exists to help push that orientation motion).

nav2_mppi_controller/src/critics/path_angle_critic.cpp
SteveMacenski: I believe this is bad to do, that is. I'm not 100% certain

SteveMacenski (changes requested):

doisyg:
Giving that problem a bit more thoughts, I pushed a new commit with a different method that should not need a dedicated parameter to activate or not the symmetric behavior:
The new method makes use of the goal poses yaw to decide which orientation (direction to goal or direction to goal + pi) to use. Tested in practice, it seems to work great. This however assumes that the global path poses are filled with proper orientations, but that's usually the case for modern planner. WDYT ?

nav2_mppi_controller/src/critics/path_angle_critic.cpp
doisyg: See new commit, by using the plan poses orientation info, we can support all use cases transparently:
- following in the forward direction a plan which has poses with orientations pointing in the direction of path increment
- following in the backward direction a plan which has poses with orientations pointing in the opposite direction of path increment
- doing an inversion manoeuvre  when starting to follow in the backward direction a plan which has poses with orientations pointing in the direction of path increment
- doing an inversion manoeuvre  when starting to follow in the forward direction a plan which has poses with orientations pointing in the opposite direction of path increment

nav2_mppi_controller/src/critics/path_angle_critic.cpp
doisyg: It was kindof working but definitly not ideal. Switching to a better method

nav2_mppi_controller/src/critics/path_angle_critic.cpp
SteveMacenski: Not all planners provide the orientation though - like [NavFn](https://github.com/ros-planning/navigation2/blob/main/nav2_navfn_planner/src/navfn_planner.cpp#L426-L429) or [Theta*](https://github.com/ros-planning/navigation2/blob/main/nav2_theta_star_planner/src/theta_star_planner.cpp#L214-L216) just set the XY coordinates, not the full SE2 poses (because its impossible for them to - they know nothing of kinematics). 

It is a slick solution though if we had that :/ I suppose we could update those 2 planners to use orientations with something like https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/smoother.cpp#L284-L300 (broken out into a `nav2_util`)

SteveMacenski:
See comment in thread

nav2_mppi_controller/include/nav2_mppi_controller/tools/utils.hpp
```diff
@@ -418,13 +418,20 @@ inline void setPathCostsIfNotSet(
...
+  const geometry_msgs::msg::Pose & pose, double point_x, double point_y,
+  double point_yaw)
 {
   double pose_x = pose.position.x;
   double pose_y = pose.position.y;
   double pose_yaw = tf2::getYaw(pose.orientation);
 
   double yaw = atan2(point_y - pose_y, point_x - pose_x);
+
+  if (abs(angles::shortest_angular_distance(yaw, point_yaw)) > M_PI_2) {
```
SteveMacenski, line 431: Not a bad solution, but again I think this requires explicit parameterization if this is allowable. If you have a robot system with a preference **not** to reverse if possible, then we don't want to set the angle this way. Else, we should be using the actual orientation. It also feels weird to me that you're hiding this detail into a generic utility which is just supposed to be finding the angle between a point and a pose. This feels wrong.

nav2_mppi_controller/src/critics/path_angle_critic.cpp
```diff
@@ -57,16 +57,30 @@ void PathAngleCritic::score(CriticData & data)
...
+      goal_yaw) < max_angle_to_furthest_)
+  {
     return;
   }
 
   const auto yaws_between_points = xt::atan2(
     goal_y - data.trajectories.y,
     goal_x - data.trajectories.x);
+
+  const auto yaws_between_points_corrected = xt::eval(
```
SteveMacenski, line 73: I think I made this same comment before, this is doing elementwise selection of which orientation to use by what's the most convenient, that's not a valid application of a critic function I don't believe. I think we have to decide which orientation we're trying to achieve and penalize applicable to that. Else, some trajectories in a set will be penalized by different criteria than others. I have to think about that a little bit though, maybe it is OK to do elementwise but it feels like that needs to be really well thought out, not on a whim. 

If we have a system preferring forward, then only forward should be considered. If we have a system where reversing is viable and we're reversing or wanting to reverse, then it makes sense to use the inverted orientation. If we have a system where reversing is viable and we're not reversing or wanting to reverse, then it makes sense to use the real orientation.

SteveMacenski: I'm thinking it might be OK to do elementwise as long as a trajectory is generally going forward or reverse -- but what happens if we have a backup maneuver where the trajectory legitimately has a portion of the path which is backing up and another section that is moving forward within the time horizon? In that situation, we'd be applying different orientation direction penalties which makes my head hurt in trying to figure out if that's a good idea or not

nav2_mppi_controller/include/nav2_mppi_controller/tools/utils.hpp
SteveMacenski: ... Or maybe not? As long as moving backward is possible I suppose the other critic should take care of that, but it needs to be tested to make sure this doesn't incentivize more backward motion. 

The main concern I see is if a user literally has `min_vel_x` set to `0.0` so it cannot at at all move backward, this would incentivize reversing motions that are not possible at all rather than fully using the force of the purpose of this plugin to make the robot do a full 180 turn to get to its desired forward-facing heading.

So I think a parameter is needed for the situation, at minimum, where reversing is 100% unequivocally impossible by the users' settings.  The question I then have is if we should only apply it in that situation or if we should also apply it when we would _prefer_ not to move backwards. I'm just not entirely sure from a quick glance - I'd need to think about it and probably test.

SteveMacenski:
PR to supersede shortly

PR #3457, opened 2023-03-08
Support for Dev Containers and Codespaces

ruffsl:
Work in progress for adding support for Dev Container and Codespaces:

- https://code.visualstudio.com/docs/devcontainers/containers
- https://containers.dev
- https://docs.github.com/en/codespaces

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,27 @@
+{
+    "name": "Nav2",
+    "image": "ghcr.io/ros-planning/navigation2:main-dev",
+    "runArgs": [
+        "--privileged",
+        "--network=host"
+    ],
+    "workspaceMount": "source=${localWorkspaceFolder},target=/opt/overlay_ws/src/navigation2,type=bind",
+    "workspaceFolder": "/opt/overlay_ws/src/navigation2",
+    // "features": {
+    //     "ghcr.io/devcontainers/features/desktop-lite:1": {},
+    //     "ghcr.io/devcontainers/features/github-cli:1": {}
+    // },
+    "customizations": {
+        "codespaces": {
+            // "openFiles": [
+            //     "doc/development/codespaces.md"
+            // ]
+        },
+        "vscode": {
+            "settings": {},
+            "extensions": [
+                "streetsidesoftware.code-spell-checker"
+            ]
+        }
+    }
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
         "--network=host"
     ],
     "workspaceMount": "source=${localWorkspaceFolder},target=/opt/overlay_ws/src/navigation2,type=bind",
-    "workspaceFolder": "/opt/overlay_ws/src/navigation2",
+    "workspaceFolder": "/opt/overlay_ws/src",
     // "features": {
     //     "ghcr.io/devcontainers/features/desktop-lite:1": {},
     //     "ghcr.io/devcontainers/features/github-cli:1": {}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -20,7 +20,12 @@
         "vscode": {
             "settings": {},
             "extensions": [
-                "streetsidesoftware.code-spell-checker"
+                "eamodio.gitlens",
+                "esbenp.prettier-vscode",
+                "GitHub.copilot",
+                "ms-iot.vscode-ros",
+                "streetsidesoftware.code-spell-checker",
+                "twxs.cmake"
             ]
         }
     }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -13,9 +13,9 @@
     },
     "customizations": {
         "codespaces": {
-            // "openFiles": [
-            //     "doc/development/codespaces.md"
-            // ]
+            "openFiles": [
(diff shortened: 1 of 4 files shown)


SteveMacenski:
Not necessarily related to this, but it might be so just mentioning it here. From the Slack thread: https://navigation2.slack.com/archives/C012L20NQLV/p1677595926445949 the containers don't seem to have the built packages in them to use as a basis (only a clone). Would it be useful to have the installed builds from CI stored in the containers for this application as well?    

I suggested Tony that perhaps we should have a different container package for Nav2 "end user" use that is built nightly which clones + builds Nav2 in the container construction so that it can be used for end-user development / testing (separately from our CI container). I have it on my short list over the next few weeks (to months) in things to toy around with to learn the GitHub Actions / Package release pipelines, but it sounds like it might be more directly necessary for codespaces

ruffsl:
> the containers don't seem to have the built packages in them to use as a basis (only a clone).

Indeed, the current images for our CI does not include a pre-compiled overlay workspace, but merely a partial copy of the source code including only the `package.xml` files needed to pre-install overlay dependencies. This is because the CI image is built only to target the `builder` stage in the project's Dockerfile:

https://github.com/ros-planning/navigation2/blob/afd96bebbf7a66a88cd37ed879dcd596f540d80a/.github/workflows/update_ci_image.yaml#L104

https://github.com/ros-planning/navigation2/blob/afd96bebbf7a66a88cd37ed879dcd596f540d80a/Dockerfile#L38

This helps keep the CI image small to minimize the amount of time it take for our CI runners to pull the image, a frequent enough task at the state of any CI job. This also prevents the CI image from being invalidated too frequently, given an image is the foundation to the rest of our caching gestates, as serves as the initial seed to our lock files used to derives cache keys:

https://github.com/ros-planning/navigation2/blob/afd96bebbf7a66a88cd37ed879dcd596f540d80a/Dockerfile#L122-L124

https://github.com/ros-planning/navigation2/blob/afd96bebbf7a66a88cd37ed879dcd596f540d80a/.circleci/config.yml#L321-L323

---

> Would it be useful to have the installed builds from CI stored in the containers for this application as well?

Copying and building all source files into the image would inherently invalidate the docker image build cache when building the CI image tagged as `ghcr.io/ros-planning/navigation2:main`, and thus subsequently all downstream caches that derive from the image. However, is something we could do instead for a new child image tagged as `ghcr.io/ros-planning/navigation2:main-dev`. This image tag could be built to target the `tester` stage in the project's Dockerfile, and so would include pre-comiled binaries for the overlay workspace, as well as a warm ccace folder and initial run of test results:

https://github.com/ros-planning/navigation2/blob/8e5a3a33d786219bf5784b491ec5a548045b0fba/.github/workflows/update_ci_image.yaml#L122-L128

Downstream users such as those developing using Dev Containers and Codespaces could then checkout their own source branches, then use `colcon cache`, just as our CI already does, to re-build and re-test only the effected packages with invalidated caches due to branch modifications. This could save individuals a good deal of time when debugging or reviewing multiple branches and pull requests.

---

I had hoped on reusing the native caching infrastructure in GitHub Actions, such that users running Codespaces could restore from the same caches stored by job runners from GitHub Actions, thus consolidating the source of truth for current cached artifacts and its reuse for everywhere else. However, despite my objections, it seems that GitHub does not intend on supporting such use cases for the Actions cache API any time soon:

- https://github.com/actions/cache/issues/105#issuecomment-1008351664
- https://github.com/community/community/discussions/44567#discussioncomment-4758120

The approach I think I'll opt for here instead is to just bake in the workspace artifacts into a `dev` image tag, and let that serve as the cache instead. Sort of like this old approach using multistage builds to cache artifacts in separate image layers:

- https://github.com/codespaces-contrib/dependency-caching

Even though this means such a `dev` image tag may encounter a lot of churn, as well as slow to rebuild given docker would be building and testing the workspace from scratch upon each new commit into the `main` branch, I don't think that to be as much of an issue, given users would be rebuilding devcontainers or starting new codespaces with a lower cadence than CI jobs anyways.

---

> I suggested Tony that perhaps we should have a different container package for Nav2 "end user" use that is built nightly which clones + builds Nav2 in the container construction so that it can be used for end-user development / testing (separately from our CI container). I have it on my short list over the next few weeks (to months) in things to toy around with to learn the GitHub Actions / Package release pipelines, but it sounds like it might be more directly necessary for codespaces

I'd be hesitate to recommend the use of such `dev` image tag, as described above, for production purposes or broader alpha/beta distribution. For example, I had to deliberately disable our `FAIL_ON_TEST_FAILURE` build arg so that failing tests wouldn't also then block rebuilds of such `dev` image tag. I may need to do something similar to `RUN colcon build` for the overlay as well, although the `main` branch has be rather well behaved in terms of seldom breaking the build. That would of course results in shipping images with broken overlay workspace, and so not quite as useful.

I suppose ROS package maintainers would prefer we iterate instead with the official apt `testing` repo for `rolling` instead, as shipping Debian binaries through such channels would be a much leaner distribution medium than monolithic docker images. Of course, we could also look into that recently announced GitHub action for publishing standalone debians, but resorting to out-of-band channels could lead to issues with support fragmentation and issue reporting.

- https://discourse.ros.org/t/packaging-ros-with-github-actions-software-release-and-blog-post/30047

Perhaps @nuclearsandwich may have more to say on those matters.

ruffsl:
Merging so I can proceed with generating and do live testing with the `dev` tag using codespaces.

PR #3470, opened 2023-03-11
Use Codespaces prebuilds

ruffsl:
instead of baking a cache of the workspace in the docker image:

https://docs.github.com/en/codespaces/prebuilding-your-codespaces

Codespaces prebuilds also have some snazzy features to gracefully deal with branching and frequent commits:

https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds#about-pushing-changes-to-prebuild-enabled-branches


ruffsl:
@SteveMacenski , I was about to try and configure prebuilds to test this PR, as documented here:

https://docs.github.com/en/codespaces/prebuilding-your-codespaces/configuring-prebuilds#configuring-prebuilds

However, does the ros-planning org still have access to Codespaces and Prebuilds?
I'm seeing these notices on this repo's setting for Prebuild configuration:

image
> https://github.com/ros-planning/navigation2/settings/codespaces

I poked around the other repos on this org, and didn't see any other using Codespaces at all, so I'm not sure how we'd have reached a spending limit yet.

We should follow up with our email thread with Craig, and check if we've enabled prebuilds for this orgs OSS projects.

tonynajjar:
This is an awesome start! :partying_face:

PR #3471, opened 2023-03-13
Fix for different ceres isinf() API

AlexeyMerzlyakov:
Fix for different `ceres::IsInfinite()`/`ceres::isinf()` API on different Ceres Solver versions (`2.0.0-`/`2.1.0+` respectively), causing build regression on Ubuntu 20.04 where Ceres is being distributed under older version.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3158 |
| Robotic platform tested on | `colcon test --packages-select nav2_constrained_smoother`, TB3 simulation |

---

* The regression appeared since on Ubuntu 20.04 since #3158 commit. This commit is adjusting for newer `ceres::isinf()` API used in Ubuntu 22.04, but Ubuntu 20.04 begins out of support
* Since Ubuntu 20.04 is in the [officially supported](https://docs.ros.org/en/rolling/Installation/Alternatives/Ubuntu-Development-Setup.html#system-requirements) by ROS2 Rolling list, it should be fixed
* In the [docs](http://ceres-solver.org/version_history.html#backward-incompatible-api-changes,), Ceres Solver proposes to switch on `std::isinf()` API, that was made in this change:
> Classification functions like `IsFinite` are deprecated. Use the `C++11` functions (`isfinite`, `isnan` etc) going forward. However to maintain consistent behaviour with comparison operators, these functions only inspect the scalar part of the `Jet`.

* The fix is made for both instantiations of templated typename `T`:
  * `std::isinf(d)`: for `T = Scalar (a.k.a. double)` and
  * for `T = ceres::Jet(D, N)` as comparison for scalar part of Ceres jets
* Using of `std::isinf()` API will allow to be independent of Ceres API version and thus support both Ubuntu 22.04 and Ubuntu 20.04 builds

No documentation changes required

---

* No future works are expected

#### For Maintainers: 

### DIFF
--- a/nav2_constrained_smoother/include/nav2_constrained_smoother/smoother_cost_function.hpp
+++ b/nav2_constrained_smoother/include/nav2_constrained_smoother/smoother_cost_function.hpp
@@ -159,7 +159,7 @@ protected:
     Eigen::Matrix<T, 2, 1> center = arcCenter(
       pt_prev, pt, pt_next,
       next_to_last_length_ratio_ < 0);
-    if (ceres::isinf(center[0])) {
+    if (CERES_ISINF(center[0])) {
       return;
     }
     T turning_rad = (pt - center).norm();
--- a/nav2_constrained_smoother/CMakeLists.txt
+++ b/nav2_constrained_smoother/CMakeLists.txt
@@ -17,6 +17,10 @@ find_package(Ceres REQUIRED COMPONENTS SuiteSparse)
 
 set(CMAKE_CXX_STANDARD 17)
 
+if(${CERES_VERSION} VERSION_LESS_EQUAL 2.0.0)
+  add_definitions(-DUSE_OLD_CERES_API)
+endif()
+
 nav2_package()
 
 set(library_name nav2_constrained_smoother)
--- a/nav2_constrained_smoother/include/nav2_constrained_smoother/utils.hpp
+++ b/nav2_constrained_smoother/include/nav2_constrained_smoother/utils.hpp
@@ -20,6 +20,16 @@
 
 #define EPSILON 0.0001
 
+/**
+ * Compatibility with different ceres::isinf() and ceres::IsInfinite() API
+ * used in Ceres Solver 2.1.0+ and 2.0.0- versions respectively
+ */
+#if defined(USE_OLD_CERES_API)
+  #define CERES_ISINF(x) ceres::IsInfinite(x)
+#else
+  #define CERES_ISINF(x) ceres::isinf(x)
+#endif
+
 namespace nav2_constrained_smoother
 {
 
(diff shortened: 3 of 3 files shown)


nav2_constrained_smoother/include/nav2_constrained_smoother/smoother_cost_function.hpp
```diff
@@ -159,7 +159,7 @@ class SmootherCostFunction
     Eigen::Matrix<T, 2, 1> center = arcCenter(
       pt_prev, pt, pt_next,
       next_to_last_length_ratio_ < 0);
-    if (ceres::isinf(center[0])) {
+    if (isInfinite(center[0])) {
```
SteveMacenski, line 162: Why not just `ifdef` on the version number of ceres? 

By the way, Tier 3 means that things might not really work but we generally try to make it work if at all possible, but I don't think we need to worry about it too much as long as 20.04's distribution's API are correct in their respective branches

SteveMacenski (changes requested):

nav2_constrained_smoother/include/nav2_constrained_smoother/smoother_cost_function.hpp
AlexeyMerzlyakov: > Why not just ifdef on the version number of ceres?

I relied mainly on Ceres official docs where they are recommending to use `std::isinf()` functions for such comparisons. Although, the solution with `#ifdef` will also be fine, I think. Something like:
```
#if defined(USE_OLD_CERES_API)
  #define CERES_ISINF(x) ceres::IsInfinite(x)
#else
  #define CERES_ISINF(x) ceres::isinf(x)
#endif
```

SteveMacenski: I don’t mind much one way or another, that’s just how we deal with such problems in PCL so its familiar to me. If you prefer this, that works too. 

Can you confirm for me this builds given our CI issues?

AlexeyMerzlyakov: Okay, I finally decided to use latest solution with Ceres versions defined in compile-time, as more straightforward and elegant (we will re-use existing API, instead of making something again) solution. Force-pushed the commit.

I could confirm that the build works with:
* Tier-III system
  * Ubuntu 20.04
  * ROS2 Rolling built from sources (2023-03-10)
  * Ceres v1.14.0-4ubuntu1.1
  * Nav2 build ... OK
* navigation2:main Docker container
  * Ubuntu 22.04
  * ROS2 Rolling installed from binaries ~2023.01.31
  * Ceres v2.0.0+dfsg1-5
  * Nav2 build ... OK
* navigation2:main Docker container w/ latest Ceres (pre-2.2.0)
  * Ubuntu 22.04
  * ROS2 Rolling installed from binaries ~2023.01.31
  * Ceres was replaced to built one from 2023.03.14 sources
  * Nav2 build ... OK

PR #3482, opened 2023-03-16
Improve Dev Container ergonomics

ruffsl:
- setting up autocompletion for bash and apt
- while also sourcing the overlay workspace so that vscode extensions and task processes have a working ros environment
- reduce need for internet after initial image build
- reduce log verbosity when building Dev Container
- add and setup more extensions


ruffsl:
The added docs for these changes can be reviewed over here:
- https://github.com/ros-planning/navigation.ros.org/pull/407

SteveMacenski (approved):

PR #3488, opened 2023-03-17
Use cSpell and add words

ruffsl:
By adding words to cSpell settings using text corpus from current source:

```
npx cspell-cli "nav2_*/**/*.{py,cpp,h,hpp}" --words-only --unique --no-progress
```

- https://stackoverflow.com/a/73188763/2577586
- https://github.com/streetsidesoftware/cspell-cli

TODO:

- simplify dictionary to lower case?
- rerun cspell-cli to use same default setting as vscode cspell


SteveMacenski:
I suppose now that we have the .vscode directory from the dev containers stuff may as well add this too

ruffsl:
It might be nice to look into how to use this for CI. For compiled languages with errors for unused or uninitialized variables, this isn't as pertinent, but for interpreted languages such as Python, I've still found it useful for catching typos and misspellings.

For example, this is how tests for colcon use `scspell`:
- https://github.com/colcon/colcon-core/blob/99c03262ad7edd4af5954c424c9248e0c4fe7307/test/test_spell_check.py

SteveMacenski:
@ruffsl should something happen with this?

ruffsl:
Ideally, it might be nice to have this check run as a test for CI, that way it always stays up-to-date, similar to how it's used for osrf's colcon python packages.

Also, at that point, it may be better split this into a per package dictionary, to avoid leaking typos across packages. But that could also add some inconvenience when importing commonly named interfaces and types, as the vocab would have to be updated for each package's dictionary, compared to just having one global dictionary for the entire repo.

I don't have immediate cycles to finish this, so we could close for now. But this could also be a "good first issue" for new contributors.

SteveMacenski:
OK, do you want to file a ticket with that first issue tag with a link here + a description? I'll add it to my queue of potential student projects!

PR #3493, opened 2023-03-20
Change directory to inside checked out repo

ruffsl:
or relative path under $GITHUB_WORKSPACE
that actions/checkout places the repository

Fixes: https://github.com/ros-planning/navigation2/pull/3492
Grr... I wish I could easily test these patches to action locally.

### DIFF
--- a/.github/workflows/update_ci_image.yaml
+++ b/.github/workflows/update_ci_image.yaml
@@ -27,6 +27,7 @@ jobs:
         id: check
         if: github.event_name == 'push'
         run: |
+          cd navigation2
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I believe there is, but no worries. As long as you get there eventually, I'm not too worried about it. I do some testing in production for tooling (think: mergify) as well - its difficult to setup locally and its the path of least resistance. If this was algorithmic, that might be different, but dev ops is difficult

ruffsl:
This happened to be a simpler way to iterate on the branch itself, then I can just revert the commit before merging:

- https://github.com/ros-planning/navigation2/pull/3494/commits/d0a25533da3b659e92748f410cec46c1a3b39553

PR #3497, opened 2023-03-22
Ackermann cost update

ObeidHaidar:
Is the cost when using an Ackermann motion model meant to be modified twice?
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | custom hardware |

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/critics/constraint_critic.cpp
+++ b/nav2_mppi_controller/src/critics/constraint_critic.cpp
@@ -66,12 +66,13 @@ void ConstraintCritic::score(CriticData & data)
         std::move(out_of_turning_rad_motion)) *
         data.model_dt, {1}, immediate) * weight_, power_);
   }
-
-  data.costs += xt::pow(
-    xt::sum(
-      (std::move(out_of_max_bounds_motion) +
-      std::move(out_of_min_bounds_motion)) *
-      data.model_dt, {1}, immediate) * weight_, power_);
+  else {
+    data.costs += xt::pow(
+      xt::sum(
+        (std::move(out_of_max_bounds_motion) +
+        std::move(out_of_min_bounds_motion)) *
+        data.model_dt, {1}, immediate) * weight_, power_);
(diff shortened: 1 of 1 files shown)


ObeidHaidar:
Also @SteveMacenski , this probably isn't the right place to ask, but any recommendation for trying to increase the prediction horizon? The current horizon doesn't work as good when controlling robots with a bigger footpath.

nav2_mppi_controller/src/critics/constraint_critic.cpp
```diff
@@ -66,12 +66,13 @@ void ConstraintCritic::score(CriticData & data)
         std::move(out_of_turning_rad_motion)) *
         data.model_dt, {1}, immediate) * weight_, power_);
   }
-
-  data.costs += xt::pow(
-    xt::sum(
-      (std::move(out_of_max_bounds_motion) +
-      std::move(out_of_min_bounds_motion)) *
-      data.model_dt, {1}, immediate) * weight_, power_);
+  else {
```
SteveMacenski, line 69: Linting issue, if nothing else `} else {`

SteveMacenski (changes requested):
Please target the `main` branch as the bot indicates. I'll backport to Humble immediately but as a point of process to keep the git history clean it needs to go there first

SteveMacenski:
> any recommendation for trying to increase the prediction horizon? The current horizon doesn't work as good when controlling robots with a bigger footpath.

Sure, do it! You can increase the number of trajectory timesteps, though it comes obviously with proportionally higher computational costs

nav2_mppi_controller/src/critics/constraint_critic.cpp
ObeidHaidar: i'll fix it and add pull request to main. Just wanted to double check that it's not done on purpose. I'll add a return instead of an else statement and will create a new pull request to main.

Issue #3496, opened 2023-03-21
[Collision monitor] Publish `collision_monitor` state on a ROS topic.


kaichie:
#### Feature description
Publish `collision_monitor` state on a ROS topic.

Currently, the `collision_monitor` package only prints the Action state on the terminal, it will be useful to publish this as a message so that other parts of the system can use this information.

#### Implementation considerations
The proposed feature should add a new publisher that publishes the state as a ROS topic. Depending on the level of details required, it should at least publish:
- action_type (`<int>` `DO_NOTHING=0`, `STOP=1`, `SLOWDOWN=2`, `APPROACH=3`) 

and/or more of the following fields:

- velocity: the requested velocity that the current state is reporting
- polygon_name: the name of the polygon that is currently activated
- points: the points of the polygon that is currently activated

If more information is required beyond the `action_type`, we probably want to create a new nav2 msgs and also update the `Action` struct to store more of these informations.

SteveMacenski:
Sounds reasonable! 

I don't think velocity is as important as the cmd_vel is going to contain that velocity as the normal output. The polygon name might be helpful context, I agree!

tonynajjar:
For the msg published by this new node, I would have liked to reuse [CollisionMonitorState](https://github.com/ros-planning/navigation2/blob/70c6ba13ea49e911e9c0fd3c19b33fb9ae4cb965/nav2_msgs/msg/CollisionMonitorState.msg), but in the context of this new node, we can have more than one polygon triggered + we don't care about action_type. Would you recommend creating another msg?

BriceRenaudeau:
This issue isn't closed?

SteveMacenski:
Thanks!

PR #3504, opened 2023-03-23
Publish collision monitor state

kaichie:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3496  |
| Robotic platform tested on | Dexory Simulation |

---

* Added a new msg in `nav2_msgs` - `CollisionMonitorState.msg`
* Publish CollisionMonitor action state
* Also fixed that the first action will not be printed if it is `DO_NOTHING`
* Also updated the state change logic to check with `polygon_name` instead of `action_type` (To handle multiple `slowdown` polygons correctly)

* Added new topic + param, so need to add that to default configs and documentation page

---

* This PR report the CM's action state change based on the input `cmd_vel`, but I think it will be very useful to publish the polygon state even when the robot is not moving too. Especially for cases like when the `stop` polygon is triggered(regardless of `cmd_vel` speed), or when there are points in `approach` polygon where the robot will not be able to move. (Probably similar requirement as mentioned [here](https://github.com/ros-planning/navigation2/pull/3500#issuecomment-1480890637))

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -15,6 +15,8 @@
 #ifndef NAV2_COLLISION_MONITOR__TYPES_HPP_
 #define NAV2_COLLISION_MONITOR__TYPES_HPP_
 
+#include <string>
+
 namespace nav2_collision_monitor
 {
 
@@ -73,6 +75,7 @@ struct Action
 {
   ActionType action_type;
   Velocity req_vel;
+  std::string polygon_name;
 };
 
 }  // namespace nav2_collision_monitor
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -68,7 +68,6 @@ enum ActionType
   STOP = 1,  // Stop the robot
   SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
   APPROACH = 3,  // Keep constant time interval before collision
-  ACTION_UNKNOWN = 4 // Default for initializing action type
 };
 
 /// @brief Action for robot
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -67,7 +67,7 @@ enum ActionType
   DO_NOTHING = 0,  // No action
   STOP = 1,  // Stop the robot
   SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
-  APPROACH = 3,  // Keep constant time interval before collision
+  APPROACH = 3  // Keep constant time interval before collision
 };
 
 /// @brief Action for robot
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -67,7 +67,8 @@ enum ActionType
   DO_NOTHING = 0,  // No action
   STOP = 1,  // Stop the robot
   SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
-  APPROACH = 3  // Keep constant time interval before collision
+  APPROACH = 3,  // Keep constant time interval before collision
+  ACTION_UNKNOWN = 4 // Default for initializing action type
 };
 
 /// @brief Action for robot
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -64,11 +64,11 @@ struct Pose
 /// @brief Action type for robot
 enum ActionType
 {
-  DO_NOTHING = 0,  // No action
-  STOP = 1,  // Stop the robot
-  SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
-  APPROACH = 3,  // Keep constant time interval before collision
-  ACTION_UNKNOWN = 4 // Default for initializing action type
+  ACTION_UNKNOWN = 0, // Default for initializing action type
+  DO_NOTHING = 1,  // No action
+  STOP = 2,  // Stop the robot
+  SLOWDOWN = 3,  // Slowdown in percentage from current operating speed
+  APPROACH = 4  // Keep constant time interval before collision
 };
 
 /// @brief Action for robot
--- a/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
@@ -64,11 +64,11 @@ struct Pose
 /// @brief Action type for robot
 enum ActionType
 {
-  ACTION_UNKNOWN = 0, // Default for initializing action type
-  DO_NOTHING = 1,  // No action
-  STOP = 2,  // Stop the robot
-  SLOWDOWN = 3,  // Slowdown in percentage from current operating speed
-  APPROACH = 4  // Keep constant time interval before collision
+  DO_NOTHING = 0,  // No action
+  STOP = 1,  // Stop the robot
+  SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
+  APPROACH = 3,  // Keep constant time interval before collision
+  ACTION_UNKNOWN = 4 // Default for initializing action type
 };
 
 /// @brief Action for robot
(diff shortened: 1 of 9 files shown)


nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
```diff
@@ -65,14 +67,16 @@ enum ActionType
   DO_NOTHING = 0,  // No action
   STOP = 1,  // Stop the robot
   SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
-  APPROACH = 3  // Keep constant time interval before collision
+  APPROACH = 3,  // Keep constant time interval before collision
+  ACTION_UNKNOWN = 4 // Default for initializing action type
```
SteveMacenski, line 71: How is this semantically different from just DO_NOTHING?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -105,7 +109,7 @@ CollisionMonitor::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
   process_active_ = false;
 
   // Reset action type to default after worker deactivating
-  robot_action_prev_ = {DO_NOTHING, {-1.0, -1.0, -1.0}};
+  robot_action_prev_ = {ACTION_UNKNOWN, {-1.0, -1.0, -1.0}, ""};
```
SteveMacenski, line 112: L120 below you need to deactivate the new `collision_monitor_state_pub_`

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -480,6 +495,16 @@ void CollisionMonitor::printAction(
   }
 }
 
+void CollisionMonitor::publishActionState(const Action & robot_action) const
```
SteveMacenski, line 498: I'd prefer if this as combined with the print function into a `notifyActionState` method logging + publishing

nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
SteveMacenski: If this is truly different, a default action should be `0` not `4` by convention and auto-initialization

SteveMacenski (changes requested):
Looks good in general, most are nit picks and a couple of questions

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -377,9 +386,12 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
     }
   }
 
-  if (robot_action.action_type != robot_action_prev_.action_type) {
+  if (robot_action_prev_.action_type == ACTION_UNKNOWN ||
+    robot_action.polygon_name != robot_action_prev_.polygon_name)
```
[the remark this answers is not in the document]
kaichie: If I understand correctly, CM support multiple `slowdown` polygon with different `slowdown_ratio`, when there is a change of `slowdown` polygon from lower ratio to higher ratio, the `action_type` will stay the same but the activated polygon will be different. This is the reason I swap to `polygon_name` to reflect the correct polygon.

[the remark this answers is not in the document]
SteveMacenski: Oh ok. Lets see what @AlexeyMerzlyakov thinking but first glance that makes sense to me.

nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
SteveMacenski: I don't see why you should publish any action if the first is `DO_NOTHING` - by definition it's not doing anything. Only just to check that its running? 

I'm not against this idea, just wanting to make sure its solving a real and not a perceived problem

kaichie: > I don't see why you should publish any action if the first is DO_NOTHING - by definition it's not doing anything.

Ah okay, I thought I should, but if that's the case then there is no need to add this new state. I was just using it to publish the first `DO_NOTHING`.

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -4,6 +4,7 @@ collision_monitor:
     odom_frame_id: "odom"
     cmd_vel_in_topic: "cmd_vel_raw"
     cmd_vel_out_topic: "cmd_vel"
+    collision_monitor_state_topic: "collision_monitor_state"
```
AlexeyMerzlyakov, line 7: How about to name parameter just `state_topic` (and for all variable names ditto)?

nav2_collision_monitor/test/collision_monitor_node_test.cpp
```diff
@@ -579,6 +610,8 @@ TEST_F(Tester, testProcessStopSlowdown)
   ASSERT_NEAR(cmd_vel_out_->linear.x, 0.5 * SLOWDOWN_RATIO, EPSILON);
   ASSERT_NEAR(cmd_vel_out_->linear.y, 0.2 * SLOWDOWN_RATIO, EPSILON);
   ASSERT_NEAR(cmd_vel_out_->angular.z, 0.1 * SLOWDOWN_RATIO, EPSILON);
+  ASSERT_EQ(action_state_->action_type, SLOWDOWN);
```
AlexeyMerzlyakov, line 613: For correct test work, we need ensure that `action_state_` was received. For this it is need to:
1. Initialize action_state_ with `nullptr`, as made it for `cmd_vel_out_` in `Tester::publishCmdVel()` routine
2. Add new `Tester::waitActionState()` method like `Tester::waitCmdVel()`
3. Use this method before checking of `action_state_` propierties in each testcase

AlexeyMerzlyakov (commented):
There are some items, I've mentioned, but in overall, this seems a good approach for me

nav2_collision_monitor/src/collision_monitor_node.cpp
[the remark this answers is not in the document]
AlexeyMerzlyakov: This change should be OK: https://github.com/ros-planning/navigation2/pull/3504#discussion_r1147726879.
We are not publishing anything more that needed, or missing something.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -190,6 +197,9 @@ bool CollisionMonitor::getParameters(
   nav2_util::declare_parameter_if_not_declared(
     node, "cmd_vel_out_topic", rclcpp::ParameterValue("cmd_vel"));
   cmd_vel_out_topic = get_parameter("cmd_vel_out_topic").as_string();
+  nav2_util::declare_parameter_if_not_declared(
+    node, "collision_monitor_state_topic", rclcpp::ParameterValue("collision_monitor_state"));
+  collision_monitor_state_topic = get_parameter("collision_monitor_state_topic").as_string();
```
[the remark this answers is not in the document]
SteveMacenski: I disagree, we should always have a logical default for topics.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -77,6 +82,9 @@ CollisionMonitor::on_activate(const rclcpp_lifecycle::State & /*state*/)
 
   // Activating lifecycle publisher
   cmd_vel_out_pub_->on_activate();
+  if (state_pub_ != nullptr) {
```
SteveMacenski, line 85:
```suggestion
  if (state_pub_) {
```

Is enough - its the same but cleaner. Same for deactivate.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -478,6 +497,15 @@ void CollisionMonitor::printAction(
       get_logger(),
       "Robot to continue normal operation");
   }
+
+  if (state_pub_ != nullptr) {
```
SteveMacenski, line 501: Ditto

SteveMacenski (changes requested):
Otherwise LGTM, just need the complimentary documentation updates for the new parameter in the config guide + migration guide for the new feature

kaichie:
Thanks the review @SteveMacenski and @AlexeyMerzlyakov! Created a [PR](https://github.com/ros-planning/navigation.ros.org/pull/405) for updating the config guide + migration guide too.

SteveMacenski (approved):
LGTM. On Alexey's final call to merge.

AlexeyMerzlyakov (approved):
Change reviewed: OK.
Local `colcon test --packages-select nav2_collision_monitor`: OK
Codecov test: OK
Local TB3 simulation verification: OK (CM is working well and publishing a new topic)

PR #3509, opened 2023-03-24
Add an argument to select if the smoother server should be launched

arslogavitabrevis:
---

## Basic Info

| ------ | ----------- |
| Ros version tested on | Humble |

---

I add an argument to the `navigation_launch.py` launch file to allow to select if the smoother server should be launched.

As some planner output are smooth enough to avoid  using a smoother and some  planner have an included smoother, the smoother server may be unnecessary. It may be interesting not to launch it to have one less node.

The argument default value is set to True to keep the actual behavior by default.

---

I wasn't able to add the condition to the smoother composed node as the composed does not have the condition argument. 

#### For Maintainers: 


arslogavitabrevis:
See #3510 for the same pull request in the main branch.

PR #3510, opened 2023-03-24
Add an argument to select if the smoother server should be launched

arslogavitabrevis:
---

## Basic Info

| ------ | ----------- |
| Ros version tested on | Humble |

---

I add an argument to the `navigation_launch.py` launch file to allow to select if the smoother server should be launched.

As some planner output are smooth enough to avoid  using a smoother and some  planner have an included smoother, the smoother server may be unnecessary. It may be interesting not to launch it to have one less node.

The argument default value is set to True to keep the actual behavior by default.

Same as #3509 but in main branch.

---

I wasn't able to add the condition to the smoother composed node as the composed does not have the condition argument. 

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/navigation_launch.py
+++ b/nav2_bringup/launch/navigation_launch.py
@@ -35,19 +35,22 @@ def generate_launch_description():
     autostart = LaunchConfiguration('autostart')
     params_file = LaunchConfiguration('params_file')
     use_composition = LaunchConfiguration('use_composition')
+    use_smoother_server = LaunchConfiguration('use_smoother_server')
     container_name = LaunchConfiguration('container_name')
     container_name_full = (namespace, '/', container_name)
     use_respawn = LaunchConfiguration('use_respawn')
     log_level = LaunchConfiguration('log_level')
 
-    lifecycle_nodes = ['controller_server',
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I appreciate the contribution, but I think that hits a level of granularity at launch that I think would best belong in your company's specific bringup stack (or just remove smoother, as I suspect that's what you actually want to do). These launch files are a good launching point (he he he) for building out your most appropriate navigation system. Its common for companies to have `<company_name>_nav` package containing your specific launch and configuration files to use the features you want to use and drop the ones you don't. 

There are naturally some servers here some users will not use - like the smoother or the waypoint follower. Its totally fine to drop them if you don't need them, and its encouraged! But for the purposes that this package serves as a "sampling" of what Nav2 offers, I think its important to have everything. Since this also doubles as the defacto-reference for launch files, I don't want to overcomplicate what is already fairly complex. 

Does that make sense? I appreciate very much the thought and PR, but I don't think this is sensible to add here, unless I'm missing something :smile:

arslogavitabrevis:
It makes a lot of sense! Thanks you for your word of wisdom and for following up with me on this pull request! :)

Issue #3499, opened 2023-03-22
[Collision monitor] max_points should be called min_points


tonynajjar:
Just a nitpick issue. The max_points parameter would be more intuitive if called min_points

![image](https://user-images.githubusercontent.com/14879660/226908050-40c82bde-b2b8-4782-92f8-016ee0d915bd.png)

becomes: "Minimum number of data readings within a zone to trigger the action. "  

The way it's used [here](https://github.com/ros-planning/navigation2/blob/0822a14b34c412cc4f8e5f2d8486d1bb2aa28fb1/nav2_collision_monitor/src/collision_monitor_node.cpp#LL404C2-L404C2) also sounds better with min_points (check if the collision points are more that the minimum number of points required -> if yes, trigger action)

AlexeyMerzlyakov:
Agree, `min_points` sounds more reasonable. I'll prepare the PR on the days

SteveMacenski:
Renamed!

PR #3513, opened 2023-03-27
Rename max_points to min_points polygon parameter

AlexeyMerzlyakov:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3499 |
| Primary OS tested on | Ubuntu 22.04 w/ ROS2 Rolling built from source (from 20230110) |
| Robotic platform tested on | `colcon test --packages-select nav2_collision_monitor` |

---

* `<polygon_name>.max_points` parameter was renamed to `<polygon_name>.min_points` to track its meaning
* Default value for this parameter was adjusted accordingly to not change default CM's behavior
* Old `<polygon_name>.max_points` is still used  (to be removed in the future releases), but now with deprecation warning, if parameter was set by developer
* New `testPolygonMaxPoints` testcase for `max_points` deprecated parameter

* Update parameter description on CM parameters page
* Switch parameters on Tutorial page

---

* Remove the support of `<polygon_name>.max_points` parameter in the further releases

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/polygon.cpp
+++ b/nav2_collision_monitor/src/polygon.cpp
@@ -128,9 +128,9 @@ ActionType Polygon::getActionType() const
   return action_type_;
 }
 
-int Polygon::getMaxPoints() const
+int Polygon::getMinPoints() const
 {
-  return max_points_;
+  return min_points_;
 }
 
 double Polygon::getSlowdownRatio() const
@@ -212,7 +212,7 @@ double Polygon::getCollisionTime(
     transformPoints(pose, points_transformed);
     // If the collision occurred on this stage, return the actual time before a collision
     // as if robot was moved with given velocity
-    if (getPointsInside(points_transformed) > max_points_) {
+    if (getPointsInside(points_transformed) >= min_points_) {
       return time;
     }
   }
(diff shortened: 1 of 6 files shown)


nav2_collision_monitor/src/polygon.cpp
```diff
@@ -264,8 +264,21 @@ bool Polygon::getCommonParameters(std::string & polygon_pub_topic)
     }
 
     nav2_util::declare_parameter_if_not_declared(
-      node, polygon_name_ + ".max_points", rclcpp::ParameterValue(3));
-    max_points_ = node->get_parameter(polygon_name_ + ".max_points").as_int();
+      node, polygon_name_ + ".min_points", rclcpp::ParameterValue(4));
```
SteveMacenski, line 267: Why change the defaults?

SteveMacenski:
LGTM. Just the docs PR for migration / config guide updates

nav2_collision_monitor/src/polygon.cpp
AlexeyMerzlyakov: This appears since `max_points` parameter meant the maximum number of points where we are still not triggering the action, while `min_points` - is a number starting from the action to be initiated. In other words, `min_points = max_points + 1`.
Here we are:
```
-    if (getPointsInside(points_transformed) > max_points_) {
+    if (getPointsInside(points_transformed) >= min_points_) {
```

PR #3515, opened 2023-03-29
Rpp dubins constraint

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | N/A ||
| Primary OS tested on | Ubuntu ||
| Robotic platform tested on | Custom sim |

---

In ackermann vehicles, if the lookahead point is too short when the vehicle gets significantly off the planned path, it oscillates around the lookahead point since it is trying to reach a point that it cannot reach and straighten out.

This adds the capability to add a lower bound to the lookahead distance based on the minimum turning radius of the vehicle, and how much cross-track error there is. For the sake of simplicity, this doesn't actually generate the dubins curve, but instead just assumes that if the robot is parallel to a path, then the minimum lookahead distance it can reach the path and still be on the correct heading is `2*R*sin(a)`, where `R` is the minimum turning radius, and `a` is the arc length of two symmetric arcs of a dubins curve.

So far, this has helped a lot on my robot.

Added new parameters:
- `use_dubins_min_lookahead_dist`
- `dubins_min_turning_radius`
---

The cross track error may be useful to break out and use for all controllers as a heuristic to track for performance.

#### For Maintainers: 

### DIFF
--- a/nav2_msgs/msg/CrossTrackError.msg
+++ b/nav2_msgs/msg/CrossTrackError.msg
@@ -1,0 +1,2 @@
+float64 position_error
+float64 heading_error
--- a/nav2_msgs/msg/CrossTrackError.msg
+++ b/nav2_msgs/msg/CrossTrackError.msg
@@ -1,2 +1,4 @@
+std_msgs/Header header
+
 float64 position_error
 float64 heading_error
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
@@ -76,7 +76,10 @@ public:
     const geometry_msgs::msg::PoseStamped & in_pose,
     geometry_msgs::msg::PoseStamped & out_pose) const;
 
-  void setPlan(const nav_msgs::msg::Path & path) {global_plan_ = path;}
+  void setPlan(const nav_msgs::msg::Path & path) {
+    global_plan_ = path;
+    path_progress_cursor_ = global_plan_.poses.begin();
+  }
 
   nav_msgs::msg::Path getPlan() {return global_plan_;}
 
@@ -92,6 +95,7 @@ protected:
   std::shared_ptr<tf2_ros::Buffer> tf_;
   std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
   nav_msgs::msg::Path global_plan_;
+  std::vector<geometry_msgs::msg::PoseStamped>::iterator path_progress_cursor_;
 };
 
 }  // namespace nav2_regulated_pure_pursuit_controller
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
@@ -84,6 +84,8 @@ public:
 
   nav_msgs::msg::Path getPlan() {return global_plan_;}
 
+  nav_msgs::msg::Path & getPlan() {return global_plan_;}
+
   std::vector<geometry_msgs::msg::PoseStamped>::iterator getCurrentPose()
   {
     return path_progress_cursor_;
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
@@ -84,8 +84,6 @@ public:
 
   nav_msgs::msg::Path getPlan() {return global_plan_;}
 
-  nav_msgs::msg::Path & getPlan() {return global_plan_;}
-
   std::vector<geometry_msgs::msg::PoseStamped>::iterator getCurrentPose()
   {
     return path_progress_cursor_;
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
@@ -76,12 +76,18 @@ public:
     const geometry_msgs::msg::PoseStamped & in_pose,
     geometry_msgs::msg::PoseStamped & out_pose) const;
 
-  void setPlan(const nav_msgs::msg::Path & path) {
+  void setPlan(const nav_msgs::msg::Path & path)
+  {
     global_plan_ = path;
     path_progress_cursor_ = global_plan_.poses.begin();
   }
 
   nav_msgs::msg::Path getPlan() {return global_plan_;}
+
+  std::vector<geometry_msgs::msg::PoseStamped>::iterator getCurrentPose()
+  {
+    return path_progress_cursor_;
+  }
 
 protected:
   /**
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -143,7 +143,9 @@ double RegulatedPurePursuitController::getLookAheadDistance(
   }
 
   if (params_->use_dubins_min_lookahead_dist) {
-    double dubins_min_lookahead_dist = std::min(getDubinsMinLookAheadDistance(cross_track_error), params_->max_lookahead_dist);
+    double dubins_min_lookahead_dist = std::min(
+      getDubinsMinLookAheadDistance(
+        cross_track_error), params_->max_lookahead_dist);
     lookahead_dist = std::max(lookahead_dist, dubins_min_lookahead_dist);
   }
 
@@ -176,7 +178,10 @@ double RegulatedPurePursuitController::getDubinsMinLookAheadDistance(
   // symmetrical arcs of `arc_length` radians to reach the path.
   // These serve to give us a reasonable bound on the lookahead
   // without generating full dubins curves
-  double arc_length = std::acos(1.0 - cross_track_error.position_error / (2.0 * params_->dubins_min_turning_radius));
+  double arc_length =
+    std::acos(
+    1.0 - cross_track_error.position_error /
+    (2.0 * params_->dubins_min_turning_radius));
   return 2 * params_->dubins_min_turning_radius * std::sin(arc_length);
 }
 
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -120,7 +120,9 @@ std::unique_ptr<geometry_msgs::msg::PointStamped> RegulatedPurePursuitController
 }
 
 double RegulatedPurePursuitController::getLookAheadDistance(
-  const geometry_msgs::msg::Twist & speed)
+  const geometry_msgs::msg::Twist & speed,
+  const nav_msgs::msg::Path & path,
+  const geometry_msgs::msg::PoseStamped & pose)
 {
   // If using velocity-scaled look ahead distances, find and clamp the dist
   // Else, use the static look ahead distance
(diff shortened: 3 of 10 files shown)


Aposhian:
I need to add parameter docs, and doc comments, etc. but wanted to get buy off on the overall approach first.

nav2_msgs/msg/CrossTrackError.msg
vinnnyr, line 1: are you opposed to stamping this message? that was systems external to nav2 could use it  along side message filters.

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -179,6 +219,120 @@ inline double calculate_path_length(const nav_msgs::msg::Path & path, size_t sta
...
+    [&pose](const auto & path_pose) {
+      return squared_euclidean_distance(path_pose, pose);
+    }
+  );
+  // We want to find the first local minima,
+  // so we don't loop around and get something from later
+  // if we start out by increasing distance away,
+  // then we will accept the beginning
+  auto distance_it = distances.begin();
+  for (; distance_it != std::prev(distances.end()); ++distance_it) {
```
SteveMacenski, line 249: I don't think this necessarily works. Imagine you have a path going laterally then does a 90 degree turn up. The robot is somewhere in the upward portion of the path.

If the path going laterally has even a slight disturbance in it, then the differences in distances could be increasing by a slight amount to trigger this. 

Further, in the `std::transform` you're already owning an entire path-wise iteration. Why not just own it then and do the logic in that loop instead of using the transform function? This is actually now doing nearly double the compute, in the worst case.

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -119,6 +121,44 @@ inline double euclidean_distance(
   return std::hypot(dx, dy);
 }
 
+/**
```
SteveMacenski, line 124: All functions need unit testing

nav2_msgs/msg/CrossTrackError.msg
SteveMacenski: Agreed!

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
```diff
@@ -76,10 +76,19 @@ class PathHandler
     const geometry_msgs::msg::PoseStamped & in_pose,
     geometry_msgs::msg::PoseStamped & out_pose) const;
 
-  void setPlan(const nav_msgs::msg::Path & path) {global_plan_ = path;}
+  void setPlan(const nav_msgs::msg::Path & path)
+  {
+    global_plan_ = path;
+    path_progress_cursor_ = global_plan_.poses.begin();
```
SteveMacenski, line 82: This seems to accomplish the same thing as the path erasing logic did previously without storing extra information (and erasing that which is no longer relevant). Can you explain this change?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -73,6 +73,10 @@ void RegulatedPurePursuitController::configure(
 
   global_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("received_global_plan", 1);
   carrot_pub_ = node->create_publisher<geometry_msgs::msg::PointStamped>("lookahead_point", 1);
+  projected_pose_pub_ =
```
SteveMacenski, line 76: The projected pose looks mostly for debugging value. Is there a reason to keep this in the final form? I can understand publishing the cross track error

SteveMacenski (changes requested):
My sincerest apologies for not getting back to you sooner on this. As you can imagine, this coming in about 2 weeks after I was laid off, I had other things on my mind and in my calendar. 

I've had a chance now to look this over and I generally approve of all of it beyond just details. In addition to the specific comments I made, I think that we should create a breakout class (or at least file of utils) for the ackermann-specific things so that we don't muddle up the main controller code as much as possible. The architecture here is already a little flimpsy and I'm trying to make a half-hearted attempt to keep this understandable to the average human

nav2_util/include/nav2_util/geometry_utils.hpp
Aposhian: You may be right that this algorithm does not yield the closest segment to the path in general. However, for the context in which it is used, so far it has worked quite well, since RPP already continuously prunes up to the waypoint closest to the robot. So this is just a search after the fact, given the assumption that the robot is within a few waypoints of the beginning of the path. I think your counter-example is correct, but it would also be extremely unlikely given how RPP is used. However, if cross-track error is to be generalized, then I think this would need to be fixed.

Yes, these calculations could be consolidated into one for loop. The std::transform isn't needed.

nav2_util/include/nav2_util/geometry_utils.hpp
Aposhian: - [ ] add unit tests for new helper functions

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
Aposhian: Yes, this is designed to do the same thing as the path erasing logic. However, the key difference is retaining the erased poses so that the cross-track error calculation can look backward to those if it needs to. Otherwise, the cross-track error is too spiky since you are always off the end of the pruned path.

As for the pruning/erasing/cursor logic in general, it remains relevant since it is half of what keeps the closest waypoint logic from snapping to waypoints that have already been traveled past.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Aposhian: Yes this is for debugging only, so I could add a parameter to enable and disable it. It helps to visualize where the cross-track error is being calculated from.

- [ ] Make projected pose pub configurable

nav2_msgs/msg/CrossTrackError.msg
Aposhian: - [ ] Add header to cross track error message

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: I meant more a development artifact, but if you think its useful as a life debugging instrument you think folks will use, no problem. Might want to rename the topic to give some more context then

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
SteveMacenski: Is there a reason not to just do the erase but just keep `i-1`'s value? I think the cross track error calculation just uses the `i-1` and `i+1` th values

Aposhian: Yes that would work, but what do you gain by erasing vs just keeping a cursor? It's probably negligible, but erasing probably uses more computation (requires reallocating the path every time).

Aposhian: Well, maybe that part about reallocation is not true, but it will relocate all elements after the erased portion: https://en.cppreference.com/w/cpp/container/vector/erase
> Complexity
Linear: the number of calls to the destructor of T is the same as the number of elements erased, the assignment operator of T is called the number of times equal to the number of elements in the vector after the erased elements

nav2_util/include/nav2_util/geometry_utils.hpp
Aposhian: Thanks for the ideas. I will take a closer look at this, and when I do I will try to come up with some unit tests that test this.
- [ ] Make projected pose algorithm more robust
- [ ] Add unit tests for different paths to project onto.

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/path_handler.hpp
SteveMacenski: Its one less thing being stored as state that can get screwed up in the future :-) 

Or someone that didn't pay attention closely not realizing they needed to start at the cursor would just start at `begin()` because its seems reasonable for a new feature. 

Mostly, just simpler bookkeeping

Aposhian: That makes sense that it could be a point of mistakes where you accidentally use begin. Although I would argue that explicit state is better than implicit state.

Also, if you have to store one pose backwards, then you have the same problem where you could use `begin()` accidentally erroneously, since most of the references will have to change to `begin() + 1`.

SteveMacenski: I think this is just a point of preference. In general, if it were just you and I modifying this code, I think that I wouldn't care very much. However, that `begin()` versus using the cursor is a subtle mistake that wouldn't be immediately clear to me when reviewing a PR if it wasn't loaded in my brain's RAM that day and I could see that introducing a bug because of the nature of open-source. Since all the other instances of things like this operate in this way, this would be a single special case. If that was absolutely necessary, then I wouldn't argue with it. 

But if there's a way to keep that consistency, its less likely that 3 years from now we introduce a sutble bug when this is long since out of thought / mind.

> Also, if you have to store one pose backwards, then you have the same problem where you could use begin() accidentally erroneously, since most of the references will have to change to begin() + 1.

I don't understand. We don't assume that the first point is the closest necessarily. We're just assuming that we've removed absolutely erroneous information.

Aposhian: No what I'm saying is that if you remove the cursor, and use the existing erase logic but back pad one pose, then you have to convert all other references of begin() to begin() + 1 to compensate. And it's very possible that we need to back pad more than one pose, so then you should probably store that constant offset in a variable, like pose_look behind or something. And then all the references that iterate through that path (except path error) need to change to begin() + pose_lookbehind. And my point is that just storing an iterator that points to the point in the path you are interested in is more explicit than remembering to use begin and an offset variable. No matter whether you use an iterator cursor or not, there are many places in the code where begin() alone is not correct, and erasing doesn't make that pitfall less likely.

Aposhian: So I think the choice is not whether to have a separate variable or not. The choice is whether that variable is an offset (which would likely be parameterized), or an iterator (which would not require a separate parameter for how long the lookbehind should be).

SteveMacenski: > convert all other references of begin() to begin() + 1 to compensate

Why? I agree if we had to do something like that its better to use the cursor. 

I don't see why back padding by 1 path point is necessary in any of the loops. We're never assuming that the first point has any special characteristic (e.g. closest point). We just assume that we've erased any points before that point which are no longer relevant.

SteveMacenski: OK that makes sense, lets keep the cursor. Lets just make sure to update all appropriate locations for it to be used :-) I don't know if there's somewhere natural to document this to make it clear (or rename the path object), but spend a minute or two trying to see if we can prevent future idiots like me from screwing this up :wink: 

The "best" solution, which I'm not necessarily asking you to do, would be to make a new struct containing the path and the cursor with some functions for iterating through it that keeps track of the cursor. That seems to me to be the most idiot-proof solution, but you might have something more clever!

Aposhian:
This implementation has a problem: when cross track error is greater than 2 * dubins_min_turning_radius, the acos starts getting smaller instead of larger.

- [ ] Fix formula to handle larger off-path errors

SteveMacenski:
Thanks for the note! Good to know :-)

Aposhian:
I probably will not get around to making the rest of the changes on this.

Issue #3514, opened 2023-03-28
[Collision monitor] Slowdown field that limit speed, not reduce it


BriceRenaudeau:
## Feature request
We would like a zone that clamps the command if there is an obstacle inside.
In our use case, we prefer to have a slowdown field that limits the speed to a fixed value. Now it is multiplying it with a coef.

Now, if there is an obstacle in the field and I send a small command the output will be even slower.

#### Feature description
We would like a field that limits the command to a fixed value.

That's the behavior of a real low-level safety field. 

#### Implementation considerations
What about robots with different min(-) and max(+) velocities? Use a single value with a *-1 or add a parameter for min_vel?

It can be another model `Limit model` or a parameter to change the behavior of the `slowdown model`.

BriceRenaudeau:
Thanks, we will make it soon.

BriceRenaudeau:
While implementing the new mode for the slowdown, I encountered an issue. It's pretty sure we will want to have no one limit for the 3 components (x, y, tw) but 3 different limits. Also `Action robot_action` doesn't have a place for a different mode of slowdown.

Hence I think it's better to add a new type (LIMIT) than to add several parameters to the slowdown type.

SteveMacenski:
How we've done this in the speed limiting costmap filter is to use the transitional velocity absolute limit and then scale back the angular velocity proportional to it (e.g. if the max is 1m/s and you set it to 0.5m/s via absolute, we calculate the percentage reduction that represents and apply it to the angular velocity).

Would that be sufficient? If not, then having separate values is sensible.

SteveMacenski:
PR imminently to be merged

PR #3519, opened 2023-03-30
(collision monitor) add limit polygon type

BriceRenaudeau:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3514 |

---

* I added  `LIMIT = 4` in the polygon types enum
* I added 3 separate parameters `limit_...` for x, y and tw.
* I perform a simple clamp on the incoming velocity
* I added the test for this polygon

* I updated the README
* I added a limit polygon example in the .yaml

---

* Update the docs

#### For Maintainers: 


BriceRenaudeau:
Hi, The build failed to build the test for the last merge: collision monitor max_point -> min_point #3513.
I have no modification on that line. I will try to fix it.

BriceRenaudeau:
Ci Tests failed in other packages.

BriceRenaudeau:
Good, now it's working. The CI failed on another node.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -436,6 +436,19 @@ bool CollisionMonitor::processStopSlowdown(
         robot_action.req_vel = safe_vel;
         return true;
       }
+    } else {  // Limit
+      Velocity safe_vel;
+      safe_vel.x = std::clamp(velocity.x, -polygon->getLimitX(), polygon->getLimitX());
+      safe_vel.y = std::clamp(velocity.y, -polygon->getLimitY(), polygon->getLimitY());
```
AlexeyMerzlyakov, line 442: For omni-drives the linear velocity could have two OX and OY components simultaneously. There could be an issue when we are limiting each component on a some value, while total maximum allowed speed of robot could exceed this maximum value. For example, the maximum speed in some area is `0.5 m/s`. Omni-drive is moving with `Vx = 0.5` and `Vy = 0.5`. Summary linear speed of robot is `Vlin = ~1.41 * 0.5 = 0.7 m/s` which is higher than limit.

Thus, I'd like recommend to calculate the absolute value of linear velocity, and then restrict each OX/OY component proportionally on clamped value. Something, like made in https://github.com/ros-planning/navigation2/blob/ae887b1eb91bbce5de8f4c8787946acc9af2e384/nav2_dwb_controller/dwb_plugins/src/kinematic_parameters.cpp#L153-L155 but instead of `base_max_vel_x` it will be used current absolute linear speed value.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -35,7 +35,8 @@ Polygon::Polygon(
   const std::string & base_frame_id,
   const tf2::Duration & transform_tolerance)
 : node_(node), polygon_name_(polygon_name), action_type_(DO_NOTHING),
-  slowdown_ratio_(0.0), footprint_sub_(nullptr), tf_buffer_(tf_buffer),
+  slowdown_ratio_(0.0), limit_x_(0.0), limit_y_(0.0), limit_tw_(0.0),
```
AlexeyMerzlyakov, line 38: The same: I do not think, we need to have separate `limit_x` and `limit_y` values. It is more reasonable to have `limit_linear` and `limit_tw` values.

AlexeyMerzlyakov (commented):
Good approach to set an absolute values for limiting the maximum robot's velocities. One note/change about linear velocities from this review comments, and I think it is OK for merge.

nav2_collision_monitor/src/collision_monitor_node.cpp
BriceRenaudeau: I think it's not the duty of the collision monitor to manage the maximum velocity of the X/Y combination, it's more velocity smoother.

But you are right on one point if we reduce only one component, the motion line will change and that's not good behavior.

I used separate limits to fit the X/Y/TW nav2::velocity design. I am not an expert in omni-drive and I don't know the expected behavior. If the robot is following a wall, it may want to only reduce speed in Y or X axis?

The approach you propose looks good to me. I will implement it.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -426,7 +426,7 @@ bool CollisionMonitor::processStopSlowdown(
       robot_action.req_vel.y = 0.0;
       robot_action.req_vel.tw = 0.0;
       return true;
-    } else {  // SLOWDOWN
+    } else if (polygon->getActionType() == SLOWDOWN) {  // SLOWDOWN
```
AlexeyMerzlyakov, line 429: There is no need to keep the comment: it is excplicitly seen which branch we are using from on now.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -436,6 +436,26 @@ bool CollisionMonitor::processStopSlowdown(
         robot_action.req_vel = safe_vel;
         return true;
       }
+    } else {  // Limit
+      // Compute linear velocity
+      const double linear_vel = std::sqrt(velocity.x * velocity.x + velocity.y * velocity.y); // absolute
```
AlexeyMerzlyakov, line 441: std::hypot(velocity.x, velocity.y) ?

AlexeyMerzlyakov (commented):
A few nitpick comments yet, and this is OK ToGo.

Local testing details:
colcon test: OK
code coverage: 100% new functionality covered

nav2_collision_monitor/src/collision_monitor_node.cpp
AlexeyMerzlyakov: We need to preserve the robot moving direction. If we will reduce only one component of speed, robot will change its direction, which will be at least confusing.
Moreover, usually we are talking about maximum allowed linear and angular velocities for robots in some areas, so that makes also sense from this point of view.

nav2_collision_monitor/src/collision_monitor_node.cpp
BriceRenaudeau: Done

nav2_collision_monitor/src/collision_monitor_node.cpp
BriceRenaudeau: Done

AlexeyMerzlyakov (approved):
LGTM. @SteveMacenski, could you please add your opinion about the merging of this PR?

AlexeyMerzlyakov:
Also, @BriceRenaudeau, please prepare the documentation update PR on https://github.com/ros-planning/navigation.ros.org with according new type of behavior.

BriceRenaudeau:
Any news?

nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
```diff
@@ -67,7 +67,8 @@ enum ActionType
   DO_NOTHING = 0,  // No action
   STOP = 1,  // Stop the robot
   SLOWDOWN = 2,  // Slowdown in percentage from current operating speed
-  APPROACH = 3  // Keep constant time interval before collision
+  APPROACH = 3,  // Keep constant time interval before collision
+  LIMIT = 4  // Limit current operating speed
```
SteveMacenski, line 71:
```suggestion
  LIMIT = 4  // Limit absolute velocity from current operating speed
```

SteveMacenski:
Whoops sorry, I thought Alexey was going to handle this one so I muted myself. Getting to this right now. Thanks for the ping @doisyg

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -286,6 +299,15 @@ bool Polygon::getCommonParameters(std::string & polygon_pub_topic)
       slowdown_ratio_ = node->get_parameter(polygon_name_ + ".slowdown_ratio").as_double();
     }
 
+    if (action_type_ == LIMIT) {
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + ".linear_limit", rclcpp::ParameterValue(0.5));
+      linear_limit_ = node->get_parameter(polygon_name_ + ".linear_limit").as_double();
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + ".angular_limit", rclcpp::ParameterValue(0.5));
```
SteveMacenski, line 307: Is 0.5 a reasonable default for angular velocities? I think that seems a bit low

SteveMacenski (approved):
The only thing I want to see if the tutorial updated to include the limit stuff https://navigation.ros.org/tutorials/docs/using_collision_monitor.html 

@AlexeyMerzlyakov can you handle the tutorial update? I think it could be as easy as adding a new `Note` to mention that there are also LIMIT zones to limit max velocities when obstacles are detected within a zone.

SteveMacenski:
@BriceRenaudeau you also need to rebase for CI to turn over so I can check on the test coverage metrics

nav2_collision_monitor/src/polygon.cpp
BriceRenaudeau: On a circular robot, 0.5rad/s may be slow, but for a robot that needs to check collisions when turning it is a reasonable value when an obstacle is in the limited area.

For the linear part, the safety devices are often set with multiple polygons with decreasing velocity limits when closer to the robot. 0.5m/s is a middle choice between the possible values.

Issue #2508, opened 2021-08-13
[Planner] possible use-after-free bug in nav2_costmap_2d


easylyou:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04.2 LTS
- ROS2 Version:
  - Foxy
- Version or commit hash:
  - a7955d8
- DDS implementation:
  - default

#### Steps to reproduce issue

I had done some tests. And it can be reproduced.

```
[INFO] [1628747203.189238115] [global_costmap.global_costmap]: Configuring
[INFO] [1628747203.199641398] [global_costmap.global_costmap]: Using plugin "static_layer"
[INFO] [1628747203.205795632] [global_costmap.global_costmap]: Subscribing to the map topic (/map) with transient local durability
[INFO] [1628747203.208597874] [global_costmap.global_costmap]: Initialized plugin "static_layer"
[INFO] [1628747203.208680902] [global_costmap.global_costmap]: Using plugin "obstacle_layer"
[INFO] [1628747203.211488645] [global_costmap.global_costmap]: Subscribed to Topics: scan
[INFO] [1628747203.218346306] [global_costmap.global_costmap]: Initialized plugin "obstacle_layer"
[INFO] [1628747203.218425410] [global_costmap.global_costmap]: Using plugin "voxel_layer"
[INFO] [1628747203.220963960] [global_costmap.global_costmap]: StaticLayer: Resizing costmap to 384 X 384 at 0.050000 m/pi
```

According to the log of a real launch, it is possible that a resizing costmap of staticLayer happens before voxel_layer complete the initializing.

```
in navigation2/nav2_voxel_grid/src/voxel_grid.cpp

void VoxelGrid::resize(unsigned int size_x, unsigned int size_y, unsigned int size_z)
{
  ...
  //2. a resize happends
  delete[] data_;
  
  ...

  data_ = new uint32_t[size_x_ * size_y_];

  ...
}

void VoxelGrid::reset()
{
  //1. VoxelGrid is initializing
  //3. boom! use after free
  uint32_t * col = data_;
  for (unsigned int i = 0; i < size_x_ * size_y_; ++i) {
    *col = unknown_col;
    ++col;
  }
}
```

#### Expected behavior
Process should not crash.

#### Actual behavior
```
==178418==ERROR: AddressSanitizer: heap-use-after-free on address 0x6020000560f0 at pc 0x7fc6b81ae43a bp 0x7ffc328e9410 sp 0x7ffc328e9408
[...]

PR #3522, opened 2023-03-31
fix data race: prohibit resizeMap() during plugin/filter initialization

dirkmb:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #2508 |
| Primary OS tested on | Ubuntu 22.04 & 20.04 |
| Robotic platform tested on | my hardware robot |

---

- all filters and plugins have a `layered_costmap_` variable. 
- it's set when `initialize()` is called
- so it's not yet set when the filter/plugin is added to the costmap
- when `resizeMap()` is called after a filter/plugin was added but before it was initialized a segmentation fault could happen.

---

#### For Maintainers: 


SteveMacenski (approved):
We're having some CI issues. Doesn't seem related to you but need to get it to turn over before merging

PR #3523, opened 2023-04-01
Enable Visualizations for Dev Container

ruffsl:
Enable the use of visualization tools such as:
- https://github.com/ros2/rviz
- https://github.com/foxglove/studio
- https://github.com/osrf/gzweb

PR #3527, opened 2023-04-06
Add headless and use_rviz LaunchConfigurations to demo launch files

ruffsl:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3523 |

---

Add headless and use_rviz LaunchConfigurations in nav2_simple_commander demo launch files, for whether to start rviz or gzclient, to simplify their use in headless environments:

- https://github.com/ros-planning/navigation2/pull/3523

None, as the defaults maintain the existing launch file behavior.

---

#### For Maintainers: 

### DIFF
--- a/nav2_simple_commander/launch/assisted_teleop_example_launch.py
+++ b/nav2_simple_commander/launch/assisted_teleop_example_launch.py
@@ -18,8 +18,10 @@ import os
 from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import ExecuteProcess, IncludeLaunchDescription
+from launch.actions import DeclareLaunchArgument, ExecuteProcess, IncludeLaunchDescription
+from launch.conditions import IfCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
+from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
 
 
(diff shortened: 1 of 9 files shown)


SteveMacenski:
I think this isn't a bad idea of these were meant to be more than just simple demos. I don't want to complicate the simple introduction demos with too many options and bells and whistles. These were intentionally barebones (just like the demo code itself) so that people could work up from here to the more complex launch systems and demos available in the nav2_bringup package. 

So, overall, I'd actually close this PR. Its not a bad thought, but this is the `simple_commander`  package and I'd like to keep it true to that name as much as possible since this package is primarily for new-user tutorial value.

ruffsl:
Well, but then these demos can't easily be accessed from within a dev container, which is explicitly a headless environment.

I'd really like to reuse these demos to showcase and document the visualization PR I've been working on. The rest of the turtlebot 2 simulation demos require a lot of manual interaction that would normally be done through rviz, but not easily replicable with Foxglove out of the box.

In contrast these demos, such as the security demo, are convenient in that they spawn the simulation, then start moving the robot automatically, without user intervention, allowing me to focus on presenting the user with the tools as soon as they launch, rather than repeating instructions on just how to get the robot to localize, move, and perpetually repeat to see anything in action. While following along with dev container documentation, users may be switching between many windows or browser tabs, so having the robot perpetually move in the background offloads this task, allowing them to focus on new design tools.

I could copy just the security launch file at least to somewhere else, but that seems like needless duplication.

ruffsl:
Additionally, given the target audience for these tutorials, i.e. Python programers, are more likely to be running mainstream OSs such as MacOS or Windows, as compared to conventional ROS users, it would behove such an audience to provide the means to quickly run and experiment with the demos without having to muck about with dual booting, virtual machines, X11 forwarding, or VNC port mapping on their own development setups. Installing and setting up ROS natively is not a trivial task for either MacOS, Windows, or WSL, especially when teaching large class sizes of young and novice python and robotics programers.

Enabling headless runtime and dev containers would afford them a simpler one-click jump start into the `simple_commander` package, as compared to what is currently possible.

SteveMacenski:
> Well, but then these demos can't easily be accessed from within a dev container, which is explicitly a headless environment.

Ah I see, OK if there's a particular reason that isn't locally changable I think its OK. I'll take another look.

SteveMacenski (approved):

Issue #3526, opened 2023-04-06
[velocity_smoother] accel and deccel are inverted for negative speeds


doisyg:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble
- Version or commit hash:
  - lastest main
- DDS implementation:
  - Cyclone

#### Steps to reproduce issue

```
    velocity_smoother = Node(
        package='nav2_velocity_smoother',
        name='root_velocity_smoother',
        executable='velocity_smoother',
        parameters=[
            # {'use_sim_time': use_sim_time},
            {'smoothing_frequency': 20.0},
            {'scale_velocities': False},
            {'feedback': 'OPEN_LOOP'},
            {'max_velocity': [0.5, 0.0, 0.8]},
            {'min_velocity': [-0.5, 0.0, -0.8]},
            {'velocity_timeout': 0.5},
            {'max_accel': [0.1, 0.0, 0.2]},
            {'max_decel': [-0.5, 0.0, -1.2]},
            ],
        remappings=[
            ('cmd_vel', 'cmd_vel_protected'),
            ('cmd_vel_smoothed', 'cmd_vel')],
        output='screen')
```

#### Expected behavior

accel and deccel are applied similarly on both direction

#### Actual behavior

accel and deccel are inverted when going backward (here 0.1m/s2 accel and 0.5 m/s2 deccel)

Forward:
![Screenshot from 2023-04-06 10-00-10](https://user-images.githubusercontent.com/15727892/230331966-b71b2659-45b7-45ab-920b-b2c730ea9ea5.png)

Backward:
![Screenshot from 2023-04-06 10-02-11](https://user-images.githubusercontent.com/15727892/230332007-d726859b-b60f-4fd1-8e34-50d98fbc8361.png)

Same for angular accel/deccel

![Screenshot from 2023-04-07 23-15-23](https://user-images.githubusercontent.com/15727892/230685667-68b1f976-587e-429d-ab88-694ba2a5ec50.png)

![Screenshot from 2023-04-07 23-15-50](https://user-images.githubusercontent.com/15727892/230685677-a78e13ff-a12d-4645-b66e-5c590b42ba58.png)

EDIT: angular velocities
Reporting for now, PR later

PR #3529, opened 2023-04-07
[velocity_smoother] Fix accel and deccel inverted for negative speeds

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/3526 |
| Robotic platform tested on | Dexory Sim |

---

Fix https://github.com/ros-planning/navigation2/issues/3526

Tested only in sim, can test on real hardware next week. I would not mind a second pair of eyes / independent test on this one before merging as it is quite critical (in our system it is the last layer before the hardware commands). @richardw347 

With :

```
    velocity_smoother = Node(
        package='nav2_velocity_smoother',
        name='root_velocity_smoother',
        executable='velocity_smoother',
        parameters=[
            # {'use_sim_time': use_sim_time},
            {'smoothing_frequency': 20.0},
            {'scale_velocities': False},
            {'feedback': 'OPEN_LOOP'},
            {'max_velocity': [0.5, 0.0, 0.8]},
            {'min_velocity': [-0.5, 0.0, -0.8]},
            {'velocity_timeout': 0.5},
            {'max_accel': [0.1, 0.0, 0.2]},
            {'max_decel': [-0.5, 0.0, -1.2]},
            ],
        remappings=[
            ('cmd_vel', 'cmd_vel_protected'),
            ('cmd_vel_smoothed', 'cmd_vel')],
        output='screen')
```
Screenshot from 2023-04-07 22-44-22
Screenshot from 2023-04-07 22-44-53
Screenshot from 2023-04-07 23-06-32
Screenshot from 2023-04-07 23-07-07

#### For Maintainers: 

### DIFF
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -206,9 +206,17 @@ double VelocitySmoother::applyConstraints(
   double v_component_max;
   double v_component_min;
 
-  if (abs(v_cmd) >= abs(v_curr)) {
-    v_component_max = accel / smoothing_frequency_;
-    v_component_min = -accel / smoothing_frequency_;
+  // Accelerating if magnitude of v_cmd is above magnitude of v_curr
+  // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
+  // Deccelerating otherwise
+  if (v_curr * v_cmd >= 0.0) {
+    if (abs(v_cmd) >= abs(v_curr)) {
+      v_component_max = accel / smoothing_frequency_;
+      v_component_min = -accel / smoothing_frequency_;
+    } else {
+      v_component_max = -decel / smoothing_frequency_;
+      v_component_min = decel / smoothing_frequency_;
+    }
   } else {
     v_component_max = -decel / smoothing_frequency_;
     v_component_min = decel / smoothing_frequency_;
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -190,7 +190,7 @@ double VelocitySmoother::findEtaConstraint(
   // Accelerating if magnitude of v_cmd is above magnitude of v_curr
   // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
   // Deccelerating otherwise
-  if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
+  if (abs(v_cmd) >= abs(v_curr) && v_curr * v_cmd >= 0.0) {
     v_component_max = accel / smoothing_frequency_;
     v_component_min = -accel / smoothing_frequency_;
   } else {
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -221,7 +221,7 @@ double VelocitySmoother::applyConstraints(
   // Accelerating if magnitude of v_cmd is above magnitude of v_curr
   // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
   // Deccelerating otherwise
-  if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
+  if (abs(v_cmd) >= abs(v_curr) && v_curr * v_cmd >= 0.0) {
     v_component_max = accel / smoothing_frequency_;
     v_component_min = -accel / smoothing_frequency_;
   } else {
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -182,9 +182,21 @@ double VelocitySmoother::findEtaConstraint(
   const double v_curr, const double v_cmd, const double accel, const double decel)
 {
   // Exploiting vector scaling properties
-  const double v_component_max = accel / smoothing_frequency_;
-  const double v_component_min = decel / smoothing_frequency_;
-  const double dv = v_cmd - v_curr;
+  double dv = v_cmd - v_curr;
+
+  double v_component_max;
+  double v_component_min;
+
+  // Accelerating if magnitude of v_cmd is above magnitude of v_curr
+  // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
+  // Deccelerating otherwise
+  if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
+    v_component_max = accel / smoothing_frequency_;
+    v_component_min = -accel / smoothing_frequency_;
+  } else {
+    v_component_max = -decel / smoothing_frequency_;
+    v_component_min = decel / smoothing_frequency_;
+  }
 
   if (dv > v_component_max) {
     return v_component_max / dv;
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -202,8 +202,18 @@ double VelocitySmoother::applyConstraints(
   const double accel, const double decel, const double eta)
 {
   double dv = v_cmd - v_curr;
-  const double v_component_max = accel / smoothing_frequency_;
-  const double v_component_min = decel / smoothing_frequency_;
+
+  double v_component_max;
+  double v_component_min;
+
+  if (abs(v_cmd) >= abs(v_curr)) {
+    v_component_max = accel / smoothing_frequency_;
+    v_component_min = -accel / smoothing_frequency_;
+  } else {
+    v_component_max = -decel / smoothing_frequency_;
+    v_component_min = decel / smoothing_frequency_;
+  }
+
   return v_curr + std::clamp(eta * dv, v_component_min, v_component_max);
 }
 
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -209,14 +209,9 @@ double VelocitySmoother::applyConstraints(
   // Accelerating if magnitude of v_cmd is above magnitude of v_curr
   // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
   // Deccelerating otherwise
-  if (v_curr * v_cmd >= 0.0) {
-    if (abs(v_cmd) >= abs(v_curr)) {
-      v_component_max = accel / smoothing_frequency_;
-      v_component_min = -accel / smoothing_frequency_;
-    } else {
-      v_component_max = -decel / smoothing_frequency_;
-      v_component_min = decel / smoothing_frequency_;
-    }
+  if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
+    v_component_max = accel / smoothing_frequency_;
+    v_component_min = -accel / smoothing_frequency_;
   } else {
     v_component_max = -decel / smoothing_frequency_;
     v_component_min = decel / smoothing_frequency_;
--- a/nav2_velocity_smoother/src/velocity_smoother.cpp
+++ b/nav2_velocity_smoother/src/velocity_smoother.cpp
@@ -189,7 +189,7 @@ double VelocitySmoother::findEtaConstraint(
 
   // Accelerating if magnitude of v_cmd is above magnitude of v_curr
   // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
-  // Deccelerating otherwise
+  // Decelerating otherwise
   if (abs(v_cmd) >= abs(v_curr) && v_curr * v_cmd >= 0.0) {
     v_component_max = accel / smoothing_frequency_;
     v_component_min = -accel / smoothing_frequency_;
@@ -220,7 +220,7 @@ double VelocitySmoother::applyConstraints(
 
   // Accelerating if magnitude of v_cmd is above magnitude of v_curr
   // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
-  // Deccelerating otherwise
+  // Decelerating otherwise
   if (abs(v_cmd) >= abs(v_curr) && v_curr * v_cmd >= 0.0) {
     v_component_max = accel / smoothing_frequency_;
     v_component_min = -accel / smoothing_frequency_;
--- a/nav2_velocity_smoother/test/test_velocity_smoother.cpp
+++ b/nav2_velocity_smoother/test/test_velocity_smoother.cpp
@@ -179,17 +179,56 @@ TEST(VelocitySmootherTest, testfindEtaConstraint)
   // default frequency is 20.0
   smoother->configure(state);
 
-  // In range
-  EXPECT_EQ(smoother->findEtaConstraint(1.0, 1.0, 1.5, -2.0), -1);
-  EXPECT_EQ(smoother->findEtaConstraint(0.5, 0.55, 1.5, -2.0), -1);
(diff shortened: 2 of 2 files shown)


AlexeyMerzlyakov:
@doisyg, thank you for pointing this problem out!

Locally checked on TB3 rotating at the place - the change is working properly.
Making acceleration and deceleration to be different as follows at the parameter file below:
```
velocity_smoother:
  ros__parameters:
    smoothing_frequency: 20.0
    scale_velocities: False
    feedback: "OPEN_LOOP"
    max_velocity: [0.15, 0.0, 0.7]
    min_velocity: [-0.15, 0.0, -0.7]
    max_accel: [0.1, 0.0, 0.2] # <- angular acceleration = 0.2
    max_decel: [-0.5, 0.0, -0.5] # <- angular deceleration = 0.5
    odom_topic: "odom"
    odom_duration: 0.1
    deadband_velocity: [0.0, 0.0, 0.0]
    velocity_timeout: 1.0
```

The picture on vanilla velocity smoother is getting unsymmetrical: robot accelerates and decelerates differently depending on rotation: clock- or counterclock- wise, which is incorrect behavior; depicted below (OY - angular rotational velocity):
vanilla_angular

Applying the patch making the robot to accelerate and decelerate correctly in all cases (velocity is higher/lower than required, having both signs), as follows from the measurements below:
patched_angular

Detailed clockwise rotation:
patched_angular_clockwise

Detailed counter clockwise rotation:
patched_angular_counter_clockwise

nav2_velocity_smoother/src/velocity_smoother.cpp
```diff
@@ -202,8 +202,18 @@ double VelocitySmoother::applyConstraints(
   const double accel, const double decel, const double eta)
 {
   double dv = v_cmd - v_curr;
-  const double v_component_max = accel / smoothing_frequency_;
-  const double v_component_min = decel / smoothing_frequency_;
+
+  double v_component_max;
+  double v_component_min;
+
+  if (abs(v_cmd) >= abs(v_curr)) {
```
AlexeyMerzlyakov, line 209: What will be if `v_curr = 1.0` while `v_cmd = -1.2`? I think the comparison should also count the sign of current and required velocities. In the current example, we have the deceleration and then acceleration as depicted at the graph below:

nav2_velocity_smoother/src/velocity_smoother.cpp
AlexeyMerzlyakov, line 1: Also, could you also please add the testcase(s) covering this change (acceleration/deceleration, linear movement, rotation at the place, positive/negative speeds, speed passing through `0.0`, etc...)?

doisyg: Yes, this is indeed quite needed. I will take care of it a bit later

nav2_velocity_smoother/src/velocity_smoother.cpp
```diff
@@ -202,8 +202,26 @@ double VelocitySmoother::applyConstraints(
...
+      v_component_max = accel / smoothing_frequency_;
+      v_component_min = -accel / smoothing_frequency_;
+    } else {
+      v_component_max = -decel / smoothing_frequency_;
+      v_component_min = decel / smoothing_frequency_;
+    }
+  } else {
+    v_component_max = -decel / smoothing_frequency_;
+    v_component_min = decel / smoothing_frequency_;
+  }
```
AlexeyMerzlyakov, line 223: I think, it could be fold into:
```
if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
  do acceleration
} else {
  do deceleration
}
```

AlexeyMerzlyakov:
BTW, just a question: shouldn't the same rules (velocities are more/less than `0.0`, `v_cmd` and `v_curr` are the same sign or not) apply to `v_component_max` and `v_component_min` calculation in `VelocitySmoother::findEtaConstraint()`, which contains similar to `applyConstraints` logic (that should also depend on which sign the speeds are)?

nav2_velocity_smoother/src/velocity_smoother.cpp
doisyg: Added a couple of applyConstraints tests

AlexeyMerzlyakov:
Update on `VelocitySmoother::findEtaConstraint()` routine. Consider the following example partially repeating the `smootherTimer()` [logic](https://github.com/ros-planning/navigation2/blob/ff4fd54076c64f6782b821c60319686413f2551b/nav2_velocity_smoother/src/velocity_smoother.cpp#L248-L276) (checked on latest patch from this PR):
```
  double eta_positive_accel = smoother->findEtaConstraint(1.0, 2.0, 1.5, -2.0);
  printf("Positive accel vel = %f (eta=%f)\n",
    smoother->applyConstraints(1.0, 2.0, 1.5, -2.0, eta_positive_accel), eta_positive_accel);

  double eta_positive_decel = smoother->findEtaConstraint(1.0, 0.0, 1.5, -2.0);
  printf("Positive decel vel = %f (eta=%f)\n",
    smoother->applyConstraints(1.0, 0.0, 1.5, -2.0, eta_positive_decel), eta_positive_decel);

  double eta_negative_accel = smoother->findEtaConstraint(-1.0, -2.0, 1.5, -2.0);
  printf("Negative accel vel = %f (eta=%f)\n",
    smoother->applyConstraints(-1.0, -2.0, 1.5, -2.0, eta_negative_accel), eta_negative_accel);

  double eta_negative_decel = smoother->findEtaConstraint(-1.0, 0.0, 1.5, -2.0);
  printf("Negative decel vel = %f (eta=%f)\n",
    smoother->applyConstraints(-1.0, 0.0, 1.5, -2.0, eta_negative_decel), eta_negative_decel);
```

The output is:
```
Positive accel vel = 1.075000 (eta=0.075000)
Positive decel vel = 0.900000 (eta=0.100000)
Negative accel vel = -1.075000 (eta=0.100000)
Negative decel vel = -0.925000 (eta=0.075000) <- incorrect
```

`eta_negative_accel` and `eta_negative_devel` will be calculated incorrectly in this case giving `0.075` for deceleration and `0.1` for acceleration case values. The velocity when accelerating in negative case will be correct and equal to the `-1.075000`, due to `clamp` barrier in `applyConstraints()`. However, for deceleration negative case, the velocity change will smaller, that in positive case (`clamp` from `applyConstraints()` in this case won't work), which means the produced by smoother accelerations will be asymmetrical.

This most probably will work if apply the same of  `applyConstraints()` logic, to the `findEtaConstraint()`, e.g.:
```
@@ -182,10 +182,22 @@ double VelocitySmoother::findEtaConstraint(
   const double v_curr, const double v_cmd, const double accel, const double decel)
 {
   // Exploiting vector scaling properties
-  const double v_component_max = accel / smoothing_frequency_;
-  const double v_component_min = decel / smoothing_frequency_;
   const double dv = v_cmd - v_curr;
 
+  double v_component_max;
+  double v_component_min;
+
+  // Accelerating if magnitude of v_cmd is above magnitude of v_curr
+  // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
+  // Deccelerating otherwise
+  if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
+    v_component_max = accel / smoothing_frequency_;
+    v_component_min = -accel / smoothing_frequency_;
+  } else {
+    v_component_max = -decel / smoothing_frequency_;
+    v_component_min = decel / smoothing_frequency_;
+  }
+
```

nav2_velocity_smoother/test/test_velocity_smoother.cpp
```diff
@@ -217,6 +217,301 @@ TEST(VelocitySmootherTest, testapplyConstraints)
...
+  for (size_t i = 1; i < steps_to_target + 1; i++) {
+    v_curr = smoother->applyConstraints(v_curr, target_vel, accel, deccel, no_eta);
+    EXPECT_NEAR(v_curr, init_vel + i * dv_accel, 0.001);
+  }
+  EXPECT_NEAR(v_curr, target_vel, 0.001);
+  v_curr = smoother->applyConstraints(v_curr, target_vel, accel, deccel, no_eta);
+  EXPECT_NEAR(v_curr, target_vel, 0.001);
+}
+
+TEST(VelocitySmootherTest, testapplyConstraintsZeroToPositiveAccel)
```
AlexeyMerzlyakov, line 248: Is there any difference between this and `testapplyConstraintsPositiveToPositiveAccel` test? It seems that in both cases we are in OY+ positive velocity axis, and thus the behavior is expected to be the same?

nav2_velocity_smoother/test/test_velocity_smoother.cpp
```diff
@@ -217,6 +217,301 @@ TEST(VelocitySmootherTest, testapplyConstraints)
...
+  for (size_t i = 1; i < steps_to_target + 1; i++) {
+    v_curr = smoother->applyConstraints(v_curr, target_vel, accel, deccel, no_eta);
+    EXPECT_NEAR(v_curr, init_vel - i * dv_accel, 0.001);
+  }
+  EXPECT_NEAR(v_curr, target_vel, 0.001);
+  v_curr = smoother->applyConstraints(v_curr, target_vel, accel, deccel, no_eta);
+  EXPECT_NEAR(v_curr, target_vel, 0.001);
+}
+
+TEST(VelocitySmootherTest, testapplyConstraintsZeroToNegativeAccel)
```
AlexeyMerzlyakov, line 338: The same question for `testapplyConstraintsNegativeToNegativeAccel` vs `testapplyConstraintsZeroToNegativeAccel`; `testapplyConstraintsPositiveToPositiveDeccel` vs `testapplyConstraintsPositiveToZeroDeccel` and `testapplyConstraintsNegativeToNegativeDeccel` vs `testapplyConstraintsNegativeToZeroDeccel`

nav2_velocity_smoother/test/test_velocity_smoother.cpp
```diff
@@ -217,6 +217,301 @@ TEST(VelocitySmootherTest, testapplyConstraints)
...
+    } else {
+      EXPECT_NEAR(v_curr, init_vel - i * dv_deccel, 0.001);
+    }
+  }
+  EXPECT_NEAR(v_curr, target_vel, 0.001);
+  v_curr = smoother->applyConstraints(v_curr, target_vel, accel, deccel, no_eta);
+  EXPECT_NEAR(v_curr, target_vel, 0.001);
+}
+
+TEST(VelocitySmootherTest, testapplyConstraintsPositiveToPositiveDeccel)
```
AlexeyMerzlyakov, line 400: Deccel -> Decel here and in all other places

doisyg:
Makes sense. Just pushed the same logic to `findEtaConstraint`. But give me time to write proper tests for this before merging

AlexeyMerzlyakov (commented):
Made a review on testcases added.

As for main code logic, it appeared, we need to have to apply similar change to the `eta` calculation routine, as proposed in https://github.com/ros-planning/navigation2/pull/3529#issuecomment-1508341528. Sorry for missing this case this in first review. Also, please support the testcases for different `eta` calculation cases (positive/negative velocities; accel-decel) as well.

nav2_velocity_smoother/src/velocity_smoother.cpp
```diff
@@ -182,9 +182,21 @@ double VelocitySmoother::findEtaConstraint(
...
-  const double dv = v_cmd - v_curr;
+  double dv = v_cmd - v_curr;
+
+  double v_component_max;
+  double v_component_min;
+
+  // Accelerating if magnitude of v_cmd is above magnitude of v_curr
+  // and if v_cmd and v_curr have the same sign (i.e. speed is NOT passing through 0.0)
+  // Deccelerating otherwise
+  if ((abs(v_cmd) >= abs(v_curr)) && (v_curr * v_cmd >= 0.0)) {
```
SteveMacenski, line 193:
```suggestion
  if (abs(v_cmd) >= abs(v_curr) && v_curr * v_cmd >= 0.0) {
```

Extra separators are distracting

nav2_velocity_smoother/src/velocity_smoother.cpp
```diff
@@ -182,9 +182,21 @@ double VelocitySmoother::findEtaConstraint(
   const double v_curr, const double v_cmd, const double accel, const double decel)
 {
   // Exploiting vector scaling properties
-  const double v_component_max = accel / smoothing_frequency_;
-  const double v_component_min = decel / smoothing_frequency_;
-  const double dv = v_cmd - v_curr;
+  double dv = v_cmd - v_curr;
+
```
SteveMacenski, line 186: remove space

SteveMacenski (changes requested):

doisyg:
Added findEtaConstraint tests

nav2_velocity_smoother/test/test_velocity_smoother.cpp
doisyg: These should be equivalent yes, but you never know how a change in the code logic could affect the behavior from and to zero. And the test is cheap

nav2_velocity_smoother/test/test_velocity_smoother.cpp
doisyg: Similar to above. And I think the tests to zero are actually more important than from zero

nav2_velocity_smoother/test/test_velocity_smoother.cpp
doisyg: ?

nav2_velocity_smoother/test/test_velocity_smoother.cpp
AlexeyMerzlyakov: Okay, this makes sense. Let's leave it for additional check.

nav2_velocity_smoother/test/test_velocity_smoother.cpp
AlexeyMerzlyakov: I mean the `Deccel` keyword typo: to fix it to `Decel`

AlexeyMerzlyakov (commented):
A few small items and I am OK for the merge

nav2_velocity_smoother/test/test_velocity_smoother.cpp
```diff
@@ -179,17 +179,56 @@ TEST(VelocitySmootherTest, testfindEtaConstraint)
...
-  EXPECT_EQ(smoother->findEtaConstraint(1.0, 2.0, 1.5, -2.0), 0.075);
-  // Too low
-  EXPECT_EQ(smoother->findEtaConstraint(1.0, 0.0, 1.5, -2.0), 0.1);
-
-  // In a more realistic situation accelerating linear axis
-  EXPECT_NEAR(smoother->findEtaConstraint(0.40, 0.50, 1.5, -2.0), 0.75, 0.001);
+  // Constant positive
+  EXPECT_EQ(smoother->findEtaConstraint(1.0, 1.0, accel, deccel), -1);
+  // Constant negative
+  EXPECT_EQ(smoother->findEtaConstraint(1.0, 1.0, accel, deccel), -1);
```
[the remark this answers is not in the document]
doisyg: Good catch!

nav2_velocity_smoother/test/test_velocity_smoother.cpp
doisyg: Thanks! Pardon my french

SteveMacenski:
The build failure is a broader issue with Rolling's update totally crippling Nav2 (I'd recommend _not_ updating). If @AlexeyMerzlyakov is happy with the test updates and it works locally (without updating!) then I'm fine merging this.

AlexeyMerzlyakov (approved):
@doisyg, thank you for the consistent work on this issue!
The patch LGTM (one small change is required yet).

Local verification was performed. The results are as follows -
* Local colcon testing for `nav2_velocity_smoother` package: PASS
* Codecov: `98.9 %` of `nav2_velocity_smoother/src/velocity_smoother.cpp` lines covered (all new changes were covered by proposed testcases).
* Local robot rotation-in-place verification: The picture seems to be symmetrical for clock- and counterclock-wise rotations:
exp_01
exp_02
exp_03

PR #3536, opened 2023-04-21
Nav2 route bfs

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

This pr adds a breadth first search algorithm to the nav2_route. It returns the closest goal in a list of goals. 

---

#### For Maintainers: 


nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+void BreadthFirstSearch::setCostmap(nav2_costmap_2d::Costmap2D *costmap)
+{
+  std::cout << "Set costmap " << std::endl;
+  costmap_ = costmap;
+
+  unsigned int x_size = costmap_->getSizeInCellsX();
+  unsigned int y_size = costmap_->getSizeInCellsY();
+
+  initMotionModel(static_cast<int>(x_size), static_cast<int>(y_size));
```
SteveMacenski, line 43: I'd just roll the init motion model stuff instead of set costmap since we're only doing 2 things that don't change

nav2_route/include/nav2_route/breadth_first_search.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+  bool search(Coordinates & closest_goal);
+
+ private:
+
+  inline Coordinates getCoords(
+      const unsigned int & index) const
+  {
+    const unsigned int & width = x_size_;
+    return {static_cast<float>(index % width), static_cast<float>(index / width)};
```
SteveMacenski, line 66: Any reason we're not asking the costmap to translate this for us?

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::cout << "Queue Size: " << queue_.size() << std::endl;
+
+    unsigned int current = queue_.front();
+    std::cout << "grab front" << std::endl;
+
+    Coordinates current_coord = getCoords(current);
+    std::cout << "Current Node: " << current_coord.x << " " << current_coord.y << std::endl;
+
+    queue_.pop();
+    states_[current].visited = true;
```
SteveMacenski, line 86: Woof, these are each individually lookups. These are pricy. Definitely only do this lookup once `auto & curr_state = states_[current];` and use that reference so you don't double that overhead. 

Ideally, we should try to make this something that doesn't involve a lookup. If you store in the `queue_` a class containing a pointer to the node instead of the index, then you can insert an expanded node into the graph that is an unordered map, but then never have to query for it again.

If you look at the smac planner, that's basically what I do. I have a holding class containing the node pointer. So when we get the node from the queue, we have the memory to ask `node->isQueued()` without having to ever look up the data structure which contains the graph. While unordered maps are O(1), constant time doesn't mean _fast_ time when compared to direct memory access

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,170 @@
...
+  states_.insert(std::make_pair(start_, State()));
+}
+
+void BreadthFirstSearch::setGoals(std::vector<unsigned int> mxs, std::vector<unsigned int> mys)
+{
+  goals_.clear();
+  for(unsigned int i = 0; i < (mxs.size() && mys.size()); ++i) {
+    unsigned int index = getIndex(mxs[i], mys[i], x_size_);
+    goals_.push_back(index);
+    states_.insert(std::make_pair(index, State()));
```
SteveMacenski, line 57: You end up clearing states first thing in `search`, I think you can remove that here?

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,170 @@
...
+void BreadthFirstSearch::setStart(unsigned int mx, unsigned int my) {
+  start_ = getIndex(mx, my, x_size_);
+  states_.insert(std::make_pair(start_, State()));
+}
+
+void BreadthFirstSearch::setGoals(std::vector<unsigned int> mxs, std::vector<unsigned int> mys)
+{
+  goals_.clear();
+  for(unsigned int i = 0; i < (mxs.size() && mys.size()); ++i) {
+    unsigned int index = getIndex(mxs[i], mys[i], x_size_);
```
SteveMacenski, line 55: Why not use costmaps' version?

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+bool BreadthFirstSearch::search(Coordinates & closest_goal)
+{
+  std::cout << "Start " << std::endl;
+  states_.clear();
+
+  // clear the queue
+  std::queue<unsigned int> empty_queue;
+  queue_.swap(empty_queue);
```
SteveMacenski, line 68: `queue_` isn't used outside of this function, you could just create it here locally as a function's member and it would just reset itself when the function goes out of scope. Given this search should be small, even if the queue  doesn't save its allocated memory block, it doesn't matter much.

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+    std::cout << "Checked goals" << std::endl;
+
+    std::vector<unsigned int> neighbors;
+    getNeighbors(current, neighbors);
+    std::cout << "Got neigbors" << std::endl;
+
+    for(const auto neighbor : neighbors)
+    {
+      if(!states_[neighbor].queued && !states_[neighbor].visited) {
```
SteveMacenski, line 105: Same here, don't lookup more than once!

nav2_route/CMakeLists.txt
```diff
@@ -111,6 +111,14 @@ ament_target_dependencies(graph_file_loaders
 
 pluginlib_export_plugin_description_file(nav2_route plugins.xml)
 
+# Breadth first search library
+add_library(breadth_first_search SHARED
```
SteveMacenski, line 115: Reason this is a separate library? I'm actually not sure what impact this has on compute time - I'd think potentially faster. I tend to just stick everything into a single library that doesn't need to be explicitly separated like plugins. That might not be good, but I haven't experimented with it.

nav2_route/include/nav2_route/breadth_first_search.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+  std::unordered_map<unsigned int, Node> graph_;
+
+  NodePtr start_;
+  NodeVector goals_;
+
+  unsigned int x_size_;
+  unsigned int y_size_;
+  unsigned int max_index_;
+  std::vector<int> neighbors_grid_offsets;
```
SteveMacenski, line 82:
```suggestion
  std::vector<int> neighbors_grid_offsets_;
```

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
...
+  goals_.clear();
+  for (unsigned int i = 0; i < (mxs.size() && mys.size()); ++i) {
+    goals_.emplace_back(addToGraph(costmap_->getIndex(mxs[i], mys[i])));
+  }
+}
+
+bool BreadthFirstSearch::search(Coordinates & closest_goal)
+{
+  // clear the graph
+  std::unordered_map<unsigned int, Node> empty_graph;
```
SteveMacenski, line 70: `graph_.clear();` is totally fine. We didn't do that for the queue because the queue has no such functionality for some bizarre reason

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
...
+  std::cout << "Goal index " << goals_[0]->index << std::endl;
+
+  int iter = 0;
+  while (!queue.empty()) {
+    iter += 1;
+    std::cout << "Iteration " << iter << std::endl;
+    std::cout << "Queue Size: " << queue.size() << std::endl;
+
+    auto & current = queue.front();
+
```
SteveMacenski, line 88: pop at this point and also check if `current->visited`, if so, `continue` the while loop since we know its already visited

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
...
+    std::cout << "grab front" << std::endl;
+    std::cout << std::endl;
+
+    Coordinates current_coord = getCoords(current->index);
```
SteveMacenski, line 92: `current_coord` is never used

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
...
+    }
+    std::cout << "Checked goals" << std::endl;
+
+    NodeVector neighbors;
+    getNeighbors(current->index, neighbors);
+    std::cout << "Got neigbors" << std::endl;
+
+    for (const auto neighbor : neighbors) {
+      if (!neighbor->visited) {
+        neighbor->visited = true;
```
SteveMacenski, line 115: This is not true, this is not visited. This is queued. Visited is when it appears the top of the queue for being the parent of search.

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
...
+    c_coord = getCoords(index);
+
+    // Check for wrap around conditions
+    if (std::fabs(p_coord.x - c_coord.x) > 1 || std::fabs(p_coord.y - c_coord.y) > 1) {
+      std::cout << "Index wraps" << std::endl;
+      continue;
+    }
+
+    // Check for out of bounds
+    if (index >= max_index_) {
```
SteveMacenski, line 146: Or < 0

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    queue.pop();
+
+    // Check goals
+    for (const auto & goal : goals_) {
+      std::cout << "Goal index " << goal->index << std::endl;
+      std::cout << "current index " << current->index << std::endl;
+      if (current->index == goal->index) {
+        std::cout << "Found goal" << std::endl;
+        closest_goal = getCoords(current->index);
```
SteveMacenski, line 103: Perhaps a more natural API would be to return the index of the goal vector who's goal pose is closest. I'm not 100% sure about that, but something to check into on the actual inputs and outputs that would be most natural in the context of the code this is going to integrate into.

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) 2023 Joshua Wallace
```
SteveMacenski, line 1: You can leave your copyright here for the adaptation, but put the original (C) line for Smac above it since this is derivative.

SteveMacenski (commented):
Close!

nav2_route/src/breadth_first_search.cpp
jwallace42: yeah, it will get removed with the printout :)

nav2_route/src/breadth_first_search.cpp
jwallace42: it is a unsigned int. If we are below zero C++ has failed us :)

nav2_route/src/breadth_first_search.cpp
jwallace42: yeah, 

I think once the search looks good I am going to fix up the api. I am still not sure if we want to check for multiple goals at the same time or just check one.

nav2_route/CMakeLists.txt
jwallace42: I have no idea how it affects comp time. I had a seperate lib for the test. I didn't think there was any need to link against other files that I don't need for the test.

nav2_route/CMakeLists.txt
```diff
@@ -138,5 +142,9 @@ endif()
 
 ament_export_include_directories(include)
 ament_export_dependencies(${dependencies})
-ament_export_libraries(${library_name} edge_scorers route_operations graph_file_loaders)
+ament_export_libraries(${library_name}
+    edge_scorers
+    route_operations
+    graph_file_loaders
+    breadth_first_search)
```
SteveMacenski, line 149: Missed one :-)

nav2_route/include/nav2_route/breadth_first_search.hpp
```diff
@@ -0,0 +1,114 @@
...
+   * @param my The array y map coordinate of the goals
+   */ 
+  void setGoals(std::vector<unsigned int> mxs, std::vector<unsigned int> mys);
+
+  /**
+   * @brief Find the closest goal to the start given a costmap, start and goal
+   * @param closest_goal The coordinates of the closest goal
+   * @return True if the search was successful
+   */ 
+  bool search(unsigned int & closest_goal_index);
```
SteveMacenski, line 79: I forget off the top of my head what the algorithmic failure modes I use across this project. If I do the whole "return true/false to handle failures" thing, then this is perfect no problems.

If I do more of a "I assume you succeed to return the results in the return value and throw exceptions otherwise" thing, then we should do that here too. I imagine in the use of this in the system, if `false` is returned, we would throw a failure exception. But I forget how low level I introduce exceptions in this project vs keeping them at medium-level software to keep the pure algorithms exception free.

nav2_route/include/nav2_route/breadth_first_search.hpp
```diff
@@ -0,0 +1,114 @@
...
+   */ 
+  void getNeighbors(unsigned int parent_index, NodeVector & neighbors);
+
+  /**
+   * @brief Checks if the index is in collision
+   * @param index The index to check
+   */ 
+  bool inCollision(unsigned int index);
+
+  std::unordered_map<unsigned int, SimpleNode> graph_;
```
SteveMacenski, line 101: If you look at Smac, I updated this to use a specailized unordered map implementation. Lets adapt that here too. Its API compatible, there's just a header you need to include in this project + use that namespace for the instantiation

SteveMacenski: It has a very fast insertion rate

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <iostream>
+#include <queue>
+#include <unordered_map>
+
+namespace nav2_route
+{
+
+void BreadthFirstSearch::setCostmap(nav2_costmap_2d::Costmap2D * costmap)
+{
+  std::cout << "Set costmap " << std::endl;
```
SteveMacenski, line 27: Still a print

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,150 @@
...
+  costmap_ = costmap;
+
+  int x_size = static_cast<int>(costmap_->getSizeInCellsX());
+  int y_size = static_cast<int>(costmap_->getSizeInCellsY());
+
+  max_index_ = x_size * y_size;
+  x_size_ = x_size;
+
+  neighbors_grid_offsets_ = {-1, +1,
```
SteveMacenski, line 36: I'd do some testing and make sure that 8 connected has a benefit over 4 connected. I assume for a BFS it does, but I wouldn't assume that without testing

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+void BreadthFirstSearch::setStart(unsigned int mx, unsigned int my)
+{
+  start_ = addToGraph(costmap_->getIndex(mx, my));
+}
+
+void BreadthFirstSearch::setGoals(std::vector<unsigned int> mxs, std::vector<unsigned int> mys)
+{
+  goals_.clear();
+  for (unsigned int i = 0; i < (mxs.size() && mys.size()); ++i) {
```
SteveMacenski, line 60: I'd throw an exception if the sizes are not equal and only just check 1's size. This might mask a real problem of a missing X or Y coordinate

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,150 @@
...
+void BreadthFirstSearch::getNeighbors(unsigned int parent_index, NodeVector & neighbors)
+{
+  unsigned int p_mx, p_my; 
+  costmap_->indexToCells(parent_index, p_mx, p_my);
+
+  unsigned int neigbhor_index;
+  for (const int & neighbors_grid_offset : neighbors_grid_offsets_) {
+    // Check if index is negative
+    int index = parent_index + neighbors_grid_offset;
+    if(index < 0){
```
SteveMacenski, line 109: linting problem here

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,150 @@
...
+      continue;
+    }
+
+    neigbhor_index = static_cast<unsigned int>(index);
```
SteveMacenski, line 113: misspelled neighbor

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // Check for out of bounds
+    if (neigbhor_index >= max_index_) {
+      continue;
+    }
+
+    if (inCollision(neigbhor_index)) {
+      continue;
+    }
+
+    auto neighbor = addToGraph(neigbhor_index);
```
SteveMacenski, line 133: These 2 lines can be put into 1 line

SteveMacenski (changes requested):
You can tell when its close to being done with something had looks easy and obvious. Nice work!

nav2_route/src/breadth_first_search.cpp
SteveMacenski: ex https://medium.com/analytics-vidhya/part-i-breadth-first-search-using-grid-dc41a5f41663 you can see that using a 4 connected neighborhood that the expansion waves aren't equal, so one might be explored first that's axially aligned with the X/Y axis which is actually up to sqrt(2) further away from one that's closer on a diagonal.

jwallace42: Do you have an tips for the visualization? Publish out a occgrid?

nav2_route/include/nav2_route/breadth_first_search.hpp
jwallace42: You are using the same flow as the rest of the servers. (try catch list) 

All the errors that I could throw are the same as the planners. Would using those be alright?

nav2_route/src/breadth_first_search.cpp
SteveMacenski: Depends on what strikes your fancy, but probably an occ grid. Maybe take a snapshot every so often of the number of expanded cells and color them black to see how they expand over time. Or, do them all at once and visualize the full grid color-coded by visitation time so you can see the pattern emerging.

nav2_route/include/nav2_route/breadth_first_search.hpp
SteveMacenski: Check where they're caught, but I suppose that would be fine unless that would get caught somewhere separate. Just make sure the string of the exception makes clear its from here and not there.

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -53,6 +59,15 @@ void GoalIntentExtractor::configure(
   nav2_util::declare_parameter_if_not_declared(
     node, "min_dist_from_start", rclcpp::ParameterValue(0.10));
   min_dist_from_start_ = static_cast<float>(node->get_parameter("min_dist_from_start").as_double());
+
+  std::string global_costmap_topic;
+  nav2_util::declare_parameter_if_not_declared(
+    node, "global_costmap_topic", rclcpp::ParameterValue(std::string("global_costmap/costmap_raw")));
+  node->get_parameter("global_costmap_topic", global_costmap_topic);
+
+  costmap_sub_ = std::make_unique<nav2_costmap_2d::CostmapSubscriber>(node, global_costmap_topic);
```
SteveMacenski, line 68: Where is the nearest costmap subscriber, can we not just pull one in from somewhere else (or are the rest in plugins?) 

If there's no centrally located one, can we make a centrally located one that can be pulled in here and also used in the plugins API?

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -122,6 +137,10 @@ GoalIntentExtractor::findStartandGoal(const std::shared_ptr<const GoalT> goal)
             "Could not determine node closest to start or goal pose requested!");
   }
 
+  if (!isNodeValid(start_route, start_))
```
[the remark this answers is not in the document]
SteveMacenski: We need to adjust the node spatial tree to return `N` results and make `N` a parameter. Get the parameter here and give it to `node_spatial_tree_` in the constructor (I don't want to make it ROS-y)

SteveMacenski (changes requested):

nav2_route/package.xml
```diff
@@ -22,6 +22,7 @@
   <depend>nav2_core</depend>
   <depend>lib-nanoflann</depend>
   <depend>nlohmann-json</depend>
+  <depend>nav2_smac_planner</depend>
```
SteveMacenski, line 25: Why? I don't see any includes

nav2_route/include/nav2_route/goal_intent_extractor.hpp
```diff
@@ -15,10 +15,14 @@
 #ifndef NAV2_ROUTE__GOAL_INTENT_EXTRACTOR_HPP_
 #define NAV2_ROUTE__GOAL_INTENT_EXTRACTOR_HPP_
 
+#include <algorithm>
+#include <geometry_msgs/msg/detail/pose_stamped__struct.hpp>
```
SteveMacenski, line 19: Why not just pose stamped?

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,180 @@
...
+  queue.push(start_);
+
+  while (!queue.empty()) {
+    auto & current = queue.front();
+    queue.pop();
+
+    std::cout << "Current index: " << current->index << std::endl;
+
+    // Check goals
+    for (unsigned int index = 0; index < goals_.size(); ++index) {
```
SteveMacenski, line 83: Use another variable that isn't index, that renders in some code editors as a keyword

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -12,11 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <boost/concept/detail/has_constraints.hpp>
```
SteveMacenski, line 15: Woah, whaddup with boost here?

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -122,6 +138,10 @@ GoalIntentExtractor::findStartandGoal(const std::shared_ptr<const GoalT> goal)
             "Could not determine node closest to start or goal pose requested!");
   }
 
+  if (!isNodeValid(start_route, start_))
+  {
+    RCLCPP_INFO(logger_, "Invalid Starting Route index");
```
SteveMacenski, line 143: Should this not be throwing an exception of some kind of its truly an error for the action to reject a request? More context on none of the nearest starting points are reachable would be good for the message instead.

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -53,6 +60,15 @@ void GoalIntentExtractor::configure(
...
     node, "min_dist_from_start", rclcpp::ParameterValue(0.10));
   min_dist_from_start_ = static_cast<float>(node->get_parameter("min_dist_from_start").as_double());
+
+  std::string global_costmap_topic;
+  nav2_util::declare_parameter_if_not_declared(
+    node, "global_costmap_topic", rclcpp::ParameterValue(std::string("global_costmap/costmap_raw")));
+  node->get_parameter("global_costmap_topic", global_costmap_topic);
+
+  costmap_sub_ = std::make_unique<nav2_costmap_2d::CostmapSubscriber>(node, global_costmap_topic);
+  bfs_ = std::make_unique<BreadthFirstSearch>();
```
SteveMacenski, line 70: BFS costmap checking should be a parameterized feature. It might not be viable for many real situations.

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +213,36 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
 
+bool GoalIntentExtractor::isNodeValid(unsigned int node_index, 
+                                      const geometry_msgs::msg::PoseStamped & pose)
+{
+  unsigned int s_mx, s_my, g_mx, g_my; 
+  costmap_sub_->getCostmap()->worldToMap(pose.pose.position.x, pose.pose.position.y, 
+                                         s_mx, s_my);
+
+  // double check the frames, probably need to move into the map frame...
+  float goal_x = (*graph_)[node_index].coords.x;
```
SteveMacenski, line 224: Don't look up multiple times, also use the api `graph_->at(idx)` which goes around the length checking to increase speed

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +213,36 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+
+  bfs_->setStart(s_mx, s_my);
+  bfs_->setGoals(goals);
+  if (bfs_->isNodeVisible()) {
+    bfs_->clearGraph();
+    return true;
+  }
+
+  unsigned int goal;
+  if (bfs_->search(goal)) {
```
SteveMacenski, line 238: You never use this output of search

SteveMacenski (changes requested):
You appear to have some IO detail problems where things aren't being set or outputs actually used and given to the requester to complete the task usefully

nav2_route/src/goal_intent_extractor.cpp
SteveMacenski: Or set to the goal -- nor in the case it has visibility either

nav2_route/src/route_server.cpp
```diff
@@ -100,6 +101,7 @@ nav2_util::CallbackReturn
 RouteServer::on_activate(const rclcpp_lifecycle::State & /*state*/)
 {
   RCLCPP_INFO(get_logger(), "Activating");
+  goal_intent_extractor_->activate();
```
SteveMacenski, line 104: Deactivate, reset on cleanup, etc?

nav2_route/src/route_server.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License. Reserved.
 
 #include "nav2_route/route_server.hpp"
+#include "nav2_core/planner_exceptions.hpp"
```
SteveMacenski, line 16: Why planner exceptions?

nav2_route/src/graph_loader.cpp
```diff
@@ -32,7 +32,9 @@ GraphLoader::GraphLoader(
   route_frame_ = frame;
 
   nav2_util::declare_parameter_if_not_declared(
-    node, "graph_filepath", rclcpp::ParameterValue(std::string("")));
+    node, "graph_filepath", rclcpp::ParameterValue(
+      ament_index_cpp::get_package_share_directory("nav2_route") +
```
SteveMacenski, line 36: Remove for production? I think this can be empty due to having a service to change?

nav2_route/src/node_spatial_tree.cpp
```diff
@@ -36,7 +41,7 @@ NodeSpatialTree::~NodeSpatialTree()
 
 void NodeSpatialTree::computeTree(Graph & graph)
 {
-  if (kdtree_) {
+  if (kdtree_ != nullptr) {
```
SteveMacenski, line 44: That should be the same, `if(pointer)` checks for nullptr explicitly

nav2_route/src/node_spatial_tree.cpp
```diff
@@ -65,7 +70,10 @@ bool NodeSpatialTree::findNearestGraphNodeToPose(
     return false;
   }
 
-  node_id = ret_index[0];
+  for (int i = 0; i < num_of_nearest_nodes_ && i < static_cast<int>(ret_index.size()); ++i) {
+    node_ids.push_back(ret_index[i]);
```
SteveMacenski, line 74: You mentioned a potential ordering issue - is it possible that these aren't promised to be in order?

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -16,7 +16,14 @@
 #include <memory>
 #include <vector>
 
+#include "nav2_core/planner_exceptions.hpp"
```
SteveMacenski, line 19: Planner exceptions?

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -53,6 +58,49 @@ void GoalIntentExtractor::configure(
...
+
+  if (enable_search_) {
+    std::string global_costmap_topic;
+    nav2_util::declare_parameter_if_not_declared(
+      node, "global_costmap_topic",
+      rclcpp::ParameterValue(std::string("global_costmap/costmap_raw")));
+    node->get_parameter("global_costmap_topic", global_costmap_topic);
+
+    nav2_util::declare_parameter_if_not_declared(
+      node, "max_iterations", rclcpp::ParameterValue(500));
```
SteveMacenski, line 81: Set the default to much higher, try 10,000

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -53,6 +58,49 @@ void GoalIntentExtractor::configure(
...
+  node_spatial_tree_->computeTree(graph);
+
+  nav2_util::declare_parameter_if_not_declared(
+    node, "enable_search", rclcpp::ParameterValue(false));
+  enable_search_ = node->get_parameter("enable_search").as_bool();
+
+  if (enable_search_) {
+    std::string global_costmap_topic;
+    nav2_util::declare_parameter_if_not_declared(
+      node, "global_costmap_topic",
```
SteveMacenski, line 76: Does this strictly need to be the global costmap?

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -53,6 +58,49 @@ void GoalIntentExtractor::configure(
...
+      rclcpp::ParameterValue(std::string("global_costmap/costmap_raw")));
+    node->get_parameter("global_costmap_topic", global_costmap_topic);
+
+    nav2_util::declare_parameter_if_not_declared(
+      node, "max_iterations", rclcpp::ParameterValue(500));
+    max_iterations_ = node->get_parameter("max_iterations").as_int();
+
+    nav2_util::declare_parameter_if_not_declared(
+      node, "enable_search_viz", rclcpp::ParameterValue(false));
+    enable_search_viz_ = node->get_parameter("enable_search_viz").as_bool();
```
SteveMacenski, line 86: I think both this param and the visualization it enables can be removed in the final version. Useful for debugging but shouldn't be necessary afterwards

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -53,6 +58,49 @@ void GoalIntentExtractor::configure(
...
+      1);
+    goal_expansion_viz_ =
+      node->create_publisher<nav_msgs::msg::OccupancyGrid>("goal_expansions", 1);
+  }
+}
+
+void GoalIntentExtractor::activate()
+{
+  if (enable_search_) {
+    bfs_->initialize(costmap_sub_->getCostmap().get(), max_iterations_);
```
SteveMacenski, line 102: getCostmap() returns a shared pointer, that's means for exactly this purpose silly :-)

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +253,90 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+}
+
+void GoalIntentExtractor::visualizeExpansions(
+  rclcpp::Publisher<nav_msgs::msg::OccupancyGrid>::SharedPtr occ_grid_pub)
+{
+  if (!enable_search_viz_) {
+    return;
+  }
+  nav_msgs::msg::OccupancyGrid occ_grid_msg;
+  unsigned int width = costmap_sub_.get()->getCostmap()->getSizeInCellsX();
```
SteveMacenski, line 321: Why are you `.get()` this? you can just `->` its a shared pointer, you don't need its underlying memory, that's what the accessor does

nav2_route/src/goal_intent_extractor.cpp
SteveMacenski: I think from a quick look it doesn't necessarily need to be

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +253,90 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+    throw nav2_core::StartOutsideMapBounds(
+            "Start Coordinates of(" + std::to_string(pose.pose.position.x) + ", " +
+            std::to_string(pose.pose.position.y) + ") was outside bounds");
+  }
+
+  if (costmap_sub_->getCostmap()->getCost(s_mx, s_my) == nav2_costmap_2d::LETHAL_OBSTACLE) {
+    throw nav2_core::StartOccupied("Start was in lethal cost");
+  }
+
+  //TODO(jwallace42): double check the frames, probably need to move into the map frame...
```
SteveMacenski, line 274: TODO? I think this is all correct, no?

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +253,90 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+  std::vector<unsigned int> valid_node_indices;
+  std::vector<nav2_costmap_2d::MapLocation> goals;
+  for (const auto & node_index : node_indices) {
+    float goal_x = (*graph_)[node_index].coords.x;
```
SteveMacenski, line 278: graph_->at(node_index)

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +253,90 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+      RCLCPP_WARN_STREAM(
+        logger_, "Goal coordinate of(" + std::to_string(goal_x) + ", " +
+        std::to_string(goal_y) + ") was outside bounds. Removing from goal list");
+      continue;
+    }
+
+    if (costmap_sub_->getCostmap()->getCost(g_mx, g_my) == nav2_costmap_2d::LETHAL_OBSTACLE) {
+      RCLCPP_WARN_STREAM(
+        logger_, "Goal corrdinate of(" + std::to_string(goal_x) + ", " +
+        std::to_string(goal_y) + ") was in lethal cost. Removing from goal list.");continue;
```
SteveMacenski, line 290: continue on new line

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +253,90 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+  }
+
+  if (goals.empty()) {
+    throw nav2_core::PlannerException("All goals were invalid");
+  }
+
+  bfs_->clearGraph();
+  bfs_->setStart(s_mx, s_my);
+  bfs_->setGoals(goals);
+  if (bfs_->isNodeVisible()) {
```
SteveMacenski, line 303: isFirstGoalVisible

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+bool BreadthFirstSearch::isNodeVisible()
+{
+  unsigned int s_mx, s_my, g_mx, g_my;
+  costmap_->indexToCells(start_->index, s_mx, s_my);
+  costmap_->indexToCells(goals_.front()->index, g_mx, g_my);
+
+  for (nav2_util::LineIterator line(s_mx, s_my, g_mx, g_my); line.isValid(); line.advance()) {
+    double cost = costmap_->getCost(line.getX(), line.getX());
```
SteveMacenski, line 155: getY I think for the second one?

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto & current = queue.front();
+    queue.pop();
+
+    if (iteration > max_iterations_) {
+      throw nav2_core::PlannerTimedOut("Exceeded maximum iterations");
+    }
+
+
+    // Check goals
+    for (unsigned int i = 0; i < goals_.size(); ++i) {
```
SteveMacenski, line 89: This is is what I mentioned about continuing until the cost is equal so that we get the closest one, not necessarily the first one since the search pattern isn't radial

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+  int iteration = 0;
+  while (!queue.empty()) {
+    auto & current = queue.front();
+    queue.pop();
+
+    if (iteration > max_iterations_) {
+      throw nav2_core::PlannerTimedOut("Exceeded maximum iterations");
+    }
+
```
SteveMacenski, line 86: extra space

SteveMacenski (changes requested):

SteveMacenski:
You have a few functions where you do `costmap_sub->getCostmap()->Fn()` multiple times. Store a ref to `costmap_sub->getCostmap()` so its more readable that you're just calling `costmap->Fn()`

nav2_route/src/route_server.cpp
jwallace42: We talked about this while back. The breadth first search throws a lot of the same errors as the planner server. We could add a separate set for errors for bfs but it would be just be a subset of the planner server errors.

nav2_route/src/node_spatial_tree.cpp
jwallace42: I am not quite sure which ordering issue you are referring too :).

nav2_route/src/goal_intent_extractor.cpp
jwallace42: same as above.

nav2_route/src/goal_intent_extractor.cpp
jwallace42: Nope, 

but then the assumption would be that the local cost map always overlaps a node in the route graph.

nav2_route/src/goal_intent_extractor.cpp
jwallace42: no idea haha, 

this was late one night :). I should stop programming after 11.

nav2_route/src/goal_intent_extractor.cpp
jwallace42: Well, 

the route graph could be in map frame and the local costmap could be in odom frame.

SteveMacenski: True, if using the local costmap this could check if the `costmap->getGobalFrame()` is the same as the route serverframe. If not, do that transform

nav2_route/src/goal_intent_extractor.cpp
SteveMacenski: hah No worries :-)

nav2_route/src/goal_intent_extractor.cpp
SteveMacenski: Sorry, I'm not saying we should change it to the local costmap frame, only that we should be able to accept it if someone wants to set it up that way. So the param shouldn't require the global costmap and handling for the local costmap should be considered

nav2_route/src/node_spatial_tree.cpp
SteveMacenski: Something in slack about the order of ... something not being as you expected. Maybe my memory is failing me

nav2_route/src/route_server.cpp
SteveMacenski: Sorry, I was unclear. My comment is more: why? If we have an `activate`, shouldn't we at least have the complimentary `deactivate` to undo whatever `activate` did? If this is more just an `initialize` that doesn't need deactivation, rename to `initialize` so we don't misinterpret this as a lifecycle necessary thing.

nav2_route/src/route_server.cpp
SteveMacenski: Ah ok, but we're returning in all actions the route-specific failure codes, right?

jwallace42: Yeah, 

the error code we use have been added to the action message.

nav2_route/src/node_spatial_tree.cpp
jwallace42: I think that was in reference to the order of nodes found by the breath first search algorithm. :). At any rate, it would be good to double check that the order is deterministic from the rtree.

SteveMacenski (changes requested):
And still some of the open things that are more major :-)

nav2_route/src/breadth_first_search.cpp
```diff
@@ -0,0 +1,187 @@
...
+    // Check for wrap around conditions
+    if (std::fabs(static_cast<float>(p_mx) - static_cast<float>(n_mx)) > 1 ||
+      std::fabs(static_cast<float>(p_my) - static_cast<float>(n_my)) > 1)
+    {
+      continue;
+    }
+
+    // Check for out of bounds
+    if (neighbor_index >= max_index_) {
```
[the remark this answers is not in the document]
jwallace42: It is a unsigned int so it should never be negative or something else can seriously wrong.

nav2_route/src/goal_intent_extractor.cpp
```diff
@@ -193,6 +255,101 @@ Route GoalIntentExtractor::pruneStartandGoal(
...
+    auto node = graph_->at(node_index);
+    route_goal.pose.position.x = node.coords.x;
+    route_goal.pose.position.y = node.coords.y;
+    route_goal.header.frame_id = node.coords.frame_id;
+    goal = transformPose(route_goal, costmap_frame_);
+
+    float goal_x = goal.pose.position.x;
+    float goal_y = goal.pose.position.y;
+    if (!costmap_sub_->getCostmap()->worldToMap(goal_x, goal_y, g_mx, g_my)) {
+      RCLCPP_WARN_STREAM(
```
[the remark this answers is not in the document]
jwallace42: I think we should think about the cases were we plan to use the local costmap. 

The current implementation doesn't quite make sense. I good example when we are trying to do goal association. We will likely be outside of the local costmap bounds. 

I think it only makes sense if, use start is false and we only associate the start.

nav2_route/src/breadth_first_search.cpp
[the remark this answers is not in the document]
SteveMacenski:
```
    int index = parent_index + neighbors_grid_offset;
    if (index < 0) {
      continue;
    }
```

Check the >= there.

nav2_route/src/goal_intent_extractor.cpp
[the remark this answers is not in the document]
SteveMacenski: Its less about the specifics of a rolling local costmap as much as it as supporting a **rolling** costmap in general. The GPS threads are making me think about the need to make all the servers accept rolling global costmaps (which most will actually already) due to navigating very large spaces that you either don't have a map or its unrealistic to have it all loaded at once.

jwallace42:
Things left. 

1. test local costmap usage
2. Benchmark

SteveMacenski:
Finishing this work and rolling it into https://github.com/ros-navigation/navigation2/pull/3536!

PR #3539, opened 2023-04-24
Behavior server error codes

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | turtlebot simulation |

---

Updated the behavior server to handle error codes. 

Added new entries in error code table. 

---

#### For Maintainers: 


nav2_behavior_tree/plugins/action/back_up_action.cpp
```diff
@@ -46,6 +46,24 @@ void BackUpAction::on_tick()
...
+
+BT::NodeStatus BackUpAction::on_aborted()
+{
+  setOutput("error_code_id", result_.result->error_code);
+  return BT::NodeStatus::FAILURE;
+}
+
+BT::NodeStatus BackUpAction::on_cancelled()
+{
+  setOutput("error_code_id", result_.result->error_code);
```
SteveMacenski, line 63: Shouldn't this be none if canceled?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/spin_action.hpp
```diff
@@ -55,11 +59,28 @@ class SpinAction : public BtActionNode<nav2_msgs::action::Spin>
       {
         BT::InputPort<double>("spin_dist", 1.57, "Spin distance"),
         BT::InputPort<double>("time_allowance", 10.0, "Allowed time for spinning"),
-        BT::InputPort<bool>("is_recovery", true, "True if recovery")
+        BT::InputPort<bool>("is_recovery", true, "True if recovery"),
+        BT::OutputPort<ActionResult::_error_code_type>(
+            "error_code_id", "The spin behavior error code")
```
SteveMacenski, line 64: New port to the BT.xml file for use + the behavior tree node xml configuration for Groot

nav2_msgs/action/Wait.action
```diff
@@ -1,8 +1,13 @@
+# Error codes
+# Note: The expected priority order of the error should match the message order
+uint16 NONE=0
```
SteveMacenski, line 3: This feels unnecessary if there's actually no real error to be had here

nav2_msgs/action/Spin.action
```diff
@@ -1,9 +1,18 @@
+# Error codes
+# Note: The expected priority order of the error should match the message order
+uint16 NONE=0
+uint16 UNKNOWN=700
+uint16 TIMEOUT=701
+uint16 TF_ERROR=702
+uint16 COLLISION_AHEAD=703
```
SteveMacenski, line 7: No invalid input?

nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -83,25 +83,25 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+  ResultStatus onCycleUpdate()
   {
     rclcpp::Duration time_remaining = end_time_ - this->steady_clock_.now();
     if (time_remaining.seconds() < 0.0 && command_time_allowance_.seconds() > 0.0) {
       this->stopRobot();
       RCLCPP_WARN(
         this->logger_,
         "Exceeded time allowance before reaching the DriveOnHeading goal - Exiting DriveOnHeading");
-      return Status::FAILED;
+      return ResultStatus{Status::FAILED, 0};
```
SteveMacenski, line 104: `0` should be a named error for readability

SteveMacenski (changes requested):
Generally LGTM, usual docs + updates for xml and groot vis.

nav2_msgs/action/Spin.action
jwallace42: Spin currently has no check to invalidate any input.

nav2_msgs/action/Wait.action
```diff
@@ -3,6 +3,7 @@ builtin_interfaces/Duration time
 ---
 #result definition
 builtin_interfaces/Duration total_elapsed_time
+uint16 error_code
```
SteveMacenski, line 6: This feels unnecessary if there's no error codes

nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
```diff
@@ -45,6 +45,12 @@ enum class Status : int8_t
   RUNNING = 3,
 };
 
+struct ResultStatus
+{
+  Status status;
+  unsigned short error_code{0};  //NOLINT
```
SteveMacenski, line 51: Why does this need no linting?

SteveMacenski (commented):
I think last is for the BTs to use it. Anything else you think we need? Maybe I suppose unit tests for the new functions added if not already covered

nav2_msgs/action/Wait.action
jwallace42: Yeah, but anything using the timed behavior needs a error code.

nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
SteveMacenski: ^ not much of a reason for this not to just be in unsigned int, really. But I don't really care

SteveMacenski:
Just on CI

nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
jwallace42: Switched this to a uint16_t :)

jwallace42:
Sweet. 
 
I will finish those up :)

SteveMacenski:
Should the default BTs be updated to use them? I suppose not necessarily required, but might be good at least for the main 2-3 trees since those are minimal examples folks will work from

SteveMacenski:
This is actually a time when mergify is correct, rebase :smile:

SteveMacenski (approved):
Waiting on CI. I'll look tomorrow

PR #3542, opened 2023-04-25
ignore collision monitor

jwallace42:
Isolate the build farm issue.


SteveMacenski:
This isn't an acceptable solution and isn't the cause. Its actually any messages with the constants, including the error codes that impact every server.

jwallace42:
Yeah, I was just trying to find the core issue. I didn't want it to be merged :).

Issue #3545, opened 2023-04-28
[bt_navigator] Crashes if lifecycle configure again


BriceRenaudeau:
## Bug report
The bt_navigator node crashes when it is restarted using lifecycle process.
```
[INFO] [1682672061.610984122] [lifecycle_manager_navigation]: Configuring bt_navigator
[INFO] [1682672061.611233341] [bt_navigator]: Configuring
[ERROR] [1682672061.612488199] []: Caught exception in callback for transition 10
[ERROR] [1682672061.612506160] []: Original error: parameter 'navigators' has already been declared
[WARN] [1682672061.612526313] []: Error occurred while doing error handling.
[FATAL] [1682672061.612535295] [bt_navigator]: Lifecycle node bt_navigator does not have error state implemented
[ERROR] [1682672061.613942386] [lifecycle_manager_navigation]: Failed to change state for node: bt_navigator
[ERROR] [1682672061.613985222] [lifecycle_manager_navigation]: Failed to bring up all requested nodes. Aborting bringup.
```

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble
- Version or commit hash:
  - Humble
- DDS implementation:
  - cyclon_dds

#### Steps to reproduce issue
- Start the navigation stack. S
- Stop it using the lifecycle `deactivate` and `cleanup`.
- Restart it using the lifecycle `configure` and `activate`

#### Expected behavior
The node is restarted properly.

#### Actual behavior
The node is not started hence the navigation doesn't work.

#### Additional information
It happens because `declare_parameter()` is called in the `on_configure()` method.
https://github.com/ros-planning/navigation2/blob/1cdc277b84115ee97c94b60b2a14630d097c6975/nav2_bt_navigator/src/bt_navigator.cpp#L140

So when the node is configured again, it crashes on the already declared parameter.

BriceRenaudeau:
I will make a PR to fix this.

SteveMacenski:
Ah thanks!

PR #3546, opened 2023-04-28
Fix bt_navigator crashes when restarted

BriceRenaudeau:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3545 |
| Robotic platform tested on | gazebo simulation  |

---

* move `declare_parameter("navigators")` in the constructor
* remove useless `default_navigator_types`

#### For Maintainers: 

### DIFF
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -90,12 +90,12 @@ BtNavigator::BtNavigator(const rclcpp::NodeOptions & options)
   declare_parameter("global_frame", std::string("map"));
   declare_parameter("robot_base_frame", std::string("base_link"));
   declare_parameter("odom_topic", std::string("odom"));
-    
+
   // Navigator defaults
   const std::vector<std::string> default_navigator_ids = {
     "navigate_to_pose",
     "navigate_through_poses"
-  };    
+  };
   declare_parameter("navigators", default_navigator_ids);
 }
 
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -20,8 +20,11 @@
 #include <vector>
 
 #include "nav2_util/geometry_utils.hpp"
+#include "nav2_util/node_utils.hpp"
 #include "nav2_util/robot_utils.hpp"
 #include "nav2_behavior_tree/bt_conversions.hpp"
+
+using nav2_util::declare_parameter_if_not_declared;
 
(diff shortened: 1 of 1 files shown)


nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -126,26 +133,9 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
...
-  const std::vector<std::string> default_navigator_types = {
-    "nav2_bt_navigator/NavigateToPoseNavigator",
-    "nav2_bt_navigator/NavigateThroughPosesNavigator"
-  };
-
+  // Load navigator plugins
   std::vector<std::string> navigator_ids;
-  declare_parameter("navigators", default_navigator_ids);
   get_parameter("navigators", navigator_ids);
-  if (navigator_ids == default_navigator_ids) {
```
SteveMacenski, line 142: This block is strictly required to declare the types of the defaults. This is non-optional. Please readd `default_navigator_types` definition and this block. You can use the `nav2_utils::declare_if_not_declared` function so that it'll only redeclare on lifecycle transition if not already declared. 

You could actually do that for `navigators` too so that you didn't have to change anything other than 2 lines. That would actually be closer in line with best practices.

SteveMacenski (changes requested):

SteveMacenski:
Just minor fix to merge :-)

nav2_bt_navigator/src/bt_navigator.cpp
BriceRenaudeau: Using `declare_if_not_declared` was my first thought but it doesn't exist anymore.

Maybe default navigators are not necessary (yaml is always required) or I can also declare them in the constructor.

SteveMacenski: I'm not sure what you mean by that, we have that in quite a few places in Nav2 :smile: 

https://github.com/ros-planning/navigation2/blob/main/nav2_navfn_planner/src/navfn_planner.cpp#L84

Its a `nav2_utils` function

BriceRenaudeau: My bad, I was looking for `declare_if_not_declared` instead of `declare_parameter_if_not_declared` ...
I will make the modification.

SteveMacenski:
@BriceRenaudeau every single system test failed - maybe you should rebase? I also retriggered the tests in case it was a fluke

SteveMacenski:
Actually I confirmed this was happening in other days builds. I'll need to look into that. Thanks!

PR #3551, opened 2023-05-01
Improve Dev Container Web App Visualization

ruffsl:
This improves the ergonomics of Web App Visualization for Dev Containers, particularly those running from Codespaces, by consolidating the serving of HTTP traffic over a single network port. This simplifies the user experience when traversing the apps themselves, but also facilitates cookie-based authentication when using the secure and private by default port forward settings in Codespaces. This is achieved by using reverse proxies to reroute websocket traffic for each web app to their respective backend bridge interface. This PR also includes a number of other improvements, including a simple landing page for users to open and auto connect each web app from a web browser.


ruffsl:
Merging for integration testing and to rebuild repo's codespace.

ruffsl:
## Refactoring

After conducting further experimentation and closely studying Caddy's documentation, I concluded that some refactoring was necessary. My intention was to simplify the Caddyfile and make it more formulaic, resembling a template that others could easily extend. For example, they could add more apps, modify the static file locations, or even reuse it outside of development containers.

Furthermore, while testing on mobile devices, I found the user experience to be unsatisfactory. Specifically, I was not pleased with the consistency of each web app's behavior upon installation. For instance, while Foxglove could be opened in a full-screen window or standalone view on Android or iOS, other apps like Gzweb or the system monitor would only reopen using regular browser tabs. This inconsistency prompted me to seek ways to further enhance the overall experience.

Now, let's delve into how we can improve these issues.

### Caddyfile

The prior Caddyfile configurations employed a basic declarative style, which was simple but had certain drawbacks. The reliance on a monolithic root and symlinks made the setup fragile, not easily extensible, and not very portable. Adding or extending hosted apps required synchronizing symlink changes to the root path. Moreover, using the same Caddyfile outside of a development container necessitated replicating these symlinks in other locations.

To address these issues, we can serve each app using its own configurable directory by leveraging user defined environment variables. This approach allows us to dynamically specify the root paths for each app during file server startup. To streamline the use of multiple `root` directives, we can utilize the `handle_path` directive. It enables us to remove the matching prefix from the request path, ensuring that requests for each app are handled within separate scopes. This scoping mechanism helps prevent conflicts with directives specific to each app, such as the `templates` directive, which previously required manual matching to avoid unintentionally overwriting index files.

This approach also enables us to adopt a more formulaic and DRY (Don't Repeat Yourself) approach with code snippets. Subsequent path matchers can now be left relative and reused between apps. However, there is a caveat when it comes to incorporating the `rewrite` directive. Previously, with a single `file_server` directive under a monolithic root, using the `rewrite` directive to gracefully reroute websocket connections when the trailing slash in the directory path was omitted was relatively simple. But now, with multiple `handle_path` directives used to match each app's root directory, achieving the same result becomes more complex. 

Fortunately, the newly added redirect already automates any websocket setup by programmatically populating URL parameters used by Foxglove, avoiding the issue of typographical errors when addressing websocket URLs altogether.

### Mobile View

When using web apps on smaller screen devices like student-owned tablets or school-issued Chromebooks, the app's display is crucial for enhancing usability. In most browsers, the URL bar and navigation buttons take up a significant portion of the vertical real estate, especially on modern screens with wider aspect ratios. While enabling the browser's full-screen mode can alleviate this issue, it can limit the use of multitasking features like multi-window or split-screen views, which are supported even on modest devices.

To address this, users have the option to install web apps as standalone apps. These apps can be launched from the home screen or app drawer, allowing them to run in their own resizable windows separate from the browser. While Foxglove can already be installed as a standalone web app natively, others may not have this capability. To enable this feature, we can add the following `<meta>` tags to the `index.html` page of each app:

```html
<head>
  <meta name="apple-mobile-web-app-capable" content="yes">
  <meta name="mobile-web-app-capable" content="yes">
  ...
```

- Making Fullscreen Experiences
  - https://web.dev/fullscreen/

In fact, you can see this added to Foxglove here; which first clued me in on this feature:

- Hide Safari chrome when launched from Home screen on iOS # 4149
  - https://github.com/foxglove/studio/pull/4149

While we could of course open similar PRs towards every such project we incorporate into our web stack, that could turn our current rabbit whole into a never ending fractal of forks. To keep forking to a minimum, we can instead patch our single page web apps as we serve them using Caddy's `replace_response` module. This module allows us to inject arbitrary strings into the response body of each app given a matching pattern, allowing Gzweb, our system monitor, and even our simple landing page to be displayed as standalone apps as well.

- Caddy replace_response handler module
  - https://github.com/caddyserver/replace-response
- Module http.handlers.replace_response
  - https://caddyserver.com/docs/modules/http.handlers.replace_response

However, this module is not part of the standard Caddy build, although it is maintained by the official Caddy organization. To use it, we have two options: either build Caddy from source or utilize the `xcaddy` tool to create a custom Caddy binary that includes the module. It's worth noting that the `xcaddy` tool is not distributed through Caddy's official Debian repository. Initially, I was concerned that this would lead to a complex detour involving the setup of Golang and friends. However, to my delight, I discovered the preinstalled `xcaddy` docker image tag in Docker Hub's Official Library. This simplified the process of building and installing custom Caddy binaries, even when compared to using `apt-get`.

- Docker Hub: Caddy - Adding custom Caddy modules
  - https://hub.docker.com/_/caddy

### Web App Manifest

While the aforementioned `<meta>` tags effectively enable the display of the app as a standalone web app on all browsers, the default startup behavior still lacks consistency across various platforms. For example, when installing Foxglove as a standalone app on desktop or Android devices, the app state and the complete URL are utilized for installation to the desktop or home screen. However, this is not the case on iOS devices, where only the URL path is utilized, resulting in the removal of any query parameters. Consequently, reopening the standalone web app of Foxglove on iOS does not automatically establish a connection, whereas reopening the web app from a bookmarked URL, the same URL that was used during the web app installation, does auto connect.

To address this issue, we can instead adopt a more contemporary and robust approach by leveraging the Web App Manifest standard to declare the apps as compatible with mobile views. This enables us to define the start URL that is utilized when the app is installed as a standalone application, giving us the ability to include a customized query parameter to ensure consistent launching behavior. To do this, we'll instead add the following `<link>` tag to the `index.html` page of each app:

```html
<head>
  <link rel="manifest" href="manifest.json" />
  ...
```
- Web app manifests
  - https://developer.mozilla.org/en-US/docs/Web/Manifest
- Add a web app manifest
  - https://web.dev/add-manifest/

Then we'll configure the Caddy web server to add a redirect to the respective manifest file. Not only does this allow us to specify the start URL and display mode for each app, but also customize the default name of the app as it appears to the user. We can then again use server side templates to dynamically include the host and port information into the manifest file, clueing the user into what domain the web app is hosted from. This is particularly useful when switching between local dev container or remote Codespace environments, or when switching between Codespaces with different names and thus different port forwarded web domains.

```json
{
    "name": "Foxglove: {{placeholder "http.request.hostport"}}",
    ...
    "start_url": "/foxglove/autoconnect",
    "display": "standalone",
    "shortcuts" : [
        {
          "name": "Auto Connect",
          "url": "/foxglove/autoconnect",
          "description": "Auto connect to default data source"
        },
        {
          "name": "Auto Layout",
          "url": "/foxglove/autolayout",
          "description": "Auto connect using default layout"
        },
        {
          "name": "Manual Connect",
          "url": "/foxglove/",
          "description": "Manually connect to data source"
        }
      ]
}
```

While we're at it, we can also incorporate shortcuts into the manifest file, enabling users to quickly select different startup options through various redirects. Instead of directly templating the complete Foxglove URL, including query parameters, into the server-side generated manifest, we'll just utilize more redirects. This approach avoids hardcoding the URLs into the installed web app, allowing us to centrally update these configurations from the server, while avoiding the need to update each individually installed web app on user devices.

Although, as of writing, `shortcuts` still seem to be an experimental feature:

- W3C: Web App Manifest - Shortcuts
  - https://w3c.github.io/manifest/#shortcuts-member
- MDN: Manifest - Shortcuts
  - https://developer.mozilla.org/en-US/docs/Web/Manifest/shortcuts

While experimenting with redirects, I discovered that Caddy offers a matcher directive to filter URL query parameters by matching specific value patterns. Although using query matching instead of path matching might be preferable in this case, especially when injecting custom redirects onto an existing web app frontend, I found paths to be more reliable in practice. This is because using paths exclusively for redirects ensures that the browser does not inadvertently cache outdated responses, as it could when requesting the root path of the Single Page Application (SPA). Short of overriding the app's own caching policy, we can side step such caching issues by matching on reserved redirect paths instead.

> Another annoyance specific to iOS: the browser cache in Safari doesn't seem to be shared with standalone web apps. In the example manifest provided earlier, you can see that the `start_url` member is set to the `autoconnect` redirect instead of the `autolayout` redirect. This avoids the need to download duplicate nav2 layouts every time the standalone web app is launched and ensures that the same layout selected from the previous session is reused.

> However, this also means that when iOS users first navigate to the Foxglove web app from Safari, using the `autolayout` redirect from the landing page, download the nav2 layout, and then install the web app: subsequent relaunches of the web app in standalone view will not include the cached nav2 layout. This deviates from the behavior of every other platform. Uhg, mobile web development in 2023... 🤦‍♂️

### Miscellaneous

With respect to the dev container setup, we can incorporate a few additional miscellaneous improvements. These include leveraging more environment variables to avoid hardcoding container-side paths in config files and enhancing the available tasks in VS Code. As mentioned in previous dev container PRs, these tasks, which facilitate starting and stopping the runtime dependencies for web-based visualizations, are particularly useful when restarting simulations or managing ROS launch files. Given that Gzweb and Foxglove do not handle simulation restarts gracefully, such as when encountering connection drops to gzserver or non-causality of time when caching TF trees in Foxglove Studio, having a quick way to restart the entire simulation stack is beneficial.

Previously, the use of `pkill` was somewhat haphazard to enable tasks to stop other running tasks. While functional, a better approach is to command VS Code directly to terminate tasks. This can be achieved by utilizing input variable substitution to specify the labeled task to terminate. This eliminates the need for convoluted pattern matching based on process command strings, ensuring that the correct process is always terminated correctly.

To learn more about variable substitution and input variables in VS Code, you can refer to the following documentation:

- VS Code Docs: Variable substitution
  - https://code.visualstudio.com/docs/editor/tasks#_variable-substitution
- VS Code Docs: Input variables
  - https://code.visualstudio.com/docs/editor/variables-reference#_input-variables

Additionally, this cleanup effort resolves issues related to the list of tasks displayed under the terminal panel in VS Code. Previously, terminated task terminals would persist, or reused terminals would exhibit inconsistent icons and names. Now, terminated tasks are removed from the display, providing clearer visibility to students regarding running and inactive processes. This helps avoid confusion and issues like the one mentioned in the following example:

- Task icon for terminal tab entry does not update with task label when terminal is reused
  - https://github.com/microsoft/vscode/issues/179337

Lastly, there was an issue where `xvfb-run` and `gzserver` were occasionally orphaned by ROS launch when restarting simulations. This resulted in subsequent relaunches failing due to display server errors or port usage conflicts, which could be perplexing for younger students. By properly using `exec` in the wrapper bash script for the headless version of `gzserver`, hanging child processes now receive escalated `SIGKILL` signals. This is a known issue, as Gazebo classic often requires such measures.

- gzserver not dying after killing the simulation # 751
  - https://github.com/ros-simulation/gazebo_ros_pkgs/issues/751

Issue #1975, opened 2020-09-03
Docking controller + go-to-dock BT node


SteveMacenski:
Add a controller suited specifically for docking with some open standards on docks. Then add in a go-to-dock Bt node to go to a staging point and then enable the docking controller to dock the robot. This will allow autonomous docking with the is-charge-low BT node for a robot to go back home when it needs to. 

The docking controller should work on a configuration of the dock shape, using april tags, or some other form of standardizable relative pose estimation for the docking controller to target an approach at a particular vector and end position. 

https://datascience.stackexchange.com/questions/6838/when-to-use-random-forest-over-svm-and-vice-versa

Also offer a Bt with this built in (if charge < X, go to docking subtree. Once charge > Y, reenter navigation subtree)

tal-grossman:
I'v been working on such a controller and BT node my self for our robot, and I would like to have it out-of-the-box in the nav2 stack.

I can also guess [this issue](https://github.com/ros-planning/navigation2/issues/1979) and [adding BTActionServer PR](https://github.com/ros-planning/navigation2/pull/2010/files) are some how related to this issue and will provide an easy implementation method. 

I still have a design conflict and i was wondering if should it be addressed in the PR.
the use-case is **transition** from one application server to another. for example:

go-to-dock to navigateToPose
Or
navigateToPose to FollowDynamicObject

should one server stop and perform some kind of procedure and only then start the second server (FSM like behavior)?
I know that we kinda get that behavior when alternating between action servers during the behavior tree runtime.

Another way to resolve this case is to have some-kind of application server manager which will abort one server and trigger the other one. In my opinion this way is messier and less deterministic. 

anyway I think this use-case it worth considering @SteveMacenski .

SteveMacenski:
Those 2 issues are not related at all. If you can create a controller plugin for docking, then in the BT you must only use the `FollowPath` BT node with the `controller_id` set to the docking controller plugin name in the controller server. There shouldn't be any additional need for those capabilities except when making autonomy logic through the BT (which we aim to support, but is not required to work in our framework for that; NavigateToPose BT node exists so any BT library can use it). 

Nav2 can have different Bts run for different tasks. If you send a navigate to pose action to BT navigator, it can include an optional BT to be able to use any arbitrary BT for your task. But the going to dock could be trivially baked into the main BT by just checking if battery is low condition node (provided) or requested, then execute a sub-tree that gets a robot to a staging pose and switch to the docking specific controller. 

[...]

PR #3553, opened 2023-05-02
Add IsBatteryChargingCondition

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | (Ubuntu 22.04) |
| Robotic platform tested on | Metralabs Scitos|

---

* Added a new BT condition node to check if the battery is charging. Including is BT XML.

* Added new BT condition node, so need to add that to https://navigation.ros.org/plugins/index.html#behavior-tree-nodes

---

#### For Maintainers: 


BriceRenaudeau:
You will get FALSE if the battery is full. Is it the intended result?
```
uint8 POWER_SUPPLY_STATUS_CHARGING = 1
uint8 POWER_SUPPLY_STATUS_DISCHARGING = 2
uint8 POWER_SUPPLY_STATUS_NOT_CHARGING = 3
uint8 POWER_SUPPLY_STATUS_FULL = 4
```
In my case, I use it to know if the robot is on its charging station. So I set the result to TRUE when the battery is full.

nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp
```diff
@@ -0,0 +1,65 @@
...
+  is_battery_charging_ = (msg->power_supply_status == sensor_msgs::msg::BatteryState::POWER_SUPPLY_STATUS_CHARGING);
+}
+
+}  // namespace nav2_behavior_tree
+
+#include "behaviortree_cpp_v3/bt_factory.h"
+BT_REGISTER_NODES(factory)
+{
+  factory.registerNodeType<nav2_behavior_tree::IsBatteryChargingCondition>("IsBatteryCharging");
+}
```
SteveMacenski, line 65: New line at end of file

nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2022 Alberto J. Tudela Roldán
```
SteveMacenski, line 1: 2023

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_battery_charging_condition.hpp
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2022 Alberto J. Tudela Roldán
```
SteveMacenski, line 1: Ditto

nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+IsBatteryChargingCondition::IsBatteryChargingCondition(
+  const std::string & condition_name,
+  const BT::NodeConfiguration & conf)
+: BT::ConditionNode(condition_name, conf),
+  battery_topic_("/battery_status"),
+  is_battery_charging_(false)
+{
+  getInput("battery_topic", battery_topic_);
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
```
SteveMacenski, line 30: Why store the ROS node in the BT node? Its not used outside of this function's scope

nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+#include <string>
+
+#include "nav2_behavior_tree/plugins/condition/is_battery_charging_condition.hpp"
+
+namespace nav2_behavior_tree
+{
+
+IsBatteryChargingCondition::IsBatteryChargingCondition(
```
SteveMacenski, line 22: Needs unit test coverage, see `/test` directory for the template, it should be pretty easy to add the true, false cases to test

SteveMacenski (changes requested):
Generally LGTM, just details to finish it up. We'll also needs a docs PR to update the list of BT nodes on the plugins page, add a BT node entry page for the configuration guide, and a really brief migration guide entry

ajtudela:
Hi,

@BriceRenaudeau  Yes, I use it to know if our robots are in the docking station too. In our case, our robots can be fully charged outside the docking station so set the result to TRUE when the battery is full won't work for us.

@SteveMacenski I'll fix the typos, test coverage and docs.

For both, to allow a wide use cases, it could be implemented both options:
- Add a port bool param to include FULL or not in this BT node. That will work for robots like BriceRenaudeau's
- Add a new "isBatteryCharged" BT node to identify charging vs charged. For example, in cases where you have to send the robot to perform a task when is fully charged.

SteveMacenski:
>   Add a port bool param to include FULL or not in this BT node. That will work for robots like BriceRenaudeau's
  
OK - I think maybe the condition when this should be evaluated is `FULL && !DISCHARGING`  or perhaps `FULL && NOT_CHARGING` -- I think the first is more correct, since it could be full & not charging right after leaving a dock for a few minutes. 

> Add a new "isBatteryCharged" BT node to identify charging vs charged. For example, in cases where you have to send the robot to perform a task when is fully charged.

That could work too. That pushes the logic into the BT, which I usually like, but I feel like that's a bit much. This seems to be all easily put into one node.

ajtudela:
> OK - I think maybe the condition when this should be evaluated is `FULL && !DISCHARGING` or perhaps `FULL && NOT_CHARGING` -- I think the first is more correct, since it could be full & not charging right after leaving a dock for a few minutes.

I'm not exactly sure what you want to accomplish with that first condition or I'm missing the point here. 

If I understood correctly to @BriceRenaudeau, he wants something like:

```
if (include_full){
  is_charging = (status == CHARGING || status == FULL);
}else{
  is_charging = (status == CHARGING)
}
```

Or you want to "detect" the other status?

SteveMacenski:
I'm trying to find a way to address the problem you mentioned, which is a very good point that we should be able to know when its on the dock but fully charged vs off the dock and fully charged

>  In our case, our robots can be fully charged outside the docking station so set the result to TRUE when the battery is full won't work for us.

Instead of just looking at `FULL`, if we did `FULL && !DISCHARGING` that would tell us that we're full, but still located on the dock because we're not discharging. That would differentiate full + driving around. I'm suggesting the same thing as @BriceRenaudeau but including the additional check so that it restricts to the intended situation (fully charged but still on the dock) so that the node can optionally be used as a general "am I on the dock" condition.

Does that make sense?

ajtudela:
Ok, ok. I think I got it. 

Because the power_supply_status enum is limited to one status at the same time: FULL, CHARGING, DISCHARGING, etc, that condition will be address better by checking current (-) and percentage (1.0) instead.

SteveMacenski:
> enum is limited to one status 

.... good point. That seems limiting. My bad.

I think using status `FULL` and discharge would be best (checking full as `1.0` leaves floating point issues to consider), but that's starting to rely pretty heavily on a fully-implemented battery system. You tell me: is it realistic to try to do that? Would discharge be negative, zero, or still positive if fully charged on the dock? Is that consistent across most types of electrical systems? I don't have a setup like this to get numbers myself

ajtudela:
Agreed with the floating issues. 

I think, at least, power_supply_status, current, voltage and percentage will be filled. Only voltage is mandatory as BatteryState.msg definition, though. In our robots (based on MetraLabs scitos) that ones and cell_voltage are filled. 

If I remember the test I made the other day, current values are oscilating between low negatives and low positive values (-0.2...., +0.2..) when the battery is full but I will check again tomorrow.

SteveMacenski:
OK. I suppose also good reason for this to be an optional setting and we should make a note in the docs that this is sensitive to a properly populated battery message

SteveMacenski:
> only relying on the BatteryState

I don't believe you can, kind of why I mention this would be a separate topic :upside_down_face:. But if tied in with a docking node, we could create a message or expose a service to ask the question "am I docked?" and the docking node would know the current state, regardless of the battery. 

> In fact, I'm the process of migrating to ROS 2 our docking system and I was wondering if it would be a good addition to the navigation2 code.

I'd be open to chatting about that! Want to start a thread in slack or on the Docking ticket? https://github.com/ros-planning/navigation2/issues/1975 I basically haven't worked on it because I assumed someone at some point would have a company based system they might want to donate as a starting point :smile:

ajtudela:
I just took this picture of the battery state (current and percentage) this morning after a whole night of charging. 
battery
I was on the right track, the current oscilate between negatives and positive values when fully charged. 

So, instead checking the currrent for negatives value, maybe an option for this case could be something like ```fabs(current) < LOW_VALUE ``` with LOW_VALUE on optional input. Including ```FULL``` when it's charging is more complex than I thought as first.

Yes, I like the ide of a new message and BT for docking. I'll check the details to talk about this. I'm not in a company but an university; but our robots has docking successfully aver the last three years.

BriceRenaudeau:
> Does that make sense? Do you agree @BriceRenaudeau ?

Yes, it will work for us. We are not using the full range of states. For us, the battery is always `DISCHARGING` when it is not docked. So we only check this state to know if we are docked. So the FULL state is seen as NOT DISCHARGING.

DISCHARGING and NOT_CHARGING are often the same thing and that's why the question is difficult to answer. It depends on the implementation.

The current is not a good value to check. It is really battery dependent and some charging stations a doing sleep cycles where the robot is docked but discharging.

ajtudela:
Ok.

So, I think the best thing to do to simplify things is back to the original idea to check against ```CHARGING || FULL``` in this BT node and use it only to check battery status. Maybe add a little brief note in the docs about this.

And the move to a new BT node where we can actually check the dock status.

SteveMacenski:
Sure, that makes sense!

ajtudela:
> Sure, that makes sense! Or I don't know, does it make sense to include `FULL` now? It can be `FULL` and not on the dock for a short while, which could mess up some users, but I suppose it depends where this condition BT node is used. If its nested under something else that only passes when docked, we're good. If its in the main tree being used to check docked state, that could put you into a loop until the battery state updates to 99%

Yeah,... maybe it's better to left this with its original purpose, just charging.

What do you think @BriceRenaudeau ? Left this to check only charging and in a, hopefully, near future provide a new message and BT for docking.

ajtudela:
Added PR Info here: https://github.com/ros-planning/navigation.ros.org/pull/420

SteveMacenski (changes requested):
Otherwise LGTM with these fixes and the ones in the docs PR

SteveMacenski:
@ajtudela thanks for the contribution!

PR #3554, opened 2023-05-03
Split nav2_system_tests into separate CI jobs

ruffsl:
to avoid CI timeout issues with CircleCI.

Related:

- https://github.com/ros-planning/navigation2/pull/3543

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,7 +123,7 @@ _commands:
           type: string
         mixins:
           type: string
-        skip:
+        packages_skip:
           default: ""
           type: string
         restore:
@@ -200,7 +200,7 @@ _commands:
                     . << parameters.underlay >>/install/setup.sh
                     colcon build \
                       --packages-select ${BUILD_PACKAGES} \
-                      --packages-skip << parameters.skip >> \
+                      --packages-skip << parameters.packages_skip >> \
                       --mixin << parameters.mixins >>
               - ccache_stats:
                   workspace: << parameters.workspace >>
@@ -226,6 +226,9 @@ _commands:
           type: string
         cache_test:
           type: boolean
+        packages_skip:
+          type: string
+          default: ""
       steps:
         - run:
             name: Test Workspace | << parameters.workspace >>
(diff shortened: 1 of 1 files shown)


ruffsl:
@SteveMacenski , would you happen see why this is listing no packages, rather than all excluding `nav2_system_tests`?

``` bash
  BUILD_PACKAGES=$(
    colcon list \
      --names-only \
      --packages-select-regex * \
      --packages-skip-regex nav2_system_tests \
      --packages-above \
        $BUILD_UNFINISHED \
        $BUILD_FAILED \
        $BUILD_INVALID \
    | xargs)
```
> https://app.circleci.com/pipelines/github/ros-planning/navigation2/9252/workflows/f47a38db-cabd-4c01-a6b3-b0855478d735/jobs/29837?invite=true#step-117-35

ruffsl:
One caveat here is that we can either run the `system_build` job as soon as the `release_build` jobs succeeds, but then never cache the test results from `system_test` to avoid cache overwrites of test results from `release_test`.

OR

We can run the jobs linearly from `release_build -> release_test -> system_build -> system_test`, but this means that `system_test` may not get to run very often given how frequent tests in `release_test` fail. But perhaps that kind of fail fast and fail early is feature for the `build_and_test` workflow that is used for PRs. Not sure.

This is represented with before and after the commit: 95ff181b7750ab0e30edf9ecfbd4c9923ae98342 . Perhaps we could ad some more conditional glue for reviewers to override this fail fast behavior, but it might take a few hacks or more refactoring:

- https://discuss.circleci.com/t/run-a-job-even-if-required-jobs-fail/45466
- https://discuss.circleci.com/t/workaround-run-jobs-sequentially-regardless-of-the-outcome-of-the-required-jobs/40807

ruffsl:
I opted to keep the tests consolidated to one jobs and make it require the new `system_build` job instead, to simplify the caching of tests, given that the job wasn't timing out before.

SteveMacenski:
@ruffsl does this have a job for testing system tests or is that still in the main test job?

PR #3555, opened 2023-05-03
[controller_server] [FollowPath] Allow multiple progress checkers

doisyg:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Dexory sim |

---

The parameters `goal_checker_plugins` and `controller_plugins` are list of strings, allowing multiple plugin configuration, whereas `progress_checker_plugin` is a single string allowing the declaration of only one goal checker.

This PR initializes the progress checker plugin(s) in the same way as for the goal checker and controller plugins: it is now a list of string and was renamed from `progress_checker_plugin` to `progress_checker_plugins`. This allows the initialization of multiple goal checker that can be chosen from the added `progress_checker_id` field of the `FollowPath` action

This is a breaking change as the configuration yaml has to be updated from

```
 controller_server:
   ros__parameters:
     progress_checker_plugin: "progress_checker"
```

to

```
 controller_server:
   ros__parameters:
     progress_checker_plugins: ["progress_checker"]
```

- Name of parameter and default configuration in all examples
- Migration guide

---

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -301,12 +301,12 @@ ControllerServer::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
   controllers_.clear();
 
   goal_checkers_.clear();
+  progress_checkers_.clear();
   if (costmap_ros_->get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
   {
     costmap_ros_->cleanup();
   }
-  progress_checkers_.clear();
 
   // Release any allocated resources
   action_server_.reset();
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -679,9 +679,9 @@ void ControllerServer::updateGlobalPath()
         RCLCPP_INFO(
           get_logger(), "Change of progress checker %s requested, resetting it",
           goal->progress_checker_id.c_str());
-        progress_checkers_[current_progress_checker]->reset();
+        current_progress_checker_ = current_progress_checker;
+        progress_checkers_[current_progress_checker_]->reset();
       }
-      current_progress_checker_ = current_progress_checker;
     } else {
       RCLCPP_INFO(
         get_logger(), "Terminating action, invalid progress checker %s requested.",
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -84,11 +84,25 @@ ControllerServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
 
   RCLCPP_INFO(get_logger(), "getting progress checker plugins..");
   get_parameter("progress_checker_plugins", progress_checker_ids_);
+  try {
+    nav2_util::declare_parameter_if_not_declared(
+      node, "progress_checker_plugin", rclcpp::PARAMETER_STRING);
+    std::string progress_checker_plugin;
+    progress_checker_plugin = node->get_parameter("progress_checker_plugin").as_string();
+    progress_checker_ids_.clear();
+    progress_checker_ids_.push_back(progress_checker_plugin);
+    RCLCPP_WARN(
+      get_logger(),
+      "\"progress_checker_plugin\" parameter was deprecated. Use "
+      "\"progress_checker_plugins\" instead to specify a list of plugins");
+  } catch (const std::exception &) {
+    // This is normal situation: progress_checker_plugin parameter should not being declared
+  }
   if (progress_checker_ids_ == default_progress_checker_ids_) {
     for (size_t i = 0; i < default_progress_checker_ids_.size(); ++i) {
       nav2_util::declare_parameter_if_not_declared(
         node, default_progress_checker_ids_[i] + ".plugin",
-            rclcpp::ParameterValue(default_progress_checker_types_[i]));
+        rclcpp::ParameterValue(default_progress_checker_types_[i]));
     }
   }
 
@@ -663,15 +677,15 @@ void ControllerServer::updateGlobalPath()
     if (findProgressCheckerId(goal->progress_checker_id, current_progress_checker)) {
       if (current_progress_checker_ != current_progress_checker) {
         RCLCPP_INFO(
-            get_logger(), "Change of progress checker %s requested, resetting it",
-            goal->progress_checker_id.c_str());
-            progress_checkers_[current_progress_checker]->reset();
+          get_logger(), "Change of progress checker %s requested, resetting it",
+          goal->progress_checker_id.c_str());
+        progress_checkers_[current_progress_checker]->reset();
       }
       current_progress_checker_ = current_progress_checker;
     } else {
       RCLCPP_INFO(
-          get_logger(), "Terminating action, invalid progress checker %s requested.",
-          goal->progress_checker_id.c_str());
+        get_logger(), "Terminating action, invalid progress checker %s requested.",
+        goal->progress_checker_id.c_str());
       action_server_->terminate_current();
       return;
     }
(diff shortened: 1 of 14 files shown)


SteveMacenski (approved):
I generally approve, but I want to see the docs PR to match since there will be a bunch of places these new ports and vectorized plugins need to be set. Last time we did this, there were vestiges in docs of it for _way_ too long causing problems with copy+pasting of YAML

nav2_controller/src/controller_server.cpp
```diff
@@ -382,8 +426,16 @@ void ControllerServer::computeControl()
...
+    std::string current_progress_checker;
+    if (findProgressCheckerId(pc_name, current_progress_checker)) {
+      current_progress_checker_ = current_progress_checker;
+    } else {
+      throw nav2_core::ControllerException("Failed to find progress checker name: " + pc_name);
+    }
+
     setPlannerPath(action_server_->get_current_goal()->path);
-    progress_checker_->reset();
+    progress_checkers_[current_progress_checker_]->reset();
```
SteveMacenski, line 438: Shouldn't you reset all of them? How do you synchronize state that they have when switching between them within a single navigation request? That doesn't seem like a crazy thing to do / need if you're having multiple progress checkers for a single navigation task

SteveMacenski: What about preemption in `updateGlobalPath`? You don't change the progress checker to the new request's value. Is that intentional? I suppose that would handle the request synchronization based isuses I describe above

doisyg: > What about preemption in `updateGlobalPath`? You don't change the progress checker to the new request's value. Is that intentional? I suppose that would handle the request synchronization based isuses I describe above. But you'd need to be explicit about that and check if there's been a change to reject + log the invalid preemption request. Or, handle the switching state between them.

My bad, I missed it.
I see 3 options:

1. rejecting the goal in case of a switch of progress checker during a preempt
2. resetting the new progress checker in case of a switch.
3. some form of synchronization but I would need to think more about it

nav2_controller/src/controller_server.cpp
```diff
@@ -50,7 +50,7 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
 
   declare_parameter("controller_frequency", 20.0);
 
-  declare_parameter("progress_checker_plugin", default_progress_checker_id_);
+  declare_parameter("progress_checker_plugins", default_progress_checker_ids_);
```
SteveMacenski, line 53: Relatedly: I think it would be good to check for `progress_checker_plugin` and `goal_checker_plugin` (no `s`) and log a visible warning about the deprecation of that parameter and to use the updated parameter. 

In earlier times, we were changing so much I promised nothing in terms of compatibility between distributions (but would be documented in migration guides). I think now, things are stable enough that it makes sense to have a cycle for warnings for a distribution.

nav2_controller/src/controller_server.cpp
```diff
@@ -607,6 +659,23 @@ void ControllerServer::updateGlobalPath()
...
+    std::string current_progress_checker;
+    if (findProgressCheckerId(goal->progress_checker_id, current_progress_checker)) {
+      if (current_progress_checker_ != current_progress_checker) {
+        RCLCPP_INFO(
+            get_logger(), "Terminating action, change od progress checker %s requested in preempt.",
+            goal->progress_checker_id.c_str());
+        action_server_->terminate_current();
+        return;
+      }
+      current_progress_checker_ = current_progress_checker;
```
SteveMacenski, line 671: This line is unnecessary, its the same already to get to that point

nav2_controller/src/controller_server.cpp
```diff
@@ -607,6 +659,23 @@ void ControllerServer::updateGlobalPath()
       action_server_->terminate_current();
       return;
     }
+    std::string current_progress_checker;
+    if (findProgressCheckerId(goal->progress_checker_id, current_progress_checker)) {
+      if (current_progress_checker_ != current_progress_checker) {
+        RCLCPP_INFO(
+            get_logger(), "Terminating action, change od progress checker %s requested in preempt.",
```
SteveMacenski, line 666:
```suggestion
            get_logger(), "Terminating action, changing progress checker %s is not currently supported in preemption.",
```

nav2_controller/src/controller_server.cpp
SteveMacenski: > rejecting the goal in case of a switch of progress checker during a preempt

That's kind of kicking the can down the road, but I'm OK with that

> resetting the new progress checker in case of a switch.

I think that's the best solution. It supports preemption and hopefully a user knows that if they change the checkers the state will be reset. I don't foresee a situation where someone's doing that alot (... but now that I say that ...) and even if replanning comes in at a regular rate from the BT, as long as it doesn't change the progress checker in the request, we just don't reset it so it should be fine.

---

I leave it to you among the two. I think 2 is the best option but I don't feel strongly about it.

nav2_controller/src/controller_server.cpp
doisyg: I handled `progress_checker_plugin` deprecation in the same way `min_points` is here:
https://github.com/ros-planning/navigation2/blob/3cfc15e065bb1c83b7be8ddbb618567a99147785/nav2_collision_monitor/src/polygon.cpp#L270-L281

nav2_controller/src/controller_server.cpp
```diff
@@ -607,6 +673,22 @@ void ControllerServer::updateGlobalPath()
...
+      }
+      current_progress_checker_ = current_progress_checker;
+    } else {
+      RCLCPP_INFO(
+        get_logger(), "Terminating action, invalid progress checker %s requested.",
+        goal->progress_checker_id.c_str());
+      action_server_->terminate_current();
+      return;
+    }
     setPlannerPath(goal->path);
```
SteveMacenski, line 692: I think you need to `progress_checkers_[current_progress_checker_]->reset();` after `setPlannerPath()`. This is done in the main function on an action request to reset the state to a fresh start but not in preemption. This could have stale state from the last time this goal checker was used (or never, if never used)

SteveMacenski: You only reset the _old_ one above, which I'm not necessarily convinced you need to do (?)

SteveMacenski (changes requested):

nav2_controller/src/controller_server.cpp
SteveMacenski: Oh I see, you do `progress_checkers_[current_progress_checker]->reset();`, I read that the first time as `progress_checkers_[current_progress_checker_]->reset();`

Whereas the second is the old goal checker I thought you were resetting unnecessarily (but then not resetting the newly set one). That makes sense. That was my mistake. 

Why not move the `current_progress_checker_ = current_progress_checker;` into the nested if statement? I wouldn't do anything useful just within the first if statement when they are equal.

SteveMacenski (approved):

nav2_controller/src/controller_server.cpp
doisyg: > Why not move the `current_progress_checker_ = current_progress_checker;` into the nested if statement? I wouldn't do anything useful just within the first if statement when they are equal.

Agreed, I moved it in the if statement and now use **current_progress_checker_** for the reset, so it is clearer to read

nav2_controller/src/controller_server.cpp
```diff
@@ -82,11 +82,28 @@ ControllerServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
...
-      node, default_progress_checker_id_ + ".plugin",
-      rclcpp::ParameterValue(default_progress_checker_type_));
+      node, "progress_checker_plugin", rclcpp::PARAMETER_STRING);
+    std::string progress_checker_plugin;
+    progress_checker_plugin = node->get_parameter("progress_checker_plugin").as_string();
+    progress_checker_ids_.clear();
+    progress_checker_ids_.push_back(progress_checker_plugin);
+    RCLCPP_WARN(
+      get_logger(),
+      "\"progress_checker_plugin\" parameter was deprecated. Use "
```
SteveMacenski, line 96:
```suggestion
      "\"progress_checker_plugin\" parameter was deprecated and will be removed soon. Use "
```

SteveMacenski (approved):

Issue #3558, opened 2023-05-07
Simple Commander API - BasicNavigator - GetPath | planner_id and use_start get overriden


nakai-omer:
**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble     
- Version or commit hash:
  - ros-humble-navigation2	1.1.6-1jammy.20230327.054544
- DDS implementation:
  - cyclone-dds

#### Steps to reproduce issue
https://github.com/ros-planning/navigation2/blob/9b56556dfd9c50f4046d253317cd0949a4f34583/nav2_simple_commander/nav2_simple_commander/robot_navigator.py#L353

You can see in above line that parameters are being overridden by default values, causing planner_id and use_start to get the default values.

#### Expected behavior
`getPath` should use the parameters

#### Actual behavior
`getPath` overrides them with default values

nakai-omer:
Added a PR to fix this and a couple of more places with the same bug

SteveMacenski:
Will merge shortly!

nakai-omer:
@SteveMacenski Also added a PR for humble branch #3562

PR #3559, opened 2023-05-07
Fix overridden parameter values

nakai-omer:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3558 |

---

* Fixed a recurring bug where parameter values were ignored, and overridden by trying to set default values

N/A

#### For Maintainers: 


SteveMacenski (approved):
Thank you, this makes sense!

PR #3564, opened 2023-05-08
Add ros2/geometry2 to underlay

ruffsl:
Add ros2/geometry2 to underlay in interim until next rolling sync.

Related:

- https://github.com/ros2/geometry2/pull/601


ruffsl:
System tests look a lot better now.

Issue #3563, opened 2023-05-08
Feature Request 


nakai-omer:

nakai-omer:
@SteveMacenski A glitch while writing the feature request, I have updated it with the details, can you please reopen

SteveMacenski:
See the planner selector (or other selector) BT nodes. You can change them at runtime however you like. That "hardcode" is just the default BT. You can make that a BT blackboard variable like `{path}` which can pass the planner any ID you like

PR #3565, opened 2023-05-09
Easily enable dynamic planners

nakai-omer:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | TBD |
| Primary OS tested on | (Ubuntu 22.04 |
| Robotic platform tested on | Custom robot running Ros2 Humble |

---

- Add the ability to select the planner which will run the navigate to pose, or navigate through poses, action. 
- Backward compatible by defaulting to GridBased planner in case no planner is passed.

Default planner id configuration has been added to navigate_to_pose.cpp and navigate_through_poses.cpp, but I couldn't find an existing documentation to add this to. 

---

#### For Maintainers: 


SteveMacenski:
This looks like an extension of the exchange in https://github.com/ros-planning/navigation2/issues/3563#issuecomment-1540343412. I don't think this is the right answer, necessarily.

nakai-omer:
@SteveMacenski Thanks for the detailed answer. 
> This would not allow you to have dynamically changing planners, just 1 planner per navigation request.
IMHO, the definition of dynamic here is subjective to why you need to change planners. If it is because a planner is not good enough or fails, I can understand your point. But for us, we need to change planners in order to perform different tasks, just like the docs about the planners in navigation concepts state. Sometimes, we need to perform coverage path, sometimes we need to find shortest path back to charger, and sometimes we want a path in order to perform exploration mapping. For this 1 planner per request is perfect. (I believe this answers your second bullet as well)

> Your navigation request defines the BT to use, which could have the hardcoded planner ID to use alongside the navigation behavior of interest.
Maybe I misunderstood you, but our navigation request only defined the planner id to use, which is no an internal implementation details as the user most likely defined it in the nav2 params. If not, there is always the default GridBased planner used, even if no planner id is passed in any of the requests.

Also, this change doesn't contradict using a BT in case you really need to, it just comes to bring a simpler API in case you need to control which planner to use for your nav task.

SteveMacenski:
> For this 1 planner per request is perfect.

> Maybe I misunderstood you, but our navigation request only defined the planner id to use, which is no an internal implementation details as the user most likely defined it in the nav2 params. If not, there is always the default GridBased planner used, even if no planner id is passed in any of the requests.

The navigation request include a `behavior_tree` field which you can specify the filepath to the BT you'd like to use. That is configurable as you know 'in order to perform different tasks'. You could simply specify the planner id you'd like to use (e.g. GridBased or otherwise) in the BT file you specify and have different BT files for different tasks involving different planners - which it sounds like you approximately do today. 

From a design perspective, defining the exact planner to use from the highest-level navigation request is breaking encapsulation and separation of concerns. The behavior tree by definition is the behavior you'd like your navigation system to abide by for the request, so it seems very natural if there's a specific and single planner you'd like to use for a given type of request, you should be defining it in that tree, not as a separate field. If you need to dynamically change it throughout the request, I can see a strong case to be made to including the PlannerSelector BT nodes (or similar) to control the planner to use throughout execution.

I believe these are better solutions. However, the navigators themselves are plugins for a reason, you could define your own Navigator plugin type that accepts a new action definition specifying the planner ID to use and placing it on the blackboard. But I would likely encourage you to take my advice on the BTs instead, it could save you alot of headache down the line.

nakai-omer:
@SteveMacenski Got it, closing this. Thanks for the quick and detailed responses. If you want to change the behavior use the behavior tree :)

PR #3571, opened 2023-05-11
Docking (1): Add dock message and IsDocked condition BT

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Clara robot |

---

* Added a new message with the state of the dock.
* Added a new BT condition node the check that state.

* Add to index, migration, etc.

---

#### For Maintainers: 


SteveMacenski:
Hi,

No issues with this PR in particular on the code quality front (good job!) but I think its premature as we need to discuss what kinds of messages we need and what kinds of information might be stored in them before we can talk about adding in interfaces. 

I think we should chat a bit more in the ticket and come back to this work on the BT and interface once we have established what we want to do and have a reference implementation we're also designing for. For now, I would close this PR just to keep the list of PRs actionable since this is a draft of a draft for the BT ports and msg\action interfaces. Keep the branch around, it may well be easy once we come to a conclusion on the interface definitions to quickly modify this for use!

Issue #3590, opened 2023-06-01
Dynamic polygons in Collision Monitor are not being corrected for different frames


AlexeyMerzlyakov:
When we are using dynamic polygon, read from input `PolygonStamped` topic once, the Collision Monitor won't update its polygon correctly. During the first read, it converts all vertices from input `PolygonStamped` frame to the `base_frame_id_`, but later doesn't do it anymore.

This leads the CM to behave incorrectly, e.g. for the case when action polygon is defined in the global `map` frame (not in `base_frame`):
![alternative_usage_p1](https://github.com/ros-planning/navigation2/assets/60094858/2cacb7f2-588e-45e1-9b91-fd3518d1bf29)

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - ROS2 rolling built from sources (20230110)
- Version or commit hash:
  - be597dcf
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
* Enable Collision Monitor with dynamic polygon read from input `polygon_sub_topic`
* Send `PolygonStamped` message once, e.g.
```
ros2 topic pub /area_stop geometry_msgs/msg/PolygonStamped "{header: {frame_id: "map"}, polygon: {points: [{x: 0.0, y: 0.5}, {x: -3.0, y: 0.5}, {x: -3.0, y: -5.5}, {x: 0.0, y: -5.5}]}}"
```
* Move robot and ensure that polygon in global frame is moving along with robot (robot suddenly stops w/o any reason)

#### Expected behavior
Collision Monitor corrects the `poly_` vertices to the latest TF state

#### Actual behavior
Collision Monitor converts only once the `poly_` vertices to the `base_frame`, but does not correct them during the further operation

PR #3591, opened 2023-06-01
Fix dynamic polygons vertices not updated for different frame

AlexeyMerzlyakov:
This is the fix for #3590 issue: Collision Monitor to correct polygon vertices to the actual TF state, for the case when this polygon was obtained from another than `base_link` frame. 

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3590|
| Robotic platform tested on | TB3 simulation, `colcon test --packages-select nav2_collision_monitor`, code coverage |

---

Fix dynamic polygons vertices not updated for different frame
* In `updatePolygon()`, correct `poly_` vertice coordinates to the latest frame state, when polygon is published in another (than `base_frame_id_`) frame
* Re-use `geometry_msgs::msg::PolygonStamped polygon_` to preserve latest dynamic polygon message for this correction
* Call `updatePolygon()` for all cases, but not only for the footprint approaches in the main `CollisionMonitorNode` loop
* Added the testcase on the written in the issue situation

No documentation updates is required for this bugfix.

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/collision_monitor_node.cpp
+++ b/nav2_collision_monitor/src/collision_monitor_node.cpp
@@ -379,6 +379,9 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
       break;
     }
 
+    // Update polygon coordinates, when it necessary
+    polygon->updatePolygon();
+
     const ActionType at = polygon->getActionType();
     if (at == STOP || at == SLOWDOWN || at == LIMIT) {
       // Process STOP/SLOWDOWN for the selected polygon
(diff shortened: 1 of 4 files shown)


nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -379,6 +379,9 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
       break;
     }
 
+    // Update polygon coordinates, when it necessary
```
SteveMacenski, line 382:
```suggestion
    // Update polygon coordinates
```

SteveMacenski (approved):

SteveMacenski:
Great, looks good to me!

PR #3600, opened 2023-06-06
[MPPI] Enforce inversion MOD (NO TEST)

doisyg:
---

## Basic Info

| ------ | ----------- |

---

THIS IS A DRAFT
TESTS ARE DISABLED
Tested and used in production for a couple of months but doesn't necessarily work with all use cases.
Needs https://github.com/ros-planning/navigation2/pull/3454 to work properly.

This PR adds a mechanism to MPPI to enforce global path inversions: the part of the global path taken into account by the controller will be cropped to the closest orientation inversion found on the path until the robot is within the tolerances of the inversion pose.
This new behavior of the controller is controlled by the following new parameters:
```
      enforce_inversion (bool)
      inversion_xy_tolerance (double)
      inversion_yaw_tolerance (double)
```

Default behavior:

https://github.com/ros-planning/navigation2/assets/15727892/0f87d27b-55cd-43fc-8d1f-4a6db6680c33

With path inversion:

https://github.com/ros-planning/navigation2/assets/15727892/5fb50ade-e320-49ac-9947-d7a202acca1f

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/CMakeLists.txt
+++ b/nav2_mppi_controller/CMakeLists.txt
@@ -89,14 +89,14 @@ install(DIRECTORY include/
   DESTINATION include/
 )
 
-if(BUILD_TESTING)
-  find_package(ament_lint_auto REQUIRED)
-  find_package(ament_cmake_gtest REQUIRED)
-  set(ament_cmake_copyright_FOUND TRUE)
-  ament_lint_auto_find_test_dependencies()
-  add_subdirectory(test)
-  # add_subdirectory(benchmark)
-endif()
+#if(BUILD_TESTING)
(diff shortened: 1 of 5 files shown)


SteveMacenski:
Note to self for tomorrow; consider if we can handle this better in the critics themselves rather than in the path handler so we can process more context all at once rather than only dishing out a little at a time (which causes slow-at-goal problem to appear mid-use). This works and is probably the easiest way to go, but if we can instead have the path-related critics look at the cusp points / orientations for inverting, that would be great. That would allow us to deviate from them a little bit where optimal, but generally still do the directional changes where explicitly requested.

SteveMacenski:
Commit to supersede in your inbox shortly for testing.

PR #3602, opened 2023-06-06
nlamprian/controller server plugin handling

nlamprian:
## Basic Info

| ------ | ----------- |

---

Take the goal_checker_id as an example. It's never set (at least within this project). This results in a warning always shown. Furthermore, if more than one plugin is configured, you are forced to specify which one you want to use. If you don't, nothing works. The logic here is kinda broken (in one case, an empty id is accepted, in another, it's not).

Ideally, the behavior I would have expected would be on par with the bt filenames. I believe choosing the first available id is the next best thing to having a default id parameter (which could also be done, of course).

The normal behavior doesn't change. The extended behavior could be documented.

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -188,8 +188,11 @@ ControllerServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
     }
   }
 
-  for (size_t i = 0; i != goal_checker_ids_.size(); i++) {
-    goal_checker_ids_concat_ += goal_checker_ids_[i] + std::string(" ");
+  for (size_t i = 0; i < goal_checker_ids_.size(); i++) {
+    goal_checker_ids_concat_ += goal_checker_ids_[i];
+    if (i < goal_checker_ids_.size() - 1) {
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Hi,

I disagree with this change on a philosophical basis and so I'm not going to review it for semantic changes which would simply waste your time. Specifying / deconflicting a request when there are multiple viable options is very important to not produce unexpected behavior which could be dangerous. Some of these controllers might not do active collision avoidance and if mistakenly activated in a crowded setting could cause injuries or damage. This is why we require that users specify which they'd like to use at any given time and provide the BT nodes to easily make those changes inline as they choose. 

The warning that prints once if you don't specify I agree is somewhat annoying, but since there are no other options to need to deconflict, its not worth outright rejecting a request since there's no actual option that could cause a problem. 

If we wanted to remove the note, we could make it a debug log instead so it doesn't print to screen or make the default BTs specify what is already the default controller ID in the controller server's configurations. 

So in summary, I'm declining to review this change on safety and defined behavior grounds. I think selecting just the first specified controller when no other deconfliction information is available is dangerous and unintuitive

Issue #3603, opened 2023-06-07
ObstacleLayer above StaticLayer always overrides unknown parts


RBT22:
#### Feature description
If I'm not mistaken currently there is no easy way to prevent the planner generating a path through areas that are not on a static map if there is an obstacle layer above it. When the obstacle layer updates the costmap's values in `ObstacleLayer::updateCosts` with both combination_method parameter options the unknown parts are overriden with new values.
```
  switch (combination_method_) {
    case 0:  // Overwrite
      updateWithOverwrite(master_grid, min_i, min_j, max_i, max_j);
      break;
    case 1:  // Maximum
      updateWithMax(master_grid, min_i, min_j, max_i, max_j);
      break;
    default:  // Nothing
      break;
  }
```
"If the master value is NO_INFORMATION, it is overwritten." says the docs about CostmapLayer::updateWithMax.

#### Implementation considerations
Our team wrote an ObstacleLayer::updateWithMaxWithoutUnknownOverride method to address this issue. Is this something that could be supported, or there is a better way to achive this behavior?

SteveMacenski:
> If I'm not mistaken currently there is no easy way to prevent the planner generating a path through areas that are not on a static map if there is an obstacle layer above it

Why would you require an obstacle layer before a static layer? I'm trying to understand the situation where that's a semantically different situation that requires you to do that (e.g. versus a misconfiguration).

RBT22:
Thank you for the quick response! Sorry if it was a bit confusing, exactly the last sentence you wrote. 

The supported way of using our robot is to create a map first about the environment you want to use it in, and then you can start the navigation. In this case, it is better to avoid situations where the robot is going through places that are not on the map, even if the current sensor measurements says they are clear.

PR #3612, opened 2023-06-09
Option for ObstacleLayer to not override StaticLayer's unknown parts

RBT22:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [#3603](https://github.com/ros-planning/navigation2/issues/3603) |

---

* I added an updateWithMaxWithoutUnknownOverride method to the CostmapLayer
* I added an value option to the ObstacleLayer's  combination_method parameter that uses the previous method.

* Added new parameter value, so need to add that to the documentation page
* 

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -92,10 +92,14 @@ void ObstacleLayer::onInitialize()
   node->get_parameter(name_ + "." + "footprint_clearing_enabled", footprint_clearing_enabled_);
   node->get_parameter(name_ + "." + "min_obstacle_height", min_obstacle_height_);
   node->get_parameter(name_ + "." + "max_obstacle_height", max_obstacle_height_);
-  node->get_parameter(name_ + "." + "combination_method", combination_method_);
   node->get_parameter("track_unknown_space", track_unknown_space);
   node->get_parameter("transform_tolerance", transform_tolerance);
   node->get_parameter(name_ + "." + "observation_sources", topics_string);
+
+  int combination_method_param{};
+  node->get_parameter(name_ + "." + "combination_method", combination_method_param);
+  combination_method_ = combination_method_from_int(combination_method_param,
+    "ObstacleLayer::onInitialize()");
 
   dyn_params_handler_ = node->add_on_set_parameters_callback(
     std::bind(
@@ -311,7 +315,8 @@ ObstacleLayer::dynamicParametersCallback(
       }
     } else if (param_type == ParameterType::PARAMETER_INTEGER) {
       if (param_name == name_ + "." + "combination_method") {
-        combination_method_ = parameter.as_int();
+        combination_method_ = combination_method_from_int(parameter.as_int(), "ObstacleLayer::"
+          "dynamicParametersCallback()");
       }
     }
   }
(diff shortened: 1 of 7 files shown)


AlexeyMerzlyakov (commented):
In general, good idea and approach.
Please follow the review items from below, and also add the testcase(s) covering new functionality. + Documentation updates on https://navigation.ros.org/configuration/packages/costmap-plugins/obstacle.html parameters page are necessary for this change.

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -546,6 +546,9 @@ ObstacleLayer::updateCosts(
     case 1:  // Maximum
       updateWithMax(master_grid, min_i, min_j, max_i, max_j);
       break;
+    case 2: // Maximum without overwrite if unknown
```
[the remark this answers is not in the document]
RBT22: I made some updates. Is this a good way to handle the conversions?

SteveMacenski:
Your PR has some linting issues, make sure to run ament_uncrustify and ament_cpplint https://app.circleci.com/pipelines/github/ros-planning/navigation2/9493/workflows/a75513af-b769-408c-ad12-b576c2a60575/jobs/30458/tests

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -760,4 +768,22 @@ ObstacleLayer::resetBuffersLastUpdated()
   }
 }
 
+CombinationMethod combination_method_from_int (const int value, const std::string function_name){
```
SteveMacenski, line 771: Remove function name, that's a pretty odd choice. Also, use the node's logger, don't create your own

SteveMacenski (changes requested):

RBT22:
Thanks for the feedback! I tried to fix the problems you wrote.
Could you help me what kind of test is required for this functionality?

SteveMacenski:
Looks good to me now. I think there's a need for documentation on the new enum type in the configuration guides as Alexey pointed out. 

>  Could you help me what kind of test is required for this functionality?

First, I'd just test locally on your robots to make sure it works in general. Then, add a new test to the `integration/obstacle_tests.cpp` file for setting the different combination method with a costmap with unknown background costs and making sure even if you insert information that its never updated on `updateCosts`. this probably involves making a new test fixture, so you could also add a new file to `unit` tests to do this as well, but that other file might be a good reference for you!

RBT22:
I'm not sure if I understood correctly what you meant, but I uploaded an integration test. @SteveMacenski 
Also started the docs. :)

nav2_costmap_2d/test/integration/obstacle_tests.cpp
```diff
@@ -554,3 +554,50 @@ TEST_F(TestNode, testDynParamsSetStatic)
...
+  ~TestNodeWithoutUnknownOverwrite() {}
+
+protected:
+  std::shared_ptr<TestLifecycleNode> node_;
+};
+
+/**
+ * Test CombinationMethod::MaxWithoutUnknownOverwrite in ObstacleLayer.
+ */
+TEST_F(TestNodeWithoutUnknownOverwrite, testMaxWithoutUnknownOverwriteCombinationMethod) {
```
SteveMacenski, line 583: Can you show that these same inputs with MaxOverwrite does change the unknown count?

SteveMacenski:
Just that and otherwise looks good to me! It would be good to test the counter-case

SteveMacenski (approved):

PR #3624, opened 2023-06-12
updating mppi's path angle critic for optional bidirectionality

SteveMacenski:
@doisyg please review

This adapts https://github.com/ros-planning/navigation2/pull/3454 for optionality and not using the path angles -- which might not be populated if not using the smac planners and/or including one of the optional Path Smoothers.

First step in path inversion work this week.

This method includes a check on if reversing is possible at all at the controller level. If not, uses only the forward values. It also has a new parameter (docs PR to match to be linked) if forward is preferred. That way, we consider using the inverted angle schema **only** when reversing is possible and that there genuinely is no preference one way or another. Otherwise, if there is a preference, we do penalize to encourage rotating to go forward.   

I also removed the use of the path point's orientations with the trajectory orientations. **Please check me on this.**

Also added a unit test for the utils function :-)

### DIFF
--- a/nav2_mppi_controller/README.md
+++ b/nav2_mppi_controller/README.md
@@ -112,7 +112,6 @@ This process is then repeated a number of times and returns a converged solution
  | trajectory_point_step      | double | Default 4. Step of trajectory points to evaluate for path distance to reduce compute time. Between 1-10 is typically reasonable.   |
  | max_path_occupancy_ratio   | double | Default 0.07 (7%). Maximum proportion of the path that can be occupied before this critic is not considered to allow the obstacle and path follow critics to avoid obstacles while following the path's intent in presence of dynamic objects in the scene.  |
 
-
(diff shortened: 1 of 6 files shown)


doisyg:
Looks good to me and I tested with `forward_preference` set to `true` and `false` (on a system with over-weighted `PathAngleCritic`) and could observe the same improvement/fix as highlighted in the videos of the previous PR description (https://github.com/ros-planning/navigation2/pull/3454)

I am not sure I understand the modications of the `ConstraintCritic` though.

SteveMacenski:
> getParentParam(vy_min, "vy_min", 0.0);

`vy_min` doesn't exist :-) If you can move in the Y direction, that's left and right. That doesn't make sense to have different max velocities based on left or right side. Its actually not a parameter at all so that was just a bug I found when I was searching for the `parent` code to steal and use here!

Issue #3626, opened 2023-06-14
Error_code parameters were not set warning message


RBT22:
## Bug report

**Required Info:**

- Operating System: 
  - Ubuntu 22.04
- ROS2 Version: 
  - Rolling
- DDS implementation: 
  - CycloneDDS

#### Steps to reproduce issue

Simply run the getting started's bringup code:

`ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False`

#### Expected behavior
There should be no warning message.

#### Actual behavior

BtActionServer logs the following warning message, even when the `error_code_names` parameter is set.
```
Error_code parameters were not set. Using default values of: follow_path_error_code
[component_container_isolated-6] compute_path_error_code
[component_container_isolated-6] Make sure these match your BT and there are not other sources of error codes you want reported to your application
```
#### Additional information

In `nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp ` the warning logging happens because the parameter is not declared.

```
if (!node->has_parameter("error_code_names")) {
    std::string error_codes_str;
    for (const auto & error_code : error_code_names) {
      error_codes_str += error_code + "\n";
    }
    RCLCPP_WARN_STREAM(
      logger_, "Error_code parameters were not set. Using default values of: "
        << error_codes_str
        << "Make sure these match your BT and there are not other sources of error codes you want "
        "reported to your application");
    node->declare_parameter("error_code_names", error_code_names);
  }
```

A workaround could be to declare the variable without default value first and check if it is set:

```
if (!node->has_parameter("error_code_names")) {
    const rclcpp::ParameterValue value = node->declare_parameter(
      "error_code_names",
      rclcpp::PARAMETER_STRING_ARRAY);

    if (value.get_type() == rclcpp::PARAMETER_NOT_SET) {
      node->set_parameter(rclcpp::Parameter("error_code_names", error_code_names));
      std::string error_codes_str;
      for (const auto & error_code : error_code_names) {
        error_codes_str += error_code + "\n";
      }
      RCLCPP_WARN_STREAM(
        logger_, "Error_code parameters were not set. Using default values of: "
          << error_codes_str
          << "Make sure these match your BT and there are not other sources of error codes you want "
          "reported to your application");
    }
}
```

doisyg:
I was about to report the same. I think all the `has_parameter` usage should be rethink if there is no chance that a `declare_parameter` was used before (it will always fail).
[...]

PR #3631, opened 2023-06-16
correct error message

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/3626 |

---

Used suggested solution to only print warning message if parameter was given a default value. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -16,8 +16,6 @@
 #define NAV2_BEHAVIOR_TREE__BT_ACTION_SERVER_IMPL_HPP_
 
 #include <memory>
-#include <rclcpp/parameter.hpp>
-#include <rclcpp/parameter_value.hpp>
 #include <string>
 #include <fstream>
 #include <set>
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -75,11 +75,11 @@ BtActionServer<ActionT>::BtActionServer(
     if (value.get_type() == rclcpp::PARAMETER_NOT_SET) {
       std::string error_codes_str;
       for (const auto & error_code : error_code_names) {
-        error_codes_str += error_code + "\n";
+        error_codes_str += "\n" + error_code;
       }
       RCLCPP_WARN_STREAM(
           logger_, "Error_code parameters were not set. Using default values of: "
-          << error_codes_str
+          << error_codes_str + "\n"
           << "Make sure these match your BT and there are not other sources of error codes you"
           "reported to your application");
     node->declare_parameter("error_code_names", error_code_names);
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -16,6 +16,7 @@
 #define NAV2_BEHAVIOR_TREE__BT_ACTION_SERVER_IMPL_HPP_
 
 #include <memory>
+#include <rclcpp/parameter_value.hpp>
 #include <string>
 #include <fstream>
 #include <set>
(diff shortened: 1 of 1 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -67,16 +68,22 @@ BtActionServer<ActionT>::BtActionServer(
   };
 
   if (!node->has_parameter("error_code_names")) {
```
SteveMacenski, line 70: This has parameter will always be false if its not been declared, so it can be removed.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -67,16 +68,22 @@ BtActionServer<ActionT>::BtActionServer(
...
-    RCLCPP_WARN_STREAM(
-      logger_, "Error_code parameters were not set. Using default values of: "
-        << error_codes_str
-        << "Make sure these match your BT and there are not other sources of error codes you want "
-        "reported to your application");
+
+    const rclcpp::ParameterValue value = node->declare_parameter(
+      "error_code_names", 
+      rclcpp::PARAMETER_STRING_ARRAY);
+    if (value.get_type() == rclcpp::PARAMETER_NOT_SET) {
```
SteveMacenski, line 75: Did you test this works? This isn't how I've seen it implemented before. It might work fine, but worth testing.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -67,16 +68,22 @@ BtActionServer<ActionT>::BtActionServer(
...
+      std::string error_codes_str;
+      for (const auto & error_code : error_code_names) {
+        error_codes_str += error_code + "\n";
+      }
+      RCLCPP_WARN_STREAM(
+          logger_, "Error_code parameters were not set. Using default values of: "
+          << error_codes_str
+          << "Make sure these match your BT and there are not other sources of error codes you"
+          "reported to your application");
     node->declare_parameter("error_code_names", error_code_names);
```
SteveMacenski, line 85: This redeclared the parameter that was already declared... 

I think it should be setting

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
SteveMacenski: Its indentation also looks wrong, so not sure if its also an issue of being in the wrong conditional

RBT22: Maybe this way, without redeclaration?

```
if (!node->has_parameter("error_code_names")) {
  node->declare_parameter(
    "error_code_names",
    rclcpp::PARAMETER_STRING_ARRAY);
}

rclcpp::Parameter tmp;
if (!node->get_parameter("error_code_names", tmp)) {
  node->set_parameter(rclcpp::Parameter("error_code_names", error_code_names));
  
  std::vector<std::string> error_code_names = {
  "follow_path_error_code",
  "compute_path_error_code"
  };

  std::string error_codes_str;
  for (const auto & error_code : error_code_names) {
    error_codes_str += error_code + "\n";
  }
  RCLCPP_WARN_STREAM(
    logger_, "Error_code parameters were not set. Using default values of: "
      << error_codes_str
      << "Make sure these match your BT and there are not other sources of error codes you want "
      "reported to your application");
}
```

SteveMacenski: That wouldn't compile for a few reasons (e.g. setting error code names before its defined) but yes, that's generally speaking what I think would work well.

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
jwallace42: Yep, 

I tested with no parameter and then set the param. Is there a better way of doing this? I couldn't find much.

SteveMacenski: This works, typically I've just seen declaring and then using a `get` command to either have the return value be `false` that it couldn't be received (as the user shows below) or use the other `get` API which returns the value checking if it throws an exception.

I don't care one way or another though, if this works I'm perfectly happy with that

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
jwallace42: This needs to be here when the parameter in defined in the yaml. Otherwise there is a declaration error. 

`[component_container_isolated-5] [ERROR] [1687029994.187670094] []: Original error: parameter 'error_code_names' has already been declared`

SteveMacenski (approved):

RBT22:
Do you plan to backport this to Humble?

SteveMacenski:
Next sync yes

Issue #2997, opened 2022-06-02
Migrate to Ignition / TB4 simulation


SteveMacenski:
We want to use the time in Humble to migrate to Ignition for Nav2 moving forward. 

- [x] Migrate AWS warehouse world to Ignition
- [x] Provide TB4 simulation launch file (`tb4_simulation_launch.py`) for Ignition analog to TB3 for Gazebo using whatever standard launch file provided by TB4 (I believe its a warehouse)
- [x] Update simple commander examples to using Ignition with the AWS warehouse world migration
- [ ] Migrate system tests to using ignition with AWS warehouse world after https://github.com/ros-planning/navigation2/pull/2797 is merged
- [ ] Consider updating TB4 simulation launch examples to using the AWS warehouse world that the system tests and simple commander API uses. Or migrate all a different warehouse. Doesn't have to all be the same, but consider it.

padhupradheep:
Additionally there are some more suggestions:

- would be also nice to see an Omnidirectional robot part of the stack (will definitely be useful for people who wants to get started with MPPIC ). 
- On my initial experience with the Ignition, I see that, Ignition messages already have namespaces in them, so it would be nice to remap the same to the ROS messages.

vinnnyr:
I found a robot model that has a public (Ignition) Gazebo sim: https://github.com/iRobotEducation/create3_sim 
Isn't Create3 what TB4 is based off of?

SteveMacenski:
A bit too is having the binaries released in rolling and such so that we can use it in CI without an underlay space (and so users can have a nice out-of-the-box experience). I think there were things before, but not quite ready and binaries not released yet.

borongyuan:
Hi, I have migrated AWS Warehouse World to Ignition Gazebo.
https://github.com/borongyuan/aws-robomaker-small-warehouse-world
Currently it's just an empty world and lacks testing. Currently Ignition Gazebo already supports ros2_control, so there should be no problem integrating TB4 simulation. But there is still a lot of work to be done.

SteveMacenski:
Thanks! That's a great step forward though! We were just talking about this at the working group meeting yesterday.

ahcorde:
Hi, 

I'm planning to contribute with the new Gazebo simulation. But it's not clear for me where I should start and I have some questions:

 - Do you want to get rid of Gazebo-classic ? 
 - Are you thinking about living with both simulator at the same time in the repository ?
 - is there any roadmap or issue for this ?
[...]

PR #3637, opened 2023-06-19
Added nav2 simple commander with the new Gazebo

ahcorde:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/2997 |

---

I created a new package called `nav2_gz_simple_commander` to replicate the package `nav2_simple_commander` but this time with the new Gazebo. 

I have some issues that maybe you can help me to solve. The `odom` frame should be locate close to the `map` frame (compared with the gazebo-classic simulation), but for some reason is located where the robots is spawned and sometimes the robot is lost. You can check this in the gif. any thoughts?

NOTE: You need to add this repository to your workspace https://github.com/ahcorde/aws-robomaker-small-warehouse-world

## Test it

```bash
ros2 launch nav2_gz_simple_commander nav_to_pose_example_launch.py
```

nav2_lost
nav2_lost2


SteveMacenski:
The only differences are the launch files, no? If so, delete this package and just add the complimentary launch files.

ahcorde:
> The only differences are the launch files, no? If so, delete this package and just add the complimentary launch files to the existing package. I don't think there's any reason to have an entirely separate package.

Done here https://github.com/ros-planning/navigation2/pull/3637/commits/6e708e9c5e2aad4ccffdbd1190d76f4a6a0390cb

Any thoughts about why the robot is lost ?

SteveMacenski:
Probably some of the URDF things I pointed out in the other PR. A 5m lidar is completely insufficient to be localizing in a warehouse of that size.

Otherwise, your changes look good if you make the updates for the rest of these, but I'd recommend just waiting until we get the other PR done since this is blocked by that anyway

ahcorde:
> Probably some of the URDF things I pointed out in the other PR. A 5m lidar is completely insufficient to be localizing in a warehouse of that size.
> 
> Otherwise, your changes look good if you make the updates for the rest of these, but I'd recommend just waiting until we get the other PR done since this is blocked by that anyway

I noticed that in this case the [warehouse.world](https://github.com/ros-planning/navigation2/blob/main/nav2_simple_commander/launch/warehouse.world) file contains a different instance of the [turtlebot3](https://github.com/ros-planning/navigation2/blob/main/nav2_simple_commander/launch/warehouse.world#L243-L680) which is defined inside the file. In particular the model working on Gazebo classic has a [3.5m lidar](https://github.com/ros-planning/navigation2/blob/main/nav2_simple_commander/launch/warehouse.world#L387-L391).

the only different that I have between models are the type caster wheel, one is `ball` and in the other one if `fixed`.

SteveMacenski:
I can't answer that for you then, maybe the Ignition model has problems?

SteveMacenski:
Figure out the reason and/or run each of these launch files to verify they work?

ahcorde:
> Figure out the reason and/or run each of these launch files to verify they work?
> 
> This PR I'll broadly just give a once over and merge as long as someone's run them and the demos still work

I checked all the new launch files and they have the same behaviour as Gazebo classic.

SteveMacenski:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/9552/workflows/684b5f23-95df-44ff-85f6-20c9897cc0e9/jobs/30626/tests

Some linting problems

SteveMacenski:
Approved beyond the minor linting problems needed for CI to not have errors

ahcorde:
humm the CI failures look odd, I can see two tests failing in `nav2_gz_simple_commander.test.test_copyright` but this package was removed, same happens with `nav2_gz_simple_commander.test.test_flake8`. 

should I need to fix linters that not included in the files that I included ?

SteveMacenski:
Might be a caching issue, change all 3 instances of `v14` to `v15` in this file and push it: https://github.com/ros-planning/navigation2/blob/main/.circleci/config.yml

SteveMacenski:
That worked

nav2_simple_commander/launch/gz_nav_through_poses_example_launch.py
```diff
@@ -0,0 +1,98 @@
...
+        default_value='True',
+        description='Whether to start RVIZ')
+
+    declare_simulator_cmd = DeclareLaunchArgument(
+        'headless',
+        default_value='False',
+        description='Whether to execute gzclient)')
+
+    # start the simulation
+    start_gazebo_server_cmd = ExecuteProcess(
```
SteveMacenski, line 51: I just noticed this - this appears to be the old server, no?

ahcorde: Done https://github.com/ros-planning/navigation2/pull/3637/commits/55bd8bf3dc186f5971b76b5e7b80ac8e1f425c00

nav2_simple_commander/launch/gz_security_demo_launch.py
```diff
@@ -0,0 +1,206 @@
...
+
+    os.environ['IGN_GAZEBO_RESOURCE_PATH'] = os.path.join(
+        get_package_share_directory('turtlebot3_gazebo'), 'models')
+    os.environ['IGN_GAZEBO_RESOURCE_PATH'] += ':' + os.path.join(
+        get_package_share_directory('aws_robomaker_small_warehouse_world'), 'models')
+    os.environ['IGN_GAZEBO_RESOURCE_PATH'] += ':' + os.path.join(
+        get_package_share_directory('turtlebot3_gazebo'), '..')
+
+    # Clock bridge
+    clock_bridge = Node(
```
SteveMacenski, line 119: It just occurred to me that we abstracted out the gz tb3 robot simulation launch file to deal with the gz specifics / bridges and whatnot. Can't that be used in all these launch files as well if we send it the updated starting pose?

ahcorde: I made the PRs independent, because the tb3 for Gazebo classic is defined in the world file inside this package. We are using a two versions of the tb3, which are a little bit different between them.

But I'm happy to use the new resources included in `nav2_bringup`.  What fo you prefer ?

 - Keep models independent from nav2_bringup 
 - Use the same models here and in nav2_bringup

SteveMacenski: Same - unless there are key differences... which I don't believe there should be? I _think_ when I made this I may have made the lidar for the simple commander range much higher to deal with the larger nature of the space that would need to be kept. 

So I'd say I'd rather re-use as much as possible. Where there are changes where they're not the same, err on the side of liberal use of resources (e.g. longer ranges and such)

SteveMacenski:
https://github.com/ros-navigation/navigation2/pull/4417 supersedes based on https://github.com/ros-navigation/nav2_minimal_turtlebot_simulation

Issue #3633, opened 2023-06-16
error_code_names parameter for BT Navigator plugins


RBT22:
#### Feature description

It would be nice to have the option to give different `error_code_names` parameters to BT Navigator plugins. Currently you can pass only one list to the parent node and that is used in every navigator. This new customization option would help to handle structually more different navigators.

#### Implementation considerations

In the config file the `error_code_name` parameter for the navigator could be placed near the plugin:
```
bt_navigator:
  ros__parameters:
    navigators: ["navigate_to_pose", "navigate_through_poses"]
    navigate_to_pose:
      plugin: "nav2_bt_navigator/NavigateToPoseNavigator"
      error_code_names:
        - compute_path_error_code
        - follow_path_error_code
```
We can reach the parent node's parameters in the BtActionServer implementation, so my proposal would be to use these new values as default and fallback on the parent node's values, and if those are not set also, we could use the default.

This way the parameter can be declared as "<navigator_name>.error_code_names". One question is how to get the first part of this string in the BtActionServer. We have `action_name`, which is "task/<navigator_name>" so technically we could take the second part of it, but maybe there is a better way.

SteveMacenski:
Is there a reason not to include them all? If there are some of the error code IDs which are not used in a particular navigation task, that wouldn't cause any issues as they'd continue to be zeros https://github.com/ros-planning/navigation2/blob/01b84136493a47c2fe6ffa9d0b1132cbf132b189/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp#L272-L291

Also, for a given NavigateToPose navigation task, there could be many different BTs that could be sent to use for the task, each of which could have their own unique cross section of error codes. I don't think specifying on a per-navigator basis makes any sense. It would either need to be global or on a per-action basis and needs to be somehow parsed from the BT itself (perhaps metadata?).

Though, I don't think that level of optimization benefits anyone really given the overhead of having `N` error code IDs checked on the blackboard is eq. to an unordered map `O(1)` lookup, there's not much computational reason to do so. 

So I'd ask why you want this feature and what it accomplishes for you?

RBT22:
Yeah if we can change the error message to debug, that is also fine by me. :D

Although one reason for using specialized parameters could be exactly the ability to keep the ERROR level messages. Logging the missing value as an error could be a good feature. Using one big list for the parent and logging the differences all the time might be confusing and not that helpful for debugging, but it's not that big of a deal.

[...]

PR #3642, opened 2023-06-21
Let Navigators have different error codes

RBT22:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [#3633](https://github.com/ros-planning/navigation2/issues/3633) |

---

* Change "Failed to get error code: %s from blackboard" log from ERROR to DEBUG
* Add INFO log on initialization about which errors codes are being used to help debug

* 

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -285,7 +285,7 @@ void BtActionServer<ActionT>::populateErrorCode(
         highest_priority_error_code = current_error_code;
       }
     } catch (...) {
-      RCLCPP_ERROR(
+      RCLCPP_DEBUG(
         logger_,
         "Failed to get error code: %s from blackboard",
         error_code.c_str());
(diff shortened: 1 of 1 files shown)


RBT22:
Is there a reason why we use \n between the error codes, instead of spaces @SteveMacenski?  It looks harder to grep this way.

SteveMacenski:
No, I missed that on my scan through, please replace with spaces

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -82,6 +82,15 @@ BtActionServer<ActionT>::BtActionServer(
...
       node->set_parameter(error_code_names_param);
+    } else {
+      error_code_names = value.get<std::vector<std::string>>();
+      std::string error_codes_str;
+      for (const auto & error_code : error_code_names) {
+        error_codes_str += "\n" + error_code;
+      }
+      RCLCPP_INFO_STREAM(
+        logger_, "Error_code parameters were set to: "
+          << error_codes_str);
```
SteveMacenski, line 93: Can this not fit on the next line up?

SteveMacenski (commented):
Generally LGTM

SteveMacenski (approved):

SteveMacenski:
Thanks! Looks great!

jwallace42:
I am kind of confused on how this lets navigators have different error codes? All navigators would still be sharing the same set of error codes.

SteveMacenski:
It doesn't, the PR was not well named but named along with the ticket. This is just to silence warnings mostly

RBT22:
Sorry if the naming was not perfect.

> It appears that all navigators would still be sharing the same set of error codes.

I mean the checklist is the same for all of them, but now there is a supported way to propagate different error codes from different Navigators. It wasn't the case before.

PR #3650, opened 2023-06-23
Enable multiple MPPI path angle modes depending on preferences in behavior

SteveMacenski:
CC @doisyg 

Complimentary docs PR: https://github.com/ros-planning/navigation.ros.org/pull/439

- [x] needs tests for the other 2 operating modes
- [x] Needs testing from @doisyg to make sure it meets his specs


doisyg:
Very nice, tested all modes, all quadrants and all orientations. 
Behave as expected!

SteveMacenski:
Just waiting on CI + checking code cov to make sure everything is properly covered.

PR #3652, opened 2023-06-24
Fix CERES failing build from compilation warnings

Ryanf55:
## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu 23 with GCC12 |

---

* Ceres has compiler warnings that fail nav2 build with werr
* This treats them as SYSTEM includes so we can ignore them

#### For Maintainers: 

### DIFF
--- a/nav2_constrained_smoother/CMakeLists.txt
+++ b/nav2_constrained_smoother/CMakeLists.txt
@@ -27,7 +27,6 @@ set(library_name nav2_constrained_smoother)
 
 include_directories(
   include
-  ${CERES_INCLUDES}
 )
 
 set(dependencies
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

Ryanf55:
Better way, according to my stack overflow answer: https://stackoverflow.com/questions/67556313/how-to-link-3rd-party-library-ceres-solver-in-cmake/76544373#76544373

SteveMacenski:
Failed to compile 😟

SteveMacenski:
Maybe try rebasing? I've retriggered CI multiple times and there are tests that should not be failing. One has to do with the smoother server which I need to be green before merging this since it impacts that system.

Ryanf55:
> Maybe try rebasing? I've retriggered CI multiple times and there are tests that should not be failing. One has to do with the smoother server which I need to be green before merging this since it impacts that system - just in case.

I did a rebase, CI is still failing on tests.

Issue #3574, opened 2023-05-15
RemovePassedGoals Plugin frame parameter inheritance


HarunTeper:
## Bug report

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Galactic
- Version or commit hash:
  - 2de3f92c0f476de4bda21d1fc5268657b499b258
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
git clone https://github.com/HarunTeper/AuNa
cd AuNa
git checkout 24c114b3aaf0cd6da65316245e08b98669ca948b

./auto_setup.sh
colcon build --symlink-install
source install/setup.bash
ros2 launch auna_scenarios scenario_single_robot_arena_navigation.launch.py

2nd Terminal:
source install/setup.bash
ros2 launch auna_mqtt mqtt_waypoint_receiver.launch.py

#### Expected behavior
The robot receives 4 waypoints to use the navigate_through_poses behavior tree to move

#### Actual behavior
The robot starts the recovery behavior and the following error occurs:
[bt_navigator-12] Warning: Invalid frame ID "base_link" passed to canTransform argument source_frame - frame does not exist
[bt_navigator-12]          at line 156 in /tmp/binarydeb/ros-galactic-tf2-0.17.5/src/buffer_core.cpp

#### Additional information

The remove_passed_goals_action.cpp plugin does not natively use the robot_base_frame specified for the bt_navigator. Instead, it uses the frame names specified in the navigate_through_poses_bt_xml file.

## Feature request

Make the BtNavigator plugins automatically use the same parameters as the bt_navigator, specifically the "global_frame" and "robot_base_frame" parameters for this case.

SteveMacenski:
I see 4 instances where we have BT nodes with their own global and robot base frame ports https://github.com/search?q=repo%3Aros-planning%2Fnavigation2%20getInput(%22robot_base_frame%22&type=code.

The intent here is that there's a potential that we may want to perform these actions wrt other frames than those that are used for global navigation. I will admit looking over these 4, it seems unlikely that the base frame would ever change (but the global frame may wrt the distance controller and condition). 

I think what could be reasonable is to change this node from having a default value in the BT port to:
- Check if its set in the port, then use that value
- If not, then the port result should be an empty string, in which case, use the ROS 2 params API to get the BT navigator's version

[...]

PR #3653, opened 2023-06-24
Modified nav2_behavior_tree and nav2_bt_navigator

HarunTeper:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3574 |

---

For navigators in nav2_bt_navigator, added parameter pass through from bt_navigator to bt_navigator/client_node_name for robot_base_frame and global_frame.
For plugins in nav2_behavior_tree, removed default parameters for behavior tree input ports for robot_base_frame and global_frame, and added the following behavior:

Check behavior tree for user-specified values in behavior tree.
If those are not available, check client_node parameters.
If those are not available, use default values.

The parameters global_frame and robot_base frame can now be used by the navigators, instead of the behavior tree specification.

---

This could be extended for other parameters or adapted for more plugins/navigators.
For future plugins, the same procedure as in this commit needs to be applied to enable these functionalities.

#### For Maintainers: 


nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -42,6 +42,13 @@ NavigateThroughPosesNavigator::configure(
 
   path_blackboard_id_ = node->get_parameter("path_blackboard_id").as_string();
 
+  // Get the action server client node
+  rclcpp::Node::SharedPtr client_node = bt_action_server_->getBlackboard()->get<rclcpp::Node::SharedPtr>("node");
```
SteveMacenski, line 46: Where is this node inserted? Would it not be possible to do this around there instead to group blackboard operations together?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/distance_traveled_condition.hpp
```diff
@@ -79,4 +79,4 @@ class DistanceTraveledCondition : public BT::ConditionNode
 
 }  // namespace nav2_behavior_tree
 
-#endif  // NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__DISTANCE_TRAVELED_CONDITION_HPP_
+#endif  // NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__DISTANCE_TRAVELED_CONDITION_HPP_
```
SteveMacenski, line 82: This EOL missing is going to cause a linting error, make sure to check the CI output or run `ament_cpplint` and `ament_uncrustify` locally to fix any such issues

SteveMacenski: Looking over the code changes, there are definitely alot of linting errors, so you should definitelly run those locally to fix them all

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -31,12 +31,44 @@ RemovePassedGoals::RemovePassedGoals(
...
   getInput("radius", viapoint_achieved_radius_);
-
-  getInput("global_frame", global_frame_);
-  getInput("robot_base_frame", robot_base_frame_);
+  
   tf_ = config().blackboard->get<std::shared_ptr<tf2_ros::Buffer>>("tf_buffer");
   auto node = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
   node->get_parameter("transform_tolerance", transform_tolerance_);
+
+  if(!getInput("global_frame", global_frame_)) {
```
SteveMacenski, line 39: I assume these 4 blocks are essentially the same, so I'll review this one but apply to all the other 4 blocks in the other BT nodes.

This is alot of logic. This feels like we might be served by having this be an external function that these all use to abstract out the code 1 time and use `N` times (and make it easily unit testable)

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -31,12 +31,44 @@ RemovePassedGoals::RemovePassedGoals(
...
+    RCLCPP_INFO(node->get_logger(), "Parameter 'global_frame' not provided by behavior tree xml file, trying to get it from ros2 parameter file");
+    if(!node->get_parameter("global_frame", global_frame_)){
```
SteveMacenski, line 41: If you declare it with the default as `node->get_parameter("global_frame").as_string()`, it will never not have a response to `get_parameter` with the default of `node->get_parameter("global_frame")` so this particular log/check is not necessary

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -31,12 +31,44 @@ RemovePassedGoals::RemovePassedGoals(
...
+      RCLCPP_INFO(node->get_logger(), "Parameter 'global_frame' provided by ros2 parameter file, using value '%s'",
+        global_frame_.c_str());
+    }
+  }
+  else{
+    RCLCPP_INFO(node->get_logger(), "Parameter 'global_frame' provided by behavior tree xml file, using value '%s'",
+      global_frame_.c_str());
+  }
+
+  if(!getInput("robot_base_frame", robot_base_frame_)) {
```
SteveMacenski, line 56: This logic for `robot_base_frame` and `global_frame` is the same. So if you make the function, you can also make 1 function that both of these checks share if you make the actual parameter to check on a parameter of the method. E.g. 

```
param_value  deconflictPortAndParamFrame(node, param_name, this)
```

`this` is so that you can call `getInput`

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -31,12 +31,44 @@ RemovePassedGoals::RemovePassedGoals(
...
+      RCLCPP_INFO(node->get_logger(), "Parameter 'robot_base_frame' not provided by ros2 parameter file, using default value '%s'",
+        robot_base_frame_.c_str());
+    }
+    else{
+      RCLCPP_INFO(node->get_logger(), "Parameter 'robot_base_frame' provided by ros2 parameter file, using value '%s'",
+        robot_base_frame_.c_str());
+    }
+  }
+  else{
+    RCLCPP_INFO(node->get_logger(), "Parameter 'robot_base_frame' provided by behavior tree xml file, using value '%s'",
```
SteveMacenski, line 69: You can make these debug logs instead!

SteveMacenski (changes requested):
Good first stab at it!

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
HarunTeper: This node is inserted in the bt_action_server_impl.hpp file in Line 149. I think I could not directly access it, as the client_node_ in bt_action_server.hpp is protected. You would need to add a function to directly access it.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
HarunTeper: Where should this function be inserted? I could create a bt_utils in nav2_utils for example, as I didn't find a class that all bt nodes extend from.

SteveMacenski: Rename this file to bt_utils + add it here: https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_conversions.hpp

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
SteveMacenski: https://github.com/ros-planning/navigation2/blob/78c693a5d29da86d696ee535e303897220be4663/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp#L123

The client node is created at the top of that function - you have direct access to it

HarunTeper: I added the parameters directly in the bt_action_server_impl.hpp.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/distance_traveled_condition.hpp
HarunTeper: Fixed the current linting errors.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
HarunTeper: Renamed and added the function with some template arguments.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
HarunTeper: Fixed.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
HarunTeper: Added this function to bt_utils.hpp

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
HarunTeper: Done.

SteveMacenski:
> What should I do about the test errors, and the error about the missing bt_conversions.hpp code coverage test.

Look into why they're failing, these tests should not be failing if your PR did not introduce breaking changes and/or caused potentially unintended problem. 

W.r.t. adding new tests, we need test coverage for all new features - so the `deconflictPortAndParamFrame` function needs some tests

nav2_behavior_tree/plugins/decorator/distance_controller.cpp
```diff
@@ -35,17 +36,16 @@ DistanceController::DistanceController(
   const BT::NodeConfiguration & conf)
 : BT::DecoratorNode(name, conf),
   distance_(1.0),
-  global_frame_("map"),
-  robot_base_frame_("base_link"),
   first_time_(false)
 {
-  getInput("distance", distance_);
```
SteveMacenski, line 42: Why remove distance?

nav2_behavior_tree/test/plugins/condition/test_goal_reached.cpp
```diff
@@ -45,7 +45,7 @@ class GoalReachedConditionTestFixture : public nav2_behavior_tree::BehaviorTreeT
       R"(
       <root main_tree_to_execute = "MainTree" >
         <BehaviorTree ID="MainTree">
-            <GoalReached goal="{goal}" />
+            <GoalReached goal="{goal}" global_frame="map" robot_base_frame="base_link"/>
```
SteveMacenski, line 48: Should this be strictly required if its in the declared parent parameters?

nav2_bt_navigator/behavior_trees/navigate_through_poses_w_replanning_and_recovery.xml
```diff
@@ -10,7 +10,7 @@
         <RateController hz="0.333">
           <RecoveryNode number_of_retries="1" name="ComputePathThroughPoses">
             <ReactiveSequence>
-              <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.7"/>
+              <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.7" global_frame="map" robot_base_frame="base_link"/>
```
SteveMacenski, line 13: This shouldn't be required, remove

nav2_bt_navigator/behavior_trees/navigate_w_replanning_distance.xml
```diff
@@ -5,7 +5,7 @@
 <root main_tree_to_execute="MainTree">
   <BehaviorTree ID="MainTree">
     <PipelineSequence name="NavigateWithReplanning">
-      <DistanceController distance="1.0">
+      <DistanceController distance="1.0" global_frame="map" robot_base_frame="base_link">
```
SteveMacenski, line 8: Ditto

SteveMacenski (changes requested):
Just a couple of small things!

nav2_behavior_tree/plugins/decorator/distance_controller.cpp
HarunTeper: Mistake on my side, will revert it back.

nav2_behavior_tree/test/plugins/condition/test_goal_reached.cpp
HarunTeper: I think I need to either add it as a ROS2 parameter, or directly in the BT tree for this test to work correctly.

HarunTeper:
The last error occurs because I renamed bt_conversions.hpp to bt_utils hpp, and now the code coverage cannot find bt_conversions anymore. Did I forget to rename it somewhere?

SteveMacenski:
Try incrementing any instances of `v14` to `v15` to break the cache https://github.com/ros-planning/navigation2/blob/main/.circleci/config.yml -- you need to do all 3 of them as the same thing (`v15`)

HarunTeper:
Did that, but now a test didn't generate a results file.

SteveMacenski:
Retriggering

SteveMacenski (approved):

SteveMacenski:
@HarunTeper thanks for your contribution here, this is a really nice update! I appreciate your responsiveness and diligence during this!

PR #3655, opened 2023-06-25
Fix rviz panel node arguments

nlamprian:
## Basic Info

| ------ | ----------- |

---

Fixes this warning

```
[rviz2-1] [WARN 1687712615.161990346 rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```

Node arguments are not parsed correctly when given as a single string.

---

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #3687, opened 2023-07-10
Properly handling resetting of costmap publishers

SteveMacenski:


BriceRenaudeau:
Thanks for the very quick fix! :+1:

PR #3693, opened 2023-07-11
moved error code enumerations to the result of action messages

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

Previously, the error code enumerations were define in the goal section of the action message. This was incorrect because the the error code is a result. This pr moves the error code enumerations for the actions to the result section.  

---

#### For Maintainers: 

### DIFF
--- a/nav2_msgs/action/SmoothPath.action
+++ b/nav2_msgs/action/SmoothPath.action
@@ -1,5 +1,12 @@
-# Error codes
-# Note: The expected priority order of the errors should match the message order
+#goal definition
+nav_msgs/Path path
+string smoother_id
+builtin_interfaces/Duration max_smoothing_duration
+bool check_for_collisions
+---
+#result definition
+#error codes
+#note: The expected priority order of the errors should match the message order
 uint16 NONE=0
 uint16 UNKNOWN=500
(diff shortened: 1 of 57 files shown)


nav2_msgs/action/SmoothPath.action
```diff
@@ -1,5 +1,12 @@
-# Error codes
-# Note: The expected priority order of the errors should match the message order
+#goal definition
```
SteveMacenski, line 1: All of the action definitions need proper capitalization + a space between the `#` and the comment for the changes. I'm not sure how those got that way with just copy pasting :wink:

SteveMacenski (changes requested):
I generally approve, seems like a reasonable change!

nav2_msgs/action/SmoothPath.action
jwallace42: I added a space and capitalized the `error codes` and `note`. The definitions were always lower case. 

i.e `#goal definition`

SteveMacenski (approved):

SteveMacenski:
@jwallace42 can you open a PR for migration guide?

Issue #3699, opened 2023-07-17
velocity smoother deadband velocity broken since 1.1.7


Aposhian:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble binaries
- Version or commit hash:
  - 1.1.8
- DDS implementation:
  - N/A

#### Steps to reproduce issue

```
ros2 run nav2_velocity_smoother velocity_smoother --ros-args -p 'deadband_velocity:=[0.25,0.0,0.0]'
```
In a separate terminal:
```
ros2 lifecycle set /velocity_smoother configure && ros2 lifecycle set /velocity_smoother activate
ros2 topic pub -r 20 /cmd_vel geometry_msgs/msg/Twist '{ linear: { x: 0.5 }}'
```

In a separate terminal:
```
ros2 topic echo /cmd_vel_smoothed --field linear.x
```

#### Expected behavior
It works on 1.1.6. `/cmd_vel_smoothed.linear.x` ramps up to full speed (0.5 in this case).

#### Actual behavior
On 1.1.8. `/cmd_vel_smoothed.linear.x` stays at 0 forever.

#### Additional information
I believe this was broken in 1.1.7 because the `last_cmd_` is set after the deadband velocities are applied. This means that the last_cmd is clamped to 0 forever.

https://github.com/ros-planning/navigation2/blob/da5edbfd12efd0ba74cd6bb6bc8c8ad8ca9bd88a/nav2_velocity_smoother/src/velocity_smoother.cpp#L308

Workaround is to disable deadband_velocity.

Also, I think this only affects open loop mode.

BriceRenaudeau:
Sorry, the fix is here soon :https://github.com/ros-planning/navigation2/pull/3690

SteveMacenski:
Merged! Backporting to iron / humble shortly

PR #3700, opened 2023-07-17
Velocity smoother natural deadband

Aposhian:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3699 |

---

This fixes the bug that specifying a non-zero deadband velocity on open loop mode makes it so you are stuck at 0 forever. This also adds a separate construct, the natural deadband velocity, which snaps up to that minimum value instead of snapping down like the existing deadband velocity. It's purpose is to fix trying to use closed loop mode on a system with an inherent deadband.

Added new parameter `natural_deadband_velocities`

---

- maybe rename this new parameter

#### For Maintainers: 


Aposhian:
Still needs some significant work, so I will close for now

PR #3704, opened 2023-07-21
Behaviors access result

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
@@ -90,7 +90,7 @@ public:
   // Implement the behavior such that it runs some unit of work on each call
   // and provides a status. The Behavior will finish once SUCCEEDED is returned
   // It's up to the derived class to define the final commanded velocity.
-  virtual ResultStatus onCycleUpdate() = 0;
+  virtual ResultStatus onCycleUpdate(std::shared_ptr<typename ActionT::Result> result) = 0;
 
   // an opportunity for derived classes to do something on configuration
   // if they chose
@@ -248,7 +248,7 @@ protected:
         return;
       }
 
-      ResultStatus on_cycle_update_result = onCycleUpdate();
+      ResultStatus on_cycle_update_result = onCycleUpdate(result);
       switch (on_cycle_update_result.status) {
         case Status::SUCCEEDED:
           RCLCPP_INFO(
--- a/nav2_behaviors/plugins/assisted_teleop.cpp
+++ b/nav2_behaviors/plugins/assisted_teleop.cpp
@@ -77,7 +77,7 @@ void AssistedTeleop::onActionCompletion()
   preempt_teleop_ = false;
 }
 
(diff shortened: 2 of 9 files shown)


nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
```diff
@@ -90,7 +90,7 @@ class TimedBehavior : public nav2_core::Behavior
   // Implement the behavior such that it runs some unit of work on each call
   // and provides a status. The Behavior will finish once SUCCEEDED is returned
   // It's up to the derived class to define the final commanded velocity.
-  virtual ResultStatus onCycleUpdate() = 0;
+  virtual ResultStatus onCycleUpdate(std::shared_ptr<typename ActionT::Result> result) = 0;
```
SteveMacenski, line 93: Wouldn't this make alot more sense in `onActionCompletion`?

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -77,7 +77,7 @@ void AssistedTeleop::onActionCompletion()
   preempt_teleop_ = false;
 }
 
-ResultStatus AssistedTeleop::onCycleUpdate()
+ResultStatus AssistedTeleop::onCycleUpdate(std::shared_ptr<AssistedTeleopActionResult>)
```
SteveMacenski, line 80: Nit: but add the commented out name `/*result*/` like we do elsewhere in the codebase for all eq.

SteveMacenski (changes requested):

nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
SteveMacenski: Please do. I dont think cycle updates should need to populate the result. The ‘action complete’ callback seems like the right place

nav2_behaviors/plugins/assisted_teleop.cpp
```diff
@@ -71,7 +71,7 @@ ResultStatus AssistedTeleop::onRun(const std::shared_ptr<const AssistedTeleopAct
   return ResultStatus{Status::SUCCEEDED, AssistedTeleopActionResult::NONE};
 }
 
-void AssistedTeleop::onActionCompletion()
+void AssistedTeleop::onActionCompletion(std::shared_ptr<AssistedTeleopActionResult>/*result*/)
```
SteveMacenski, line 74: Anything worth populating since we're updating it anyway?

SteveMacenski: As well with the other actions

SteveMacenski (commented):
Overall, looks good to me. Docs update per usual + I tink we need to update the tutorial on behavior API to mention this (and perhaps your other recent update)

SteveMacenski (approved):

jwallace42:
@SteveMacenski,

can I backport this into iron?

SteveMacenski:
No, this is an ABI breaking change

PR #3706, opened 2023-07-23
Use ParameterFile (allow_substs)

ymd-stella:
## Basic Info

| ------ | ----------- |

---

- Enable substitution in parameter file. For example, you can write the following

```yaml
default_nav_to_pose_bt_xml: $(find-pkg-share my_package)/behavior_tree/my_nav_to_pose_bt.xml
```

---

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/navigation_launch.py
+++ b/nav2_bringup/launch/navigation_launch.py
@@ -22,7 +22,7 @@ from launch.conditions import IfCondition
 from launch.substitutions import LaunchConfiguration, PythonExpression
 from launch_ros.actions import LoadComposableNodes, SetParameter
 from launch_ros.actions import Node
-from launch_ros.descriptions import ComposableNode
+from launch_ros.descriptions import ComposableNode, ParameterFile
 from nav2_common.launch import RewrittenYaml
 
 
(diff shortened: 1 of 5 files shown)


nav2_bringup/launch/navigation_launch.py
```diff
@@ -61,11 +61,13 @@ def generate_launch_description():
     param_substitutions = {
         'autostart': autostart}
 
-    configured_params = RewrittenYaml(
-        source_file=params_file,
-        root_key=namespace,
-        param_rewrites=param_substitutions,
-        convert_types=True)
+    configured_params = ParameterFile(
```
SteveMacenski, line 64: Totally approve of this action, but don't you need to do this for all instances of `RewrittenYaml`, not just this one so it could be used for anywhere this kind of pattern may be useful?

SteveMacenski (changes requested):
Also, add documentation updates for the migration guide that this is now an option, thanks!

Also, I'd add to the configuration docs for that parameter that this is a possible pattern to use -- that's worth explicitly documenting

nav2_bringup/launch/navigation_launch.py
ymd-stella: Updated.

SteveMacenski (approved):

SteveMacenski:
@ymd-stella check CI, there are linting problems and suspiciously all of the system tests are now failing. Did you check that this brings up fine?

ymd-stella:
Fixed. Old files were being pushed.

SteveMacenski:
Thanks @ymd-stella this is a really nice feature and I feel silly that I didn’t know it was already possible to do this in the launch API. I’ve been wanting this for ages for this EXACT reason!

Issue #3313, opened 2022-12-05
[Collision monitor] feature discussion: Stop zone with allowed rotation or inverted direction


doisyg:
First of all, thanks a lot to @AlexeyMerzlyakov and @SteveMacenski for the collision monitor. I am finding myself in a new company and having yet again to implement something similar to the collision monitor. Just started experimenting with it. It is great to be able to find this brick part of nav2.

A bit of context about my use case: a robot with a full certified low-level functional safety where we are using nav2 on top and with the collision monitor configured with slightly more conservative zones in order to never trigger the low level safety.

#### Feature description
A stop zone with optional allowed degree of freedoms in order to avoid having a robot stuck.
For instance:

##### For a circular robot
When obstacle in the zone, forbid linear speeds, but allow rotations / angular z speeds

##### For a rectangular bi-directional robot
When obstacle in the zone, forbid angular speeds, forbid linear speed in one direction but allow in the inverted direction (i.e. when an obstacle is in front of the robot, allow to go backward)

#### Implementation considerations
TBD if this feature is seen as useful. I will happily propose an implementation and a PR.

SteveMacenski:
Any other bricks we should add? :wink: I only know what to add based on my personal experiences and what people tell me they want! 

Consideration: there is the collision monitor approach mode which uses the velocity of the robot in accounting for collision within immediate future. If you use that, you can back out or move laterally, no problem, since it won't be in the directions of collision. But I assume you know that and want to still use the polygon zones?

One thing I'm concerned by here is generality. Right now, there are just zones without any specificity about their positioning relative to the robot base (they could be in front, behind, off to the right by 2000 meters, enveloping the robot etc) and knowing which of these its sensible to apply the conditioning to may be non-trivial. Or the policies that would be relevant for each (e.g. omni, diff, ackermann). Though, I think the policies are an easier thing to embed by what you describe as "degrees of freedom" as configurations.

Additionally, how do you deal with conflicts if there are multiple zones? Lets say you have a forward and a reverse zone setup: frontal says stop, backwards is OK. So you start going backwards, then the backwards one triggers because you're running near a wall. Or if you have a polygon enveloping the entire robot so forward has points for collision so you start reversing and now there are points in the reverse direction. Since its part of the same polygon, that could cause a collision.  

The secondary problem set is providing the tooling (if not simply the recovery behaviors?) to perform the actions described as admissible by the policy. What were you thinking there? Just the default recovery behaviors or something else?

[...]

PR #3708, opened 2023-07-23
Add velocity based polygon

kaichie:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #3313  |
| Robotic platform tested on | Nav2 TB3 gazebo simulation |

---

* Adds velocity based polygon as the third priority in Collision Monitor

* Added new parameter - `polygon_velocity`, so need to add that to default configs and documentation page

---

* Add odom speed condition check (linear, direction + theta too) for speed based polygon. (Better control on when to switch based on actual speed)

## Pending 
* Full range velocity coverage check(linear, direction, theta)

## Test
1. Use `tb3_simulation_launch.py`
2. Update collision_monitor [config](https://github.com/ros-planning/navigation2/blob/main/nav2_collision_monitor/params/collision_monitor_params.yaml#L17) to use `FootprintApproach`
3. Use `rqt_robot_steering` for convenient bi-directional testing

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/CMakeLists.txt
+++ b/nav2_collision_monitor/CMakeLists.txt
@@ -45,6 +45,7 @@ set(library_name ${executable_name}_core)
 add_library(${library_name} SHARED
   src/collision_monitor_node.cpp
   src/polygon.cpp
+  src/polygon_velocity.cpp
   src/circle.cpp
   src/source.cpp
   src/scan.cpp
--- a/nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
@@ -28,6 +28,7 @@
 #include "nav2_util/lifecycle_node.hpp"
 #include "nav2_costmap_2d/footprint_subscriber.hpp"
 
+#include "nav2_collision_monitor/polygon_velocity.hpp"
 #include "nav2_collision_monitor/types.hpp"
 
 namespace nav2_collision_monitor
@@ -126,6 +127,16 @@ public:
    * Othewise, prints a warning and returns false.
    */
   virtual bool isShapeSet();
+
+  /**
+   * @brief Returns true if using velocity based polygon
+   */
+  bool isUsingPolygonVelocitySelector();
+
+  /**
+   * @brief Updates polygon by velocity (if any)
+   */
+  void updatePolygonByVelocity(const Velocity & cmd_vel_in);
 
   /**
    * @brief Updates polygon from footprint subscriber (if any)
@@ -242,6 +253,9 @@ protected:
   geometry_msgs::msg::PolygonStamped polygon_;
   /// @brief Polygon publisher for visualization purposes
   rclcpp_lifecycle::LifecyclePublisher<geometry_msgs::msg::PolygonStamped>::SharedPtr polygon_pub_;
+
+  /// @brief Polygon velocity (if any)
+  std::vector<std::shared_ptr<PolygonVelocity>> polygon_velocity_;
 
   /// @brief Polygon points (vertices) in a base_frame_id_
   std::vector<Point> poly_;
--- a/nav2_collision_monitor/include/nav2_collision_monitor/polygon_velocity.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/polygon_velocity.hpp
@@ -1,0 +1,103 @@
+// Copyright (c) 2022 Samsung R&D Institute Russia
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NAV2_COLLISION_MONITOR__POLYGON_VELOCITY_HPP_
+#define NAV2_COLLISION_MONITOR__POLYGON_VELOCITY_HPP_
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "rclcpp/rclcpp.hpp"
+
+#include "nav2_collision_monitor/types.hpp"
+
+namespace nav2_collision_monitor
+{
+
+/**
+ * @brief Polygon velocity class.
+ * This class contains all the points of the polygon and
+ * the expected condition of the velocity based polygon.
+ */
+class PolygonVelocity
+{
+public:
+  /**
+   * @brief PolygonVelocity constructor
+   * @param poly Points of the polygon
+   * @param polygon_name Name of polygon
+   * @param linear_max Maximum twist linear velocity
+   * @param linear_min Minimum twist linear velocity
+   * @param direction_max Maximum twist direction angle
+   * @param direction_min Minimum twist direction angle
+   * @param theta_max Maximum twist rotational speed
+   * @param theta_min Minimum twist rotational speed
+   */
+  PolygonVelocity(
+    const std::vector<Point> & poly,
+    const std::string & polygon_name,
+    const double & linear_max,
+    const double & linear_min,
+    const double & direction_max,
+    const double & direction_min,
+    const double & theta_max,
+    const double & theta_min);
+  /**
+   * @brief PolygonVelocity destructor
+   */
+  virtual ~PolygonVelocity();
+
+  /**
+   * @brief Check if the velocities and direction is in expected range.
+   * @param cmd_vel_in Robot twist command input
+   * @return True if speed and direction is within the condition
+   */
+  bool isInRange(const Velocity & cmd_vel_in);
+
+  /**
+   * @brief Check if the velocities and direction is in expected range.
+   * @param cmd_vel_in Robot twist command input
+   * @return True if speed and direction is within the condition
+   */
+  std::vector<Point> getPolygon();
+
+protected:
+  // ----- Variables -----
+
+  /// @brief Collision monitor node logger stored for further usage
+  rclcpp::Logger logger_{rclcpp::get_logger("collision_monitor")};
+
+  // Basic parameters
+  /// @brief Points of the polygon
+  std::vector<Point> poly_;
+  /// @brief Name of polygon
+  std::string polygon_name_;
+  /// @brief Maximum twist linear velocity
+  double linear_max_;
+  /// @brief Minimum twist linear velocity
+  double linear_min_;
+  /// @brief Maximum twist direction angle
+  double direction_max_;
+  /// @brief Minimum twist direction angle
+  double direction_min_;
+  /// @brief Maximum twist rotational speed
+  double theta_max_;
+  /// @brief Minimum twist rotational speed
+  double theta_min_;
+};  // class PolygonVelocity
+
+}  // namespace nav2_collision_monitor
+
+#endif  // NAV2_COLLISION_MONITOR__POLYGON_VELOCITY_HPP_
--- a/nav2_collision_monitor/params/collision_monitor_params.yaml
+++ b/nav2_collision_monitor/params/collision_monitor_params.yaml
@@ -42,11 +42,45 @@ collision_monitor:
     FootprintApproach:
       type: "polygon"
       action_type: "approach"
-      footprint_topic: "/local_costmap/published_footprint"
+      # footprint_topic: "/local_costmap/published_footprint"
       time_before_collision: 2.0
       simulation_time_step: 0.1
       min_points: 6
-      visualize: False
+      visualize: True
+      polygon_pub_topic: "polygon_approach"
+      polygon_velocity: ["stop", "rotation", "translation_forward", "translation_backward"] # third priority
+      stop:
+        points: [0.25, 0.25, 0.25, -0.25, -0.25, -0.25, -0.25, 0.25]
+        linear_min: 0.0
+        linear_max: 0.01
+        direction_min: -3.14159
+        direction_max: 3.14159
+        theta_min: 0.0
+        theta_max: 0.01
+      rotation:
+        points: [0.3, 0.3, 0.3, -0.3, -0.3, -0.3, -0.3, 0.3]
+        linear_min: 0.0
+        linear_max: 0.05
+        direction_min: -3.14159
+        direction_max: 3.14159
+        theta_min: -1.0
+        theta_max: 1.0
+      translation_forward:
+        points: [0.35, 0.3, 0.35, -0.3, -0.2, -0.3, -0.2, 0.3]
+        linear_min: 0.0
+        linear_max: 1.0
+        direction_min: -1.5708
+        direction_max: 1.5708
+        theta_min: -1.0
+        theta_max: 1.0
+      translation_backward:
+        points: [0.2, 0.3, 0.2, -0.3, -0.35, -0.3, -0.35, 0.3]
+        linear_min: 0.0
+        linear_max: 1.0
+        direction_min: 1.5708
+        direction_max: -1.5708
+        theta_min: -1.0
+        theta_max: 1.0
     observation_sources: ["scan"]
     scan:
       type: "scan"
--- a/nav2_collision_monitor/src/collision_monitor_node.cpp
+++ b/nav2_collision_monitor/src/collision_monitor_node.cpp
@@ -423,6 +423,10 @@ bool CollisionMonitor::processStopSlowdownLimit(
   const Velocity & velocity,
   Action & robot_action) const
 {
+  if (polygon->isUsingPolygonVelocitySelector()) {
+    polygon->updatePolygonByVelocity(velocity);
+  }
+
   if (!polygon->isShapeSet()) {
     return false;
   }
@@ -478,6 +482,10 @@ bool CollisionMonitor::processApproach(
   const Velocity & velocity,
   Action & robot_action) const
 {
+  if (polygon->isUsingPolygonVelocitySelector()) {
+    polygon->updatePolygonByVelocity(velocity);
+  }
+
   if (!polygon->isShapeSet()) {
     return false;
   }
--- a/nav2_collision_monitor/src/polygon.cpp
+++ b/nav2_collision_monitor/src/polygon.cpp
@@ -168,6 +168,34 @@ bool Polygon::isShapeSet()
   return true;
 }
 
+bool Polygon::isUsingPolygonVelocitySelector()
+{
+  return !polygon_velocity_.empty();
+}
+
+void Polygon::updatePolygonByVelocity(const Velocity & cmd_vel_in)
+{
+  for (auto & polygon : polygon_velocity_) {
+
+    if (polygon->isInRange(cmd_vel_in)) {
+      // Set the polygon that is within the speed range
+      poly_ = polygon->getPolygon();
+
+      // Update visualization polygon
+      polygon_.polygon.points.clear();
+      for (const Point & p : poly_) {
+        geometry_msgs::msg::Point32 p_s;
+        p_s.x = p.x;
+        p_s.y = p.y;
+        // p_s.z will remain 0.0
+        polygon_.polygon.points.push_back(p_s);
+      }
+
+      return;
+    }
+  }
+}
+
 void Polygon::updatePolygon()
 {
   if (footprint_sub_ != nullptr) {
@@ -420,19 +448,121 @@ bool Polygon::getParameters(
         polygon_name_.c_str());
     }
 
-    if (action_type_ == STOP || action_type_ == SLOWDOWN || action_type_ == LIMIT) {
-      // Dynamic polygon will be used
+    try {
+      if (action_type_ == STOP || action_type_ == SLOWDOWN || action_type_ == LIMIT) {
+        // Dynamic polygon will be used
+        nav2_util::declare_parameter_if_not_declared(
+          node, polygon_name_ + ".polygon_sub_topic", rclcpp::PARAMETER_STRING);
+        polygon_sub_topic =
+          node->get_parameter(polygon_name_ + ".polygon_sub_topic").as_string();
+        return true;
+      } else if (action_type_ == APPROACH) {
+        // Obtain the footprint topic to make a footprint subscription for approach polygon
+        nav2_util::declare_parameter_if_not_declared(
+          node, polygon_name_ + ".footprint_topic", rclcpp::PARAMETER_STRING);
+        footprint_topic =
+          node->get_parameter(polygon_name_ + ".footprint_topic").as_string();
+        return true;
+      }
+    } catch (const rclcpp::exceptions::ParameterUninitializedException &) {
+      RCLCPP_INFO(
+        logger_,
+        "[%s]: Polygon topic are not defined. Using polygon velocity instead.",
+        polygon_name_.c_str());
+    }
+
+    // Polygon velocity will be used
+    nav2_util::declare_parameter_if_not_declared(
+      node, polygon_name_ + ".polygon_velocity", rclcpp::PARAMETER_STRING_ARRAY);
+    std::vector<std::string> polygon_velocity_array = node->get_parameter(
+      polygon_name_ + ".polygon_velocity").as_string_array();
+
+    for (std::string polygon_velocity_name : polygon_velocity_array) {
+
+      std::string polygon_name = polygon_velocity_name;
+
       nav2_util::declare_parameter_if_not_declared(
-        node, polygon_name_ + ".polygon_sub_topic", rclcpp::PARAMETER_STRING);
-      polygon_sub_topic =
-        node->get_parameter(polygon_name_ + ".polygon_sub_topic").as_string();
-    } else if (action_type_ == APPROACH) {
-      // Obtain the footprint topic to make a footprint subscription for approach polygon
+        node, polygon_name_ + "." + polygon_velocity_name + ".points",
+        rclcpp::PARAMETER_DOUBLE_ARRAY);
+      std::vector<double> polygon_points = node->get_parameter(
+        polygon_name_ + "." + polygon_velocity_name + ".points").as_double_array();
+
+      // Check for points format correctness
+      if (polygon_points.size() <= 6 || polygon_points.size() % 2 != 0) {
+        RCLCPP_ERROR(
+          logger_,
+          "[%s]: Polygon has incorrect points description", polygon_name_.c_str());
+        return false;
+      }
+
+      // Obtain polygon vertices
+      Point point;
+      std::vector<nav2_collision_monitor::Point> poly;
+      bool first = true;
+      for (double val : polygon_points) {
+        if (first) {
+          point.x = val;
+        } else {
+          point.y = val;
+          poly.push_back(point);
+        }
+        first = !first;
+      }
+
+      // linear_max param
       nav2_util::declare_parameter_if_not_declared(
-        node, polygon_name_ + ".footprint_topic",
-        rclcpp::ParameterValue("local_costmap/published_footprint"));
-      footprint_topic =
-        node->get_parameter(polygon_name_ + ".footprint_topic").as_string();
+        node, polygon_name_ + "." + polygon_velocity_name + ".linear_max",
+        rclcpp::ParameterValue(0.0));
+      double linear_max =
+        node->get_parameter(
+        polygon_name_ + "." + polygon_velocity_name +
+        ".linear_max").as_double();
+
+      // linear_min param
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + "." + polygon_velocity_name + ".linear_min",
+        rclcpp::ParameterValue(0.0));
+      double linear_min =
+        node->get_parameter(
+        polygon_name_ + "." + polygon_velocity_name +
+        ".linear_min").as_double();
+
+      // direction_max param
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + "." + polygon_velocity_name + ".direction_max",
+        rclcpp::ParameterValue(0.0));
+      double direction_max =
+        node->get_parameter(
+        polygon_name_ + "." + polygon_velocity_name +
+        ".direction_max").as_double();
+
+      // direction_min param
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + "." + polygon_velocity_name + ".direction_min",
+        rclcpp::ParameterValue(0.0));
+      double direction_min =
+        node->get_parameter(
+        polygon_name_ + "." + polygon_velocity_name +
+        ".direction_min").as_double();
+
+      // theta_max param
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + "." + polygon_velocity_name + ".theta_max", rclcpp::ParameterValue(
+          0.0));
+      double theta_max =
+        node->get_parameter(polygon_name_ + "." + polygon_velocity_name + ".theta_max").as_double();
+
+      // theta_min param
+      nav2_util::declare_parameter_if_not_declared(
+        node, polygon_name_ + "." + polygon_velocity_name + ".theta_min", rclcpp::ParameterValue(
+          0.0));
+      double theta_min =
+        node->get_parameter(polygon_name_ + "." + polygon_velocity_name + ".theta_min").as_double();
+
+      polygon_velocity_.push_back(
+        std::make_shared<PolygonVelocity>(
+          poly, polygon_name, linear_max, linear_min, direction_max,
+          direction_min, theta_max, theta_min));
     }
   } catch (const std::exception & ex) {
     RCLCPP_ERROR(
--- a/nav2_collision_monitor/src/polygon_velocity.cpp
+++ b/nav2_collision_monitor/src/polygon_velocity.cpp
@@ -1,0 +1,74 @@
+// Copyright (c) 2022 Samsung R&D Institute Russia
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "nav2_collision_monitor/polygon_velocity.hpp"
+
+#include "nav2_util/node_utils.hpp"
+#include "nav2_util/robot_utils.hpp"
+
+
+namespace nav2_collision_monitor
+{
+
+PolygonVelocity::PolygonVelocity(
+  const std::vector<Point> & poly,
+  const std::string & polygon_name,
+  const double & linear_max,
+  const double & linear_min,
+  const double & direction_max,
+  const double & direction_min,
+  const double & theta_max,
+  const double & theta_min)
+: poly_(poly), polygon_name_(polygon_name), linear_max_(linear_max), linear_min_(linear_min),
+  direction_max_(direction_max), direction_min_(direction_min),
+  theta_max_(theta_max), theta_min_(theta_min)
+{
+  RCLCPP_INFO(logger_, "[%s]: Creating PolygonVelocity", polygon_name_.c_str());
+}
+
+
+PolygonVelocity::~PolygonVelocity()
+{
+  RCLCPP_INFO(logger_, "[%s]: Destroying PolygonVelocity", polygon_name_.c_str());
+  poly_.clear();
+}
+
+bool PolygonVelocity::isInRange(const Velocity & cmd_vel_in)
+{
+  const double twist_linear = std::hypot(cmd_vel_in.x, cmd_vel_in.y);
+
+  // check if direction in angle range(min -> max)
+  double direction = std::atan2(cmd_vel_in.y, cmd_vel_in.x);
+  bool direction_in_range;
+  if (direction_min_ <= direction_max_) {
+    direction_in_range = (direction >= direction_min_ && direction <= direction_max_);
+  } else {
+    direction_in_range = (direction >= direction_min_ || direction <= direction_max_);
+  }
+
+  return twist_linear <= linear_max_ &&
+         twist_linear >= linear_min_ &&
+         direction_in_range &&
+         cmd_vel_in.tw <= theta_max_ &&
+         cmd_vel_in.tw >= theta_min_;
+}
+
+
+std::vector<Point> PolygonVelocity::getPolygon()
+{
+  return poly_;
+}
+
+
+}  // namespace nav2_collision_monitor


nav2_collision_monitor/include/nav2_collision_monitor/polygon_velocity.hpp
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2022 Samsung R&D Institute Russia
```
AlexeyMerzlyakov, line 1: Ditto

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
```diff
@@ -243,6 +254,9 @@ class Polygon
   /// @brief Polygon publisher for visualization purposes
   rclcpp_lifecycle::LifecyclePublisher<geometry_msgs::msg::PolygonStamped>::SharedPtr polygon_pub_;
 
+  /// @brief Polygon velocity (if any)
+  std::vector<std::shared_ptr<PolygonVelocity>> polygon_velocity_;
```
AlexeyMerzlyakov, line 258: Since it is vector of `PolygonVelocity` type, this, I think, is better to rename to something, like:
```
std::vector<std::shared_ptr<PolygonVelocity>> polygons_velocity_;
```
... in order to avoid further confusions

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,6 +168,34 @@ bool Polygon::isShapeSet()
...
 }
 
+bool Polygon::isUsingPolygonVelocitySelector()
+{
+  return !polygon_velocity_.empty();
+}
+
+void Polygon::updatePolygonByVelocity(const Velocity & cmd_vel_in)
+{
+  for (auto & polygon : polygon_velocity_) {
```
AlexeyMerzlyakov, line 178: `polygon` in this case will be of `PolygonVelocity` type. So, I'd like to suggest to rename it for better clarity to something like:
```
for (auto & polygon_velocity : polygons_velocity_) {
 ...
```

Does it make sense?

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,6 +168,34 @@ bool Polygon::isShapeSet()
...
+      polygon_.polygon.points.clear();
+      for (const Point & p : poly_) {
+        geometry_msgs::msg::Point32 p_s;
+        p_s.x = p.x;
+        p_s.y = p.y;
+        // p_s.z will remain 0.0
+        polygon_.polygon.points.push_back(p_s);
+      }
+
+      return;
```
AlexeyMerzlyakov, line 194: The first found PolygonVelocity will be applied. In case if we will have many overlapped polygons_velocity, responding to the same velocity, safest rule may not be found. However, this is the big question of Collision Monitor logic design, as the decision of making safest solution is made at `CollisionMonitorNode` level, which is above `Polygon` logic. If you will have any proper solution/suggestion for this situation, please check how could it be applied w/o large re-work of CM, or not? Otherwise, we have to assume that velocity ranges in {linear, angular} space should not intersect for one polygon.

AlexeyMerzlyakov (commented):
Overall approach seems good for me.

The rotation at the place could be handled by making small polygon inside when the linear velocity is equal to `0.0 (+ epsilon)`.

Another point, that developer could intentionally leave some ranges in the velocities, where no one polygon_velocity will be applied. This case should be tracked (informed, or warned), but I do not see any reason to check all the velocity ranges in {linear, angular} space preliminary for the consistency before we will apply the rules.

The question with overlapped polygons_velocity for one polygon remains to be discussed: should we find safest polygon_velocity responsing to the input velocity, or agree that velocity ranges within one polygon shouldn't be overlapped?

Please note, that circles will not have an ability of polygons_velocity utilization (which is OK from my point of view, and already handled by this change as `polygons_velocity_` vector is not being initialized in the constructor).

There are some changes to be noted in the review items below, please refer to them

nav2_collision_monitor/include/nav2_collision_monitor/polygon_velocity.hpp
```diff
@@ -0,0 +1,103 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NAV2_COLLISION_MONITOR__POLYGON_VELOCITY_HPP_
+#define NAV2_COLLISION_MONITOR__POLYGON_VELOCITY_HPP_
+
+#include <memory>
```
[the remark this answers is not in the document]
kaichie: moved to `polygon.cpp` instead

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -478,6 +482,10 @@ bool CollisionMonitor::processApproach(
   const Velocity & velocity,
   Action & robot_action) const
 {
+  if (polygon->isUsingPolygonVelocitySelector()) {
+    polygon->updatePolygonByVelocity(velocity);
+  }
+
```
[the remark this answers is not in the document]
kaichie: updated as suggested

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -420,19 +448,121 @@ bool Polygon::getParameters(
...
+      bool first = true;
+      for (double val : polygon_points) {
+        if (first) {
+          point.x = val;
+        } else {
+          point.y = val;
+          poly.push_back(point);
+        }
+        first = !first;
+      }
```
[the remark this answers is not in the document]
kaichie: Added `setPolyShape` to optimise the duplicated code

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,6 +168,34 @@ bool Polygon::isShapeSet()
...
+        geometry_msgs::msg::Point32 p_s;
+        p_s.x = p.x;
+        p_s.y = p.y;
+        // p_s.z will remain 0.0
+        polygon_.polygon.points.push_back(p_s);
+      }
+
+      return;
+    }
+  }
```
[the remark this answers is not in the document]
kaichie: Added a warn logging for the uncovered speed.

nav2_collision_monitor/src/polygon_velocity.cpp
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) 2022 Samsung R&D Institute Russia
```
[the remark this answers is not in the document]
kaichie: fixed

nav2_collision_monitor/src/polygon.cpp
kaichie: Much better :100:

nav2_collision_monitor/src/polygon.cpp
kaichie: If we want to support finding the safest velocity for overlapping polygons without a major re-work, we can do something like this(but which polygon to visualise for `PolygonVelocity`?):

```cpp
    if (polygon->isUsingPolygonVelocitySelector()) {
      // Get all the polygons that is in range
      polygonsToProcess = polygon-> getPolygonsByVelocity(cmd_vel_in);
    } else {
      // Otherwise, dynamically update polygon coordinates when it necessary
      polygon->updatePolygon();
      polygonsToProcess.push_back(polygon);
    }

    for (auto & polygon: polygonsToProcess){
      const ActionType at = polygon->getActionType();
      if (at == STOP || at == SLOWDOWN || at == LIMIT) {
        // Process STOP/SLOWDOWN for the selected polygon
        if (processStopSlowdownLimit(polygon, collision_points, cmd_vel_in, robot_action)) {
          action_polygon = polygon;
        }
      } else if (at == APPROACH) {
        // Process APPROACH for the selected polygon
        if (processApproach(polygon, collision_points, cmd_vel_in, robot_action)) {
          action_polygon = polygon;
        }
      }
    }
```
I would agree that velocity ranges within one polygon shouldn't be overlapped, and if it does, we should choose the first polygon per the config.

AlexeyMerzlyakov: @kaichie, thanks for the update of your thoughts on it. The problem to draw the safest polygon could be resolved by adding a `for (auto sub_polygon : polygon->getPolygonsByVelocity(cmd_vel_in)) {...}` sub-cycle with constant `cmd_vel_in` as input to realize which sub-polygon is the best locally and then comparing produced by each polygon velocities globally to make the safest decision. We could switch to `std::vector<Velocity> generated_cmd_vel` per each polygon / sub-polygon and then select the best solution per each polygon though its sub-polygons, and globally. The knowledge about best polygon shape selected from sub-polygons could be stored (or re-used, it should be carefully checked - could we do this) in some field of `Polygon` class.

Another alternative I see - to move the best velocity selection `if() {processStopSlowdownLimit()} else if() {processApproach()}` logic into one separate routine, and then pass this routine callback to the polygon selector to compare each sub-polygon on its safest solution. The safest velocities should be stored again, in some `Velocity` vector. This is not straightforward solution, but also could take place.

Anyway, the main subject here is - which use-case might we cover by overlapping sets of different velocities for one polygon. I can not find any of them for now: the ticket initially was made to move out robot from some situations when CM was already stuck the robot, and we thus need to change the safety shape to move robot backward; or rotate. Or another case will use the robot's speed to change dynamically safety shape, but there is no intention to set multiple shapes per different velocities, that could work at the same time. For this, we have initial CM's multi-polygons mechanism, and thus, I do not see any reason why we need to double this logic into sub-polygons. In case if developers wants to have multiple choices for one robot, they could define multiple polygons (with its own velocity rules per each) , instead of intentionally making sub-polygons overlapped by velocities.

Anyway, if you have any example, when the above won't work, I think, we need to think about supporting safest selection through sub-polygons, as well.

kaichie: > However, the main subject here is - which use-case might we cover by overlapping sets of different velocities for one polygon. I can not find any of them for now: the ticket initially was made to move out robot from some situations when CM was already stuck the robot, and we thus need to change the safety shape to move robot backward; or rotate. 

@AlexeyMerzlyakov Yes this is the use-case for now, not aware of other scenario that would require finding the safest sub-polygons with overlapping velocity. However it is a good discussion to know that we can still do this, if needed, without a major re-work on the CM.

kaichie:
> Another point, that developer could intentionally leave some ranges in the velocities, where no one polygon_velocity will be applied. This case should be tracked (informed, or warned), but I do not see any reason to check all the velocity ranges in {linear, angular} space preliminary for the consistency before we will apply the rules.

@AlexeyMerzlyakov Regarding of this point, the warning(on run-time) of uncovered velocity is added in this [commit](https://github.com/ros-planning/navigation2/commit/727060ca4a2762e303947fb85dcaaa879ebd2e33). If the user is intentionally leaving some ranges in the velocities, then this might conflict with what we were discussing in the issue to check the full range velocity(@SteveMacenski). We might need to specify the intended gap(s) in the config but I think this will greatly increase the complexity of setting it up.

SteveMacenski:
I commented back on the main ticket thread on that topic. Maybe just logging is enough if we provide good example profiles with the documentation

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -42,11 +42,45 @@ collision_monitor:
...
-      visualize: False
+      visualize: True
+      polygon_pub_topic: "polygon_approach"
+      polygon_velocity: ["stop", "rotation", "translation_forward", "translation_backward"] # third priority
+      stop:
+        points: [0.25, 0.25, 0.25, -0.25, -0.25, -0.25, -0.25, 0.25]
+        linear_min: 0.0
+        linear_max: 0.01
+        direction_min: -3.14159
+        direction_max: 3.14159
```
SteveMacenski, line 57: I don't understand "direction" min / max. If we have the linear velocity min/max and the angular velocity min/max (which should be signed), why is this relevant?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -386,8 +386,13 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
       break;
     }
 
-    // Update polygon coordinates
-    polygon->updatePolygon();
+    if (polygon->isUsingPolygonVelocitySelector()) {
+      // Update polygon using its velocity ranges
+      polygon->updatePolygonByVelocity(cmd_vel_in);
```
SteveMacenski, line 391: We could just make updatePolygon take a velocity field that the normal polygon ignores that doesn't care about it. Simplifies this because we don't need to check anything

nav2_collision_monitor/include/nav2_collision_monitor/polygon_velocity.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+namespace nav2_collision_monitor
+{
+
+/**
+ * @brief Polygon velocity class.
+ * This class contains all the points of the polygon and
+ * the expected condition of the velocity based polygon.
+ */
+class PolygonVelocity
```
SteveMacenski, line 33: VelocityPolygon

I think the main thing here is that this is a velocity profile to apply a polygon, not a polygon to apply a velocity profile. The conditional input is the velocity which we apply the attribute of the polygon.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,6 +169,37 @@ bool Polygon::isShapeSet()
...
+  RCLCPP_WARN(
```
SteveMacenski, line 198: Make this a throttled warning every few seconds. This will blast the terminal at controller rate which would get really annoying

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -388,51 +420,114 @@ bool Polygon::getParameters(
...
+      std::vector<double> polygon_points = node->get_parameter(
+        polygon_name_ + "." + polygon_velocity_name + ".points").as_double_array();
+
+      std::vector<Point> poly;
+      if (!setPolygonShape(polygon_points, poly)) {
+        return false;
+      }
+
+      // linear_max param
+      nav2_util::declare_parameter_if_not_declared(
```
SteveMacenski, line 478: Why can't the VelocityPolygon object get  its own parameters? That would be better separation of concerns

SteveMacenski:
Blah blah blah, the usual bits on documentation (readme + configuration guide + migration guide) and test coverage :-)

nav2_collision_monitor/params/collision_monitor_params.yaml
AlexeyMerzlyakov: As I understood, this to be a sector of the velocities direction whether the required dynamic polygon actions are applied to:
directions

This might be useful, if velocity is not always parallel to OX coordinate in robot base frame, e.g. for omni-drives.

SteveMacenski: I don't understand, velocity is signed, why don't we just specify velocity ranges with signs? That should still establish the cmd_vel range. I don't think we should overoptimize here to get this out. We can always come back if there's a tangible issue for omni robot platforms. 

Though, I don't really see this example being really useful to even omni robot providers. They can still describe the polygon to use if the X velocity is in some range and the Y velocity is in some range, so that would still describe this envelope (e.g. atan2(y,x) gives you the relative angle -- so its redundant information)

AlexeyMerzlyakov: This might be overoptimiation, if we could use the velocity corridors in velocity сartesian coordinate system. However, how could we set this space if we want to reflect some 45 degrees direction (e.g. for omni-drives) with velocity from 0 ... to max? From other side I am no sure, that this has a real use-case applications - from initial talks, we we used to say about forward/backward moving directions and rotation in a place. Plus your example when we could neglect one of the velocity components (that could not be resolved in polar coordinate system).

Thinking twice about it: yes, this seems to be unnecessary complication.

SteveMacenski: I'd be curious what @kaichie thinks. 

OK, I can get on board with this. I think we should have some param for `holonomic` and if `true` we consider `direction`, if not that part of of the `isInRange` is skipped. While that's not a really formal abstraction of the different motion model types, I think that's sufficient. 

I think including a good piece of documentation in the readme + a version of that diagram you provided will be important to describe what that means in detail for omni robots

nav2_collision_monitor/src/polygon_velocity.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  const double twist_linear = std::hypot(cmd_vel_in.x, cmd_vel_in.y);
+
+  // check if direction in angle range(min -> max)
+  double direction = std::atan2(cmd_vel_in.y, cmd_vel_in.x);
+  bool direction_in_range;
+  if (direction_min_ <= direction_max_) {
+    direction_in_range = (direction >= direction_min_ && direction <= direction_max_);
+  } else {
+    direction_in_range = (direction >= direction_min_ || direction <= direction_max_);
```
SteveMacenski, line 54: Wouldn't this be `direction >= direction_max_ || direction <= direction_min_`?

This if/else for the directions I think may need some work, but I'm in currently in a position to do a table of all the options to validate. Mostly a note for myself / @kaichie to potentially look at this again. Especially when passing the PI line or when dealing with negatives (including potentially making sure that the direction settings are in the correct -PI, PI frame)

nav2_collision_monitor/params/collision_monitor_params.yaml
kaichie: For holonomic platform, I personally believe that it will be more user friendly to define the `linear` magnitude and `direction` rather than two separate ranges in `x` and `y`. It will be valuable to gather insights from the holonomic platform community to determine the more advantageous approach.

Sounds good to add the flag to simplify the process of configuring polygons for non-holonomic robot types.

As for the `direction_min` and `direction_max` configuration, my current concept and implementation assume that the polygon's covered direction always spans from `direction_min` to `direction_max`. With this, we can specify the angle within any range. If we are defining the range from `0 to min` and `0 to max`, then we are not able to specify ranges like the backward direction. For example:

velocityranges

nav2_collision_monitor/src/polygon_velocity.cpp
kaichie: Please see the [comment](https://github.com/ros-planning/navigation2/pull/3708#discussion_r1292861566) here. I think `direction >= direction_max_ || direction <= direction_min_` will not cover the current assumption. But we can discuss the concept first and see if there is a better way to encompass the direction range.

nav2_collision_monitor/src/polygon.cpp
kaichie: I was thinking to validate all the parameters before creating the `VelocityPolygon` object. I think this is do-able, will update to pass the `node_` pointer and handle it's own parameter in the `VelocityPolygon` object.

nav2_collision_monitor/params/collision_monitor_params.yaml
AlexeyMerzlyakov: I am OK with the idea of using `holonomic` parameter which specifies which restrictions are we operating with: polar-based `direction_min/direction_max` + `linear_min/linear_max`; or cartesian-based `vel_x_min/vel_x_max` + `vel_y_min/vel_y_max` for differential and Ackermann/legged drives.
This is some kind of controller-like approach.

From another hand, to not make Collision Monitor to be controller-independent, as it originally conceived, I'd like to propose to make `direction_min`/`direction_max` to be an optional parameters. If one of them was specified, we will use their values. If none of them were specified - check velocity component restrictions only (and thus, do not do `std::atan()` cost-ineffective operations).
Moreover, if we will use polar-based and cartesian-based restrictions at the same time, it could allow us to be even more flexible in making robot velocity application rules (say, forward movement in 15 degree sector, but x-component of speed no less than some minimal value with total ignoring of y-value of velocity).

@kaichie, and yes, the clarification of directions and its signs (clock/counter clock wise rotations) and the pictures from above would be extremely helpful to include it into final documentation.

SteveMacenski: > I personally believe that it will be more user friendly to define the linear magnitude and direction rather than two separate ranges in x and y.

I would agree conceptually in reducing the number of parameters down to a minimum. It would though make configuring a bit more painful, but I don't see a way around that. 

> Sounds good to add the flag to simplify the process of configuring polygons for non-holonomic robot types.

:+1: So only check on the direction bits when `holonomic = true`, else only consider `X` and `theta` vel.

> As for the direction_min and direction_max configuration, my current concept and implementation assume that the polygon's covered direction always spans from direction_min to direction_max. With this, we can specify the angle within any range. If we are defining the range from 0 to min and 0 to max, then we are not able to specify ranges like the backward direction

I don't understand. If we call the direction from -PI to PI with `0` being forward, then reverse would be from -PI/2 to PI/2. This is why I asked in another part of the code review to review how you handle the ranges of direction when checking if the polygon should be applied for wrap around. Just checking `> min` and `< max` doesn't work with angles. 

>  This is some kind of controller-like approach.

I don't think this is a controller approach, this is just an approach that is cognizant of robot kinematic constraints - which this would obviously need to have since different robots move differently and thus would represent their motion abilities to define polygons for separately. If we had to change anything more then `direction` checking, I would ask for an abstraction of the motion models in this node. But while its simple and only 1 change, I think its fine as a global param as to which behavior to use.

kaichie: Just to summarise the proposed ideas here, I think so far these are the ideas that we have agreed on:
- Add `holonomic` flag
- If `holonomic` is `false`: consider velocity `x` within `vel_x_min`/`vel_x_max` and `theta_min/theta_max`

The ideas still ongoing for discussion where `holonomic` is `true` are as follows:
**Option 1:**  consider velocity within `linear_min/linear_max`, `direction_min/direction_max` and `theta_min/theta_max`  
**Option 2:** consider velocity within `vel_x_min/vel_x_max/vel_y_min/vel_y_max` and `theta_min/theta_max`.
**Option 3:** consider velocity within `linear_min/linear_max` and `direction_min/direction_max`(if provided as an optional parameter). Otherwise, consider `vel_x_min/vel_x_max/vel_y_min/vel_y_max`. +  `theta_min/theta_max`

I think **Option 3** would cover most of the concern we have discussed, offering the flexibility to define either within a linear+direction range or using a Cartesian-based range approach. 

For the direction range , lets discuss on the other part of the code review. I will include the table for all the possible range options(probably tomorrow).

SteveMacenski: I think we need to do Option 1. As proposed Option 3 is Option 1 but without Theta, but keep in mind translational velocity in some component direction doesn't preclude rotations at the same time, so we still need to consider a rotational component. 

Option 3 seems to cover 1 and 2 as you mention, but it provides redundent methods of embedding the same information which should be avoided unless there are reasons to have both (e.g. some things you can't model in one you need the other for)

nav2_collision_monitor/src/polygon_velocity.cpp
kaichie: As mentioned in the previous comment, the polygon's covered direction always spans from `direction_min` to `direction_max` in clockwise. 
The direction is calculated with `std::atan2(y, x)`, resulting in an output range of `-π` to `π`.
The direction should be defined as follow:
reference

To validate the logic, the following examples are considered:
direction_ranges
table1

In cases such as 5 and 6, a user could potentially use -π instead of π; however the direction verification should remains `True`:
table2

Here's a basic Python script I used to test the logic:
```python
#!/usr/bin/env python3

import sys

def check_direction_in_range(direction, direction_min, direction_max):
    if direction_min <= direction_max:
        direction_in_range = direction >= direction_min and direction <= direction_max
    else:
        direction_in_range = direction >= direction_min or direction <= direction_max
    return direction_in_range

if len(sys.argv) != 4:
    print("Invalid usage. Usage: python3 script.py direction direction_min direction_max")
    sys.exit()

direction, direction_min, direction_max = map(float, sys.argv[1:4])
result = check_direction_in_range(direction, direction_min, direction_max)
print("Direction is in range:", result)
```

nav2_collision_monitor/params/collision_monitor_params.yaml
kaichie: I will proceed with Option 1 to make the relevant changes. Much appreciated for the discussion and clarification @AlexeyMerzlyakov @SteveMacenski! :pray: I will include the related diagrams in the documentation as discussed.

nav2_collision_monitor/src/polygon_velocity.cpp
kaichie: > So if you were to have something from just forward to left, you'd do it 0.5 PI as min and 0 as max? That doesn't make sense... That's counter-clockwise to do that in a logical way. I assume that's just a typo.

Apologies for the typo; it should be counter-clockwise. Therefore, the forward and left would be 0 as min and 0.5 PI as max.

> Then if you did it always counter-clockwise, then having 0.75 PI be the min and -0.75PI be the max equally makes little sense. I think you need to abandon this dirctionality logic of defining the range.

Okay I believe there might be some confusion regarding the parameter names. Technically, we are trying to determine if an angle is between two angle(range of -PI to PI). A more intuitive way to refer to these parameter names could be `direction_start_angle` and `direction_end_angle`. This naming would help avoid the confusion of having to set the `min` angle numerically greater than the `max` angle. However, we still need the underlying assumption of `direction_start_angle` to `direction_end_angle` is in the counter-clockwise direction. How does this sounds?

As for the logic, I've come across a couple of interesting reads([here](https://stackoverflow.com/a/29721295) and another one [here](https://stackoverflow.com/questions/11406189/determine-if-angle-lies-between-2-other-angles)) that have a similar implementation. I will try if it is possible to `sort them at configuration time` as mentioned. Totally open for any ideas and suggestions!

> But that's not a objective thought, but my subjective opinion. I just see lots of people complaining or misconfiguring how the positive value is somehow the "minimum" where as the negative value is the "maximum" by the range definitions.

I understand your perspective, lets avoid this :100:

SteveMacenski: Perhaps this is OK as long as we document it explicitly. You have that nice graphic now so I suppose if that was linked in the readme/config guide to explain how to configure it, there's little excuse for people to mess it up. I know alot of people **don't** read the docs which is what I was trying to combat against, but I suppose if there's a visual and clear explanation I can point them to, then the overhead on me answering questions is limited to a link. 

Nothing's wrong with your method, I just try to find the most bullet proof way of handling it as possible to head-off user errors / issues. 

For the parameters, I think renaming also helps. Not required but you could make it an array param `translational_direction: [0.5, 0.6]` (or direction or linear direction, I think just being consistent with the other things is key) instead of specifying min/max or start/end explicitly. That might actually help by forcing folks into the docs to understand what it means :-)

SteveMacenski:
Any update? I think we're on the same page now, its just about adding your graphic to the docs so that things are clear to end-users & a trivial renaming of the params

kaichie:
Yes I think we are on the same page now. Sorry I was quite occupied in the last two weeks. I will jump back into this on the coming weekend!

kaichie:
The core features have been updated based on the reviews and discussions above. Documentation and test cases are currently in progress, with more to follow later this week.

SteveMacenski:
Great!

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+
+VelocityPolygon::~VelocityPolygon()
+{
+  RCLCPP_INFO(logger_, "[%s]: Destroying VelocityPolygon", velocity_polygon_name_.c_str());
+  poly_.clear();
+}
+
+
+bool VelocityPolygon::getParameters()
```
SteveMacenski, line 41: Some of the line extension breaks in this function for the getting / declaring of parameters are pretty awkward (e.x. `linear_max`), please take a look at this and make more readable and don't use excessive breaks. This is probably just ament_uncrustify's automatic response

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,8 +169,46 @@ bool Polygon::isShapeSet()
...
 
-void Polygon::updatePolygon()
+bool Polygon::isUsingVelocityPolygonSelector()
 {
+  return !velocity_polygons_.empty();
+}
+
+void Polygon::updatePolygon(const Velocity & cmd_vel_in)
+{
+  if (isUsingVelocityPolygonSelector()) {
```
SteveMacenski, line 179: I'm trying to figure out here if we have this master bool `isUsingVelocityPolygonSelector` why then we have the VelocityPolygon inside of the Polygon itself and not be a separate class object that has the base type of Polygon to be used in the application separately (or have a Polygon composed object within the VelocityPolygon class). At the end of this `if` condition, we either `return` from getting a polygon in range or logging a warning that there is none, but we never actually enter the logic of the rest of this function.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,8 +169,46 @@ bool Polygon::isShapeSet()
...
+        polygon_.polygon.points.clear();
+        for (const Point & p : poly_) {
+          geometry_msgs::msg::Point32 p_s;
+          p_s.x = p.x;
+          p_s.y = p.y;
+          // p_s.z will remain 0.0
+          polygon_.polygon.points.push_back(p_s);
+        }
+
+        return;
```
SteveMacenski, line 196: So... what happens if there are multiple polygons that apply to the range potentially?

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: Can the velocity polygon not leverage the subscriptions to change the footprint values that are implemented below? Or setting in different reference frames?

I think some thought needs to be placed into these other features w.r.t. the velocity polygon addition. I don't think there's any philosophical reason that we cannot have the velocity polygon's polygon be adjustable from the subscription like any other polygon?

SteveMacenski (changes requested):

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -422,22 +446,48 @@ bool Polygon::getParameters(
...
+        logger_,
+        "[%s]: Polygon topic are not defined. Using velocity polygon instead.",
+        polygon_name_.c_str());
     }
+
+
+    // Velocity polygon will be used
+    nav2_util::declare_parameter_if_not_declared(
+      node, polygon_name_ + ".velocity_polygon", rclcpp::PARAMETER_STRING_ARRAY);
+    std::vector<std::string> velocity_polygon_array = node->get_parameter(
```
SteveMacenski, line 478: If there is no prameter set, this can throw an exception when there are no velocity polygons specified. This is why the other `get`s are under the try/catch that this also needs to be placed under above.

nav2_collision_monitor/src/polygon.cpp
AlexeyMerzlyakov: This might be related to https://github.com/ros-planning/navigation2/pull/3708#discussion_r1272210118 thread discussion. If so, I thing we agreed to select first suitable VelocityPolygon in the list and reflect this rule in the documentation

nav2_collision_monitor/CMakeLists.txt
```diff
@@ -57,6 +58,7 @@ add_library(${monitor_library_name} SHARED
 add_library(${detector_library_name} SHARED
   src/collision_detector_node.cpp
   src/polygon.cpp
+  src/velocity_polygon.cpp
```
AlexeyMerzlyakov, line 61: It seems that VelocityPolygons are not going to be added to a CollisionDetector. Thus, it seems that source-file is unnecessary for CD, only used for CM.

nav2_collision_monitor/src/polygon.cpp
AlexeyMerzlyakov: That is really good question, we've previously touched in the original ticket, I recall. One way of the solution that was to make separate module, publishing velocities for CM node, but this idea (for some reasons, I do not recall, maybe performance) we've finally refused. Your idea to make `VelocityPolygon` class to be inherited from `Polygon` seems not much crazy for me. We will have two `updatePolygon()` methods that could be generalized, and depending on some conditions (like velocity), could react and update its shape. Great idea, especially in case of if we:
1. Want to have some unified API of CM & its shapes for future re-usage
2. Refuse the number of if-else cases not related to robot behavior in CM main cycle logic (that I don't like myself, but anyway, it is working solution :smile:, that need be moved into behavior handling logic in a foreseen future).

@kaichie, what do you think about it?

AlexeyMerzlyakov (commented):
Review in progress. More comments will arrive shortly

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: The other option is to have dynamic polygon setting support within the velocity polygon to set the different velocity polygons to different values. Its less "ooooh nice design" but would be more straight forward. 

If making the VelocityPolygon derive from Polygon would make things cleaner, that's preferable in my book, but I'm not 100% convinced that wouldn't just be a huge headache and still not fully solve the issue.

kaichie: My thought on this are as follows: if VelocityPolygons support subscriptions to change polygons, wouldn't this be the same as using a normal polygon with the subscription topic? Unless the use case involves users preferring to set the polygon using the topic, even if it's a static polygon.

Conceptually, VelocityPolygon is quite different from the Polygon base class as it contains multiple sub-polygons. Even if it is a derived class of Polygon, I believe it still needs the if-else cases, as CM always checks for the topic first, followed by static points, and finally the VelocityPolygon. But yes then we can use `this->` instead. Or maybe I did not get this part correctly?

If I understand correctly, we are also discussing whether we need to derive VelocityPolygon's sub-polygons from the Polygon base class so that it can support the dynamic changing of polygons with subscriptions. It makes sense to derive it from the Polygon class, but there will likely be some code duplication too (for example, the logic to decide whether to use the topic or static points will be duplicated in the VelocityPolygon class itself). I would be happy to draft a version of this if it helps in the discussion.

SteveMacenski: I agree, I think deriving is not the right answer to get this behavior. The interface for the subscriber would have to be different to specify the name to change - else we'd have to create `N` topics where the polygon name was in the topic name. That's doable, but unwieldy. 

I have two separate topics:
- (1) We skip an entire and massive block of code creating a giant switch in a core function
- (2) The code that the skipped block of code represents is not represented in the VelocityPolygon case and might as well be

I think (2) can be added as a service call or topic within the VelocityPolygon code. (1) I think might just need different abstractions (e.g. break into functions or make explicitly a switch statement) to make clear that its a binary operation being done

AlexeyMerzlyakov: @SteveMacenski, I am not sure what did you mean in the latest message.
However, the following idea:

> The other option is to have dynamic polygon setting support within the velocity polygon to set the different velocity polygons to different values.

as I understand well - is to make a type of shapes, called `VelocityPolygon`. This shape might not be derived from `Polygon`, as you've previously mentioned for concept and convenience issues, but rather to be a separate parallel instance existing along with usual `Polyhons` and `Circles` (but again, not derived from them). Am I correct, or you meant something else?

AlexeyMerzlyakov: Or do you mean from one `PolygonVelocity` -> to make N sub-polygons of `Polygon` type (and add them into main `polygons_` array); and then enable/disable these sub-polygons depending on robot velocity by using `PolygonVelocity` as a polygons main brain / handler?

SteveMacenski: I'm suggesting that we change nothing about this PR but:
- VelocityPolygon now when it loads the different velocity regime shapes, exposes a service / topic to change them dynamically via a ROS interface
- Inside of `updatePolygon` we break up this function into 2 separate functions in a `switch()` statement to make it clear that there are two, mutually exclusive paths that are done depending on the nature of the polygon. 

@kaichie does that seem straight forward enough? Lets just try to get this in before ROSCon!

kaichie: Yeap sounds good to me. For each of the sub-polygon in the VelocityPolygon, I will expose it with a topic. Default to `/collision_monitor/set_velocity_polygon/<sub-polygon name>`, unless it is specified in the parameter config(`polygon_sub_topic`). I think this part will be quite similar with the one in `Polygon` base class.

SteveMacenski: OK, works for me. That's alot of topics, but it would be nice to be able to use the Polygon message still instead of having to create a custom one to add a name field. 

Maybe naming as `/collision_monitor/<polygon name>/set_polyon`?

AlexeyMerzlyakov: @SteveMacenski, thank you for the explanation. This also seems good to me: not change too much in existing PR, and make it to look more logical and usable (topics or services-based handling). Let's make next round of review after update then.

nav2_collision_monitor/CMakeLists.txt
kaichie: I think this is needed because velocity_polygon is included in polygon.cpp

nav2_collision_monitor/src/polygon.cpp
kaichie: Added the dynamic update from topic for velocity polygon(quite abit of duplicated code with the polygon base class), and replaced the huge if-else flow with switch case for handling `updatePolygon()`.

SteveMacenski:
@kaichie please make sure to scroll up and address all the remaining comments - there appear to be several that no changes were made to address

nav2_collision_monitor/include/nav2_collision_monitor/velocity_polygon.hpp
```diff
@@ -0,0 +1,134 @@
...
+   */
+  void updatePolygon(geometry_msgs::msg::PolygonStamped::ConstSharedPtr msg);
+
+  // Basic parameters
+  /// @brief Points of the polygon
+  std::vector<Point> poly_;
+  /// @brief Name of polygon
+  std::string polygon_name_;
+  /// @brief Topic for dynamic polygon subscriber
+  std::string polygon_sub_topic_;
```
SteveMacenski, line 103: Remove parameter, make it set to what we discussed in the comments. If someone wants to remap, they can always do that at launch time. There hits a point where its not beneficial to saturate people with parameters they'll likely never change

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -42,11 +42,42 @@ collision_monitor:
     FootprintApproach:
       type: "polygon"
       action_type: "approach"
-      footprint_topic: "/local_costmap/published_footprint"
+      # footprint_topic: "/local_costmap/published_footprint"
```
SteveMacenski, line 45: Commented out?

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -42,11 +42,42 @@ collision_monitor:
...
       time_before_collision: 2.0
       simulation_time_step: 0.1
       min_points: 6
-      visualize: False
+      visualize: True
+      polygon_pub_topic: "polygon_approach"
+      velocity_polygons: ["rotation", "translation_forward", "translation_backward", "stop"] # third priority
```
SteveMacenski, line 51: Why would this be under approach footprint? this doesn't make sense

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
```diff
@@ -157,6 +158,14 @@ class Polygon
    */
   void publish();
 
+  /**
+   * @brief Set the polygon shape based on the polygon points
+   * @param poly_points Polygon points in the format of x1, y1, x2, y2 ...
+   * @param poly Output polygon points (vertices)
+   * @return True if successfully set the polygon shape
+   */
+  static bool setPolygonShape(std::vector<double> & poly_points, std::vector<Point> & poly);
```
SteveMacenski, line 167: This doesn't ever appear to use its staticness?

nav2_collision_monitor/include/nav2_collision_monitor/types.hpp
```diff
@@ -79,6 +79,15 @@ struct Action
   std::string polygon_name;
 };
 
+/// @brief Source for setting the polygon
+enum PolygonSource
```
SteveMacenski, line 83: enum class

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,239 @@
...
+    if (!Polygon::setPolygonShape(polygon_points, poly_)) {
+      RCLCPP_ERROR(
+        logger_,
+        "[%s][%s]: Polygon has incorrect points description",
+        polygon_name_.c_str(), velocity_polygon_name_.c_str());
+      return false;
+    }
+
+    // holonomic param
+    nav2_util::declare_parameter_if_not_declared(
```
SteveMacenski, line 68: Move this down with the other kinematic params

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,51 +170,95 @@ bool Polygon::isShapeSet()
...
-      tf2::Vector3 p_v3_b = tf_transform * p_v3_s;
+  switch (polygon_source) {
+    case STATIC_POINTS:
+      {
+        if (!polygon_.header.frame_id.empty() && polygon_.header.frame_id != base_frame_id_) {
+          // Polygon is published in another frame: correct poly_ vertices to the latest frame state
+          std::size_t new_size = polygon_.polygon.points.size();
+
+          // Get the transform from PolygonStamped frame to base_frame_id_
+          tf2::Transform tf_transform;
```
SteveMacenski, line 183: Breakout into new method `transformPolygon(polygon)`

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,51 +170,95 @@ bool Polygon::isShapeSet()
...
+          }
+        }
+        break;
+      }
+    case DYNAMIC_SUB:
+      {
+        // Get latest robot footprint from footprint subscriber
+        std::vector<geometry_msgs::msg::Point> footprint_vec;
+        std_msgs::msg::Header footprint_header;
+        footprint_sub_->getFootprintInRobotFrame(footprint_vec, footprint_header);
```
SteveMacenski, line 210: You don't check if dereferencing nullptr. I don't see anywhere where `footprint_sub_` and the dynamic sub mode are linked such that we have any promises

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,51 +170,95 @@ bool Polygon::isShapeSet()
...
-    }
-
-    // Correct main poly_ vertices
-    poly_.resize(new_size);
-    for (std::size_t i = 0; i < new_size; i++) {
-      // Transform point coordinates from PolygonStamped frame -> to base frame
-      tf2::Vector3 p_v3_s(polygon_.polygon.points[i].x, polygon_.polygon.points[i].y, 0.0);
-      tf2::Vector3 p_v3_b = tf_transform * p_v3_s;
+  switch (polygon_source) {
+    case STATIC_POINTS:
```
SteveMacenski, line 176: Why isn't static points the default?

SteveMacenski (changes requested):
I don't think this is complete? I don't see where the different polygon sources are properly set or initialized for all edge cases. I wish you hadn't changed so much and only just implemented the dynamic velocity polygons internally, this is a much more complicated PR now

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,239 @@
...
+        0.0));
+    theta_min_ =
+      node->get_parameter(polygon_name_ + "." + velocity_polygon_name_ + ".theta_min").as_double();
+
+  } catch (const std::exception & ex) {
+    RCLCPP_ERROR(
+      logger_,
+      "[%s][%s]: Error while getting polygon parameters: %s",
+      polygon_name_.c_str(), velocity_polygon_name_.c_str(), ex.what());
+    return false;
```
SteveMacenski, line 153: Nothing of substance happens here, you simply drop the error on the floor and continue as if nothing happened!

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,239 @@
...
+    std::vector<double> polygon_points = node->get_parameter(
+      polygon_name_ + "." + velocity_polygon_name_ + ".points").as_double_array();
+
+    if (!Polygon::setPolygonShape(polygon_points, poly_)) {
+      RCLCPP_ERROR(
+        logger_,
+        "[%s][%s]: Polygon has incorrect points description",
+        polygon_name_.c_str(), velocity_polygon_name_.c_str());
+      return false;
+    }
```
SteveMacenski, line 65: Why aren't you setting this as a static footprint mode?

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: Also I'd think that a bunch of this can be made into a util and shared between mutliple instance where this is done

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: Where is this set even for the velocity polygon?!

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -422,22 +454,49 @@ bool Polygon::getParameters(
...
+        polygon_sub_topic =
+          node->get_parameter(polygon_name_ + ".polygon_sub_topic").as_string();
+        return true;
+      } else if (action_type_ == APPROACH) {
+        // Obtain the footprint topic to make a footprint subscription for approach polygon
+        nav2_util::declare_parameter_if_not_declared(
+          node, polygon_name_ + ".footprint_topic", rclcpp::PARAMETER_STRING);
+        footprint_topic =
+          node->get_parameter(polygon_name_ + ".footprint_topic").as_string();
+        polygon_source = DYNAMIC_SUB;
```
SteveMacenski, line 473: Huh?? This has nothing to do with approach

SteveMacenski (changes requested):
I believe if I'm not misunderstanding that you should revert all of your recent changes since the last review. It introduces so many bugs and potential misunderstandings that its not reviewable.

All I asked was for the velocity polygons to be dynamically updatable, this reformatting adds restrictions and enums that aren't properly implemented it doesn't seem with a number of incorrect or unhandled edge cases. 

I'm not sure I can review this as it stands. The flaws are so large and subtle that I'm bound to miss them myself.

SteveMacenski:
@doisyg I think you need to review some of this with @kaichie internally

nav2_collision_monitor/src/velocity_polygon.cpp
SteveMacenski: How are you deciding if its a static footprint mode vs if a message has come in yet on the topic for the subscription's footprint to be useful?

nav2_collision_monitor/src/velocity_polygon.cpp
kaichie: Yeah most of them are from `ament_uncrustify --reformat` (is there any dedicated formatter to use for this repo?). I have updated most of them manually now.

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
kaichie: This is used in PolygonVelocity [here](https://github.com/kaichie/navigation2/blob/b97cbb7a490de8c928650ebf74e5cb3f104b1ce5/nav2_collision_monitor/src/velocity_polygon.cpp#L57) and in the polygon class [here](https://github.com/kaichie/navigation2/blob/b97cbb7a490de8c928650ebf74e5cb3f104b1ce5/nav2_collision_monitor/src/polygon.cpp#L434). Or should I create a new `polygon_utils.hpp` for this?

nav2_collision_monitor/params/collision_monitor_params.yaml
kaichie: This params file is currently modified to run the CM example node with velocity_polygon (for ease of testing without any changes). To discuss if we want to use velocity_polygon by default for the example node, or else I will revert this comment and remove "FootprintApproach" from the `polygons` too.

nav2_collision_monitor/params/collision_monitor_params.yaml
kaichie: This feature is available for all action type. The logic is [here](https://github.com/kaichie/navigation2/blob/b97cbb7a490de8c928650ebf74e5cb3f104b1ce5/nav2_collision_monitor/src/polygon.cpp#L470), if `polygon_sub_topic`(`STOP`, `SLOWDOWN`, `LIMIT`, `DO_NOTHING`) or `footprint_topic`(`APPROACH`) is not available, it will try to use the velocity polygon feature. This follows the pattern of the original design.
For real world use case, we do use the dynamic update of polygon via the `footprint_topic` for `APPROACH` model.

nav2_collision_monitor/src/velocity_polygon.cpp
kaichie: fixed with proper return of failure case

nav2_collision_monitor/src/velocity_polygon.cpp
kaichie: Added the similar try catch pattern as in the `polygon` class for waiting the subscription topic. In current design, the sub polygon will always expose the topic for reconfiguring, do we want to skip the topic subscription if `points` are provided in the param file?

kaichie:
@SteveMacenski Sorry for the confusion, I misunderstood some of the internal logic and the state changes was incorrectly implemented. I have reverted the state switch case and addressed the comments.

I will wait for another round of review/discussion before making the changes((It should be literally as described below):
> Inside of updatePolygon we break up this function into 2 separate functions in a switch() statement to make it clear that there are two, mutually exclusive paths that are done depending on the nature of the polygon.

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
SteveMacenski: Or just add to `utils.hpp` file

nav2_collision_monitor/params/collision_monitor_params.yaml
SteveMacenski: Lets make the demo file of this based on the stop zones, that's how I believe most users will be using it. 

It doesn't seem good form to me to do this in the approach style. The approach takes into account the **actual** footprint, so there's no need to artificially inflate or adjust it based on the velocity profile. Its already only projecting in the velocity directions. I would not recommend that you use it that way either.

nav2_collision_monitor/params/collision_monitor_params.yaml
SteveMacenski: Dont remove anything that was already here.

SteveMacenski:
Update

```
if (!footprint_sub_->getFootprintInRobotFrame(footprint_vec, footprint_header)) {return;}
```

so that if we don't get any footprint yet, we don't try overriding the default one. I believe this is a bug, not your bug, but something that we should resolve here

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -422,22 +446,48 @@ bool Polygon::getParameters(
...
-      polygon_sub_topic =
-        node->get_parameter(polygon_name_ + ".polygon_sub_topic").as_string();
-    } else if (action_type_ == APPROACH) {
-      // Obtain the footprint topic to make a footprint subscription for approach polygon
-      nav2_util::declare_parameter_if_not_declared(
-        node, polygon_name_ + ".footprint_topic",
-        rclcpp::ParameterValue("local_costmap/published_footprint"));
-      footprint_topic =
-        node->get_parameter(polygon_name_ + ".footprint_topic").as_string();
+    try {
```
SteveMacenski, line 449: I don't think this handling of try/catch is correct. Why would we assume because there's no subscription topic that it means we're using the velocity polygons?

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -422,22 +446,48 @@ bool Polygon::getParameters(
...
+        logger_,
+        "[%s]: Polygon topic are not defined. Using velocity polygon instead.",
+        polygon_name_.c_str());
     }
+
+
+    // Velocity polygon will be used
+    nav2_util::declare_parameter_if_not_declared(
+      node, polygon_name_ + ".velocity_polygons", rclcpp::PARAMETER_STRING_ARRAY);
+    std::vector<std::string> velocity_polygons = node->get_parameter(
```
SteveMacenski, line 478: This will throw

SteveMacenski (changes requested):

SteveMacenski:
I think we need to have a conversation on Thursday about what variables to initialize or what to throw an exception if not set. And, when we do either of those operations, what does that actually mean about a user configuration, if anything. It seems to me there are some mistakes w.r.t. the polygon, polygon topic, and velocity polygon handling of these -- then also checking what happens in the global system once something is thrown or not thrown and returns false to make sure we don't have edge cases unhandled for any arbitrary but valid user configuration

AlexeyMerzlyakov (commented):
Please check my comments for the logic update/simplification stage below

SteveMacenski:
@kaichie what's the story here? I'd really like to get this in ASAP rather than hanging around indefinitely

nav2_collision_monitor/src/polygon.cpp
kaichie: If I understand correctly, in the current design, it is using the try-catch method -- if static points are defined, it will [return early](https://github.com/ros-planning/navigation2/blob/be370db6ea74280eff7b2075b9f31cc6eca56d31/nav2_collision_monitor/src/polygon.cpp#L429C7-L431).  if static points are not defined, it will throw a `ParameterUninitializedException` and the [catch block](https://github.com/ros-planning/navigation2/blob/be370db6ea74280eff7b2075b9f31cc6eca56d31/nav2_collision_monitor/src/polygon.cpp#L432-L437) will log a message and skip to the next part to initialise the parameters for dynamic subscriptions.

In the changes here, I follow the same pattern -- if `polygon_sub_topic` or `footprint_topic`  is defined, it will return early. if `polygon_sub_topic` or `footprint_topic` is not defined, the catch block will skip to the next part to initialise the parameters for velocity-based polygon.

SteveMacenski: This logic needs to be adjusted to check the nature of the static footprint information -- whether we have a static single footprint or we have the parameters associated with the velocity polygons to use. If neither of those are true, then we fail. We can check this explicitly. 

I think its worth from this pause in the development of this feature to take a pause and potentially rethink areas like this that with the new feature might just need to be refactored how they work now that we're adding in a 3rd modality. Think more openly about the best way to do this explicitly, not relying on cascading failures. We know what params might be filled in for each mode -- we should be able to check which mode is set by them (and if multiple params set making ambiguous, fail the configuration)

kaichie:
@SteveMacenski @AlexeyMerzlyakov I have rewrite the `VelocityPolygon` class, using `Polygon` as it's base class and will be added as a new type -- `velocity_polygon`. All parameter reading and logic are now consolidated within the same file, which looks much cleaner. I have also updated the code based on the feedback and reviews.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -390,30 +429,15 @@ bool Polygon::getParameters(
         node, polygon_name_ + ".points", rclcpp::PARAMETER_DOUBLE_ARRAY);
       std::vector<double> poly_row =
         node->get_parameter(polygon_name_ + ".points").as_double_array();
-      // Check for points format correctness
-      if (poly_row.size() <= 6 || poly_row.size() % 2 != 0) {
+      // Do not need to proceed further, if "points" parameter is defined.
+      // Static polygon will be used.
```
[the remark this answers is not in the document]
kaichie: not relevant anymore in the latest changes.

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -168,8 +169,46 @@ bool Polygon::isShapeSet()
...
+    // Log for uncovered velocity
+    auto node = node_.lock();
+    if (!node) {
+      throw std::runtime_error{"Failed to lock node"};
+    }
+    RCLCPP_WARN_THROTTLE(
+      logger_,
+      *node->get_clock(), 2.0, "Velocity is not covered by any of the velocity polygons. x: %.3f y: %.3f tw: %.3f ",
+      cmd_vel_in.x, cmd_vel_in.y, cmd_vel_in.tw);
+    return;
```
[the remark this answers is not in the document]
kaichie: There is no "default" polygon to be used, but to achieve the same functionality, the user can create a polygon that cover the full range of the robot velocity and rotation/direction. (this is also why I added `stop` in the `velocity_polygons` parameter in the default config -- to be added as recommendation in the documentation)

nav2_collision_monitor/src/polygon.cpp
kaichie: not relevant anymore in the latest changes.

nav2_collision_monitor/src/polygon.cpp
kaichie: not relevant anymore in the latest changes.

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
kaichie: not relevant anymore in the latest changes.

nav2_collision_monitor/params/collision_monitor_params.yaml
kaichie: reverted and added a sample for `velocity_polygon` type(not used by default)

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -14,6 +14,7 @@ collision_monitor:
     # Footprint could be "polygon" type with dynamically set footprint from footprint_topic
     # or "circle" type with static footprint set by radius. "footprint_topic" parameter
     # to be ignored in circular case.
+    # or "velocity_polygon" type with dynamically set polygon from velocity_polygons
```
SteveMacenski, line 17: Adjust this comment

> Footprint could be "polygon" type with  ...

to be numbered (1) (2) (3) for the 3 options instead of tacking on the extra "or" fragment

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -51,6 +52,41 @@ collision_monitor:
...
+        linear_max: 1.0
+        theta_min: -1.0
+        theta_max: 1.0
+      translation_backward:
+        points: "[[0.2, 0.3], [0.2, -0.3], [-0.35, -0.3], [-0.35, 0.3]]"
+        linear_min: -1.0
+        linear_max: 0.0
+        theta_min: -1.0
+        theta_max: 1.0
+      stop:
```
SteveMacenski, line 84: `stopped` -> just to be clear its not a "stop"ing polygon, but a polygon to apply when stopped

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -51,6 +52,41 @@ collision_monitor:
...
+        points: "[[0.25, 0.25], [0.25, -0.25], [-0.25, -0.25], [-0.25, 0.25]]"
+        linear_min: -1.0
```
SteveMacenski, line 86: These linear / angular vleocity limits seem not like stopped ranges?

nav2_collision_monitor/params/collision_monitor_params.yaml
```diff
@@ -51,6 +52,41 @@ collision_monitor:
       min_points: 6
       visualize: False
       enabled: True
+    VelocityPolygonApproach:
+      type: "velocity_polygon"
+      action_type: "approach"
```
SteveMacenski, line 57: Is this really using the approach model? I thought these were stop polygons? Is the approach type even possible with this velocity polygon? I'm pretty sure its just for stop / slowdown,

nav2_collision_monitor/include/nav2_collision_monitor/velocity_polygon.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+namespace nav2_collision_monitor
+{
+
+/**
+ * @brief Velocity polygon class.
+ * This class contains all the points of the polygon and
+ * the expected condition of the velocity based polygon.
+ */
+class VelocityPolygon : public Polygon
```
SteveMacenski, line 36: Its worth carefully looking at the `Polygon` API methods and making sure that they're all applicable to this new velocity polygon type (else, mark them virtual for this class to override appropriately). Many will be called by the application layer

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,195 @@
...
+        node, polygon_name_ + "." + velocity_polygon_name + ".points", rclcpp::PARAMETER_STRING);
+      std::string poly_string =
+        node->get_parameter(polygon_name_ + "." + velocity_polygon_name + ".points").as_string();
+
+      if (!getPolygonFromString(poly_string, poly)) {
+        return false;
+      }
+
+      // linear_min param
+      double linear_min;
```
SteveMacenski, line 74: For linear & theta min, max, no `0.0` defaults, do the same `PARAMETER_DOUBLE` that they need to be  specified. The min/max angle make sense to have defaulted to `-PI, PI` so it covers all angles unless otherwise specified, don't change those.

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,195 @@
...
+        direction_start_angle =
+          node
+          ->get_parameter(polygon_name_ + "." + velocity_polygon_name + ".direction_start_angle")
+          .as_double();
+      }
+
+      SubPolygonParameter sub_polygon = {
+        poly, velocity_polygon_name, linear_min, linear_max, theta_min,
+        theta_max, direction_end_angle, direction_start_angle};
+      sub_polygons_.emplace_back(sub_polygon);
```
SteveMacenski, line 128: You can emplace back with ` poly, velocity_polygon_name, linear_min, linear_max, theta_min,
        theta_max, direction_end_angle, direction_start_angle`. There's no need for the object instantiation

nav2_collision_monitor/src/velocity_polygon.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+  }
+
+  // Log for uncovered velocity
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+  RCLCPP_WARN_THROTTLE(
+    logger_, *node->get_clock(), 2.0,
```
SteveMacenski, line 165: Store `clock_` in class definition so you don't need to lock the node every time we call this. Do `node->get_clock()` in the get params method

SteveMacenski (changes requested):
@kaichie pull in `main`, any remaining linting failure are then things this PR introduces. Obviously, we need to address test coverage of this feature but its understandable that you haven't gotten there yet given the level of rewriting we've asked for. You should prioritize this now. 

Very close now :-)

nav2_collision_monitor/params/collision_monitor_params.yaml
kaichie: "stopped" is the final sub-polygon in `velocity_polygons`, it is recommended to cover the entire range of the robot's linear and angular limits, serving as the fallback if velocities are not covered in any other case.

nav2_collision_monitor/include/nav2_collision_monitor/velocity_polygon.hpp
kaichie: the only method that need to be override in this new velocity polygon is `updatePolygon()`, which is updated in `Polygon` class to make it virtual.

SteveMacenski:
- [ ] Docs: migration guide, configuration guide, tutorial, and readme
- [ ] Docs: explain the open comments on the direction for omni, definitions of angles using the table you made, etc
- [ ] A nice video

---

Then we can move on to the other PRs
- [ ] Approach model using dynamics for slowing down based on decelleration distances
- [ ] Approach model report in status message if the actual robot footprint (t = 0) is in collision causing a deadlock
- [ ] Option to use cmd_vel or odom for each velocity polygon

nav2_collision_monitor/README.md
```diff
@@ -48,7 +50,7 @@ The data may be obtained from different data sources:
 The Collision Monitor is designed to operate below Nav2 as an independent safety node.
 This acts as a filter on the `cmd_vel` topic coming out of the Controller Server. If no such zone is triggered, then the Controller's `cmd_vel` is used. Else, it is scaled or set to stop as appropriate.
 
-The following diagram is showing the high-level design of Collision Monitor module. All shapes (Polygons and Circles) are derived from base `Polygon` class, so without loss of generality we can call them as polygons. Subscribed footprint is also having the same properties as other polygons, but it is being obtained a footprint topic for the Approach Model.
+The following diagram is showing the high-level design of Collision Monitor module. All shapes (`Polygon`, `Circle` and `VelocityPolygon`) are derived from base `Polygon` class, so without loss of generality we can call them as polygons. Subscribed footprint is also having the same properties as other polygons, but it is being obtained a footprint topic for the Approach Model.
```
SteveMacenski, line 53: Adding your dexory gif somewhere in here would be nice

SteveMacenski (approved):
@doisyg please review 

@kaichie your test coverage is being picked up now, that's all good

doisyg:
Finally tested and looks good to me! Great work!
Once thing that could be improved (probably as a separate PR) is an example of how to configure a `velocity_polygon` with `action_type: "limit"`. When trying, I naively filled `linear_limit` and `angular_limit` in the sub-polygon parameter spaces (with no effect). Maybe worth considering allowing different limits for different sub-polygon ?

PR #3716, opened 2023-07-27
nav2_behaviors: export all available plugins

DylanDeCoeyer-Quimesis:
It allows external packages to include those headers and create child classes through inheritance.

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


SteveMacenski:
Why not simply adjust the cmake to export those headers as well?

DylanDeCoeyer-Quimesis:
> Why not simply adjust the cmake to export those headers as well?

Isn't it more conventional to place all exported headers in the "include" folder? I took inspiration from the `nav2_behavior_tree` directory.

SteveMacenski:
Fair enough. The headers probably don't belong where they are now in the `nav2_behaviors` package - that's that way for primarily historical reasons that I (apparently) never got to resolving :smile: 

Please resolve the merge conflict and I'll merge

SteveMacenski (approved):

PR #3692, opened 2023-07-11
Install nav2_behavior_tree utils outside of BUILD_TESTING

AadityaRavindran:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/3683 |
| Primary OS tested on | Ubuntu |
| Robotic platform tested on | Built with `--cmake-args "-DBUILD_TESTING=OFF"` and verified that `install/nav2_behavior_tree/include` contains the utils folder |

---

## Description of contribution in a few bullet points

- Moved nav2_behavior_tree/test/utils/ installs to be outside of BUILD_TESTING

## Description of documentation updates required from your changes

---

#### For Maintainers: 
- [ ] Check that any new parameters added are updated in navigation.ros.org
- [ ] Check that any significant change is added to the migration guide
- [ ] Check that any new features **OR** changes to existing behaviors are reflected in the tuning guide
- [ ] Check that any new functions have Doxygen added
- [ ] Check that any new features have test coverage
- [ ] Check that any new plugins is added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

Issue #3683, opened 2023-07-07
Exposing test utility headers for use in other packages


paulbovbel:
#### Feature description
While creating tests for internal behaviours, we've reused some of the handy utility helpers in https://github.com/ros-planning/navigation2/tree/main/nav2_behavior_tree/test/utils. However, since those headers are behind a BUILD_TESTING flag, they're not installed in binary distributions, and so aren't possible to use in tests from another source package. It would be great to find a way to make those reusable in other packages.

#### Implementation considerations
IMO there's basically two approaches here:
- install the test/utility headers regardless of BUILD_TESTING. This leaks those headers outside of a test builds, but.. it's just a few headers
- Create a ~nav2_test_utils type of package, that contains these test utils, is a test_depends for nav2_behavior_tree and other packages that require these kinds of utilities.
- Suboptimally nothing, and we can duplicate the headers in our test packages, for better or worse.

SteveMacenski:
I’d say option 1, should merge that PR a few hours after submission

PR #3718, opened 2023-07-27
Move nav2_behavior_tree test utils into the right folder

AadityaRavindran:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [#3683](https://github.com/ros-planning/navigation2/issues/3683) |
| Robotic platform tested on | Built with --cmake-args "-DBUILD_TESTING=OFF" and verified that install/nav2_behavior_tree/include contains the utils folder |

---

- The original PR https://github.com/ros-planning/navigation2/pull/3692 only worked if `--merge-install` wasn't used to build. Moving the test utils to it's own test/include/<package name> folder not only makes it comply with ros2 package standards, but also make this work with and without `--merge-install`
- Back port to iron is needed.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/CMakeLists.txt
+++ b/nav2_behavior_tree/CMakeLists.txt
@@ -224,7 +224,7 @@ install(DIRECTORY include/
 )
 
 install(DIRECTORY test/utils/
-        DESTINATION include/utils/
+        DESTINATION include/${PROJECT_NAME}/utils/
 )
 
 install(FILES nav2_tree_nodes.xml DESTINATION share/${PROJECT_NAME})


SteveMacenski (approved):
Please ensure you properly test this time and mark this as ready for merging once ready (this is a draft)

SteveMacenski:
These file paths are totally nonsensical: https://github.com/ros-planning/navigation2/tree/35a0f6b8cd4bae1d023b5bd163b24d8485467443/nav2_behavior_tree/test/include/nav2_behavior_tree/utils

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -223,8 +223,8 @@ install(DIRECTORY include/
   DESTINATION include/
 )
 
-install(DIRECTORY test/utils/
-        DESTINATION include/utils/
+install(DIRECTORY test/include/
```
SteveMacenski, line 226: Revert, there's no reason to have an `include` inside of the `test` directory. Utils is correct.

AadityaRavindran:
> Please ensure you properly test this time and mark this as ready for merging once ready (this is a draft)

Yes. I have the change active in the locus build farm. I have tested this locally with and without `--merge-install` but I will wait for the locus pipeline to finish and succeed before marking this as ready.

nav2_behavior_tree/CMakeLists.txt
AadityaRavindran: Actually, if you consider the include path in src, it is `package_name/src/include/package_name/header.hpp`.
Which installs to
- `ros2_workspace_dir/install/package_name/package_name/header.hpp` if `--merge-install` is used with colcon build.
- `ros2_workspace_dir/install/package_name/include/package_name/header.hpp` if `--merge-install` is absent.

Therefore, by adding an include directory in test, we ensure that the utils package installs to
- `install/nav2_behavior_tree/utils` instead of `install/utils` when `--merge-install` is used
- `install/nav2_behavior_tree/include/nav2_behavior_tree/utils` instead of `install/nav2_behavior_tree/include/utils/` when `--merge-install` is absent

SteveMacenski: This directory schema seems nonsensical to me and I see no other examples (e.g. rclcpp, moveit2) of this structure being used. The install rule you simply update from `utils` to `include`, so there should be little difference here if we just name it `utils`. I don't see why it wouldn't be possible to still have it appear in `install/nav2_behavior_tree/test/utils` if we set the destination in the install rule. I don't see why 

> install/nav2_behavior_tree/include/nav2_behavior_tree/utils instead of install/nav2_behavior_tree/include/utils/ when --merge-install is absent

is preferable.

SteveMacenski (approved):
Much better :smile:

PR #3719, opened 2023-07-28
Fix Goal updater QoS

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

* Updated QoS of GoalUpdater so that the latest msg is retrieved. The problem is that when we have a QoS of history depth more than 1,  `spin_some()`  will only retrieve one message in the queue. I actually expected it to retrieve all messages available at the moment it's called but apparently not -> `spin_all()` does that. Not sure what's the difference between `spin_once` and `spin_some` is that this point

* Added some tests

---

#### For Maintainers: 


nav2_behavior_tree/test/plugins/decorator/test_goal_updater_node.cpp
```diff
@@ -95,42 +95,101 @@ TEST_F(GoalUpdaterTestFixture, test_tick)
...
   geometry_msgs::msg::PoseStamped goal;
   goal.header.stamp = node_->now();
   goal.pose.position.x = 1.0;
   config_->blackboard->set("goal", goal);
 
-  // tick until node succeeds
-  while (tree_->rootNode()->status() != BT::NodeStatus::SUCCESS) {
-    tree_->rootNode()->executeTick();
-  }
-
```
tonynajjar, line 109: not sure what's the point in ticking in a loop, it should succeed from the first time

nav2_behavior_tree/test/plugins/decorator/test_goal_updater_node.cpp
```diff
@@ -95,42 +95,101 @@ TEST_F(GoalUpdaterTestFixture, test_tick)
...
+  tree_->rootNode()->executeTick();
+  geometry_msgs::msg::PoseStamped updated_goal;
   config_->blackboard->get("updated_goal", updated_goal);
 
-  EXPECT_NE(updated_goal, goal);
-  EXPECT_EQ(updated_goal, goal_to_update);
+  // expect to succeed
+  EXPECT_EQ(tree_->rootNode()->status(), BT::NodeStatus::SUCCESS);
+  // expect to update goal with latest goal update
+  EXPECT_EQ(updated_goal, goal_to_update_2);
```
tonynajjar, line 192: With the old QoS (depth 10), this would fail -> updated_goal would be equal to goal_to_update_1 since we tick once and therefore `spin_some()` once

tonynajjar: I'm assuming here that the desired behavior is to always get the latest goal published on a topic of course. Is this a safe assumption?

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -46,7 +46,7 @@ GoalUpdater::GoalUpdater(
   sub_option.callback_group = callback_group_;
   goal_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
     goal_updater_topic,
-    10,
+    rclcpp::SystemDefaultsQoS(),
```
tonynajjar, line 49: Tested with cyclonedds. SystemDefaultsQoS has KeepLast policy by default. Maybe we should make that explicit in case some other vendor doesn't have that as default?

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -59,8 +59,13 @@ inline BT::NodeStatus GoalUpdater::tick()
...
   callback_group_executor_.spin_some();
 
-  if (rclcpp::Time(last_goal_received_.header.stamp) > rclcpp::Time(goal.header.stamp)) {
-    goal = last_goal_received_;
+  if (last_goal_received_.header.stamp != rclcpp::Time(0)) {
+    if (rclcpp::Time(last_goal_received_.header.stamp) > rclcpp::Time(goal.header.stamp)) {
+      goal = last_goal_received_;
+    }
+    else {
+      RCLCPP_WARN(node_->get_logger(), "Goal is older than the current one. Not updating");
```
SteveMacenski, line 67: Add the timestamps here. I think this could be useful to ID problems with ROS clocks which would be immediately clear if we had them printed

nav2_behavior_tree/test/plugins/decorator/test_goal_updater_node.cpp
SteveMacenski: I think this is just a general pattern in a bunch of the tests whereas the first tick might not be the end when testing different control flow nodes.

nav2_behavior_tree/test/plugins/decorator/test_goal_updater_node.cpp
SteveMacenski: Yeah that's a good assumption if we have multiple goals in the queue at a given time, use the last one

SteveMacenski:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/9769/workflows/a2a94cc4-4193-4900-ac8c-a64cf5d04c44/jobs/31236/tests Linting issues

SteveMacenski:
A couple small updates and we can merge this :-)

SteveMacenski:
The WPF test is flaky and I haven't had time to look into why, retriggering

SteveMacenski (approved):

PR #3733, opened 2023-08-02
Completing Hybrid-A* visualization of expansion footprints PR

SteveMacenski:


BriceRenaudeau:
There might be a typo here: "activate" instead of "deactivate".

It, therefore, crashes when the planner is clean_up.

https://github.com/ros-planning/navigation2/blob/f1f968b221ceef7019358495ad96347c1abdddf7/nav2_smac_planner/src/smac_planner_hybrid.cpp#L264-L267

SteveMacenski:
Done, thanks!

PR #3751, opened 2023-08-09
Same orientation of coordinate frames in rviz ang gazebo

ct2034:
---

## Basic Info

| ------ | ----------- |

---

* The coordinate frames in rviz and gazebo are oriented such that the positive x-axis points right
* This could simplify things for beginner users.
rviz
gazebo

Is there maybe some tutorial that references this?

---

None

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/rviz/nav2_default_view.rviz
+++ b/nav2_bringup/rviz/nav2_default_view.rviz
@@ -560,7 +560,7 @@ Visualization Manager:
   Value: true
   Views:
     Current:
-      Angle: -1.6150002479553223
+      Angle: 0
       Class: rviz_default_plugins/TopDownOrtho
       Enable Stereo Rendering:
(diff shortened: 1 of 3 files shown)


SteveMacenski:
Both the gazebo and rviz bringup by default are bottom-down orientations. This would rotate both to be on its side, which is no different in being in the same relative heading on launch. If you wanted to adjust the gazebo view to match the rviz view exactly, I would have no issue with that, but I would reject this PR as currently written. I think the current default view of Rviz is quite god

ct2034:
I understand. Now it looks like this:
Screenshot from 2023-08-10 17-54-18
Screenshot from 2023-08-10 17-54-11

SteveMacenski (approved):

SteveMacenski:
Great!

Issue #3172, opened 2022-09-05
Zig-zag due to unfavorable angle increments and starting angle


mangoschorle:
I am experiencing some interesting behavior when using SmacPlanner Hybrid. I am not quite sure whether this is a common mistake/behavior maybe caused by incorrect usage, or whether it is a known limitation. I currently have no idea how to approach this / solve this. If ideas pop up along this thread, I am more than willing to contribute.

So the issue is that depending on which angle-bin the start node falls into, I am getting either a straight line or a very messy zig-zag.

We have **64** angle bins. Here **the start is in bin 32** (pi)
![Screenshot from 2022-09-05 21-53-44](https://user-images.githubusercontent.com/94619337/188508744-3fe1c1ac-1165-4c6d-b902-8b0c2190291f.png)

In this image **the starting angle is in bin 31**
![Screenshot from 2022-09-05 21-54-39](https://user-images.githubusercontent.com/94619337/188508760-19e5a036-c859-4d32-b797-30f1fbfe57ab.png)

I also found that due to my choice of resolution (10cm) and min_turning_radius (50cm), the turning motions of the minimal primitive set (Dubins) cover **four** angle bins at a time, i.e., increments=4 
https://github.com/ros-planning/navigation2/blob/634d2e3d9b6bde5558c90fe6583d52b3ed66cf55/nav2_smac_planner/src/node_hybrid.cpp#L104)

This is actually quite bad since you one will therefore never end up in odd angle bins. But only in every fourth bin that are multiples of four. This also causes the zig-zagging because just can't reach an angle bin that would directly go in the right direction.
Also as can be seen in the image, the planner has to expand WAY more nodes to make progress.

This is a distribution of the angle bins. As you can see, we only reach every fourth angle bin.
![Screenshot from 2022-08-04 16-42-16](https://user-images.githubusercontent.com/94619337/188509434-07c6adb0-fc10-45d3-9f2c-ef25de2f33bf.png)

I wonder / am collecting ideas, on how we could find remedy for this zig-zagging issue. I know, we could double the resolution or use a larger turning radius (>2 meters if we want to cover only a single bin) but we are running into other issues if we do so. Maybe there are other ideas. I have tried adding **additional** turning primitives which turn much less than the min_turning_radius but didn't get good results. Again, if I understand the idea / solution, I'd be more than happy to implement, test, and submit a PR.

SteveMacenski:
Interesting 

[...]

PR #3752, opened 2023-08-10
Fix zig-zaggin in smac planner due to discrete angular bin quantizations

SteveMacenski:
Per https://github.com/ros-planning/navigation2/issues/3172, we sometimes see zig-zagging behavior due to the discrete increments of angular quanitizations not fully utilizing the requested angular resolution. This issue is written out in detail in this comment: https://github.com/ros-planning/navigation2/issues/3172#issuecomment-1672249118

This solves that by creating additional motion primitives to fill in the angular space for each expansion to hit every angular bin possible. As a result, keeping angular bin densities of 64/72 as previously utilized will create **many** more new primitives for certain configurations than previously. It may be wise to reduce your resolution to not create additional primitives that your application does not require.

TODO
- [x]  performance testing across reeds-shepp, dubin, various turning rads (0.05, 0.1, 0.4, 0.5, 1.0, 2.0, 10.0), and non-circular robots for path quality and runtime. 
- [x] Unit testing of new methods like populateExpansionsLog
- [x] Traversal cost for primitives of turns
- [x] Dynamic parameter support
- [x] Document change: reduce resolution (72 no longer needed), new param in guide + migration guide. Post on Discourse about this major change for folks to update their config files https://github.com/ros-planning/navigation.ros.org/pull/459
- [x] Find new best practice resolution to use (e.g. 16, 32, 64, 72, ...) with rationale. More bins = more prims; lower $R_t$ = more prims
- [ ] test on realistic maps, not just benchmarking scripts -- why don't I see more change between the paths with various turning radii? Shouldn't the curves be very different? Need to look into that manually, something doesn't seem quite right (as if the static members are being shared between them?) or does the cost-awareness generally just guide it into the center so they have similar influences? I need to plot some turns. Or has to do with the number of smaller primitives available?

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
@@ -110,7 +110,7 @@ public:
    */
   bool createPath(
     CoordinateVector & path, int & num_iterations, const float & tolerance,
-    std::vector<std::tuple<float, float>> * expansions_log = nullptr);
+    std::vector<std::tuple<float, float, float>> * expansions_log = nullptr);
 
   /**
    * @brief Sets the collision checker to use
@@ -239,6 +239,14 @@ protected:
    * @brief Clear graph of nodes searched
    */
   inline void clearGraph();
+
+  /**
+   * @brief Populate a debug log of expansions for Hybrid-A* for visualization
+   * @param node Node expanded
+   * @param expansions_log Log to add not expanded to
+   */
+  inline void populateExpansionsLog(const NodePtr & node,
+  std::vector<std::tuple<float, float, float>> * expansions_log);
 
   int _timing_interval = 5000;
 
--- a/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
@@ -245,8 +245,8 @@ protected:
    * @param node Node expanded
    * @param expansions_log Log to add not expanded to
    */
-  inline void populateExpansionsLog(const NodePtr & node,
-  std::vector<std::tuple<float, float, float>> * expansions_log);
+  inline void populateExpansionsLog(
+    const NodePtr & node, std::vector<std::tuple<float, float, float>> * expansions_log);
 
   int _timing_interval = 5000;
 
--- a/nav2_smac_planner/include/nav2_smac_planner/node_basic.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/node_basic.hpp
@@ -76,6 +76,7 @@ public:
   MotionPrimitive * prim_ptr;  // Used by NodeLattice
   unsigned int index, motion_index;
   bool backward;
+  TurnDirection turn_dir;
 };
 
 }  // namespace nav2_smac_planner
--- a/nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
@@ -122,10 +122,13 @@ struct HybridMotionTable
   float cost_penalty;
   float reverse_penalty;
   float travel_distance_reward;
+  bool downsample_obstacle_heuristic;
+  bool use_quadratic_cost_penalty;
   ompl::base::StateSpacePtr state_space;
   std::vector<std::vector<double>> delta_xs;
   std::vector<std::vector<double>> delta_ys;
   std::vector<TrigValues> trig_values;
+  std::vector<float> travel_costs;
 };
 
 /**
--- a/nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/node_hybrid.hpp
@@ -232,9 +232,10 @@ public:
    * @brief Sets the motion primitive index used to achieve node in search
    * @param reference to motion primitive idx
    */
-  inline void setMotionPrimitiveIndex(const unsigned int & idx)
+  inline void setMotionPrimitiveIndex(const unsigned int & idx, const TurnDirection & turn_dir)
   {
     _motion_primitive_index = idx;
+    _turn_dir = turn_dir;
   }
 
   /**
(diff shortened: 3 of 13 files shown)


SteveMacenski:
Benchmarking:

```
-----------------  -----------------------  --------------------  ------------------  --------
Planner            Average path length (m)  Average Time (s)      Average cost        Max cost
NORMAL_DUBIN       49.67057981917864        0.056411659300000014  26.42893242594106   98.44
NORMAL_REEDSSHEPP  49.66429946425317        0.056731877410000006  26.44040309517375   98.47
ON_DUBIN           49.670592919255476       0.05622163887999999   26.428906231633896  98.44
ON_REEDSSHEPP      49.66603145368965        0.05642396648999999   26.437818514359627  98.47
-----------------  -----------------------  --------------------  ------------------  --------

```

For previous changes to current changes in both the dubins and reeds-shepp models on the random benchmark map kindly provided by @jwallace42 

Turning radius tests next to see how it scales with primitive numbers

SteveMacenski:
```
------------  -----------------------  -------------------  ------------------  --------
Planner       Average path length (m)  Average Time (s)     Average cost        Max cost
DUBIN_005     48.0687865862458         0.34696313113000005  33.63145014373353   113.14
DUBIN_01      48.05450925271065        0.33147114150999996  33.680642759452276  113.14
DUBIN_04      48.052733500317366       0.33054843583        33.683575324139866  113.14
DUBIN_10      48.052764603818105       0.32961878808        33.65879248833779   113.03
DUBIN_20      48.15930088892668        0.33270151082999994  33.91807622385019   115.74
DUBIN_005_ON  48.05101334085401        0.33693136072999996  33.69104135579259   113.14
DUBIN_01_ON   48.05975069796279        0.3672095316         33.66681511255692   113.14
DUBIN_04_ON   48.052900739188914       0.33240715681000005  33.68058711888695   113.14
DUBIN_10_ON   48.052764603818105       0.33412878039        33.65879248833779   113.03
DUBIN_20_ON   48.15930088892668        0.33433825997        33.91807622385019   115.74
------------  -----------------------  -------------------  ------------------  --------

```

For different turning radius'. First set are normal, second set are with the addtl primitives set on. More or less all the same or not notably different worth additional optimization

stevedanomodolor:
There seems to be not much of a difference between the activation of the new parameter, I tried using 64, 72 and with the new parameter on, and the behavior continues. 
Figure_1
Figure_2
Figure_3
Figure_4

SteveMacenski:
Thanks for the quick analysis! I think its due to me not updating this function yet (https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/node_hybrid.cpp#L303) that was brought up in the ticket. I think because its not 0,1,2, we're penalizing it the most and also giving it the reversing penalty. I think that's why you see it happening sometimes (e.g. the graph shows occasional uses of the other bins) but not as regularly as it should be. 

I'll work on that tomorrow, hopefully that will resolve the issue! I'll need to re-run the testing metrics above, but that's relatively straight forward. I appreciate your experimentation with the charts, it was illuminating!

SteveMacenski:
Traversal cost function updates made, but it involved some surgical changes so I'm going to be extremely careful in testing to make sure this still works as expected since it involved modifications that could be subtly error prone. I looked carefully over them and added a couple more unit tests to sanity check the actual traversal cost function, but since there were updates to the projections computation and how the data is propagated during search, its worth careful consideration.  

@stevedanomodolor please test again :-)

SteveMacenski:
With updates:

```
-----------------  -----------------------  -------------------  ------------------  --------
Planner            Average path length (m)  Average Time (s)     Average cost        Max cost
NORMAL_DUBIN       48.62226744406985        0.08540430985999999  2.315272051418369   91.66
NORMAL_REEDSSHEPP  48.620874918802116       0.08607285177000001  2.327718953474646   91.66
ON_DUBIN           48.62135754274054        0.08432577783999999  2.3343595784746456  91.8
ON_REEDSSHEPP      48.619247989185006       0.08586728219999998  2.3044361077010693  91.78
-----------------  -----------------------  -------------------  ------------------  --------

```
I'm messing with costmap settings, so the actual values aren't really important. The comparison with the original method is the important part. 

```
-------------  -----------------------  -------------------  ------------------  --------
Planner        Average path length (m)  Average Time (s)     Average cost        Max cost
NORMAL_DUBIN   48.46639022624647        0.38991617460999994  33.311832403525656  111.88
NORMAL_DUBIN2  48.46688734930349        0.37671986905        33.31204292528036   111.88
NORMAL_DUBIN3  48.5784431139699         0.37746076980999993  33.517786595032454  114.44
ON_DUBIN       48.46639022624647        0.37777031248999987  33.311832403525656  111.88
ON_DUBIN2      48.46010691573797        0.3620015351         33.32492302248469   111.88
ON_DUBIN3      48.57142159711799        0.36241974367000007  33.53521806264665   114.44
-------------  -----------------------  -------------------  ------------------  --------

```

(for 0.05, 0.4, 2.0 turning radii). 

Something seems odd at the moment, we seem to be using alot more search iterations than before. Also, the tests had to be updated for a larger number of iterations as well which shouldn't be happening when the feature is off (which is default and the best I can tell is the context for the test). So, something seems a little off to me but I can't quite place it

stevedanomodolor:
I performed some more tests based on the new changes, the behavior does not appear and the planner is able to use more of the bin space, although one thing to watch out for is the increment of planning time which make sense because we have increased the number of primitives used during the planning. With respect to planning time, when using the default minimum turning radius, 72 seems to be the optimal number of quantization bins. The results of the graphs are for 64, 72, 124, and 256 while the table includes more tests in between. 
It would be nice to try this on a different map to see how it behaves, if you have any more map that i could try this on too. 
## Comparision 
image


## Figures
### 64 bins no interpolation
64_bins_no_interpolation_figure
64_bins_no_interpolation

### 64 bins interpolation

64_bins_interpolation
64_bins_interpolation_figure

### 72 bins no interpolation
72_bins_no_interpolation
72_bins_no_interpolation_figure

### 72 bins interpolation
72_bins_interpolation
72_bins_interpolation_figure

### 124 bins no interpolation 
124_bins_no_interpolation
124_bins_no_interpolation_figure

### 124 bins interpolation 

124_bins_interpolation
124_bins_interpolation_figure

### 256 bins no interpolation 

256_bins_no_interpolation
256_bins_no_interpolation_figure



### 256 bins  interpolation 

256_bins_interpolation
256_bins_interpolation_figure

SteveMacenski:
Great thanks for the detail!

SteveMacenski:
@stevedanomodolor see new changes adding the additional features which can be enabled using the parameters 

```
      downsample_obstacle_heuristic: false
      use_quadratic_cost_penalty: true
```

PR #3764, opened 2023-08-17
nav2_smac_planner: log start and goal position when creating plan

DylanDeCoeyer-Quimesis:
This information is useful for debug purposes.

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


DylanDeCoeyer-Quimesis:
In fact, there is already a debug log available in the planner_server: https://github.com/ros-planning/navigation2/blob/b6f3b028f67001be38b7ab3cd7dc6d2dd5b0c58e/nav2_planner/src/planner_server.cpp#L578

PR #3778, opened 2023-08-28
Declare param use_maximum on static _layer

Sunart24:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (N/A |
| Robotic platform tested on | Currently tested on simulated robot|

---

* I recover the "use_maximum" param declaration to use it in a similar way the "combination_method" is used in every layer as obstacle_layer. It is useful to use the enabled param dynamically and keep the maximum value from the data of the layers.  All the code was done, but the declaration was missing or broken. 

* Add the parameter to the official navigation web. 

#### For Maintainers: 


SteveMacenski:
I don't believe this is correct since the API guide states that this will throw if its not declared (https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Node.html#a0bf1077fba623d72fe1b49805f6c0a5a) so it must already be declared. Declaring again will throw an exception from being declared twice.

Sunart24:
You are right,  is declared as a parameter of Costmap2DROS. Thanks! I'll close this PR.

Issue #3765, opened 2023-08-18
rclcpp_action goal handle timeout logic clearing long-running goals


AlexeyMerzlyakov:
This issue describes the flaky issues appeared with `test_waypoint_follower` system test

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04.3
- ROS2 Version:
  - ROS2 Rolling (built from sources)
- Version or commit hash:
  - b6f3b028 (and older 822560ce)
- DDS implementation:
  - Cyclone DDS

#### Steps to reproduce issue
1. Enable visualization for Waypoint Follower system test by following patch:
```
diff --git a/nav2_system_tests/src/waypoint_follower/test_case_py.launch b/nav2_system_tests/src/waypoint_follower/test_case_py.launch
index adeb839c..c7194fe9 100755
--- a/nav2_system_tests/src/waypoint_follower/test_case_py.launch
+++ b/nav2_system_tests/src/waypoint_follower/test_case_py.launch
@@ -40,6 +40,8 @@ def generate_launch_description():
     bringup_dir = get_package_share_directory('nav2_bringup')
     params_file = os.path.join(bringup_dir, 'params/nav2_params.yaml')
 
+    rviz_config_file = os.path.join(bringup_dir, 'rviz', 'nav2_default_view.rviz')
+
     # Replace the `use_astar` setting on the params file
     configured_params = RewrittenYaml(
         source_file=params_file,
@@ -59,6 +61,13 @@ def generate_launch_description():
                  '--minimal_comms', world],
             output='screen'),
 
+        # Launch rviz
+        Node(
+            package='rviz2',
+            executable='rviz2',
+            arguments=['-d', rviz_config_file],
+            output='screen'),
+
         # TODO(orduno) Launch the robot state publisher instead
         #              using a local copy of TB3 urdf file
         Node(
```
2. Run `test_waypoint_follower` manually, or using the attached below stress-test script:
[run_stress.sh.txt](https://github.com/ros-planning/navigation2/files/12380528/run_stress.sh.txt)

#### Expected behavior
Test passes 30 times in a row.

#### Actual behavior
The following problems were found during the test:

##### Problem1: No action
When the robot is moving near the obstacle, it could stuck. Sometimes, this causes recovery behaviors that being triggered.
[...]

PR #3785, opened 2023-08-30
Fixes for flaky WPF test

AlexeyMerzlyakov:
* New RewrittenYaml ability to add non-existing parameters
* Prune distance fix for WPF test
* Treat UNKNOWN status as error in WPF
* Clear error codes after BT run
* Remove unnecessary setInitialPose() from WPF test

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3765 |
| Robotic platform tested on | colcon test |

---

The following problems described in #3765, were fixed:

| Problem | Solution description |
| ------------- | ------------- |
| Problem1: No action | Return back BT action server timeout to 15 minutes: #3787 |
| Problem1: No action  | New RewrittenYaml ability to add non-existing parameters & prune distance fix for lower probability of recovery situations |
| Problem1: No action | Treat UNKNOWN status as error in WPF to avoid hangs when action goal was deleted |
| Problem2: Incorrect out-of-map status | Clear error codes after BT run |
| Problem3: Robot lost | Remove unnecessary setInitialPose() from testing |

## Verification status

Performed `colcon test` for `nav2_waypoint_follower`, `nav2_behavior_tree` and `nav2_system_tests` packages, touched in this PR. No regressions appear during the local testing.

* Describe new ability of `RewrittenYaml` in Iron->Jazzy MG

---

* Remove Costmap Filter parameters YAML-s in favor of using new ability of RewrittenYaml to add new ROS-parameters. This will reduce boilerplate works to synchronize Costmap Filter parameters YAML-s with main one.
* Fix RCLCPP action 10-second timeout in either RCLCPP or make a workaround for Nav2

Both above items are subject of separate PR-s.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -284,6 +284,8 @@ void BtActionServer<ActionT>::executeCallback()
       action_server_->terminate_all(result);
       break;
   }
+
+  cleanErrorCodes();
 }
 
 template<class ActionT>
@@ -307,6 +309,14 @@ void BtActionServer<ActionT>::populateErrorCode(
 
   if (highest_priority_error_code != std::numeric_limits<int>::max()) {
     result->error_code = highest_priority_error_code;
+  }
+}
+
+template<class ActionT>
+void BtActionServer<ActionT>::cleanErrorCodes()
+{
+  for (const auto & error_code : error_code_names_) {
+    blackboard_->set<unsigned short>(error_code, 0);  //NOLINT
   }
 }
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -230,6 +230,7 @@ void BtActionServer<ActionT>::executeCallback()
 {
   if (!on_goal_received_callback_(action_server_->get_current_goal())) {
     action_server_->terminate_current();
+    cleanErrorCodes();
     return;
   }
 
--- a/nav2_waypoint_follower/src/waypoint_follower.cpp
+++ b/nav2_waypoint_follower/src/waypoint_follower.cpp
@@ -275,9 +275,7 @@ WaypointFollower::followWaypoints()
       }
     }
 
-    if (current_goal_status_.status != ActionStatus::PROCESSING &&
-      current_goal_status_.status != ActionStatus::UNKNOWN)
-    {
+    if (current_goal_status_.status != ActionStatus::PROCESSING) {
       // Update server state
       goal_index++;
       new_goal = true;
--- a/nav2_waypoint_follower/src/waypoint_follower.cpp
+++ b/nav2_waypoint_follower/src/waypoint_follower.cpp
@@ -217,7 +217,10 @@ WaypointFollower::followWaypoints()
     feedback->current_waypoint = goal_index;
     action_server_->publish_feedback(feedback);
 
-    if (current_goal_status_.status == ActionStatus::FAILED) {
+    if (
+      current_goal_status_.status == ActionStatus::FAILED ||
+      current_goal_status_.status == ActionStatus::UNKNOWN)
+    {
       nav2_msgs::msg::MissedWaypoint missedWaypoint;
       missedWaypoint.index = goal_index;
       missedWaypoint.goal = goal->poses[goal_index];
@@ -324,6 +327,7 @@ WaypointFollower::resultCallback(
       current_goal_status_.status = ActionStatus::FAILED;
       return;
     default:
+      RCLCPP_ERROR(get_logger(), "Got unknown result from BT");
       current_goal_status_.status = ActionStatus::UNKNOWN;
       return;
   }
--- a/nav2_waypoint_follower/src/waypoint_follower.cpp
+++ b/nav2_waypoint_follower/src/waypoint_follower.cpp
@@ -327,7 +327,7 @@ WaypointFollower::resultCallback(
       current_goal_status_.status = ActionStatus::FAILED;
       return;
     default:
-      RCLCPP_ERROR(get_logger(), "Got unknown result from BT");
+      RCLCPP_ERROR(get_logger(), "Received an UNKNOWN result code from navigation action!");
       current_goal_status_.status = ActionStatus::UNKNOWN;
       return;
   }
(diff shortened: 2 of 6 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -284,6 +284,8 @@ void BtActionServer<ActionT>::executeCallback()
       action_server_->terminate_all(result);
       break;
   }
+
+  cleanErrorCodes();
```
SteveMacenski, line 288: What about the other return above if not `on_goal_received_callback_`

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -217,7 +217,10 @@ WaypointFollower::followWaypoints()
     feedback->current_waypoint = goal_index;
     action_server_->publish_feedback(feedback);
 
-    if (current_goal_status_.status == ActionStatus::FAILED) {
+    if (
+      current_goal_status_.status == ActionStatus::FAILED ||
+      current_goal_status_.status == ActionStatus::UNKNOWN)
```
SteveMacenski, line 222: Is this really a failed case if unknown from this timeout? Seems like we should instead temporarily increase the timeout back to 15 minutes (@pepisg ) while working through a potential patch to rclcpp to better keep actively processing goals around.

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -324,6 +327,7 @@ WaypointFollower::resultCallback(
       current_goal_status_.status = ActionStatus::FAILED;
       return;
     default:
+      RCLCPP_ERROR(get_logger(), "Got unknown result from BT");
```
SteveMacenski, line 330:
```suggestion
      RCLCPP_ERROR(get_logger(), "Received an UNKNOWN result code from navigation action!");
```

nav2_waypoint_follower/src/waypoint_follower.cpp
pepisg: Exposed that as a parameter in #3787

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
AlexeyMerzlyakov: In current implementations (`NavigateToPose` and `NavigateThroughPoses`) `on_goal_received_callback_` will fail only if BT action server will fail to load behavior tree (`loadBehaviorTree(bt_xml_filename)`). If I understand correctly, `error_code` in the blackboard is being filled during BT run, rather than during XML-file load.
In other words, status clean mechanism is required to keep clean/correct errors between two sequential BT runs. And if some problem will appear on `on_goal_received_callback_` stage, we will not go through BT run in this case, but on next `BtActionServer::executeCallback()` call again will try to execute `on_goal_received_callback_` which will load/initalize BT tree from XML again.
So, this should be unnecessary in this case.

However, here is another side of view: if we won't rely on current implementations and new navigator might appear in the future (for example `NavigateAroundPoses`), and this new navigator will do something with blacklist's error codes at `goalReceived()` stage, we need to make it safe-and-sane, and clean all error status whenever it possible.

I'd rather to rely on current implementations, but if you think we need to do this anyway, please let me know, I'll give an update.

AlexeyMerzlyakov: Thinking about it twice made me decide that we'd better to have resilience to further code changes, and thus will avoid unnecessary failures. So added additional error clean in https://github.com/ros-planning/navigation2/pull/3785/commits/2912c92364a012e2da055e9878bd1053da1af4ec

SteveMacenski:
Beyond that last question, I think this is good to go!

nav2_waypoint_follower/src/waypoint_follower.cpp
AlexeyMerzlyakov: This PR is complementary with PR3787, that improves the situation and fixes some bugs, that 3787 does not. So both PR-s ought to be be merged (moreover, new parameters in RewrittenYaml I plan to use to comb-up Costmap Filters system tests).

SteveMacenski: @AlexeyMerzlyakov I meant only the addition of `|| current_goal_status_.status == ActionStatus::UNKNOWN`. I thought this line was to mask the unknown return type problem? If we increase the timeout, wouldn't that make it so we didn't see `UNKNOWN`? This logic treats `UNKNOWN` as a failure.

If we keep this here should https://github.com/ros-planning/navigation2/pull/3785/files#diff-872945d6a63001626ab7379639b5b3d5f11982ea3bdf08da01e7f111a9c2dfecR279 be changed for incrementing new goals? You're treating UNKNOWN as a terminal condition in your change, so this doesn't make as much sense anymore, correct?

AlexeyMerzlyakov: This status is being treated as terminal one for WPF server in order to avoid infinite cycles whether for some reasons the goal got into "UNKNOWN" state. This is intended to fix/mask not only the case when the goal is getting after it was destroyed, but rather for any other potential problems. Briefly looking to RCLCPP and RCL code shows that UNKNOWN state might appear in case if goal was invalidated, destroyed, does not exists at the moment of if the transition was invalid or any other error occurred. So, receiving UNKNOWN goal for action server seems to be treated as problem state, if I got the RCLCPP / RCL intention correctly.

When I've wrote this change, I've also focused on `BtActionNode` server, receiving the result from downstream nodes. For this case, "UNKNOWN" status is [being treated](https://github.com/ros-planning/navigation2/blob/be4d40711bbaf811aa3a0d5de346216ce67d0083/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L289) as error situation, and error is being thrown, interrupting normal operation.
From this point of view, looking on the UNKNOWN goal status as on error seems to be logical.

> If we keep this here should https://github.com/ros-planning/navigation2/pull/3785/files#diff-872945d6a63001626ab7379639b5b3d5f11982ea3bdf08da01e7f111a9c2dfecR279 be changed for incrementing new goals?

Yes, this should be removed for UNKNOWN stage, if we will treat UNKNOWNS as errors. Thanks for the notice!
If we will agree on main idea, I will remove it in next update.

SteveMacenski: Yup that works for me!

AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3785/commits/85b6b71f6793f6740db7cd0d4d930bd19de3ca38

SteveMacenski (approved):

SteveMacenski:
@AlexeyMerzlyakov thanks so much for your time on this!

PR #3787, opened 2023-08-31
Expose action server result timeout as a parameter in bt navigator servers

pepisg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3765  |
| Primary OS tested on | Ubuntu 22.04 (docker) |
| Robotic platform tested on | gazebo turtlebot |

---

* As described in #3765, the waypoint follower is getting `UNKNOWN` results if the call to `/navigate_to_pose` takes more than 10 seconds to complete. @AlexeyMerzlyakov traced this back the problem to https://github.com/ros2/rcl/pull/1012. I'm exposing the result timeout of the action server as a parameter in the `bt_navigator` node following @AlexeyMerzlyakov suggestion in #3765

* Added the new parameter `action_server_result_timeout`, so need to add that to default configs and documentation page

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -91,6 +91,16 @@ BtActionServer<ActionT>::BtActionServer(
       RCLCPP_INFO_STREAM(logger_, "Error_code parameters were set to:" << error_codes_str);
     }
   }
+
+  // Result timeout for the action server.
+  // In https://github.com/ros2/rcl/pull/1012 a change was introduced
+  // which makes action servers discard a goal handle if the result is not produced
+  // within 10 seconds. Since this may not be the case for all actions in
+  // Nav2, this timeout is exposed as a parameter and default to the previous
+  // expiration value of 15 minutes.
+  if (!node->has_parameter("action_server_result_timeout")) {
+    node->declare_parameter("action_server_result_timeout", 15 * 60);
+  }
 }
 
 template<class ActionT>
@@ -128,15 +138,21 @@ bool BtActionServer<ActionT>::on_configure()
   client_node_->declare_parameter(
     "global_frame", node->get_parameter("global_frame").as_string());
 
+  // set the timeout in seconds for the action server to discard goal handles
+  int timeout;
+  node->get_parameter("action_server_result_timeout", timeout);
+  rcl_action_server_options_t server_options = rcl_action_server_get_default_options();
+  server_options.result_timeout.nanoseconds = RCL_S_TO_NS(timeout);
+
   action_server_ = std::make_shared<ActionServer>(
     node->get_node_base_interface(),
     node->get_node_clock_interface(),
     node->get_node_logging_interface(),
     node->get_node_waitables_interface(),
-    action_name_, std::bind(&BtActionServer<ActionT>::executeCallback, this));
+    action_name_, std::bind(&BtActionServer<ActionT>::executeCallback, this),
+    nullptr, std::chrono::milliseconds(500), false, server_options);
 
   // Get parameters for BT timeouts
-  int timeout;
   node->get_parameter("bt_loop_duration", timeout);
   bt_loop_duration_ = std::chrono::milliseconds(timeout);
   node->get_parameter("default_server_timeout", timeout);
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -61,6 +61,9 @@ BtActionServer<ActionT>::BtActionServer(
   if (!node->has_parameter("default_server_timeout")) {
     node->declare_parameter("default_server_timeout", 20);
   }
+  if (!node->has_parameter("action_server_result_timeout")) {
+    node->declare_parameter("action_server_result_timeout", 900.0);
+  }
 
   std::vector<std::string> error_code_names = {
     "follow_path_error_code",
@@ -91,16 +94,6 @@ BtActionServer<ActionT>::BtActionServer(
       }
       RCLCPP_INFO_STREAM(logger_, "Error_code parameters were set to:" << error_codes_str);
     }
-  }
-
-  // Result timeout for the action server.
-  // In https://github.com/ros2/rcl/pull/1012 a change was introduced
-  // which makes action servers discard a goal handle if the result is not produced
-  // within 10 seconds. Since this may not be the case for all actions in
-  // Nav2, this timeout is exposed as a parameter and defaults to the previous
-  // expiration value of 15 minutes.
-  if (!node->has_parameter("action_server_result_timeout")) {
-    node->declare_parameter("action_server_result_timeout", 900.0);
   }
 }
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
@@ -161,7 +161,7 @@ bool BtActionServer<ActionT>::on_configure()
   // Get parameters for BT timeouts
   int bt_loop_duration;
   node->get_parameter("bt_loop_duration", bt_loop_duration);
-  bt_loop_duration_ = std::chrono::milliseconds(timeout);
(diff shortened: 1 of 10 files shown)


pepisg:
@SteveMacenski should this parameter be exposed in all nodes that have action servers?

AlexeyMerzlyakov (approved):
Everything seems to be set well: this should be applied to `BehaviorTreeNavigator` having BT Action Server inside, and be utilized by both `NavigateToPose` and `NavigateThroughPoses` navigators, that both required long periods of running time. So, from my side everything is LGTM

SteveMacenski:
> should this parameter be exposed in all nodes that have action servers?

As much as I don't like it, but yes, I think we should expose it to all the servers. If there's no dynamic replanning, the controller server could run for > 10s. If there's a really, really slow path planner it could as well technically. You could have these default to the same default in rclcpp. This will require documentation updates to the Configuration Guide accordingly to show users this param exists (but not necessarily in the yaml file, the yaml file isn't a complete set of params).

The only 2 that need non-default values are the high level navigation action and the WPF action. 

Wait - you didn't actually add this to the WPF server, the one we see the issue with :laughing: That one is **definitely** necessary

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -91,6 +91,16 @@ BtActionServer<ActionT>::BtActionServer(
...
   }
+
+  // Result timeout for the action server.
+  // In https://github.com/ros2/rcl/pull/1012 a change was introduced
+  // which makes action servers discard a goal handle if the result is not produced
+  // within 10 seconds. Since this may not be the case for all actions in
+  // Nav2, this timeout is exposed as a parameter and defaults to the previous
+  // expiration value of 15 minutes.
+  if (!node->has_parameter("action_server_result_timeout")) {
+    node->declare_parameter("action_server_result_timeout", 15 * 60);
```
SteveMacenski, line 102:
```suggestion
    node->declare_parameter("action_server_result_timeout", 900.0);
```

Make it clearly a double since declaring the parameter also declares its type

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -128,15 +138,21 @@ bool BtActionServer<ActionT>::on_configure()
   client_node_->declare_parameter(
     "global_frame", node->get_parameter("global_frame").as_string());
 
+  // set the timeout in seconds for the action server to discard goal handles if not finished
+  int timeout;
```
SteveMacenski, line 142: make a double

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -128,15 +138,21 @@ bool BtActionServer<ActionT>::on_configure()
...
     node->get_node_base_interface(),
     node->get_node_clock_interface(),
     node->get_node_logging_interface(),
     node->get_node_waitables_interface(),
-    action_name_, std::bind(&BtActionServer<ActionT>::executeCallback, this));
+    action_name_, std::bind(&BtActionServer<ActionT>::executeCallback, this),
+    nullptr, std::chrono::milliseconds(500), false, server_options);
 
   // Get parameters for BT timeouts
-  int timeout;
```
SteveMacenski, line 139: DONT reuse variables

SteveMacenski (changes requested):

pepisg:
> As much as I don't like it, but yes, I think we should expose it to all the servers. If there's no dynamic replanning, the controller server could run for > 10s. If there's a really, really slow path planner it could as well technically. You could have these default to the same default in rclcpp. This will require documentation updates to the Configuration Guide accordingly to show users this param exists (but not necessarily in the yaml file, the yaml file isn't a complete set of params).

Should I do that on this PR? Should I remove that from the yaml files?

> Wait - you didn't actually add this to the WPF server, the one we see the issue with 😆 That one is **definitely** necessary

So, to fix the problem with the WPF this had to be added in the bt_navigator servers because the WPF has a client to the `bt_navigator`. This client is the one that is getting unknown results when the BT navigator server discards the goal handle as described in the bug, thus timeout needs to be increased for the server, not for the client

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
AlexeyMerzlyakov: As I understand, the `timeout` is being already re-used later in the code for storing "bt_loop_duration" and "default_server_timeout", so one more parameter re-usage wouldn't be an issue, I believe. How about to leave the comment before each (re)usage to get it clear?

SteveMacenski: Yikes. That's not good either. Lets make them all their own variables.

pepisg:
I think the parameter is now exposed in all action servers, however I have a couple of questions:
1. Should all node's have the parameter default to 15 min? or should 15 min be the default only for navigator and wpf and 10 seconds for the other nodes? (what about the timed behavior in nav2_behaviors?)
2. Should I leave the comment on why this parameter was exposed everywhere? or should I remove that and put in the configuration guides instead?

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -92,6 +92,16 @@ BtActionServer<ActionT>::BtActionServer(
...
     }
   }
+
+  // Result timeout for the action server.
+  // In https://github.com/ros2/rcl/pull/1012 a change was introduced
+  // which makes action servers discard a goal handle if the result is not produced
+  // within 10 seconds. Since this may not be the case for all actions in
+  // Nav2, this timeout is exposed as a parameter and defaults to the previous
+  // expiration value of 15 minutes.
+  if (!node->has_parameter("action_server_result_timeout")) {
```
SteveMacenski, line 102: do this logic in the constructor where the other instances of the `!has -> declare` are for timeouts

SteveMacenski (changes requested):

nav2_smoother/src/nav2_smoother.cpp
```diff
@@ -53,6 +53,14 @@ SmootherServer::SmootherServer(const rclcpp::NodeOptions & options)
     rclcpp::ParameterValue(std::string("base_link")));
   declare_parameter("transform_tolerance", rclcpp::ParameterValue(0.1));
   declare_parameter("smoother_plugins", default_ids_);
+
+  // Result timeout for the action server.
```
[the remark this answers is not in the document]
SteveMacenski: Also, configuration guide entries for the new parameters + a migration guide entry for the change in software (and that's where we should definitely link back to that rclcpp ticket about this issue)

nav2_smoother/src/nav2_smoother.cpp
```diff
@@ -53,6 +53,14 @@ SmootherServer::SmootherServer(const rclcpp::NodeOptions & options)
...
+  // In https://github.com/ros2/rcl/pull/1012 a change was introduced
+  // which makes action servers discard a goal handle if the result is not produced
+  // within 10 seconds. Since this may not be the case for all actions in
+  // Nav2, this timeout is exposed as a parameter and defaults to the previous
+  // expiration value of 15 minutes.
+  declare_parameter("action_server_result_timeout", 900.0);
```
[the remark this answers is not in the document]
pepisg: What about the timed behavior in nav2_behaviors?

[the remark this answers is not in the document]
SteveMacenski: I think a default of 10s also makes sense. Spin/wait/backup are very unlikely to exceed that in any practical sense. Teleop is based on the length of the user command, but if we have to generalize, keeping default seems to make sense

pepisg:
docs & migration guide for this PR: https://github.com/ros-planning/navigation.ros.org/pull/461

SteveMacenski (approved):
Approved, just need the small change to the documentation and I'll merge the set!  Thanks for this work @pepisg !

SteveMacenski:
Oh whoops, you have a linting issue:

```
Diff with 5 lines
--- src/controller_server.cpp
+++ src/controller_server.cpp.uncrustify
@@ -52 +52 @@
-  
+
```

PR #3788, opened 2023-08-31
[RPP] Project carrot beyond end of path if interpolating

### LINKED ISSUE #3784: [RPP] Carrot at end of path does not respect look ahead distance
#### Feature description
At the moment the `getLookAheadPoint()` function takes the last point on the path if the look ahead distance would put it past the last point. This has the effect of increasing the gain of the steering inputs (ie the look ahead distance is reduced).
https://github.com/ros-planning/navigation2/blob/210977790bfb6e809f2c2ef3b7e1660662d630cc/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L323-L325

We are using an Ackermann steered vehicle and having the carrot be so close to the robot makes for very unstable motion. When the robot pose is close to the final point in the path small deviations create large steering efforts, and inevitably full-lock steering just before the path is completed.

Project carrot off the end of the last path segment in RPP controller. There is various collision checking, etc that is based on the carrot position, so this "augmented carrot" would have to be implemented such that those functions are unchanged.

If this sounds like the thing to do, I'll get on to writing something.

james-ward:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Closes #3784  |
| Robotic platform tested on | Ackermann steered EV |

---

* Projects the carrot past the end of the path
  * Only when interpolation mode is on (users that want actual path poses will still get them)
  * Prevents the effective gain of the steering system being increased when approaching the goal
  * Uses vector between last two path poses for direction - ignores orientation of last path pose

None

---

None

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -14,30 +14,30 @@
 // limitations under the License.
 
 #include <algorithm>
-#include <string>
 #include <limits>
 #include <memory>
-#include <vector>
+#include <string>
 #include <utility>
+#include <vector>
 
-#include "nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp"
 #include "nav2_core/controller_exceptions.hpp"
-#include "nav2_util/node_utils.hpp"
-#include "nav2_util/geometry_utils.hpp"
 #include "nav2_costmap_2d/costmap_filters/filter_values.hpp"
+#include "nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp"
+#include "nav2_util/geometry_utils.hpp"
+#include "nav2_util/node_utils.hpp"
 
-using std::hypot;
-using std::min;
-using std::max;
 using std::abs;
-using namespace nav2_costmap_2d;  // NOLINT
+using std::hypot;
+using std::max;
+using std::min;
+using namespace nav2_costmap_2d; // NOLINT
 
 namespace nav2_regulated_pure_pursuit_controller
 {
 
 void RegulatedPurePursuitController::configure(
-  const rclcpp_lifecycle::LifecycleNode::WeakPtr & parent,
-  std::string name, std::shared_ptr<tf2_ros::Buffer> tf,
+  const rclcpp_lifecycle::LifecycleNode::WeakPtr & parent, std::string name,
+  std::shared_ptr<tf2_ros::Buffer> tf,
   std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros)
 {
   auto node = parent.lock();
@@ -63,7 +63,8 @@ void RegulatedPurePursuitController::configure(
     tf2::durationFromSec(params_->transform_tolerance), tf_, costmap_ros_);
 
   // Checks for imminent collisions
-  collision_checker_ = std::make_unique<CollisionChecker>(node, costmap_ros_, params_);
+  collision_checker_ =
+    std::make_unique<CollisionChecker>(node, costmap_ros_, params_);
 
   double control_frequency = 20.0;
   goal_dist_tol_ = 0.25;  // reasonable default before first update
@@ -71,8 +72,10 @@ void RegulatedPurePursuitController::configure(
   node->get_parameter("controller_frequency", control_frequency);
   control_duration_ = 1.0 / control_frequency;
 
-  global_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("received_global_plan", 1);
-  carrot_pub_ = node->create_publisher<geometry_msgs::msg::PointStamped>("lookahead_point", 1);
+  global_path_pub_ =
+    node->create_publisher<nav_msgs::msg::Path>("received_global_plan", 1);
+  carrot_pub_ = node->create_publisher<geometry_msgs::msg::PointStamped>(
+    "lookahead_point", 1);
 }
 
 void RegulatedPurePursuitController::cleanup()
@@ -108,7 +111,8 @@ void RegulatedPurePursuitController::deactivate()
   carrot_pub_->on_deactivate();
 }
 
-std::unique_ptr<geometry_msgs::msg::PointStamped> RegulatedPurePursuitController::createCarrotMsg(
+std::unique_ptr<geometry_msgs::msg::PointStamped>
+RegulatedPurePursuitController::createCarrotMsg(
   const geometry_msgs::msg::PoseStamped & carrot_pose)
 {
   auto carrot_msg = std::make_unique<geometry_msgs::msg::PointStamped>();
@@ -128,7 +132,8 @@ double RegulatedPurePursuitController::getLookAheadDistance(
   if (params_->use_velocity_scaled_lookahead_dist) {
     lookahead_dist = fabs(speed.linear.x) * params_->lookahead_time;
     lookahead_dist = std::clamp(
-      lookahead_dist, params_->min_lookahead_dist, params_->max_lookahead_dist);
+      lookahead_dist, params_->min_lookahead_dist,
+      params_->max_lookahead_dist);
   }
 
   return lookahead_dist;
@@ -138,8 +143,7 @@ double calculateCurvature(geometry_msgs::msg::Point lookahead_point)
 {
   // Find distance^2 to look ahead point (carrot) in robot base frame
   // This is the chord length of the circle
-  const double carrot_dist2 =
-    (lookahead_point.x * lookahead_point.x) +
+  const double carrot_dist2 = (lookahead_point.x * lookahead_point.x) +
     (lookahead_point.y * lookahead_point.y);
 
   // Find curvature of circle (k = 1 / R)
@@ -150,7 +154,8 @@ double calculateCurvature(geometry_msgs::msg::Point lookahead_point)
   }
 }
 
-geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocityCommands(
+geometry_msgs::msg::TwistStamped
+RegulatedPurePursuitController::computeVelocityCommands(
   const geometry_msgs::msg::PoseStamped & pose,
   const geometry_msgs::msg::Twist & speed,
   nav2_core::GoalChecker * goal_checker)
@@ -158,7 +163,8 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
   std::lock_guard<std::mutex> lock_reinit(param_handler_->getMutex());
 
   nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
-  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> lock(*(costmap->getMutex()));
+  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> lock(
+    *(costmap->getMutex()));
 
   // Update for the current goal checker's state
   geometry_msgs::msg::Pose pose_tolerance;
@@ -182,7 +188,8 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
     // Cusp check
     const double dist_to_cusp = findVelocitySignChange(transformed_plan);
 
-    // if the lookahead distance is further than the cusp, use the cusp distance instead
+    // if the lookahead distance is further than the cusp, use the cusp distance
+    // instead
     if (dist_to_cusp < lookahead_dist) {
       lookahead_dist = dist_to_cusp;
     }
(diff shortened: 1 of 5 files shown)


nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -322,23 +322,42 @@ geometry_msgs::msg::PoseStamped RegulatedPurePursuitController::getLookAheadPoin
   // If the no pose is not far enough, take the last pose
   if (goal_pose_it == transformed_plan.poses.end()) {
     goal_pose_it = std::prev(transformed_plan.poses.end());
-  } else if (params_->use_interpolation && goal_pose_it != transformed_plan.poses.begin()) {
+  }
+  if (params_->use_interpolation && goal_pose_it != transformed_plan.poses.begin()) {
```
SteveMacenski, line 326: Interpolation is regarding interpolating between path points for sparse paths / large vehicles for selecting the lookahead points at steady state. 

I think this approach to goal stuff is something separate. Mostly because people might want this feature you're adding, but not care about path tracking interpolation

SteveMacenski:
@doisyg please review :-)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
james-ward: My reasoning is that if someone isn't turning on the interpolation then they (presumably) care about only having carrots that are poses from the path. Projecting off the end of the path would violate that as we are creating a new pose to track.

WRT the default behaviour - yes, I believe it should always be enabled. At the moment if you are asking for interpolated carrots, you are in effect asking for the planner to always produce a carrot that is exactly the lookahead distance from the robot. At the moment when approaching the goal this is not the case - the carrot gets closer and closer.
For an Ackermann steered vehicle where the `base_link` is on the rear axle as you approach the goal the carrot ends up behind the front steering wheels. This makes it very unstable, and small cross-track errors lead to big deviations in yaw. With very little path left this very often ends up with the robot reaching the goal with very poor alignment (on our platform anyway!).

I have seen the other work done here on slowing down on approach to goal, and that is something that we are interested in because we also have a large platform with limits on how quickly it can speed up and slow down. But this PR isn't meant to address any of those things - just tracking of the path as the robot approaches the goal. All the calcs about when to start slowing down are done with reference to the last pose in the path, so this PR does not change that.

SteveMacenski: The interpolation is mostly for smooth motion purposes so that between iterations that the velocity curves are more continuous -- which is important for larger vehicles that might have pretty sparse paths (this was created by an ag-tech company). I don't think the rationale is really about being on the exact path with exact path points. 

> WRT the default behaviour - yes, I believe it should always be enabled ... I have seen the other work done here on slowing down on approach to goal, But this PR isn't meant to address any of those things 

Ah, do you think we could? For instance
- Do this forward projection to the lookahead distance
- But reduce the velocity in proportion to the distance remaining to the actual goal pose

If we had something like that, we could simply make it default then without a parameterization. In fact, isn't that the behavior with this enabled + `approachVelocityConstraint` (which is non-optional)? I believe the slowdown logic is already decoupled the lookahead distance

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -322,23 +322,42 @@ geometry_msgs::msg::PoseStamped RegulatedPurePursuitController::getLookAheadPoin
...
+      // This will ensure that the minimum lookahead distance constraint is preserved
+      // and that the gain of the steering isn't effectively increased near the goal
+      auto dx = goal_position.x - prev_position.x;
+      auto dy = goal_position.y - prev_position.y;
+
+      auto d = std::hypot(dx, dy);
+      // Projecting past the last point by the lookahead distance ensures that the point
+      // is at least the lookahead distance away
+      // (maybe more, but we will interpolate in the next step)
+      goal_position.x += dx / d * lookahead_dist;
```
SteveMacenski, line 345: I... don't think this is correct

If the total distance between points is 1 unit and equal component X and Y, then `dx` and `dy` would be `0.70710678118` for a total `d` of `1.0`. In that case, we're adding more to the goal pose than would be the length of the `lookahead_distance`. 

I think you need to do some trig with the angles to see what component of each you need to apply like the commit I originally linked to did, unless I'm missing something.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
james-ward: > The interpolation is mostly for smooth motion purposes so that between iterations that the velocity curves are more continuous -- which is important for larger vehicles that might have pretty sparse paths (this was created by an ag-tech company). I don't think the rationale is really about being on the exact path with exact path points.

I was just guessing - I would never want to run without the interpolation on for exactly the reasons you state! ;) Even if the path wasn't sparse.

> If we had something like that, we could simply make it default then without a parameterization. In fact, isn't that the behavior with this enabled + approachVelocityConstraint (which is non-optional)? I believe the slowdown logic is already decoupled the lookahead distance

Yes, the slowdown is already decoupled - it uses the current pose and last pose in the path as parameters. The carrot is not involved in those calculations.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
james-ward: Yes, that is deliberate because the direction is the important bit. The distance is truncated in the following code by the call:
https://github.com/ros-planning/navigation2/blob/681d7e15acbb27618272e0530d7b8f385d614fad/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L353

> I think you need to do some trig with the angles to see what component of each you need to apply like the commit I originally linked to did, unless I'm missing something.

This is the same as in the linked commit https://github.com/ros-planning/navigation2/blob/a9d8c4c9889c7d8684296c5d7733f507255520c1/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L232-L239 just without using trig functions to convert to an intermediate angle.

The distance is set to the lookahead distance so that it is outside the radius and `circleSegmentationIntersection()` succeeds.

The difference is approach is that in the previous work in https://github.com/ros-planning/navigation2/commit/a9d8c4c9889c7d8684296c5d7733f507255520c1 the remaining lookahead distance is calculated by taking the "path end as carrot" distance, and working out how much shorter than the lookahead distance it is. This distance is then added to the last segment of the path and treated as the new carrot. (There is an implicit assumption that the direction from the robot to the last path pose is the same as the direction of the last segment, but I suspect that in practice this would make very little difference/be very unlikely to cause problems as it would need odd dog-leg paths at the scale of the lookahead distance for it to matter).
My approach checks to see if the last pose is being selected before the call to find the circle-segment intersection (which runs for all interpolations along the path), and shifts the goal pose before this call. So every iteration uses the circle-segment intersection approach, even when it is on the last segment.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
SteveMacenski: Honestly, this might be a kick for me to disable the option for interpolation and just have it happen. That was a new feature so I didn't want to break users, but its had enough baking time now I think it can just be always included. But something to do after this PR is figured out!

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
doisyg: Trying to find a slot to test on a real big robot. Give me 24h

> (There is an implicit assumption that the direction from the robot to the last path pose is the same as the direction of the last segment, but I suspect that in practice this would make very little difference/be very unlikely to cause problems as it would need odd dog-leg paths at the scale of the lookahead distance for it to matter).

Indeed, I developed and tested https://github.com/ros-planning/navigation2/commit/a9d8c4c9889c7d8684296c5d7733f507255520c1 for straight paths only. I agree this should be mostly fine for general paths but I did not make the exercise of thinking and testing about corner cases.

doisyg:
> Just a reminder to split this from the use_interpolation parameter, these are different behaviors and I've gotten the same remark from another reviewer that they want to see that as well 

Agreed, one could want to interpolate on the path without interpolating the lookahead point after the goal. And if interpolating on the path becomes the default and the parameter disappears, we still want to be able to control if we want to interpolate the lookahead point or not.

SteveMacenski:
@james-ward any update?

doisyg:
I re did it from scratch started from main branch in https://github.com/botsandus/navigation2/commit/4d570f85af869af8ff9544f50f5dd6f6191b7f1d
Currently testing (so far so good after one week). I will clean and PR soon

james-ward:
> @james-ward any update?

Sorry, did not get any notifications of new comments on this thread. Odd...

The approach by @doisyg above is basically what I have done on a different branch - I chose to start again. https://github.com/james-ward/navigation2/tree/preserve-last-segment
The approach in the branch pointed to by this current PR is not good. The key, as in @doisyg's commit above, is that if the path is pruned by the controller to only have one pose left then you cannot do the projection properly. Projecting from the robot through the one remaining pose won't work because when the robot is close and has a small crosstrack error it will project the point a long way to the side of the robot, and cause large steering outputs.
I did it by injecting an additional pose into the path upon receipt an infinitesimally small distance from the endpoint, but perhaps @doisyg's method is better in just preventing the controller from pruning if there are only two points left.

In all of these approaches there is still an issue with projecting past cusps in paths with reversals. I cannot see a nice way to do that at the present time.

james-ward:
> The approach by @doisyg above is basically what I have done on a different branch - I chose to start again. https://github.com/james-ward/navigation2/tree/preserve-last-segment

For reference we have been using that branch on our platform for the last three weeks and it has been a great improvement.

SteveMacenski:
@james-ward do you think we should use your new branch or @doisyg's for merging a fix into Nav2? Just trying to make sure we can rectify these two similar directions to get something that satisfies everyone!

james-ward:
I'm happy with @doisyg's approach. I didn't do it that way because I was unsure if there might be problems if the controller has poses behind the robot after it has finished pruning.

I was also going to update the documentation and add a warning if the minimum lookahead distance is non-zero and the new projection flag isn't turned on.

If you are happy for me to make the history in this PR invalid I can force push my new branch to replace the one referenced in this PR which will make looking at my patch set easier.

SteveMacenski:
Yes please! 

@doisyg do you agree we should use his as well? `[rectification in progress. . . . 90%]`

SteveMacenski:
I'll let @doisyg give the first review analysis when he has some time

james-ward:
Hi @doisyg @SteveMacenski 
What can I do to help get this progressed? I've had to rebase twice due to other PRs getting merged - the longer it goes on the more likely I have to do it again. ;)
FWIW we are still running this patch on our platforms and gives much better behaviour at the end of the path (ie no heaving the steering to one side or the other).

SteveMacenski:
I think its on @doisyg list for this week or next?

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -345,9 +373,47 @@ geometry_msgs::msg::PoseStamped RegulatedPurePursuitController::getLookAheadPoin
...
+  const auto d = std::hypot(dx, dy);
+
+  // Create a carrot the lookahead distance from the goal pose
+  auto pose = goal_pose;
+  pose.pose.position.x += dx / d * lookahead_dist;
+  pose.pose.position.y += dy / d * lookahead_dist;
+
+  // Interpolate as usual
+  auto point = circleSegmentIntersection(
+    prev_pose.pose.position, goal_pose.pose.position, lookahead_dist);
```
RhysMcK, line 408: Is this correct?, or should it be...
`circleSegmentIntersection(prev_pose.pose.position, pose.pose.position, lookahead_dist);`

james-ward: You are correct. Changed.

(Tests passed because the circle segment intersection calculation will calculate an intersection for a segment projected past the end point anyway)

james-ward:
@RhysMcK Thanks for looking at this. I had to rebase a couple of times due to upstream changes, and some formatting changes came through. I have added a commit to remove them, so the only changes in this PR are now the algorithmic ones. Much easier to review!

SteveMacenski:
#4140 supersedes. Thanks @james-ward for your help and prototype here!

PR #3792, opened 2023-09-01
fixing external users of the BT action node template

SteveMacenski:
@doisyg


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -122,7 +123,12 @@ bool BtActionServer<ActionT>::on_configure()
   // Support for handling the topic-based goal pose from rviz
   client_node_ = std::make_shared<rclcpp::Node>("_", options);
 
-  // Declare parameters for client node
+  // Declare parameters for client node to share with BT nodes
+  // Declare if not declared in case being used an an external application
```
doisyg, line 127:
```suggestion
  // Declare if not declared in case being used an external application
```

doisyg (commented):
Thanks! Works for me

SteveMacenski:
Green CI is so nice. Thanks #dexory

PR #3795, opened 2023-09-05
Fix `min_points` comparison check

tonynajjar:
Fix `min_points` comparison check in collision_detector to match [collision_monitor](https://github.com/ros-planning/navigation2/blob/910a9f2793b102d3fdf0e1ae46f4a7f5ca1de889/nav2_collision_monitor/src/collision_monitor_node.cpp#L431)


SteveMacenski (approved):

AlexeyMerzlyakov (approved):
Yes, CM has [similar](https://github.com/ros-planning/navigation2/blob/f3a041a7d7081ab35d6c4fcd208d0949829a77e1/nav2_collision_monitor/src/collision_monitor_node.cpp#L431) logic in collision detection. This was initially missed point during the CD review. The test should not be changed, as one point is enough to trigger collision appearance.

Issue #3824, opened 2023-09-19
Collision Monitor: dynamic enabling/disabling of polygons


tonynajjar:
#### Feature description

I'd to be able to dynamically enable and disable specific polygons of the collision monitor node

#### Implementation considerations

Implementation proposition:

- Add an `enabled` member and ROS parameter to `Polygon` (default: true)
- Add dynamic setting of this parameter
- in `collision_monitor_node`, skip polygon if not enabled

PR #3825, opened 2023-09-19
collision_monitor: dynamic polygon and source enable/disable

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/3824 |
| Robotic platform tested on | Own gazebo simulation |

---

* Added dynamic enabling/disabling of collision monitor polygons and sources
* Added test

---

#### For Maintainers: 


nav2_collision_monitor/src/polygon.cpp
```diff
@@ -487,6 +500,26 @@ void Polygon::updatePolygon(geometry_msgs::msg::PolygonStamped::ConstSharedPtr m
...
+{
+  rcl_interfaces::msg::SetParametersResult result;
+
+  for (auto parameter : parameters) {
+    const auto & param_type = parameter.get_type();
+    const auto & param_name = parameter.get_name();
+
+    if (param_type == rcl_interfaces::msg::ParameterType::PARAMETER_BOOL) {
+      if (param_name == polygon_name_ + "." + "enabled") {
+        enabled_ = parameter.as_bool();
```
tonynajjar, line 515: setter function better?

SteveMacenski (changes requested):
Minor changes but LGTM. Some test would be nice though

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: Nope, this is fine!

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -382,6 +382,10 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
   std::shared_ptr<Polygon> action_polygon;
 
   for (std::shared_ptr<Polygon> polygon : polygons_) {
+    if (!polygon->getEnabled()) {
+      // If the polygon is disabled, skip it
```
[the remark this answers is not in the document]
tonynajjar: Done. One more question, do you think I should also skip publishing the polygon in `publishPolygons` if it's disabled?

tonynajjar:
Also, since collision monitor and detector use the same Polygon, detector polygons will also have this new parameter. We could implement the same logic in collision_detector but I don't see much benefit to it except consistency/parallelism between the two nodes. What do you think?

tonynajjar:
> Doesn't hurt anyone and really easy.

One drawback I see to disabling a polygon in collision_detector is that it would mess up the indexing of polygons in `CollisionDetectorState` msg; if some node subscribing to this topic has saved at initialization that a polygon is at a certain index, then disabling a polygon would cause this node to get the wrong index in `detections` (or even access an index out of range).

The reason why I said I see no value in disabling a polygon in collision detector is because if you don't want the status of a certain polygon, you just don't use it.

If you still think that we should add the functionality and it's safe to expect that the user should always check if `CollisionDetectorState` msg has changed, then I'm fine with implementing it

tonynajjar:
@SteveMacenski in my last commit I also added the dynamic enabling/disabling of a source

tonynajjar (commented):
I have a general question about dynamic parameter setting: from the Nav2 codebase I see that you prefer having these dynamicParametersCallback in which we set internal member variables with the new parameters.

Another way would be to directly read the parameter from the parameter server, roughly:


```
    auto node = polygon->node_.lock();
    if (!node->get_parameter(polygon->getName() + ".enabled").as_bool()) {
      continue;
    }
```

The benefit here is less boilerplate code. What's the con, is it slower? Or why do you prefer the first option?

SteveMacenski:
> if some node subscribing to this topic has saved at initialization that a polygon is at a certain index

That's really not good practice and I'd put that into the category of "not our problem". Also, aren't only the polygons in collision published? If so, then there's no order to be assumed because there's no consistent polygon collision. 

> If you still think that we should add the functionality

I do

> dynamicParametersCallback in which we set internal member variables with the new parameters

Because updating a parameter may require reinitializing other areas that parameter is being used in or you could be triggering an action mid-parameter update if trying to update many - so then the state of that action is not fully defined. Further, design wise, its not wise to inject ROS code into your deep algorithm logic, so getting the ROS params and converting to internal types lets you keep ROS out of the library cores + gives you control to explicitly accept or reject parameter updates if they're bogus or invalid, which happens

nav2_collision_monitor/src/source.cpp
```diff
@@ -37,10 +37,23 @@ Source::Source(
   transform_tolerance_(transform_tolerance), source_timeout_(source_timeout),
   base_shift_correction_(base_shift_correction)
 {
+  RCLCPP_INFO(logger_, "[%s]: Creating Source", source_name_.c_str());
```
SteveMacenski, line 40: Remove here and destructor

SteveMacenski:
@AlexeyMerzlyakov please review

SteveMacenski (approved):

SteveMacenski:
@tonynajjar just needs migration guide entry for navigation.ros.org + configuration guide for the new parameters

tonynajjar:
https://github.com/ros-planning/navigation.ros.org/pull/471

SteveMacenski:
Great! Thanks! This is nice to have!

Issue #3805, opened 2023-09-12
Nav2 Lifecycle Manager Diagnostic Publishing


Plaqueoff:
## Bug report

**Required Info:**

Operating System: Ubuntu 22.04
ROS2 Version: Humble
Version or commit hash: ead48a9039c80cadb18dcfbb21fff0bdd496ef03
Package: nav2_lifecycle_manager

#### Description of the Bug

When the lifecycle manager's pause method is called, a diagnostic status of ERROR is published with the message "Nav2 is inactive". We have a robot with a few states, where we want nav2 to be paused, and I don't see why the paused state would be an error state, so I wished to ask if this is the wanted behavior for publishing diagnostics.

#### Expected behavior

Navigation nodes could be paused through lifecycle manager without error diagnostics.

#### Actual behavior

Error diagnostic is published when navigation nodes are paused through lifecycle manager.

SteveMacenski:
Seems reasonable! This is low on the priority list for Open Navigation to work on, so if this is something you need on a reasonable timeline, I suggest you contribute a PR that we can review and merge!

Plaqueoff:
I'll try to get a PR out if not today by the end of the week!

SteveMacenski:
Imminently to merge!

PR #3830, opened 2023-09-22
Stop error diagnostics when pausing nav

Plaqueoff:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3805  |
| Robotic platform tested on | gazebo simulation of Turtlebot |

---

* Added enum of node states and a variable to keep track of current state of managed nodes.
* Creating diagnostic message based on the current state of managed nodes.

Could add a table of diagnostic messages with situations where they are published.

---

#### For Maintainers: 


nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -150,12 +150,16 @@ LifecycleManager::isActiveCallback(
...
-  if (system_active_) {
-    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::OK, "Nav2 is active");
-  } else {
-    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::ERROR, "Nav2 is inactive");
+  if (state_of_managed_nodes_ == NodeState::ACTIVE || state_of_managed_nodes_ == NodeState::PAUSED) {
+    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::OK, "Managed nodes' status is OK");
+  } else if (state_of_managed_nodes_ == NodeState::UNKNOWN){
+    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::ERROR, "An error has occurred during node state transition");
+  } else if (state_of_managed_nodes_ == NodeState::UNCONFIGURED){
+    stat.summary(diagnostic_msgs::msg::DiagnosticStatus::WARN, "Managed nodes have not been configured");
```
SteveMacenski, line 160: Question, not a request (yet) for a change: Is it really a warning that its just not setup yet? 

@mikeferguson I'd appreciate your opinion here if you had a few moments

Plaqueoff: My rationale was simply that it might be useful to get warning if the nodes have not been configured (and additionally since it was an error before users might expect some yelling from diagnostics), but I can see it being just a normal state. Don't have strong feelings on it one way or another.

SteveMacenski: Lets make it a normal state. Its not totally unheard of for folks to have many operating modes for certain nodes (e.g. swapping from GPS to lidar-localization when going from outdoor to indoor) which are implemented by transitioning up / down the appropriate items. I think its not an warning state. Lets go ahead and change that

Plaqueoff: Also changed the if-else chain into a switch case.

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -38,6 +38,17 @@ namespace nav2_lifecycle_manager
 using namespace std::chrono_literals;  // NOLINT
 
 using nav2_msgs::srv::ManageLifecycleNodes;
+
+/// @brief Enum to for keeping track of the state of managed nodes
+enum NodeState
+{
+  UNCONFIGURED,
```
SteveMacenski, line 45: The 4 primary states are unconfigured, inactive, active, and finalized. I don't know what "paused" is

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -150,13 +150,33 @@ LifecycleManager::isActiveCallback(
...
+    case NodeState::FINALIZED:
+      error_level = diagnostic_msgs::msg::DiagnosticStatus::WARN;
+      message = "Managed nodes have been shut down";
+      break;
+    case NodeState::UNKNOWN:
+      error_level = diagnostic_msgs::msg::DiagnosticStatus::ERROR;
+      message = "An error has occurred during a node state transition";
+      break;
   }
+  stat.summary(error_level, message);
```
SteveMacenski, line 179: Missed `shutdownAllNodes()`

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -295,6 +317,7 @@ bool
 LifecycleManager::shutdown()
 {
   system_active_ = false;
```
SteveMacenski, line 319: You can remove `system_active_` and replace the times its checked with `state_of_managed_nodes_ == NodeState::ACTIVE` as an inlined separate method `isActive()`.

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -283,10 +303,12 @@ LifecycleManager::startup()
     !changeStateForAllNodes(Transition::TRANSITION_ACTIVATE))
   {
     RCLCPP_ERROR(get_logger(), "Failed to bring up all requested nodes. Aborting bringup.");
+    state_of_managed_nodes_ = NodeState::UNKNOWN;
     return false;
   }
   message("Managed nodes are active");
   system_active_ = true;
+  state_of_managed_nodes_ = NodeState::ACTIVE;
```
SteveMacenski, line 311: Rename: managed_nodes_state_

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -150,13 +150,33 @@ LifecycleManager::isActiveCallback(
...
+  unsigned char error_level;
+  std::string message;
+  switch (state_of_managed_nodes_) {
+    case NodeState::ACTIVE:
+      error_level = diagnostic_msgs::msg::DiagnosticStatus::OK;
+      message = "Managed nodes are active";
+      break;
+    case NodeState::PAUSED:
+      error_level = diagnostic_msgs::msg::DiagnosticStatus::OK;
+      message = "Managed nodes are paused";
```
SteveMacenski, line 164: inactive

SteveMacenski (changes requested):
After these clean up steps, happy to merge!

SteveMacenski:
Also, pull in main, there's a test failure that getting the latest will resolve so this can pass CI

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
Plaqueoff: Indeed it is, change made!

nav2_lifecycle_manager/src/lifecycle_manager.cpp
Plaqueoff: Moved setting the managed_nodes_state_ from `shutdown()` to `shutdownAllNodes()`. Didn't like setting the variable twice in a row and I think its better to set it in the latter method.

nav2_lifecycle_manager/src/lifecycle_manager.cpp
Plaqueoff: Good point, change made!

nav2_lifecycle_manager/src/lifecycle_manager.cpp
Plaqueoff: Done!

nav2_lifecycle_manager/src/lifecycle_manager.cpp
Plaqueoff: Done!

SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #3863, opened 2023-10-06
Thready safety fixes

LeifHookedWireless:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | No known tickets. |
| Robotic platform tested on | Peanut's in-house robot|

---

Thread safety:
 * Own a copy of data provided by messages.
 * Keep same smart pointer to object within a scope.
 * Atomic copy of smart pointers.
Fixes issues with NaNs in MPPI controller and probably other places.

* No documentation changes needed.

---

* Thread safety is always a concern. Thereare probably cases that I missed.

#### For Maintainers: 


nav2_waypoint_follower/plugins/photo_at_waypoint.cpp
```diff
@@ -138,7 +138,7 @@ bool PhotoAtWaypoint::processAtWaypoint(
 void PhotoAtWaypoint::imageCallback(const sensor_msgs::msg::Image::SharedPtr msg)
 {
   std::lock_guard<std::mutex> guard(global_mutex_);
-  curr_frame_msg_ = msg;
+  curr_frame_msg_ = std::make_shared<sensor_msgs::msg::Image>(*msg);
```
SteveMacenski, line 141: Please explain why storing shared pointers of shared pointer stored data. This should be OK.

nav2_costmap_2d/src/footprint.cpp
```diff
@@ -42,12 +42,14 @@ namespace nav2_costmap_2d
 
 void calculateMinAndMaxDistances(
   const std::vector<geometry_msgs::msg::Point> & footprint,
-  double & min_dist, double & max_dist)
+  double & min_dist_out, double & max_dist_out)
 {
-  min_dist = std::numeric_limits<double>::max();
-  max_dist = 0.0;
+  double min_dist = std::numeric_limits<double>::max();
```
SteveMacenski, line 47: Explain, this should also be fine. This is also not going to be fixing any threading problems since there is still no locking. So I doubt this does anything fix-wise

SteveMacenski (changes requested):
I need some informational context here about the solutions, many of these don't seem like hard fixes or aren't obvious to me why this is necessary. Please provide some explanation

nav2_costmap_2d/src/footprint.cpp
LeifHookedWireless: before this change, referenced `min_dist_out` aka `LayeredCostmap::inscribed_radius_` is set to an illegal value (std::numeric_limits<double>::max) for a substantial amount of time while the actual radius is being calculated.
any code trying to use `inscribed_radius_`, will get float `inf` instead.
the Obstacle critic was getting NaNs this way.
after this change,` inscribed_radius_` remains a legal value throughout the function and is only changed when the new value is known.

nav2_waypoint_follower/plugins/photo_at_waypoint.cpp
LeifHookedWireless: My concern is that the shared object is owned by the Inters-Process Manager, not by the PhotoAtWaypoint node.
The data in the object can be modified outside of the callback.

SteveMacenski: Huh?

nav2_costmap_2d/src/footprint.cpp
SteveMacenski: But there's still a time period that its being set / potentially gotten during the write. This is not a full solution to a threading problem where one exists. Perhaps instead we should just lock `setFootprint` https://github.com/ros-planning/navigation2/blob/57dd8d4f51e8431dcad419d5da843ba9b0ed67c2/nav2_costmap_2d/src/layered_costmap.cpp#L275C22-L275C34

`  std::unique_lock<Costmap2D::mutex_t> lock(*(combined_costmap_.getMutex()));`

nav2_velocity_smoother/src/velocity_smoother.cpp
```diff
@@ -187,7 +187,7 @@ void VelocitySmoother::inputCommandCallback(const geometry_msgs::msg::Twist::Sha
     return;
   }
 
-  command_ = msg;
```
LeifHookedWireless, line 190: after the first time, `command_` is pointing to the same object pointed to by `msg`
the check for NaN doesn't help because `command_ == msg` and if `msg` has NaNs, then so does `command_`.

SteveMacenski: I don't think that's true. When you have `N` messages coming through a subscription, I don't think each message as the same shared pointer object that its underlying content is changed. I believe each message is constructed as a separate shared pointer object so there's no reason for `command_ == msg` yet in `inputCommandCallback` before `command_ = msg;` is set.  

https://github.com/ros2/rclcpp/blob/13abc1beed02a995863c5c115ef758d35485b46b/rclcpp/src/rclcpp/executor.cpp#L666

nav2_costmap_2d/src/footprint.cpp
LeifHookedWireless: the reason I didn't do lock during all of setFootprint() is that the function isn't trivial and takes fair amount of time
better to work on a temp variable and then lock on the assignment.
or simply use atomic assignment.

nav2_velocity_smoother/src/velocity_smoother.cpp
LeifHookedWireless: It was definitely true for me. the same object got reused and modified outside of the callback.
Is this middleware dependent?
what if `can_load_messages` is true?

LeifHookedWireless: not seeing the messages get reused using `rmw_cyclonedds_cpp`

LeifHookedWireless: confirmed that the message objects get reused with `rmw_fastrtps_cpp`
`cmd_sub_->can_loan_messages()` returns true in this case.

nav2_waypoint_follower/plugins/photo_at_waypoint.cpp
LeifHookedWireless: most of my changes are protecting against the `msg` object being modified outside the callback.
if you assume that can't happen, then the code is fine as-was.

SteveMacenski: That can't happen in the case that I'm mentioning above. I didn't check all of them, but I suspect many of them fall into that same category. The executor is single threaded so if this is processing, that promises that another cannot be

nav2_costmap_2d/src/costmap_subscriber.cpp
```diff
@@ -88,7 +88,7 @@ void CostmapSubscriber::toCostmap2D()
 
 void CostmapSubscriber::costmapCallback(const nav2_msgs::msg::Costmap::SharedPtr msg)
 {
-  std::atomic_store(&costmap_msg_, msg);
+  std::atomic_store(&costmap_msg_, std::make_shared<nav2_msgs::msg::Costmap>(*msg));
```
SteveMacenski, line 91: This is already atomically stored/loaded

nav2_costmap_2d/plugins/costmap_filters/speed_filter.cpp
```diff
@@ -168,7 +168,7 @@ void SpeedFilter::maskCallback(
     filter_mask_.reset();
   }
 
-  filter_mask_ = msg;
+  std::atomic_store(&filter_mask_, std::make_shared<nav_msgs::msg::OccupancyGrid>(*msg));
```
SteveMacenski, line 171: I could see this atomic store/load as sensible, but the extra copy over seems not. Keep that for the speed/keepout/binary

nav2_costmap_2d/src/footprint.cpp
SteveMacenski: Looking at its use...

`calculateMinAndMaxDistances` called in `LayeredCostmap::setFootprint`, which is called when a new footprint is set on a callback, dynamic parameter, or initialization.  

Dynamic parameter is registered with `this` as is the footprint subscriber, so minus the lifecycle initialization, its mutually exclusive for being `set`. But the map update thread is separate which calls costmap layers that depend on it. 

So... the right answer is to either lock `setFootprint` or change the set semantics for `inscribed_radius_, circumscribed_radius_` to be atomic. But just making the time of the write smaller to only 

```
  min_dist_out = min_dist;
  max_dist_out = max_dist;
```

Isn't a full solution to promise a complete resolution

nav2_velocity_smoother/src/velocity_smoother.cpp
SteveMacenski: Yicks. @EduPonz are you aware of this? It appears Fast-DDS reuses the same underlying memory for multiple messages. If we store the shared pointer from a ROS callback (required for IPC zero copy), then we're actually updating other information!

SteveMacenski: @LeifHookedWireless what version of ROS / Fast-DDS are you using. OR wants to know to potentially look into this

nav2_waypoint_follower/plugins/photo_at_waypoint.cpp
LeifHookedWireless: it happens using `rmw_fastrtps_cpp` as discussed in the other threads.

nav2_velocity_smoother/src/velocity_smoother.cpp
SteveMacenski: @LeifHookedWireless this should address https://github.com/ros2/rclcpp/pull/2335 can you test?

LeifHookedWireless: I am using a fairly new version of Iron.
rod-iron-rclpp/jammy 21.03-1
ros-iron-fastrtps-cmake-module/jammy 3.0.1-1
ros-iron-fastrtps/jammy 2.10.2-1
ros-iron-rmw-fastrtps-cpp/jammy 7.1.1-2
ros-iron-rmw-fastrtps-shared-cpp/jammy 7.1.1-2

I've switched to Cyclone (`can_loan_messages == false`), but I can reproduce these problems on Fast-DDS (`can_loan_messages==true`).
So I'm not blocked, but I want to make sure the code handles those cases I found.

LeifHookedWireless: is this a bug though?
I thought the point of `can_loan_messages` is that there is a single object shared by the publisher and subscriber to prevent unnecessary copies.
if you are using it the object only in the callback, this is fine.
if you want to use the object outside of the callback, then you need to make your own copy.
that's what this PR does.

SteveMacenski: We can talk about it synchronously, but yes this is a definite bug. Temporary solve here: https://github.com/ros2/rclcpp/pull/2335 to get the broken behavior out as a temporary work around. You should see in the next sync that this works fine. For now, use Cyclone. We'll work on a full solution later

nav2_waypoint_follower/plugins/photo_at_waypoint.cpp
SteveMacenski: ^ if not in Cyclone (so rclcpp bug isn't impacting you) we can revert these since it'll be fixed in the next sync

nav2_velocity_smoother/src/velocity_smoother.cpp
LeifHookedWireless: Sounds like they are talking about not reusing buffers if the ref count is >1.
I think that will work. 
New PR coming.

EduPonz: > Yicks. @EduPonz are you aware of this? It appears Fast-DDS reuses the same underlying memory for multiple messages. If we store the shared pointer from a ROS callback (required for IPC zero copy), then we're actually updating other information! I think that's probably reasonable to do as long as the user doesn't store the pointer (eg you can check the reference counter on the shared pointer if the user has cached it) but when we do store the pointers, that breaks alot of internal program state. If this report is correct, it would break every person using Fast-DDS and storing shared pointers

Hi everyone. Yes, we are aware, let me elaborate. As you know, Zero-copy is a combination of Data sharing and sample loaning. When using data sharing, the history of the DataWriter is stored in a memory mapped file accessible from other processes. This way, DataReaders can read samples directly from this DataWriter's process owned memory. If you add loans (as the ROS 2 API does if possible on the reception side), then the application reads the data directly from there.

Now, in the case of DataSharing, the samples are acknowledged as soon as the receiving application retrieves them for the first time (in opposition of waiting until they return the loan). If we were to block the sample's memory until the readers returned the loan, then one misbehaving reader could potentially completely block the DataWriter's operation as the writer will not be able to find any available memory to write new samples. This means that one reader not returning loans would prevent other readers to receive data.

In DDS, DataWriters are in fact allowed to reuse history samples on keep last history policies for both reliable and best effort communications. If the history policy is set to keep all, then only best effort writers can reuse "unacknowledged" samples. This still applies to data sharing capable writers, so samples will be indeed reused under those circumstances. For this very reason, Fast DDS DataReaders expose an [`is_sample_valid`](https://fast-dds.docs.eprosima.com/en/latest/fastdds/api_reference/dds_pim/subscriber/datareader.html#_CPPv4NK8eprosima7fastdds3dds10DataReader15is_sample_validEPKvPK10SampleInfo) so that applications can check whether their samples have been overidden; sadly this API is not exposed in ROS 2.

Long story short, when receiving data on a zero-copy pipeline, samples are meant to be processed as soon as they arrive. If as in this case the application needs to store the data for later use, then it is advisable to copy it over (or to tune history sizes accordingly if possible). Another possibility would be to [disable loan messages](https://docs.ros.org/en/iron/How-To-Guides/Disabling-ZeroCopy-loaned-messages.html) altogether (which is what you're implicitly doing when switching implementations).

I'll leave here some links for further reading:

* [Fast DDS Zero-copy](https://fast-dds.docs.eprosima.com/en/latest/fastdds/use_cases/zero_copy/zero_copy.html#use-case-zero-copy)
* [Fast DDS Datasharing](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/datasharing.html)
* [Fast DDS accessing received data](https://fast-dds.docs.eprosima.com/en/latest/fastdds/dds_layer/subscriber/dataReader/readingData.html)

SteveMacenski: @EduPonz  This is not a zero-copy pipeline nor did we enable anything to make that the case. From the thread we have in the ROS core developers chat, I think its clear that this is a bug in rclcpp (but also as you point out possibly a bug in Fast-DDS?). We should not have loaned messages enabled unless the user enables them so that users know that they shouldn't be storing memory received in the callback. We have done no such thing, so this shouldn't be a problem for us and is considered a bug -- but possibly a bug in rclcpp enabling Fast-DDS zero-copy without being enabled?

If I did enable zero-copy, then you're right, what we're doing here is bad news. Still, Fast-DDS should be able to detect the reference counter of the shared pointers and know that that data is still being used. Before overwriting, it should be throwing an exception / error message. Simply overwriting memory is totally unacceptable

nav2_waypoint_follower/plugins/photo_at_waypoint.cpp
kfabian: Should it not be sufficient to change the callback signature to `void PhotoAtWaypoint::imageCallback(const sensor_msgs::msg::Image::ConstSharedPtr msg)`?

SteveMacenski: From other conversations in other threads, no, I think not. I think that makes it const for us in the function, but the middleware doesn't have that const-ness

LeifHookedWireless:
Replaced by:
 * https://github.com/ros-planning/navigation2/pull/3876 (can_loan_messages)
 * https://github.com/ros-planning/navigation2/pull/3875 (other thread-safety issues)

nav2_velocity_smoother/src/velocity_smoother.cpp
fujitatomoya: @SteveMacenski @EduPonz @clalancette @wjwwood 

sorry i did not see this thread at all.

@EduPonz as far as i see this issue, this is not related to Data Sharing in Fast-DDS, because Fast-DDS Data Sharing is default disabled (https://github.com/ros2/rmw_fastrtps#enable-zero-copy-data-sharing). besides, Nav2 does not use [LoanedMessage](https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/include/rclcpp/loaned_message.hpp#L32), so publisher does not do any related work neither `Data Sharing` nor `LoanedMessage` at all.

i think the problem is `DataReader` returns its own memory (possibly its own HistoryCache?) to the ROS 2 executor via `__rmw_take_loaned_message_internal`. this is the problem just between `DataReader` and `subscription`.
i do not think this is expected behavior. i guess that probably we want to enable loaned interfaces on subscription only if Fast-DDS Data Sharing is enabled, that is aligned with your scenario described above.

what do you think?

fujitatomoya: Draft PR for above: https://github.com/ros2/rmw_fastrtps/pull/726

wjwwood: > Nav2 does not use [LoanedMessage](https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/include/rclcpp/loaned_message.hpp#L32),

But the executor does use the loaned message feature, just not via the rclcpp class:

https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/src/rclcpp/executor.cpp#L626-L663

wjwwood: I commented on your draft @fujitatomoya, but I think it's not correct. I think loaning messages while not using "DataSharing" is a valid use case.

EduPonz: > @EduPonz This is not a zero-copy pipeline nor did we enable anything to make that the case. From the thread we have in the ROS core developers chat, I think its clear that this is a bug in rclcpp (but also as you point out possibly a bug in Fast-DDS?). We should not have loaned messages enabled unless the user enables them so that users know and made the explicit choice that they shouldn't be storing memory received in the callback. We have done no such thing, so this shouldn't be a problem for us and this behavior is considered a bug -- but possibly a bug in rclcpp enabling Fast-DDS zero-copy without being enabled?

I read zero-copy above, hence my comment. I want to point out that so far this is not a bug on Fast DDS. Furthermore, enabling data sharing on ROS 2 takes some user involvement as it is disabled by default, so I don't think is a risk here.

> If I did enable zero-copy, then you're right, what we're doing here is bad news. Still, Fast-DDS should be able to detect the reference counter of the shared pointers and know that that data is still being used. Before overwriting, it should be throwing an exception / error message. Simply overwriting memory is totally unacceptable

It's not that Fast DDS is not able to detect that the reader is still using the memory, it's more a design decision on what to do on that case. In our opinion being able to stop the remote publishing application from the subscribing one by either consuming all its resources or making it throw is more unacceptable than overriding the readers data and giving and API for the reader to do a sanity check, specially when the behaviour is well documented.

> I commented on your draft @fujitatomoya, but I think it's not correct. I think loaning messages while not using "DataSharing" is a valid use case.

I agree, it is indeed a valid and I'd say common use case. It's even the use-case foresaw in the DDS spec.

fujitatomoya: @EduPonz @wjwwood thanks for sharing insights!

Issue #3749, opened 2023-08-09
Costmap activation blocks for ever if required tf transform is not available


kfabian:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- ROS2 Version:
  - Galactic
- Version or commit hash:
  - 1.0.12-1focal.20221208.082109
- DDS implementation:
  - Cyclone DDS

#### Steps to reproduce issue
Activate the controller server without providing the transformation from `robot_base_frame` to `global_frame`

#### Expected behavior
The transition from configured to active fails

#### Actual behavior
The transitions blocks indefinitely

#### Additional information
In the costmaps `on_activate` method it is waited indefinitely for the transform from `robot_base_frame` to `global_frame` to become available. In my opinion a behavior where `nav2_util::CallbackReturn::FAILURE` is returned after some timeout would make more sense.

SteveMacenski:
A timeout parameter could be made that could fail a transition after `N` seconds of trying. Should be a pretty trivial change in https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_ros.cpp#L276. Would that be something you'd be open to contributing?

SteveMacenski:
Any update? :smile:

SteveMacenski:
I think that you might be overthinking it, but I also could be missing an important detail :-)

The planner/controller servers call `activate` on the costmap node. Configure/Activate, [according to the API](https://docs.ros2.org/latest/api/rclcpp_lifecycle/classrclcpp__lifecycle_1_1LifecycleNode.html) returns the state which you would immediately know in the planner/controller server if there was an issue for itself to fail its own lifecycle transition. There might be some glue to pass around that `false` or throw/catch an exception to get to the right method, but I think that should be straight forward - no?

The costmap transitions mirror the main server's transitions, so as part of planner/controller configuring/activating, we can just pay attention to what the costmap is telling us and rely that as a failure to the main server's transitions if the costmap is unable to do so.

SteveMacenski:
Merged!

PR #3866, opened 2023-10-07
Add a timeout to the wait for transforms step of the costmap activation.

kfabian:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3749 |
| Robotic platform tested on | only tested without robot by optionally manually providing the transform |

---

* Previously the `on_activate` method of the costmap was blocking indefinitely if the transform from the robot base frame to the global frame is not available
* This adds a timeout so that the transition from `configured` to `active` fails if the transform is not available
* The timeout can be configured by a costmap parameter
* The controller server transition from `configured` to `active` fails if the corresponding costmap transitions fails

* There is a new costmap parameter: `wait_for_transforms_timeout`
---

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -242,7 +242,10 @@ ControllerServer::on_activate(const rclcpp_lifecycle::State & /*state*/)
 {
   RCLCPP_INFO(get_logger(), "Activating");
 
-  costmap_ros_->activate();
+  const auto costmap_ros_state = costmap_ros_->activate();
+  if (costmap_ros_state.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
+    return nav2_util::CallbackReturn::FAILURE;
+  }
   ControllerMap::iterator it;
   for (it = controllers_.begin(); it != controllers_.end(); ++it) {
     it->second->activate();
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -129,6 +129,7 @@ void Costmap2DROS::init()
   declare_parameter("rolling_window", rclcpp::ParameterValue(false));
   declare_parameter("track_unknown_space", rclcpp::ParameterValue(false));
   declare_parameter("transform_tolerance", rclcpp::ParameterValue(0.3));
+  declare_parameter("wait_for_transforms_timeout", rclcpp::ParameterValue(10.0));
   declare_parameter("trinary_costmap", rclcpp::ParameterValue(true));
   declare_parameter("unknown_cost_value", rclcpp::ParameterValue(static_cast<unsigned char>(0xff)));
   declare_parameter("update_frequency", rclcpp::ParameterValue(5.0));
@@ -279,6 +280,9 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
 
   RCLCPP_INFO(get_logger(), "Checking transform");
   rclcpp::Rate r(2);
+  const auto wait_for_transforms_timeout = rclcpp::Duration::from_seconds(
+    wait_for_transforms_timeout_);
+  const auto start_wait_for_transforms = now();
   while (rclcpp::ok() &&
     !tf_buffer_->canTransform(
       global_frame_, robot_base_frame_, tf2::TimePointZero, &tf_error))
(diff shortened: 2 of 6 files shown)


nav2_controller/src/controller_server.cpp
```diff
@@ -242,7 +242,10 @@ ControllerServer::on_activate(const rclcpp_lifecycle::State & /*state*/)
 {
   RCLCPP_INFO(get_logger(), "Activating");
 
-  costmap_ros_->activate();
+  const auto costmap_ros_state = costmap_ros_->activate();
```
SteveMacenski, line 245: For planner server too?

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -129,6 +129,7 @@ void Costmap2DROS::init()
   declare_parameter("rolling_window", rclcpp::ParameterValue(false));
   declare_parameter("track_unknown_space", rclcpp::ParameterValue(false));
   declare_parameter("transform_tolerance", rclcpp::ParameterValue(0.3));
+  declare_parameter("wait_for_transforms_timeout", rclcpp::ParameterValue(10.0));
```
SteveMacenski, line 132: How about `initial_transform_timeout`?

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -288,6 +292,21 @@ Costmap2DROS::on_activate(const rclcpp_lifecycle::State & /*state*/)
       " to become available, tf error: %s",
       robot_base_frame_.c_str(), global_frame_.c_str(), tf_error.c_str());
 
+    // Check timeout
+    if (now() > (start_wait_for_transforms + wait_for_transforms_timeout)) {
+      RCLCPP_ERROR(
+        get_logger(), "Failed to activate %s because transform from %s to %s did not become available before timeout",
+        get_name(), robot_base_frame_.c_str(), global_frame_.c_str());
+      // Deactivate already activated publishers
+      footprint_pub_->on_deactivate();
```
SteveMacenski, line 301: Don't deactivate. This will trigger the error state and its not the role of a failed activation to deactivate things according to the ROS 2 lifecycle state machine docs

SteveMacenski (changes requested):
Generally speaking, LGTM!

nav2_costmap_2d/src/costmap_2d_ros.cpp
SteveMacenski: ^ also needs a docs PR for navigation.ros.org adding it + a migration guide entry about the change. Please submit that for the next PR review cycle (I'll probably be able to merge the pair together)

SteveMacenski:
Alot of the system tests failed. I'm going to rerun them again in case that was a fluke, but that might be pointing to some activation errors this PR introduces

nav2_controller/src/controller_server.cpp
kfabian: I haven't considered the planner server, it should get the same change.

nav2_costmap_2d/src/costmap_2d_ros.cpp
kfabian: I am deactivating the lifecycle publishers not lifecycle nodes, I don't think this will trigger the error state. According to the lifecycle design doc, a failure in onActivate will result in the inactive state. So I think if I won't deactivate the publishers they will still be active despite the node being in the inactive state. It would be easiest though to just change the order inside the `on_activate` method, so that the transforms are checked at the top of the method.

kfabian:
> Alot of the system tests failed. I'm going to rerun them again in case that was a fluke, but that might be pointing to some activation errors this PR introduces

Maybe the new timeout is just too low for tests with the gazebo simulation? Unfortunately I could not get the turtlebot simulation running on my Ubuntu arm64 virtual machine.

SteveMacenski (approved):
Still needs docs PR I described but otherwise LGTM!

kfabian:
@SteveMacenski the CI is complaining about reduced test coverage. Looking at the costmap test folder I'm wondering where a suitable test should be located.

SteveMacenski:
A new file would be fine!

kfabian:
> Still needs docs PR I described but otherwise LGTM!

I created the docs PR: https://github.com/ros-planning/navigation.ros.org/pull/477

SteveMacenski:
Waiting on CI

SteveMacenski:
Thanks!!

PR #3875, opened 2023-10-12
Thread safety fixes for footprint

LeifHookedWireless:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | No ticket known |
| Robotic platform tested on |Peanut's in-house robot (sim) |

---

* Footprint calculation was not thread safe and could lead to crashes. e.g. attempting to do collision detection while footprint was updating.
* Code inspection revealed that sensors PointCloud, Range, and Scan also appear to be unsafe.
* Replaces https://github.com/ros-planning/navigation2/pull/3863

* Changed signature of calculateMinAndMaxDistances()

---

* None expected.

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/scan.cpp
+++ b/nav2_collision_monitor/src/scan.cpp
@@ -64,17 +64,17 @@ void Scan::configure()
     std::bind(&Scan::dataCallback, this, std::placeholders::_1));
 }
 
-void Scan::getData(
+bool Scan::getData(
   const rclcpp::Time & curr_time,
   std::vector<Point> & data) const
 {
   // Ignore data from the source if it is not being published yet or
   // not being published for a long time
   if (data_ == nullptr) {
-    return;
+    return false;
   }
   if (!sourceValid(data_->header.stamp, curr_time)) {
-    return;
+    return false;
   }
 
   tf2::Transform tf_transform;
@@ -87,7 +87,7 @@ void Scan::getData(
         base_frame_id_, curr_time, global_frame_id_,
         transform_tolerance_, tf_buffer_, tf_transform))
     {
-      return;
+      return false;
     }
   } else {
     // Obtaining the transform to get data from source frame to base frame without time shift
@@ -98,7 +98,7 @@ void Scan::getData(
         data_->header.frame_id, base_frame_id_,
         transform_tolerance_, tf_buffer_, tf_transform))
     {
-      return;
+      return false;
     }
   }
 
@@ -118,6 +118,7 @@ void Scan::getData(
     }
     angle += data_->angle_increment;
   }
+  return true;
 }
 
 void Scan::dataCallback(sensor_msgs::msg::LaserScan::ConstSharedPtr msg)
--- a/nav2_collision_monitor/src/scan.cpp
+++ b/nav2_collision_monitor/src/scan.cpp
@@ -123,8 +123,7 @@ void Scan::getData(
 
 void Scan::dataCallback(sensor_msgs::msg::LaserScan::ConstSharedPtr msg)
 {
-  sensor_msgs::msg::LaserScan::ConstSharedPtr msg_copy = std::make_shared<sensor_msgs::msg::LaserScan>(*msg);
-  std::atomic_store(&data_, msg_copy);
+  std::atomic_store(&data_, msg);
 }
 
 }  // namespace nav2_collision_monitor
--- a/nav2_collision_monitor/src/scan.cpp
+++ b/nav2_collision_monitor/src/scan.cpp
@@ -70,10 +70,11 @@ void Scan::getData(
 {
   // Ignore data from the source if it is not being published yet or
   // not being published for a long time
-  if (data_ == nullptr) {
+  auto data_in = std::atomic_load(&data_);
+  if (data_in == nullptr) {
     return;
   }
-  if (!sourceValid(data_->header.stamp, curr_time)) {
+  if (!sourceValid(data_in->header.stamp, curr_time)) {
     return;
   }
 
@@ -83,7 +84,7 @@ void Scan::getData(
     // to the base frame and current time
     if (
       !nav2_util::getTransform(
-        data_->header.frame_id, data_->header.stamp,
+        data_in->header.frame_id, data_in->header.stamp,
         base_frame_id_, curr_time, global_frame_id_,
         transform_tolerance_, tf_buffer_, tf_transform))
     {
@@ -95,7 +96,7 @@ void Scan::getData(
     // frames.
     if (
       !nav2_util::getTransform(
-        data_->header.frame_id, base_frame_id_,
+        data_in->header.frame_id, base_frame_id_,
         transform_tolerance_, tf_buffer_, tf_transform))
     {
       return;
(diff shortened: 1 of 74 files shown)


nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -275,9 +275,8 @@ bool LayeredCostmap::isCurrent()
 void LayeredCostmap::setFootprint(const std::vector<geometry_msgs::msg::Point> & footprint_spec)
 {
   footprint_ = footprint_spec;
-  nav2_costmap_2d::calculateMinAndMaxDistances(
-    footprint_spec,
-    inscribed_radius_, circumscribed_radius_);
+  std::tie(inscribed_radius_, circumscribed_radius_) = nav2_costmap_2d::calculateMinAndMaxDistances(
```
SteveMacenski, line 278: This is in effect the same as before, its still not protected from a simultaneous write/read. Needs either atomic access/change or a mutex around the set footprint callback

nav2_collision_monitor/src/scan.cpp
```diff
@@ -70,10 +70,11 @@ void Scan::getData(
 {
   // Ignore data from the source if it is not being published yet or
   // not being published for a long time
-  if (data_ == nullptr) {
+  auto data_in = std::atomic_load(&data_);
```
SteveMacenski, line 73: For all the collision monitor changes: `getData` is triggered by the velocity cmd vel callback. `dataCallback` is triggered by a subscriber that shares the same single threaded executor from the velocity callback. Thus, these are already mutually exclusive and shouldn't require this

SteveMacenski: Thus, I don't think any of the changes to collision monitor are needed?

LeifHookedWireless: This looked like an issue to me, but only by inspecting the code. I didn't verify it with a test.
I will double-check.

SteveMacenski: At one point awhile ago I went through and made sure we locked everything that needed it knowing we have callback groups setup. I skipped costmap 2D because I assumed that'd been stable since ROS 1, that its probably OK. That was a mistake, but I would be surprised if I missed many others. I'm definitely not perfect, but I did think about some of these before :- )

LeifHookedWireless: Yes, you are right. Sensors are OK as-is.
I will revert those changes.

nav2_costmap_2d/src/layered_costmap.cpp
LeifHookedWireless: Yes, of course you are right. I'll protect the assignments.

SteveMacenski:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/10204/workflows/72cd1995-6da4-4624-afd0-0cc42fcfb145/jobs/32392/tests

Linting problems

nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -274,10 +274,12 @@ bool LayeredCostmap::isCurrent()
 
 void LayeredCostmap::setFootprint(const std::vector<geometry_msgs::msg::Point> & footprint_spec)
 {
-  footprint_ = footprint_spec;
-  nav2_costmap_2d::calculateMinAndMaxDistances(
-    footprint_spec,
-    inscribed_radius_, circumscribed_radius_);
+  auto min_max = nav2_costmap_2d::calculateMinAndMaxDistances(footprint_spec);
```
SteveMacenski, line 277: Ok, let me sorry to make you have to do this another time, but let me just suggest a different way here. I really don't love the idea of costmap having any more mutexes than it already has, this could easily create unforeseen problems because of the number of other mutexes at play. I don't know that it _will_ cause any issues, but I will say that the relationship with all the threads, costmap layers, and dynamic footprints are too complex for me to want to only rely on my own 15-minutes-of-code-reviewing to guarantee that this doesn't introduce some rare deadlock condition. 

- Keep this structure where `calculateMinAndMaxDistances` returns the pair 
- Make  circumscribed_radius_, inscribed_radius_ atomics 
- Use atomic set/get to promise they're set changed safely here (no need for the scoped lock then) 
- Perhaps footprint an atomic too?

SteveMacenski (changes requested):

nav2_costmap_2d/src/layered_costmap.cpp
LeifHookedWireless: Atomics could fix `circumscribed_radius_`, `inscribed_radius_`, but `footprint_` is an array of points (`std::vector<geometry_msgs::msg::Point>`) which can't easily be atomic.
A mutex seems like the right solution here.
If you're dead set against it, maybe getFootprint() could return a shared_ptr?

LeifHookedWireless: I made `footprint_` a shared_ptr internally and kept the interface the same for now.
(I would consider returning a shared pointer and not having to copy the array.)
This should handle the worst of the threading problems.
There is still a possibility that a client could do something like grab a inscribed_radius from one update and a circumscribed_radius from a different one, but there is no way that those values would be simply garbage like before.

nav2_costmap_2d/src/layered_costmap.cpp
```diff
@@ -274,10 +274,13 @@ bool LayeredCostmap::isCurrent()
...
 {
-  footprint_ = footprint_spec;
-  nav2_costmap_2d::calculateMinAndMaxDistances(
-    footprint_spec,
-    inscribed_radius_, circumscribed_radius_);
+  std::pair<double, double> inside_outside = nav2_costmap_2d::calculateMinAndMaxDistances(
+    footprint_spec);
+  std::atomic_store(
+    &footprint_,
+    std::make_shared<std::vector<geometry_msgs::msg::Point>>(footprint_spec));
```
SteveMacenski, line 281: This can be a unique pointer if you're not returning the pointer to anyone else but always dereferencing it to send a reference of the object in any `getFootprint` calls. Shared pointer has overhead, unique has effectively none.

Also, add a note here that we have these as atomics because the sets here and gets in algorithms' updates aren't locked on the same thread 

After that, this LGTM

SteveMacenski (changes requested):

nav2_costmap_2d/src/layered_costmap.cpp
SteveMacenski: Also, a couple of unusual tests failed on this PR that aren't considered flaky. I'm retriggering, but if that happens again, you may need to look into issues this is introducing

LeifHookedWireless: `atomic_load` and `atomic_store` do not support `std::unique_ptr`.
I think keeping the shared_ptr would be the best option.

SteveMacenski (approved):

SteveMacenski:
@LeifHookedWireless retriggering CI, but you had a failure with obstacle tests in costmap 2D. If the new CI job also fails, that means this has a regression that you should look at. 

But pending that, I approve this PR

SteveMacenski:
Yeah @LeifHookedWireless multiple tests are failing

SteveMacenski (approved):

PR #3876, opened 2023-10-12
Make copy of shared objects in messages (can_loan_messages)

LeifHookedWireless:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | None known |
| Robotic platform tested on | Peanut's in-house robot simulator |

---

* Under certain circumstances, messages passed by shared pointer cannot be used outside the callback.
* Save a copy of message in callbacks, so they can safely be used outside the callback.
* This is necessary when using Fast-DDS which currently enables `can_loan_messages`.
* Not necessary for Cyclone DDS, which disables `can_loan_messages`.
* Might be a RCLPP bug.
* Might be a Middleware bug
* Replaces https://github.com/ros-planning/navigation2/pull/3863

No documentation changes needed.

---

* If the RCLPP/MW behavior changes and the bug is fixed, this PR can be discarded.
* May want to check a flag and not do these copies if can_loan_messages is false.

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/pointcloud.cpp
+++ b/nav2_collision_monitor/src/pointcloud.cpp
@@ -139,7 +139,7 @@ void PointCloud::getParameters(std::string & source_topic)
 
 void PointCloud::dataCallback(sensor_msgs::msg::PointCloud2::ConstSharedPtr msg)
 {
-  data_ = msg;
+  data_ = std::make_shared<sensor_msgs::msg::PointCloud2>(*msg);
 }
 
 }  // namespace nav2_collision_monitor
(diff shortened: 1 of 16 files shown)


SteveMacenski:
This is being resolved so that it isn't an issue with Fast-DDS anymore very shortly. I guess we could merge this in knowing that I'll revert it once done. Would that be OK with you? I don't want to own copies if not strictly necessary (especially wrt sensor and map data which can be _quite large_)

SteveMacenski:
Ugh, I just realized that if we merge this, the same sync that'll fix rclcpp will also add this to the binaries, so there's not really much point in actually merging this for the short term since it'll be fixed for you in binary format at the same time...

Closing :( Lets get those other PRs in though that are permanent fixes to Nav2 problems!

PR #3877, opened 2023-10-12
Use mutex to protect costmap reads.

LeifHookedWireless:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | No tickets known |
| Robotic platform tested on | Peanut's robot (sim) |

---

* Protect costmap reads with the existing mutex.
* Otherwise costmap can be read during a map update and return 0.

* No documentation changes needed.

---

* None expected.

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/controller.cpp
+++ b/nav2_mppi_controller/src/controller.cpp
@@ -83,8 +83,11 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   auto start = std::chrono::system_clock::now();
 #endif
 
-  std::lock_guard<std::mutex> lock(*parameters_handler_->getLock());
+  std::lock_guard<std::mutex> param_lock(*parameters_handler_->getLock());
   nav_msgs::msg::Path transformed_plan = path_handler_.transformPath(robot_pose);
+
+  nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
+  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> cosmtap_lock(*(costmap->getMutex()));
 
   geometry_msgs::msg::TwistStamped cmd =
     optimizer_.evalControl(robot_pose, robot_speed, transformed_plan, goal_checker);
--- a/nav2_mppi_controller/src/controller.cpp
+++ b/nav2_mppi_controller/src/controller.cpp
@@ -87,7 +87,7 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   nav_msgs::msg::Path transformed_plan = path_handler_.transformPath(robot_pose);
 
   nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
-  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> cosmtap_lock(*(costmap->getMutex()));
+  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> costmap_lock(*(costmap->getMutex()));
 
   geometry_msgs::msg::TwistStamped cmd =
     optimizer_.evalControl(robot_pose, robot_speed, transformed_plan, goal_checker);
--- a/nav2_costmap_2d/src/costmap_2d.cpp
+++ b/nav2_costmap_2d/src/costmap_2d.cpp
@@ -265,19 +265,16 @@ unsigned char * Costmap2D::getCharMap() const
 
 unsigned char Costmap2D::getCost(unsigned int mx, unsigned int my) const
 {
-  std::unique_lock<mutex_t> lock(*access_);
   return costmap_[getIndex(mx, my)];
 }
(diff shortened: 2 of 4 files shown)


SteveMacenski:
Where is this a problem, specifically. That's a lot of overhead to obtain a mutex when we call get/set costs thousands of times a loop. I would not merge this without an explicitly application example.

LeifHookedWireless:
A specific example would be:
ObstaclesCritic::score()
ObstaclesCritic::costAtPose()
CollisionChecker::costAtPose()
Costmap2D::getCost()

which can occur simultaneously with

LayeredCostmap::updateMap()
which modifies Costmap2D in several places.

The result I see is that costAtPose() frequently returns 0 cost, in space that is non-zero and possibly even lethal.

LeifHookedWireless:
Given how often getCost() is used, I suspect this is a problem in many places.
It's going to be tricky to optimize for the same reason, but I will take a look.

SteveMacenski:
@LeifHookedWireless 

The algorithms lock the layered costmap's mutex when starting to plan: https://github.com/ros-planning/navigation2/blob/461a7ba1d296ef7224dde10d3e314a8f8d46fb81/nav2_smac_planner/src/smac_planner_hybrid.cpp#L316

Thus, updateMap is blocked: https://github.com/ros-planning/navigation2/blob/461a7ba1d296ef7224dde10d3e314a8f8d46fb81/nav2_costmap_2d/src/layered_costmap.cpp#L136

LeifHookedWireless:
Sure, locking the update is half of it.
Locking the read is the (missing) other half.

LeifHookedWireless:
So you're saying you want the costmap locked at a higher level?
Maybe in ObstaclesCritic::score()?
That will fix the case that I am aware of.

Personally, I would make the sure the reads are blocked everywhere, so that every call to getCost is protected, current and future.

SteveMacenski:
I'm saying that the master costmap's mutex is already locked at the start of all of the planner/controller algorithms, for example as I linked above

LeifHookedWireless:
I think that's true for all of the planners, and I see a lock in the DWB controller.
I don't see it in the MPPI controller, which is the one I'm trying to fix.
I think you're saying you want the lock in MPPIController::computeVelocityCommands()

FWIW, I don't see in a costmap lock in the RPP or RotationShim controllers either.

LeifHookedWireless:
I locked the costmap in the MPPI controller.
That seems to fix the problem.

nav2_mppi_controller/src/controller.cpp
```diff
@@ -83,9 +83,12 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   auto start = std::chrono::system_clock::now();
 #endif
 
-  std::lock_guard<std::mutex> lock(*parameters_handler_->getLock());
+  std::lock_guard<std::mutex> param_lock(*parameters_handler_->getLock());
   nav_msgs::msg::Path transformed_plan = path_handler_.transformPath(robot_pose);
 
+  nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
+  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> cosmtap_lock(*(costmap->getMutex()));
```
SteveMacenski, line 90:
```suggestion
  std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> costmap_lock(*(costmap->getMutex()));
```

SteveMacenski: Your branch appears to be protected so I can't commit in this renaming fix. Please commit this in and I'll merge!

SteveMacenski:
>  RPP or RotationShim controllers either

Can you add?

SteveMacenski (approved):

SteveMacenski:
Ignore the linting failures in CI, apparently OSRF must have updated the profile. 

If you get that lock into the other algos, I can merge this in and backport immediately

BriceRenaudeau:
Thanks for this fix, I was also wondering why I get 0 cost sometimes. :+1:

LeifHookedWireless:
>https://github.com/ros-planning/navigation2/blob/461a7ba1d296ef7224dde10d3e314a8f8d46fb81/nav2_smac_planner/src/smac_planner_hybrid.cpp#L316
Shouldn't the planners and controllers use `std::shared_lock()`?
That way they both have read access to the costmap and won't block each other.
I guess we need to compile with C++17 for that...

LeifHookedWireless:
I verified that RotationShim controller also had the issue and that adding the costmap lock fixes it.
I also added the lock to RPP controller, as requested, though I haven't had time to test it.
Will revisit this after ROSCon!

SteveMacenski (approved):

LeifHookedWireless:
> I also added the lock to RPP controller, as requested, though I haven't had time to test it.
Confirmed that the bug was present in RPP (though not nearly as bad).
This PR does seem to fix it.

PR #3878, opened 2023-10-13
Fix for  robot footprint collision in obstacles critic

LeifHookedWireless:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | No known tickets |
| Robotic platform tested on |Peanut's in-house robot (sim) |

---

* Inscribed/Circumscribed costs must be updated to take into account the current shape of the robot.
* `ObstaclesCritic::findCircumscribedCost()` was only being called once, in initialize()
* in `initialize()`, `findCircumscribedCost()` uses placeholder inner/outer radius values which come from` robot_radius`.
* `findCircumscribedCost()` was never called again, and the circumscribed cost was never updated based on the footprint.
* depending on the value of `robot_radius`, this could cause too much or too little footprint checking in `costAtPose()`

* No documentation changes.

---

* The SMAC planner may have a similar problem. I will investigate when I have time.

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/critics/obstacles_critic.cpp
+++ b/nav2_mppi_controller/src/critics/obstacles_critic.cpp
@@ -91,6 +91,7 @@ float ObstaclesCritic::findCircumscribedCost(
 
   circumscribed_radius_ = static_cast<float>(circum_radius);
   circumscribed_cost_ = static_cast<float>(result);
+
   return circumscribed_cost_;
 }
 
@@ -115,8 +116,17 @@ void ObstaclesCritic::score(CriticData & data)
   if (!enabled_) {
     return;
   }
-
-  possibly_inscribed_cost_ = findCircumscribedCost(costmap_ros_);
+  auto parent = parent_.lock();
+  auto t0 = parent->now(); // xx!!
+  if (consider_footprint_) {
+    // footprint has almost certainly changed since initialize()
+    possibly_inscribed_cost_ = findCircumscribedCost(costmap_ros_);
+  }
+  auto t1 = parent->now();
+  auto dt = (t1 - t0).seconds();
+  static double fcc_dt = 0.05;
+  fcc_dt = 0.9 * fcc_dt + 0.1 * dt;
+  RCLCPP_WARN_THROTTLE(logger_, *(parent->get_clock()), 500, "findCircumscribedCost took %f ms (%f)", 1000.0*dt, 1000.0*fcc_dt);
 
   // If near the goal, don't apply the preferential term since the goal is near obstacles
   bool near_goal = false;
@@ -170,6 +180,12 @@ void ObstaclesCritic::score(CriticData & data)
     (repulsion_weight_ * repulsive_cost / traj_len),
     power_);
   data.fail_flag = all_trajectories_collide;
+
+  auto t2 = parent->now();
+  dt = (t2 - t0).seconds();
+  static double score_dt = 0.05;
+  score_dt = 0.9 * score_dt + 0.1 * dt;
+  RCLCPP_WARN_THROTTLE(logger_, *(parent->get_clock()), 500, "score took %f ms (%f)", 1000.0*dt, 1000.0*score_dt);
 }
 
 /**
--- a/nav2_mppi_controller/src/critics/obstacles_critic.cpp
+++ b/nav2_mppi_controller/src/critics/obstacles_critic.cpp
@@ -55,6 +55,13 @@ float ObstaclesCritic::findCircumscribedCost(
 {
   double result = -1.0;
   bool inflation_layer_found = false;
+
+  const double circum_radius = costmap->getLayeredCostmap()->getCircumscribedRadius();
+  if (static_cast<float>(circum_radius) == circumscribed_radius_) {
+    // early return if footprint size is unchanged
+    return circumscribed_cost_;
+  }
+
   // check if the costmap has an inflation layer
   for (auto layer = costmap->getLayeredCostmap()->getPlugins()->begin();
     layer != costmap->getLayeredCostmap()->getPlugins()->end();
@@ -66,7 +73,6 @@ float ObstaclesCritic::findCircumscribedCost(
     }
 
     inflation_layer_found = true;
-    const double circum_radius = costmap->getLayeredCostmap()->getCircumscribedRadius();
     const double resolution = costmap->getCostmap()->getResolution();
     result = inflation_layer->computeCost(circum_radius / resolution);
     inflation_scale_factor_ = static_cast<float>(inflation_layer->getCostScalingFactor());
@@ -83,7 +89,9 @@ float ObstaclesCritic::findCircumscribedCost(
       "significantly slow down planning times and not avoid anything but absolute collisions!");
   }
 
-  return static_cast<float>(result);
+  circumscribed_radius_ = static_cast<float>(circum_radius);
+  circumscribed_cost_ = static_cast<float>(result);
+  return circumscribed_cost_;
 }
 
 float ObstaclesCritic::distanceToObstacle(const CollisionCost & cost)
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
I approve, I’m waiting until I’m in front of a computer again this evening to test the runtime considerations.

BriceRenaudeau:
I had similar issues with the `obstacles_critic`. and that's why we developed a similar critic.

The radii are not updated correctly as @LeifHookedWireless mentioned. This modification will help with dynamic footprint.

But there is still several issues with the `obstacles_critic`:

- The `obstacles_critic` is using the cost from inflate layer to compute distances by using the invert function of the cost. It makes inflation_layer `cost_scaling_factor` tunning useless, its good because it reduces tunning and it may be dangerous as it has a quantization effect on the distance).
- The score is computed differently when the robot is far from obstacles and when it is close.
https://github.com/ros-planning/navigation2/blob/7009ffba5f85c50ac97fd0057924b0f1447c5e85/nav2_mppi_controller/src/critics/obstacles_critic.cpp#L147-L151
Why is the inflation radius used in the obstacle score? There is an X20 scale difference as `collision_margin_distance_ ~ 0.1` and `inflation_radius_ ~ 2`.
- If the footprint has some points in the inscribed area, the score is the same regardless of the distance to the obstacle. Because `footprintCostAtPose` returns the highest cost which will be "INSCRIBED" whatever the distance to the obstacle, `distanceToObstacle(cost)` will return the same distance so the score will be the same in a situation where it is important to evade obstacles.

SteveMacenski:
I won't argue that this is perfect, but I'll say that I tested _alot_ of formulations that didn't work nearly as well as this one. I'm not stuck to it what so ever if we have some alternatives that would be better / faster / stronger. 

>  that's why we developed a similar critic

Plz share what you have that works better?

> The obstacles_critic is using the cost from inflate layer to compute distances by using the invert function of the cost. It makes inflation_layer cost_scaling_factor tunning useless, its good because it reduces tunning and it may be dangerous as it has a quantization effect on the distance).

That's intentional. Decoupling the potential field so that the critic can decide what to do using the same units (meters) as the other critics to keep things proportionate so that the tuning weights are the only difference is a nice to have thing. Mostly though, I did this because when trying to use raw costs and normalizing them, I could never really get good performance and made the tuning way-way-way-way too sensitive to minuscule changes in the cost scaling factor and inflation radius. It was absolutely insane the degree that which you had to tune to get even remotely reasonable results. It was untenable in the formulations I tried. 

> The score is computed differently when the robot is far from obstacles and when it is close.

Of course, I don't see why you mention this is a 'bad' thing. We need to have very high cost for things that basically in collision, but you don't want to necessarily have high cost when you're further away from things. Decoupling "nice to have" behavior from "critical" behavior allows you to tune more finely. If we could find a mathematical model that gave us both of those at the same time, I'd be more than happy to test / use it. I think this is a really nice behavior to be able to tune how we act in "normal" situations separately from how we penalize near-obstacle behavior. 

> There is an X20 scale difference

Yes, but they're all in meters, like all other critics using SI units to keep the scales consistent between critics. The critical weight should just be 20x larger. That seems completely normal and reasonable. We weight things more highly when we care about them or they're smaller. 

> If the footprint has some points in the inscribed area, the score is the same regardless of the distance to the obstacle. Because footprintCostAtPose returns the highest cost which will be "INSCRIBED" whatever the distance to the obstacle, distanceToObstacle(cost) will return the same distance so the score will be the same in a situation where it is important to evade obstacles.

This is a fair point. There's a reason why TSDF is on [my potential improvements list](https://github.com/ros-planning/navigation2/issues/3351)

BriceRenaudeau:
Thans for you detailed reply. 
I think it is beyond to this issue now so I will keep it short and I will be happy to discuss in another thread.

The critic we are using is a simplified version of the `obstacles_critcs`. It uses inflation costs directly as the exponential creates the behavior we want (close = very detailed high scores and far = low scores).  There is no distance computation (m). It works well enough and allow custom inflations to be taken into acount.

SteveMacenski:
@BriceRenaudeau can you share a branch or open a ticket to chat?

@LeifHookedWireless did you try the references item so we don't recompute every iteration? I know you were at ROSCon like I, so I don't expect yet, just jogging our collective memories

LeifHookedWireless:
> @LeifHookedWireless did you try the references item so we don't recompute every iteration? I know you were at ROSCon like I, so I don't expect yet, just jogging our collective memories
@SteveMacenski Sorry, been busy. I'll look into it soon.

LeifHookedWireless:
I'm not seeing any way to use references here.
The CircumscribedCost needs to be recalculated whenever the costmap or footprint changes, which will happen at least once between `ObstaclesCritic::initialize()` and `ObsctacleCritics::score()`.
It's not stored anywhere.

It's not trivial to keep track of when those two have changed.
Since it is a fairly lightweight calculation, I suggest just computing it before `score()` every time.

SteveMacenski:
Can you show that its a lightweight operation with some timers? How long is this taking?

LeifHookedWireless:
> Can you show that its a lightweight operation with some timers? How long is this taking?
Compared to?

LeifHookedWireless:
If we assume that the costmap resolution and inflation layer parameters don't change during navigation, then we only have to recalculate the cost if `getCircumscribedRadius()` changes, so that could be an optimization.

SteveMacenski:
Seems reasonable, can you just put a timer around that function and benchmark it to know that this isn't going to be some big performance hit?

LeifHookedWireless:
I added some timers and ran some tests in my simulated hallway including passing through a narrow doorway.
With `consider_footprint: false`, I saw `ObstaclesCritc:score()` take about 4ms on average (7ms max) in the hallway and doorway.
With `consider_footprint: true` I was `ObstaclesCritc:score()` take about 4ms on average (8ms max) in the hallway, and over 100ms passing through the doorway.
`findCircumscribedCost` is less than 1% of the cost in any case.

LeifHookedWireless:
I have a suggestion to address the high cost of collision detection when using the footprint.
Since the footprint code is only active near obstacles, a lot of the checks are very close to each other and I suspect that many are effectively identical to each other.
Caching results in the FootprintCollisionChecker may result in a dramatic speed-up.
I'm interested in trying this, though it may be a week or two away depending on how busy I am.

nav2_mppi_controller/src/critics/obstacles_critic.cpp
```diff
@@ -108,6 +117,12 @@ void ObstaclesCritic::score(CriticData & data)
     return;
   }
 
+  if (consider_footprint_) {
+    // footprint has almost certainly changed since initialize()
```
SteveMacenski, line 121: Replace:

`// footprint may have changed since initialization if user has dynamic footprints

SteveMacenski:
Sorry, maybe I should have been more explicit. Put a timer only around that `findCircumscribedCost` in score and let me know specifically how long that takes alone. 

> Caching results in the FootprintCollisionChecker may result in a dramatic speed-up.

Perhaps, but how do you plan to cache anything if we're working in continuous coordinates? There will be almost guaranteed to be no 2 identical poses in any of the trajectories being scored. They might be similar, but not the same. You'd have to bin both the orientation and the cells to approximately collision check, so you'd have to decide how it is you'd want to do that binning.

We do something related in the Smac Planner, where its version of the collision checker precomputes the footprints at the binned orientations so that we're not owning sin/cos operations in the checks and are simply translating the footprints to the current position to check at the right angle for each bin. You could do something like that for approximate checks so that you have a cache of `x,y` in costmap cells and `yaw` of precomputed angular quantizations.  

Something of that matter is also on my mind if one had a GPU: where it might make sense to literally brute force the entire costmap at every cell / angle its collision state using the GPU and doing lookups here from that computed cache.

SteveMacenski:
>  You can hash those 4 ints as a key and use it to cache the results.

I think that's unlikely to result in any improvements, but I could be wrong! I doubt there are that many redundant checks for the exact same line segment. But hey, I'm happy to be surprised. Just remember that hashtable lookups are O(1), but that doesn't mean they're fast. I've found from experience with the Smac Planner and others in Nav2 that creating a storage semantic in `std::vector` is far, far faster.

SteveMacenski (approved):

PR #3879, opened 2023-10-13
Publish collision points for debug purposes

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Custom Simulation |

---

* Added publish of collision points in form of MarkerArray for debugging purposes

* Docs + Migration guide

---

#### For Maintainers: 


nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -376,6 +395,36 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
     source->getData(curr_time, collision_points);
   }
 
+  if (collision_points_marker_pub_) {
+    // visualize collision points with markers
+    visualization_msgs::msg::MarkerArray marker_array;
+    if (collision_points.size() > 0) {
+      visualization_msgs::msg::Marker marker;
+      marker.header.frame_id = get_parameter("base_frame_id").as_string();
+      marker.header.stamp = rclcpp::Time(0, 0);
```
tonynajjar, line 404: for some reason it shows up in rviz but not in foxglove, still debugging why

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -68,11 +72,16 @@ CollisionMonitor::on_configure(const rclcpp_lifecycle::State & /*state*/)
     std::bind(&CollisionMonitor::cmdVelInCallback, this, std::placeholders::_1));
   cmd_vel_out_pub_ = this->create_publisher<geometry_msgs::msg::Twist>(
     cmd_vel_out_topic, 1);
+
   if (!state_topic.empty()) {
     state_pub_ = this->create_publisher<nav2_msgs::msg::CollisionMonitorState>(
       state_topic, 1);
   }
 
+  if (visualize_collision_points) {
```
SteveMacenski, line 81: Why return? Just store as a class member and grab it here. Lets not complicate the APIs (and would make this not backportable since you change ABI!!)

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -68,11 +72,16 @@ CollisionMonitor::on_configure(const rclcpp_lifecycle::State & /*state*/)
...
+    collision_points_marker_pub_ = this->create_publisher<visualization_msgs::msg::MarkerArray>(
+      "~/collision_points_marker", 1);
```
SteveMacenski, line 83:
```suggestion
      "collision_points_marker", 1);
```

Is the same thing

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -378,6 +397,36 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
     }
   }
 
+  if (collision_points_marker_pub_) {
+    // visualize collision points with markers
+    visualization_msgs::msg::MarkerArray marker_array;
+    if (collision_points.size() > 0) {
```
SteveMacenski, line 403: I'd remove this so that an empty message is published when empty so that we clear out old stuff

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -378,6 +397,36 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
...
+      visualization_msgs::msg::Marker marker;
+      marker.header.frame_id = get_parameter("base_frame_id").as_string();
+      marker.header.stamp = rclcpp::Time(0, 0);
```
SteveMacenski, line 406: ... `now()`?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -378,6 +397,36 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
...
+      marker.ns = "collision_points";
+      marker.id = 0;
+      marker.type = visualization_msgs::msg::Marker::POINTS;
+      marker.action = visualization_msgs::msg::Marker::ADD;
+      marker.scale.x = 0.1;
```
SteveMacenski, line 411: 10cm seems a little big, don't you think?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -378,6 +397,36 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
...
+      marker.scale.y = 0.1;
+      marker.color.r = 1.0;
+      marker.color.a = 1.0;
+      marker.lifetime = rclcpp::Duration(2, 0);
```
SteveMacenski, line 415: Set to zero so that its only just the most recent message perhaps if we publish the empty message?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -378,6 +397,36 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
...
+        p.x = point.x;
+        p.y = point.y;
+        p.z = 0.0;
+        marker.points.push_back(p);
+      }
+      marker_array.markers.push_back(marker);
+    }
+    collision_points_marker_pub_->publish(marker_array);
+  }
+
```
SteveMacenski, line 428: Remove the second space

SteveMacenski (changes requested):
Generally look good!

nav2_collision_monitor/src/collision_monitor_node.cpp
tonynajjar: I would have done it like you proposed as well but that's how it's done for all other parameters of the collision monitor like `cmd_vel_in_topic`, `cmd_vel_out_topic`, `state_topic`. So what do you prefer, breaking ABI or introducing some inconsistency?

nav2_collision_monitor/src/collision_monitor_node.cpp
tonynajjar: Not really, `~/collision_points_marker` will result in `/<node_name>/collision_points_marker`. `collision_points_marker` will result in `/collision_points_marker`. Since all other topics are without `~/`, I'll remove it

nav2_collision_monitor/src/collision_monitor_node.cpp
tonynajjar: Tried that, for some reason it breaks rviz, which complains about the TF map->baselink requiring extrapolation in the future. My `get_parameter("base_frame_id").as_string()` is `base_link` and my Rviz fixed frame is `map`.

I guess the current implementation is also fine according to [this](https://wiki.ros.org/rviz/DisplayTypes/Marker#:~:text=Note%20that%20the,the%20current%20time.)

nav2_collision_monitor/src/collision_monitor_node.cpp
tonynajjar: yep agree, reducing to 0.2

nav2_collision_monitor/src/collision_monitor_node.cpp
SteveMacenski: OK, its just visualization anyway

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -308,6 +323,33 @@ void CollisionDetector::process()
     }
   }
 
+  if (collision_points_marker_pub_) {
```
SteveMacenski, line 326: If you check if the publishers have subscriber > 1, then you can remove the parameter and just publish if anyone is listening. We didn't do this historically because that API used to not always work, but it does now. Also for things like the MPPI trajectories or costmaps, we do want to make sure you want them published at all since they're very heavy. This is not

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -308,6 +323,33 @@ void CollisionDetector::process()
...
+    // visualize collision points with markers
+    visualization_msgs::msg::MarkerArray marker_array;
```
SteveMacenski, line 328: unique ptr

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -308,6 +323,33 @@ void CollisionDetector::process()
...
+
+    for (const auto & point : collision_points) {
+      geometry_msgs::msg::Point p;
+      p.x = point.x;
+      p.y = point.y;
+      p.z = 0.0;
+      marker.points.push_back(p);
+    }
+    marker_array.markers.push_back(marker);
+    collision_points_marker_pub_->publish(marker_array);
```
SteveMacenski, line 350: std::,move(marker_array) once a ptr

This helps with publishing jitter

SteveMacenski (changes requested):
Then should be good to go!

nav2_collision_monitor/src/collision_monitor_node.cpp
SteveMacenski: See new comments below, this may be moot

nav2_collision_monitor/test/collision_monitor_node_test.cpp
```diff
@@ -1170,6 +1200,42 @@ TEST_F(Tester, testSourcesNotSet)
...
+  publishCmdVel(0.5, 0.2, 0.1);
+  ASSERT_TRUE(waitCollisionPointsMarker(500ms));
+  ASSERT_EQ(collision_points_marker_msg_->markers[0].points.size(), 0u);
+
+  publishCmdVel(0.5, 0.2, 0.1);
+  publishScan(0.5, curr_time);
+  ASSERT_TRUE(waitData(0.5, 500ms, curr_time));
+  
+  ASSERT_TRUE(waitCollisionPointsMarker(500ms));
+  ASSERT_NE(collision_points_marker_msg_->markers[0].points.size(), 0u);
```
tonynajjar, line 1234: fails but not sure why yet

nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
```diff
@@ -107,6 +108,7 @@ class CollisionMonitor : public nav2_util::LifecycleNode
    * @param cmd_vel_in_topic Output name of cmd_vel_in topic
    * @param cmd_vel_out_topic Output name of cmd_vel_out topic
    * is required.
+   * @param state_topic topic name for publishing collision monitor state
```
SteveMacenski, line 111: rm

SteveMacenski (commented):
Otherwise LGTM once the test works!

nav2_collision_monitor/test/collision_monitor_node_test.cpp
SteveMacenski: The points are empty... not sure, you should check!

nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
SteveMacenski: Ah.... I was never here...

nav2_collision_monitor/src/collision_monitor_node.cpp
tonynajjar: Actually I just ran into a problem with this: I have both a collision detector and monitor running and since they have the same topic name, not having the namespace makes them publish to the same topic. I would add the `~/` if you don't object

SteveMacenski:
@tonynajjar there are multiple collision monitor pkg test failures (ignore python linting… I’ll try to handle that Friday)

tonynajjar:
> @tonynajjar there are multiple collision monitor pkg test failures (ignore python linting… I’ll try to handle that Friday)

Ahh that came for the last topic change, I forgot to do the same in the test

tonynajjar:
Backport to iron please?

SteveMacenski:
beep boop, mergify, do yo thing

SteveMacenski:
the auto backports failed horribly. The two options are for you to wait until I do a manual backport in bulk before a release or you own a PR to do the backport manually yourself. Yours will be done much faster than mine (weeks)

PR #3880, opened 2023-10-13
[Collision Monitor] Add a watchdog mechanism

doisyg:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Dexory's ARRI |

---

Submitting as a draft as there are a couple of points to discuss.

This PR adds a per source blocking watchdog mechanism, i.e. stop the robot if a source is not publishing at the expected rate (building on the existing `source_timeout` mechanism).

- It adds the parameter `block_if_invalid` (other name suggestions are welcome) either per source or globally. If set to `true`, a source triggering its `source_timeout` will make the robot stop.
- In order to make it work with a set of source/sensor of different frequencies, I made possible to use the parameter `source_timeout` per source. I.e. having different value for each sensors. In order not break the current behavior, if not set, each source specific `source_timeout` are set by default to the global `source_timeout`.
- I modified the `getData` method to return `false` if a source is considered invalid AND `block_if_invalid: true`. @AlexeyMerzlyakov, can you maybe check/have a look if the invalid conditions make sense ?
- I modified symmetrically the collision detector with the new parameters but did not change the behavior. Any thought on what we should do there regarding `block_if_invalid` ?
- Tests ?

TBD once points above are resolved

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/include/nav2_collision_monitor/collision_detector_node.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/collision_detector_node.hpp
@@ -114,6 +114,7 @@ protected:
    * @param source_timeout Maximum time interval in which data is considered valid
    * @param base_shift_correction Whether to correct source data towards to base frame movement,
    * considering the difference between current time and latest source time
+   * @param block_if_invalid Whether to stop the robot if a source is invalid
    * @return True if all sources were configured successfully or false in failure case
    */
   bool configureSources(
@@ -121,7 +122,8 @@ protected:
     const std::string & odom_frame_id,
     const tf2::Duration & transform_tolerance,
     const rclcpp::Duration & source_timeout,
-    const bool base_shift_correction);
+    const bool base_shift_correction,
+    const bool block_if_invalid);
 
   /**
    * @brief Main processing routine
--- a/nav2_collision_monitor/include/nav2_collision_monitor/collision_detector_node.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/collision_detector_node.hpp
@@ -114,7 +114,6 @@ protected:
    * @param source_timeout Maximum time interval in which data is considered valid
    * @param base_shift_correction Whether to correct source data towards to base frame movement,
    * considering the difference between current time and latest source time
-   * @param block_if_invalid Whether to stop the robot if a source is invalid
    * @return True if all sources were configured successfully or false in failure case
    */
   bool configureSources(
@@ -122,8 +121,7 @@ protected:
     const std::string & odom_frame_id,
     const tf2::Duration & transform_tolerance,
     const rclcpp::Duration & source_timeout,
-    const bool base_shift_correction,
-    const bool block_if_invalid);
+    const bool base_shift_correction);
 
   /**
    * @brief Main processing routine
--- a/nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
@@ -133,7 +133,6 @@ protected:
    * @param source_timeout Maximum time interval in which data is considered valid
    * @param base_shift_correction Whether to correct source data towards to base frame movement,
    * considering the difference between current time and latest source time
-   * @param block_if_invalid Whether to stop the robot if a source is invalid
    * @return True if all sources were configured successfully or false in failure case
    */
   bool configureSources(
@@ -141,8 +140,7 @@ protected:
     const std::string & odom_frame_id,
     const tf2::Duration & transform_tolerance,
     const rclcpp::Duration & source_timeout,
-    const bool base_shift_correction,
-    const bool block_if_invalid);
+    const bool base_shift_correction);
 
   /**
    * @brief Main processing routine
--- a/nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
@@ -133,6 +133,7 @@ protected:
    * @param source_timeout Maximum time interval in which data is considered valid
    * @param base_shift_correction Whether to correct source data towards to base frame movement,
    * considering the difference between current time and latest source time
+   * @param block_if_invalid Whether to stop the robot if a source is invalid
    * @return True if all sources were configured successfully or false in failure case
    */
   bool configureSources(
@@ -140,7 +141,8 @@ protected:
     const std::string & odom_frame_id,
     const tf2::Duration & transform_tolerance,
     const rclcpp::Duration & source_timeout,
-    const bool base_shift_correction);
+    const bool base_shift_correction,
+    const bool block_if_invalid);
 
   /**
    * @brief Main processing routine
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -322,7 +322,9 @@ void CollisionDetector::process()
       if (!source->getData(curr_time, collision_points)) {
         RCLCPP_WARN(
           get_logger(),
-          "Invalid source %s detected",
+          "Invalid source %s detected."
+          " Either due to data not published yet, or to lack of new data received within the"
+          " sensor timeout, or if impossible to transform data to base frame",
           source->getSourceName().c_str());
       }
     }
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -263,8 +263,7 @@ bool CollisionDetector::configureSources(
         node, source_name + ".source_timeout",
         rclcpp::ParameterValue(source_timeout.seconds()));    // node source_timeout by default
       const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
-        get_parameter(
-          source_name + ".source_timeout").as_double());
+        get_parameter(source_name + ".source_timeout").as_double());
 
       if (source_type == "scan") {
         std::shared_ptr<Scan> s = std::make_shared<Scan>(
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -356,9 +356,6 @@ void CollisionDetector::process()
     collision_points_marker_pub_->publish(std::move(marker_array));
   }
 
-  std::unique_ptr<nav2_msgs::msg::CollisionDetectorState> state_msg =
-    std::make_unique<nav2_msgs::msg::CollisionDetectorState>();
-
   for (std::shared_ptr<Polygon> polygon : polygons_) {
     if (!polygon->getEnabled()) {
       continue;
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -259,16 +259,10 @@ bool CollisionDetector::configureSources(
         rclcpp::ParameterValue("scan"));  // Laser scanner by default
       const std::string source_type = get_parameter(source_name + ".type").as_string();
 
-      nav2_util::declare_parameter_if_not_declared(
-        node, source_name + ".source_timeout",
-        rclcpp::ParameterValue(source_timeout.seconds()));    // node source_timeout by default
-      const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
-        get_parameter(source_name + ".source_timeout").as_double());
-
       if (source_type == "scan") {
         std::shared_ptr<Scan> s = std::make_shared<Scan>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, sensor_specific_source_timeout, base_shift_correction);
+          transform_tolerance, source_timeout, base_shift_correction);
 
         s->configure();
 
@@ -276,7 +270,7 @@ bool CollisionDetector::configureSources(
       } else if (source_type == "pointcloud") {
         std::shared_ptr<PointCloud> p = std::make_shared<PointCloud>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, sensor_specific_source_timeout, base_shift_correction);
+          transform_tolerance, source_timeout, base_shift_correction);
 
         p->configure();
 
@@ -284,7 +278,7 @@ bool CollisionDetector::configureSources(
       } else if (source_type == "range") {
         std::shared_ptr<Range> r = std::make_shared<Range>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, sensor_specific_source_timeout, base_shift_correction);
+          transform_tolerance, source_timeout, base_shift_correction);
 
         r->configure();
 
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -313,7 +313,9 @@ void CollisionDetector::process()
   // Fill collision_points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
     if (source->getEnabled()) {
-      if (!source->getData(curr_time, collision_points)) {
+      if (!source->getData(curr_time, collision_points) &&
+        source->getSourceTimeout().seconds() != 0.0)
+      {
         RCLCPP_WARN(
           get_logger(),
           "Invalid source %s detected."
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -167,10 +167,6 @@ bool CollisionDetector::getParameters()
     node, "base_shift_correction", rclcpp::ParameterValue(true));
   const bool base_shift_correction =
     get_parameter("base_shift_correction").as_bool();
-  nav2_util::declare_parameter_if_not_declared(
-    node, "block_if_invalid", rclcpp::ParameterValue(false));
-  const bool block_if_invalid =
-    get_parameter("block_if_invalid").as_bool();
 
   if (!configurePolygons(base_frame_id, transform_tolerance)) {
     return false;
@@ -178,7 +174,7 @@ bool CollisionDetector::getParameters()
 
   if (!configureSources(
       base_frame_id, odom_frame_id, transform_tolerance, source_timeout,
-      base_shift_correction, block_if_invalid))
+      base_shift_correction))
   {
     return false;
   }
@@ -248,8 +244,7 @@ bool CollisionDetector::configureSources(
   const std::string & odom_frame_id,
   const tf2::Duration & transform_tolerance,
   const rclcpp::Duration & source_timeout,
-  const bool base_shift_correction,
-  const bool block_if_invalid)
+  const bool base_shift_correction)
 {
   try {
     auto node = shared_from_this();
@@ -271,17 +266,10 @@ bool CollisionDetector::configureSources(
         get_parameter(
           source_name + ".source_timeout").as_double());
 
-      nav2_util::declare_parameter_if_not_declared(
-        node, source_name + ".block_if_invalid",
-        rclcpp::ParameterValue(block_if_invalid));    // node block_if_invalid by default
-      const bool sensor_specific_block_if_invalid =
-        get_parameter(source_name + ".block_if_invalid").as_double();
-
       if (source_type == "scan") {
         std::shared_ptr<Scan> s = std::make_shared<Scan>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, sensor_specific_source_timeout, base_shift_correction,
-          sensor_specific_block_if_invalid);
+          transform_tolerance, sensor_specific_source_timeout, base_shift_correction);
 
         s->configure();
 
@@ -289,8 +277,7 @@ bool CollisionDetector::configureSources(
       } else if (source_type == "pointcloud") {
         std::shared_ptr<PointCloud> p = std::make_shared<PointCloud>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, sensor_specific_source_timeout, base_shift_correction,
-          sensor_specific_block_if_invalid);
+          transform_tolerance, sensor_specific_source_timeout, base_shift_correction);
 
         p->configure();
 
@@ -298,8 +285,7 @@ bool CollisionDetector::configureSources(
       } else if (source_type == "range") {
         std::shared_ptr<Range> r = std::make_shared<Range>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, sensor_specific_source_timeout, base_shift_correction,
-          sensor_specific_block_if_invalid);
+          transform_tolerance, sensor_specific_source_timeout, base_shift_correction);
 
         r->configure();
 
--- a/nav2_collision_monitor/src/collision_detector_node.cpp
+++ b/nav2_collision_monitor/src/collision_detector_node.cpp
@@ -167,6 +167,10 @@ bool CollisionDetector::getParameters()
     node, "base_shift_correction", rclcpp::ParameterValue(true));
   const bool base_shift_correction =
     get_parameter("base_shift_correction").as_bool();
+  nav2_util::declare_parameter_if_not_declared(
+    node, "block_if_invalid", rclcpp::ParameterValue(false));
+  const bool block_if_invalid =
+    get_parameter("block_if_invalid").as_bool();
 
   if (!configurePolygons(base_frame_id, transform_tolerance)) {
     return false;
@@ -174,7 +178,7 @@ bool CollisionDetector::getParameters()
 
   if (!configureSources(
       base_frame_id, odom_frame_id, transform_tolerance, source_timeout,
-      base_shift_correction))
+      base_shift_correction, block_if_invalid))
   {
     return false;
   }
@@ -244,7 +248,8 @@ bool CollisionDetector::configureSources(
   const std::string & odom_frame_id,
   const tf2::Duration & transform_tolerance,
   const rclcpp::Duration & source_timeout,
-  const bool base_shift_correction)
+  const bool base_shift_correction,
+  const bool block_if_invalid)
 {
   try {
     auto node = shared_from_this();
@@ -259,10 +264,24 @@ bool CollisionDetector::configureSources(
         rclcpp::ParameterValue("scan"));  // Laser scanner by default
       const std::string source_type = get_parameter(source_name + ".type").as_string();
 
+      nav2_util::declare_parameter_if_not_declared(
+        node, source_name + ".source_timeout",
+        rclcpp::ParameterValue(source_timeout.seconds()));    // node source_timeout by default
+      const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
+        get_parameter(
+          source_name + ".source_timeout").as_double());
+
+      nav2_util::declare_parameter_if_not_declared(
+        node, source_name + ".block_if_invalid",
+        rclcpp::ParameterValue(block_if_invalid));    // node block_if_invalid by default
+      const bool sensor_specific_block_if_invalid =
+        get_parameter(source_name + ".block_if_invalid").as_double();
+
       if (source_type == "scan") {
         std::shared_ptr<Scan> s = std::make_shared<Scan>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, source_timeout, base_shift_correction);
+          transform_tolerance, sensor_specific_source_timeout, base_shift_correction,
+          sensor_specific_block_if_invalid);
 
         s->configure();
 
@@ -270,7 +289,8 @@ bool CollisionDetector::configureSources(
       } else if (source_type == "pointcloud") {
         std::shared_ptr<PointCloud> p = std::make_shared<PointCloud>(
           node, source_name, tf_buffer_, base_frame_id, odom_frame_id,
-          transform_tolerance, source_timeout, base_shift_correction);
+          transform_tolerance, sensor_specific_source_timeout, base_shift_correction,
+          sensor_specific_block_if_invalid);
 
         p->configure();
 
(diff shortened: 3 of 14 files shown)


SteveMacenski (changes requested):
Ok, so on reviewing this, something I don't quite understand is why introduce `block_if_invalid`? If its too old / not updating, that's a pretty critical issue. What's the use-case where we wouldn't have this set to true? 

If there isn't one and just trying to be general purpose, I think we can remove that parameter and just outright return `false` when in an error state. Then we keep the getData change to return bool for logging the warning / stopping the system. 

I think also the collision detector node should use that `getData` bool to log the warning too. I'm not sure - should we somehow adjust the publication to note that to an external listener @tonynajjar? 

Some tests are failing, usual bits on documentation updates! Great addition! Can't believe myself and Alexey missed that

doisyg:
> Ok, so on reviewing this, something I don't quite understand is why introduce block_if_invalid? If its too old / not updating, that's a pretty critical issue. What's the use-case where we wouldn't have this set to true?

To not change the current behavior. Maybe there are people using the CM on sensors with flaky rate. Or maybe with a dynamic array of sensors that can be enabled/disabled depending on the robot mission. But I agree that blocking on invalid should _ideally_ be the default behavior if that's acceptable to change.
Hum, what about getting rid of `block_if_invalid` but allow to ignore an invalid source if `source_timeout` is set inferior 0.0 ?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -374,7 +396,16 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
...
+      if (!source->getData(curr_time, collision_points)) {
+        RCLCPP_WARN(get_logger(), "Invalid blocking source detected, stopping the robot");
+        Velocity stop_vel;
+        stop_vel.tw = 0.0;
+        stop_vel.x = 0.0;
+        stop_vel.y = 0.0;
+        Action robot_action{STOP, stop_vel, ""};
+        publishVelocity(robot_action);
+        return;
+      }
```
AlexeyMerzlyakov, line 408: The return will break the `process()` logic: in the tail of this routine there are some things which we can't miss.
If we just return from this, the notify action state logic won't be started, and CM won't do the change notification. So, it is better to update `CollisionMonitor::notifyActionState()` with ability to pass `action_polygon` as `nullptr` and report abnormal situation for that cases.

Additionally, we need to save action with 0-velocity to `robot_action_prev_` to have notifications to work properly; and still publish polygons, since they might be in different frames moving while robot is being stopped.

I think, here is better to set main `robot_action` local variable and go to `process()` routine's tail, calling `notifyActionState()` with null `action_polygon` that will tell this routine to report abnormal behavior.

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -318,26 +324,42 @@ bool CollisionMonitor::configureSources(
...
+      const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
+        get_parameter(
+          source_name + ".source_timeout").as_double());
+
+      nav2_util::declare_parameter_if_not_declared(
+        node, source_name + ".block_if_invalid",
+        rclcpp::ParameterValue(block_if_invalid));    // node block_if_invalid by default
+      const bool sensor_specific_block_if_invalid =
+        get_parameter(source_name + ".block_if_invalid").as_bool();
+
```
AlexeyMerzlyakov, line 339: Please move each sources' timeout and `ignore_invalid` (or whatever you call it) logic inside `Source::getCommonParameters()` since both parameters belong to their own data source

nav2_collision_monitor/src/collision_detector_node.cpp
AlexeyMerzlyakov, line 1: We need to think, how to report the robot needs to stop, when the sources were outdated for Collision Detector. In the initial message, there is a detected polygons string array, but there is not ability what to do. So, I am thinking about publishing `CollisionDetectorState.msg` with empty `polygons[]` array, that will give the developer an information that something (not polygon) is causing robot to report the "collision". Or add one more bool variable into this message, as an alternative for it.

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -253,26 +258,42 @@ bool CollisionDetector::configureSources(
...
+      const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
+        get_parameter(
+          source_name + ".source_timeout").as_double());
+
+      nav2_util::declare_parameter_if_not_declared(
+        node, source_name + ".block_if_invalid",
+        rclcpp::ParameterValue(block_if_invalid));    // node block_if_invalid by default
+      const bool sensor_specific_block_if_invalid =
+        get_parameter(source_name + ".block_if_invalid").as_double();
+
```
AlexeyMerzlyakov, line 273: The same here: please move this logic inside `Source` class

nav2_collision_monitor/test/sources_test.cpp
AlexeyMerzlyakov, line 1: Regarding testcase idea 
* sources_test.cpp
 You could publish implicitly outdated sources and check that `getData()` will return `true` / `false` depending on `ignore_invalid` state
* collision_monitor_node_test.cpp
 The idea here is to again, publish implicitly outdated source and read the out `cmd_vel` wheels message: it should be 0 or non-0, depending on `ignore_invalid` state. Additionally, `CollisionMonitorState.msg` state should be received and checked.
* collision_detector_node_test.cpp
  The similar logic as for collision_monitor_node_test.cpp, but using `CollisionDetectorState.msg` message

AlexeyMerzlyakov (commented):
Thank you for pointing this situation. Yes, I'd agree that we better to stop robot in case of broken one of the source: this is serious situation (unless developer intentionally tells that this is unreliable source, by setting `block_if_invalid` or `ignore_invalid` option).

Please check my comments below:

AlexeyMerzlyakov:
> Hum, what about getting rid of `block_if_invalid` but allow to ignore an invalid source if `source_timeout` is set inferior 0.0 ?

I am not sure, that negative timeout values won't cause a confusion. As from me, it is better to use `ignore_invalid` indicator per each source, since it will be more clear for end-developer.

SteveMacenski:
>   To not change the current behavior. Maybe there are people using the CM on sensors with flaky rate. Or maybe with a dynamic array of sensors that can be enabled/disabled depending on the robot mission. But I agree that blocking on invalid should ideally be the default behavior if that's acceptable to change.

I think that's acceptable to change. If they have flaky sensors, they can set the timeout to something outrageously high (or below). Disable/enable can be done through this API with Tony's recent work, no? 

> Hum, what about getting rid of block_if_invalid but allow to ignore an invalid source if source_timeout is set inferior 0.0 ?

Sure, or `-1` or something clearly invalid. I  think `-1` is better

nav2_collision_monitor/src/collision_monitor_node.cpp
SteveMacenski: I think he's also made this so that there can be a global version as well, not just per-sensor if they want to specify in the parent namespace for all

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -374,7 +396,16 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
   // Fill collision_points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
     if (source->getEnabled()) {
-      source->getData(curr_time, collision_points);
+      if (!source->getData(curr_time, collision_points)) {
+        RCLCPP_WARN(get_logger(), "Invalid blocking source detected, stopping the robot");
```
tonynajjar, line 400: Would be useful to know which source is faulty

tonynajjar:
> I think also the collision detector node should use that getData bool to log the warning too. I'm not sure - should we somehow adjust the publication to note that to an external listener @tonynajjar?

Logging is an obvious improvement yes. I'm also not too sure about adjusting the publication; it wouldn't hurt but I can't think of satisfying semantics. Another alternative to modifying the msg is to publish this information as diagnostics; seems like a suitable place for something similar to a failing heartbeat.

nav2_collision_monitor/src/collision_monitor_node.cpp
doisyg: Indeed

nav2_collision_monitor/src/collision_detector_node.cpp
doisyg: Same as above. Allow both per global and per sensor config

nav2_collision_monitor/src/collision_monitor_node.cpp
doisyg: See last commit

doisyg:
> Sure, or `-1` or something clearly invalid. I think `-1` is better

Actually going for `0.0` as the `source_timeout` is stored as a `Duration` and can't be negative

doisyg:
Okay, I just pushed some changes:
- Default behavior now to blocking if a source is invalid
- Got rid of the `block_if_invalid` parameter and allow to have a non blocking invalid source when the`source_timeout` parameter is equal to `0.0` (internally `source_timeout_.seconds() == 0.0`)
- Also allow valid state (getData true) if a source did not publish yet AND `source_timeout: 0.0`, otherwise block. 
- Added source name in the log (with new `std::string Source::getSourceName()` method)

If that looks fine to you, I will move forward with the next steps:
1. Not breaking the `process()` logic and proper feedback
2. Collision Detector
3. Tests

nav2_collision_monitor/include/nav2_collision_monitor/collision_detector_node.hpp
```diff
@@ -113,14 +113,16 @@ class CollisionDetector : public nav2_util::LifecycleNode
...
    * @return True if all sources were configured successfully or false in failure case
    */
   bool configureSources(
     const std::string & base_frame_id,
     const std::string & odom_frame_id,
     const tf2::Duration & transform_tolerance,
     const rclcpp::Duration & source_timeout,
-    const bool base_shift_correction);
+    const bool base_shift_correction,
+    const bool block_if_invalid);
```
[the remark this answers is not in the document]
doisyg: Changed see below

doisyg:
If we consider the 0.0 equality check dirty, I will change `source_timeout` type to `double` and check for negativity

nav2_collision_monitor/src/source.cpp
```diff
@@ -75,6 +75,10 @@ bool Source::sourceValid(
   const rclcpp::Time & source_time,
   const rclcpp::Time & curr_time) const
 {
+  // Ignore timestamp check if source_timeout == 0.0
+  if (source_timeout_.seconds() == 0.0) {
+    return true;
+  }
```
AlexeyMerzlyakov, line 81: If source is outdated too much, `spourceValid()` will return true and we still try to interpolate sources on current transform time. I think, it is more correct to interpolate the sources if time shift is between 0 ... and some `source_timeout`, and ignore outdated sources, if time shift is lager than `source_timeout` (if developer want this). Otherwise, we could get some unpredictable failures (e.g. what will be, if some data source is not available within 5 min?)

So, I consider to have `ignore_invalid` option, or using negative timeouts for ignored sources.

AlexeyMerzlyakov:
Regarding https://github.com/ros-planning/navigation2/pull/3880#discussion_r1360488856 and https://github.com/ros-planning/navigation2/pull/3880#discussion_r1360501611: if we are changed the code to message the failed source, I think it would be fair to add the failed source name in `CollisionMonitorState.msg` and `CollisionMonitorDetector.msg`. @SteveMacenski, what do you think about it?

nav2_collision_monitor/src/source.cpp
doisyg: Changed the logic in last push

doisyg:
I added the source_names feedback in the last commit as an experiment so we can discuss (but there is probably an overhead of doing this). 
For the collision monitor, new field `string[] source_names` in `CollisionMonitorState.msg ` containing:
- if all sources are valid: list of first sources that triggered the matching `polygon_name`. Possibly multiple sources as we trigger a polygon with a min number of points that can be more than 1.
- Otherwise, if a source is invalid: the list of invalid source, and polygon_name will be left empty

For the collision detector, new field `string[] source_names` in `CollisionDetectedState.msg ` containing:
- if all sources are valid: list of list of sources (combined in a string to avoid a multi array field).
- Otherwise, if a source is invalid: the list of invalid sources, and `polygons` + `detections` will be left empty

nav2_collision_monitor/src/collision_detector_node.cpp
doisyg: See last commit

nav2_collision_monitor/src/collision_monitor_node.cpp
doisyg: See last commit

nav2_collision_monitor/include/nav2_collision_monitor/collision_monitor_node.hpp
```diff
@@ -154,6 +154,7 @@ class CollisionMonitor : public nav2_util::LifecycleNode
    * @param robot_action Output processed robot action
    * @return True if returned action is caused by current polygon, otherwise false
    */
+  void extracted(std::vector<std::string> & source_names) const;
```
SteveMacenski, line 157: Unused?

nav2_collision_monitor/src/circle.cpp
```diff
@@ -58,14 +58,20 @@ void Circle::getPolygon(std::vector<Point> & poly) const
...
   int num = 0;
-  for (Point point : points) {
+  std::unordered_set<std::string> unordered_source_names;
+  for (const Point & point : points) {
     if (point.x * point.x + point.y * point.y < radius_squared_) {
       num++;
+      unordered_source_names.insert(point.source_name);
     }
   }
+  source_names.assign(unordered_source_names.begin(), unordered_source_names.end());
```
SteveMacenski, line 73: Unless the source names stuff is needed for the watchdog, lets remove it. This PR is starting to get a little too complicated to review with this change in here if its a new feature.

Also, this seems a little convoluted. If we were to add this, I'd think changing points to being a vector of sources, whereas each source is a vector of points + a source string name. That way we iterate through a vector of a vector of points, where each nested vector as a single source name attached to it so we don't have these massively long source_name vectors of repeating values

doisyg:
Removed source feedback and rebased on latest.
Source feedback should probably be added only after https://github.com/ros-planning/navigation2/pull/3885 comes in

doisyg:
If that looks good to you, I quickly add a test and update documentation

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -259,26 +259,33 @@ bool CollisionDetector::configureSources(
         rclcpp::ParameterValue("scan"));  // Laser scanner by default
       const std::string source_type = get_parameter(source_name + ".type").as_string();
 
+      nav2_util::declare_parameter_if_not_declared(
+        node, source_name + ".source_timeout",
+        rclcpp::ParameterValue(source_timeout.seconds()));    // node source_timeout by default
+      const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
+        get_parameter(
+          source_name + ".source_timeout").as_double());
```
SteveMacenski, line 267: `source_name + ".source_timeout").as_double());` can be on the same line as `get_parameter`, looks like an autoformatting oddity

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -325,26 +325,33 @@ bool CollisionMonitor::configureSources(
         rclcpp::ParameterValue("scan"));  // Laser scanner by default
       const std::string source_type = get_parameter(source_name + ".type").as_string();
 
+      nav2_util::declare_parameter_if_not_declared(
+        node, source_name + ".source_timeout",
+        rclcpp::ParameterValue(source_timeout.seconds()));    // node source_timeout by default
+      const rclcpp::Duration sensor_specific_source_timeout = rclcpp::Duration::from_seconds(
+        get_parameter(
```
SteveMacenski, line 332: Ditto

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -307,10 +314,18 @@ void CollisionDetector::process()
...
+    std::make_unique<nav2_msgs::msg::CollisionDetectorState>();
+
   // Fill collision_points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
     if (source->getEnabled()) {
-      source->getData(curr_time, collision_points);
+      if (!source->getData(curr_time, collision_points)) {
+        RCLCPP_WARN(
+          get_logger(),
+          "Invalid source %s detected",
```
SteveMacenski, line 326: Here and in the monitor: I want this log to be more descriptive.

"Source (%s) has timed out due to lack of new data received within the sensor timeout."

Or to that effect

SteveMacenski (changes requested):
Superficial stuff, but then I'm happy with it. @AlexeyMerzlyakov any other blockers?

nav2_collision_monitor/src/collision_detector_node.cpp
SteveMacenski: ^ I wouldn't have bothered if I didn't ask for the log update below

nav2_collision_monitor/src/pointcloud.cpp
```diff
@@ -65,17 +65,24 @@ void PointCloud::configure()
...
+    }
+    return false;
   }
   if (!sourceValid(data_->header.stamp, curr_time)) {
-    return;
+    // Don't block if source_timeout == 0.0
+    if (source_timeout_.seconds() == 0.0) {
+      return true;
+    }
+    return false;
```
AlexeyMerzlyakov, line 85: It looks like we have to face the same as reported in https://github.com/ros-planning/navigation2/pull/3880#discussion_r1361601175 issue.

Let's check the case when `source_timout_` is equal to `0.0`: this is the case, when we should ignore outdated source. However, how could we determine that the source is outdated?
In other words, during the execution `sourceValid()` routine contains the following comparison inside:
```
 const rclcpp::Duration dt = curr_time - source_time;
  if (dt > source_timeout_) {
    // Fail-case
  }
```
which always will return `false`. Thus, if we set 0-value of `source_timeout` for specific source, this source will be always ignored, independently on which timeout we want to bring on, after the ignoring became to be.

nav2_collision_monitor/src/collision_monitor_node.cpp
AlexeyMerzlyakov: I still inclined to move this logic inside `Source::getCommonParameters()`: by its design all parameters belonging to any specific source or polygons should be declared/get in its routines. We could construct the source with common for all sources timeout, and then later on `configure()` stage (when getCommonParameters is being called), re-assign the `source_timeout_` class' value for specific for its source value. I think, this will be more clear to not making the part of code outside its handling class.

nav2_collision_monitor/test/collision_monitor_node_test.cpp
```diff
@@ -908,6 +908,7 @@ TEST_F(Tester, testCrossOver)
   // 1. Obstacle is not in the slowdown zone, but less than TIME_BEFORE_COLLISION (ahead in 1.5 m).
   // Robot should approach the obstacle.
   publishPointCloud(2.5, curr_time);
+  publishRange(2.5, curr_time);
```
AlexeyMerzlyakov, line 911: This is very strange - what was happened here: why the test became to fail? Initially, it was developed to pointcloud to be enough to trigger the approach model. What is the reason, why pointcoud was not enough to do that?

nav2_collision_monitor/src/collision_detector_node.cpp
doisyg: Fixed

nav2_collision_monitor/src/collision_detector_node.cpp
doisyg: Fixed

nav2_collision_monitor/src/collision_monitor_node.cpp
doisyg: fixed

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -307,10 +314,18 @@ void CollisionDetector::process()
...
   std::vector<Point> collision_points;
 
+  std::unique_ptr<nav2_msgs::msg::CollisionDetectorState> state_msg =
+    std::make_unique<nav2_msgs::msg::CollisionDetectorState>();
+
   // Fill collision_points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
     if (source->getEnabled()) {
-      source->getData(curr_time, collision_points);
+      if (!source->getData(curr_time, collision_points)) {
```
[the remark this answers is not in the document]
doisyg: I had a much more complicated approach with proper source feedback, but per @SteveMacenski, I agree it is best to leave it to another PR.
Effort archived here: https://github.com/doisyg/navigation2/tree/with_source_feedback

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -378,10 +385,23 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
...
+  // Polygon causing robot action (if any)
+  std::shared_ptr<Polygon> action_polygon;
+
   // Fill collision_points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
     if (source->getEnabled()) {
-      source->getData(curr_time, collision_points);
+      if (!source->getData(curr_time, collision_points)) {
+        action_polygon = nullptr;
+        robot_action.polygon_name = "invalid source";
```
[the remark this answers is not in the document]
doisyg: Same as above.
For now, without proper source tracing/feedback, that would not make sense as we are stopping at the detection of the first invalid source, so if we have multiple invalid, it will just be random which one we are reporting.
Alternative (complex) is here: https://github.com/doisyg/navigation2/tree/with_source_feedback
Better for another PR

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -545,10 +560,16 @@ void CollisionMonitor::notifyActionState(
...
 {
   if (robot_action.action_type == STOP) {
-    RCLCPP_INFO(
-      get_logger(),
-      "Robot to stop due to %s polygon",
-      action_polygon->getName().c_str());
+    if (robot_action.polygon_name == "invalid source") {
+      RCLCPP_WARN(
+        get_logger(),
+        "Robot to stop due to invalid source");
```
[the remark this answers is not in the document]
doisyg: Same as above. See https://github.com/doisyg/navigation2/tree/with_source_feedback but better for another PR

nav2_collision_monitor/src/pointcloud.cpp
doisyg: Gosh. I get it know. Nice catch. This is problematic, a refactor is needed. I don't see a way to solve it without re-introducing the additional bool parameter `block_if_invalid` (or better name). Thoughts @SteveMacenski ?

nav2_collision_monitor/test/collision_monitor_node_test.cpp
doisyg: It is actually a proof that this PR kind of works ,)
New default behavior is to stop if a source is invalid. In this test, a range source was declared but was never published. So as we have a default source timeout of 5s, the test was failing. (I think)

nav2_collision_monitor/src/pointcloud.cpp
SteveMacenski: @doisyg why not just move this check inside of `sourceValid`? 

```

if (source_timeout_.seconds() != 0.0 && dt > source_timeout_) {
```

doisyg: That will push the problem down to the next part of `getData` and we will have a tf error at some point :
https://github.com/doisyg/navigation2/blob/e54293f9b327174de619519317b6bb5bc0d2e282/nav2_collision_monitor/src/pointcloud.cpp#L89-L111

Though if we also don't block on tf error and just skip the data on `source_timeout == 0.0`, we could add a check higher here (which will require a new source::getSourceTimeout method) and we should be fine: https://github.com/doisyg/navigation2/blob/e54293f9b327174de619519317b6bb5bc0d2e282/nav2_collision_monitor/src/collision_monitor_node.cpp#L389)

```
  // Fill collision_points array from different data sources
  for (std::shared_ptr<Source> source : sources_) {
    if (source->getEnabled()) {
      if (!source->getData(curr_time, collision_points) && source->getSourceTimeout() != 0.0) {
        action_polygon = nullptr;
        robot_action.polygon_name = "invalid source";
        robot_action.action_type = STOP;
        robot_action.req_vel.x = 0.0;
        robot_action.req_vel.y = 0.0;
        break;
      }
    }
  }

```
Pushing this now

SteveMacenski: > we will have a tf error at some point 

So thinking aloud: 
- If we check if source is valid `sourceValid` with the check on timeout, then we either return `false` due to stale data with a timeout or...
- We continue on allowing stale data without a timeout. Then:
- We try to TF transform it. As long as that data isn't insanely stale, that should be fine as long as its still in the TF buffer
- If its not in the TF buffer, then the `return false` is warranted since its literally impossible to use 

That seems like a fine workflow. If we set no timeout, then it should work fine for as long as the data is in any way possibly actionable. 

So then going to the `node` code block you added, that would make sense to not `STOP` in that case. We don't want to consider the data, but we also don't want to stop either. That seems sensible.

SteveMacenski:
@AlexeyMerzlyakov please re-review and close your previous comments if they are outdated / fixed. Its hard to review this when its unclear how many of your items are still actionable / blocking or not.

nav2_collision_monitor/src/source.cpp
```diff
@@ -69,6 +69,12 @@ void Source::getCommonParameters(std::string & source_topic)
   nav2_util::declare_parameter_if_not_declared(
     node, source_name_ + ".enabled", rclcpp::ParameterValue(true));
   enabled_ = node->get_parameter(source_name_ + ".enabled").as_bool();
+
+  nav2_util::declare_parameter_if_not_declared(
+    node, source_name_ + ".source_timeout",
+    rclcpp::ParameterValue(source_timeout_.seconds()));      // node source_timeout by default
```
SteveMacenski, line 75: which calls into question this -- can you test if this actually is not just full whole seconds?

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -307,10 +307,22 @@ void CollisionDetector::process()
...
 
+  std::unique_ptr<nav2_msgs::msg::CollisionDetectorState> state_msg =
+    std::make_unique<nav2_msgs::msg::CollisionDetectorState>();
+
   // Fill collision_points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
     if (source->getEnabled()) {
-      source->getData(curr_time, collision_points);
+      if (!source->getData(curr_time, collision_points) &&
+        source->getSourceTimeout().seconds() != 0.0)
```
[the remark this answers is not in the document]
doisyg: It is a `double` behind, it will be fine (apart if we want a very accurate and not small `source_timeout` like 10000.0000000001). `nanoseconds() `returns a `rcl_duration_value_t`, more messy

nav2_collision_monitor/src/source.cpp
doisyg: I swear it is not. This can be confusing as message header `stamp` is split between full seconds and full nanoseconds int: https://docs.ros2.org/latest/api/builtin_interfaces/msg/Time.html
But that is not the case for `rclcpp::Duration` and `rclcpp::Time` `seconds()` methods

nav2_collision_monitor/src/source.cpp
```diff
@@ -78,7 +84,7 @@ bool Source::sourceValid(
   // Source is considered as not valid, if latest received data timestamp is earlier
   // than current time by source_timeout_ interval
   const rclcpp::Duration dt = curr_time - source_time;
-  if (dt > source_timeout_) {
+  if (source_timeout_.seconds() != 0.0 && dt > source_timeout_) {
```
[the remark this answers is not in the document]
doisyg: Same

nav2_collision_monitor/src/scan.cpp
```diff
@@ -64,17 +64,21 @@ void Scan::configure()
...
+bool Scan::getData(
   const rclcpp::Time & curr_time,
   std::vector<Point> & data) const
 {
   // Ignore data from the source if it is not being published yet or
   // not being published for a long time
   if (data_ == nullptr) {
-    return;
+    // Don't block if source_timeout == 0.0
+    if (source_timeout_.seconds() == 0.0) {
```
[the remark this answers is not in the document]
doisyg: Indeed

nav2_collision_monitor/src/range.cpp
```diff
@@ -65,17 +65,20 @@ void Range::configure()
...
-void Range::getData(
+bool Range::getData(
   const rclcpp::Time & curr_time,
   std::vector<Point> & data) const
 {
   // Ignore data from the source if it is not being published yet or
   // not being published for a long time
   if (data_ == nullptr) {
-    return;
+    if (source_timeout_.seconds() == 0.0) {
```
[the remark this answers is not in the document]
doisyg: Indeedx3

nav2_collision_monitor/src/collision_detector_node.cpp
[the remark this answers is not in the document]
SteveMacenski: @doisyg just checking -- if you've tested this and you know it works, I'll trust you. The `count()` though makes me highly suspicious though, since chrono's count usually returns a whole numbered count from my experience:

```
  return std::chrono::duration<double>(std::chrono::nanoseconds(rcl_duration_.nanoseconds)).count();
```

But the `duration` might do differently. I both love/hate chrono. Very useful, but also the API isn't very consistent at times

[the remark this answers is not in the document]
doisyg: > @doisyg just checking 

As you should, I get confused very often with all these slightly different Time API.

> if you've tested this and you know it works

Confirming that it is tested and working

SteveMacenski:
CI failed with 

> unknown file. C++ exception with description "parameter 'Range.source_timeout' cannot be set because it was not declared" thrown in the test body.

SteveMacenski (approved):
Otherwise, I approve pending @AlexeyMerzlyakov's rereview

doisyg:
Hopefully fixed CI.
Doc PR: https://github.com/ros-planning/navigation.ros.org/pull/483

SteveMacenski:
Looks good to me. Any other open topics from @AlexeyMerzlyakov, @doisyg? I know he's not fully tasked on ROS anymore, so if you think you resolved all his concerns, we can merge this to save him the time. If he has issues later, we can address them in the follow up PR doing the logging stuff

doisyg:
All good to me, I am happy to support and iterate fast if this PR causes any unforeseen issue

nav2_collision_monitor/test/collision_monitor_node_test.cpp
AlexeyMerzlyakov: Ah, yes this takes place, as previously in this testcase range source was added: `addSource(RANGE_NAME, RANGE)`

PR #3885, opened 2023-10-16
CM/CD: Add polygon source

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Custom gazebo simulation |

---

We would like to add a source to the collision monitor and that is an array of arbitrary PolygonStamped fed from a topic. Our specific use case is that we would like to prevent collisions with some tracked objects. 

Note: Adding these tracked objects to the global/local costmap was tried, however this was not enough as the controller makes a "best effort" to avoid the collisions; we need more decisive logic to stop the robot and therefore need the collision monitor

* Add Polygon as a source
* Create new msg type PolygonArray to feed in polygons to the Polygon source

---

#### For Maintainers: 


nav2_util/src/robot_utils.cpp
```diff
@@ -143,6 +143,66 @@ bool getTransform(
   return true;
 }
 
+bool getTransform(
+  const std::string & source_frame_id,
+  const std::string & target_frame_id,
+  const tf2::Duration & transform_tolerance,
+  const std::shared_ptr<tf2_ros::Buffer> tf_buffer,
+  geometry_msgs::msg::TransformStamped & transform_stamped)
```
tonynajjar, line 151: I needed the output as geometry_msgs::msg::TransformStamped to keep the stamp. We can think of a less verbose solution maybe later

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,166 @@
...
+  // get the oldest time stamp from the polygon array
+  rclcpp::Time oldest_stamp = rclcpp::Time(data_->polygons[0].header.stamp);
+  for (const auto& polygon : data_->polygons) {
+    if (rclcpp::Time(polygon.header.stamp) < oldest_stamp) {
+      oldest_stamp = rclcpp::Time(polygon.header.stamp);
+    }
+  }
+  if (!sourceValid(oldest_stamp, curr_time)) {
+    return;
+  }
```
tonynajjar, line 86: could maybe be refined to not reflect the whole message

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  for (const auto& polygon : data_->polygons) {
+    geometry_msgs::msg::PolygonStamped poly_out;
+    tf2::doTransform(polygon, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+
+}
+void PolygonSource::convertPolygonStampedToPoints(const geometry_msgs::msg::PolygonStamped& polygon, std::vector<Point>& data) const
```
tonynajjar, line 120: result:

image

tonynajjar:
@SteveMacenski @AlexeyMerzlyakov that's a first draft, a lot of small improvements can still be made but wanted your opinion on the idea before spending more time refining it. Thanks

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -342,6 +342,13 @@ bool CollisionMonitor::configureSources(
         r->configure();
 
         sources_.push_back(r);
+      } else if (source_type == "polygon_source") {
```
SteveMacenski, line 345: Why not just `polygon`? (param + file name + class name)

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  for (const auto & polygon : data_->polygons) {
+    geometry_msgs::msg::PolygonStamped poly_out;
+    tf2::doTransform(polygon, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+
+}
+void PolygonSource::convertPolygonStampedToPoints(
+  const geometry_msgs::msg::PolygonStamped & polygon, std::vector<Point> & data) const
```
SteveMacenski, line 121: We do have a `LineIterator` that you may be able to use instead to make this easer. You can set the start/end coordinates and march along it by `increments`. We use it mostly in the `map` coordinate space so we march by "1" cell, but you can give it world coordinates and march by another increment if you prefer

nav2_util/src/robot_utils.cpp
```diff
@@ -143,6 +143,66 @@ bool getTransform(
...
+{
+  geometry_msgs::msg::TransformStamped transform;
+
+  if (source_frame_id == target_frame_id) {
+    // We are already in required frame
+    return true;
+  }
+
+  try {
+    // Obtaining the transform to get data from source to target frame
```
SteveMacenski, line 161: Could this be called by the nonstamped version above so we reduce duplicate code?

SteveMacenski (commented):
Test coverage?

I'd think for polygon-source types, we would want to make the minimum number of points 1 for this special case

nav2_msgs/msg/PolygonsArray.msg
```diff
@@ -0,0 +1,3 @@
+std_msgs/Header header
```
AlexeyMerzlyakov, line 1: Since each `PolygonStamped` in the array below has its own header, I think, main header is unnecessary

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return;
+  }
+
+  geometry_msgs::msg::TransformStamped tf;
+  if (base_shift_correction_) {
+    // Obtaining the transform to get data from source frame and time where it was received
+    // to the base frame and current time
+    if (
+      !nav2_util::getTransform(
+        data_->header.frame_id, data_->header.stamp,
```
AlexeyMerzlyakov, line 94: Since each polygon in the array might be in its own frame, I believe, we need to get transform per each polygon it the message instead of using main headers' frame_id

nav2_util/src/robot_utils.cpp
```diff
@@ -143,6 +143,66 @@ bool getTransform(
...
+
+bool getTransform(
+  const std::string & source_frame_id,
+  const rclcpp::Time & source_time,
+  const std::string & target_frame_id,
+  const rclcpp::Time & target_time,
+  const std::string & fixed_frame_id,
+  const tf2::Duration & transform_tolerance,
+  const std::shared_ptr<tf2_ros::Buffer> tf_buffer,
+  geometry_msgs::msg::TransformStamped & transform_stamped)
```
AlexeyMerzlyakov, line 184: I'd also like to get rid of duplicate code, as @SteveMacenski mentioned by utilizing `tf2::Transform`. Another alternative could be making this routine as inline and calling it from inside non-stamped `getTransform` version

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,169 @@
...
+void PolygonSource::convertPolygonStampedToPoints(
+  const geometry_msgs::msg::PolygonStamped & polygon, std::vector<Point> & data) const
+{
+  // Calculate the total perimeter of the polygon
+  double perimeter = 0.0;
+  double spacing = 0.1;
+  for (size_t i = 0; i < polygon.polygon.points.size(); ++i) {
+    const auto & currentPoint = polygon.polygon.points[i];
+    const auto & nextPoint = polygon.polygon.points[(i + 1) % polygon.polygon.points.size()];
+    perimeter += sqrt(pow(nextPoint.x - currentPoint.x, 2) + pow(nextPoint.y - currentPoint.y, 2));
```
AlexeyMerzlyakov, line 129: std::hypot can do this

nav2_collision_monitor/src/collision_monitor_node.cpp
tonynajjar: Just made it the same as the filename/class but you're right, doesn't have to be, changing

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: hmm I gave it a quick look but it's not clear to me how to use it in this situation, the increment doesn't seem to be configurable. I removed an obsolete part of it, maybe now it's a bit simpler. Is it a must for you to use the LineIterator here?

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: good point, changed

nav2_util/src/robot_utils.cpp
tonynajjar: I think I found a satisfying solution, I changed the output of `getTransform` from `tf2::Transform` to `tf2::Stamped<tf2::Transform>` so that we can keep the stamp information. The other sources are not affected since `    tf2::Vector3 p_v3_b = tf_transform * p_v3_s;` should work with both stamped and unstamped Transform

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,166 @@
...
+    tf2::doTransform(polygon, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+
+}
+void PolygonSource::convertPolygonStampedToPoints(const geometry_msgs::msg::PolygonStamped& polygon, std::vector<Point>& data) const
+{
+    // Calculate the total perimeter of the polygon
+    double perimeter = 0.0;
+    double spacing = 0.1;
```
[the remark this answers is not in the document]
tonynajjar: done

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: Any opinions on what to consider a valid message?

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
```diff
@@ -0,0 +1,108 @@
...
+   */
+  void dataCallback(nav2_msgs::msg::PolygonsArray::ConstSharedPtr msg);
+
+  // ----- Variables -----
+
+  /// @brief PolygonSource data subscriber
+  rclcpp::Subscription<nav2_msgs::msg::PolygonsArray>::SharedPtr data_sub_;
+
+  /// @brief Latest data obtained
+  nav2_msgs::msg::PolygonsArray::ConstSharedPtr data_;
```
SteveMacenski, line 100: This should use standard messages, not custom ones. You can have it store a vector of data and then clear the vector after its been populated by the monitor

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+  // Ignore data from the source if it is not being published yet or
+  // not published for a long time
+  if (data_ == nullptr || data_->polygons.empty()) {
+    return;
+  }
+  // get the oldest time stamp from the polygon array
+  rclcpp::Time oldest_stamp = rclcpp::Time(data_->polygons[0].header.stamp);
+  for (const auto & polygon : data_->polygons) {
+    if (rclcpp::Time(polygon.header.stamp) < oldest_stamp) {
```
SteveMacenski, line 82: FYI @doisyg for timeout-ing: how should we handle if its multiple data? If any single one is bad or if all are bad?

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,172 @@
...
+void PolygonSource::getData(
+  const rclcpp::Time & curr_time,
+  std::vector<Point> & data) const
+{
+  // Ignore data from the source if it is not being published yet or
+  // not published for a long time
+  if (data_ == nullptr || data_->polygons.empty()) {
+    return;
+  }
+  // get the oldest time stamp from the polygon array
```
SteveMacenski, line 79: Space above this line

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,172 @@
...
+      {
+        return;
+      }
+    }
+    geometry_msgs::msg::PolygonStamped poly_out;
+    geometry_msgs::msg::TransformStamped tf = tf2::toMsg(tf_transform);
+    tf2::doTransform(polygon, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+
```
SteveMacenski, line 121: No space

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+void PolygonSource::convertPolygonStampedToPoints(
```
SteveMacenski, line 123: Space between functions

Overall, linting / styling here is weird. Please fix.

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,172 @@
...
+  const geometry_msgs::msg::PolygonStamped & polygon, std::vector<Point> & data) const
+{
+  // Iterate over the vertices of the polygon
+  for (size_t i = 0; i < polygon.polygon.points.size(); ++i) {
```
SteveMacenski, line 127: While this works, again, why not use the line iterator we have in nav2_utils?

SteveMacenski (changes requested):

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: I'm thinking we just only include the current ones and log a warning about any expired ones that aren't being used -- but not returning false if any are valid.

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
tonynajjar: hmm, I see one challenge, how would you not store duplicates in the vector? By cross-checking every field of the msg::Polygon against the Polygons already in the vector and not adding it if it matches?

SteveMacenski: Will these polygons have headers? Perhaps keep the latest of each header typed? Maybe as a parameter so that others might want to use all

tonynajjar: They will need to have headers to check individually if they are too old to be valid. What do yoh mean with keep it "typed", can you clarify?

SteveMacenski: I mean if you have N polygons, check the header values and keep only the newest of each reference frame, but as a parameterized option

SteveMacenski: What would you propose?

tonynajjar: I stated the only alternatives I have already, either custom message or tracking with std::set + custom comparator as mentioned above. The custom msg seems simpler but I understand wanting to stick to standard messages...

SteveMacenski: OK, lets go with that then. I'm not sure a custom comparator is needed, I believe `msg == msg` is implemented for all ROS message structs.

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,172 @@
...
+    const auto & current_point = polygon.polygon.points[i];
+    const auto & next_point = polygon.polygon.points[(i + 1) % polygon.polygon.points.size()];
+
+    // Calculate the distance between the current and next points
+    double segment_length =
+      std::hypot(next_point.x - current_point.x, next_point.y - current_point.y);
+
+    // Calculate the number of points to sample in the current segment
+    size_t num_points_in_segment =
+      std::max(static_cast<size_t>(segment_length / sampling_distance_), static_cast<size_t>(1));
```
AlexeyMerzlyakov, line 137: Is not just `1` w/o static_cast won't be suitable here?

nav2_collision_monitor/test/sources_test.cpp
```diff
@@ -156,10 +161,47 @@ class TestNode : public nav2_util::LifecycleNode
...
+      std::make_unique<nav2_msgs::msg::PolygonsArray>();
+
+    geometry_msgs::msg::PolygonStamped polygon_msg;
+
+    polygon_msg.header.frame_id = SOURCE_FRAME_ID;
+    polygon_msg.header.stamp = stamp;
+
+    geometry_msgs::msg::Point32 point;
+    point.x = 1.0;
+    point.y = 0.0;
```
AlexeyMerzlyakov, line 179: The question to make sure that everything is syched: the first point seems to asymmetrical towards to the rest 3. Is it the initial intention to make non-bar polygon, or misprint?

nav2_collision_monitor/test/sources_test.cpp
AlexeyMerzlyakov, line 1: Also, the case to cover the working with the `PolygonSource` is needed in `collision_monitor_node_test.cpp` and `collision_detector_node_test.cpp`. It could be new simple testcase, just create the (1,1,-1,-1) stop polygon and add there polygon source with intersection with stop polygon, like depicted below, and make sure that robot will stop (for CM; or report stopping for CD) after polygon source will entered/interlaced the stop polygon:

AlexeyMerzlyakov (commented):
There is almost nothing to add to @SteveMacenski comments from last review on the main logic. Please check the remaining items for the testcases from below

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
tonynajjar: I implemented the logic that will store incoming polygons to a vector plus some additional logic to keep this vector up to date:

- if a new polygon arrives that is equivalent to an existent one in the vector but is more recent, then it replaces it
- if any polygon is older than source_timeout it's cleared from the vector

The problem I'm facing is that if we're publishing a "moving polygon", then there will be a trail (of the old poses of this polygon) because the old poses in the vector do not satisfy any of the above stated conditions initially. Only after they are older than source_timeout will they be cleared

image

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: probably, GPT generated that part :p

nav2_collision_monitor/test/sources_test.cpp
tonynajjar: good point, not intentional

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: nope, fails:

```
/code/ros2_ws/src/navigation2/nav2_collision_monitor/src/polygon_source.cpp:125:15: error: no matching function for call to ‘max(size_t, int)’
  125 |       std::max(static_cast<size_t>(segment_length / sampling_distance_), 1);
```

nav2_collision_monitor/src/polygon_source.cpp
doisyg: I did not dig in this PR. I will probably do once https://github.com/ros-planning/navigation2/pull/3880 is in and if this one is rebased on it to check all timeout related behavior.

doisyg: > FYI @doisyg for timeout-ing: how should we handle if its multiple data? If any single one is bad or if all are bad?

Behavior wise, I think from the moment one source is bad (but with https://github.com/ros-planning/navigation2/pull/3880 we could have more complex behavior with independent per-source configuration)

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: It doesn't need to be size_t.

Also, please add a comment anywhere generative AI was used to generate code. I'm working on a proposal around that and it would be good practice for me to see if my suggestions are helpful :-)

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
SteveMacenski: That's what I thought about the frame_id containing some unique element of the particular polygon - but I could see how that wouldn't work if everything is pre-transformed. I'd really ideally like to **not** create a new collision-monitor specific message type, since that wouldn't be portable & it would require a shim to any natural messages they might use. A name or uuid would do the trick, but not having something like that in the msg definition causes a problem. 

I suppose another option is to have `N` polygon sources each of their own topic for each individual polygon. For these polygons you're using, how many are we talking about would be of interest? 1? 2?  10? 100? Would having 1 topic-source per robot footprint be doable?

Another option: if the polygons are stamped and we make some statement that the polygons to run multiple in 1 instance need to be sychronized, we could check for any that have a timestamp older than the latest by `eta` and remove those. Think something like a depth processing pipeline with stereo image pairs. Its a little hacky. 

The last option is the first you had; just have a custom message. I suppose we could do that, though I'd probably want that to include the polygon, stamp, and some unique identifier so we can fully solve the problem if we had that. The array might not be necessary at that point then to allow it to be more distributed. Though, I suppose it could continue to be a vector as long as we don't just store the last message but buffer `N` data sources

tonynajjar: I implemented my proposal and it works quite well, I tried to minimize the risk of wrongfully clearing polygons by making `arePolygonsSimilar` check 2 things, the polygons are considered similar if:

1- they have the same number of points
2- Pair by pair, their respective points all have a distance smaller than `polygon_similarity_threshold_`

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: I ran ament_uncrustify on it, please specify what you consider weird so I can fix it.

nav2_collision_monitor/test/sources_test.cpp
tonynajjar: fixed

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -400,6 +407,7 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in)
     marker.color.r = 1.0;
     marker.color.a = 1.0;
     marker.lifetime = rclcpp::Duration(0, 0);
+    marker.frame_locked = true;
```
tonynajjar, line 410: follow up for https://github.com/foxglove/studio/issues/6991

tonynajjar:
I thin most points have been addressed, please have another look.

SteveMacenski:
@AlexeyMerzlyakov please re-review

@tonynajjar this file has by far the most missing coverage of the Collision Monitor files https://app.codecov.io/gh/ros-planning/navigation2/blob/add-polygon-source-main/nav2_collision_monitor%2Fsrc%2Fpolygon_source.cpp#L179 please fill in the gaps. I'm not worried about the `node.lock()` exceptions, but there is a complete function missing an a for loop totally unexecuted.

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * @brief Adds latest data from polygon source to the data array.
+   * @param curr_time Current node time for data interpolation
+   * @param data Array where the data from source to be added.
+   * Added data is transformed to base_frame_id_ coordinate system at curr_time.
+   */
+  void getData(
+    const rclcpp::Time & curr_time,
+    std::vector<Point> & data) const;
+
+  void convertPolygonStampedToPoints(
```
SteveMacenski, line 77: Missing doxygen

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+void PolygonSource::getData(
+  const rclcpp::Time & curr_time,
+  std::vector<Point> & data) const
+{
+  // Ignore data from the source if it is not being published yet or
+  // not published for a long time
+  if (data_.empty()) {
+    return;
+  }
+
```
SteveMacenski, line 78: Missing a check on timeouts if this data is valid using the object constructor argument `source_timeout`

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+    geometry_msgs::msg::TransformStamped tf = tf2::toMsg(tf_transform);
+    tf2::doTransform(polygon_stamped, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+}
+
+void PolygonSource::convertPolygonStampedToPoints(
+  const geometry_msgs::msg::PolygonStamped & polygon, std::vector<Point> & data) const
+{
+  /* Generated by GPT */
```
SteveMacenski, line 114: What part of it, specifically? Or the entire function?

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+    throw std::runtime_error{"Failed to lock node"};
+  }
+  auto curr_time = node->now();
+
+  // remove old data
+  data_.erase(
+    std::remove_if(
+      data_.begin(), data_.end(),
+      [this, curr_time](const geometry_msgs::msg::PolygonStamped & polygon_stamped) {
+        return curr_time - rclcpp::Time(polygon_stamped.header.stamp) > source_timeout_;
```
SteveMacenski, line 174: I can see why you thought you could remove the check with source timeout from the main file -- but there's no promise that this callback is run in any particular frequency or regularity. We still need to check the polygon's timeout in the main getData function

SteveMacenski: Better yet, move this up there -- we're looping through the data anyway in `getData`, do the deletion there and remove here

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
tonynajjar: @smac is the proposes solution satisying to you? (I'll worry abou the testing at the end to avoid double work)

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+      }), data_.end());
+
+  // check if older similar polygon exists already and replace it with the new one
+  for (auto & polygon_stamped : data_) {
+    if (msg->header.frame_id == polygon_stamped.header.frame_id &&
+      rclcpp::Time(msg->header.stamp) > rclcpp::Time(polygon_stamped.header.stamp) &&
```
SteveMacenski, line 180: or equal to, if they're literally the same?

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+  if (polygon1.points.size() != polygon2.points.size()) {
+    return false;
+  }
+
+  // Compare the distance between each pair of corresponding points
+  for (size_t i = 0; i < polygon1.points.size(); ++i) {
+    double dx = polygon1.points[i].x - polygon2.points[i].x;
+    double dy = polygon1.points[i].y - polygon2.points[i].y;
+    double dz = polygon1.points[i].z - polygon2.points[i].z;
+    double distance = std::sqrt(dx * dx + dy * dy + dz * dz);
```
SteveMacenski, line 206: How does this generalize?

SteveMacenski (changes requested):
Generally looks good, but I don't think the are polygons similar approach is valid general approach to this. This assumes alot about what specifically these polygons are, how often they're published, and that they're spatially near each other due to that regular frequency

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: I think if you want to replace data, we need to have (1) some kind of unique identifier or (2) have unique frames of each thing that we use TF to transform. (2) feels fragile, so I think the answer here is to have some unique identifier in the messages of the physical object the polygon belongs to. This feels like a good add to the standard messages too, let me inquire if we can make an API change

https://github.com/ros2/common_interfaces/issues/230

tonynajjar: For the unique ID, we could abuse the 'seq' of the Header msg. I say "abuse" because it's more meant as a sequence counter.

Regarding your PR, you could also make the case to adding the unique identifier to the Header msg, since this uid could be needed for many other msgs, e.g. Point, Vector, etc...

(I'm on vacation for the whole week; I'll try to answer but it might take a while or not be as detailled as I would wish)

SteveMacenski: There is no seq in ROS 2 :smiling_face_with_tear: 

No worries, enjoy vacation. I'd be disappointed if I heard from you the rest of the week :P

tonynajjar: Damn it, you're right I was looking at the ROS 1 version. Second part of my comment is still valid though.

SteveMacenski:
@tonynajjar when you're back, rebase and fix the API changes for `bool getData` and checking the timeout via https://github.com/ros-planning/navigation2/pull/3880 merging

nav2_collision_monitor/src/polygon_source.cpp
AlexeyMerzlyakov, line 206: For the case, if no Unique ID to be used: this could be done through `std::hypot(dx, dy, dz)`

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  // remove old data
+  data_.erase(
+    std::remove_if(
+      data_.begin(), data_.end(),
+      [this, curr_time](const geometry_msgs::msg::PolygonStamped & polygon_stamped) {
+        return curr_time - rclcpp::Time(polygon_stamped.header.stamp) > source_timeout_;
+      }), data_.end());
+
+  // check if older similar polygon exists already and replace it with the new one
```
AlexeyMerzlyakov, line 177: How do you handle moving polygon challenge? For example, the polygon is moving in some frame and its coordinates are being updated&published e.g. 5 times per second. In this case, this routine will add N-polygons at the same time unless old ones won't exceed `source_timeout_`. Changing `source_timeout_` to be `0.2 sec` to meet 5-times per a second rate handling one polygon at a time, seems to be not a good idea as `Source::sourceValid()` checker is also using the same timeout to determine whether the data is rust. And robot in this case might go with some "jerking" by stopping-running-stopping...

AlexeyMerzlyakov: I'd more prefer the idea with polygon `UUID` that @SteveMacenski mentioned in previous messages.
... or using polygons array from previous implementation seem also doable idea for me.

AlexeyMerzlyakov (commented):
The main challenge here - is how to solve movin polygon task, where polygon changing its coordinates during the time.
Whether should we use `unique_identifier_msgs/msg/UUID.msg`, or get back to previous implementation when all polygons are being published once per a time, or might be something else (which I did not noted).

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: You're right that the removal of outdated polygons can't be done here; I had it in the getData function originally but the problem is that the getData function is marked as `const`, meaning I can't modify `data_` in it. Are you fine with removing the const from that function? Keep in mind it has to be removed for all sources.

nav2_collision_monitor/include/nav2_collision_monitor/polygon_source.hpp
tonynajjar: done

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: all of it (checked by me of course), edited

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: discussed in comment below about getData being const

nav2_collision_monitor/src/polygon_source.cpp
tonynajjar: true, done

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "nav2_collision_monitor/polygon_source.hpp"
+
+#include <functional>
+#include <cmath>
```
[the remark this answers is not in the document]
tonynajjar: done

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+#include "geometry_msgs/msg/polygon_stamped.hpp"
+#include "tf2/transform_datatypes.h"
+
+#include "nav2_util/node_utils.hpp"
+#include "nav2_util/robot_utils.hpp"
+
+using rcl_interfaces::msg::ParameterType;
```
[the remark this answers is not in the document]
tonynajjar: no, removed

tonynajjar:
Addressed most of the comments, I'll wait for the result of https://github.com/ros2/common_interfaces/issues/230 to continue this PR  and the tests. It's also my preferred solution to the problem

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: While I agree that an id in the header could be useful, I think that is a uphill battle, but is worth a discussion!

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: Yeah, I'm fine with that!

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,215 @@
...
+    tf2::doTransform(polygon_stamped, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+  return true;
+}
+
+void PolygonSource::convertPolygonStampedToPoints(
+  const geometry_msgs::msg::PolygonStamped & polygon, std::vector<Point> & data) const
+{
+  /*  Full function generated by GPT */
```
SteveMacenski, line 114: @tonynajjar I'm looking into this more for writing ROS community guidelines and it looks like this is illegal. You can't copy right chatgpt code and can't use it unless you found the source it pulled it from to check the appropriate license. You would need to have substantively modified the gpt-generated code for it to be legal.

Please reimplement yourself

tonynajjar: What if you asked GPT to generate a bubble sort algorithm for you, would that be illegal to use? Some things don't have multiple implementations... is changing the variable names + minor tweaks enough?

SteveMacenski: In my current understanding of the law: yes, it is illegal. You should reimplement without looking at this methods current implementation.

SteveMacenski:
@tonynajjar Polygons with IDs have been added! https://github.com/ros2/common_interfaces/pull/232#event-11529615113

That can be used now to implement this PR to get over the issue :-)

tonynajjar:
Awesome! In the meantime I left my old company that needed this and am quite busy with my new one. @jplapp maybe you want to take this over? If not I'll try to eventually find the time but can't promise when.

SteveMacenski:
It should be straight forward now if using those messages. All of the `arePolygonsSimilar` functionality can be deleted and just check the IDs. Add a few unit tests and a nice video and I think its done.

jplapp:
Thanks for review! Sure, I'll find someone to complete this, but cannot promise a timeline

oldtopos:
@jplapp I have cycles to pick this up; it seems there is one use of arePolygonsSimilar to replace, so this should be straightforward.  LMK if that works for you

oldtopos:
I am attempting to copy the branch associated with the PR to my own fork of nav2 in github, as I presume (and have not tested) I do not have commit access to the repo currently associated with the merge.  Will see how the build goes tomorrow, fingers crossed

Issue #3891, opened 2023-10-19
[Suggestion] Code-modification to improve the compatibility with the latest version of the compilation environment


GoesM:
# [Suggestion] for code-modification for Navigation2 [within ROS2-humble, Ubuntu 22.04] 
## To improve the compatibility of navigation2-[ROS2-humble] with the latest version of the compilation environment

# Problem Description

For most current versions of compilers, ` nav2_ costmap_ 2d` package often fails to compile smoothly,

Compiling ROS2 humble navigation2 using the latest version of gcc (such as gcc-11,gcc-12) or LLVM compiler (such as clang-11, clang-12, clang-14) will result in the following error:

```bash
/...../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_pair.h:314:17: error: the parameter for this explicitly-defaulted copy constructor is const, but a member or base requires it to be non-const
      constexpr pair(const pair&) = default;    ///< Copy constructor
```

**The compilation instructions we used are as follows**

we had tried CC&CXX as clang-11, clang-12, clang-14, gcc-11, gcc-12 and so on, our compilation instruction format is similar to:
```
source /opt/ros/humble/setup.bash
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++
colcon build \
	--cmake-clean-cache \
	--cmake-args \
		-DBUILD_TESTING=OFF \
		-DCMAKE_CXX_STANDARD=17 \
		-DCMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS} \
			-w -Wno-error -Wno-everything" \
		-DCMAKE_C_FLAGS="${CMAKE_C_FLAGS}  \
			-w -Wno-error -Wno-format-security"
```
**It seems that only gcc-9 could complie the program smoothly, however it's a very old version**

# Our Suggestions for Modification

## suggestion for [nav2_costmap_2d] pkg

**here's our core suggestion**

In  `navigation2/nav2_costmap_2d/include/nav2_costmap_2d/denoise/image.hpp`,  modify the variable type as const:

The specific modification location is indicated in the following code comments:
```cpp
template<class T>
class Image
{
public:
   ...
  /**
   * @brief Create image referencing to the other
   * Share image data between new and old object.
   * Changing data in a new object will affect the given one and vice versa
   */
//Image(Image & other)
  Image(const Image & other); // add " const " here !!!
   ...

template<class T>
//Image<T>::Image(Image & other) 
[...]

PR #3895, opened 2023-10-23
Adjust the Variable types in Nav2_costmap_2d pkg in [nav2_humble] 

GoesM:
#3891 

To improve the compatibility of navigation2-[ROS2-humble] with the latest version of the compilation environment , we modify the code in `nav2_costmap_2d` refer to the ISSUE #3891


SteveMacenski:
@GoesM please retarget to `main` and I'll backport once approved. We want the git history to move in one direction.

PR #3914, opened 2023-10-28
Enabling soft realtime prioritization to the Controller Server

SteveMacenski:
Requires modifying `/etc/security/limits.conf` with 

```
<username> hard rtprio 99
<username> soft rtprio 99
```

to give the userspace the permissions to set thread priorities. After which, enable `use_realtime_priority` in the controller server to have it use elevated permissions to meet scheduling deadlines of the sensitive server 

Docs PR: https://github.com/ros-planning/navigation.ros.org/pull/482

### DIFF
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -179,6 +179,26 @@ public:
   }
 
   /**
+   * @brief Sets thread priority level
+   */
+  void setSoftRealTimePriority()
+  {
+    if (use_realtime_prioritization_) {
+      sched_param sch;
+      sch.sched_priority = 90;
+      if (sched_setscheduler(0, SCHED_FIFO, &sch) == -1) {
+        std::string errmsg(
+          "Cannot set as real-time thread. Users must set: <username> hard rtprio 99 and "
+          "<username> soft rtprio 99 in /etc/security/limits.conf to enable "
+          "realtime prioritization! Error: ");
+        throw std::runtime_error(errmsg + std::strerror(errno));
+      } else {
+        debug_msg("Soft realtime prioritization successfully set!");
+      }
+    }
+  }
+
+  /**
    * @brief Handles accepted goals and adds to preempted queue to switch to
    * @param Goal A server goal handle to cancel
    */
@@ -211,19 +231,7 @@ public:
       // Return quickly to avoid blocking the executor, so spin up a new thread
       debug_msg("Executing goal asynchronously.");
       execution_future_ = std::async(std::launch::async, [this]() {
-        if (use_realtime_prioritization_) {
-          sched_param sch;
-          sch.sched_priority = 90;
-          if (sched_setscheduler(0, SCHED_FIFO, &sch) == -1) {
-            std::string errmsg(
-              "Cannot set as real-time thread. Users must set: <username> hard rtprio 99 and "
-              "<username> soft rtprio 99 in /etc/security/limits.conf to enable "
-              "realtime prioritization! Error: ");
-            throw std::runtime_error(errmsg + std::strerror(errno));
-          } else {
-            debug_msg("Soft realtime prioritization successfully set!");
-          }
-        }
+        setSoftRealTimePriority();
         work();
       });
     }
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -230,10 +230,11 @@ public:
 
       // Return quickly to avoid blocking the executor, so spin up a new thread
       debug_msg("Executing goal asynchronously.");
-      execution_future_ = std::async(std::launch::async, [this]() {
-        setSoftRealTimePriority();
-        work();
-      });
+      execution_future_ = std::async(
+        std::launch::async, [this]() {
+          setSoftRealTimePriority();
+          work();
+        });
     }
   }
 
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -185,7 +185,7 @@ public:
   {
     if (use_realtime_prioritization_) {
       sched_param sch;
-      sch.sched_priority = 90;
+      sch.sched_priority = 49;
       if (sched_setscheduler(0, SCHED_FIFO, &sch) == -1) {
         std::string errmsg(
           "Cannot set as real-time thread. Users must set: <username> hard rtprio 99 and "
--- a/nav2_util/include/nav2_util/simple_action_server.hpp
+++ b/nav2_util/include/nav2_util/simple_action_server.hpp
@@ -57,6 +57,8 @@ public:
    * @param server_timeout Timeout to to react to stop or preemption requests
    * @param spin_thread Whether to spin with a dedicated thread internally
    * @param options Options to pass to the underlying rcl_action_server_t
+   * @param realtime Whether the action server's worker thread should have elevated
+   * prioritization (soft realtime)
    */
   template<typename NodeT>
   explicit SimpleActionServer(
@@ -66,13 +68,15 @@ public:
     CompletionCallback completion_callback = nullptr,
     std::chrono::milliseconds server_timeout = std::chrono::milliseconds(500),
     bool spin_thread = false,
-    const rcl_action_server_options_t & options = rcl_action_server_get_default_options())
+    const rcl_action_server_options_t & options = rcl_action_server_get_default_options(),
+    const bool realtime = false)
   : SimpleActionServer(
       node->get_node_base_interface(),
       node->get_node_clock_interface(),
       node->get_node_logging_interface(),
       node->get_node_waitables_interface(),
-      action_name, execute_callback, completion_callback, server_timeout, spin_thread, options)
+      action_name, execute_callback, completion_callback,
+      server_timeout, spin_thread, options, realtime)
   {}
 
   /**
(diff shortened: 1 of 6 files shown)


nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -202,7 +210,22 @@ class SimpleActionServer
 
       // Return quickly to avoid blocking the executor, so spin up a new thread
       debug_msg("Executing goal asynchronously.");
-      execution_future_ = std::async(std::launch::async, [this]() {work();});
+      execution_future_ = std::async(std::launch::async, [this]() {
```
embeddedadam, line 213: Consider extracting added code to function like: setSoftRealTimePriority() and call it here. It would improve readability.

nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -170,6 +178,26 @@ class SimpleActionServer
...
 
+  /**
+   * @brief Sets thread priority level
+   */
+  void setSoftRealTimePriority()
+  {
+    if (use_realtime_prioritization_) {
+      sched_param sch;
+      sch.sched_priority = 90;
+      if (sched_setscheduler(0, SCHED_FIFO, &sch) == -1) {
```
JanStaschulat, line 189: This PR looks great to me. 

The only comment I have is: that `sched_setscheduler()` works only on Linux machines.
If you want support for Windows and Mac, see e.g. the function `configure_native_thread`, which @ralph-lange has written to assign the thread priority to support Linux/Windows/Mac operating system:
 
https://github.com/ros2/examples/blob/rolling/rclcpp/executors/cbg_executor/src/examples_rclcpp_cbg_executor/utilities.hpp

It is used in the callback group-level executor example:
https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/cbg_executor/src/pong.cpp#L78

You might consider using a similar utility function here as well.

ralph-lange: In my opinion, `configure_native_thread` should factored out for this purpose (to the rcutils package or to rclcpp?) and generalized with numeric priorities instead of the enum supporting low and high only. Happy to review such PR.

SteveMacenski: Thanks for the remark - we don't typically think about windows/apple but if there's a utility that would allow us to have support for it easily, more than happy to ensure cross-compatibility! 

However, it looks like the signature of those tools requires an `std::thread` input, where we're using `std::async` to have a future result. It could be refactored to use `std::thread`, but it would be alot less clean.

So, even if that utility was in rclutils, it wouldn't, without modifications, be able to help us here. There would need to be modifications to the apple/qnx/windows implementations to use the current thread instead of passing in the thread object. Would that be something possible / Bosch interested in doing?

tonynajjar:
To fully benefit of this, wouldn't it make sense to also raise the priority of all the nodes part of the `/cmd_vel` pipeline, i.e. collision_monitor and velocity_smoother?

SteveMacenski:
That's a good point - I'd like to get to the bottom of the discussion with @JanStaschulat in the thread above first, but once we have priority levels that are proportionate to the task at hand (sounds like 90 is too high) and understanding if there's tools in `rcutils` future to use, then expanding to those as well would make sense.

JanStaschulat:
the ros2_control guys use priority 50 (same priority as system calls): 

https://github.com/ros-controls/ros2_control/blob/cf405bc3d3ad6c999a35a98faccf871a9b00c008/controller_manager/src/ros2_control_node.cpp#L32

I would start experimenting with a priority below system calls, say 40, and then evaluate if the soft real-time behavior is good enough. If not, I would also raise the priority to system calls, as in ros2_control framework.

SteveMacenski:
Thanks! This is on my agenda today to complete testing and evaluation

SteveMacenski:
@JanStaschulat quick question on rclcpp::Timers: can I use the default callback group / executor of the component given by the container,  can I simply set the thread priorities in the node's constructor? The velocity smoother [1] is another area where increased priority would be helpful. I believe this is true of subscriptions, which solves the collision monitor case [2].

[1] https://github.com/ros-planning/navigation2/blob/main/nav2_velocity_smoother/src/velocity_smoother.cpp#L145-L147
[2] https://github.com/ros-planning/navigation2/blob/main/nav2_collision_monitor/src/collision_monitor_node.cpp#L66-L68

I went through the real-time workshop slides and all of the examples I found using current tooling involved creating an executor, adding nodes to it, and then spinning it in a separate thread where the priorities were set. When using Composition, which Nav2 extensively uses, we don't create our own executors to be able to control them -- the component container does.  I wanted to touch base before I spent too much time on those nodes [1-2] in case there was a significant technical reason why you didn't add them in-line in the code and did so at the `main.cpp` level with the executor population. I suspect it was just the most general for demonstration without getting into the specifics of the nodes, but just verifying that. 

PS: Perhaps that's a good area to set soft real-time prioritization as well as an argument option? The composition containers could have options about which node(s) or if all nodes in the container should be set with higher priority. For the single threaded and multithreaded containers; its all. But the Isolated Single Threaded Container we added and use in Nav2 can have a per-node selection since each nodes gets its own isolated executor.

JanStaschulat:
Hi @SteveMacenski ,
my apologies for the late reply: prio 49 seems to be a good starting point. 

Regarding components:
>... if I use the default callback group / executor of the component given by the container, can I simply set the thread priorities in the node's constructor?

Which API to set the priority in a node's constructor are you refering to?
I am not aware that such a parameter to assign a priority exists in [NodeOptions](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp/include/rclcpp/node_options.hpp#L35).

> When using Composition, which Nav2 extensively uses, we don't create our own executors to be able to control them -- the component container does.

The component container creates the Executors, yes: e.g. here: with [SingleThreadedExecutor](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/src/component_container.cpp#L25C1-L25C1), or with a [MultiThreadedExecutor](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/src/component_container_mt.cpp#L30). But the Component Container does not set the priority of the threads. This might be a good contribution.

>  I wanted to touch base before I spent too much time on those nodes [1-2] in case there was a significant technical reason why you didn't add them in-line in the code and did so at the main.cpp level with the executor population

We wanted to keep it simple. Actually, we did both. But we did not highlight it in the workshop:
in-line thread creation [exercise4-2/ImagePublisherNode](https://github.com/ros-realtime/roscon-2023-realtime-workshop/blob/10c9f9652c33cf7577363391202c8c0920061933/exercise4-2/src/camera_demo/src/system_nodes.cc#L41)
in some [main.cpp](https://github.com/ros-realtime/roscon-2023-realtime-workshop/blob/10c9f9652c33cf7577363391202c8c0920061933/exercise4-2/solutions/main.cc#L35)

> PS: Perhaps it's a good area to set soft real-time prioritization as an argument option for the component containers in rclcpp?

Fully agree.

> The composition containers could have options about which node(s) or if all nodes in the container should be set with higher priority. For the single threaded and multithreaded containers; its all since they share a single executor. But the Isolated Single Threaded Container we added and use in Nav2 can have a per-node selection since each nodes gets its own isolated executor. That would be another way to generically decide at launch-time which components to make high priority without modifying the nodes themselves. That would let all of your demos then work analog to the workshop presentation slides, but now as a matter of yaml param or launch API argument instead of modifying any C++ code at all to have to manually-specify which at compile time needs what priority. That seems like an easy win!

Fully agree. 

I briefly looked at the source code: The ComponentManager (SingleThreaded/MultiThreaded) don't create additional threads, when adding nodes:
- [ComponentManager::add_node_to_executor](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/src/component_manager.cpp#L202)
and then they call `spin()` in main:
- [component container (SingleThreaded)](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/src/component_container.cpp#L25)
- [component container (MultiThreaded)](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/src/component_container_mt.cpp#L26)

But the IsolatedContainer does create new threads when `add_node_to_executor` is called: 
[ComponentManagerIsolated::add_node_to_executor](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp#L80)

This could be easily be extended with the assignment of some scheduling policy.

SteveMacenski:
> Which API to set the priority in a node's constructor are you refering to?

I'm saying to put the following in the object's constructor (tweaked for context): 

```

  void setSoftRealTimePriority()
  {
    if (use_realtime_prioritization_) {
      sched_param sch;
      sch.sched_priority = 49;
      if (sched_setscheduler(0, SCHED_FIFO, &sch) == -1) {
SteveMacenski marked this conversation as resolved.
        std::string errmsg(
          "Cannot set as real-time thread. Users must set: <username> hard rtprio 99 and "
          "<username> soft rtprio 99 in /etc/security/limits.conf to enable "
          "realtime prioritization! Error: ");
        throw std::runtime_error(errmsg + std::strerror(errno));
      } else {
        debug_msg("Soft realtime prioritization successfully set!");
      }
    }
  }
```

Or, perhaps in the configure state transitions? What do you think is the right place for that change

> But the IsolatedContainer does create new threads when add_node_to_executor is called:
[ComponentManagerIsolated::add_node_to_executor](https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp#L80) This could be easily be extended with the assignment of some scheduling policy.

Is that something you'd be open to contributing back or something I should put on my queue?

SteveMacenski:
https://github.com/ros2/rclcpp/pull/2373 I took a stab at it. I know once your RT work is ready to be merged / reviewed, this will change, but I think this is a good middle-term solution

Issue #3894, opened 2023-10-22
Goal frame_id is not respected consistently


meyerj:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Focal 20.04 (ROS and navigation2 built from source)
- ROS2 Version:
  - Humble (built from source)
- Version or commit hash:
  - 9876fe1b4faa79bcfe8a8fa50b3c4da2eeaf9d1a
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue

- Robot spawned at position (10, 0), or drove close to position (10, 0) before.
- Empty costmap, no obstacle detection involved.
- Request the `navigate_to_pose` action:
  ```sh
  ros2 action send_goal /navigate_to_pose nav2_msgs/action/NavigateToPose "
  pose:
    header:
      frame_id: base_link
    pose:
      position:
        x: 10.0
        y: 0.0
      orientation:
        w: 1.0
  "
  ```

#### Expected behavior

The robot navigates to a point 10 meters ahead of the current pose, then the action succeeds. With the patch from #2222 (for #2186), which addressed the issue for the `planner_server`, I would have expected that it is possible to send goals in any frame, including `base_link` or whatever the robot's base frame is. But apparently the goal's header frame_id is still ignored in various other places, for example in [`NavigateToPoseNavigator::initializeGoalPose()`](https://github.com/ros-planning/navigation2/blob/57f55c4dca45d93b61adbf7bcd24ad83106192bd/nav2_bt_navigator/src/navigators/navigate_to_pose.cpp#L204-L225) and [`GoalReachedCondition::isGoalReached()`](https://github.com/ros-planning/navigation2/blob/57f55c4dca45d93b61adbf7bcd24ad83106192bd/nav2_behavior_tree/plugins/condition/goal_reached_condition.cpp#L71-L88).

Like the original reporter of #2186, I would expect the header frame_id to be respected in _each_ consumer of the current goal if it is represented as `geometry_msgs/msg/PoseStamped`, independent of whether it has been already transformed to the global navigation frame before or not. Otherwise, if the `frame_id` is expected to match the global frame by design and the header stamp also does not play a role, then the goal could have been represented by a `geometry_msgs/msg/Pose` message only internally and all action requests.

[...]

PR #3917, opened 2023-10-29
Enable goals to be sent in other frames than the pre-configured global_frame

meyerj:
This patch was meant to address the issue described in https://github.com/ros-planning/navigation2/issues/3894, at least partially: Some ROS and behavior tree nodes do not check the header frame_id of the goal and assume that goals are provided in their respective global frame.

I tested this patch with our own setup for an Ackermann-steered quad-like robotic platform, and also with the `tb3_simulation_launch.py` from the [Getting Started](https://navigation.ros.org/getting_started/index.html) guide. While this patch does solve the issue for the former, unfortunately I could not reproduce the issue with the latter though. We are using the `nav2_smac_planner/SmacPlannerHybrid` planner and the `nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController` with the [`navigate_w_replanning_only_if_path_becomes_invalid.xml`](https://github.com/ros-planning/navigation2/blob/57f55c4dca45d93b61adbf7bcd24ad83106192bd/nav2_bt_navigator/behavior_trees/navigate_w_replanning_only_if_path_becomes_invalid.xml) behavior tree, so a different setup than the Turtlebot example setup.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/3894 |
| Robotic platform tested on | - Custom Ackermann-steered quad-like robotic platform<br>- `tb3_simulation_launch.py` from the [Getting Started](https://navigation.ros.org/getting_started/index.html) guide |

---

- The stamped goal pose will be transformed to the configured `global_frame` of the `bt_navigator` node in `NavigateToPoseNavigator::initializeGoalPose()` before initialization of the BehaviorTree blackboard.

  The following behavior nodes _maintained in this repository_ consume the `goal` input and may be affected:
  - [`ComputePathToPose`](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp)
    - forwarded to the `ComputePathToPose` action server (`planner_server`) in the action request.
    - The `planner_server` transforms a goal to its own global frame, if that would differ from the global frame of the `bt_navigator` node. Otherwise the transform by the `planner_server` is a no-op.
  - [`NavigateToPose`](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/action/navigation_to_pose_action.cpp)
    - forwarded to the `NavigateToPose` action server (`bt_navigator`) in the action request.
    - This action is not used by any behavior tree example in this repository. If it would be used in a behavior tree that implements the `NavigateToPose` action, that would be a recursion?
  - [`GoalReached`](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/condition/goal_reached_condition.cpp)
    - for comparison of the current pose in the goal pose.
    - Before this patch the header frame_id of the goal was not taken into account here. If this node is part of a behavior tree executed by the `bt_navigator` node, then the `goal` on the blackboard was already transformed to the global frame before now, and the change in `GoalReachedCondition::isGoalReached()` would not strictly be necessary because the `goal.header.frame_id` and the global frame do already match. However, because the input of this node is a stamped goal pose and it may also be used in other contexts, I suggest to do the comparison in the goal's frame in any case. The value read `global_frame` is redundant and only applied if the `goal.pose.frame_id` is empty.

- None. Implications of sending goal poses or paths in other than the respective configured global and local frame is not documented explicitly, or I have not found it yet.

---

- Apply a similar patch to [`NavigateThroughPosesNavigator`](https://github.com/meyerj/navigation2/blob/19a8165a75c9479e9d1bc017eab87255dc09c579/nav2_bt_navigator/src/navigators/navigate_through_poses.cpp), for consistency?

- The `planner_server` does already transform the goal point(s) to the global frame [here](https://github.com/ros-planning/navigation2/blob/4b4465dfc9427b95e98aef70620bedd933cfbe56/nav2_planner/src/planner_server.cpp#L517-L521) and [here](https://github.com/ros-planning/navigation2/blob/4b4465dfc9427b95e98aef70620bedd933cfbe56/nav2_planner/src/planner_server.cpp#L398-L402), since https://github.com/ros-planning/navigation2/pull/2222. However, the `controller_server` does not and it depends on the controller implementation whether the transform to the local frame is applied only once upon receiving a new plan or continuously, at each [call of `computeVelocityCommands()`](https://github.com/ros-planning/navigation2/blob/4b4465dfc9427b95e98aef70620bedd933cfbe56/nav2_controller/src/controller_server.cpp#L590-L594). It also transforms the goal pose to the local frame (the "global frame" of the local costmap) in [`ControllerServer::isGoalReached()`](https://github.com/ros-planning/navigation2/blob/4b4465dfc9427b95e98aef70620bedd933cfbe56/nav2_controller/src/controller_server.cpp#L731-L735) (since https://github.com/ros-planning/navigation2/pull/2780).

  In case of the `controller_server` not transform the goal path initially may be desirable, because updates of the transform between the global and local frame are supposed to have a direct effect on the controller for a path provided in map coordinates, without the need to replan. But maybe it would be good to document this implicit behavior better? Or even to add a parameter and optionally transform the received path only once, in [`ControllerServer::computeControl()`](https://github.com/ros-planning/navigation2/blob/4b4465dfc9427b95e98aef70620bedd933cfbe56/nav2_controller/src/controller_server.cpp#L416)?

- The configured [goal checker](https://navigation.ros.org/plugins/index.html#goal-checkers) used by the controller server may disgree with the behavior tree's `GoalReachedCondition`, which is only based on the Euclidian 2D distance. Depending on the concrete behavior tree this may have unwanted implications? However, I noticed that this condition node is not even used by any of the behavior trees in `nav2_bt_navigator`, so this concern cannot be the culprit of the behavior described in https://github.com/ros-planning/navigation2/issues/3894.

#### For Maintainers: 


nav2_bt_navigator/src/navigators/navigate_to_pose.cpp
```diff
@@ -209,18 +209,29 @@ NavigateToPoseNavigator::initializeGoalPose(ActionT::Goal::ConstSharedPtr goal)
     feedback_utils_.global_frame, feedback_utils_.robot_frame,
     feedback_utils_.transform_tolerance);
 
+  geometry_msgs::msg::PoseStamped goal_pose;
+  if (!nav2_util::transformPoseInTargetFrame(
```
SteveMacenski, line 213: A change here should also be made in the Navigate Though Poses navigator as well

nav2_bt_navigator/src/navigators/navigate_to_pose.cpp
```diff
@@ -209,18 +209,29 @@ NavigateToPoseNavigator::initializeGoalPose(ActionT::Goal::ConstSharedPtr goal)
...
+        goal->pose, goal_pose, *feedback_utils_.tf, feedback_utils_.global_frame,
+        feedback_utils_.transform_tolerance)) {
+    RCLCPP_ERROR(
+      logger_,
+      "Failed to transform a goal pose provided with frame_id '%s' to the global frame '%s'.",
+      goal->pose.header.frame_id.c_str(), feedback_utils_.global_frame.c_str());
+    throw std::runtime_error("transformation failed");
```
SteveMacenski, line 220: What catches this throw? Also make the message more verbose

nav2_behavior_tree/plugins/condition/goal_reached_condition.cpp
```diff
@@ -73,17 +73,18 @@ void GoalReachedCondition::initialize()
 
 bool GoalReachedCondition::isGoalReached()
 {
-  geometry_msgs::msg::PoseStamped current_pose;
+  geometry_msgs::msg::PoseStamped goal;
+  getInput("goal", goal);
+  if (goal.header.frame_id.empty()) goal.header.frame_id = global_frame_;
```
SteveMacenski, line 78: Lets make a variable `std::string global_frame` that we populate in this if condition and then use in the `getCurrentPose` method. Lets not overwrite things that we don't expect the node to do

SteveMacenski: Also properly format this, don't use one line.

SteveMacenski (changes requested):
I generally approve of both of these actions - but is this exhaustive? It seems to me that potentially other updates wrt `global_frame` could be made to other BT nodes. I don't want to only partially solve the issue

nav2_behavior_tree/plugins/condition/goal_reached_condition.cpp
SteveMacenski: ^ Actually I wonder if we should even bother. If a goal header isn't set, wouldn't alot of other things also have errors and fail at various stages? Maybe we can actually just delete the global_frame port completely.

SteveMacenski:
@meyerj its been 2 weeks - can you please respond to the review so we can get this resolved?

nav2_bt_navigator/src/navigators/navigate_to_pose.cpp
meyerj: What is the expected behavior then if the goal pose (or one of the goal poses) cannot be transformed? Is it okay to change the return type of `NavigateToPoseNavigator::initializeGoalPose()` and `NavigateThroughPosesNavigator::initializeGoalPoses()` to `bool`, such that the error will be propagated to the respective `goalReceived()` callback? I assume that returning `false` here implies that the goal is rejected?

What additional information would you expect to be logged? We could add something like "goal rejected..."?

Also note that `nav2_util::transformPoseInTargetFrame()` already logs the failure [here](https://github.com/ros-planning/navigation2/blob/19a8165a75c9479e9d1bc017eab87255dc09c579/nav2_util/src/robot_utils.cpp#L50-L75) depending on the exact type of `tf2::TransformException`, so this log statement is already kind of redundant.

I also noticed that the return value of `getCurrentPose()` is currently not checked in the `onLoop()` callbacks and in `NavigateToPoseNavigator::initializeGoalPose()`. That could also fail, for example if global localization was lost and transforms timed out. Then the `current_pose` is left default initialized, with undefined behavior?

meyerj:
> > In case of the controller_server not transform the goal path initially may be desirable, because updates of the transform between the global and local frame are supposed to have a direct effect on the controller for a path provided in map coordinates, without the need to replan. But maybe it would be good to document this implicit behavior better? Or even to add a parameter and optionally transform the received path only once, in [ControllerServer::computeControl()](https://github.com/ros-planning/navigation2/blob/4b4465dfc9427b95e98aef70620bedd933cfbe56/nav2_controller/src/controller_server.cpp#L416)?
> 
> The transform between map and odom (the typical global and local frames) can change over time, so it is actually important to update the path each iteration to account for drift / odometric motion. Can you clarify what you think is implicit and might be desired to disable?

Maybe "implicit" is not the right word: What I meant is that some action servers transform the goals (or whatever input pose?) only once initially, and others keep looking up the transform continuously in each update cycle, like the controller. I think the behavior is fine as-is and no extra parameter is needed. But the navigator actions now transform the goal(s) only once initially, although multiple replanning steps may be triggered while running the behavior. Those decisions do have an effect on the behavior if the goals are not provided in the respective pre-configured frame, and hence would be worth to be mentioned in the documentation somewhere?

Or was that even by design that another trigger of the planner action (e.g. `ComputePathToPose`) may actually run with another _transformed_ goal pose, if the transform to the global frame changed in between? That would be an indirect way to update the goal pose then, that is not covered by the `GoalUpdated` condition, for example, and this patch would be invalid.

SteveMacenski:
> But the navigator actions now transform the goal(s) only once initially, although multiple replanning steps may be triggered while running the behavior.

I think we've beaten this topic with a stick. You should navigate with respect to some external frame. This is standard practice but certainly an explicit remark about it does not hurt in documentation, I would agree. Use `odom` if you're doing local moves without desiring for global localization correction -- or `map` for longer term actions that require updates in positioning from localization. This is established in [REP 105](https://www.ros.org/reps/rep-0105.html) for ROS and hasn't changed since the early days of robotics for a reason. Navigating w.r.t. an internal frame at an original time is incredibly fragile as we've discussed before due to transformation buffer rolling sizes, among others.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -57,16 +55,15 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
-      transform_tolerance_))
-  {
-    return BT::NodeStatus::FAILURE;
-  }
-
   double dist_to_goal;
   while (goal_poses.size() > 1) {
+    geometry_msgs::msg::PoseStamped current_pose;
+    if (!nav2_util::getCurrentPose(
+        current_pose, *tf_, goal_poses[0].header.frame_id, robot_base_frame_,
```
SteveMacenski, line 62: This can be moved back. The paths are all in the same frameID and you're pulling `[0]`s every time anyway.

nav2_bt_navigator/src/navigators/navigate_to_pose.cpp
SteveMacenski: > I assume that returning false here implies that the goal is rejected?

Correct, I think that's a good way to do it. I think your logs look good to me!

In `onPreempt`, you can also use the new return type of `initializeGoalPose` to `bt_action_server_->terminatePendingGoal();` as you see in the else statement next to it. 

In `onLoop`, you're right. Things are a little fragile there. These are primarily high level "FYI" metrics so I'm not overly concerned about them, but simple:

```
if (!  nav2_util::getCurrentPose(
    current_pose, *feedback_utils_.tf,
    feedback_utils_.global_frame, feedback_utils_.robot_frame,
    feedback_utils_.transform_tolerance))
{
  // LOG SOMETHING
  // return;
}
```
and not publish feedback - that sounds very reasonable

SteveMacenski (changes requested):

SteveMacenski:
With those small updates, this looks good to go to me! Please `rebase` or merge in `main` so that CI is green

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
meyerj: The first element with index 0 is a different one in each cycle, because the loop either terminates or the first element gets removed. But assuming that the path that is processed here is always the output of a planner, the assumption that the frame_id is the same for all poses is valid. I will move the call back up, as requested.

In general, I am not aware of a specification besides "best practice" and "that is how the interface is meant to be used" that paths could not be provided with a different `frame_id` for each `PoseStamped` instance? The [`nav_msgs/Path`](https://github.com/ros2/common_interfaces/blob/ad69eab30975b967b8c840c690ab491b2d6234cd/nav_msgs/msg/Path.msg) definition is not explicit with respect to restrictions on nested frame_ids and whether the timestamps per pose play a role or not. If they don't, and the `frame_id` must also match that of the message header in all cases, then the definition could have used unstamped `geometry_msgs/Pose[] poses` instead a long time ago... In my experience different tools and libraries use the message type in different ways, some don't even populate the individual pose headers. Related: https://github.com/ros2/common_interfaces/issues/97.

meyerj:
> Navigating w.r.t. an internal frame at an original time is incredibly fragile due to transformation buffer rolling sizes, among others.

Not necessarily imho: It only implies that the transformation to the internal frame needs to happen only once early on, when the goals are received and before buffer timeouts could occur, as it is done with this patch now. rviz for example sends goal poses in whatever frame is selected as the fixed frame, and with a "now" timestamp. (Remeber to run rviz with the `use_sim_time` parameter set when running in simulation!). So strictly speaking that implies that the goal is meant to be transformed to the map frame (`global_frame`) only once, because if the goal's header stamp would be respected by the navigation stack and buffers would have infinite size, the transform would not change anymore while following the goal (with the exception of static transform updates). On the other hand, if the goal would be provided with a zero timestamp by the action client or publisher, that indicates that it is meant to be transformed repeatedly on each planner or behavior cycle using the latest available transform at the time, which is what actually happened in the navigation stack before this patch, just not consistently in all action nodes, and error-prone indeed because tf buffers are not infinite and the goal stamps are usually not zero.

nav2_behavior_tree/plugins/condition/goal_reached_condition.cpp
```diff
@@ -33,8 +33,6 @@ GoalReachedCondition::GoalReachedCondition(
 {
   auto node = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
 
-  global_frame_ = BT::deconflictPortAndParamFrame<std::string, GoalReachedCondition>(
```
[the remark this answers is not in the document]
meyerj: Done: https://github.com/ros-planning/navigation.ros.org/pull/489

SteveMacenski (approved):

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
SteveMacenski: The discussion you linked to I generally agreed that the `frame` was redundant and the `header` is mostly useful in the paths in case its a proper trajectory and each of the path poses have a timestamp attached to them - which could be made use of in the medium-term future.

SteveMacenski:
> It only implies that the transformation to the internal frame needs to happen only once early on, when the goals are received and before buffer timeouts could occur, as it is done with this patch now

That's also what you could do with the request; transform it into odom frame and send it. That's SOP. 

> On the other hand, if the goal would be provided with a zero timestamp by the action client or publisher, that indicates that it is meant to be transformed repeatedly on each planner or behavior cycle using the latest available transform at the time,

That is not a correct understanding of the intent, but may be able to be leveraged in the way you describe. A zero timestamp says to use the most recently available transformation in TF transformations when its transformed, not to do so every single time. That is again, incredibly fragile and highly not recommended. So many of the things needed to make robotics highly accurate after you hit a POC level comes down to time synchronization and properly propagating timestamps.

PR #3929, opened 2023-11-02
Handle NaNs in AMCL beam sensor model

hidmic:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Data-driven benchmark |

---

Pretty much what the title says. I accidentally came across this bug while running AMCL against a dataset with LaserScan messages that happened to bear NaNs every now and then.

Here I apply the same NaN check that the other sensor models use, for consistency.

None.

---

None.
#### For Maintainers: 

### DIFF
--- a/nav2_amcl/src/sensors/laser/beam_model.cpp
+++ b/nav2_amcl/src/sensors/laser/beam_model.cpp
@@ -74,7 +74,7 @@ BeamModel::sensorFunction(LaserData * data, pf_sample_set_t * set)
       obs_range = data->ranges[i][0];
 
       // Check for NaN
-      if (obs_range != obs_range) {
+      if (isnan(obs_range)) {
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


nav2_amcl/src/sensors/laser/beam_model.cpp
```diff
@@ -72,6 +72,12 @@ BeamModel::sensorFunction(LaserData * data, pf_sample_set_t * set)
     step = (data->range_count - 1) / (self->max_beams_ - 1);
     for (i = 0; i < data->range_count; i += step) {
       obs_range = data->ranges[i][0];
+
+      // Check for NaN
+      if (obs_range != obs_range) {
```
SteveMacenski, line 77: Actually check for NaNs properly, this is a really not-awesome way of doing it. https://cplusplus.com/reference/cmath/isnan/

SteveMacenski (changes requested):
I approve in concept, will backport to humble/iron once resolved

nav2_amcl/src/sensors/laser/beam_model.cpp
hidmic: Sure thing. See https://github.com/ros-planning/navigation2/pull/3929/commits/8ac39c06d0285f46389828b248806a26528737de.

SteveMacenski (approved):

Issue #3017, opened 2022-06-15
Feature Request: Custom/Virtual Obstacle Layer


tonynajjar:
Hey, in our use case, we would like to use https://github.com/GMahmoud/virtual_costmap_layer with Nav2. Would you consider it a generic costmap layer that could be part of Nav2? Could it also be an extension of the Obstacle Layer?
We could work on contributing it.

@GMahmoud FYI

SteveMacenski:
I would indeed, I haven't looked in excessive detail how this works exactly, but the concept seems good to me to have in Nav2 and would be actually very convenient for testing purposes (mainly). Curious though for a more production reason what you might be doing with this? 

I'd want to know
- How the polygons/shapes are being filled in, and with what cost value
- How the polygons / shapes are being specified (service, topic, etc) and how to add / remove them 
- How it handles rolling costmaps or costmap updates efficiently 

But the general principle seems good.

SteveMacenski:
> Add RPC instead of published msgs in order to have the possibility of deleting obstacles with id identification,

I might recommend just adding a "add" and "remove" service and part of the request includes some string or int index for the objects. Not using pub/sub. 

> I have developed also some navigation plugins (global and local planning) that I want to make them compatible with ROS2 and Nav2 robot rocket. I will let you know in slack or in another Github issue/thread about this work.

Great to hear it and look forward to hearing about it!

SteveMacenski:
@tonynajjar so what's your plan here? Remake something using that as a reference or wait for it to be ported and we can go through some design reviews to patch it up? Or should we start design reviews here now with the intention of including the final results into Nav2 itself?

Its also totally fine if you keep it separate, no need to put it into nav2 if you don't want. We can always add it to our [Navigation Plugins](https://navigation.ros.org/plugins/index.html) list in our docs so that its searchable for users (and possible a tutorial if you're interested in that?)

jplapp:
I work on the same project as @tonynajjar and can give some context on our use:

We use multiple robots in a shared environment. There are some cases in which the onboard-sensors of robot A cannot fully see robot B. Thus, robots share their location (and planned path) with each other. Now, to help with local and global planning, we want to put the footprint of robot B on the costmap of robot A. 

 For this, robot A would publish the received footprint of robot B on a topic, and this new plugin would then add it to the costmap. I think services would add too much latency here, in case robot B is moving.

GMahmoud:
[...]

PR #3930, opened 2023-11-02
Add Vector Object server

AlexeyMerzlyakov:
Adding Vector Objects support to costmaps (and OccupancyGrid maps).

This is done by new Vector Object server, working as a separate node. In this server, Vector Objects are being putted into OccupancyGrid map. OccupancyMap is being published via output map topic (the performance question of map publishing was [discussed](https://github.com/ros-planning/navigation2/issues/3017#issuecomment-1440029323) in the original ticket). VO-server have an ability to support moving objects working dynamically with a given rate. If at least one vector object is being published in separate than "map" frame, resulting OccupancyGrid is being updated dynamically with this rate.

Support of vector objects in costmaps could be enabled by switching published by VO-server map as an input mask for Keepout Costmap Filter (it can't be done via Static Layer as it reads input map once and thus can not support moving objects; transfers input OccupancyGrid map to costmap w/o objects overlay on initial background; works with fixed-size input maps while VO-server can change its origin and dimensions).
Also, they are being able to work with other Costmap Filters (Speed Filter), to have a support for speed restricting areas specified by polygons).

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3017 |
| Robotic platform tested on | TB3 simulation in Gazebo |

---

* Added new Vector Object server into `nav2_map_server`
* Supported polygons and circles vector objects for now
* New messages: PolygonVO.msg and CircleVO.msg for them
* New services to work with Vector Objects: `AddShapes.srv`, `GetShapes.srv` and `RemoveShapes.srv`
* raytraceLine(), bresenham2D() algorithms were moved into common `nav2_utils` part
* Such functions as `maskToWorld()`/`worldToMask()`, `getMaskData()` were moved to `nav2_util/occ_grid_utils.hpp` common module working OccupancyGrid-s

Use-cases to be covered by this feature (from the #3017 discussions):
* Testing/virtual objects
* No-access-zone
* Speed restriction areas
* Robot footprint (or any other moving objects) replacement
* Areas restriction for UAVs to define a flying zone (outer / boundary) with obstacles (inner rings / holes)
* Remove some area from costmap (need KeepoutFilter modification: new "always overlay" overlay parameter)

* Chapter in Iron to Jazzy migration guide
* Parameters page on Configuration guide
* Demo tutorial of using the VO-server in costmaps

---

* Consider to move point-in-polygon implementation (which is currently written twice in the code: initially for Collision Monitor, and for VO-server) to common `nav2_utils` function
* Consider to move Bresenham3D algorithm into common `nav2_utils` part as well as it was made for Bresenham2D one
* Testcase coverage once design will be agreed

#### For Maintainers: 


nav2_map_server/src/vo_server/vector_object_shapes.cpp
```diff
@@ -0,0 +1,579 @@
...
+    if (point.x > max_x) {
+      max_x = point.x;
+    }
+    if (point.y > max_y) {
+      max_y = point.y;
+    }
+  }
+}
+
+bool Polygon::isPointInside(const double px, const double py) const
```
AlexeyMerzlyakov, line 184: This need to be considered move to common part (e.g. to `nav2_util/include/nav2_util/vector_object_utils.hpp`), as Collision Monitor has also the similar routine

nav2_util/include/nav2_util/raytrace_line_2d.hpp
AlexeyMerzlyakov, line 1: Here also could be moved `bresenham3D()`/`raytraceLine3D()` methods, for unification: they are currently using by VoxelGrid costmap layer. File in this case might be renamed nav2_util/include/nav2_util/raytrace_line_2d.hpp` -> `nav2_util/include/nav2_util/raytrace_line.hpp

nav2_map_server/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ find_package(tf2 REQUIRED)
 find_package(nav2_util REQUIRED)
 find_package(GRAPHICSMAGICKCPP REQUIRED)
 
+pkg_search_module(UUID REQUIRED uuid)
```
SteveMacenski, line 19: package.xml dependency?

nav2_map_server/CMakeLists.txt
```diff
@@ -40,6 +44,11 @@ add_executable(${map_saver_server_executable}
 add_executable(${costmap_filter_info_server_executable}
   src/costmap_filter_info/main.cpp)
 
+add_executable(${vector_object_server_executable}
```
SteveMacenski, line 47: No component?

nav2_util/include/nav2_util/occ_grid_utils.hpp
```diff
@@ -0,0 +1,129 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Author: Eitan Marder-Eppstein
+//         David V. Lu!!
+//         Alexey Merzlyakov
+
+#ifndef NAV2_UTIL__OCC_GRID_UTILS_HPP_
+#define NAV2_UTIL__OCC_GRID_UTILS_HPP_
+
+#include "nav_msgs/msg/occupancy_grid.hpp"
```
SteveMacenski, line 41: I would really prefer not to propagate more occupancy grid dependent things outside of the costmap and map server packages. Its in the roadmap to replace costmap implementations and abstracting this out makes that a bit more complicated. I'd prefer if that abstraction was reverted back into the costmaps

nav2_util/include/nav2_util/raytrace_line_2d.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+#ifndef NAV2_UTIL__RAYTRACE_LINE_2D_HPP_
+#define NAV2_UTIL__RAYTRACE_LINE_2D_HPP_
+
+#include <limits.h>
+#include <algorithm>
+#include <cmath>
+
+namespace nav2_util
+{
```
SteveMacenski, line 44: Ditto with all of this; I expect the future data structure we use will have a better implemented raytracing method that I would rather not keep hanging around after we refactor and remove costmap2D

nav2_msgs/msg/PolygonVO.msg
```diff
@@ -0,0 +1,5 @@
+std_msgs/Header header
+geometry_msgs/Point32[] points
+bool closed
+int8 value
+unique_identifier_msgs/UUID uuid
```
SteveMacenski, line 5: Move uuid fields right after `header` in both messages. Just a formatting thing

nav2_msgs/msg/CircleVO.msg
```diff
@@ -0,0 +1,6 @@
+std_msgs/Header header
```
SteveMacenski, line 1: The naming here I think can be more clear potentionally for the polygon/circle.

`CircleObject` `PolygonObject` `CircleObstacle` `PolygonObstacle` `GenericObstacle`, any thoughts? `VO` I don't think is necessarily all that clear

nav2_msgs/srv/RemoveShapes.srv
```diff
@@ -0,0 +1,5 @@
+# Remove vector objects from map
+
+unique_identifier_msgs/UUID[] uuids
```
SteveMacenski, line 3: Have a bool for `all_objects` so you don't specify uuid, just removes them all

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+  map_pub_ = create_publisher<nav_msgs::msg::OccupancyGrid>(
+    map_topic_,
+    rclcpp::QoS(rclcpp::KeepLast(1)).transient_local().reliable());
+
+  // Make name prefix for services
+  const std::string service_prefix = get_name() + std::string("/");
+
+  add_shapes_service_ = create_service<nav2_msgs::srv::AddShapes>(
+    service_prefix + std::string("add_shapes"),
```
SteveMacenski, line 62: `~/add_shapes` does the same thing without needing `  const std::string service_prefix = get_name() + std::string("/");`

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+nav2_util::CallbackReturn
+VectorObjectServer::on_activate(const rclcpp_lifecycle::State & /*state*/)
+{
+  RCLCPP_INFO(get_logger(), "Activating");
+
+  map_pub_->on_activate();
+
+  // Creating bond connection
+  createBond();
```
SteveMacenski, line 84: This needs to be the last thing in the function for timeout reasons

nav2_map_server/include/nav2_map_server/vector_object_shapes.hpp
```diff
@@ -0,0 +1,456 @@
...
+   */
+  bool getShapeUUID(const std::string & shape_name, unsigned char * out_uuid);
+
+  /**
+   * @brief Supporting routine obtaining ROS-parameters for the given vector object.
+   * Empty virtual method intended to be used in child implementations
+   * @param shape_name Name of the shape
+   * @return True if all parameters were obtained or false in failure case
+   */
+  virtual bool getROSParameters(const std::string & shape_name) = 0;
```
SteveMacenski, line 80: All methods of this name: all the parameters are "ROS" parameters, so that is redundant

SteveMacenski (commented):
This is alot of code to review, I really didn't have a chance yet to go through the cpp files in much detail, but I didn't want to delay the few comments I have now another few days while I continue to review the rest (its a big PR).

Higher level comment that this will really need alot of documentation on Navigation.ros.org to support (e.g. how to use tutorial / demo) and its missing test coverage

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+  switchMapUpdate();
+}
+
+bool VectorObjectServer::getROSParameters()
+{
+  // Main ROS-parameters
+  map_topic_ = getROSParameter(
+    shared_from_this(), "map_topic", "vo_map").as_string();
+  frame_id_ = getROSParameter(
```
SteveMacenski, line 521: make `global_frame_id_` + the variable name too

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+    shared_from_this(), "frame_id", "map").as_string();
+  resolution_ = getROSParameter(
+    shared_from_this(), "resolution", 0.05).as_double();
+  default_value_ = getROSParameter(
+    shared_from_this(), "default_value", nav2_util::OCC_GRID_UNKNOWN).as_int();
+  overlay_type_ = static_cast<OverlayType>(getROSParameter(
+      shared_from_this(), "overlay_type", static_cast<int>(OverlayType::OVERLAY_SEQ)).as_int());
+  update_frequency_ = getROSParameter(
```
SteveMacenski, line 529: Is there any logic to skip updates if all of the objects are in the map frame so its static and no new shapes were added?

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+{
+  // Initialize result with true. If one of the required vector object was not added properly,
+  // set it to false.
+  response->success = true;
+
+  auto node = shared_from_this();
+
+  // Process polygons
+  for (auto req_poly : request->polygons) {
+    nav2_msgs::msg::PolygonVO::SharedPtr new_params =
```
SteveMacenski, line 349: What's with the copy?

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+      std::make_shared<nav2_msgs::msg::PolygonVO>(req_poly);
+
+    bool new_shape = true;  // Whether to add a new shape
+    std::shared_ptr<Polygon> polygon;
+    for (auto shape : shapes_) {
```
SteveMacenski, line 354: If the shapes were stored in an unordered map hashed by the uuid, then this would be a trivial lookup

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+  if (size_x < 0) {
+    throw std::runtime_error("Incorrect map x-size");
+  }
+  if (size_y < 0) {
+    throw std::runtime_error("Incorrect map y-size");
+  }
+
+  if (!map_) {
+    map_ = std::make_shared<nav_msgs::msg::OccupancyGrid>();
```
SteveMacenski, line 206: Perhaps some of this difficulty is because you're working with `nav_msgs::msg::OccupancyGrid` instead of the `nav2_costmap_2d::Costmap2D`? I believe that has the constructor to take in an Occ Grid + can be converted back out to an occ grid easily. That would give you access to all the convert world to map coordinates and 2D raytracing functions for free without need to put them into nav2_utils since its being used here too

SteveMacenski: This is tied to the costmap2D layer anyway, so I don't think its at all a problem to use costmap2d here too

nav2_map_server/src/vo_server/vector_object_shapes.cpp
```diff
@@ -0,0 +1,579 @@
...
+bool Shape::getShapeUUID(const std::string & shape_name, unsigned char * out_uuid)
+{
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+
+  try {
+    // Try to get shape UUID from ROS-parameters
+    std::string uuid_str = getROSParameter(
```
SteveMacenski, line 56: What if this is a shape set by a service call?

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: The idea here is to make vector object server as an OccGrid map server for flexibility, e.g. for having used not only in Costmaps2D, but also in mind to have an ability to apply vector objects to/over OccupancyGrid-s. If we will move this to Costmap2D, there is not sense to keep vector object server inside map server package. So, if we want to use costmaps only, it should be moved in there.

Another question, is that OccGrid-s are being published and read by Costmap as standard message, that is already compatible with Costmap Filters (just subscribe to the OccGrid masks, published by VO-server, so it seem to me OK). If we want to switch to costmaps, we need to have a VO-node instead, embedded inside all necessary layers, and Costmap Filters code modification is needed. Is this the way we want to have it (it seem that initially we've agreed to have a separate VO-server, instead of embedding node).

Anyway, this is not a problem to change it to be embedded VO-node inside Costmaps2D, but I think, I need to understand an intention to be on the same page.

nav2_util/include/nav2_util/raytrace_line_2d.hpp
AlexeyMerzlyakov: I am not sure, that understand it well. Could you please give more context about replacing of Costamp2D? Is is planned to remove&replace all algorithms related to costmaps (including this VO-server/VO-node contribution)?

AlexeyMerzlyakov:
@SteveMacenski, thank you for having time to review the PR. I'll prepare the changes meet the review comments soon. Only the moving from OccGrid maps -> to Costmap2D is still doubtful for me, so I've left the responds for a few review items from above.

nav2_util/include/nav2_util/raytrace_line_2d.hpp
SteveMacenski: This is still very TBD so I don't have alot of details to share, but Costmap2D is a real rat's nest that needs to be reworked from the ground up with modern data structures and formatting to be more understandable and controllable. Also, it would be great to simultaneously think about AI methods and other data other than `cost` that users might want to store. So, a rework is planned as my principal project of 2024, but concrete details aren't available yet since its not 2024 and I haven't started thinking about the finite details in ... well detail :laughing: 

The VO and filters work will be ported over to work with the new system, those are useful / important capabilities. There will be no loss in capability in the migrartion; only expansion. But it will start with a fresh package rather than building on costmap2D

nav2_map_server/src/vo_server/vector_object_server.cpp
SteveMacenski: >  for having used not only in Costmaps2D, but also in mind to have an ability to apply vector objects to/over OccupancyGrid-s

Its still published out as an `OccupancyGrid` message, but you can take advantage that the Costmap2D object internally that has the same kind of data and alot of useful utilities that you can leverage (e.g. raytracing, world to map conversions, default values, etc) rather than abstracting that out or storing the same kinds of information as independent objects. 

>  If we will move this to Costmap2D, there is not sense to keep vector object server inside map server package.

I would agree. The map server location wasn't my first thought about where this would live. 

> If we want to switch to costmaps, we need to have a VO-node instead, embedded inside all necessary layers, and Costmap Filters code modification is needed. Is this the way we want to have it (it seem that initially we've agreed to have a separate VO-server, instead of embedding node)?

I think my remark above clarifies that. Use `nav2_costmap_2d::Costmap2D internal to the node in order to create the occupancy grid of the shapes. Then convert to the `OccupancyGrid` message type for publishing

AlexeyMerzlyakov: > Its still published out as an OccupancyGrid message, but you can take advantage that the Costmap2D object internally

OK, got it. However, this will cause the Nav2 to do unnecessary job twice. First time, we will convert costmaps -> to OccupancyGrid maps before publishing by VO-server. Then, if it will be used by costmaps, so it is being converted back to Costmaps by Keepout Filter (or Speed Filter or something else).

Moreover, Costmaps have no messages interfaces, while `OccupancyGrid.msg` is a far old-known standard, everyone works with it well. This is the same way, as we've used OccGrids for filter masks, to exist along with maps.

In this way, VO-server having the concept to be an independent module, working with vector objects and OccupancyGrids, and nothing common with Costmaps. All responsibilities for costmaps are taking by costmaps (Costmap Filter layers). If they will be re-worked, VO-server could be untouched - I see no problem with it.
 So I am still not sure, that using Costmaps for internal representation of what will be converted to Occupancy Grids - is a right idea or not?

> I would agree. The map server location wasn't my first thought about where this would live.

Initially, I've developing VO-server inside Costmap2D package. But after I've realized that there is nothing, that keeps it to work with Costmaps (please refer to the prev. sentence discourse), I've decided to move it to map_server package.
If we will decide to switch to Costmaps, there is no problem to move it back to `nav2_costmap_2d`

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Yes, it is switched be the following comparison method:
```
bool VectorObjectServer::isMapUpdate()
{
  for (auto shape : shapes_) {
    if (shape->getFrameID() != global_frame_id_ && !shape->getFrameID().empty()) {
      return true;
    }
  }

  return false;
}
```

If shape's `frame_id` differs from `global_frame_id_` and shape has it (empty frame id for the shape is treated to be equal to global frameID by-default), we are publishing map dynamically. Otherwise - statically once (per each shape update/remove call).

nav2_map_server/src/vo_server/vector_object_shapes.cpp
AlexeyMerzlyakov: During the creation of new Polygon/Circle shape, parameters are being passed as an input argument from a service call (here is [construction](https://github.com/AlexeyMerzlyakov/navigation2/blob/52d2c12628f3c3d2ef8225fe97b0a9fe9805a5db/nav2_map_server/src/vo_server/vector_object_server.cpp#L390) of new Polygon).
Inside both Polygon and Circle constructors, if UUID was specified by developer in service request, we just passing it to shape' parameters, as `params_ = params;`. If not set, [generating](https://github.com/AlexeyMerzlyakov/navigation2/blob/52d2c12628f3c3d2ef8225fe97b0a9fe9805a5db/nav2_map_server/src/vo_server/vector_object_shapes.cpp#L100) a new UUID (which could be obtained later by calling `get_shapes` service).

During the update of existing Polygon/Circle, UUID is being [searched](https://github.com/AlexeyMerzlyakov/navigation2/blob/52d2c12628f3c3d2ef8225fe97b0a9fe9805a5db/nav2_map_server/src/vo_server/vector_object_server.cpp#L355) (I'll try to replace it with unordered map, in accordance with comment from above) by requested by service call UUID.

SteveMacenski: This seems a little convoluted. to add new virtual parameters for incoming shapes rather than storing the attributes of the shape within the shape itself (and params only gotten for the shapes set in files on initialization -- then no more ROS parameters in the system).

nav2_map_server/src/vo_server/vector_object_server.cpp
SteveMacenski: I'm not saying don't publish to the costmap filter as a OccupancyGrid message, I'm asking you to use Costmap2D internally to this server to do operations so that you don't need to abstract out the map2world, world2map, getData, raycasting in 2D/3D and other functions into nav2_util which I do not want expanded outside of costmap2D given the road to come. 

The Costmap2D object contains all of these nice utilities which you'd like to use as a cost-map object. It seems purpose made for what you want to do. A good chunk of your parameters in the server are also natural constructor arguments for it 

```

  map_topic_ = getROSParameter(
    shared_from_this(), "map_topic", "vo_map").as_string();
  frame_id_ = getROSParameter(
    shared_from_this(), "frame_id", "map").as_string();
  resolution_ = getROSParameter(
    shared_from_this(), "resolution", 0.05).as_double();
  default_value_ = getROSParameter(
    shared_from_this(), "default_value", nav2_util::OCC_GRID_UNKNOWN).as_int();
  transform_tolerance_ = getROSParameter(
    shared_from_this(), "transform_tolerance", 0.1).as_double();
```

nav2_map_server/src/vo_server/vector_object_shapes.cpp
AlexeyMerzlyakov: Hm... It already works like you've mentioned: these parameters are not virtual. Abstract class `Shape` does not contain any of shape parameters. Parameters are already belonging to each shape. For `Polygon` shape class it is:
```
class Polygon : public Shape
{
...
protected:
  nav2_msgs::msg::PolygonVO::SharedPtr params_;
};
```
And for `Circle` shape class it is:
```
class Circle : public Shape
{
...
protected;
  nav2_msgs::msg::CircleVO::SharedPtr params_;
};
```

> and params only gotten for the shapes set in files on initialization -- then no more ROS parameters in the system

This is also how it already works: ROS-parameters are being set during initialization stage inside `Polygon::getROSParameters()` and `Circle::getROSParameters()`. Aha, these methods could be called from polygon/circle constructor. I've got it, if this is what you mean in this review comment.

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: In the `AddShapes.srv` request, there are `std::vector` of polygon and circle parameters:
```
CircleObject[] circles
PolygonObject[] polygons
```
Parameters in the `Polygon` or `Circle` C++ classes - are shared pointer to the object of this type. At this stage, the object will be created and copied to `new_params` smart pointer. However, at the later setting parameters stage (`polygon->setParams(new_params);` / `circle->setParams(new_params);`) these objects will be already created, so we need to have only one object copy, you've pointed out.

Indeed, we could do without copying (0-copy). However, in this case, we need to use usual `PolygonObject *` pointer in all interfaces instead of smart pointer, since smart pointer at the creation stage require to do a object copy. I do not really want to have usual pointers in the code, and performance overheads to do this copy are not extremely large, it is used to go the way as written above.

nav2_map_server/CMakeLists.txt
AlexeyMerzlyakov: [Added](https://github.com/ros-planning/navigation2/pull/3930/commits/d6a831d90d1440436538a61c771f017bddb59422#diff-a56612c357ec64656195abe672ad17eae27b8239e1f0d6067bf4ed925d7fdc9c) missed dependency

nav2_msgs/msg/PolygonVO.msg
AlexeyMerzlyakov: Done for [CircleObject.msg](https://github.com/ros-planning/navigation2/pull/3930/commits/d6a831d90d1440436538a61c771f017bddb59422#diff-455a4f390eb3e15399b91e734fd767cd265c8f21b9736f4cc3694468b46f1acd) and [PolygonObject.msg](https://github.com/ros-planning/navigation2/pull/3930/commits/d6a831d90d1440436538a61c771f017bddb59422#diff-69588642e05a866e11ca8c89c7c583cb4c4eef4ca26f25f71b4b1fa5180ffbdc)

nav2_msgs/srv/RemoveShapes.srv
AlexeyMerzlyakov: [Added](https://github.com/ros-planning/navigation2/pull/3930/commits/d6a831d90d1440436538a61c771f017bddb59422)

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3930/commits/d6a831d90d1440436538a61c771f017bddb59422

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Moved `createBond()` to be the last one in `on_activate()` and `destroyBond()` to be first one in `on_deactivate()` in https://github.com/ros-planning/navigation2/pull/3930/commits/d6a831d90d1440436538a61c771f017bddb59422

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/commit/d6a831d90d1440436538a61c771f017bddb59422

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Nice item for optimization. Thank you for pointing this out! Made `shapes_` to be `std::unordered_map` in https://github.com/ros-planning/navigation2/commit/d6a831d90d1440436538a61c771f017bddb59422

nav2_map_server/src/vo_server/vector_object_shapes.cpp
AlexeyMerzlyakov: Afte code double-checking, `Polygon`/`Circle` parameters initialization / setting were simplified:
Now the `Polygon`/`Circle` constructor is almost empty. Two routines are responsible for shape/parameters setting:

- `obtainParams(name)` - used to obtain ROS-parameters. Called once during ROS-node configuring.
- `setParams(params)` - used to set ROS-parameters during add/update shape service call.

The changes were reflected in https://github.com/ros-planning/navigation2/pull/3930/commits/f1fe2479025920c4927c4931ea9fd9f01910a93b

nav2_util/include/nav2_util/raytrace_line_2d.hpp
AlexeyMerzlyakov: After latest Nav2 WG discussion, there is no need to move 3D-version of `raytraceLine3D()` from Voxel Grid: it could be re-worked in foreseen future.

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Since last Nav2 WG, it was decided to keep VO-server to be based on OccupancyGrids to not make double Costmap2D->OccGrid->Costmap2D conversion and thus not loose performance and precision.
Also, it was decided to return back `getMaskData()` routine to `nav2_costmap_2d` as it used only there (and there is no sense to doubling the code), which was made in https://github.com/ros-planning/navigation2/pull/3930/commits/d6b9cf123c5bb1a26ab4ffbbed22356d402dcb85

nav2_map_server/CMakeLists.txt
AlexeyMerzlyakov: Added composition node support in https://github.com/ros-planning/navigation2/pull/3930/commits/5cfde7ce3ff11deeeaef2b0d163a94fe95d44384

nav2_costmap_2d/plugins/costmap_filters/costmap_filter.cpp
```diff
@@ -189,29 +189,6 @@ bool CostmapFilter::transformPose(
   return true;
 }
 
-bool CostmapFilter::worldToMask(
```
SteveMacenski, line 192: I think we discussed moving this back since its only used in the filters anyway?

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,565 @@
...
+    }
+  } catch (const std::exception & ex) {
+    RCLCPP_ERROR(get_logger(), "Can not upate map: %s", ex.what());
+    return;
+  }
+
+  publishMap();
+}
+
+bool VectorObjectServer::isMapUpdate()
```
SteveMacenski, line 307: Move this into `switchMapUpdate` please. Mostly because `isMapUpdate` and `updateMap` / `processMap` are so related in naming its a little confusing at first glance. Could rename this, but this is simple enough to be put in `switchMapUpdate` and entirely disambiuate

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,565 @@
...
+}
+
+void VectorObjectServer::processMap()
+{
+  if (!process_map_) {
+    return;
+  }
+
+  try {
+    if (shapes_.size()) {
```
SteveMacenski, line 287:
```suggestion
    if (shapes_.size() > 0) {
```

I know that technically works, but I don't like casting numerical values to bool

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,565 @@
...
+  // Initialize result with true. If one of the required vector object was not found,
+  // set it to false.
+  response->success = true;
+
+  if (request->all_objects) {
+    // Clear all objects
+    for (auto shape : shapes_) {
+      shape.second.reset();
+    }
+    shapes_.clear();
```
SteveMacenski, line 469: clear() should reset all the shared pointers, no?

nav2_map_server/params/vector_object_server_params.yaml
```diff
@@ -0,0 +1,31 @@
...
+      value: 100
+      points: [0.3, 0.3, 0.3, -0.3, 0.0, -0.3, 0.0, 0.3]
+    CircleA:
+      type: "circle"
+      frame_id: "map"
+      fill: True
+      value: 10
+      center: [3.0, 3.0]
+      radius: 0.5
+      uuid: "7b3f3d7d-135c-4b6c-aca1-7a84d1050505"
```
SteveMacenski, line 24: I didn't think about this before, but if we're asking users to specify shapes in a params file, UUID might not be the best option. That would require them to offline generate some UUIDs to put into the params file. If we instead just made it an `id` field of `unsigned int` then it would be easier to enumerate them in a config file.

When its generated via code, UUID makes sense, but not when manually requiring annotation

SteveMacenski (changes requested):
Needs tests / docs :smile:

nav2_map_server/src/vo_server/vector_object_shapes.cpp
SteveMacenski: I leave that to your discretion

nav2_map_server/params/vector_object_server_params.yaml
AlexeyMerzlyakov: BTW, `uuid` ROS-param is optional, it could be skipped by developer: in this case it will be generated to a new shape automatically. In this case, does it still have a sense to have an ID?

I'm asking, because when we have an ID, we do not need to support UUID anymore: any shape could be uniquely addressed by its ID, it also could be easily generated if not specified w/o any UUID dependency. So we could just skip it.

What do you think, maybe it is better to totally refuse from UUID usage in favor of ID?
Another question that logically raises here: how about to use ID of string type, which could be called "name"? It is even much more user-friendly, than unsigned ints, I believe; and we do not need to worry about ID-s/UUID-s in ROS-parameters: name of each shape is already specified in its parameters path and `shapes` array.

nav2_costmap_2d/plugins/costmap_filters/costmap_filter.cpp
AlexeyMerzlyakov: It is used in VO-server in different places, like [here](https://github.com/AlexeyMerzlyakov/navigation2/blob/aed5df9e2671101dd87d7ff1b714984f4d885ee7/nav2_map_server/src/vo_server/vector_object_server.cpp#L244) or [here](https://github.com/AlexeyMerzlyakov/navigation2/blob/aed5df9e2671101dd87d7ff1b714984f4d885ee7/nav2_map_server/src/vo_server/vector_object_shapes.cpp#L218C10-L218C10), so if we want to have this routine in costmaps, we also need to duplicate it for map server. So, I'd like to switch in favor of using it in `nav2_util` instead (in order to avoid boilerplate works in the future by keeping synched duplicating parts of code in different places).

nav2_map_server/params/vector_object_server_params.yaml
AlexeyMerzlyakov: However, from my point of view, using UUID is still the nice way for UX:
- We could optionally specify it in ROS-parameters
- We could optionally specify it in `AddShapes.srv` service request when adding a new shape
- And we to specify it on mandatory basis when modifying existing shape while calling `AddShapes.srv` and removing some specific shape in `RemoveShape.srv`

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Yep, this is not needed: at `clear()` stage, all unused smart pointers inside to be destroyed. Fixed in https://github.com/ros-planning/navigation2/pull/3930/commits/1bb10dcbb2b00962f896bd46b81d1a9070846bbb

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Done in https://github.com/ros-planning/navigation2/pull/3930/commits/888bdbba2a4d465f6c37034b0432e6ddaf13a74c

nav2_map_server/params/vector_object_server_params.yaml
SteveMacenski: >  uuid ROS-param is optional, it could be skipped by developer

Ah, ok! I'd just make sure to document that in the param guide that its optional except if you might want to remove it, in which case specifying it is important.

nav2_map_server/include/nav2_map_server/vector_object_shapes.hpp
```diff
@@ -0,0 +1,426 @@
...
+    const OverlayType overlay_type);
+
+  /// @brief Input circle parameters (could be in any frame)
+  nav2_msgs::msg::CircleObject::SharedPtr params_;
+  /// @brief Circle center in the map's frame
+  geometry_msgs::msg::Point32::SharedPtr center_;
+};
+
+/// @brief Functor class used in raytraceLine algorithm
+class MapAction
```
SteveMacenski, line 391: In vector object shapes file?

SteveMacenski: Isn't this more a util?

nav2_map_server/src/vo_server/vector_object_server.cpp
```diff
@@ -0,0 +1,551 @@
...
+void VectorObjectServer::switchMapUpdate()
+{
+  for (auto shape : shapes_) {
+    if (shape.second->getFrameID() != global_frame_id_ && !shape.second->getFrameID().empty()) {
+      if (!map_timer_) {
+        map_timer_ = this->create_timer(
+          std::chrono::duration<double>(1.0 / update_frequency_),
+          std::bind(&VectorObjectServer::processMap, this));
+      }
+      RCLCPP_INFO(get_logger(), "Publishing map dynamically");
```
SteveMacenski, line 313: Update to mention what rate its publishing at, that's useful debug info

nav2_map_server/src/vo_server/vector_object_shapes.cpp
```diff
@@ -0,0 +1,588 @@
...
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+
+  try {
+    // Try to get shape UUID from ROS-parameters
+    std::string uuid_str = getParameter(
+      node, shape_name + ".uuid", rclcpp::PARAMETER_STRING).as_string();
+    if (uuid_parse(uuid_str.c_str(), out_uuid)) {
```
SteveMacenski, line 58: if false? Does this return an int or something?

AlexeyMerzlyakov:
Regarding https://github.com/ros-planning/navigation2/pull/3930#discussion_r1384316191, Ive missed it initially, but currently during the testcase development it was discovered that switching to `std::unordered_map`broke the `OVERALY_SEQ` updating mechanism, where the shapes are put on map in the same order, as they arrived (independently on their values). This sequential updating mechanism is important to support some use-cases, such as outer ring with inner objects inside this ring, which was reported at initial ticket.

To do this, there is a way to add `unsigned int seq` field to the `Polygon` and `Circle` classes, but I think it will overcomplicate the overall structure.
So, unfortunately, better way - is roll-back usage of `std::unordered_map` in favor of `std::vector`-s where we could save the order of shapes arrived.

SteveMacenski:
> where the shapes are put on map in the same order, as they arrived (independently on their values).

Ohhhhhhhhhh yeah, that sounds like a problem :(

nav2_map_server/include/nav2_map_server/vector_object_shapes.hpp
AlexeyMerzlyakov: Moved to `vector_object_utils.hpp` in https://github.com/ros-planning/navigation2/pull/3930/commits/43af6df733237b6f5ef8c91d0757b2ceda43a2de

nav2_map_server/src/vo_server/vector_object_server.cpp
AlexeyMerzlyakov: Fixed in https://github.com/ros-planning/navigation2/pull/3930/commits/43af6df733237b6f5ef8c91d0757b2ceda43a2de

nav2_map_server/src/vo_server/vector_object_shapes.cpp
AlexeyMerzlyakov: `uuid_parse()` returns 0 in normal case and non-zero in case of failure (according to documentation, failure value is -1). Switched to comparison with 0 in https://github.com/ros-planning/navigation2/pull/3930/commits/43af6df733237b6f5ef8c91d0757b2ceda43a2de

AlexeyMerzlyakov:
Not a problem. This is some kind of working process while making better contribution. Restored back vectors usage in https://github.com/ros-planning/navigation2/pull/3930/commits/43af6df733237b6f5ef8c91d0757b2ceda43a2de.

There are also tests covering new functionality were added in https://github.com/ros-planning/navigation2/pull/3930/commits/355b95bf23e92819c15adb421e96b3055bd09a90. What is need ToGo - it is documentation (MG, Configuration pages, tutorial).

nav2_map_server/src/vo_server/vector_object_shapes.cpp
AlexeyMerzlyakov: Finally moved to `nav2_util` in 6d04b15af60e7089f9ad771b77c95061ee26bad7 to avoid code duplication (fortunately, it could be easily combined with existing code by using structure templating in `isPointInsidePolygon()` function)

AlexeyMerzlyakov:
Everything is set to be completed for today:

- All review comments were met
- Testcases developed. Code coverage for Vector Object server is around `~97%` which is fine:
codecov_report_1
And aligns with overall Map Server code coverage:
codecov_report_2
- Documentation update made in https://github.com/ros-planning/navigation.ros.org/pull/496

PranavShevkar:
Hello @AlexeyMerzlyakov,
I would like to use this branch but the build keeps on failing. Could you please add the instructions to make it work. Thanks

AlexeyMerzlyakov:
Force-pushed the branch resolving conflicts with latest main. @SteveMacenski, do we plan to merge this PR or this feature is no more necessary?

PranavShevkar:
Tested on Rolling. Looks good and works fine

AlexeyMerzlyakov:
Resolved small merge conflict and force-pushed development branch.
@SteveMacenski, are we ready to merge this change (from my side, everything: change, tests, docs - are ready)?

ajtudela:
Bump this!

This server would be very beneficial for one our current use cases. Let me know if I can help to wrap this up.

SteveMacenski:
This is functionally on permanent hold as Alexey is no longer working on this project in a professional capacity. If someone wants to pick this up, test it, and resubmit a new PR with fixes made, we're happy to review!

ajtudela:
@SteveMacenski I want to finish this PR if Alexey can't because we use it in our system. Do I make a new PR with this exact commits and we go from there?

SteveMacenski:
Sure, I'd make sure to go through the ticket and the review comments here to understand the remaining efforts required

SteveMacenski:
https://github.com/ros-navigation/navigation2/pull/4680 to supersede

Issue #3931, opened 2023-11-02
navfn_planner GoalOccupied exception


ladianchad:
## Bug report

**Required Info:**

- Operating System:
  - 22.04
- ROS2 Version:
  - humble
- Version or commit hash:
  - source da7f85d
- DDS implementation:
  - rmw_cyclone

#### Steps to reproduce issue
send goal like below image
![Frame 4 (2)](https://github.com/ros-planning/navigation2/assets/60467877/1b6f2f45-96a4-4031-babf-2dee46065913)

#### Expected behavior
navfn throw GoalOccupied

#### Actual behavior
navfn palnner not throw GoalOccupied
#### Additional information

----

SteveMacenski:
I don't understand this ticket. Please provide detail about the error, what you expect to happen, what is happening, and what behavior you expect. This is not verbose enough for me to understand what the issue is.

SteveMacenski:
Closing due to lack of response and PR doesn't seem to solve a problem that I can identify. Feel free to comment here or on the PR and we can continue.

PR #3932, opened 2023-11-02
fix navfn planner error

ladianchad:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3931 |
| Primary OS tested on | ubuntu |
| Robotic platform tested on | local  pc |

---

add case throw plan error.

---

#### For Maintainers: 

### DIFF
--- a/nav2_navfn_planner/src/navfn_planner.cpp
+++ b/nav2_navfn_planner/src/navfn_planner.cpp
@@ -346,6 +346,10 @@ NavfnPlanner::makePlan(
         "Failed to create a plan from potential when a legal"
         " potential was found. This shouldn't happen.");
     }
+  } else {
+    throw nav2_core::GoalOccupied(
+            "Goal Coordinates of(" + std::to_string(goal.pose.position.x) + ", " +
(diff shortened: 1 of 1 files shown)


nav2_navfn_planner/src/navfn_planner.cpp
```diff
@@ -346,6 +346,10 @@ NavfnPlanner::makePlan(
         "Failed to create a plan from potential when a legal"
         " potential was found. This shouldn't happen.");
     }
+  } else {
+    throw nav2_core::GoalOccupied(
```
SteveMacenski, line 350: This shouldn't be necessary?

`return !plan.poses.empty();` causes NoPathCanBeFound exception thrown https://github.com/ros-planning/navigation2/blob/933f4a9cfcae3ebd2860e057172910b8d23360c5/nav2_navfn_planner/src/navfn_planner.cpp#L193-L194 since none of the path is populated.

SteveMacenski:
Closing due to lack of response and PR doesn't seem to solve a problem that I can identify. Feel free to comment here or on the ticket and we can continue.

Issue #3089, opened 2022-07-23
RPP cusp not detected for in-place rotation (for reversing)


automech-rb:
## Bug report

**Required Info:**

- Operating System:
  - Docker 20.04 Ubuntu
- ROS2 Version:
  - rolling
- Version or commit hash:
  - latest 104bdc51fb72e05af4ccb4fb96a2fd421239f016
- DDS implementation:
  - rmw_cyclonedds_cpp

#### Steps to reproduce issue
Use SMAC-lattice planner with **in-place rotation on**
and set **use_rotate_to_heading: false** and **allow_reversing: true** in RPP

```
  FollowPath:
      plugin: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
      desired_linear_vel: 0.8
      lookahead_dist: 0.4
      min_lookahead_dist: 0.4
      max_lookahead_dist: 0.8
      lookahead_time: 1.0
      rotate_to_heading_angular_vel: 0.5
      transform_tolerance: 0.1
      use_velocity_scaled_lookahead_dist: false
      min_approach_linear_velocity: 0.05
      approach_velocity_scaling_dist: 0.8
      max_allowed_time_to_collision_up_to_carrot: 1.0
      use_regulated_linear_velocity_scaling: true
      use_cost_regulated_linear_velocity_scaling: false
      regulated_linear_scaling_min_radius: 0.9
      regulated_linear_scaling_min_speed: 0.2
      use_rotate_to_heading: false
      allow_reversing: true
      rotate_to_heading_min_angle: 0.785 #45degree
      max_angular_accel: 0.5
      max_robot_pose_search_dist: 10.0
      use_interpolation: false
```
![image](https://user-images.githubusercontent.com/24780015/180596760-1b091ce8-7b7d-4a6a-9a28-9aa6ff7bd3b2.png)
The above figure shows a cusp with in-place rotation.
#### Expected behavior
The lookup should first be till the cusp and after robot arrives at cusp it should lookup to next point.

#### Actual behavior
Cusp is not detected and robot lookup directly crosses the cusp even when robot has not arrived to the cusp pose.

#### Additional information

[...]

PR #3934, opened 2023-11-02
[RPP] Adding support for rotate in place cusps

SteveMacenski:
@Aposhian  @vinnnyr  @doisyg  I could use a sanity check here as I'm proposing a solution to https://github.com/ros-planning/navigation2/issues/3089.

Previously, cusps that involve in-place rotations from the state lattice planner are ignored. This is an attempted solution which I don't think has any negative side effects, but its subtle and worth expanding upon for visibility and understandability.

Previously, we check the dot project is negative, meaning that we have a discontinuity in the path that is a direction change. Since in place rotations have the same position, the dot product is `0` which isn't triggered.  That's a problem.

So, I adjusted a new check if we're `0` and that the orientations of the poses are different -- that way if we simply have 2 overlapping exactly same points, this won't trigger (only true rotate in place represntations). Rather than parameterizing it, this **always triggers** with the assumption that if you're providing a feasible path with an in-place rotation, your robot can handle it since you gave the planner a motion model that can do it. Also in the case that the cusp direction changes (not just a 'going forward -> rotate -> going forward' still), you still want to identify the cusp. Its now that 'going forward -> rotate -> going forward' will **also** now trigger.

So now we've identified the cusp with rotate in place & general rotate in places. Cool. That'll impact the lookahead point selection as any cusp would that we know works, so no need to beat that analysis to death, other than to mention that the `min_by` finds the closes path point to the robot to prune all path points prior to.

To support this, I needed to change the comparison so that we find the **last, closest** point, instead of the **first, closest** point when there are N points the same distance away (e.g. rotate in place, same position). That way, we prune to the final path-pose in the rotation sequence, not the first.

In the ticket, we discussed adding a `shouldRotateAlongPath`, but actually I think that's not necessary. The `shouldRotateToPath` should actually implicitly do what we need to do. If rotate-to-path is valid and the error is sufficiently large, we do it. I think that covers our rotate to path heading needs for inplace rotations. When that rotation is very small - - then we don't really functionally need to do it here and it'll take care of itself in the immediate path tracking request curvature velocities. The same `use_rotate_to_heading` param applies. 

So, I think this fully resolves the issue - with relatively minor changes.  What do you think?

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -419,9 +419,24 @@ double RegulatedPurePursuitController::findVelocitySignChange(
     /* Checking for the existance of cusp, in the path, using the dot product
     and determine it's distance from the robot. If there is no cusp in the path,
     then just determine the distance to the goal location. */
-    if ( (oa_x * ab_x) + (oa_y * ab_y) < 0.0) {
+    const double dot_prod = (oa_x * ab_x) + (oa_y * ab_y);
+    if (dot_prod < 0.0) {
       // returning the distance if there is a cusp
       // The transformed path is in the robots frame, so robot is at the origin
+      return hypot(
+        transformed_plan.poses[pose_id].pose.position.x,
+        transformed_plan.poses[pose_id].pose.position.y);
+    } else if (
+      (hypot(oa_x, oa_y) == 0.0 &&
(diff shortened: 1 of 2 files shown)


Aposhian:
Sounds reasonable enough.

SteveMacenski:
@Aposhian do you see any regressions in this for your application? I want to make sure there aren't unintended consequences given your larger application

Aposhian:
Actually we aren't using RPP at the moment, so I can't say if it causes regressions for us.

PR #3893, opened 2023-10-22
Fix nav2_bringup flake8 tests proposal

nachovizzo:
Hello, I'm opening a PR instead of a ticket so the pipeline is also triggered and we can discuss further

## Problem: new  flake8 plugins in `ros:rolling` make the CI fail

This sounds pretty strange, but the reason is quite simple:

The CI is using [`ghcr.io/ros-planning/navigation2:main`](https://github.com/ros-planning/navigation2/blob/a1c9fd5ad29bb00e40ce6e696d899a2bcd50cde5/.circleci/config.yml#L485C23-L485C23
) as the base image, which is based on `rolling`.

By doing a simple docker run in such a container one can immediately see all the plugins that are installed there:
```
python3-flake8/now 4.0.1-2 all [installed,local]
python3-flake8-builtins/now 1.5.3-3 all [installed,local]
python3-flake8-comprehensions/now 3.8.0-1 all [installed,local]
python3-flake8-docstrings/now 1.6.0-1 all [installed,local]
python3-flake8-import-order/now 0.18.1-2 all [installed,local]
python3-flake8-quotes/now 3.3.1-2 all [installed,local]
ros-rolling-ament-cmake-flake8/now 0.16.0-1jammy.20231004.142753 amd64 [installed,local]
ros-rolling-ament-flake8/now 0.16.0-1jammy.20231004.141251 amd64 [installed,local]
```

In contrast, if one searches for the `flake8` apt packages installed in the `ros:iron` image (as an example) the output is only:
```
python3-flake8/now 4.0.1-2 all [installed,local]
ros-iron-ament-cmake-flake8/now 0.14.2-1jammy.20230908.160003 amd64 [installed,local]
ros-iron-ament-flake8/now 0.14.2-1jammy.20230908.154814 amd64 [installed,local]
```

## Why this happens

This is the thing I don't like much about `flake8`, the [plugin systems](https://flake8.pycqa.org/en/latest/plugin-development/index.html). As is the case now, I can `sudo apt install python3-flake8-quotes` and the output of the `flake8` linter is going to be different from the "vanilla" flake8.

## How to reproduce the CI failure locally?

Easy:
```sh
sudo apt install \
  python3-flake8-builtins \
  python3-flake8-comprehensions \
  python3-flake8-docstrings \
  python3-flake8-import-order \
  python3-flake8-quotes
```

And after that just run `ament_flake8 .` on the `navigation2` root, and you will see exactly the same errors that are on the CI.

## CI Test

[...]

PR #3941, opened 2023-11-05
Use ament_black as linter

nachovizzo:
## Use `ament_black` to format python code

@SteveMacenski I'm only opening this pre-PR to continue #3893. The reason is simple, most python scripts have many style errors according to `flake8` which can be automatically solved by using perhaps the most famous Python format, `black`. 

I have another branch where I'm fixing the missing linting tests, **BUT** those contain changes in the logic. The diff between main and a branch that fixes style errors + flake8 errors would be impossible to read/ follow.

This one is straightforward, I just did the following (on iron), 

```
git clone nav2...
cd nav2
ament_black --reformat .
```

The only prerequisite would be to use the linter is to install the https://github.com/botsandus/ament_black linter (Yes I'm the current maintainer as well)
```
sudo apt install ros-iron-ament-black # or sudo apt install ros-hubmle-ament-black
```

## About `rolling`

`rolling` is still waiting to get its `ament-black` package: https://github.com/ros/rosdistro/pull/38762 . In the meantime and as a temporary solution, I'm also maintaining a Python package that strictly follows the `rosdistro` package versioning and implementation, thus:

```
pip install ament-black
```

I wanted to keep this PR as straightforward as possible, so in future also the commit associated with this PR is PURELY style-change, therefore;
- [ ] If the maintainers agree to start using `black` (I'm sure many people would love to see this change coming to the stack) then I will open *another* PR, adding the linter as part of the CI tests

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/bringup_launch.py
+++ b/nav2_bringup/launch/bringup_launch.py
@@ -17,8 +17,12 @@ import os
 from ament_index_python.packages import get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import (DeclareLaunchArgument, GroupAction,
-                            IncludeLaunchDescription, SetEnvironmentVariable)
+from launch.actions import (
+    DeclareLaunchArgument,
+    GroupAction,
+    IncludeLaunchDescription,
+    SetEnvironmentVariable,
+)
 from launch.conditions import IfCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch.substitutions import LaunchConfiguration, PythonExpression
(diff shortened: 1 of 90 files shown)


SteveMacenski:
We follow the best practices of ROS 2 styling so we wouldn't be open to using another format other than what ROS 2 recommends so that Nav2 can be a "model" codebase of when all best practices are followed, here's an example of what a good codebase would look like.

If black is configured to use the same profile as ROS 2's flake8 config, then that would be really useful to have something that reformatted things. I guess I need some clarity about the intent of `ament_black` with flake8: are you seeing it as in addition to; instead of; or something that we can use to autoformat (perhaps even as part of the CI jobs having a bot push reformat updates automatically)?

nachovizzo:
@SteveMacenski I guess then `black` is not an option.. as a matter of fact, `black` will always format with double quotes (I'm not the only one against flake8-quotes)

That said, we will never manage to marry `black` and `flake8-quotes`, as long as this linter is part of the ros tooling, then black has (sadly) no room to play here.

Please confirm if you would like me to close this PR and rebase #3942 without the black changes

SteveMacenski:
If black can be reconfigured to use ament_flake8's configuration profile, having something that does auto reformatting would be great! But yes, we are following OSRF's coding style precisely. If additional rules are placed on things due to black that don't conflict with ROS 2's formatting, that's fine. But we will always take ROS 2's styling first.

nachovizzo:
Then good life to black :)

Understood and follow the ideal :+1:

PR #3950, opened 2023-11-09
Use native library haltTree()

doisyg:
---

## Basic Info

| ------ | ----------- |

---

As discussed, draft to open the discussion.

Update to use the the native cpp library `haltTree()`  method in `BehaviorTreeEngine::haltAllActions`.
https://github.com/BehaviorTree/BehaviorTree.CPP/blob/61c55ed720e0630f211686b3830942ae27e90501/include/behaviortree_cpp_v3/bt_factory.h#L162-L179

Before this change, only the active BT node was halted when finishing the action. After this change, all BT nodes `halt()` methods are called. This is very convenient to handle cleaning operation (switch off your lights when leaving) in halt().

Also updated `nav2_behavior_tree::BtActionServer::haltTree()` to use the same. It is used nowhere in nav2 but is useful for external users (like me) that want for instance to halt the tree on preemption.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/behavior_tree_engine.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/behavior_tree_engine.hpp
@@ -85,9 +85,9 @@ public:
 
   /**
    * @brief Function to explicitly reset all BT nodes to initial state
-   * @param root_node Pointer to BT root node
+   * @param tree Tree to halt
    */
-  void haltAllActions(BT::TreeNode * root_node);
+  void haltAllActions(BT::Tree &tree);
 
 protected:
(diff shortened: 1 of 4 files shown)


SteveMacenski (approved):
I see no downside - why a draft?

doisyg:
Switching to PR ;)

doisyg:
I guess a note on the migration guide ?

doisyg:
Docs: https://github.com/ros-planning/navigation.ros.org/pull/487

SteveMacenski:
Honestly not sure if that was worth a migration guide doc since its not actionable to users, but I'll always default to more docs over less docs :wink:

Issue #3920, opened 2023-10-30
Make BT Action, Cancel, and Service nodes have configurable wait times


SteveMacenski:
Currently default at `1s` hardcoded, which should be sufficient for any well engineered system. However, users have asked for this to be configurable and we should oblige

SteveMacenski:
- https://github.com/ros-planning/navigation2/blob/363b55af3a61ccdd2f429689a174e49fb7cf7523/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L96
- https://github.com/ros-planning/navigation2/blob/363b55af3a61ccdd2f429689a174e49fb7cf7523/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L80
- https://github.com/ros-planning/navigation2/blob/363b55af3a61ccdd2f429689a174e49fb7cf7523/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp#L92

embeddedadam:
@SteveMacenski I would like to take care of this ticket.

SteveMacenski:
Great!! Let me know if you have any questions!

SteveMacenski:
Done! Thanks @embeddedadam !

PR #3960, opened 2023-11-13
WIP: Make BT nodes have configurable wait times.

embeddedadam:
The previous solution provided a hardcoded 1s value.
Right now the value can be configured for BT
Action, Cancel, and Service nodes.

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3920 |
| Primary OS tested on | (Ubuntu)
| Robotic platform tested on | (None) |

---

Added configurable wait times for BT nodes.

Added a new parameter which is configurable with the time value.
Needs documentation of its meaning and usage.

(To consider):
Even the value can be provided from the outside and can be configured,
there should be also a default one being set if no one plans to configure this parameter,
so the old applications maintain the same way of using this software and it makes it more
robust in the case of unintended usage.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -61,8 +61,8 @@ public:
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
-    bt_wait_for_service_timeout_ =
-      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout_");
+    wait_for_service_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("wait_for_service_timeout_");
 
     // Initialize the input and output messages
     goal_ = typename ActionT::Goal();
@@ -95,7 +95,7 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    if (!action_client_->wait_for_action_server(bt_wait_for_service_timeout_)) {
+    if (!action_client_->wait_for_action_server(wait_for_service_timeout_)) {
       RCLCPP_ERROR(
         node_->get_logger(), "\"%s\" action server not available after waiting for 1 s",
         action_name.c_str());
@@ -465,7 +465,7 @@ protected:
   std::chrono::milliseconds bt_loop_duration_;
 
   // The timeout value for waiting for a service to response
-  std::chrono::milliseconds bt_wait_for_service_timeout_;
+  std::chrono::milliseconds wait_for_service_timeout_;
 
   // To track the action server acknowledgement when a new goal is sent
   std::shared_ptr<std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>>
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -62,7 +62,7 @@ public:
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
     wait_for_service_timeout_ =
-      config().blackboard->template get<std::chrono::milliseconds>("wait_for_service_timeout_");
+      config().blackboard->template get<std::chrono::milliseconds>("wait_for_service_timeout");
 
     // Initialize the input and output messages
     goal_ = typename ActionT::Goal();
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -61,8 +61,8 @@ public:
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
-    bt_wait_for_service_time_ =
-      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_time");
+    bt_wait_for_service_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout_");
 
     // Initialize the input and output messages
     goal_ = typename ActionT::Goal();
@@ -95,7 +95,7 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    if (!action_client_->wait_for_action_server(bt_wait_for_service_time_)) {
+    if (!action_client_->wait_for_action_server(bt_wait_for_service_timeout_)) {
       RCLCPP_ERROR(
         node_->get_logger(), "\"%s\" action server not available after waiting for 1 s",
         action_name.c_str());
@@ -465,7 +465,7 @@ protected:
   std::chrono::milliseconds bt_loop_duration_;
 
   // The timeout value for waiting for a service to response
-  std::chrono::milliseconds bt_wait_for_service_time_;
+  std::chrono::milliseconds bt_wait_for_service_timeout_;
 
   // To track the action server acknowledgement when a new goal is sent
   std::shared_ptr<std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>>
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -61,6 +61,8 @@ public:
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    bt_wait_for_service_time_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_time");
 
     // Initialize the input and output messages
     goal_ = typename ActionT::Goal();
@@ -93,7 +95,7 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    if (!action_client_->wait_for_action_server(1s)) {
+    if (!action_client_->wait_for_action_server(bt_wait_for_service_time_)) {
       RCLCPP_ERROR(
         node_->get_logger(), "\"%s\" action server not available after waiting for 1 s",
         action_name.c_str());
@@ -461,6 +463,9 @@ protected:
 
   // The timeout value for BT loop execution
   std::chrono::milliseconds bt_loop_duration_;
+
+  // The timeout value for waiting for a service to response
+  std::chrono::milliseconds bt_wait_for_service_time_;
 
   // To track the action server acknowledgement when a new goal is sent
   std::shared_ptr<std::shared_future<typename rclcpp_action::ClientGoalHandle<ActionT>::SharedPtr>>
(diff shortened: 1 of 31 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -61,6 +61,8 @@ class BtActionNode : public BT::ActionNodeBase
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    bt_wait_for_service_time_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_time");
```
SteveMacenski, line 65: How about `wait_for_service_timeout`?

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -59,6 +59,8 @@ class BtCancelActionNode : public BT::ActionNodeBase
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    bt_wait_for_service_time_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_time");
```
SteveMacenski, line 63: Where is this set on the blackboard :wink:

SteveMacenski (changes requested):
That's all right for the nodes themselves, but a question for you to search into: Where are variables set on the blackboard and how can we use that method to also set this timeout value so that users can control it? 

I know the answer, but its better for you to find it understand how the system works. If you get stumped let me know and I can point you in the right direction

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -61,6 +61,8 @@ class BtActionNode : public BT::ActionNodeBase
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    bt_wait_for_service_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout_");
```
SteveMacenski, line 65:
```suggestion
      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout");
```

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -172,6 +175,7 @@ bool BtActionServer<ActionT>::on_configure()
   blackboard_->set<rclcpp::Node::SharedPtr>("node", client_node_);  // NOLINT
   blackboard_->set<std::chrono::milliseconds>("server_timeout", default_server_timeout_);  // NOLINT
   blackboard_->set<std::chrono::milliseconds>("bt_loop_duration", bt_loop_duration_);  // NOLINT
+  blackboard_->set<std::chrono::milliseconds>("bt_wait_for_service_timeout_", bt_wait_for_service_timeout_);  // NOLINT
```
SteveMacenski, line 178: Ditto fix

nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
```diff
@@ -59,6 +59,8 @@ class BtCancelActionNode : public BT::ActionNodeBase
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    bt_wait_for_service_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout_");
```
SteveMacenski, line 63: Ditto fix

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -62,6 +62,8 @@ class BtServiceNode : public BT::ActionNodeBase
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    bt_wait_for_service_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout_");
```
SteveMacenski, line 66: Ditto fix

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -158,6 +158,9 @@ bool BtActionServer<ActionT>::on_configure()
   int default_server_timeout;
   node->get_parameter("default_server_timeout", default_server_timeout);
   default_server_timeout_ = std::chrono::milliseconds(default_server_timeout);
+  int bt_wait_for_service_timeout;
+  node->get_parameter("bt_wait_for_service_timeout", bt_wait_for_service_timeout);
```
SteveMacenski, line 162: Great! I think this needs to be added as a parameter in the parameter  guide for the BT Navigator server and that should be good to go! Here's the file: https://github.com/ros-planning/navigation.ros.org/blob/master/configuration/packages/configuring-bt-navigator.rst

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -235,6 +239,7 @@ bool BtActionServer<ActionT>::loadBehaviorTree(const std::string & bt_xml_filena
       blackboard->set<rclcpp::Node::SharedPtr>("node", client_node_);
       blackboard->set<std::chrono::milliseconds>("server_timeout", default_server_timeout_);
       blackboard->set<std::chrono::milliseconds>("bt_loop_duration", bt_loop_duration_);
+      blackboard->set<std::chrono::milliseconds>("bt_wait_for_service_timeout", bt_wait_for_service_timeout_);
```
SteveMacenski, line 242: Ditto fix

nav2_bringup/params/nav2_multirobot_params_1.yaml
```diff
@@ -45,6 +45,7 @@ bt_navigator:
     odom_topic: /odom
     bt_loop_duration: 10
     default_server_timeout: 20
+    bt_wait_for_service_timeout: 1000.0
```
SteveMacenski, line 48: Make sure to remark that the units are ms in the config guide

SteveMacenski (changes requested):
Otherwise a small typo, looks good to me! If you fix this + open the docs PR, I can merge the pair!

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
SteveMacenski: Also, for all, remove the `bt_` prefix. This is in the BT navigator node, its pretty clear in context at this is for BTs

SteveMacenski:
You have lots of test failures: https://app.circleci.com/pipelines/github/ros-planning/navigation2/10446/workflows/befffbd1-f698-4d7f-b19e-819cdd7a3490/jobs/32990/tests. Beyond the linting errors, I think you need to make sure you update the tests that rely on these BT nodes to have it set in their blackboards as well!

nav2_bringup/params/nav2_multirobot_params_1.yaml
SteveMacenski: @embeddedadam your config guide PR has this an an `int`, but this is clearly a `double`. This will cause a mismatched type exception to be thrown. Remove the `.0` from the config files

embeddedadam: @SteveMacenski yes I am aware of that, I fixed everything in new patch except uts.

SteveMacenski: OK! Please push that so I can take a look :-)

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -61,8 +61,8 @@ class BtActionNode : public BT::ActionNodeBase
     server_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
     getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
-    bt_wait_for_service_timeout_ =
-      config().blackboard->template get<std::chrono::milliseconds>("bt_wait_for_service_timeout_");
+    wait_for_service_timeout_ =
+      config().blackboard->template get<std::chrono::milliseconds>("wait_for_service_timeout_");
```
SteveMacenski, line 65: You still have the extra underscore in several places

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp
```diff
@@ -172,6 +175,7 @@ bool BtActionServer<ActionT>::on_configure()
   blackboard_->set<rclcpp::Node::SharedPtr>("node", client_node_);  // NOLINT
   blackboard_->set<std::chrono::milliseconds>("server_timeout", default_server_timeout_);  // NOLINT
   blackboard_->set<std::chrono::milliseconds>("bt_loop_duration", bt_loop_duration_);  // NOLINT
+  blackboard_->set<std::chrono::milliseconds>("wait_for_service_timeout_", wait_for_service_timeout_);  // NOLINT
```
SteveMacenski, line 178: underscore

SteveMacenski (changes requested):

SteveMacenski:
Broken tests are the main thing now!

SteveMacenski (approved):

tonynajjar:
Cool PR, I don't know why we didn't think of making it configurable back then and just gave up on it: https://github.com/ros-planning/navigation2/pull/3259

tonynajjar:
Actually I have a nitpick, `wait_for_service_timeout` seems biased towards services but we actually also use it for action servers. Wouldn't `wait_for_server_timeout` have been a better name? Even then, having `server_timeout` and `wait_for_server_timeout` is confusing as they sound similar. Maybe `server_response_timeout` and `server_bringup_timeout` would be better suitable? It would probably be a pain to deprecate `server_timeout` though

nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
```diff
@@ -93,7 +95,7 @@ class BtActionNode : public BT::ActionNodeBase
 
     // Make sure the server is actually there before continuing
     RCLCPP_DEBUG(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    if (!action_client_->wait_for_action_server(1s)) {
+    if (!action_client_->wait_for_action_server(wait_for_service_timeout_)) {
       RCLCPP_ERROR(
         node_->get_logger(), "\"%s\" action server not available after waiting for 1 s",
```
tonynajjar, line 100: missed changing this from 1 s to "wait_for_service_timeout_" seconds

christophfroehlich:
We also had issues with the fixed timeouts, and want to use this PR in our projects. @SteveMacenski would it be possible to backport this to humble+iron? Though, I'm not sure if this could break ABI, but it could be fine because all the nodes are loaded dynamically and not linked against a released binary?

christophfroehlich:
@SteveMacenski May I ping you again and ask for a backport to humble? Or maybe #4261 from iron.

SteveMacenski:
`wait_for_service_timeout` does not appear to be declared to work properly as-is for a migration. It looks like it was added later to resolve that bug https://github.com/ros-navigation/navigation2/commit/1ef462e71d12fd3e4ffb45e332a7d9883e921921 

So, I'm a little nervous about back porting this one since there were bugs identified that we missed on review / development and I haven't taken a deep look through the git history since this was merged to see if other bug fixes need to be applied with this to also make it work. 

But, same with the manual backports. If I can do another review on this and have a second set of eyes (with that patch applied) then we can try. No promises on this one though.

PR #3962, opened 2023-11-14
Prevent analytic expansions from shortcutting Smac Planner feasible paths

SteveMacenski:
Generally helps reduce the shortcutting on approach to goals by setting a maximum cost for traversal for the analytic expansion module to use, multiple turning radius refinement attempts, and an optional cost override when shortcutting is OK on approach when absolutely required.

It also addresses a minor collision detector issue that was resolved in the PR 

Needs: docs update + testing from deployed users

### DIFF
--- a/nav2_smac_planner/src/analytic_expansion.cpp
+++ b/nav2_smac_planner/src/analytic_expansion.cpp
@@ -132,7 +132,7 @@ typename AnalyticExpansion<NodeT>::NodePtr AnalyticExpansion<NodeT>::tryAnalytic
         float initial_score = scoringFn(analytic_nodes);
         float score = std::numeric_limits<float>::max();
         float min_turn_rad = node->motion_table.min_turning_radius;
-        float step = 0.10; // 10cm TODO make proportional to costmap resolution or a parameter?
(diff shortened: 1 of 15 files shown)


nav2_smac_planner/src/analytic_expansion.cpp
```diff
@@ -196,6 +245,38 @@ typename AnalyticExpansion<NodeT>::AnalyticExpansionNodes AnalyticExpansion<Node
...
+          break;
+        }
+      }
+
+      // (3) Handle exception: there may be no other option close to goal
+      // if max cost is set too low (optional)
+      if (failure) {
+        if (d < 2.0f * M_PI * goal->motion_table.min_turning_radius &&
+          _search_info.analytic_expansion_max_cost_override)
+        {
```
harderthan, line 273: nitpick.  

```suggestion
          _search_info.analytic_expansion_max_cost_override) {
```

SteveMacenski: That is incorrect formatting.

SteveMacenski:
- [ ] migration
- [ ] config / readme

SteveMacenski:
Docs: https://github.com/ros-planning/navigation.ros.org/pull/511

PR #3948, opened 2023-11-08
3732 costmap raw update msgs

emSko:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | #3732  |
| Primary OS tested on | Ubuntu |
| Robotic platform tested on | gazebo simulation |

---

## Description of contribution in a few bullet points

* Adding new msg type **CostmapUpdate.msg**,
* Adding **CostmapUpdate** publisher to  **Costmap2DPublisher** class,
* Re-implementation of the **CostmapSubscriber** class to support the **CostmapUpdate** msg. 

## Description of documentation updates required from your changes

---

## Future work that may be required in bullet points

* I see a lot of redundancy in **Costmap2DPublisher** class while creating the msgs, adding methods that are responsible for populating msgs and return the unique ptr to msgs might improve the code readability

#### For Maintainers: 
- [ ] Check that any new parameters added are updated in navigation.ros.org
- [ ] Check that any significant change is added to the migration guide
- [ ] Check that any new features **OR** changes to existing behaviors are reflected in the tuning guide
- [ ] Check that any new functions have Doxygen added
- [ ] Check that any new features have test coverage
- [ ] Check that any new plugins is added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

Issue #3732, opened 2023-08-02
Update costmap publisher and subscriber to allow for 'update' messages in raw costmaps


SteveMacenski:
Currently, all costmap publisher / subscription when using raw messages (e.g. costmap messages containing the raw costmap from 0-255 rather than the occupancy grid message of 0-100 for internal use by the task servers such as behavior and smoother) send the full costmap each time. This is unideal, even though we compose all our nodes into the same process, which does help greatly with compute time.

It would be better if we mirrored the Update methods from the OccupancyGrid publication to minimize the number of times a full, raw, costmap is sent over the network and only publish the updated regions. This would decrease compute resources and latency for the behavior and smoother server (or other servers for client applications using the raw costmaps).

This is a great first issue for a new contributor! Just make sure to look over the code and make sure that all instances of costmap pub/sub are using these classes to make sure that everyone can properly handle the updates (which I believe they all should be, minus some unit tests)

Relativiteit:
@SteveMacenski I would like to contribute to this as one of my first contributions!

SteveMacenski:
Great! Any questions you might have :-) The file I point to should pretty much have the "core basics" of demonstration of how such a feature could work using the OccupancyGrid update logic. Its mostly mirroring that for the nav2_msgs/Costmap message type and updating the publisher and subscriber to work with it. Then, checking uses of the Costmap message to make sure everyone is using the pub/sub objects to get those updates -- in case there are some raw publishers / subscribers hanging around!

Thanks and I look forward!

SteveMacenski:
@Relativiteit any update? :-)

emSko:
I would like to take over this issue.

Just to recap, you think about mimicking this behaviour [costmap_update_pub_](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_publisher.cpp#L203-L222)  for _costmap_raw_pub__ . 

So creating CostmapUpdate message and adding proper publisher and subscriber.

SteveMacenski:
And the counterpart in the costmap subscriber as well to take those published updates and populate the raw costmap!

Relativiteit:
hey guys yes, I got a bit lost on how to do it please @emSko take it over this was a bit above my head. I would still like to try it this weekend. And make a PR but please take point on it.

SteveMacenski:
@emSko any updates or thoughts? How's it going :smile:

emSko:
[...]

PR #3965, opened 2023-11-15
3732 costmap raw update msgs rebased (fixes github issues with PR #3948)

emSko:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3732  |

---

* Adding new msg type **CostmapUpdate.msg**,
* Adding **CostmapUpdate** publisher to  **Costmap2DPublisher** class,
* Re-implementation of the **CostmapSubscriber** class to support the **CostmapUpdate** msg. 

---

* I see a lot of redundancy in **Costmap2DPublisher** class while creating the msgs, adding methods that are responsible for populating msgs and return the unique ptr to msgs might improve the code readability

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_publisher.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_publisher.cpp
@@ -77,8 +77,7 @@ Costmap2DPublisher::Costmap2DPublisher(
   costmap_update_pub_ = node->create_publisher<map_msgs::msg::OccupancyGridUpdate>(
     topic_name + "_updates", custom_qos);
   costmap_raw_update_pub_ = node->create_publisher<nav2_msgs::msg::CostmapUpdate>(
-    topic_name + "_raw_updates",
-    custom_qos);
+    topic_name + "_raw_updates", custom_qos);
 
   // Create a service that will use the callback function to handle requests.
   costmap_service_ = node->create_service<nav2_msgs::srv::GetCostmap>(
@@ -192,16 +191,16 @@ nav2_msgs::msg::CostmapUpdate Costmap2DPublisher::get_raw_costmap_update_msg()
   msg.header.frame_id = global_frame_;
   msg.x = x0_;
   msg.y = y0_;
-      msg.size_x = xn_ - x0_;
-      msg.size_y = yn_ - y0_;
-      msg.data.resize(msg.size_x * msg.size_y);
-    
-    unsigned int i = 0;
-    for (unsigned int y = y0_; y < yn_; y++) {
-      for (unsigned int x = x0_; x < xn_; x++) {
-        msg.data[i++] = costmap_->getCost(x, y);
-      }
+  msg.size_x = xn_ - x0_;
+  msg.size_y = yn_ - y0_;
+  msg.data.resize(msg.size_x * msg.size_y);
+
+  unsigned int i = 0;
+  for (unsigned int y = y0_; y < yn_; y++) {
+    for (unsigned int x = x0_; x < xn_; x++) {
+      msg.data[i++] = costmap_->getCost(x, y);
     }
+  }
   return msg;
 }
 
@@ -243,15 +242,9 @@ void Costmap2DPublisher::publishCostmap()
       }
       costmap_update_pub_->publish(std::move(update));
     }
-
     if (costmap_raw_update_pub_->get_subscription_count() > 0) {
-      RCLCPP_WARN(
-      logger_,
-      "PUBLIKUJE");
       costmap_raw_update_pub_->publish(get_raw_costmap_update_msg());
     }
-
-    
   }
 
   xn_ = yn_ = 0;
--- a/nav2_costmap_2d/src/costmap_2d_publisher.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_publisher.cpp
@@ -224,7 +224,6 @@ std::unique_ptr<nav2_msgs::msg::CostmapUpdate> Costmap2DPublisher::createCostmap
   std::uint32_t i = 0;
   for (std::uint32_t y = y0_; y < yn_; y++) {
     for (std::uint32_t x = x0_; x < xn_; x++) {
-      //todo copy from array instead of calling method
       msg->data[i++] = costmap_->getCost(x, y);
     }
   }
@@ -250,7 +249,7 @@ void Costmap2DPublisher::publishCostmap()
       costmap_raw_pub_->publish(std::move(costmap_raw_));
     }
   } else if (x0_ < xn_) {
-    // Publish Just Updates
+    // Publish just update msgs
     std::unique_lock<Costmap2D::mutex_t> lock(*(costmap_->getMutex()));
     if (costmap_update_pub_->get_subscription_count() > 0) {
       costmap_update_pub_->publish(createGridUpdateMsg());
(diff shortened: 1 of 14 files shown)


emSko:
This is the continuation of work from #3948

nav2_costmap_2d/src/costmap_2d_publisher.cpp
```diff
@@ -181,13 +183,30 @@ void Costmap2DPublisher::prepareCostmap()
...
+  msg->x = x0_;
+  msg->y = y0_;
+  msg->size_x = xn_ - x0_;
+  msg->size_y = yn_ - y0_;
+  msg->data.resize(msg->size_x * msg->size_y);
+
+  std::uint32_t i = 0;
+  for (std::uint32_t y = y0_; y < yn_; y++) {
+    for (std::uint32_t x = x0_; x < xn_; x++) {
+      //todo copy from array instead of calling method
```
SteveMacenski, line 201: You have a few TODOs that should be resolved before turning this into a full non-draft PR.

SteveMacenski (commented):
Otherwise, beyond tests, this looks good :-)

nav2_costmap_2d/src/costmap_subscriber.cpp
```diff
@@ -14,84 +14,155 @@
...
 }
 
+// todo: change to const?
 std::shared_ptr<Costmap2D> CostmapSubscriber::getCostmap()
 {
   if (!costmap_received_) {
     throw std::runtime_error("Costmap is not available");
   }
-  toCostmap2D();
+  if (costmap_msg_ != nullptr) {
```
SteveMacenski, line 67:
```suggestion
  if (costmap_msg_) {
```

nav2_costmap_2d/src/costmap_subscriber.cpp
```diff
@@ -14,84 +14,155 @@
...
+    costmap_->resizeMap(
+      costmap_msg_->metadata.size_x, costmap_msg_->metadata.size_y,
+      costmap_msg_->metadata.resolution,
+      costmap_msg_->metadata.origin.position.x,
+      costmap_msg_->metadata.origin.position.y);
   }
+
+  unsigned char * master_array = costmap_->getCharMap();
+  std::copy(costmap_msg_->data.begin(), costmap_msg_->data.end(), master_array);
+  costmap_msg_ = nullptr;
```
SteveMacenski, line 141: Use `costmap_msg_.reset()` will do the same thing and in the smart-pointer-y style

nav2_costmap_2d/src/costmap_2d_publisher.cpp
emSko: resolved with 4e42850620fa8b753ab0b8c7e0517cacf759a931

nav2_costmap_2d/src/costmap_subscriber.cpp
emSko: resolved with 4e42850620fa8b753ab0b8c7e0517cacf759a931

nav2_costmap_2d/src/costmap_subscriber.cpp
emSko: resolved with 4e42850620fa8b753ab0b8c7e0517cacf759a931

nav2_costmap_2d/src/costmap_2d_publisher.cpp
```diff
@@ -115,13 +117,20 @@ void Costmap2DPublisher::onNewSubscription(const ros::SingleSubscriberPublisher&
...
 
-// prepare grid_ message for publication.
-void Costmap2DPublisher::prepareGrid()
+
+void Costmap2DPublisher::updateGridParams()
 {
-  std::unique_lock<Costmap2D::mutex_t> lock(*(costmap_->getMutex()));
+  saved_origin_x_ = costmap_->getOriginX();
+  saved_origin_y_ = costmap_->getOriginY();
   grid_resolution = costmap_->getResolution();
```
SteveMacenski, line 125: Grid resolution/width/height should have trailing underscores as class member variables

SteveMacenski:
Overall looks good / done to me! The costmap publisher seems to be missing coverage in several entire functions though: https://app.codecov.io/gh/ros-planning/navigation2/pull/3965/blob/nav2_costmap_2d/src/costmap_2d_publisher.cpp

emSko:
@SteveMacenski  the lines that are missing coverage are not related to the functionality implemented in this issue (such as grid publishing and costmap service). I feel that lines were missing the coverage before my changes and the tests should be added in different PR

SteveMacenski:
OK - still heed the underscore comment above

nav2_costmap_2d/src/costmap_2d_publisher.cpp
emSko: resolved with 9b92028155ceefbfa42881738e212f44109a26fe

SteveMacenski (approved):

SteveMacenski:
@emSko last thing to merge this: Add new feature to the migration guide! https://navigation.ros.org/migration/Iron.html

Also, please do a run of the system and add some prints to make sure its working in production and not just sending the full costmap except on initialization (e.g. update callback is triggered at runtime mostly). Do you see any major computation changes in the runtime performance of the system? I suspect this change may actually be visible in metrics

SteveMacenski:
Any metrics changes?

emSko:
@SteveMacenski  I have run the simulation with the default map for Turtlebot3 and unfortunately, I could not observe any noticeable change in performance.
```
Full costmap not navigating. Average CPU time: 28.294999999999995, average memory 1.4000000000000004.

Full costmap navigating. Average CPU time: 49.1, average memory 1.3640000000000003.

Update not navigating. Average CPU time: 29.507, average memory 1.4000000000000004.

Update navigating. Average CPU time: 50.132, average memory 1.3770000000000002. 
```
Not navigating means that the robot was just standing still, navigating means that the robot was navigating to goals. These are the measurements for a composable node after collecting 100 samples for each case. 


Then with help from Alexey, I created a synthetic map of size 1000x1000 px, however, I was not able to collect the measurements because the simulation kept crashing on my laptop, but by just eyeballing the top command I would say that there still were not any observable changes.

emSko:
I've checked the functionality with the ros2 topic echo command. It works as intended that is the full costmap is latched on the topic and only the updates are sent continuously. I also have implemented the helper node, as you suggested, that is converting the raw msgs into grid msgs to visualize it in rviz - it shows that everything is correct. 

From my analysis, I have found that it does not make sense to use the update msg for the local costmap. Due to the change of the origin while the robot is moving, the full costmap is sent and the updates are only sent when the robot is standing still. Also, the size of the full costmap is insignificantly larger than the update (3600 cells vs ~3300 cells). However, in the case of the global costmap, we are dealing with the decrease of the number of cells ~10 times (when running the simulation of the default turtlebot map/world, for larger maps the difference is larger).

emSko:
I have nothing to add at the moment. As for me, the changes can be merged at the current state.

emSko:
Thank you, Steven, for all the comments and suggestions.
I was thinking about working on the AI Integration Layers that are listed on the nav2 user survey. Could you provide me with more info about that on Slack, or can we discuss it during the working group meeting?

Issue #3971, opened 2023-11-20
fix bug mentioned in #3958 && a futhuer suggestion for child-LifecycleNode mechanism design


GoesM:
## Bug report

**Required Info:**

- Operating System: 
  - all
- ROS2 Version: 
  - all
- Version or commit hash: 
  - newest
- DDS implementation:
  - 

#### Steps to reproduce issue

```

```

#### Expected behavior
expected behavior what the degisner wish should be like 

`costmap_ros_->deactivate()` works during `planner/controller->on_deactivate()`, as well as `costmap_ros_->cleanup()` works during `planner/controller->on_cleanup()` 

[.../nav2_planner/src/planner_server.cpp#L224 and L257](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L224), and similar code in `nav2_controller`

#### Actual behavior
As an individual LifecycleNode, `costmap_ros_` may react to rcl_preshutdown **randomly**, and always be closed much earlier than planner/controller->`deactivate()`

it would lead to many unexpected crashes like bug mentioned in #3958 : 

during shutdown_period, `costmap_ros_` may react to rcl_preshutdown earlier, so that planner/controller may try to access the `costmap_ros_` though it has been a nullptr. As a result it would lead to a crash during shutdown period.

such crashes may lead to bad situation that :   

program shut down in an unplanned manner and result in **residual processes**. 

***

## Additional information

**costmap** is set as a Nav2::util **LifecycleNode** : [../nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp#L 73](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp#L73)

all LifecycleNode would react to **Ctrl+C** [../nav2_util/src/lifecycle_node.cpp#L 90](https://github.com/ros-planning/navigation2/blob/main/nav2_util/src/lifecycle_node.cpp#L90)

all LifecycleNode would react to  rcl_preshutdown **randomly**, and similar issue has been discussed before : [../ros2/rclcpp/issues/2096](https://github.com/ros2/rclcpp/issues/2096)

this explanation should have been clear enough:  [../nav2_planner/src/planner_server.cpp #L 214-L225](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L214-L225)

[...]

Issue #3303, opened 2022-11-30
Timed Behaviors do not adhere to Sim Time


bobbleballs:
## Bug report

- Operating System:
  Ubuntu 22.04

- ROS2 Version:
 Humble

- Version or commit hash:
0d86469f1c31708117e68f3f44ad5392347e9a5d

- DDS implementation:
  - Cyclone-DDS
#### Steps to reproduce issue

I have an issue where I am running a simulation at greater than real time speed. My simulation platform (Flatland2D) is publishing sim time on /clock and use_sim_time is set on the Nav2 launch procedure. My issue is that my behaviour tree contains some waits, and backups however these behaviours occur at real time not sim time.

#### Expected behavior
Timed behaviours will work to sim time not real time. e.g. in a simulation running at 2x speed with a wait of 10 seconds I would expect this to take 5 Seconds.

#### Actual behavior
Waits take the real length of time.

#### Additional information
I've had a look at nav2_behaviors/plugins/wait and back_up and it looks like wait uses a steady_time from chrono which calls now(). I don't know much about chrono but I'm guessing this gets an accurate system time?
nav2_behaviors/include/timed_behaviors.hpp also uses ` rclcpp::Clock steady_clock_{RCL_STEADY_TIME};` which I presume does something similar?

For the behaviours to adhere to use_sim_time these behaviours should be using `RCL_ROS_TIME` I think? Would this have any negative effects. I'm not sure if steady_time was used for a reason? Ideally I am trying to make the behaviour of the navigation stack at high sim rates as accurate as real time so this is a blocker to me.

SteveMacenski:
Good question, I don't see any reason why we couldn't do that. Yes, chrono is C++ system time independent of ROS / simulation times. You'd probably need to update not only the behaviors themselves, but the TimedBehavior header class which also uses chrono to dish out behaviors. You should get the clock object from the root node so that it can use real/ros time as the node is set appropriately. 

We used steady time because the ROS clock isn't guaranteed to move forward or be regular. There are certain things in the stack (e.g. server timeouts, BT node rate controllers) which should probably use wall time instead. I think there can be an argument for the behaviors to be moved to ROS time, that seems sensible. I think it would just be important as you're changing code in the behavior server to ask yourself "how could this go wrong?" and make sure to handle any edge cases that could be introduced if the clock either stopped or moved backwards (so B-A could be negative). 

@jwallace42 any objection? @bobbleballs would you be open to contributing this?

[...]

PR #3972, opened 2023-11-20
fix bug mentioned in #3958

GoesM:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3971  |
| Robotic platform tested on | turtlebot |

---

### bug description

**costmap** is set as a Nav2::util **LifecycleNode** : [../nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp#L 73](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp#L73)

all LifecycleNode would react to **Ctrl+C** [../nav2_util/src/lifecycle_node.cpp#L 90](https://github.com/ros-planning/navigation2/blob/main/nav2_util/src/lifecycle_node.cpp#L90)

all LifecycleNode would react to  rcl_preshutdown **randomly**, and similar issue has been discussed before : [../ros2/rclcpp/issues/2096](https://github.com/ros2/rclcpp/issues/2096)

this explanation should have been clear enough:  [../nav2_planner/src/planner_server.cpp #L 214-L225](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L214-L225)

**the bug happening in [ISSUE: #3958](https://github.com/ros-planning/navigation2/pull/3958) is caused by this.**

because during shutdown_period, `costmap_ros_` may react rcl_preshutdown earlier, so that planner/controller may try to access the `costmap_ros_` though it has been a nullptr.

### our provided solution
it seems that `costmap` is only used as a **child-thread** of planner/controller,  [../nav2_planner/src/planner_server.cpp#L 89](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L89)

so can we make it do nothing when reacting to Ctrl+C ?

doing so, its lifecycle would be completely bond to planner/controller's lifecycle, and also going to death in order.

```cpp
//in file `costmap_2d_ros.hpp`
  /**
   * @brief override on_rcl_preshutdown() as empty
   * [reason] costmap only could be created by its parents like planner/controller_server
   * [reason] costmap may react to ctrl+C earlier than its parents to cause nullptr-accessed
   * so the costmap's reaction must be later than its parent to avoid nullptr-accessed
   * 
   * thus, it's a more perfect way to override on_rcl_preshutdown() as empty function
   * and its deactivate must be joint in planner/controller_server->on_deactivate()
   * and its cleanup must be joint in planner/controller_server->on_cleanup()
   * 
   * !!! a mention !!!
   * if any other place use this class (Costmap2DROS)
   * it's necessary to let costmap->deactivate() joint in parent->deactivate()
   * and let costmap->cleanup() joint in parent->cleanup() 
   */
  void on_rcl_preshutdown() override{
    //do nothing
    return ;
  }
```

because the `costmp_ros_` is **just a thread** created by planner/controller,  

whether program shutdown correctly or not, or even dead unexpectably, the **child-thread** could be closed all the time since **parent-LifecycleNode** sub-process dead. 

Thus there's no need to worry whether it would be still alive after whole program exit finally.

### in addition
if our solution being adopted, many checks could be removed :

-  [double check in planner_server->on_deactivate()](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L221)

- [double check in planner_server->on_cleanup()](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L254)

- [double check in controller_server->on_deactivate()](https://github.com/ros-planning/navigation2/blob/main/nav2_planner/src/planner_server.cpp#L285)

- [double check in controller_server->on_cleanup()](https://github.com/ros-planning/navigation2/blob/main/nav2_controller/src/controller_server.cpp#L315)

these double checks are also designed to face to bugs in situation that `costmap_ros_` may be closed earlier than planner/controller. After our code modification, these checks are useless anymore.

what's more, our solution performs perfectly during our local tests, much more than our past solutions mentioned in #3958 

---

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -284,6 +284,7 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
    */
   if (costmap_ros_->get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+  //this condition could be removed
   {
     costmap_ros_->deactivate();
   }
@@ -314,6 +315,7 @@ ControllerServer::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
   progress_checkers_.clear();
   if (costmap_ros_->get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+  //this condition could be removed
   {
     costmap_ros_->cleanup();
   }
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
@@ -151,7 +151,7 @@ public:
    */
   void on_rcl_preshutdown() override
   {
-    //do nothing
+    // do nothing
     return;
   }
 
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
@@ -135,6 +135,27 @@ public:
   nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
   /**
+   * @brief override on_rcl_preshutdown() as empty
+   * [reason] costmap only could be created by its parents like planner/controller_server
+   * [reason] costmap should react to ctrl+C later than its parents to avoid nullptr-accessed
+   * so the reaction must be later than its parent
+   *
+   * thus, it's a more perfect way to make on_rcl_preshutdown() as empty function
+   * only joint in planner/controller_server->on_deactivate()
+   * and planner/controller_server->on_cleanup()
+   *
+   * !!! a mention !!!
+   * if any other place use this class (Costmap2DROS)
+   * it's neccessary to let costmap->deactivate() joint in parent->deactivate()
+   * and let costmap->cleanup() joint in parent->cleanup()
+   */
+  void on_rcl_preshutdown() override
+  {
+    //do nothing
+    return;
+  }
+
+  /**
    * @brief  Subscribes to sensor topics if necessary and starts costmap
    * updates, can be called to restart the costmap after calls to either
    * stop() or pause()
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -220,6 +220,7 @@ PlannerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
    */
   if (costmap_ros_->get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+  //this condition could be removed
   {
     costmap_ros_->deactivate();
   }
@@ -253,6 +254,7 @@ PlannerServer::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
    */
   if (costmap_ros_->get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE)
+  //this condition could be removed
   {
     costmap_ros_->cleanup();
   }


nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +134,27 @@ class Costmap2DROS : public nav2_util::LifecycleNode
...
+   * thus, it's a more perfect way to make on_rcl_preshutdown() as empty function
+   * only joint in planner/controller_server->on_deactivate()
+   * and planner/controller_server->on_cleanup()
+   *
+   * !!! a mention !!!
+   * if any other place use this class (Costmap2DROS)
+   * it's neccessary to let costmap->deactivate() joint in parent->deactivate()
+   * and let costmap->cleanup() joint in parent->cleanup()
+   */
+  void on_rcl_preshutdown() override
```
SteveMacenski, line 152: This is not a viable solution, destroying bond and shutting things down is absolutely necessary.

SteveMacenski (changes requested):

SteveMacenski:
I'm rejecting this line of thinking on premise. Lets follow up in your ticket, I think the answer here is much easier than you're seeing.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +134,27 @@ class Costmap2DROS : public nav2_util::LifecycleNode
    */
   nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
+  /**
+   * @brief override on_rcl_preshutdown() as empty
+   * [reason] costmap only could be created by its parents like planner/controller_server
```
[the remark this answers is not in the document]
SteveMacenski: So, it would be useful to write a quick unit test to destroy the object https://github.com/ros-planning/navigation2/blob/main/nav2_util/src/lifecycle_node.cpp#L43-L54 and make sure the transition is also followed even if `rclcpp::shutdown()` isn't yet called

SteveMacenski (changes requested):

SteveMacenski:
I think this is a clever and fine solution to the problem, just some questions to answer

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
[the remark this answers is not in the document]
GoesM: > Incorrect, there is a costmap node `main.cpp` file that it can be launched independently

any code file show that?  we didn't find where the node is launched independently......

or you mean it just would be tested by launching independently ?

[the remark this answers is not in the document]
GoesM: > So, it would be useful to write a quick unit test to destroy the object 

absolutely, we find it failed in CI because the existing unit test launched it independently.

however, in our suggestion, we'd like to make it as a **child-LifecycleNode** fully dependent on some other lifecycle node.

nav2_controller/src/controller_server.cpp
```diff
@@ -284,6 +284,7 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
    */
   if (costmap_ros_->get_current_state().id() ==
     lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
+  //this condition could be removed
```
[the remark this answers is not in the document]
GoesM: We test it (removing these four conditions ) by repeating  `ros2 launch` and _`pre-shutdown` at random moment_ more than 1000 times in our local environment, as a result there's no shutdown error happening.

It now appears that the unit test of CI considers Costmap as an individual node for launch testing (launch it independntly). 

If we could change this unit test into relying on a virtual lifecycle node for launch testing, it would be more fitable for our suggestion.
After that, I believe that the coverage report may help us to prove these four conditions will be 100% judged correct.

GoesM:
We specially name nodes like `costmap_ros_` as **child-LifecycleNode**, which is born from their **parent-LifecycleNode** like `planner_server` and `controller_server`.

**our suggestion is:**  
Could we set up a mechanism  specially for such situations like **child-LifecycleNode**, to make sure they're completely controlled by their parent?

these **child-LifecycleNode** should have some basic specialist like:
- they must be created by a **parent-LifecycleNode**
- they must be created as a **thread** but **not a sub-process**

such mechanism could be like: 

- `child's->activate` should be joint into its their parent's `on_configure` 
- `child->deactivate()` **must** be joint into its parent's `on_deactivate`
- `child->cleanup()` **must** be joint into its parent's `on_cleanup`
- and so on

Perhaps such mechanism could be used for any program in ROS2 ?

GoesM:
At present, the unit test of CI takes nav2_ Costmap is considered as an independent lifecycle node for running and closing tests, which goes against the premise of our proposed modification proposal
>these child-LifecycleNode should have some basic specialist like:
>-- they must be created by a parent-LifecycleNode
>-- they must be created as a thread but not a sub-process

For this, we have three suggestions to address different development needs that may exist in the future.

#### 1. modify the unit-test
If it can be guaranteed that in current software usage and subsequent software development, 
`nav2_costmap` will only be run as a **child-LifecycleNode**,  
the next step should be to modify the unit test.

It's only in need to modify the `launch` method of `costmap` based on the original unit test.
Declare and launch a virtual **parent-LifecycleNode**, through which to run and preshutdown `costmap` (just like the how planner/controller do it)

#### 2: create `class ChildCostmap2DROS` 
if you're worried that `Costmap2DROS` may be used as an individual LifecycleNode, 
our suggestion is to create a new class named as `ChildCostmap2DROS`, specifically to implement the special usage of **child-LifecycleNode** by planner/controller.

It's only in need to create some new files like `child_costmap_2d_ros.hpp` similar to `costmap_2d_ros.hpp`, and `child_*.cpp` simliar to `*.cpp`

and change declaration of `global_costmap_` and `local_costmap_` in `planner/controller_server.cpp`

#### 3. create a new class `ChildLifecycleNode` for ROS2
is there anything else, like `costmap`, utilizes the characteristics of LifecycleNode for performance optimization, but is only used as a sub-thread? 
if such utilize method is widely used by various ROS2 software,
our suggestion is to create a new class named as ``ChildLifecycleNode` for whole ROS2, to optimize the shutdown mechanism of all **child-LifecycleNode**s and **parent-LifecycleNode**s, to bring thread management optimization

by the way, such class `ChildLifecycleNode` must comply with the following mandatory requirements when used:
1. The 'activate', 'deactivate', and 'cleanup' of this class must be called by a **parent-LifecycleNode**
2. This class **must be** executed at the **thread** level and **not** be executed at the **process** level

GoesM:
whether suggestion 1 or 2 you prefer ?  Or 

If there is any work that requires our assistance, we'd like to have a try within our capabilities   O(∩_∩)O

nav2_controller/src/controller_server.cpp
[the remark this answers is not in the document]
GoesM: another way could also be tried:

change code into:
```cpp
  if (costmap_ros_->get_current_state().id() ==
    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
{
.....
}
else {
    RCLCPP_INFO(get_logger(), “??????????????????");
}
```
after test, code line `RCLCPP_INFO(get_logger(), “??????????????????");` could not be covered anymore.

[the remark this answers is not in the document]
SteveMacenski: Did you read the discussion to understand why those were there and its removal doesn't have any downsides after making this change?

>  It now appears that the unit test of CI considers Costmap as an individual node for launch testing (launch it independently). If we could change such unit test into relying on a virtual parent-LifecycleNode for launch and shutdown testing, it would be more fitable for our suggestion.

I don't understand. Are you saying that this doesn't work when the costmap nodes are launched independently? See other comment below, that needs to work as well.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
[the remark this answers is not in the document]
SteveMacenski: Its not currently used within the stack, but there are user that use it that way and it is a supported feature. You'd have to test this yourself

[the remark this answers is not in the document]
GoesM: > Its not currently used within the stack, but there are user that use it that way and it is a supported feature. 

ok, i get it. I will try to make this mechanism compatible with different usage modes, both as a child-LifecycleNode and as an independent LifecycleNode.

SteveMacenski:
> If there is anymore work that requires our assistance, we'd like to have a try within our capabilities O(∩_∩)O

What might you be interested in doing? I can always certainly find projects :-) Taking a stab at https://github.com/ros-planning/navigation2/issues/3303 would be very helpful - Alexey made some changes to rclcpp::Rate to make it better supportable of different clocks https://github.com/AlexeyMerzlyakov/rclcpp/blob/rolling/rclcpp/include/rclcpp/rate.hpp#L141

nav2_controller/src/controller_server.cpp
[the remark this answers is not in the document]
GoesM: >Did you read the discussion to understand why those were there and its removal doesn't have any downsides after making this change?

of course, our solution could solve it together. unclean shutdown issue mentioned in #3377 is very similar to #3958 . 

we believe they're all caused by:
>all LifecycleNode react to rcl_preshutdown randomly,
&&
>costmap may react to ctrl+C earlier than its parents to cause nullptr-accessed and then, unclean shutdown.

so what we achieved is to make sure that `child-LifecycleNode`'s reaction to `rcl_preshutdown` is all up to its `parent-LifecycleNode`.


what's more, as #3377 said:
>There are a few instances fixed in this PR, but likely not all (see Future work below)

that's because `double check` method used before may be just a temporary solution. it's too hard for users to check all details and add `double check` in their whole programs.

so, we believe our solution is a thorough solution to all of similar issues in nav2-programs, and more friendly for users to use it.
they don't need to check too much and they don't need to add `double check` anymore. 
What they only need to do is to run `costmap_ros_->turnChildLifecycleNode()` following they set `  costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);`
doing so, `costmap_ros_`'s state will be completely up to its parents-LifecycleNode planner/controller_server.

GoesM:
>What they only need to do is to run costmap_ros_->turnChildLifecycleNode() following they set   costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);

you can see how to do it in our latest code as a case.

GoesM:
> I can always certainly find projects :-) Taking a stab at #3303 would be very helpful - Alexey made some changes to rclcpp::Rate to make it better supportable of different clocks https://github.com/AlexeyMerzlyakov/rclcpp/blob/rolling/rclcpp/include/rclcpp/rate.hpp#L141

WOW! That's cool ! Thanks a lot ~

nav2_controller/src/controller_server.cpp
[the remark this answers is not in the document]
GoesM: >Did you read the discussion to understand why those were there and its removal doesn't have any downsides after making this change?

of course, our solution could solve it together. unclean shutdown issue mentioned in #3377 is very similar to #3958 . 

they're all caused by:
>all LifecycleNode react to rcl_preshutdown randomly,
&&
>costmap may react to ctrl+C earlier than its parents to cause nullptr-accessed and then, unclean shutdown.

so what we achieved is to make sure that `child-LifecycleNode`'s reaction to `rcl_preshutdown` is all up to its `parent-LifecycleNode`.


what's more, as #3377 said:
>There are a few instances fixed in this PR, but likely not all (see Future work below)

that's because `double check` method used before may be just a temporary solution. it's too hard for users to check all details and add `double check` in their whole programs.

so, we believe our solution is a thorough solution to all of similar issues in nav2-programs, and more friendly for users to use it.
they don't need to check too many details and they don't need to add `double check` anymore. 
What they only need to do is to run `costmap_ros_->turnChildLifecycleNode()` following they set `  costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);`
doing so, `costmap_ros_`'s state will be completely up to its parents-LifecycleNode planner/controller_server.

GoesM:
>What they only need to do is to run costmap_ros_->turnChildLifecycleNode() following they set   costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);

you can see how to do it in our latest code as a case.

GoesM:
if only considering current nav2, we're very sure that such not covered checks could be removed.

However, perhaps it may be still worried that any future changes may make these checks work again. 
 
anyway, whether remove or retain them, I am very willing to make modifications based on your futher suggestions  (●'◡'●)

nav2_controller/src/controller_server.cpp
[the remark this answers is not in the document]
GoesM: >What they only need to do is to run costmap_ros_->turnChildLifecycleNode() following they set   costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);

the case is like following:
```cpp
  // Launch a thread to run the costmap node
  costmap_thread_ = std::make_unique<nav2_util::NodeThread>(costmap_ros_);
  // work as a child-LifecycleNode of its parent planner_server
  costmap_ros_->turnChildLifecycleNode();
```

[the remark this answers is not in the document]
SteveMacenski: > of course, our solution can cover the role of these double-checks.

OK, you can remove them - don't comment them out.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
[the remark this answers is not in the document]
SteveMacenski: - https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_node.cpp

Like I said, its not run during nav2's default system, but it is used by people not only in testing but in production.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +134,42 @@ class Costmap2DROS : public nav2_util::LifecycleNode
...
+  /**
+   * @brief child-LifecycleNode mode
+   * sometimes costmap may be launched by another LifecycleNode and work as a child-thread
+   * child-LifecycleNode should react to ctrl+C later than its parents to avoid unclean shutdown
+   * Thus, it's neccessary to turnChildLifecycleNode() to change its launch mode
+   *
+   * in ChildLifecycleNode mode, it would react to rcl_preshutdown anymore
+   * all its lifecycle state would be controlled by its parent-LifecycleNode
+   */
+  void turnChildLifecycleNode()
```
SteveMacenski, line 146: Better option:
- https://github.com/ros-planning/navigation2/blob/4bc3e8a534005e983ee7a49a09ec00aeb9820661/nav2_costmap_2d/src/costmap_2d_node.cpp#L46

This non-argument constructor is only used in the node, not in any of the internal uses of costmap 2d. [Update it to take in NodeOptions with a default constructed option](https://github.com/ros-planning/navigation2/blob/4bc3e8a534005e983ee7a49a09ec00aeb9820661/nav2_planner/include/nav2_planner/planner_server.hpp#L58).

Then, we know that constructor is only used in the `Node` or `Component` node independent operation. Any internal-node will use the other constructors. That eliminates `turnChildLifecycleNode`. 

In this constructor set `is_lifecycle_follower_ = false;`

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -317,6 +353,9 @@ class Costmap2DROS : public nav2_util::LifecycleNode
   double getRobotRadius() {return robot_radius_;}
 
 protected:
+  // launch mode: as a child-thread or an independent-process
+  bool is_thread_{false};
```
SteveMacenski, line 357: `is_lifecycle_follower_{true}`

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +134,42 @@ class Costmap2DROS : public nav2_util::LifecycleNode
...
+   * in ChildLifecycleNode mode, it would react to rcl_preshutdown anymore
+   * all its lifecycle state would be controlled by its parent-LifecycleNode
+   */
+  void turnChildLifecycleNode()
+  {
+    // as a child-LifecycleNode launched by its parent-LifecycleNode
+    // and work as a child-thread
+    is_thread_ = true;
+  }
+  bool isThread()
```
SteveMacenski, line 152: Remove method

SteveMacenski (changes requested):
LGTM in concept, but we can play some games and make it cleaner

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
GoesM: I'm not very clear about how to do it, while I've tried to modify our code in our latest change.

GoesM:
OMG ! I'm sorry that I uploaded codes with wrong use of `is_lifecycle_follower_` before. 

I regard `is_lifecycle_follower_` as whether the costmap node is following another node, so It should default to false ?

is it a correct understanding from you ?

GoesM:
may I ask how to local CI test as completely same as yours? 

It seems that we failed in some code style check many times, however we passed code style check locally by ` ament_uncrustify` provided by ros2..  Is there anything else ignored by us ?

GoesM:
it seems right now, check it please ^_^

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -99,6 +99,21 @@ class Costmap2DROS : public nav2_util::LifecycleNode
...
+   * @param local_namespace Namespace to append to the parent namespace
+   * @param use_sim_time Whether to use simulation or real time
+   * @param is_lifecycle_follower Whether launch as a child-LifecycleNode following others
+   */
+  explicit Costmap2DROS(
+    const std::string & name,
+    const std::string & parent_namespace,
+    const std::string & local_namespace,
+    const bool & use_sim_time,
+    const bool & is_lifecycle_follower);
```
SteveMacenski, line 115: No, not a constructor argument. 3 contructors up there is `Costmap2DROS();` That is what should be modified with `NodeOptions` and set `is_lifecycle_follower_` in the method (e.g. not a argument)

SteveMacenski (changes requested):

SteveMacenski:
There are a whole suite of linters in ament. One way to run everything and the unit tests `colcon test --packages-select <your pkgs>` and `colcon test-result --verbose` to see failures

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
GoesM: i'm not very familiar with `NodeOptions`. you mean add following code into `Costmap2DROS::init()`?
```cpp
declare_parameter("is_lifecycle_follower", rclcpp::ParameterValue(false));
```

however,, i'm not clear how to set it into `true` when declaring a costmap node
```cpp
  costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
    "local_costmap", std::string{get_namespace()}, "local_costmap",
    get_parameter("use_sim_time").as_bool()); 
```
for example, in `controller_service.cpp`, how should I modify this declaration to make `is_lifecycle_follower_` being `true`?

SteveMacenski: [I linked what you should do](https://github.com/ros-planning/navigation2/blob/4bc3e8a534005e983ee7a49a09ec00aeb9820661/nav2_planner/include/nav2_planner/planner_server.hpp#L58)

SteveMacenski: `is_lifecycle_follower_` is a member, not a parameter. Just set it in that default contractor to `false` since that's when it is not a follower. Else, default the value in the hpp file to true since all users of other constructors are automatically not independent nodes.

GoesM: > default the value in the hpp file to true since all users of other constructors are automatically not independent nodes.

I'm not very sure about it. I default the value in the hpp file to `false` because I think it's more common to regard it as an independent node, just like how unit test does ?

GoesM: I'm so unclear about your suggested changes that I need your help to point out why our latest code is wrong....

GoesM:
I'm not sure my reply was sent here successfully or not, so I send again.

>default the value in the hpp file to true since all users of other constructors are automatically not independent nodes.

I think such defaulted value in the hpp file may conflict with current unit test ? Current test will declare costmap within defaulted values and regard `costmap` as a not independent node but they launched it indepenedently? 


so sorry that I'm still not clear enough about your suggested changes, so that I'd appreciate it if your could help to point out any misunderstanding in our latest code.

SteveMacenski:
1. change `Costmap2DROS();` to `Costmap2DROS(const rclcpp::NodeOptions & options = rclcpp::NodeOptions());` in hpp
2. Change that same method implementation in the cpp file and add in `is_lifecycle_follower_ = false;`
3. Add an inline comment about it in that constructor why its set to false, what that means, and why its false now only

nav2_controller/src/controller_server.cpp
```diff
@@ -290,11 +290,12 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
    * unordered_set iteration. Once this issue is resolved, we can maybe make a stronger
    * ordering assumption: https://github.com/ros2/rclcpp/issues/2096
    */
-  if (costmap_ros_->get_current_state().id() ==
-    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
-  {
-    costmap_ros_->deactivate();
-  }
+  /*
+   * The costmap is changed into is_lifecycle_follower_{true}
```
SteveMacenski, line 294: Remove comment

nav2_planner/src/planner_server.cpp
```diff
@@ -218,11 +218,12 @@ PlannerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
...
    * ordering assumption: https://github.com/ros2/rclcpp/issues/2096
    */
-  if (costmap_ros_->get_current_state().id() ==
-    lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
-  {
-    costmap_ros_->deactivate();
-  }
+  /*
+   * The costmap is changed into is_lifecycle_follower_{true}
+   * https://github.com/ros-planning/navigation2/pull/3972
```
SteveMacenski, line 223: remove comment

SteveMacenski (changes requested):

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -71,6 +71,15 @@ Costmap2DROS::Costmap2DROS()
...
   declare_parameter("map_topic", rclcpp::ParameterValue(std::string("map")));
+  /**
+   * @brief
+   * If launched as an independent node, set `is_lifecycle_follower_  = true`.
+   * Thus, it would react to rcl_preshutdown.
+   * Else if launched in a thread created by another LifecycleNode(its parent), set
+   * `is_lifecycle_follower_  = false`. It would not react to rcl_preshutdown anymore
+   * and its lifecycle state is controlled by its parent.
+   */
+  get_parameter("is_lifecycle_follower", is_lifecycle_follower_);
```
SteveMacenski, line 82: Not a parameter. Just set it to `false`, we know its false if this constructor is used. Its used when launched as an individual node and/or composition node. The other constructor's given their arguments require it to be a module of another system.

SteveMacenski (changes requested):

SteveMacenski (changes requested):

nav2_costmap_2d/test/unit/lifecycle_test.cpp
```diff
@@ -51,4 +51,4 @@ TEST(LifecylceTest, CheckInitialTfTimeout) {
   if (spin_thread.joinable()) {
     spin_thread.join();
   }
-}
+}
```
SteveMacenski, line 54: EOF line missing

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +135,23 @@ class Costmap2DROS : public nav2_util::LifecycleNode
    */
   nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
+  void on_rcl_preshutdown() override
```
SteveMacenski, line 138: This needs a doxygen description, didn't there used to be on there?

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -71,6 +71,15 @@ Costmap2DROS::Costmap2DROS()
     "nav2_costmap_2d::InflationLayer"}
 {
   declare_parameter("map_topic", rclcpp::ParameterValue(std::string("map")));
+  /**
+   * @brief
```
SteveMacenski, line 75: Remove. Again _only_ add the following to this function:

`is_lifecycle_follower_ = false;`

That is it. No comments, no parameters, nothing.

SteveMacenski (changes requested):

GoesM:
The failure of CI test may show that some unit tests may need to change their constructor choice ?

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +135,30 @@ class Costmap2DROS : public nav2_util::LifecycleNode
    */
   nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
+  /**
+   * @brief as a child-LifecycleNode :
+   * sometimes costmap may be launched by another LifecycleNode and work as a child-thread
+   * child-LifecycleNodes should wait for their parents to handle the shutdown, which includes this module
+   * Thus, it's neccessary to set NodeOption is_lifecycle_follower_ as true
```
SteveMacenski, line 142:
```suggestion
   * Thus, it's neccessary to set NodeOption is_lifecycle_follower_ as true
```

Delete.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +135,30 @@ class Costmap2DROS : public nav2_util::LifecycleNode
    */
   nav2_util::CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
+  /**
+   * @brief as a child-LifecycleNode :
+   * sometimes costmap may be launched by another LifecycleNode and work as a child-thread
```
SteveMacenski, line 140:
```suggestion
   * Costmap2DROS may be launched by another Lifecycle Node as a composed module
```

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -134,6 +135,30 @@ class Costmap2DROS : public nav2_util::LifecycleNode
...
+   * child-LifecycleNodes should wait for their parents to handle the shutdown, which includes this module
```
SteveMacenski, line 141:
```suggestion
   * If composed, its parents will handle the shutdown, which includes this module
```

SteveMacenski (changes requested):

SteveMacenski:
Sorry, just some changes to be clear in technical language

GoesM:
great of thanks for your patience ~

SteveMacenski (approved):

SteveMacenski:
@GoesM I retriggered CI a few times, `lifecycle_test` is failing deterministically with this PR. I think you need to look into that

SteveMacenski:
Great! Thanks for the iteration!

PR #3973, opened 2023-11-20
Move isValidVelocity back to kinematic_parameters

IPA-MKI:
---

## Basic Info

| ------ | ----------- |

---

[isValidVelocity](https://github.com/ros-planning/navigation2/blob/main/nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp#L95) should not be in xy_theta_iterator. It is only related to kineamtic_parameters and should be made available for use without xy_theta_iterator. In ROS1 (robot_navigation) this also used to be in [kinematic_parameters](https://github.com/locusrobotics/robot_navigation/blob/noetic/dwb_plugins/src/kinematic_parameters.cpp#L116). 

## NOTE
If you decide to keep the function in xy_theta_iterator you should cherry-pick the first commit of this pr which fixes a small bug that currently has no effect.

---

#### For Maintainers: 

### DIFF
--- a/nav2_dwb_controller/dwb_plugins/include/dwb_plugins/kinematic_parameters.hpp
+++ b/nav2_dwb_controller/dwb_plugins/include/dwb_plugins/kinematic_parameters.hpp
@@ -35,12 +35,13 @@
 #ifndef DWB_PLUGINS__KINEMATIC_PARAMETERS_HPP_
 #define DWB_PLUGINS__KINEMATIC_PARAMETERS_HPP_
 
+#include <limits>
 #include <memory>
 #include <string>
 #include <vector>
 
-#include "rclcpp/rclcpp.hpp"
 #include "nav2_util/lifecycle_node.hpp"
+#include "rclcpp/rclcpp.hpp"
 
 namespace dwb_plugins
 {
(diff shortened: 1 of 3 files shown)


SteveMacenski (commented):
What is the motivation for this change?

IPA-MKI:
@SteveMacenski
Motivation for this change: I have written a custom TrajectoryGenerator plugin. I use the KinematicsHandler for the parameters but I don't use the XYThetaIterator. Still I want to use the function isValidSpeed which confirms if a twist is within the limits of KinematicParameters. XYThetaIterator is also using this function to confirm twists generated by the iterator but this function does not use any members of XYThetaIterator, just members of KinematicsHandler (if it weren't for the "bug").

nav2_dwb_controller/dwb_plugins/src/xy_theta_iterator.cpp
```diff
@@ -104,7 +104,7 @@ bool XYThetaIterator::isValidSpeed(double x, double y, double theta)
   {
     return false;
   }
-  if (vmag_sq == 0.0 && th_it_->getVelocity() == 0.0) {
+  if (vmag_sq == 0.0 && theta == 0.0) {
```
IPA-MKI: @SteveMacenski the function should not use the member but the function parameter. The only place this function is called is in line 115 where `th_it_->getVelocity()` is passed as `theta`.

IPA-MKI:
Where do i find the right .clang-format file to fix linting errors?

SteveMacenski:
ament uncrustify, cpplint have the profiles built in, just run the commands `ament_cpplint` and `ament_uncrustify`

nav2_dwb_controller/dwb_plugins/include/dwb_plugins/kinematic_parameters.hpp
```diff
@@ -74,6 +75,30 @@ struct KinematicParameters
...
+   * This is based on three parameters: min_speed_xy, max_speed_xy and
+   * min_speed_theta. The speed is valid if 1) The combined magnitude hypot(x,y)
+   * is less than max_speed_xy (or max_speed_xy is negative) AND 2) min_speed_xy
+   * is negative or min_speed_theta is negative or hypot(x,y) is greater than
+   * min_speed_xy or fabs(theta) is greater than min_speed_theta.
+   */
+  inline bool isValidSpeed(double x, double y, double theta)
+  {
+    double vmag_sq = x * x + y * y;
+    if (max_speed_xy_ >= 0.0 && vmag_sq > max_speed_xy_sq_ + std::numeric_limits<float>::epsilon())
```
SteveMacenski, line 93: Readd all brackets, this will not be accepted without them

SteveMacenski (changes requested):

SteveMacenski:
Otherwise I approve!

nav2_dwb_controller/dwb_plugins/include/dwb_plugins/kinematic_parameters.hpp
IPA-MKI: ament uncrustify did the trick, thanks!

SteveMacenski (approved):

PR #3980, opened 2023-11-21
clear start cell for planning

jwallace42:
---

## Basic Info

| ------ | ----------- |

---

This pr clears the start location in the global costmap of all planners. 

---

#### For Maintainers: 


nav2_smac_planner/src/a_star.cpp
```diff
@@ -234,11 +234,6 @@ bool AStarAlgorithm<NodeT>::areInputsValid()
     throw nav2_core::GoalOccupied("Goal was in lethal cost");
   }
 
-  // Check if starting point is valid
-  if (!_start->isNodeValid(_traverse_unknown, _collision_checker)) {
```
SteveMacenski, line 238: Why not keep if it'll auto be OK?

nav2_smac_planner/test/test_a_star.cpp
```diff
@@ -92,10 +92,6 @@ TEST(AStarTest, test_a_star_2d)
   a_star_2.setCollisionChecker(checker.get());
   num_it = 0;
   EXPECT_THROW(a_star_2.createPath(path, num_it, tolerance), std::runtime_error);
-  a_star_2.setStart(50, 50, 0);  // invalid
```
SteveMacenski, line 95: I think this would remove some test coverage - I think this should still be OK and cause a failure since we still can't find a valid plan, no?

SteveMacenski:
You see that this fixes things in rviz in the sandbox? :heart:

nav2_smac_planner/src/a_star.cpp
jwallace42: I would argue that it misleads the reader. But I can add it back.

nav2_smac_planner/test/test_a_star.cpp
jwallace42: We don't throw a no valid plan inside a_star. It happen in the smac_planner. 

It shouldn't remove test coverage because that branch is removed.

jwallace42:
> You see that this fixes things in rviz in the sandbox? ❤️

Yeah, 

I don't see failures due to the planner server. :)

SteveMacenski:
Yay!

nav2_smac_planner/src/a_star.cpp
SteveMacenski: I can see that. How about putting a comment inline then saying we don't check due to the clearing under start condition? Something that makes me in 5 years not look at this and wonder why I skipped it while skimming

jwallace42: Added in the comment. 

The error that would be throw in the case you are mentioning is `NoValidPathCouldBeFound`. 

So, 

You could say that if the expansions < number of neighbors(assuming no prunning) then the start is invalid.
 
If we expect that the robot could be a few cells deep of lethal cost then we probably should clear the whole footprint of the robot rather than just a cell.

nav2_smac_planner/src/smac_planner_2d.cpp
```diff
@@ -217,6 +217,7 @@ nav_msgs::msg::Path SmacPlanner2D::createPlan(
             std::to_string(start.pose.position.y) + ") was outside bounds");
   }
   _a_star->setStart(mx_start, my_start, 0);
+  _a_star->clearStart(mx_start, my_start);
```
SteveMacenski, line 220: Move to `areInputsValid` below the comment, that seems to organize it better. Update for the 3 planners. Adding more external APIs to users of the A* is not ideal, though if there's a good reason its not the end of the world

nav2_smac_planner/src/a_star.cpp
SteveMacenski: > You could say that if the expansions < number of neighbors(assuming no prunning) then the start is invalid.

Exactly. Though I think if you tested it, it might actually end up being just `0` or `1`, since the getNeighbors would return no possible motion, so we wouldn't be `iterations++`-ing. I'd test to verify, but I think it should be one of those two. If not, then yea, less than the number of neighbors is a reasonable choice

nav2_smac_planner/src/smac_planner_2d.cpp
jwallace42: What would you prefer?

SteveMacenski: I'd prefer if clearStart was in `areInputsValid` for Smac

SteveMacenski:
Make sure to rebase / pull in main. CI is failing on a bunch of stuff that should be handled in main now

SteveMacenski:
any word? This would be good to get done ASAP

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -322,7 +322,6 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
     _collision_checker.setCostmap(costmap);
   }
 
-  // Set collision checker and costmap information
```
SteveMacenski, line 325: Why delete comment

SteveMacenski (approved):
Otherwise LGTM

SteveMacenski (approved):

skrinjarf:
I know it's been a long time since this was merged, but I am having a hard time understanding why this change was introduced? Isn't the end effect the same with and without this change? 
Ideologically it seems bad to me that planner modifies the costmap? shouldn't we plan in the exact space that we are? 
What if for some reason the robot gets locked on the obstacle right at the edge of the obstacle. Then clearing that start cell would make the robot unstuck and we would be able to plan our way out in cases where we shouldn't have.

tonynajjar:
@jwallace42 what was the context of this PR? Does it solve a bug?

I find it counter-intuitive for my planner to successfully plan when my start pose in in a lethal space

image

PR #3988, opened 2023-11-27
BT nodes not able to accept parameter values

maksymdidukh:
---

## Basic Info

| ------ | ----------- |

---

## Reasong for change
There are BT nodes that read the input ports from the constructor. This causes issues when we want to pass the variables as input.

- move every `getInput()` usage from constructor to `tick()` method  

---

#### For Maintainers: 


SteveMacenski:
I'm happy to review this in detail once you can actually build it :wink:

maksymdidukh:
> I'm happy to review this in detail once you can actually build it 😉
> 
> From a quick glance, you're moving input getting from the constructors to the ticks, which I assume is because you're varying some or all of those inputs (eg non-const)? I'm kind of surprised for many of them that you'd actually change them mid-operation, but who am I to argue.
> 
> The only ones I have somewhat substantial beef with are those that are moving the creation of subscriptions in `tick`. Those should be existing already and its impractical to do what you're doing. Those nodes just won't work if you're recreating subscriptions every 100 hz tick.

sure! I'll remove PR from the draft when it's ready for review :)

SteveMacenski:
OK. Much of your linting issues (though I'm sure you have a few of your own) will be fixed by rebasing or pulling in main. Some other things large-number of linting problems were fixed recently.

SteveMacenski:
Hey, just wanted to check in. Any questions I can help answer? It looks like you're having alot of build failures. You may want to develop locally instead of using this PR's CI as the builder to speed it up :laughing: We do have limited build resources... you're not nearing that individually, but we do ride that knifes edge in Nav2 given how active the project is, so if we can find another workflow, that would be preferable :-)

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/action/assisted_teleop_action.cpp
```diff
@@ -24,23 +24,33 @@ AssistedTeleopAction::AssistedTeleopAction(
...
-  getInput("time_allowance", time_allowance);
-  getInput("is_recovery", is_recovery_);
-
-  // Populate the input message
-  goal_.time_allowance = rclcpp::Duration::from_seconds(time_allowance);
 }
 
 void AssistedTeleopAction::on_tick()
 {
+  if (!initialized_) {
```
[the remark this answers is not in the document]
SteveMacenski: This is just shuffling code from the constructor into the tick but still executed once on first BT tick which would be microseconds after the constructors are created.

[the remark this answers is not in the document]
maksymdidukh: Perhaps I didn't write it quite accurately in the description of the PR, but the problem I'm trying to solve in this PR is as follows: the BT node constructor is run when BT is initialized, and the blackboard variables are not yet defined, so there are cases when `getInput()`, called in the constructor, reads a value that has not yet been initialized, that is, 0 is read. By moving `getInput()` to the `tick()` method, we ensure that the blackboard variables used by this node are already defined. In most nodes, I use the `initialize()` method to move the ports initialization from the constructor, and the `bool initialized_` variable to make sure we initialize the ports only once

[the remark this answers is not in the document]
SteveMacenski: Ah got it. Give me a couple of days to fully review this. There's alot of similar things happening and I want to make sure by eyes don't glaze over by trying to do them all in a single sitting.

nav2_behavior_tree/plugins/condition/transform_available_condition.cpp
```diff
@@ -72,6 +65,22 @@ BT::NodeStatus TransformAvailableCondition::tick()
   return BT::NodeStatus::FAILURE;
 }
 
+void TransformAvailableCondition::initialize()
```
SteveMacenski, line 68: Move function above tick

nav2_behavior_tree/plugins/condition/time_expired_condition.cpp
```diff
@@ -55,6 +57,14 @@ BT::NodeStatus TimeExpiredCondition::tick()
   return BT::NodeStatus::SUCCESS;
 }
 
+void TimeExpiredCondition::initialize()
```
SteveMacenski, line 60: Ditto: move above tick

nav2_behavior_tree/plugins/condition/is_battery_low_condition.cpp
```diff
@@ -27,7 +27,25 @@ IsBatteryLowCondition::IsBatteryLowCondition(
...
+  }
+
+  callback_group_executor_.spin_some();
+  if (is_battery_low_) {
+    return BT::NodeStatus::SUCCESS;
+  }
+  return BT::NodeStatus::FAILURE;
+}
+
+void IsBatteryLowCondition::initialize()
```
SteveMacenski, line 48: The creation of the callback ground / subscription needs to happen in the the constructor to give it a little extra time for DDS to deal with discovery before the first tick.

nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp
```diff
@@ -24,7 +24,25 @@ IsBatteryChargingCondition::IsBatteryChargingCondition(
   const BT::NodeConfiguration & conf)
 : BT::ConditionNode(condition_name, conf),
   battery_topic_("/battery_status"),
-  is_battery_charging_(false)
+  is_battery_charging_(false),
+  initialized_(false)
```
SteveMacenski, line 28: Ditto

nav2_behavior_tree/plugins/action/truncate_path_action.cpp
```diff
@@ -31,14 +31,20 @@ TruncatePath::TruncatePath(
...
+  distance_(1.0),
+  first_time_(true)
 {
-  getInput("distance", distance_);
 }
 
 inline BT::NodeStatus TruncatePath::tick()
 {
   setStatus(BT::NodeStatus::RUNNING);
+  getInput("distance", distance_);
```
SteveMacenski, line 42: You get distance here -- no need for the first time stuff if you're getting it just a line above every iteration

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/condition/transform_available_condition.cpp
SteveMacenski: ^ do this for all nodes, just do I don't repeat this a ton of times

nav2_behavior_tree/plugins/condition/is_battery_low_condition.cpp
SteveMacenski: Ditto in a few different nodes

nav2_behavior_tree/plugins/action/truncate_path_action.cpp
maksymdidukh: done

nav2_behavior_tree/plugins/condition/transform_available_condition.cpp
maksymdidukh: done

nav2_behavior_tree/plugins/condition/is_battery_low_condition.cpp
maksymdidukh: working on it

SteveMacenski:
LMK when you're done with the last item!

nav2_behavior_tree/plugins/condition/is_battery_low_condition.cpp
maksymdidukh: In this case, we decided to stick with the old implementation and here's why:
- we have to create a subscription in the constructor, and this node will be permanently subscribed to only one topic => the user has to use this node as follows:  
`<IsBatteryLowCondition topic_name="some_topic">`, not like this: 
`<IsBatteryLowCondition topic_name="{some_topic}">`

So, I'll revert changes in all nodes where we create a subscription in the constructor

maksymdidukh:
> LMK when you're done with the last item!

Hey, I'm done!

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -29,22 +29,33 @@ RemovePassedGoals::RemovePassedGoals(
...
+
+void RemovePassedGoals::initialize()
 {
   getInput("radius", viapoint_achieved_radius_);
 
   tf_ = config().blackboard->get<std::shared_ptr<tf2_ros::Buffer>>("tf_buffer");
   auto node = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
   node->get_parameter("transform_tolerance", transform_tolerance_);
 
+  global_frame_ = BT::deconflictPortAndParamFrame<std::string, RemovePassedGoals>(
```
SteveMacenski, line 44: I believe this was removed. Maybe you're working on an old commit?

SteveMacenski (changes requested):

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
```diff
@@ -68,6 +68,7 @@ class IsPathValidCondition : public BT::ConditionNode
   // The timeout value while waiting for a responce from the
   // is path valid service
   std::chrono::milliseconds server_timeout_;
+  bool first_time_;
```
jwallace42, line 71: I would suggest keeping the pattern you have established with all the other nodes. 

`bool initialized`
`initialized()`

nav2_behavior_tree/plugins/decorator/rate_controller.cpp
```diff
@@ -26,13 +26,14 @@ RateController::RateController(
...
   first_time_(false)
 {
-  double hz = 1.0;
-  getInput("hz", hz);
-  period_ = 1.0 / hz;
 }
 
 BT::NodeStatus RateController::tick()
 {
+  double hz = 1.0;
```
jwallace42, line 33: Why is this not wrapped by a check for initialization?

jwallace42 (changes requested):
It looks like you are missing

[bt_cancel_node](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp#L59,)

[bt_action_node](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L61.)

[planner_selector](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/action/planner_selector_node.cpp#L41.)

There are probably a few more missing so I would double check all the bt nodes and verify that there no `getInput` calls in the constructors.

nav2_behavior_tree/plugins/decorator/rate_controller.cpp
maksymdidukh: done

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
maksymdidukh: done

maksymdidukh:
> It looks like you are missing
> 
> [bt_cancel_node](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp#L59,)
> 
> [bt_action_node](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L61.)
> 
> [planner_selector](https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/plugins/action/planner_selector_node.cpp#L41.)
> 
> There are probably a few more missing so I would double check all the bt nodes and verify that there no `getInput` calls in the constructors.

bt_cancel_node and bt_action_node are fixed, but I added an extra field called `xml_tag` to log debugging messages in the initialize method, but I was thinking about logging it in the constructor? Let me know if I should change this.

And about the planner selector, we decided not to move `getInput()` from the constructor to `tick()` in case we create a subscription in the constructor, check this [comment](https://github.com/rosplanning/navigation2/pull/3988#discussion_r1426747708)

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
maksymdidukh: fixed

SteveMacenski:
I'm on the fence about the action / service nodes. 

On one hand it would be nice to actually resolve some subtree issues if we didn't get those ports in the constructor. It also then makes it no different from the other BT nodes. 

On the other hand, it delays the construction of the action client to give very little time for discovery to occur before calling the action and pushing the `wait_for_action_server` into the BT's execution context -- which seems like a poor choice on both counts especially since we know users have action servers on other computers in the network pretty regularly. 

I'm leaning towards asking you to revert the changes in those two files, but I could be convinced if someone has a strong outside reason why we should enact that change given the trade off.

maksymdidukh:
> I'm on the fence about the action / service nodes.
> 
> On one hand it would be nice to actually resolve some subtree issues if we didn't get those ports in the constructor. It also then makes it no different from the other BT nodes.
> 
> On the other hand, it delays the construction of the action client to give very little time for discovery to occur before calling the action and pushing the `wait_for_action_server` into the BT's execution context -- which seems like a poor choice on both counts especially since we know users have action servers on other computers in the network pretty regularly.
> 
> I'm leaning towards asking you to revert the changes in those two files, but I could be convinced if someone has a strong outside reason why we should enact that change given the trade off. I'd say for now to revert those two files so that we can merge in all the rest. Then I suppose we could have a follow up PR or discussion about it if someone disagrees. We don't have to have the controversial stuff happen right now to get 80% of the benefit immediately.



> I'm on the fence about the action / service nodes.
> 
> On one hand it would be nice to actually resolve some subtree issues if we didn't get those ports in the constructor. It also then makes it no different from the other BT nodes.
> 
> On the other hand, it delays the construction of the action client to give very little time for discovery to occur before calling the action and pushing the `wait_for_action_server` into the BT's execution context -- which seems like a poor choice on both counts especially since we know users have action servers on other computers in the network pretty regularly.
> 
> I'm leaning towards asking you to revert the changes in those two files, but I could be convinced if someone has a strong outside reason why we should enact that change given the trade off. I'd say for now to revert those two files so that we can merge in all the rest. Then I suppose we could have a follow up PR or discussion about it if someone disagrees. We don't have to have the controversial stuff happen right now to get 80% of the benefit immediately.

reverted)

maksymdidukh:
When this PR is merged, will we be able to make a backport to iron?

SteveMacenski (approved):
@jwallace42 any additional comments?

jwallace42 (approved):
LGTM!

maksymdidukh:
> Edit: you have a test failure:

Cannot reproduce test failure locally, I'll rerun the tests, maybe it was an accident
image

SteveMacenski:
Thanks for the contribution!

redvinaa:
> We have some examples of doing such things https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/plugins/decorator/goal_updated_controller.cpp#L35-L45 that we could duplicate here. Is that the right move in conjunction with your suggestion to move the status setting?

Yes, this is the same solution as what I suggested. As this is a decorator node, and doesn't inherit from `BtActionNode`, it does actually get the IDLE state.

Pro for the initialization fix is that it can't break existing nodes. But also it has to be implemented separately in every single bt node. And this quick fix just takes us deeper down into the rabbit hole.

Issue #3861, opened 2023-10-05
Add collision monitor approach model on by default in configs


SteveMacenski:
That way we can have it leveraged out of the box for users & show for example how to use it!

SteveMacenski:
Change to stopped goal checker too>?

SteveMacenski:
In progress - to be submitted PR today.

PR #3989, opened 2023-11-28
adding default collision monitor to bringup

SteveMacenski:
#3861


SteveMacenski:
^ need to add collision monitor to parameters files missing

PR #3990, opened 2023-11-28
Nav2 3954 consider robot footprint

jwallace42:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | 3954 |
| Robotic platform tested on | gazebo|

---

This pr adds functionality to invalidate a path based based on footprint checks. This functionality was added to the `isPathValid` service. A new parameter called `is_path_valid_check_footprint` was added to turn on/off the full footprint checks. 

---

#### For Maintainers: 


nav2_planner/src/planner_server.cpp
```diff
@@ -656,16 +666,58 @@ void PlannerServer::isPathValid(
...
+
+        geometry_msgs::msg::PolygonStamped stamped_footprint_;
+        for (const auto point : oriented_footprint) {
+          geometry_msgs::msg::Point32 point_msg;
+          point_msg.x = point.x;
+          point_msg.y = point.y;
+          stamped_footprint_.polygon.points.push_back(point_msg);
+        }
+        stamped_footprint_.header.frame_id = "map";
+        failed_footprint_->publish(stamped_footprint_);
```
SteveMacenski, line 717: Do you intend to keep the publisher or is that just a debugging artifact?

nav2_planner/src/planner_server.cpp
```diff
@@ -85,6 +88,9 @@ PlannerServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
 
   costmap_ros_->configure();
   costmap_ = costmap_ros_->getCostmap();
+  collision_checker_ =
```
SteveMacenski, line 91: Only need to create if `is_path_valid_check_footprint`

nav2_planner/src/planner_server.cpp
```diff
@@ -62,6 +63,8 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
     }
   }
 
+  get_parameter("is_path_valid_check_footprint", is_path_valid_check_footprint_);
```
SteveMacenski, line 66: Is there a standard parameter name we use for checking full footprint? If nothing else, drop the `is_` up front to shorten it a hair

SteveMacenski: ^ would be potentially be valuable to use the costmap's information about whether its a radius or a footprint to make that call instead of making it a parameter?

SteveMacenski (changes requested):

nav2_planner/src/planner_server.cpp
jwallace42: I used it to verify my work :). I think it would be useful to keep around. However, I would clean it up and only pub on sub.

SteveMacenski: What utility would this have beyond testing your code as working? I'm not sure how someone would use that in some workflow that would be useful, but maybe I'm not being imaginative enough

jwallace42: Removed :)

PR #3997, opened 2023-12-02
Add ability to visualize the state lattice tree expansions

embeddedadam:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (None) |
| Robotic platform tested on | (None) |

---

Added debug flag in smac_planner_lattice.cpp that enables visualization of state lattice tree expansions

* Added new parameter in smac_planner_lattice _debug_visualizations
* Added visualisation like in hybrid A* planner after _debug_visualizations is set

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/smac_planner_lattice.cpp
+++ b/nav2_smac_planner/src/smac_planner_lattice.cpp
@@ -287,6 +287,11 @@ nav_msgs::msg::Path SmacPlannerLattice::createPlan(
   if (!_a_star->createPath(
       path, num_iterations, _tolerance / static_cast<float>(_costmap->getResolution())))
   {
+    // Note: If the start is blocked only one iteration will occur before failure
+    if (num_iterations == 1) {
+      throw nav2_core::StartOccupied("Start occupied");
+    }
+
     if (num_iterations < _a_star->getMaxIterations()) {
       throw nav2_core::NoValidPathCouldBeFound("no valid path found");
     } else {
--- a/nav2_smac_planner/src/smac_planner_lattice.cpp
+++ b/nav2_smac_planner/src/smac_planner_lattice.cpp
@@ -126,7 +126,10 @@ void SmacPlannerLattice::configure(
   nav2_util::declare_parameter_if_not_declared(
     node, name + ".allow_reverse_expansion", rclcpp::ParameterValue(false));
   node->get_parameter(name + ".allow_reverse_expansion", _search_info.allow_reverse_expansion);
-
+  nav2_util::declare_parameter_if_not_declared(
+    node, name + ".debug_visualizations", rclcpp::ParameterValue(false));
+  node->get_parameter(name + ".debug_visualizations", _debug_visualizations);
+  
   _metadata = LatticeMotionTable::getLatticeMetadata(_search_info.lattice_filepath);
   _search_info.minimum_turning_radius =
     _metadata.min_turning_radius / (_costmap->getResolution());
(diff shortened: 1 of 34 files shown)


nav2_smac_planner/src/smac_planner_lattice.cpp
```diff
@@ -282,11 +298,27 @@ nav_msgs::msg::Path SmacPlannerLattice::createPlan(
...
+  std::unique_ptr<std::vector<std::tuple<float, float, float>>> expansions = nullptr;
+  if (_debug_visualizations) {
+    expansions = std::make_unique<std::vector<std::tuple<float, float, float>>>();
+  }
   // Note: All exceptions thrown are handled by the planner server and returned to the action
   if (!_a_star->createPath(
       path, num_iterations, _tolerance / static_cast<float>(_costmap->getResolution())))
   {
+    if (_debug_visualizations) {
+      geometry_msgs::msg::PoseArray msg;
```
SteveMacenski, line 310: These 2 blocks are not covered by any unit or system tests, we need to have that in order to merge the PR for quality control requirements

SteveMacenski (commented):
Can you show what this output looks like in rviz?

nav2_smac_planner/src/smac_planner_lattice.cpp
SteveMacenski: The WPF test failed oddly, so I retriggered CI to see if that was just a fluke

SteveMacenski:
The new parameter needs to go into the readme / [configuration guide](https://navigation.ros.org/configuration/packages/smac/configuring-smac-lattice.html)

mikeferguson:
As a general note, that lattice might become VERY large ... and very large MarkerArrays can definitely tank performance in RVIZ.

In most cases where I've seen visualization of a state lattice tree, it has been done with a OccupancyGrid type message, which simply shows which cells have had related states expanded within them. This definitely works for debugging things like "my heuristic is messed up and I'm expanding states in the wrong place" which I feel is one of the most common issues, but wouldn't help as much in debugging heading-specific issues.

mikeferguson (commented):
As a general note, that lattice might become VERY large ... and very large MarkerArrays can definitely tank performance in RVIZ.

In most cases where I've seen visualization of a state lattice tree, it has been done with a OccupancyGrid type message, which simply shows which cells have had related states expanded within them. This definitely works for debugging things like "my heuristic is messed up and I'm expanding states in the wrong place" which I feel is one of the most common issues, but wouldn't help as much in debugging heading-specific issues.

SteveMacenski:
Huh, that's an interesting idea. 

I can say that the Hybrid-A* expansions show up fine but starts to lag down rviz a _little bit._ This would introduce comparatively less of those since the individual primitives are so much longer in the state lattices than in Hybrid-A*. Right now as the PR stands, it only does the end-points of the lattices, not the intermediary poses that are used to populate the normalized dense path. I think due to that, and including the orientation information which can be important debugging information, I think this PR is OK as it is using the markers.

I definitely don't disagree though that if we want to publish the full lattice samples that we should use the method you describe. Basically that would just entail iterating over them and marking each cell a primitive lands in, which should be straight forward to implement. I think that would be a lovely additional capability which shows another part of the expansion pattern which might yield insights.

SteveMacenski:
@embeddedadam this PR now has commits from the main branch - this is not rebased / merged in correctly to only have the software you've created for this PR in your tree. I'm not able to review this until that is resolved. Try rebasing or merging in changes more standardly. 

Also, please provide the rviz screen shot I requested for what this looks like! :smile:

SteveMacenski:
Hey! Any update? I'd love to have this in :-)

SteveMacenski:
Working on this now to bring it home - PR to open shortly

Issue #3924, opened 2023-10-30
Loopback Simulator Package 


SteveMacenski:
Offer a simplistic Gazebo replacement which sends robot velocity commands to TF as the robot's odometry to completely remove the need for complex simulation for testing higher level behavioral attributes not related to perception or control.

This would be unable to generate sensor data, only robot motion, so in using it the documentation should mention that a map is required and have the static map in the local costmap to avoid obstacles. AMCL will probably need to be disabled in this mode and the simulator just publish the robot's map-odom pose since we know perfectly where it is from initial pose + perfect odometry. 

A nice thing about this is that you can spawn the robot anywhere for some deterministic tests repeatedly 

This would be a nice medium sized contribution from an outside contributor looking to make an impact!

smitdumore:
Hi @SteveMacenski,

I'm excited about the prospect of contributing to the project, and I'm reaching out to express my interest in tackling this issue.
This would mark my first venture into the realm of open source contributions. While I boast intermediate proficiency in ROS1, I'm in the early stages of acquainting myself with ROS2, with a commitment to learning and growing along the way.

For a bit of background, you can explore my portfolio here: smitdumore.github.io

I'm wondering if I could take up the aforementioned issue ? 

Thank you.

Thanks,
[Your Name]

SteveMacenski:
Absolutely! You happen to have found something I'm intentionally not going to build with the goal of it being a good first issue for a new contributor that's interested in giving back could take on -- and hopefully grow your skills to continue to contribute well afterward :-) 

I think a good place to start is to review this topic and come up with a proposal of a plan of attack. It might be worth getting the basic TB3 demo working [from our docs](https://navigation.ros.org/getting_started/index.html) to have a base-case and see what kinds of things you might need to fake out by looking at topics, TF tree, and the role that odometry + AMCL play in the success of the robot. We can review that with you to make sure we have a good plan in place before you get too deep into the implementation -- but I'd never argue with you experimenting yourself a bit as well. 

**I believe** due to the nature of this package in testing (e.g. not hardware or production use), you could use Python3 to implement this if you so choose. It would not be my preference that you did that, but that option does exist for you uniquely for this project over others -- if you happen to be more comfortable with python than C++. Future projects though would be primarily C++.  

P.S. Your docking experience might be something we can work on too, its on my roadmap with a similar holding off for the right time.

smitdumore:
[...]

PR #4006, opened 2023-12-11
Added Loopback Simulator Package

smitdumore:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #3924 |
| Primary OS tested on | Ubuntu 20.04.4 |
| Robotic platform tested on | Rviz |

---

* This new package serves as a Gazebo replacement.
* Designed for testing robot behaviors without the need for a complicated simulator.
* Uses incoming velocity commands to forward simulate the robot pose.
* A map is needed for navigation. The Robot can be spawned anywhere for repetative tests.

* A detailed readme is required for the nav2_loopback_sim package.

---

* Needs to be tested for various robot platforms and scenarios.
* Needs some nice to have features.

#### For Maintainers: 


nav2_loopback_sim/package.xml
```diff
@@ -0,0 +1,30 @@
...
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>launch</build_depend>
+  
+  <depend>rclcpp</depend>
+  <depend>geometry_msgs</depend>
+  <depend>std_msgs</depend>
+  <depend>tf2_msgs</depend>
+  <depend>tf2_ros</depend>
+  <depend>tf2_ros_msgs</depend>
```
SteveMacenski, line 21: Build failed here, you need to make sure all of these packages exist and are correct. Make sure to look at all of your header includes and cmakelists and make sure everything is in this file here or else it won't compile in the build farm

nav2_loopback_sim/.amentcpplint
```diff
@@ -0,0 +1,4 @@
+filter=-build/c++14
```
SteveMacenski, line 1: Remove file

nav2_loopback_sim/include/nav2_loopback_sim/loopback_simulator.hpp
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2019 Samsung Research America
```
SteveMacenski, line 1: This should be you, silly! Update in all files with 2023 year

nav2_loopback_sim/include/nav2_loopback_sim/loopback_simulator.hpp
```diff
@@ -0,0 +1,93 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NAV2_LOOPBACK_SIM__LOOPBACK_SIMULATOR_HPP_
+#define NAV2_LOOPBACK_SIM__LOOPBACK_SIMULATOR_HPP_
+
+#pragma once
```
SteveMacenski, line 18: You don't need once with the ifndefs

nav2_loopback_sim/include/nav2_loopback_sim/loopback_simulator.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include "tf2/LinearMath/Quaternion.h"
+#include "tf2/convert.h"
+#include "tf2_ros/buffer.h"
+#include "tf2/exceptions.h"
+#include "tf2/transform_datatypes.h"
+
+namespace nav2_loopback_sim
+{
+/**
+ * @class nav2_planner::LoopbackSimulator
```
SteveMacenski, line 40:
```suggestion
 * @class nav2_loopback_sim::LoopbackSimulator
```

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <memory>
+#include "nav2_loopback_sim/loopback_simulator.hpp"
+
+namespace nav2_loopback_sim
+{
+
+LoopbackSimulator::LoopbackSimulator()
+: Node("loopback_simulator_node")
+{
+  // Initialise subscribers
```
SteveMacenski, line 25:
```suggestion
  // Initialize subscribers
```

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+  }
+}
+
+void LoopbackSimulator::timerCallback()
+{
+  if (init_pose_set_ == false) {return;}
+  // RCLCPP_INFO(this->get_logger(), "Publishing map->odom tf");
+
+  // Publish map to odom transform
+  geometry_msgs::msg::TransformStamped map_to_odom_transform;
```
SteveMacenski, line 152: You populate alot of transform stamped messages. Perhaps make a function that does this boiler plate for you given the inputs of the frames and pose.

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+  map_to_odom_transform.header.stamp = this->now();
```
SteveMacenski, line 153: Add 1 `dt` time to the now so that this is future stamped slightly. Its something that AMCL / RL does as well to make sure that the global localization is always valid

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+void LoopbackSimulator::initposeCallback(
+  const
+  geometry_msgs::msg::PoseWithCovarianceStamped::SharedPtr msg)
+{
+  RCLCPP_INFO(this->get_logger(), "Received initial pose wrt map");
+  init_pose_ = *msg;
+  init_pose_set_ = true;
+
+  if (init_odom_base_published_ == false) {
```
SteveMacenski, line 125: Couldn't this just be in the constructor and just remove the need for this variable?

Or I suppose it could go into the timer callback if `init_pose_set_ == false` and also remove this variable and continuously publish the 0-transform until twists start coming in.

How does this handle gaps? E.g. if you have a robot navigating, then stop for a few seconds from success, then start again, do any errors pop up from TF transforms not being valid?

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+    geometry_msgs::msg::TransformStamped odom_to_base_transform;
+    odom_to_base_transform.header.stamp = this->now();
+    odom_to_base_transform.header.frame_id = "odom";
+    odom_to_base_transform.child_frame_id = "base_footprint";
+    odom_to_base_transform.transform.translation.x = 0.0;
+    odom_to_base_transform.transform.translation.y = 0.0;
+    odom_to_base_transform.transform.translation.z = 0.0;
+    tf2::Quaternion q;
+    q.setRPY(0, 0, 0);
+    odom_to_base_transform.transform.rotation.x = q.x();
```
SteveMacenski, line 136: An identity transform for quaternion is X,Y,Z all `0` and w `1`

SteveMacenski (changes requested):

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+  RCLCPP_INFO(this->get_logger(), "Received initial pose wrt map");
+  init_pose_ = *msg;
+  init_pose_set_ = true;
+
+  if (init_odom_base_published_ == false) {
+    // Publish Identity transform between odom to base_footprint
+    geometry_msgs::msg::TransformStamped odom_to_base_transform;
+    odom_to_base_transform.header.stamp = this->now();
+    odom_to_base_transform.header.frame_id = "odom";
+    odom_to_base_transform.child_frame_id = "base_footprint";
```
SteveMacenski, line 130: Use the initial poses' header frame like you do later on

SteveMacenski:
I think there's something funky happening with the frames here. The timerCallback should be publishing map->odom, but its actually publishing the `transform.translation` as the full initial pose's map->base_link. Also `twistCallback` should be publishing odom->base_link (or init_pose_ header id if fine) but there appears to be something funky going on there

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    transform_initpose_once_ = false;
+  }
+
+  // Update the transformed pose in the local variable
+  odom_updated_pose_.header.frame_id = "odom";
+  odom_updated_pose_.header.stamp = this->now();
+
+  tf2::Quaternion base_link_orientation(
```
mikeferguson, line 79: check out tf2::toMsg and fromMsg functions in tf2_geometry_msgs - this will condense a bunch of this code

nav2_loopback_sim/package.xml
```diff
@@ -0,0 +1,29 @@
...
+                focusing on sending robot velocity commands to TF as odometry, 
+                facilitating high-level behavioral testing without the complexity of full simulation</description>
+  <maintainer email="smitd@umd.edu">Smit Dumore</maintainer>
+  <license>Apache-2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>launch</build_depend>
+  
+  <depend>rclcpp</depend>
```
SteveMacenski, line 16: Add `find_package` in cmakelists should be listed as a `depend` here too

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+  //  tf
+  tf_buffer_ = std::make_unique<tf2_ros::Buffer>(this->get_clock());
+  tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
+}
+
+void LoopbackSimulator::twistCallback(const geometry_msgs::msg::Twist::SharedPtr msg)
+{
+  if (init_pose_set_ == false) {return;}
+
+  double dt = 0.1;  // Adjust the time step
```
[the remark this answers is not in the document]
smitdumore: Yes, I'll change it

nav2_loopback_sim/src/loopback_simulator.cpp
smitdumore: Sure

nav2_loopback_sim/src/loopback_simulator.cpp
smitdumore: Got it !

nav2_loopback_sim/src/loopback_simulator.cpp
smitdumore: Yes you are right  `init_odom_base_published_`  is not needed.

We need to have a 0-transform between "odom" and "base_footprint" only at initialization right ? So I think its okay to publish the 0-transform only once. 
Later the transform gets updated based on how the robot moves and we dont want to overwrite that transform with our script.

> if you have a robot navigating, then stop for a few seconds from success, then start again

I'll test this scenario and see what happens

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+  if (init_pose_set_ == false) {return;}
+  // RCLCPP_INFO(this->get_logger(), "Publishing map->odom tf");
+
+  // Publish map to odom transform
+  geometry_msgs::msg::TransformStamped map_to_odom_transform;
+  map_to_odom_transform.header.stamp = this->now();
+  map_to_odom_transform.header.frame_id = "map";
+  map_to_odom_transform.child_frame_id = "odom";
+  map_to_odom_transform.transform.translation.x = init_pose_.pose.pose.position.x;
```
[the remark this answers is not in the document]
smitdumore: This gets executed only when an initial pose estimate is given from Rviz, that means a 0-transform is also published between "odom" and "base_link". 

So it makes more sense to me to continously publish the above transform between map->odom, because odom is a fixed frame

nav2_loopback_sim/src/loopback_simulator.cpp
SteveMacenski: > We need to have a 0-transform between "odom" and "base_footprint" only at initialization right 

Correct, after that it should be updated by the robot's motion

nav2_loopback_sim/src/loopback_simulator.cpp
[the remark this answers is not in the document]
SteveMacenski: OK

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+  publishTransform(init_pose_.pose.pose, "map", "base_link");
+}
+
+void LoopbackSimulator::timerCallback()
+{
+  if (!init_pose_set_) {
+    return;
+  }
+
+  // RCLCPP_INFO(this->get_logger(), "Publishing map->odom tf");
```
SteveMacenski, line 190: Remove please

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+  // Continously publishing map->odom transform
+  publishTransform(relocalization_pose_, "map", "odom");
+
+  // publish bogus lidar data
+  auto scan_msg = std::make_shared<sensor_msgs::msg::LaserScan>();
```
SteveMacenski, line 196: Make this a unique pointer

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+  scan_msg->range_max = 10.0;
+
+  // Fill in the scan data
+  int num_readings = 180;
+  scan_msg->ranges.resize(num_readings);
+  for (int i = 0; i < num_readings; ++i) {
+    scan_msg->ranges[i] = 5.0;  // Set all ranges to 5 meters
+  }
+
+  scan_publisher_->publish(*scan_msg);
```
SteveMacenski, line 217:
```suggestion
  scan_publisher_->publish(std::move(scan_msg));
```

Once its a unique pointer

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+  scan_msg->time_increment = 0.0;
+  scan_msg->scan_time = 0.1;
+  scan_msg->range_min = 0.0;
+  scan_msg->range_max = 10.0;
+
+  // Fill in the scan data
+  int num_readings = 180;
+  scan_msg->ranges.resize(num_readings);
+  for (int i = 0; i < num_readings; ++i) {
+    scan_msg->ranges[i] = 5.0;  // Set all ranges to 5 meters
```
SteveMacenski, line 214: Set to 0 or Inf instead?

nav2_loopback_sim/src/loopback_simulator.cpp
SteveMacenski, line 196: This all feels like it can be made into a separate `sendFakeLaserScan` function

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+  // Updating relocalization pose of odom frame
+  relocalization_pose_.position.x = map_to_odom_transform.transform.translation.x;
+  relocalization_pose_.position.y = map_to_odom_transform.transform.translation.y;
+  relocalization_pose_.position.z = map_to_odom_transform.transform.translation.z;
+  relocalization_pose_.orientation = map_to_odom_transform.transform.rotation;
+
+  // Publishing updated position of odom and base_link wrt map
+  publishTransform(relocalization_pose_, "map", "odom");
+  publishTransform(init_pose_.pose.pose, "map", "base_link");
```
SteveMacenski, line 181: The transform it should be publishing is odom->base_link, the map->base_link should **never** be published. That breaks the TF tree. Its map->odom and odom->base_link. I think that largely invalidates the code above.

- Given the initial pose (map->base_link)
- Find the current odom->base_link transform (or just have it stored from the previous iteration)
- Back out the map->odom that keeps odom->base_link constant 
- Publish that map->odom

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+  // Convert the quaternion in the transform to tf2::Quaternion
+  tf2::Quaternion current_orientation;
+  tf2::fromMsg(transformStamped.transform.rotation, current_orientation);
+
+  // Calculate the change in orientation based on the twist message
+  tf2::Quaternion angular_change;
+  angular_change.setRPY(msg->angular.x * dt_, msg->angular.y * dt_, msg->angular.z * dt_);
+
+  // Integrate the change in orientation with the current orientation
+  tf2::Quaternion new_orientation = current_orientation * angular_change;
```
SteveMacenski, line 106: Shouldn't this be `+`? `t0 + velocity * dt = t1`

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+  new_position.position.x =
+    transformStamped.transform.translation.x + rotated_linear_velocity.x() * dt_;
+  new_position.position.y =
+    transformStamped.transform.translation.y + rotated_linear_velocity.y() * dt_;
+  new_position.position.z =
+    transformStamped.transform.translation.z + rotated_linear_velocity.z() * dt_;
+  new_position.orientation = tf2::toMsg(new_orientation);
+
+  // Publish the updated transform between odom and base_link
+  publishTransform(new_position, "odom", "base_link");
```
SteveMacenski, line 124: This function seems overcomplicated. Why not use TF math more to do the rotations and matrix calculation for you?

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+  if (!init_pose_set_) {
+    RCLCPP_WARN(
+      rclcpp::get_logger("logger_name"),
+      "Initial pose has not been set. Ignoring incoming velocity messages.");
+    return;
+  }
+
+  // get base_link location in the odom frame
+  geometry_msgs::msg::TransformStamped transformStamped;
```
SteveMacenski, line 89: If you just store the odom_to_base_link` transform in the class, you can just update it every time with new twist data. You don't need to look it up since you're the only publisher of that data. If you just store `new_position` as `odom_to_base_link_`, that can be used repeatedly and updated each time. It only just needs to be set to Identity on the first publication and then you're good

nav2_loopback_sim/src/loopback_simulator.cpp
```diff
@@ -0,0 +1,227 @@
...
+  init_pose_subscriber_ = create_subscription<geometry_msgs::msg::PoseWithCovarianceStamped>(
+    "initialpose", 10,
+    std::bind(&LoopbackSimulator::initPoseCallback, this, std::placeholders::_1));
+
+  // Initialize publishers
+  scan_publisher_ = this->create_publisher<sensor_msgs::msg::LaserScan>("scan", 10);
+
+  // Frequency parameter for timer function
+  double timer_frequency{0.0};
+  declare_parameter("timer_frequency", 10.0);
```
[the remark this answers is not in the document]
SteveMacenski: Question: the `dt` should that be run at the same controller rate? What happens if this is set to 0.1, but the controls are published every 0.05s (20hz)?

nav2_loopback_sim/src/loopback_simulator.cpp
smitdumore: I am converting both the orientation and the change in orientation to quaternions. So multiplication is allowed because both are in quaternions.

SteveMacenski: Correct, sorry my quaternion math is clearly rusty (which is especially embarrassing considering I took a a whole attitude control systems class in college)

nav2_loopback_sim/src/loopback_simulator.cpp
smitdumore: Absolutely! That will be so much better

nav2_loopback_sim/src/loopback_simulator.cpp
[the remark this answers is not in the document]
smitdumore: Yes, ideally, when this simulator node is launched, it should get the 'dt' from the controller node. Because if the controller is running at a faster rate, then the simulator is missing out on control commands, which can lead to inaccuracies over time.

SteveMacenski:
Hey! Checking back in here - anything I can help answer?

smitdumore:
Hey Steve, 
I apologize for being inactive. I am wrapping up my last semester and should be able to jump back within a week. Thanks

SteveMacenski:
To supersede shortly

Issue #3789, opened 2023-08-31
[Smac Planner] Enable goal orientation non-specificity


SteveMacenski:
Some users have asked for a way to compute feasible paths to the goal without considering orientation. Since they're feasible paths, there's no "real" way to do this since we need to know the target orientation to make sure its drivable

However, we can brute force this internally by making the check for `isGoal()` consider not just the XY,Yaw bin but all Yaw bins in an XY position. That gets us part of the way there. The other half is in the analytic expansions that actually computes the way to the goal and knows things about the orientation. Each expansion would need to check for each Yaw bin at the goal also to check if any of them are valid (and use the shortest, valid one).

This comes at an increased compute cost obviously, especially for analytic expansions. However its far cheaper than the alternatives some users have proposed that involve multiple entirely separate planning cycles. 

While this is not something Open Navigation is likely to work on directly in the near-term future (except by sponsor request or contracting)  this is a good potential community contribution with a clear plan laid out above to accomplish it.

BriceRenaudeau:
Hi, we have a use case where the robot goal has two possible orientations (yaw & yaw+180°) because the robot can be loaded from both sides. We want to find the fastest path so we compute it twice (one for each yaw) and execute the path we the smallest number of points.

SteveMacenski:
That is a bit of a different problem, but falls into the same category of solutions. The same strategy would work, just using mirrored quantization rather than all quantizations

stevedanomodolor:
I would love to take this task. I would more detail on this if possible so I can start working on it.

SteveMacenski:
@stevedanomodolor this would be a great contribution! I think the description above more or less gives the details. We want the ability to not only plan to an exact pose (X, Y, Yaw) but also perhaps the inverse of that (X, Y, Yaw-180) or at any angle (X, Y, 0-360). 

To do that, we need to change `setGoal` to something like `setGoals` or `setGoal` to include some flag for how to interpret the orientation. Probably the latter makes sense to be more generalized, otherwise if it was just only Yaw and Yaw-180, then setting 2 goals would be EZ-PZ.

Then in A* we need to change the check for isGoal to be OK with any Goal orientation that meets the requests spec. That gets us half the way there. 

Then we need to update the analytic expansion module with similar information about the orientation flexibility to try analytic expansions to the different orientations. If we find any that pass, we return it. If we find multiple that pass, we return the shortest. 

[...]

PR #4019, opened 2023-12-20
[Smac Planner] Enable goal orientation non-specificity

stevedanomodolor:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (#3789) |

---

* WIP

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/constants.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/constants.hpp
@@ -28,6 +28,14 @@ enum class MotionModel
   STATE_LATTICE = 4,
 };
 
+enum class GoalHeading 
+{
+  UNKNOWN = 0,
+  DEFAULT = 1,
+  BIDIRECTIONAL = 2, 
+  ANY_HEADING = 3,
+};
+
 inline std::string toString(const MotionModel & n)
 {
   switch (n) {
@@ -58,6 +66,38 @@ inline MotionModel fromString(const std::string & n)
     return MotionModel::UNKNOWN;
   }
 }
+
+inline std::string toString(const GoalHeading & n)
+{
+  switch (n)
+  {
+    case GoalHeading::DEFAULT:
+      return "DEFAULT";
+    case GoalHeading::BIDIRECTIONAL:
+      return "BIDIRECTIONAL";
+    case GoalHeading::ANY_HEADING:
+      return "ANY_HEADING";
+    default:
+      return "Unknown";
+  }
+}
+
+// TODO(@stevedanomodolor) find a better way to do this
+
+inline GoalHeading fromStringToGH(const std::string & n)
+{
+  if(n == "DEFAULT") {
+    return GoalHeading::DEFAULT;
+  } else if(n == "BIDIRECTIONAL") {
+    return GoalHeading::BIDIRECTIONAL;
+  } else if (n == "ANY_HEADING") {
+    return GoalHeading::ANY_HEADING;
+  } else {
+    return GoalHeading::UNKNOWN;
+  }
+}
+
+
 
 const float UNKNOWN = 255.0;
 const float OCCUPIED = 254.0;
--- a/nav2_smac_planner/src/a_star.cpp
+++ b/nav2_smac_planner/src/a_star.cpp
@@ -179,30 +179,85 @@ template<>
 void AStarAlgorithm<Node2D>::setGoal(
   const unsigned int & mx,
   const unsigned int & my,
-  const unsigned int & dim_3)
+  const unsigned int & dim_3,
+  const GoalHeading & goal_heading)
 {
   if (dim_3 != 0) {
     throw std::runtime_error("Node type Node2D cannot be given non-zero goal dim 3.");
   }
 
-  _goal = addToGraph(Node2D::getIndex(mx, my, getSizeX()));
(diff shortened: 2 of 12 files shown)


SteveMacenski:
Want to populate the PR template and add a name to the PR? This is opened without any context or statement abotu the state of this work and if there are any questions (or this is a 'done' implementation) :wink:

nav2_smac_planner/src/node_hybrid.cpp
```diff
@@ -337,6 +337,11 @@ float HybridMotionTable::getAngleFromBin(const unsigned int & bin_idx)
   return bin_idx * bin_size;
 }
 
+unsigned int getNumOfBins()
+{
+  return num_angle_quantization;
```
SteveMacenski, line 342: You could just use `num_angle_quantization` for the one instance this method is used? That seems in style with the rest of it. The table is treated like a global struct to grab info (for better or worse)

nav2_smac_planner/include/nav2_smac_planner/constants.hpp
```diff
@@ -59,6 +67,38 @@ inline MotionModel fromString(const std::string & n)
...
+    case GoalHeading::BIDIRECTIONAL:
+      return "BIDIRECTIONAL";
+    case GoalHeading::ANY_HEADING:
+      return "ANY_HEADING";
+    default:
+      return "Unknown";
+  }
+}
+
+// TODO(@stevedanomodolor) find a better way to do this
```
SteveMacenski, line 85: What's the concern? I think this works fine, anything more you want to expand on here?

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -154,6 +154,20 @@ void SmacPlannerHybrid::configure(
   nav2_util::declare_parameter_if_not_declared(
     node, name + ".motion_model_for_search", rclcpp::ParameterValue(std::string("DUBIN")));
   node->get_parameter(name + ".motion_model_for_search", _motion_model_for_search);
+
+  nav2_util::declare_parameter_if_not_declared(
+  node, name + ".goal_heading", rclcpp::ParameterValue("DEFAULT"));
+  node->get_parameter(name + ".goal_heading", _goal_heading_type);
```
SteveMacenski, line 160: You don't need to store the string type if you're storing the enum class

nav2_smac_planner/src/a_star.cpp
```diff
@@ -179,39 +179,98 @@ template<>
...
-  const unsigned int & dim_3)
+  const unsigned int & dim_3,
+  const GoalHeading & goal_heading)
 {
-  _goal = addToGraph(NodeT::getIndex(mx, my, dim_3));
+  _goals.clear();
+  // _goals_coordinates.clear();
+  std::vector<Coordinates> goal_coordinates;
+  unsigned int number_of_bins = NodeT::motion_table.getNumOfBins();
+  if(goal_heading == GoalHeading::DEFAULT){
```
SteveMacenski, line 206: You could do this with a `switch` statement which would be more compact

nav2_smac_planner/src/a_star.cpp
```diff
@@ -179,39 +179,98 @@ template<>
...
 void AStarAlgorithm<NodeT>::setGoal(
   const unsigned int & mx,
   const unsigned int & my,
-  const unsigned int & dim_3)
+  const unsigned int & dim_3,
+  const GoalHeading & goal_heading)
 {
-  _goal = addToGraph(NodeT::getIndex(mx, my, dim_3));
+  _goals.clear();
+  // _goals_coordinates.clear();
```
SteveMacenski, line 203: Remove this line?

nav2_smac_planner/src/a_star.cpp
```diff
@@ -179,39 +179,98 @@ template<>
...
+  else
+  {
+    throw std::runtime_error("Goal Heading not supported");
+  }
 
-  typename NodeT::Coordinates goal_coords(
-    static_cast<float>(mx),
-    static_cast<float>(my),
-    static_cast<float>(dim_3));
+  auto has_goals_changed = [](const std::vector<Coordinates> & current_goal_coordinates,const std::vector<Coordinates> & previous_goal_coordinates) -> bool
```
SteveMacenski, line 242: I don't understand this lambda. Why not simply do `A != B`? Vectors can be compared. If the `Coordinates` class doesn't implement the `operator=`, that seems like an easier solution :-) I'm also not 100% sure why you'er comparing goal_coordinates[i] != goal_coordinates[i+1]` and never seem to use `current_` or `previous_` ones. Overall, I think this can be removed with a simple `if A != B` statement inline to the code without the lambda

SteveMacenski (changes requested):
Generally this works! Now, you need to modify `isGoal()` [1] to use those goals to check if any are the item. That should be enough to start actually testing this working! 

From there, the next major step is to expand the definitions for AnalyticExpansion to take in the goals and try to expand to each, each iteration. If any are valid, take the shortest. 


[1] https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/a_star.cpp#L365-L370

stevedanomodolor:
Thanks for the review, but it is not ready yet. I have made some changes locally since I made this push. I will close it for now, untill it is fully ready.

SteveMacenski:
OK!

PR #4021, opened 2023-12-20
New Graceful Motion Controller

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #1975 |
| Robotic platform tested on | Ignition Gazebo|

---

* Added a new version of the graceful controller based on the smooth control law with similar settings to the existing controllers.
* I think I need to add more tests to increase the test coverage.

* New entry in configuration page: https://navigation.ros.org/configuration/index.html
* New entry in plugins: https://navigation.ros.org/plugins/index.html#controllers

---

#### For Maintainers: 


nav2_graceful_motion_controller/test/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+  ${library_name}
+)
+
+# Egopolar test
+ament_add_gtest(test_egopolar
+  test_egopolar.cpp
+)
+ament_target_dependencies(test_egopolar
+  geometry_msgs tf2_geometry_msgs angles
+)
```
SteveMacenski, line 18: Line at EOF

nav2_graceful_motion_controller/graceful_controller_plugin.xml
```diff
@@ -0,0 +1,7 @@
+<class_libraries>
+    <library path="nav2_graceful_motion_controller">
+        <class type="nav2_graceful_motion_controller::GracefulMotionController" base_class_type="nav2_core::Controller">
+            <description>Graceful motion controller for Nav2</description>
+        </class>
+    </library>
+</class_libraries>
```
SteveMacenski, line 7: Line at EOF

nav2_graceful_motion_controller/package.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>nav2_graceful_motion_controller</name>
+  <version>0.0.0</version>
```
SteveMacenski, line 5: Version should match nav2's version in other package.xml files

nav2_graceful_motion_controller/package.xml
```diff
@@ -0,0 +1,34 @@
...
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+    <nav2_core plugin="${prefix}/graceful_controller_plugin.xml" />
+  </export>
+
+</package>
```
SteveMacenski, line 34: Line at EOF

SteveMacenski (changes requested):

SteveMacenski:
Yes, more test coverage is definitely needed: https://app.codecov.io/gh/ros-planning/navigation2/pull/4021/tree/nav2_graceful_motion_controller/src. We accept new packages at 90%+ test coverage, though you should aim for as much as is reasonably possible (e.g. missing a stray line due to an edge case within an impossible edge case is :shrug:) 

You also have a few linting issues that CI will probably bring up, but you can find with `ament_cpplint` and `ament_uncrustify`

I'm pleasantly surprised by the quality of this PR so this shouldn't be a huge effort for me to review even though its a big update. I did a once over on obvious things in the meantime since I don't have the cycles to give it a deep review right now, but I did want to give you feedback on tests / things you can do immediately in the meantime.

It would also be good for @mikeferguson to review this one as well given his experience. He might be able to find errors faster than I and have feedback on good bells and whistles to add that are missing.

ajtudela:
Thanks for the compliment, Steve. Sorry, to make this PR later than I though but I've been very busy this past months and I preferred to upload the code as complete as possible.

In the meantime, I'll fix the linting issues and I'll working in the new tests to increase the coverage

SteveMacenski:
A nice demo video highlighting this controller's value would be great too both for here & for the documentation

SteveMacenski:
Hi! I'm really excited for this work and its very timely - a docking effort will begin shortly which can definitely leverage this core work. any updates on your side? I know we're just now back from the holidays so no worries!

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/ego_polar_coords.hpp
```diff
@@ -0,0 +1,69 @@
...
+namespace nav2_graceful_motion_controller
+{
+
+/**
+ * @brief Egocentric polar coordinates defined as the difference between the
+ * robot pose and the target pose relative to the robot position and orientation.
+ */
+struct EgocentricPolarCoordinates
+{
+  double r;       // Distance between the robot pose and the target pose.
```
SteveMacenski, line 35: You can probably set all these as `float` since you use `float` math in the other constructor. As well change to float in the first constructor while you're at it.

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/ego_polar_coords.hpp
```diff
@@ -0,0 +1,69 @@
...
+    const geometry_msgs::msg::Pose & target,
+    const geometry_msgs::msg::Pose & current = geometry_msgs::msg::Pose())
+  {
+    float dX = target.position.x - current.position.x;
+    float dY = target.position.y - current.position.y;
+    float line_of_sight = std::atan2(-dY, dX);
+
+    r = sqrt(dX * dX + dY * dY);
+    phi = angles::normalize_angle(tf2::getYaw(target.orientation) + line_of_sight);
+    delta = tf2::getYaw(current.orientation) + line_of_sight;
```
SteveMacenski, line 63: This isn't necessarily normalized, you need to use `normalize_angle` here as well

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/smooth_control_law.hpp
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * @brief Calculate the path curvature using a Lyapunov-based feedback control law from
+   * "A smooth control law for graceful motion" (Jong Jin Park and Benjamin Kuipers).
+   *
+   * @param r Distance between the robot and the target.
+   * @param phi Orientation of target with respect to the line of sight from the robot to the target.
+   * @param delta Steering angle of the robot.
+   * @return The curvature
+   */
+  double calculateCurvature(double r, double phi, double delta);
```
SteveMacenski, line 147: All functions before all variables

nav2_graceful_motion_controller/src/smooth_control_law.cpp
```diff
@@ -0,0 +1,99 @@
...
+  const geometry_msgs::msg::Pose & target,
+  const geometry_msgs::msg::Pose & current)
+{
+  geometry_msgs::msg::Twist vel = calculateRegularVelocity(target, current);
+  geometry_msgs::msg::Pose next;
+  double yaw = tf2::getYaw(current.orientation);
+  next.position.x = current.position.x + vel.linear.x * dt * cos(yaw);
+  next.position.y = current.position.y + vel.linear.x * dt * sin(yaw);
+  yaw += vel.angular.z * dt;
+  next.orientation.z = sin(yaw / 2.0);
```
SteveMacenski, line 83: I'd generally prefer using TF2 utils to convert from RPY to quaternion than do the conversion here.

nav2_graceful_motion_controller/src/path_handler.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+PathHandler::PathHandler(
+  tf2::Duration transform_tolerance,
+  std::shared_ptr<tf2_ros::Buffer> tf,
+  std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros)
+: transform_tolerance_(transform_tolerance), tf_buffer_(tf), costmap_ros_(costmap_ros)
+{
+}
+
+nav_msgs::msg::Path PathHandler::transformGlobalPlan(
```
SteveMacenski, line 41: I assume most of this file is lifted from another controller - which one and did you make any modifications? Makes reviewing this easier if I know what is different or if its all the same as another

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,356 @@
...
+}
+
+void GracefulMotionController::setSpeedLimit(
+  const double & speed_limit, const bool & percentage)
+{
+  std::lock_guard<std::mutex> param_lock(param_handler_->getMutex());
+
+  if (percentage) {
+    // Speed limit is expressed in % from maximum speed of robot
+    params_->v_linear_max = std::max(
```
SteveMacenski, line 206: What if there are multiple set speed limit updates? this overrides the existing one. If we don't store the original, how can we do a "go 50%" then later "go 75%" of original maximum? This would apply 75% to the 50%'th value

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+  // If the pose is not far enough, take the last pose
+  if (goal_pose_it == transformed_plan.poses.end()) {
+    goal_pose_it = std::prev(transformed_plan.poses.end());
+  }
+
+  return *goal_pose_it;
+}
+
+geometry_msgs::msg::PointStamped GracefulMotionController::createMotionTargetMsg(
```
SteveMacenski, line 236: This can be a util in a new utils.hpp file in the package

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,356 @@
...
+{
+  geometry_msgs::msg::PointStamped motion_target_point;
+  motion_target_point.header = motion_target.header;
+  motion_target_point.point = motion_target.pose.position;
+  motion_target_point.point.z = 0.01;
+  return motion_target_point;
+}
+
+visualization_msgs::msg::Marker GracefulMotionController::createSlowdownMsg(
+  const geometry_msgs::msg::PoseStamped & motion_target)
```
SteveMacenski, line 247: This can be a util in a new utils.hpp file in the package

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,356 @@
...
+{
+  unsigned int mx, my;
+  if (!costmap_ros_->getCostmap()->worldToMap(x, y, mx, my)) {
+    RCLCPP_WARN(
+      logger_, "The path is not in the costmap. Cannot check for collisions. "
+      "Proceed at your own risk, slow the robot, or increase your costmap size.");
+    return false;
+  }
+
+  double footprint_cost = collision_checker_->footprintCostAtPose(
```
SteveMacenski, line 340: Its worth having a check if this footprint is a circle, in which a point check is much more efficient than full footprint SE2 check

SteveMacenski (changes requested):

ajtudela:
Hi Steve, sorry for not making the video, I'm still on Christmas holidays until Monday. I did new tests on the last day before the holidays, but I left a few until we reach 90%+ coverage. So, it's almost done.

I'll review your changes; they don't seem very difficult to do.

I'm thrilled to work on the docking implementation. #1975

nav2_graceful_motion_controller/src/path_handler.cpp
ajtudela: Yes, it's from the regulated pure pursuit controller without any modifications, if there are any.

ajtudela:
I think I fixed all the requested changes.

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/ego_polar_coords.hpp
```diff
@@ -0,0 +1,69 @@
...
+    const geometry_msgs::msg::Pose & target,
+    const geometry_msgs::msg::Pose & current = geometry_msgs::msg::Pose())
+  {
+    float dX = target.position.x - current.position.x;
+    float dY = target.position.y - current.position.y;
+    float line_of_sight = std::atan2(-dY, dX);
+
+    r = sqrt(dX * dX + dY * dY);
+    phi = angles::normalize_angle(tf2::getYaw(target.orientation) + line_of_sight);
+    delta = angles::normalize_angle(tf2::getYaw(current.orientation) + line_of_sight);
```
mikeferguson, line 63: I think there is a subtle bug here if current.orientation isn't always 0 (I think we get that right now because you transform the plan into the local frame before proceeding, right?

The paper assumes that these (especially delta) is in ego-centric coordinates (so delta should just be line_of_sight - and phi should be line_of_sight + target-current). This will especially change the calculateCurvature - as it will vary depending on the orientation of the robot - which will make it quite hard to tune)

I would suggest we at least add a comment here noting this assumption

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  // Transform path to robot base frame and publish it
+  auto transformed_plan = path_handler_->transformGlobalPlan(
+    pose, params_->max_robot_pose_search_dist);
+  transformed_plan_pub_->publish(transformed_plan);
+
+  // Get the particular point on the path at the motion target distance and publish it
+  auto motion_target = getMotionTarget(params_->motion_target_dist, transformed_plan);
+  auto motion_target_point = nav2_graceful_motion_controller::createMotionTargetMsg(motion_target);
```
mikeferguson, line 127: I think you will find this performs splendidly poorly in the real world.

 * The graceful law doesn't necessarily follow the path exactly
 * If you only pick the farthest out motion point once, then if the graceful law takes you off the path, and you later fail in the collision checking, you get no velocity command and end up needing to globally replan

What my graceful controller does is an iterative approach:

 * Select a candidate motion target (it must satisfy the rules, such as not too far away from robot)
 * Run the control law
 * Forward simulate with collisions
 * If there is a collision, repeat by selecting a candidate motion target that is a bit closer on the path (until you hit a minimum distance threshold beyond which we have to return to the global planner)

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+  return true;
+}
+
+geometry_msgs::msg::Twist GracefulMotionController::rotateToTarget(const double & angle_to_target)
+{
+  double sign = angle_to_target < 0.0 ? -1.0 : 1.0;
+  geometry_msgs::msg::Twist vel;
+  vel.linear.x = 0.0;
+  vel.angular.z = sign * params_->v_angular_max;
```
mikeferguson, line 295: I would recommend you actually use the angle_to_target rather than a fixed -1/+1 - and then have a scaling parameter - otherwise this almost sure to overshoot (you'll be going max_angular_v as you cross the goal, and you still need to decelerate)

nav2_graceful_motion_controller/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+  nav2_core
+  tf2
+  tf2_geometry_msgs
+  nav_2d_utils
+  angles
+)
+
+set(library_name nav2_graceful_motion_controller)
+
+add_library(${library_name} SHARED
```
mikeferguson, line 41: for ease of use with docking, might want to have two libraries - one that is just the underlying control law, and a second that is the actual controller plugin itself

mikeferguson (commented):
The underlying control law (which is probably all we would use in a docking effort) looks sound (although a few comments should probably be added as noted).

I don't expect that the actual controller itself is going to perform very well outside of very open spaces - but we could come back and attack that later in later PRs.

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/parameter_handler.hpp
```diff
@@ -0,0 +1,94 @@
...
+namespace nav2_graceful_motion_controller
+{
+
+struct Parameters
+{
+  double transform_tolerance;
+  double motion_target_dist;
+  double max_robot_pose_search_dist;
+  double k_phi;
+  double k_delta;
```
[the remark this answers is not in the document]
ajtudela: I read an updated paper in which they changed the name from k1/k2 to kphi/kdelta but yes, I'll add a note with that change.

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/ego_polar_coords.hpp
ajtudela: Yes, transforming the plan to the local frame is the first step before the other calculations. So, in a "regular" use current pose is always zero and r, phi, delta are computed according to the paper.

The only part of the code that current pose =/= zero is in simulateTrajectory(...) when the the pose is projected into the future to generate the path.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
ajtudela: I agree. I tested in our robot and, unless you set a motion point very close to the robot, it will not follow the path exactly.

SteveMacenski:
Sorry for the delay, I was on a work trip. Would you be OK if @mikeferguson took the lead in finishing up this PR? I'm working with him on a related project that we need to move fast on which has related application of the core control law. He's offered to clean this up with some of the concerns he's listed above as part of that work and he's thought about this area quite a bit over the last couple of years on his previous version of this controller.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+      "Proceed at your own risk, slow the robot, or increase your costmap size.");
+    return false;
+  }
+
+  // Calculate the cost of the footprint at the robot's current position depending
+  // on the shape of the footprint
+  double footprint_cost;
+  if (costmap_ros_->getUseRadius()) {
+    footprint_cost = collision_checker_->pointCost(mx, my);
+    if (footprint_cost >= nav2_costmap_2d::INSCRIBED_INFLATED_OBSTACLE) {
```
SteveMacenski, line 323: I think this misses the condition of unknown space like handled in the full footprint case.

I think in general, this logic can be lifted from somewhere else in the stack to make sure it handles all the conditions. I'm only glancing at this quickly, but I don't think this is the only problem

nav2_graceful_motion_controller/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+find_package(geometry_msgs REQUIRED)
+find_package(nav_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(nav_2d_utils REQUIRED)
+find_package(angles REQUIRED)
+
+nav2_package()
+set(CMAKE_CXX_STANDARD 17)
```
SteveMacenski, line 19: Does this need to specifically be C++17? `nav2-package()` already sets it to the REP2000 value

nav2_graceful_motion_controller/CMakeLists.txt
SteveMacenski: True! Or the main control law being header-only

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/graceful_motion_controller.hpp
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * @brief Checks if the robot is in collision
+   * @param x The x coordinate of the robot in global frame
+   * @param y The y coordinate of the robot in global frame
+   * @param theta The orientation of the robot in global frame
+   * @return Whether in collision
+   */
+  bool inCollision(const double & x, const double & y, const double & theta);
+
+  rclcpp_lifecycle::LifecycleNode::WeakPtr parent_node_;
```
SteveMacenski, line 153: This doesn't appear to be ever used again. This can be removed

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/graceful_motion_controller.hpp
```diff
@@ -0,0 +1,179 @@
...
+  std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
+  std::string plugin_name_;
+  std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
+  std::unique_ptr<nav2_costmap_2d::FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>
+  collision_checker_;
+  geometry_msgs::msg::TransformStamped costmap_transform_;
```
SteveMacenski, line 159: This could be passed to the simulate function without the need for storing as a class member murking up the logic flow, since its only used twice

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/path_handler.hpp
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2023 Alberto J. Tudela Roldán
```
SteveMacenski, line 1: If this is taken from other Nav2 projects, it should use that project's header

nav2_graceful_motion_controller/src/parameter_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  dyn_params_handler_ = node->add_on_set_parameters_callback(
+    std::bind(&ParameterHandler::dynamicParametersCallback, this, std::placeholders::_1));
+}
+
+rcl_interfaces::msg::SetParametersResult
+ParameterHandler::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
+{
+  rcl_interfaces::msg::SetParametersResult result;
+  std::lock_guard<std::mutex> lock_reinit(mutex_);
```
SteveMacenski, line 109: This lock is only in a single place. You'd need to lock the parameter `get`'s too

nav2_graceful_motion_controller/src/path_handler.cpp
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2023 Alberto J. Tudela Roldán
```
SteveMacenski, line 1: Ditto on the hpp file

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+  tf_buffer_ = tf;
+  plugin_name_ = name;
+  logger_ = node->get_logger();
+
+  // Handles storage and dynamic configuration of parameters.
+  // Returns pointer to data current param settings.
+  param_handler_ = std::make_unique<ParameterHandler>(
+    node, plugin_name_, logger_,
+    costmap_ros_->getCostmap()->getSizeInMetersX());
+  params_ = param_handler_->getParams();
```
SteveMacenski, line 44: I think this causes a problem if the dynamic parameters are changing `params_` without a lock on this parameter pointer's getting of that memory. Maybe it makes sense for `getParams()` to return a parameters copy and the `getParams()` method is locked by the mutex. That way, the get/sets are mutually exclusive and the params can't change mid-execution here. We get a new `getParams()` at the start of each cycle. 

Though, that complicates anything initialized here in the configure method. I think there's a bit of a design problem that might need more than a simple solution to resolve

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+  RCLCPP_INFO(logger_, "Configured Graceful Motion Controller: %s", plugin_name_.c_str());
+}
+
+void GracefulMotionController::cleanup()
+{
+  RCLCPP_INFO(
+    logger_,
+    "Cleaning up controller: %s of type graceful_motion_controller::GracefulMotionController",
+    plugin_name_.c_str());
+  transformed_plan_pub_.reset();
```
SteveMacenski, line 74: Also collision checker, control law, path handler, and param handler's pointer resets

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+  motion_target_pub_->on_deactivate();
+  slowdown_pub_->on_deactivate();
+}
+
+geometry_msgs::msg::TwistStamped GracefulMotionController::computeVelocityCommands(
+  const geometry_msgs::msg::PoseStamped & pose,
+  const geometry_msgs::msg::Twist & /*velocity*/,
+  nav2_core::GoalChecker * goal_checker)
+{
+  std::lock_guard<std::mutex> param_lock(param_handler_->getMutex());
```
SteveMacenski, line 109: Ah I see, so you lock the param mutex here. Ok. I think that makes sense, but how does the change of the K values and other values of the control law instance get updated?

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+    pose, params_->max_robot_pose_search_dist);
+  transformed_plan_pub_->publish(transformed_plan);
+
+  // Get the particular point on the path at the motion target distance and publish it
+  auto motion_target = getMotionTarget(params_->motion_target_dist, transformed_plan);
+  auto motion_target_point = nav2_graceful_motion_controller::createMotionTargetMsg(motion_target);
+  motion_target_pub_->publish(motion_target_point);
+
+  // Publish marker for slowdown radius around motion target for debugging / visualization
+  auto slowdown_marker = nav2_graceful_motion_controller::createSlowdownMarker(
```
SteveMacenski, line 131: What's this slowdown logic?

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+    motion_target,
+    params_->slowdown_radius);
+  slowdown_pub_->publish(slowdown_marker);
+
+  // Compute distance to goal
+  geometry_msgs::msg::PoseStamped goal_pose = transformed_plan.poses.back();
+  double dist_to_goal = std::hypot(goal_pose.pose.position.x, goal_pose.pose.position.y);
```
SteveMacenski, line 138: Might want to use the path's integrated distance, there's a util for that in nav2_utils. That way its not just the L2 distance, but the integrated path distances to account for path curvatures

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+  } else if (params_->initial_rotation && // NOLINT
+    fabs(angle_to_target) > params_->initial_rotation_min_angle)
+  {
+    cmd_vel.twist = rotateToTarget(angle_to_target);
+  } else {
+    cmd_vel.twist = control_law_->calculateRegularVelocity(motion_target.pose, reversing);
+  }
+
+  // Transform local frame to global frame to use in collision checking
+  try {
```
SteveMacenski, line 180: I don't understand this. What's up with this costmap transform? If we have some point `N` in the path, the path handler should have transformed it (from `map`) into the costmap's global frame (i.e. `odom`), no? Or did you not use that part of the other code's path handler?

If we have them in the costmap's global frame (ie `odom`), then we can use `mapToWorld` and `worldToMap` easy checks to convert the path's coordinates into the costmap frame to do lookups. Or is this control law not being computed in `odom` frame? 

I think something here is being over thought. We should be doing all the work in the `odom`, local costmap frame

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,344 @@
...
+    // Apply velocities to calculate next pose
+    next_pose.pose = control_law_->calculateNextPose(
+      dt, motion_target.pose, next_pose.pose, backward);
+
+    // Add the pose to the trajectory for visualization
+    trajectory.poses.push_back(next_pose);
+
+    // Check for collision
+    geometry_msgs::msg::PoseStamped global_pose;
+    tf2::doTransform(next_pose, global_pose, costmap_transform_);
```
SteveMacenski, line 285: Here's where I mean. What's next_pose frame? Should be odom, the local costmap frame already, right? Or is this because we're in robot-centric coordinates?

SteveMacenski (changes requested):

ajtudela:
> Sorry for the delay, I was on a work trip. Would you be OK if @mikeferguson took the lead in finishing up this PR? I'm working with him on a related project that we need to move fast on which has related application of the core control law. He's offered to clean this up with some of the concerns he's listed above as part of that work and he's thought about this area quite a bit over the last couple of years on his previous version of this controller.
> 
> I wouldn't want to take this from you however if you want to complete it yourself and I totally respect the desire to do that. There's probably just a few small things Fergs can push to your branch ASAP like decoupling the controller from the control law libs so that we can use it in the docking project while you work on the controller-specific action items that Fergs mentions in his review above. Its completely up to you, no pressure. If you're under a time constraint and can't make progress, it might be a good option. If not and you want to 'own' this project, the more the merrier :-)
> 
> I'm very impressed by your updates for test coverage, really good job there at 96%! Even though there are a number of comments below, honestly they're mostly pedantic to improve things, this is very good!

No problem at all. In the meantime, I was fixing the changes you and Mike requested. I was working in the last one: the iterative selection of the motion target. The decoupling is already implemented.

I would like to finish it but @mikeferguson has a deeper knowledge of the controller as he worked on it much more time than me; and both of you seem to need to close this PR quickly so lets Mike wrap this up.

SteveMacenski:
I don't want to take it from you - so why don't you continue with the controller plugin items and Fergs will probably push a few updates to your branch to cover the core control-law changes (like decoupling the lib) needed for his work. He might come in and do a few other things, but I want to leave this squarely in your camp if you want to complete it.

@mikeferguson can you give an update of what changes are critical for your path or you feel you have expert knowledge that its better for you to handle? Then, @ajtudela can do the rest and we don't have conflict areas

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/path_handler.hpp
ajtudela: It's based on RPP path class but it's not 100% exactly. Should I the RPP header in that case?

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
ajtudela: The control law parameters only can be changed when you read the parameters at the start because they are passed to the control_law class when the node is configured. Lines 50-52 of nav2_graceful_motion_controller.cpp

I can add those values to the OnSetParametersCallback and then set the control law with the new values.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
ajtudela: > Edit: oh, is it because the control law is in robot base coordinate frame? That would make sense

Yeap, you're right. This get the transform of the robot in the map to pass it to the trajectory simulation. That is because the control law assumes that all coordinates are in the robot base frame, egocentric, as they called in the paper.

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/path_handler.hpp
SteveMacenski: Yeah. If you copy+paste files, you must retain the old headers. If you substantially change it from there, then you can add your own copyright line in addition to the original, but only for substantive changes. What that means is somewhat subjective, but I'm also not really overly picky about it. Use your best judgement.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
SteveMacenski: That would be great!

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
SteveMacenski: Got it!

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
ajtudela: Mmm, I think I don't understand this. Isn't unknown space the same as NO_INFORMATION = 255?

With that logic, any value above 253 will be treated as a collision.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -317,14 +315,19 @@ bool GracefulMotionController::inCollision(const double & x, const double & y, c
 
   // Calculate the cost of the footprint at the robot's current position depending
   // on the shape of the footprint
+  std::unique_ptr<nav2_costmap_2d::FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>
+  collision_checker =
```
SteveMacenski, line 319: Why was this moved? I think this should be moved back. This isn't a "Free" thing to construct every time.

ajtudela: This could be passed to the simulate function without the need for storing as a class member murking up the logic flow, since its only used twice

_Originally posted by @SteveMacenski in https://github.com/ros-planning/navigation2/pull/4021#discussion_r1463968006_

My bad. I think you mentioned here the collision checker instead the costmap_transform_

SteveMacenski:
Let us know when you're ready for another review!

ajtudela:
I'm ready. I made all the requested changes.

SteveMacenski:
@mikeferguson please re-review

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,354 @@
...
+    params_->v_linear_max = std::max(
+      params_->v_linear_max_initial * speed_limit / 100.0,
+      params_->v_linear_min);
+  } else {
+    // Speed limit is expressed in m/s
+    params_->v_linear_max = std::max(speed_limit, params_->v_linear_min);
+  }
+
+  // Update the speed limit in the control law
+  control_law_->setSpeedLimit(params_->v_linear_min, params_->v_linear_max, params_->v_angular_max);
```
SteveMacenski, line 234: What about dynamic parameter changes these? Those are never updated.

SteveMacenski, line 234: Angular velocity is never adjusted.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,354 @@
...
+    RCLCPP_ERROR(
+      logger_, "Could not transform %s to %s: %s",
+      costmap_ros_->getBaseFrameID().c_str(), costmap_ros_->getGlobalFrameID().c_str(),
+      ex.what());
+    return cmd_vel;
+  }
+
+  // Generate and publish local plan for debugging / visualization
+  nav_msgs::msg::Path local_plan;
+  if (!simulateTrajectory(pose, motion_target, costmap_transform, local_plan, reversing)) {
```
SteveMacenski, line 203: Fergs remarked on this - you need to back out/forward the poitn

SteveMacenski (changes requested):

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
SteveMacenski: Still a problem, I think

SteveMacenski:
But overall the code looks really good! I'm leaning on Fergs for a details algorithm review and subtle details from his expertise. He still has several of his original comments unaddressed

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
ajtudela: Those are changed in lines 130-136 of parameter_handler.cpp, in the dynamicParametersCallback

ajtudela: Fixed

ajtudela:
I'm ready. I made all the requested changes.

> But overall the code looks really good! I'm leaning on Fergs for a details algorithm review and subtle details from his expertise. He still has several of his original comments unaddressed

I agreed. I think I made most of the changes proposed by Fergs but I let the rest of that part to him.

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
```diff
@@ -0,0 +1,358 @@
...
+    params_->v_linear_max = std::max(
+      params_->v_linear_max_initial * speed_limit / 100.0,
+      params_->v_linear_min);
+  } else {
+    // Speed limit is expressed in m/s
+    params_->v_linear_max = std::max(speed_limit, params_->v_linear_min);
+  }
+
+  // Limit the angular velocity to be proportional to the linear velocity
+  params_->v_angular_max = params_->v_linear_max * params_->rotation_scaling_factor;
```
SteveMacenski, line 234: Huh? I think you need to rethink this.

SteveMacenski (changes requested):

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
SteveMacenski: @ajtudela they are never updated for use. If that is updated, they're never actually communicated to the control algorithm to **use**

nav2_graceful_motion_controller/src/graceful_motion_controller.cpp
SteveMacenski: Why would either of these values appear in this expression? It should be something like the proportion of max linear velocity reduction times the original max angular velocity to scale it down proportionately to the requested linear velocity reduction. Try checking out other implementations

SteveMacenski:
OK on the Fergs items. We discussed in the Working Group meeting today that we can merge in once you finish up these last couple details with the dynamic parameters and speed limits. The remaining items @mikeferguson will enumerate here, then we can merge this (thanks so much!) and he can pick those up in a follow on PR!

SteveMacenski:
Looks good to me! Fergs?

The last bits are just documentation! Adding the new plugin to the Configuration Guide with the parameters and a few sentences up top about what it is & announcing it on the [Migration Guide](https://navigation.ros.org/migration/index.html) and in the [Navigation Plugins](https://navigation.ros.org/plugins/index.html) table.

nav2_graceful_motion_controller/include/nav2_graceful_motion_controller/parameter_handler.hpp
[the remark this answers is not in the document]
mikeferguson: If we are going to keep this name, we should also add the name of the updated paper - the only one in the README is the original 2011 paper that uses k1/k2

nav2_graceful_motion_controller/src/smooth_control_law.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // Adjust the linear velocity as a function of the path curvature to
+  // slowdown the controller as it approaches its target
+  double v = v_linear_max_ / (1.0 + beta_ * std::pow(fabs(curvature), lambda_));
+
+  // Set some small v_min when far away from origin to promote faster
+  // turning motion when the curvature is very high
+  v = std::clamp(v, v_linear_min_, v_linear_max_);
+
+  // Slowdown when the robot is near the target to remove singularity
+  v = std::min(v_linear_max_ * (ego_coords.r / slowdown_radius_), v);
```
mikeferguson, line 72: This needs to be BEFORE the std::clamp - otherwise as you approach the goal you will violate v_linear_min_ - which usually means the robot comes to a full and complete stop (v_linear_min_ should be the absolute lowest velocity that the robot can actually control, less than that and the robot will stall in place)

mikeferguson (changes requested):
We can almost merge this - remaining issues:
 * The v_linear_min issue I just raised (I found this during some other testing)
 * Add the name of the updated paper to the README
 * Address the collision checking concerns that Steve raised
 
Things that I will do in a follow on PR:
 * Revamp the controller body to "search" for a motion target (I can re-license and merge the stuff from my controller because I personally hold the copyright on that)
 * Likely adjust some parameter defaults based on experience with running this kind of controller on a number of robots

ajtudela:
Thanks @mikeferguson!! I fixed the minor issues with the velocity and Readme and I think the collision cheking is also fixed.

I made a PR for the docs here: [517](https://github.com/ros-planning/navigation.ros.org/pull/517)

SteveMacenski:
One small typo on the documentation, but otherwise LGTM! 

@mikeferguson are these comments still pending and/or items you plan to address? Just want to make sure we don't leave anything hanging. 

> I believe this will be wrong when the robot is "reversing" (which is a case not handled by the paper, but which occurs in the real world)

> I think there is a subtle bug here if current.orientation isn't always 0 (I think we get that right now because you transform the plan into the local frame before proceeding, right?

> I would recommend you actually use the angle_to_target rather than a fixed -1/+1 - and then have a scaling parameter - otherwise this almost sure to overshoot (you'll be going max_angular_v as you cross the goal, and you still need to decelerate)

ajtudela:
> @ajtudela Overall, really great work on all of this. Definitely great code quality, functionality, and understanding of the important trade offs, I'm very impressed! I'd welcome your help anytime on any subjects your find interest in! 😄

Thank you very much! It was hard at first, especially because of the tests, I'm not used to doing them. But after so many commits I've gotten the hang of it and I'm starting to love them. :D

Of course. Next project: docking in nav2!

mikeferguson:
> I believe this will be wrong when the robot is "reversing" (which is a case not handled by the paper, but which occurs in the real world)

As close as I can tell without manually testing, yes this is now resolved. I can fix any remaining issues when doing the carrot overhaul.

> I think there is a subtle bug here if current.orientation isn't always 0 (I think we get that right now because you transform the plan into the local frame before proceeding, right?

It's not an issue for the codebase as-is - and it does make the EpiPolar code more generic as currently implemented. I only see it as a potential trap for someone who comes in and modifies this code without realizing the assumption. I figured I would leave the comment unresolved there in case anyone someday comes back through this because they didn't realize those assumptions.

> I would recommend you actually use the angle_to_target rather than a fixed -1/+1 - and then have a scaling parameter - otherwise this almost sure to overshoot (you'll be going max_angular_v as you cross the goal, and you still need to decelerate)

Yes, this is resolved

SteveMacenski:
Great! Any objections to a merge @mikeferguson ?

mikeferguson:
No objections

SteveMacenski:
@ajtudela last-last thing, you have a test failure:

> /opt/overlay_ws/src/navigation2/nav2_graceful_motion_controller/test/test_graceful_motion_controller.cpp:1119
Expected: (cmd_vel.twist.linear.x) < (0.0), actual: 0.1 vs 0/opt/overlay_ws/src/navigation2/nav2_graceful_motion_controller/test/test_graceful_motion_controller.cpp:1120
Expected: (cmd_vel.twist.angular.z) < (0.0), actual: 0.293018 vs 0

ajtudela:
ups, fixed!

Issue #3966, opened 2023-11-16
[Smac planner] Looping motion primitive list in every getNeighbor call


akchobby:
In the code snippet below , one can see the motion primitives list needs to be list iterated to find the direction change index but for a given set of motion primitives this will always be the same value. Can this be avoided ? or is there a particular reasoning behind it? 

https://github.com/ros-planning/navigation2/blob/edfe5bf6bd785f92061f880535512cddc67da5ae/nav2_smac_planner/src/node_lattice.cpp#L476-L481

this call can be done once at the start and the value can be stored. 

To reiterate what I have expressed below is a simple print that I have added to see if the direction change index ever change , in my test I don't see it change.
![direction_change_test_1](https://github.com/ros-planning/navigation2/assets/44644339/6d68c5f3-85c2-4163-8a1b-25bf22bd9552)
(test done on the sample primitives ackermann with 0.5 turning radius)

SteveMacenski:
This list is relatively small (e.g. 3-7) of the relative motion primitives of the current pose only. Its not all motion primitives of the entire lattice, but the motion primitives obtainable at the current heading. 

But you are correct, that for the case that reversing is allowed, we should know that IDX at the parse time and that could be precomputed for each angular quantization and stored in the motion table. Is that something you'd be interested in doing? That would speed up the planner a bit, though I'm not sure it would be substantially faster given how small of an operation this is. To your own logging, that loop stops after 6 cycles. But hey, I'll never complain with slightly faster ;-)

SteveMacenski:
@akchobby would you be open to contributing this?

akchobby:
> @akchobby would you be open to contributing this?

Yes I can take this up , shall do it towards the last week of December

SteveMacenski:
Merging imminent

PR #4026, opened 2023-12-26
Direction change index computation during initialization

akchobby:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #3966 ) |

---

Prior the direction change index was being computed in ever getNeighbor call, now its done at parse time.

* added an attribute (vector) to the LatticeMotionTable class to hold direction change indices
* Logic to decide if current angle index  requires a direction change  modified to point to the above attribute.

-- N.A --

---

## Future work 
* I think instead of holding the reverse indices in a vector, we could instead obtain it in the getMotionPrimitives function in the lines below:
https://github.com/akchobby/navigation2/blob/3966-reverse-index-calculation/nav2_smac_planner/src/node_lattice.cpp#L138-L146 
* this would require the modification of the getMotionPrimitives function parameters.

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/node_lattice.cpp
+++ b/nav2_smac_planner/src/node_lattice.cpp
@@ -95,12 +95,19 @@ void LatticeMotionTable::initMotionModel(
 
     if (prev_start_angle != new_primitive.start_angle) {
       motion_primitives.push_back(primitives);
+
+      // adding direction change index, used only if reversing is allowed
+      direction_change_indices.push_back(static_cast<unsigned int>(primitives.size()));
+
       primitives.clear();
       prev_start_angle = new_primitive.start_angle;
     }
     primitives.push_back(new_primitive);
   }
   motion_primitives.push_back(primitives);
+  // adding direction change index for last angle
+  direction_change_indices.push_back(static_cast<unsigned int>(primitives.size()));
+
 
   // Populate useful precomputed values to be leveraged
   trig_values.reserve(lattice_metadata.number_of_headings);
@@ -128,6 +135,7 @@ MotionPrimitivePtrs LatticeMotionTable::getMotionPrimitives(const NodeLattice * 
     if (reserve_heading > num_angle_quantization) {
       reserve_heading -= num_angle_quantization;
     }
+    // direction change index can be obtained from here
 
     MotionPrimitives & prims_at_reverse_heading = motion_primitives[reserve_heading];
     for (unsigned int i = 0; i != prims_at_reverse_heading.size(); i++) {
@@ -472,14 +480,7 @@ void NodeLattice::getNeighbors(
   Coordinates initial_node_coords, motion_projection;
   MotionPrimitivePtrs motion_primitives = motion_table.getMotionPrimitives(this);
   const float & grid_resolution = motion_table.lattice_metadata.grid_resolution;
-
-  unsigned int direction_change_idx = 1e9;
-  for (unsigned int i = 0; i != motion_primitives.size(); i++) {
-    if (motion_primitives[0]->start_angle != motion_primitives[i]->start_angle) {
-      direction_change_idx = i;
-      break;
-    }
-  }
+  unsigned int start_angle_index = static_cast<unsigned int>(motion_primitives[0]->start_angle);
 
   for (unsigned int i = 0; i != motion_primitives.size(); i++) {
     const MotionPose & end_pose = motion_primitives[i]->poses.back();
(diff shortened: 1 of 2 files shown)


nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -95,12 +95,19 @@ void LatticeMotionTable::initMotionModel(
 
     if (prev_start_angle != new_primitive.start_angle) {
       motion_primitives.push_back(primitives);
+
+      // adding direction change index, used only if reversing is allowed
+      direction_change_indices.push_back(static_cast<unsigned int>(primitives.size()));
```
SteveMacenski, line 100: I don't believe this works. The reversing primitives are not added [until here](https://github.com/ros-planning/navigation2/blob/7bb4954cdb3c8b486d10ef1e8318dd1861993465/nav2_smac_planner/src/node_lattice.cpp#L121-L147), they're not stored in the file themselves. It seems better that `getMotionPrimitives` returns the `idx` of the reversing where it is populated instead. 

How did you test this & working?

SteveMacenski (changes requested):

nav2_smac_planner/src/node_lattice.cpp
akchobby: the current implementation definitely works: 
- the method used to test was to compare the direction change index from the old implementation to the current one (the parameter file was updated to use the smac lattice planner & reversing was set to true). The gazebo simulation was used with random start and end points (manually done) to check if the direction change index is correct.
- The core logic of the implementation is that  in the function getMotionPrimitives a list is first created with **all forward primitives** for the current heading , then the reverse primitives are appended. This implies the `direction_change index` is nothing but the number of primitives in the forward direction.

The reason I added a section as future work is that in the issue  #3966 it was suggested the calculation to be done at initialization (this is how its implemented now) but I felt we could also "save memory " by not storing the values but just returning the index in getMotionPrimitives function.

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -95,12 +95,19 @@ void LatticeMotionTable::initMotionModel(
...
+
       primitives.clear();
       prev_start_angle = new_primitive.start_angle;
     }
     primitives.push_back(new_primitive);
   }
   motion_primitives.push_back(primitives);
+  // adding direction change index for last angle
+  direction_change_indices.push_back(static_cast<unsigned int>(primitives.size()));
+
```
SteveMacenski, line 110: Remove extra empty line

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -128,6 +135,7 @@ MotionPrimitivePtrs LatticeMotionTable::getMotionPrimitives(const NodeLattice *
     if (reserve_heading > num_angle_quantization) {
       reserve_heading -= num_angle_quantization;
     }
+    // direction change index can be obtained from here
```
SteveMacenski, line 138: Remove excess comment?

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -472,14 +480,7 @@ void NodeLattice::getNeighbors(
...
   const float & grid_resolution = motion_table.lattice_metadata.grid_resolution;
-
-  unsigned int direction_change_idx = 1e9;
-  for (unsigned int i = 0; i != motion_primitives.size(); i++) {
-    if (motion_primitives[0]->start_angle != motion_primitives[i]->start_angle) {
-      direction_change_idx = i;
-      break;
-    }
-  }
+  unsigned int start_angle_index = static_cast<unsigned int>(motion_primitives[0]->start_angle);
```
SteveMacenski, line 483:
```suggestion
  unsigned int direction_change_idx = motion_table.direction_change_indices[motion_primitives[0]->start_angle];
```

Then use `direction_change_idx` in L506. That prevents an additional lookup every loop. This is a minor optimization, but relative to the speed ups removing this loop will provide, its important

SteveMacenski (changes requested):
You're right - I missed some nuance. LGTM

nav2_smac_planner/src/node_lattice.cpp
SteveMacenski: Can you give me a link to that patch? Either would be technically correct. I think I'd prefer the `getMotionprimitives` version, but I'd want to see if before saying that definitely

akchobby: Just pushed it to temporary branch here: https://github.com/akchobby/navigation2/commits/tmp-direction-change/

SteveMacenski: Yeah lets go with this instead

nav2_smac_planner/src/node_lattice.cpp
```diff
@@ -128,7 +133,6 @@ MotionPrimitivePtrs LatticeMotionTable::getMotionPrimitives(const NodeLattice *
     if (reserve_heading > num_angle_quantization) {
       reserve_heading -= num_angle_quantization;
     }
-
```
SteveMacenski, line 131: Readd this line

SteveMacenski (changes requested):
Otherwise, LGTM

SteveMacenski (approved):

PR #4045, opened 2024-01-10
Bring back pep257 docstring

tonynajjar:
Works when the "Returns" is added. I can't say I know why exactly though but my hypothesis is that for the style with `Arg` (without colon) the Returns section is mandatory :man_shrugging: . For other places with `Args:` (with colon), the Returns section is missing and the test doesn't complain


SteveMacenski:
what...the...fuck.... godIhatepython

Thank you so much :-)

PR #4046, opened 2024-01-10
Handle dynamically changes in parameters.

Sunart24:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Any |
| Robotic platform tested on | (Real and gazebo-simulated robot) |

---

* As in the other controller, I added the possibility to handle the parameter changes for the MPPI controller. At this moment some parameters such as visualize are updated and their new value is changing something in the code and the robot behavior. For others such as maximum velocities, it did not happen.  

I'd like to add it to all the parameters, so we can dynamically reconfigure the MPPI controller, however, I am not sure if this controller was thought to change any parameter when a user sends a `ros2 param set` command. So, I open the PR to put it on the table present the idea, and if needed modify the code to get the best result.

* As it is related to dynamic param handling, that is a common task, there is no documentation needed
---

* Test if constraint values that depend on the parameters need to be re-calculated when a relevant parameter is updated.

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/optimizer.cpp
+++ b/nav2_mppi_controller/src/optimizer.cpp
@@ -123,6 +123,8 @@ void Optimizer::reset()
   control_history_[2] = {0.0, 0.0, 0.0};
   control_history_[3] = {0.0, 0.0, 0.0};
 
+  settings_.constraints = settings_.base_constraints;
+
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
Why draft?

Sunart24:
> Why draft?

just to wait for the tests.

PR #4047, opened 2024-01-11
Fix MPPI obstacle critic InflationLayer check

renan028:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | simulation |

---

The current condition passes when you have another layer that extends `InflationLayer`. 
We need to make sure it is the actual `InflationLayer` and not a derived class.

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/README.md
+++ b/nav2_mppi_controller/README.md
@@ -107,6 +107,7 @@ This process is then repeated a number of times and returns a converged solution
  | collision_cost       | double | Default 100000.0. Cost to apply to a true collision in a trajectory.                                          |
  | collision_margin_distance   | double    | Default 0.10. Margin distance from collision to apply severe penalty, similar to footprint inflation. Between 0.05-0.2 is reasonable. |
  | near_goal_distance          | double    | Default 0.5. Distance near goal to stop applying preferential obstacle term to allow robot to smoothly converge to goal pose in close proximity to obstacles.   
+ | inflation_layer_name        | string    | Default "". Name of the inflation layer. If empty, it uses the last inflation layer in the costmap. If you have multiple inflation layers, you may want to specify the name of the layer to use. |
 
 #### Path Align Critic
  | Parameter                  | Type   | Definition                                                                                                                         |
--- a/nav2_mppi_controller/include/nav2_mppi_controller/critics/obstacles_critic.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/critics/obstacles_critic.hpp
@@ -16,9 +16,10 @@
 #define NAV2_MPPI_CONTROLLER__CRITICS__OBSTACLES_CRITIC_HPP_
 
 #include <memory>
+#include <string>
+
 #include "nav2_costmap_2d/footprint_collision_checker.hpp"
 #include "nav2_costmap_2d/inflation_layer.hpp"
-
 #include "nav2_mppi_controller/critic_function.hpp"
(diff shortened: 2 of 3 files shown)


SteveMacenski:
I think its perfectly find to have a derived layer from the inflation layer instead of it itself.

renan028:
> I think its perfectly find to have a derived layer from the inflation layer instead of it itself. It should still have those methods available to use

It will not break the code for sure, since the methods will be available. 
But I don't think it is the expected behavior, since the `obstacle critic` will use the parameters of that new layer instead of the actual `InflationLayer`.
For instance, in our use case, we have both `InflationLayer` and a derived class that we use for other purpose, not at the end to inflate all costs. So, it is using the parameters of this new derived layer.

Hum, but I can see that for other use cases, someone may want to actually use his derived class, not the vanilla `InflationLayer`.

SteveMacenski:
> Hum, but I can see that for other use cases, someone may want to actually use his derived class, not the vanilla InflationLayer.

Indeed, I actually know a couple of users that do that.

Any workaround that you can think of which could settle this? I believe all the layers have a [name](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/include/nav2_costmap_2d/layer.hpp#L122-L125) attached to them. We could iterate looking for the inflation layer and **also** match it to a parameter `inflation_layer_name` parameter, if set to non-empty string

renan028:
> > Hum, but I can see that for other use cases, someone may want to actually use his derived class, not the vanilla InflationLayer.
> 
> Indeed, I actually know a couple of users that do that.
> 
> Any workaround that you can think of which could settle this? I believe all the layers have a [name](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/include/nav2_costmap_2d/layer.hpp#L122-L125) attached to them. We could iterate looking for the inflation layer and **also** match it to a parameter `inflation_layer_name` parameter, if set to non-empty string. That seems like one workaround, but there are likely many

Yes, that was the solution I thought about (adding a new parameter for that). I think it makes sense.
I will change this one to draft until I make the change.
Thanks.

nav2_mppi_controller/README.md
```diff
@@ -106,7 +106,8 @@ This process is then repeated a number of times and returns a converged solution
  | cost_power           | int    | Default 1. Power order to apply to term.                                                                    |
  | collision_cost       | double | Default 100000.0. Cost to apply to a true collision in a trajectory.                                          |
  | collision_margin_distance   | double    | Default 0.10. Margin distance from collision to apply severe penalty, similar to footprint inflation. Between 0.05-0.2 is reasonable. |
- | near_goal_distance          | double    | Default 0.5. Distance near goal to stop applying preferential obstacle term to allow robot to smoothly converge to goal pose in close proximity to obstacles.   
+ | near_goal_distance          | double    | Default 0.5. Distance near goal to stop applying preferential obstacle term to allow robot to smoothly converge to goal pose in close proximity to obstacles.
+ | inflation_layer_name        | string    | Default "". Name of the inflation layer. If empty, it uses the last inflation layer in the costmap. If you have multiple inflation layers, you may want to specify the name of the layer to use. |
```
SteveMacenski, line 110: Add to navigation docs: https://navigation.ros.org/configuration/packages/configuring-mppic.html

SteveMacenski:
Otherwise LGTM. Make sure to sign the commits for DCO

SteveMacenski (approved):

PR #3831, opened 2023-09-22
Fixes for compiling with clang on macOS (arm64)

srmainwaring:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | #3828 |
| Primary OS tested on | MacOS (arm64 and intel) |
| Robotic platform tested on | Gazebo / ArduPilot simulation of Iris copter |

---

## Description of contribution in a few bullet points

* Changes fix compiler errors when building on macOS 13.4 with clang (version details below).
* Broken out into changes per submodule / type of error to aid dropping changes that may not be desired (e.g. replacement of `std::experimental::filesystem` with `std::filesystem`).
* Primary development and testing against ROS 2 Humble and the nav2 `humble` branch.
* This PR rebased onto `main` and built against ROS Rolling (the port of ROS Rolling to macOS is ongoing, so cannot fully test all nav2 libraries on macOS until that is complete).    

```bash
% clang --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

## Description of documentation updates required from your changes

* No functional changes.

---

## Future work that may be required in bullet points

* There are additional changes required to support macOS - these require more work to make them compatible with other platforms so have not been included at this stage.

#### For Maintainers: 
- [ ] Check that any new parameters added are updated in navigation.ros.org
- [ ] Check that any significant change is added to the migration guide
- [ ] Check that any new features **OR** changes to existing behaviors are reflected in the tuning guide
- [ ] Check that any new functions have Doxygen added
- [ ] Check that any new features have test coverage
- [ ] Check that any new plugins is added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

PR #4051, opened 2024-01-15
Fixes for compiling with clang on macOS (arm64)

srmainwaring:
Backport #3831 to `humble`.


srmainwaring:
Hi @SteveMacenski, is there a guide for preparing a backporting request? I'd like to get the macOS / clang support from #3831 into `humble` if possible.

SteveMacenski:
For some reason, I thought this changed ABI substantially but it doesn't appear to do so. I'm just checking that it compiles fine locally in Linux and then we can merge

PR #4055, opened 2024-01-17
Fix Smac Planner confined collision checker 

SteveMacenski:


DylanDeCoeyer-Quimesis:
@SteveMacenski I’m trying to understand the rationale for the added `+0.5f` in `inCollision()`.
As written, this changes the lookup from “cost of the cell containing the continuous pose” to “cost of the nearest cell”, because
```cpp
static_cast<unsigned int>(x + 0.5f)
```
rounds rather than floors.
My concern is that, in the context of collision checking, `getCost()` is being used to evaluate the cost at a pose, so I would expect the containing cell to be queried, which normally means flooring/truncation, not rounding.

Simple example:
- suppose `x = 10.6`, `y = 20.2`
- that pose is still inside cell `(10, 20)`
- but with `+0.5f`, the code queries `(11, 20)`

So if cell `(10, 20)` is lethal and `(11, 20)` is free, `inCollision()` would read the free neighbor cell even though the pose being evaluated still lies inside the lethal cell.

If it was the intention, could you explain why that is preferable in this collision-checking context? From my understanding, this does not seem correct.

PR #4056, opened 2024-01-17
Publish collision monitor state (#3504)

gennartan:
This commits is cherry-picked from b8d077e. The collision monitor now has an additional parameter named "state_topic". If the state topic is set, the collision monitor will publish its status to the defined topic so that other nodes might perform actions based on the current state of the collision monitor.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  (See PR #3504)  |
| Primary OS tested on | Ubuntu 22.04 + (See PR #3504) |
| Robotic platform tested on |  (See PR #3504)  |
| Does this PR contain AI generated software? | (See PR #3504) |

---

* Cherry pick commit b8d077e so that the collision monitor is able to publish the current state.

* Added new parameter, so need to add that to default configs and documentation page.

---

* Add documentation of the feature in the readme (I noticed this was done in another PR)

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/CMakeLists.txt
+++ b/nav2_collision_monitor/CMakeLists.txt
@@ -14,6 +14,7 @@ find_package(tf2_geometry_msgs REQUIRED)
 find_package(nav2_common REQUIRED)
 find_package(nav2_util REQUIRED)
 find_package(nav2_costmap_2d REQUIRED)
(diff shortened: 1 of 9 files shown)


SteveMacenski:
This is not ABI compatible, and thus cannot be backported.

Issue #4059, opened 2024-01-19
[Discussion] preparing migration to BehaviorTree.CPP 4.x


facontidavide:
Hi,

I would like to help to migrate Nav2 to BT.CPP 4.x before the release of **ROS J**.

I am playing with it and... it compiles on my machine (that is a start!)

Few initial thoughts:

1. Very stupid question first: does the **main** compile only in Rolling? 

2. Some methods  in BT.CPP are now `[[nodiscard]]` (modern c++, yay) . The changes to address that can be merged earlier and are backward compatible with BT.CPP 3.8 : #4060

3. I see that Nav2 broadly use "hard-coded" access to the blackboard, basically skipping the entire **InputPort** and **OutputPort**, Subtree remapping, mechanism (that exists for a reason). I would love to discuss in the next community meeting how to address this bad practice. I will try to write a small article about why and how that can be solved the next week.

4. I need to improve the automated XML converter to upgrade from version 3.X tyo 4.X [here](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/convert_v3_to_v4.py). Hopefully, that should takle care of most of the migration pain for many users.

Thoughts?

SteveMacenski:
> Very stupid question first: does the main compile only in Rolling?

Yes :-). If it happens to compile elsewhere, thats a nice to have, but not a requirement. We move fast.

> Some methods in BT.CPP are now [[nodiscard]] (modern c++, yay) . The changes to address that can be merged earlier and are backward compatible with BT.CPP 3.8 : https://github.com/ros-planning/navigation2/pull/4060

I'll take a look at that right after this

> I see that Nav2 broadly use "hard-coded" access to the blackboard, basically skipping the entire InputPort and OutputPort, Subtree remapping, mechanism (that exists for a reason). I would love to discuss in the next community meeting how to address this bad practice. I will try to write a small article about why and how that can be solved the next week.

I'd say "one thing at a time", but I'd have to see what things you're talking about. I think we use blackboards for almost everything except a few globals like `node` and a few `timeouts` which are set as attributes of the entire system and not of specific BT nodes. Certainly you could specify `node={node}` in every single BT node, but that makes the XML file unwieldy and redundant. 

But, open to critique. I would say though that we should table that until after the BT.CPP migration is done just not to muddle things up

[...]

PR #4060, opened 2024-01-19
[behavior_tree] support both ports and blackboard

facontidavide:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4059 |

---

* cosmetic: use `haltTree()` more consistently.

* `Blackboard::get` is marked `[[nodiscard]]` in BT.CPP 4.x. To prevent future warnings, I added EXPECT_TRUE in the test where the compiler would complain.

* Simplified `deconflictPortAndParamFrame`. T2 will always be a `const BT::TreeNode*`, so why writing more verbose code?

* Accessing directly the blackboard instead of using `inputPort` is an anti-pattern. `getInputPortOrBlackboard` is a temporary workaround that doesn't change that radically. A good solution must be discussed more broadly

Ports are defined for few BT nodes. If the port is defined, we will use that, otherwise we fallback to reading the blackboard, as we are currently doing.

Comments added in the modified/added functions

---

None of this change should affect the current behavior of the existing code.
In the future, we shoulkd consider how to avoid accessing the blackboard directly.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
@@ -132,20 +132,27 @@ inline std::set<int> convertFromString(StringView key)
 }
 
 /**
- * @brief Return parameter value from behavior tree node or ros2 parameter file
+ * @brief Return parameter value from behavior tree node or ros2 parameter file.
  * @param node rclcpp::Node::SharedPtr
  * @param param_name std::string
  * @param behavior_tree_node T2
  * @return <T1>
  */
-template<typename T1, typename T2>
+template<typename T1, typename T2 = BT::TreeNode>
 T1 deconflictPortAndParamFrame(
   rclcpp::Node::SharedPtr node,
   std::string param_name,
-  T2 * behavior_tree_node)
+  const T2 * behavior_tree_node)
 {
   T1 param_value;
-  if (!behavior_tree_node->getInput(param_name, param_value)) {
+  bool param_from_input = behavior_tree_node->getInput(param_name, param_value);
+
+  if constexpr (std::is_same_v<T1, std::string>) {
+    // not valid if port doesn't exist or it is an empty string
+    param_from_input &= !param_value.empty();
+  }
+
+  if (!param_from_input) {
     RCLCPP_DEBUG(
       node->get_logger(),
       "Parameter '%s' not provided by behavior tree xml file, "
@@ -161,6 +168,37 @@ T1 deconflictPortAndParamFrame(
     return param_value;
   }
 }
+
+/**
+ * @brief Try reading an import port first, and if that doesn't work
+ * fallback to reading directly the blackboard.
+ * The blackboard must be passed explitly because config() is private in BT.CPP 4.X
+ *
+ * @param bt_node node
+ * @param blackboard the blackboard ovtained with node->config().blackboard
+ * @param param_name std::string
+ * @param behavior_tree_node the node
+ * @return <T>
+ */
+template<typename T> inline
+BT::Result getInputPortOrBlackboard(
+  const BT::TreeNode & bt_node,
+  const BT::Blackboard & blackboard,
+  const std::string & param_name,
+  T & value)
+{
+  if (auto res = bt_node.getInput<T>(param_name, value)) {
+    return res;
+  }
+  if (auto res = blackboard.get<T>(param_name, value)) {
+    return {};
+  }
+  return nonstd::make_unexpected(StrCat(param_name, " not found"));
+}
+
+// Macro to remove boiler plate when using getInputPortOrBlackboard
+#define getInputOrBlackboard(name, value) \
+  getInputPortOrBlackboard(*this, *(this->config().blackboard), name, value);
 
 }  // namespace BT
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
@@ -181,19 +181,19 @@ T1 deconflictPortAndParamFrame(
  * @return <T>
  */
 template<typename T> inline
-BT::Result getInputPortOrBlackboard(
+bool getInputPortOrBlackboard(
   const BT::TreeNode & bt_node,
   const BT::Blackboard & blackboard,
   const std::string & param_name,
   T & value)
 {
-  if (auto res = bt_node.getInput<T>(param_name, value)) {
-    return res;
+  if (bt_node.getInput<T>(param_name, value)) {
+    return true;
   }
-  if (auto res = blackboard.get<T>(param_name, value)) {
-    return {};
+  if (blackboard.get<T>(param_name, value)) {
+    return true;
   }
-  return nonstd::make_unexpected(StrCat(param_name, " not found"));
+  return false;
 }
 
 // Macro to remove boiler plate when using getInputPortOrBlackboard
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
@@ -57,7 +57,10 @@ public:
    */
   static BT::PortsList providedPorts()
   {
-    return {};
+    return {
+      BT::InputPort<std::vector<geometry_msgs::msg::PoseStamped>>("goals"),
+      BT::InputPort<geometry_msgs::msg::PoseStamped>("goal"),
+    };
   }
 
 private:
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
@@ -58,8 +58,10 @@ public:
   static BT::PortsList providedPorts()
   {
     return {
-      BT::InputPort<std::vector<geometry_msgs::msg::PoseStamped>>("goals"),
-      BT::InputPort<geometry_msgs::msg::PoseStamped>("goal"),
+      BT::InputPort<std::vector<geometry_msgs::msg::PoseStamped>>(
+        "goals", "Vector of navigation goals"),
+      BT::InputPort<geometry_msgs::msg::PoseStamped>(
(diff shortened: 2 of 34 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
```diff
@@ -57,7 +57,10 @@ class GloballyUpdatedGoalCondition : public BT::ConditionNode
    */
   static BT::PortsList providedPorts()
   {
-    return {};
+    return {
```
SteveMacenski, line 60: For all instances: the BT xml index needs to be updated: https://github.com/ros-planning/navigation2/blob/main/nav2_behavior_tree/nav2_tree_nodes.xml

SteveMacenski: And description perhaps?

SteveMacenski (changes requested):
Generally seems good to me. These ports need to be added to navigation.ros.org for each of the BT nodes that were added & to the migration guide explaining that these now have the option to take in an input port for goal/goals

nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
```diff
@@ -162,6 +169,37 @@ T1 deconflictPortAndParamFrame(
...
+  const std::string & param_name,
+  T & value)
+{
+  if (auto res = bt_node.getInput<T>(param_name, value)) {
+    return res;
+  }
+  if (auto res = blackboard.get<T>(param_name, value)) {
+    return {};
+  }
+  return nonstd::make_unexpected(StrCat(param_name, " not found"));
```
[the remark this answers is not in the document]
facontidavide: There are two different idioms in the C++ world:
1. exceptions and try/catch
2. std::optional and std::expected (soon)

The C++ community is moving steadily toward the second one.
The arguments against exceptions are the runtime overhead and more verbose code to do error handling.
I do use exceptions in BT.CPP but only for... exceptional things! Something that you want to blow up in your face and say "there is an issue in your code, fix it".

An error that will likely happen, should not use exceptions.
This is one of those examples.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
facontidavide: I don't even know what the description should be...

facontidavide: Remember that that file should be generated automatically using this function, not edited by hand...

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/include/behaviortree_cpp/xml_parsing.h#L60

let me fix that

facontidavide:
@SteveMacenski I added a small utility to generate automatically the **nav2_tree_node.xml** file (as I originally intended).

See [this commit,](https://github.com/ros-planning/navigation2/pull/4060/commits/3d35add895281d2ebcfd128d07593427771e0c0a) that also automates the creation of a list of plugins.

The auto-generated XML is radically different. See attachment.

[nav2_tree_nodes.zip](https://github.com/ros-planning/navigation2/files/14051507/nav2_tree_nodes.zip)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
SteveMacenski: The description is that this is the navigation goal or navigation goals (if using navigate through poses).

I'm not terribly concerned how you update the XML index, just that its in sync

SteveMacenski:
> @SteveMacenski I added a small utility to generate automatically the nav2_tree_node.xml file (as I originally intended).

Remove that. PRs should be single feaetures with single implementations. Open another PR if you plan to add in additional features unrelated to this one please. That's how we get megaPRs with never ending review cycles

facontidavide:
Done: #4073

nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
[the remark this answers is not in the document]
facontidavide: ok, but I am not adding any try catch in my code, right?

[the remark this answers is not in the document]
facontidavide: what about returning a boolean instead?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
facontidavide: More about syncing the XML in #4073

nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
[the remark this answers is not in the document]
SteveMacenski: We run the tree wrapped in a try/except since there are things inside of other BT nodes that can throw (i.e. TF buffers, if control nodes are given incorrect configurations of their childs, etc), so you could throw.

A bool makes sense in this context since either `goal` or `goals` is populated, not both

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/globally_updated_goal_condition.hpp
SteveMacenski: Still need the XML index updates for this PR for these items

facontidavide:
ok, I will check what to change in **navigation.ros.org**

SteveMacenski:
How's this going? I think its just peanuts and this can be merged :-). You did the hard stuff already

facontidavide:
Yes, I think we should merge it and move on with the next one

facontidavide:
Added the fixes as discussed. Ready to merge?

SteveMacenski (approved):

facontidavide:
@SteveMacenski I will submit the change to the documentation today.

Heads up! Some of the changes in this PR might be short-lived, once we migrate to BT.CPP 4.6 (to be released soon),

This new syntax is possible (4 different ways to express the default value of a port):
```XML
  static PortsList providedPorts()
  {
    return {BT::InputPort<Point2D>("input", "no default value"),
            BT::InputPort<Point2D>("pointA", Point2D{1, 2}, "default value is [1,2]"),
            BT::InputPort<Point2D>("pointB", "{point}", "default value inside blackboard {point}"),
            BT::InputPort<Point2D>("pointC", "5,6", "default value is [5,6]"),
            BT::InputPort<Point2D>("pointD", "{=}", "default value inside blackboard {pointD}")};
  }
```

This means that the new code in NAV2 will be:

```cpp
BT::InputPort<std::vector<geometry_msgs::msg::PoseStamped>>("goals","{=}", "Vector of navigation goals"),
BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "{=}, "Navigation goal"),
```

That means: "If I am lazy and don't specify the port in the XML, search in the blackboard using the SAME name of the port".
But this specific PR tries to stay compatible with BT.CPP 3.8, therefore I will postpone this change.

facontidavide:
Aaaaand... done! @SteveMacenski 
https://github.com/ros-planning/navigation.ros.org/pull/524

Note that I am not adding anything to the migration guide because it is TOO early, in my opinion

1. There will be more changes
2. The changes in this specific PR are backward compatible and no action should be taken.

So, let me postpone that documentation improvement to future PRs!

SteveMacenski:
> So, let me postpone that documentation improvement to future PRs!

Perfectly fine, this is just for the fact that ports so API changed and its easier to keep those always in sync than remember in the back of my mind the places where I know it needs to change later. The migration guide for the larger v4 migration can certainly wait, but mentioning that the blackboard or port can be used for goal/goals is worth mentioning. 

>  once we migrate to BT.CPP 4.6

Whats your stable v4 version? We're not going to be able to be on your bleeding edge version of v4, its too much risk for too large an audience to go there until its had baking time with early adopters

facontidavide:
> Whats your stable v4 version? We're not going to be able to be on your bleeding edge version of v4, its too much risk for too large an audience to go there until its had baking time with early adopters

Technically speaking all the version are stable... I have unit tests 😄 

But from a behavioral point of view (pun not intended), after 4.6 I have no plan to add features. Only bugfixes until Jazzy.

SteveMacenski:
This is just pending on a minor docs update if you can do that, we can have this merged in an hour.

> And I will release 4.6 the next week. My proposed changes to Nav2 will depend on that version. Also, I wan to point out that some of the features in 4.6 are being created specifically for Nav2, to help migration 😉

I guess my conservative "I don't want to break _100+_ companies" mind says not to use 4.6 and instead use 4.5 until 4.6 has been around for at least a quarter and the top ~5 bugs have been reported/fixed. But, I also don't know how many core changes were made in that subversion update. Was it major or minor updates (and/or major to existing features or adding new features that wouldn't effect existing ones)? I'm more open to potential breaks in new features we don't use yet in the core of Nav2 than I would be for potential breaks in core things in Nav2 folks rely on. ROS 2 has not-so-good reputation due to RWM vendor breakages and issues, I want to make sure Nav2 doesn't fall into the same camp.

mikeferguson:
> "I don't want to break _100+_ companies"

Oh come on! That strategy seems to work fine for MoveIt!

SteveMacenski:
It'd be a shame if something happened to your robot, see..

<div class="tenor-gif-embed" data-postid="8283674910861859030" data-share-method="host" data-aspect-ratio="1.77857" data-width="100%"><a href="https://tenor.com/view/luca-changretta-peaky-blinders-mafia-italian-gif-8283674910861859030">Luca Changretta Peaky Blinders GIF</a>from <a href="https://tenor.com/search/luca+changretta-gifs">Luca Changretta GIFs</a></div> <script type="text/javascript" async src="https://tenor.com/embed.js"></script>

facontidavide:
I think you are making some assumptions about "cutting edge" but.cpp that are not correct, and I need to disagree.

I can only agree with you about the importance of stability, but let me repeat myself: the latest changes are specifically created to help migration to Nav2.

Once I have the change log, we can discuss it further.

PR #4065, opened 2024-01-25
Allow path end pose deviation revive

jwallace42:
This pr revives the work done [here](https://github.com/ros-planning/navigation2/pull/3928) by @pepisg


SteveMacenski:
I think we were talking about having the param removed and always using the path's final goal as the start to the new?

pepisg:
> I think we were talking about having the param removed and always using the path's final goal as the start to the new

Yeah, I think that was what we agreed on slack. Also, discussion was gravitating around possible artifacts happening on Lattice. Finally, I wasn't sure how test should look like for this change

nav2_planner/include/nav2_planner/planner_server.hpp
```diff
@@ -241,6 +241,7 @@ class PlannerServer : public nav2_util::LifecycleNode
   std::vector<std::string> planner_types_;
   double max_planner_duration_;
   std::string planner_ids_concat_;
+  bool allow_path_through_poses_goal_deviation_;
```
SteveMacenski, line 244: Unused now :smile:

nav2_planner/src/planner_server.cpp
```diff
@@ -52,6 +52,7 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
   // Declare this node's parameters
   declare_parameter("planner_plugins", default_ids_);
   declare_parameter("expected_planner_frequency", 1.0);
+
```
SteveMacenski, line 55: remove empty line

SteveMacenski (changes requested):

SteveMacenski (approved):

PR #4067, opened 2024-01-25
[Smac Planner] Massive Improvement of Behavior for SE2 Footprint Checking (ie non-circular robots) In Confined Settings

SteveMacenski:
@doisyg please test on the map we did onsite in Verda Park to verify this still solves the problem for you after my refactor


doisyg:
Tested, amazing speed and behavior improvement ! GAME CHANGER

SteveMacenski:
All I needed to hear :-)

PR #4074, opened 2024-01-26
[behavior_tree] don't repeat yourself in "blackboard->set"

facontidavide:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | not applicable |

---

The only purpose of this PR is cosmetic.
Template deduction is doing its job, no need to write more verbose code.

I am just keeping `blackboard->set<std::chrono::milliseconds>` because is the only case where someone may accidentally introduce a bug (may fix this in the future inside BT.CPP itself, to be more permissive).

Not applicable

---

#### For Maintainers: 


SteveMacenski (approved):
Once upon a time that didn't work, but glad to now

PR #4031, opened 2024-01-04
Twist stamped migration

SteveMacenski:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   |[ (add tickets here #1)](https://github.com/ros-planning/navigation2/issues/1594) |
| Primary OS tested on | Ubuntu, |
| Robotic platform tested on |  CI |
| Does this PR contain AI generated software? | None |

---

PR #3775, opened 2023-08-27
Add TwistStamped support via TwistPublisher and TwistSubscriber

Ryanf55:
## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | #1594  |
| Primary OS tested on | Ubuntu 22.04 |
| Robotic platform tested on | CI |

---

## Description of contribution in a few bullet points

* Add a new util class for publishing either Twist or TwistStamped called `TwistPublisher`
* Add a new util class for subscribing to either Twist or TwistStamped called `TwistSubscriber`
* Add a new parameter for selecting to stamp the twist data to each util
* Consume TwistPublisher in nav2_controller, among other nodes
* Consume TwistSubscriber in nav2_velocityy_smoother, among other nodes
* Implements the approach described [here](https://github.com/ros-planning/navigation2/issues/1594#issuecomment-1675262038): `But, I suppose we could make a velocity publisher wrapper in nav2_util that deals with that parameterization outside of the application code (and in a way we can remove the non-stamped support more gradually)`.

## Description of documentation updates required from your changes

* Would need to document the new optional parameter behavior in the normal docs on top of the README.

## Future work that may be required in bullet points

* Related PR's in gazebo, etc, to convert the ecosystem to TwistStamped

## Test Instructions

```bash
colcon test --packages-select nav2_util --event-handlers=console_cohesion+ --ctest-args " -R" " test_twist"
```

Quick iteration
```bash
colcon build --packages-select nav2_util && colcon test --packages-select nav2_util --event-handlers=console_cohesion+ --ctest-args " -R" " test_twist"
```

#### For Maintainers: 
- [ ] Check that any new parameters added are updated in navigation.ros.org
- [ ] Check that any significant change is added to the migration guide
- [ ] Check that any new features **OR** changes to existing behaviors are reflected in the tuning guide
- [ ] Check that any new functions have Doxygen added
- [ ] Check that any new features have test coverage
- [ ] Check that any new plugins is added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists

PR #4077, opened 2024-01-26
1594 twist stamped publisher

Ryanf55:
See https://github.com/ros-planning/navigation2/pull/3775; this is the same PR but opened back up. 

Replaces https://github.com/ros-planning/navigation2/pull/4031, by pulling in all 3 commits, as well as more changes since then. 

This has been tested by backporting to humble against ArduPilot by @srmainwaring. No bugs were reported.


SteveMacenski (approved):
Note to self once ripping out non-stamped:
- [ ] Check constructor for stamped only is OK
- [ ] Change all uses of publisher / subscriber to the stamped only option
- [ ] Delete all non-stamped function overrides 
- [ ] Re-check all timestamps and headers are properly propagated (not reassigned, overrided, etc after time of computation of the root cmd). 
- [ ] Remove enabled stamped param from param guide + mention  in migration guide total deprecation

Issue #3269, opened 2022-11-04
Feature: select controllers and planners from RViz


padhupradheep:
#### Feature description
I had this idea of creating an RViz plugin for changing the controllers, planners, etc on the fly from the RViz. Like if someone is testing out a new controller, he or she might always have to use the controller_selector topic to publish the `controller_id` (provided they have a controller_selector node included in their BT plugin). Of course this can all be done from the terminal. This RViz plugin will just provide some additional cushion for users.  

Some may want to even tryout different combinations of planners and controllers, in that case too, it would be easy to keep track of the planners and the controllers that are in use from the RViz plugin. Also it acts as a status indicator on which controller_id is being utilized for the process at hand.  

#### Implementation considerations
Currently I just did a small prototype as shown below (yes, it comes with Nav2 logo as well :smile: ,  but not sure if I should keep it because of the space constraints )

![select_planner_controller](https://user-images.githubusercontent.com/20242192/199979039-8f7c7ab6-de68-4461-95b9-787c8c772079.png)

So, when the robot is in the navigation state, you cannot do any changes to the planner id and the controller id. The robot has to reach the goal in order to change the planner or controller for the next process. 

CC: @jwallace42

SteveMacenski:
> it comes with Nav2 logo as well smile , but not sure if I should keep it because of the space constraints

Could always put it on the left-hand side next to the controller/planner selector dropdowns instead of on top :wink: 

I think this makes sense as an independent plugin like you created. I thought about it being part of the default window, but I think this might be good for some users, but not something that's exposed to them right away by default. We'd need some docs along with this to explain how to use it (e.g. should add these selector nodes to your BT (or maybe even add all the selectors to our BTs so it works out of the box?)) 

How does that list of potential options get populated?

padhupradheep:
> Could always put it on the left-hand side next to the controller/planner selector dropdowns instead of on top wink Maybe could have it added to the Nav2 panel on the right side aligned with the navigation / localization / feedback things (since that doesn't take the full width)

Sure, that is a nice idea. I also looked into the possibility of adding a GIF, something like the parachute taking off from the ground, and maybe just loop it once or twice. But, that seems to be a lot of work :-D .

> We'd need some docs along with this to explain how to use it e.g. should add these selector nodes to your BT

[...]

PR #4091, opened 2024-02-01
Selector plugin for selecting the planner and the controller

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3269) |
| Robotic platform tested on | Neobotix MPO-700 |

---

Now we can select the controller and the planner plugins on the fly

Maybe RViz plugins after all needs a documentation somewhere?

---

* Goal checkers and smoothers could be added to the selection list if needed

#### For Maintainers: 

### DIFF
--- a/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
+++ b/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
@@ -8,11 +8,11 @@
   <BehaviorTree ID="MainTree">
     <RecoveryNode number_of_retries="6" name="NavigateRecovery">
       <PipelineSequence name="NavigateWithReplanning">
-        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath2" topic_name="controller_selector"/>
+        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath" topic_name="controller_selector"/>
         <PlannerSelector selected_planner="{selected_planner}" default_planner="GridBased" topic_name="planner_selector"/>
         <RateController hz="1.0">
           <RecoveryNode number_of_retries="1" name="ComputePathToPose">
-            <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased" error_code_id="{compute_path_error_code}"/>
+            <ComputePathToPose goal="{goal}" path="{path}" planner_id="{selected_planner}" error_code_id="{compute_path_error_code}"/>
             <Sequence>
               <WouldAPlannerRecoveryHelp error_code="{compute_path_error_code}"/>
               <ClearEntireCostmap name="ClearGlobalCostmap-Context" service_name="global_costmap/clear_entirely_global_costmap"/>
--- a/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
+++ b/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
@@ -8,6 +8,7 @@
   <BehaviorTree ID="MainTree">
     <RecoveryNode number_of_retries="6" name="NavigateRecovery">
       <PipelineSequence name="NavigateWithReplanning">
+        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath" topic_name="controller_selector"/>
         <RateController hz="1.0">
           <RecoveryNode number_of_retries="1" name="ComputePathToPose">
             <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased" error_code_id="{compute_path_error_code}"/>
@@ -18,7 +19,7 @@
           </RecoveryNode>
         </RateController>
         <RecoveryNode number_of_retries="1" name="FollowPath">
-          <FollowPath path="{path}" controller_id="FollowPath" error_code_id="{follow_path_error_code}"/>
+          <FollowPath path="{path}" controller_id="{selected_controller}" error_code_id="{follow_path_error_code}"/>
           <Sequence>
             <WouldAControllerRecoveryHelp error_code="{follow_path_error_code}"/>
             <ClearEntireCostmap name="ClearLocalCostmap-Context" service_name="local_costmap/clear_entirely_local_costmap"/>
--- a/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
+++ b/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
@@ -8,7 +8,8 @@
   <BehaviorTree ID="MainTree">
     <RecoveryNode number_of_retries="6" name="NavigateRecovery">
       <PipelineSequence name="NavigateWithReplanning">
-        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath" topic_name="controller_selector"/>
+        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath2" topic_name="controller_selector"/>
+        <PlannerSelector selected_planner="{selected_planner}" default_planner="GridBased" topic_name="planner_selector"/>
         <RateController hz="1.0">
           <RecoveryNode number_of_retries="1" name="ComputePathToPose">
             <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased" error_code_id="{compute_path_error_code}"/>
--- a/nav2_rviz_plugins/src/selector.cpp
+++ b/nav2_rviz_plugins/src/selector.cpp
@@ -95,7 +95,7 @@ Selector::timerEvent(QTimerEvent * event)
       controller->addItems(id_list);
       id_list.clear();
     } else {
-      auto parameters_con = parameter_client_con->get_parameters({"FollowPath"});
+      auto parameters_con = parameter_client_con->get_parameters({"FollowPath.plugin"});
       std::cout<<parameters_con.size()<<std::endl;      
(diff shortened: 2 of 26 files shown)


nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
```diff
@@ -8,6 +8,8 @@
   <BehaviorTree ID="MainTree">
     <RecoveryNode number_of_retries="6" name="NavigateRecovery">
       <PipelineSequence name="NavigateWithReplanning">
+        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath2" topic_name="controller_selector"/>
```
SteveMacenski, line 11: "FollowPath2"?

I assume the BT changes here and in the other file are temporary / not complete. In general, I like the first-step of setting the selectors as you did in the 2 files. Set the defaults to `GridBased` `FollowPath`, then use those blackboard entries in the FollowPath / ComputePathToPose BT nodes! 

Just make sure to update all the relevant BTs

nav2_rviz_plugins/src/selector.cpp
```diff
@@ -0,0 +1,129 @@
...
+      plugins_loaded_ = true;
+    }
+    timer_.stop();
+  }
+}
+
+void
+Selector::start_ros_timer()
+{
+  rclcpp::spin(node_);
```
SteveMacenski, line 124: Where is this `node_` set?

SteveMacenski (commented):
Generally looks good to me! What's this tool look like? :smile:

SteveMacenski:
Also, seems easy enough to add the smoother / goal checker / progress checkers here too! Nice, simple tool, I like it!

nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml
padhupradheep: > Just make sure to update all the relevant BTs

You mean the docs right?

SteveMacenski:
Progress checker too :heart: 

I think then its just the BT updates and perhaps this is good for testing / merging? What do you think?

padhupradheep:
> Progress checker too

There is no BT node for that as far as I checked. We can use the dynamic reconf server for it but that would be something unique and really won't fit to the scope of this PR. 

> I think then its just the BT updates and perhaps this is good for testing / merging? What do you think?

Yes, almost done!  I need to update the docs and probably you can give it a swing when you get some time and we are all set for the merge once after that.

padhupradheep:
Shall I add this plugin as default in nav2 bringup ? 

Since it's more of a development tool, I feel it would be better if the user selects it by hand when required.

SteveMacenski:
Yeah that sounds great!

padhupradheep:
image

bit more compact version..

Would this work?

SteveMacenski:
Yeah that new 2x2 version works, though why not have the drop down on the same line as the label instead of a new line? Either way works.

If you're going to make a new gif with the format, try to use not just `FollowPath` and `FollowPath2`, but how about `FollowPath` and `AvoidObstacles` or something else with different modality implications.

padhupradheep:
> why not have the drop down on the same line as the label instead of a new line

I tried it yesterday, It felt cramped by this way.

image

I'd stick to the other version..

padhupradheep:
Just leaving a short note on having the selector plugin has default in the rviz bringup. There is some sort of race condition that does not allow the selector plugin to populate atleast one of the component's combo boxes, atleast once in ten instances.

SteveMacenski:
> There is some sort of race condition that does not allow the selector plugin to populate atleast one of the component's combo boxes, atleast once in ten instances.

Do you see your error log about the server not being up yet? Can you change the ` auto parameters = parameter_client->get_parameters({plugin_type});` line to use the API which would give more feedback on failure?

It would be good to know where this is happening so we can see what might need to be retried / delayed to get it to work deterministically. 1/10 failures is not something we could merge :(

nav2_rviz_plugins/src/selector.cpp
```diff
@@ -0,0 +1,184 @@
...
+  rclcpp::Node::SharedPtr node,
+  const std::string & server_name,
+  const std::string & plugin_type,
+  QComboBox * combo_box)
+{
+  auto parameter_client = std::make_shared<rclcpp::SyncParametersClient>(node, server_name);
+
+  // Wait for the service to be available before calling it
+  bool server_unavailable = false;
+  while (!parameter_client->wait_for_service(3s)) {
```
padhupradheep, line 138: We get the issue exactly here. As you see indeed have a logger for it too below. It's always the `controller_server` which is not available.. Probably we could wait couple of more seconds here?

SteveMacenski: Possibly yes! Or you could wait once, then just pass the callback and come back again in another 200ms to try again, rather than `while` looping

padhupradheep:
Issue is fixed! Thanks for the suggestion! :+1:

SteveMacenski:
CI had error I assume isn't related to this PR, but retriggering to check.

In the meantime, anything else you want to do before I test and merge?

padhupradheep:
> In the meantime, anything else you want to do before I test and merge?

Well, I just added the progress_checker too. I just updated from 2x2 to 3x2 configuration, with slight changes to the way combo boxes would be placed. 

Now I think, it's ready to be tested.

SteveMacenski:
> CI had error I assume isn't related to this PR, but retriggering to check.

Huh, your PR is regularly triggering these failures. Can you reproduce locally? Maybe try pulling in `main`?

padhupradheep:
I already did a rebase early today.. not sure what's causing the issue. Let me check..

SteveMacenski:
Beyond that, I tested it locally and looks good to me

padhupradheep:
For some reason this is happening..

` [bt_navigator-14] [ERROR] [1707415321.396377533] [bt_navigator]: Exception when loading BT: Error at line 10: -> Node not recognized: ControllerSelector`

SteveMacenski:
>  [bt_navigator-14] [ERROR] [1707415321.396377533] [bt_navigator]: Exception when loading BT: Error at line 10: -> Node not recognized: ControllerSelector

Oh. Note that you have to update the BT nodes to default load in a few places (also, add in progress checker while you're at it, I totally missed it too): 

- https://github.com/ros-planning/navigation2/blob/main/nav2_bt_navigator/src/bt_navigator.cpp
- https://github.com/ros-planning/navigation2/blob/main/nav2_bringup/params/nav2_params.yaml (and other 3 next to it)

Crucially the following:
- https://github.com/ros-planning/navigation2/blob/0cdabca2f27191da7ef3f8f1edff75e9e2e925aa/nav2_system_tests/src/costmap_filters/keepout_params.yaml#L50
- https://github.com/ros-planning/navigation2/blob/0cdabca2f27191da7ef3f8f1edff75e9e2e925aa/nav2_system_tests/src/costmap_filters/speed_local_params.yaml#L50
- https://github.com/ros-planning/navigation2/blob/0cdabca2f27191da7ef3f8f1edff75e9e2e925aa/nav2_system_tests/src/gps_navigation/nav2_no_map_params.yaml#L18
- https://github.com/ros-planning/navigation2/blob/0cdabca2f27191da7ef3f8f1edff75e9e2e925aa/nav2_system_tests/src/costmap_filters/speed_global_params.yaml#L50

This requirement is not long for this world, @facontidavide opened a PR which has a tool which autopopulates this which will remove this need in the coming weeks. But for now, we manually set the BT nodes in the plugins lib list.

padhupradheep:
Ohhh.. this again.. I remember we faced the same trouble when we developed all the `cancel BTs` last year.

Glad that it's not needed anymore..

padhupradheep:
Okay, now the release_test fails because of someother reason.. :thinking:

SteveMacenski:
What do you mean? Its failing for the exact same reason: https://app.circleci.com/pipelines/github/ros-planning/navigation2/10965/workflows/724ba00e-3d4b-49ec-a289-d41bf0a1e4ac/jobs/34394/tests

It doesn't seem that you pushed any updates to add the new BT nodes like we discussed.

padhupradheep:
My mistake, somehow messed up the commit while fixing the DCO. I just was following the recommendation it was giving out, but somehow it broke my changes :man_shrugging:

padhupradheep:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/10966/workflows/2b3de681-a592-4539-866f-bcfe9e88dedc/jobs/34400/tests

Now it seems to be the `test_spin_behavior_fake.missing_result` and  `test_drive_on_heading_recovery.missing_result`

SteveMacenski:
I was messing around there recently, but it all passed on my CI jobs. Maybe increment your cache to v20 in the 3x places in the circleci file to break it all? If that still fails, I'll take that as an action item to look into, I won't block you further with it.

padhupradheep:
v20 didn't help either..

please let me know, if there is something else that I can try to sort it outt.

SteveMacenski:
Not your problem, probably me. If you wanted to take a look at it, the more the merrier :wink: 

Want to post on discourse with your new feature + gif? I think that's worthy of highlighting!

padhupradheep:
> Not your problem, probably me. If you wanted to take a look at it, the more the merrier 😉

:smile: 

> Want to post on discourse with your new feature + gif? I think that's worthy of highlighting!

Oh! Sure can do that.. :+1:

padhupradheep:
It took me one whole year to eventually get on this PR and get it up.. :face_with_spiral_eyes: 

I'll clear #3347 too..

If you have anything else, feel free to tag me in..

SteveMacenski:
Thanks and will do!

PR #4112, opened 2024-02-12
Use ament_export_targets for all targets in nav2_costmap_2d

Ryanf55:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  https://github.com/ANYbotics/grid_map/issues/403 |
| Primary OS tested on |  Ubuntu 22.04  |
| Robotic platform tested on | N/A ||

---

* Switch nav2_costmap_2d to use the latest recommended usage of ament_cmake with ament_export_targets.
* Export includes directories with targets that use it
* Add support for consuming packages to link against the `nav2_costmap_2d::nav2_costmap_2d_core` target
* Use ALIAS library for internal linkage when possible on all libraries
* Don't use ament_target_dependencies when already linking to core because it's unnecessary

N/A

---

* Concerting more of NAV2 to abide by the Ament CMake recommendations.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/CMakeLists.txt
+++ b/nav2_costmap_2d/CMakeLists.txt
@@ -26,16 +26,9 @@ find_package(visualization_msgs REQUIRED)
 find_package(angles REQUIRED)
 
 remove_definitions(-DDISABLE_LIBUSB-1.0)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3 3.3 REQUIRED)
 
 nav2_package()
-
-include_directories(
-  include
-  ${EIGEN3_INCLUDE_DIRS}
-)
-
-add_definitions(${EIGEN3_DEFINITIONS})
 
 add_library(nav2_costmap_2d_core SHARED
   src/costmap_2d.cpp
@@ -50,6 +43,13 @@ add_library(nav2_costmap_2d_core SHARED
   src/clear_costmap_service.cpp
   src/footprint_collision_checker.cpp
   plugins/costmap_filters/costmap_filter.cpp
+)
+add_library(${PROJECT_NAME}::nav2_costmap_2d_core ALIAS nav2_costmap_2d_core)
+
+target_include_directories(nav2_costmap_2d_core
+  PUBLIC
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+    "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>"
 )
 
 set(dependencies
@@ -78,6 +78,7 @@ set(dependencies
 ament_target_dependencies(nav2_costmap_2d_core
   ${dependencies}
 )
+target_link_libraries(nav2_costmap_2d_core Eigen3::Eigen)
 
 add_library(layers SHARED
   plugins/inflation_layer.cpp
@@ -88,11 +89,12 @@ add_library(layers SHARED
   plugins/range_sensor_layer.cpp
   plugins/denoise_layer.cpp
 )
+add_library(${PROJECT_NAME}::layers ALIAS layers)
 ament_target_dependencies(layers
   ${dependencies}
 )
 target_link_libraries(layers
-  nav2_costmap_2d_core
+  ${PROJECT_NAME}::nav2_costmap_2d_core
 )
 
 add_library(filters SHARED
@@ -100,11 +102,14 @@ add_library(filters SHARED
   plugins/costmap_filters/speed_filter.cpp
   plugins/costmap_filters/binary_filter.cpp
 )
+add_library(${PROJECT_NAME}::filters ALIAS filters)
+
+
 ament_target_dependencies(filters
   ${dependencies}
 )
 target_link_libraries(filters
-  nav2_costmap_2d_core
+  ${PROJECT_NAME}::nav2_costmap_2d_core
 )
 
 add_library(nav2_costmap_2d_client SHARED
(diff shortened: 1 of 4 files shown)


Ryanf55:
Looks like certain code is trying to link to executables. I have no idea why ament_export_libraries would throw executables in things you want to link to, but it does...
```
CMake Error at CMakeLists.txt:15 (add_library):
  Target "costmap_queue" links to target
  "nav2_costmap_2d::nav2_costmap_2d_markers" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
  ```
  
  Do you consider the link requirements part of ABI?

**Comment by SteveMacenski** [comment]
> Do you consider the link requirements part of ABI?

Yes

File: nav2_costmap_2d/CMakeLists.txt [hunk 1]
```diff
@@ -26,17 +26,10 @@ find_package(visualization_msgs REQUIRED)
 find_package(angles REQUIRED)
 
 remove_definitions(-DDISABLE_LIBUSB-1.0)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3 3.3 REQUIRED)
 
 nav2_package()
 
-include_directories(
-  include
```
SteveMacenski, line 34: Readd

nav2_costmap_2d/CMakeLists.txt
```diff
@@ -141,34 +147,30 @@ ament_target_dependencies(nav2_costmap_2d
...
   nav2_costmap_2d_client
+  nav2_costmap_2d_markers
+  nav2_costmap_2d_cloud
+  EXPORT export_${PROJECT_NAME}
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
 
-install(TARGETS
```
SteveMacenski, line 159: Readd. The install locations of executables and libraries are different.

SteveMacenski (changes requested):

nav2_costmap_2d/CMakeLists.txt
Ryanf55: Re-added. You can do different destinations for both in the same `install` command by changing the `RUNTIME DESTINATION`, however it looks like this causes the nodes to be exported in the export set which is would break consumers using ament_target_dependencies because you can't link to an executable.

Existing behavior is preserved for two installs, which is in line with the tutorials.

Ryanf55:
> Looks like certain code is trying to link to executables. I have no idea why ament_export_libraries would throw executables in things you want to link to, but it does... This could be a mistake in usage here by having the executable be exported in this way.
> 
> ```
> CMake Error at CMakeLists.txt:15 (add_library):
>   Target "costmap_queue" links to target
>   "nav2_costmap_2d::nav2_costmap_2d_markers" but the target was not found.
>   Perhaps a find_package() call is missing for an IMPORTED target, or an
>   ALIAS target is missing?
> ```
> 
> Do you consider the link requirements part of ABI?

This is fixed now, the rest of `nav2` builds fine without any changes, so ABI is preserved.

nav2_costmap_2d/CMakeLists.txt
```diff
@@ -26,17 +26,10 @@ find_package(visualization_msgs REQUIRED)
 find_package(angles REQUIRED)
 
 remove_definitions(-DDISABLE_LIBUSB-1.0)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3 3.3 REQUIRED)
```
SteveMacenski, line 29: Why not use the installed / OS  version? This adds fragility that we have to keep updating this by the current ubuntu version.

SteveMacenski (changes requested):
I don't necessarily have a problem with any of this, but why exactly are you making these changes? What is this enabling for you that wasn't happening before?

nav2_costmap_2d/CMakeLists.txt
Ryanf55: This specified a minimum version, not an exact match. Ubuntu 22 comes with Eigen 3.4, and the build passes, so this proves that it's not an exact match.

nav2_costmap_2d/CMakeLists.txt
Ryanf55: The packages own include directory is now handled by the `BUILD_INTERFACE` generator for `target_include_directories. I moved that call down, because you can't call `target_include_directories` before the target exists.

nav2_costmap_2d/CMakeLists.txt
SteveMacenski: How did you pick 3.3?

Ryanf55:
> I don't necessarily have a problem with any of this, but why exactly are you making these changes? What is this enabling for you that wasn't happening before?

As mentioned in the PR description, it enables using the targets from nav2_costmap_2d with `target_link_libraries` and only link to the library targets that are actually needed. Compared to `ament_target_dependencies` which links all dependencies to all targets and does not support PRIVATE linkage, target_link_libraries allow for finer control and reduced coupling. `ament_export_targets` replaces the previous method when using targets.
image

I've done this change across a number of ROS 2 projects in the open-source area with no reported issues.

Ryanf55:
Looks like the build passed, but the release test failed for unrelated reasons. I've marked it out of draft now that the build is passing.

nav2_costmap_2d/CMakeLists.txt
Ryanf55: I just tested. NAV2 compiles with Eigen3.3.9. That said, CI compiles with 3.4(what comes with Ubuntu 22), so 3.3 is untested. We can leave as-is.

Ryanf55:
Is there any way this can be backported to `iron` and `humble`?

SteveMacenski:
I see no reason why not with the possible exception for `humble` that something might be so old there that this PR added rules that don't exist yet creating a probably-minor merge conflict. I don't think that is the case though and even if it were, pretty straight forward (hopefully).

Issue #4033, opened 2024-01-05
Feature request: Graceful cancellation of `follow_path` actions


Rayman:
## Feature request
Ability to let controllers handle cancelations gracefully

#### Feature description
At the moment when the `follow_path` action is cancelled, a zero velocity is immediately published ([see the code](https://github.com/ros-planning/navigation2/blob/305f7e039f1183b94981ec6a7f294f0ed7b89d07/nav2_controller/src/controller_server.cpp#L463) and the figure below).
We have a big heavy robot that needs a lot of time to slow down so we want to keep following the current path while decelerating within the deceleration limits.

With move base flex, we could override the `cancel` method and return when the cancellation was performed (see 
https://github.com/magazino/move_base_flex/issues/171)
Is there a way to achieve this graceful cancel or can we add the same API to Navigation2?

![image](https://github.com/ros-planning/navigation2/assets/91969/68df7e99-6e55-48f2-8dca-7742f95f59c6)

Timple:
The idea is to decelerate by dynamic limits _whilst tracking the path_. 

When following a turn, it would be undesired to slow down in a straight line upon a cancel action, which I think/guess is what the `nav2_velocity_smoother` does?

SteveMacenski:
I also don't think that's actually a generically safe solution? Deceleration in turns could cause the kinds of "throwing loads off my robot top" that the instantaneous braking could due to centrifugal forces. I imagine as well there are techniques where reducing the velocity scale will actually make it deviate off the path due to its internal modeling (thinking MPC/MPPI which are maintaining some state) 

This isn't my opposition to your request - just throwing out some additional constraints perhaps that weren't considered giving me pause

SteveMacenski:
> When following a turn, it would be undesired to slow down in a straight line upon a cancel action, which I think/guess is what the nav2_velocity_smoother does?

The Vel Smoother will ramp down from the last request to the zero velocity. It won't necessarily be a straight line. If there's some non-zero angular component, that will be ramped down too by its own deceleration constraints

Rayman:
When the robot is paused during operation (no emergency stop)
```
target_x_acc: 0.25 m/s^2
target_x_vel: 3.8333 m/2. 
```
That means it takes ~15 seconds to fully slow down. The braking distance is (if I remember correctly)
```
s = v^2/2/a = 29.4 m
```
[...]

PR #4136, opened 2024-02-22
Add API to gracefully cancel a controller

Rayman:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4033 |
| Robotic platform tested on | Gazebo simulation of inhouse robot |

---

Ability to let controllers handle cancelations gracefully. See the discussion at https://github.com/ros-planning/navigation2/issues/4033.

I'm opening this PR for a first review of the API. I've added gracefull cancellation to one of our in-house controllers and tested that this API is sufficient. I hope the API fits the design of the core APIs. 

---

- [x] Add gracefull cancellation to one of the existing controllers.
- [ ] Merge https://github.com/ros-planning/navigation.ros.org/pull/535

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -460,10 +460,15 @@ void ControllerServer::computeControl()
       }
 
       if (action_server_->is_cancel_requested()) {
-        RCLCPP_INFO(get_logger(), "Goal was canceled. Stopping the robot.");
-        action_server_->terminate_all();
-        publishZeroVelocity();
-        return;
+        if (controllers_[current_controller_]->cancel()) {
+          RCLCPP_INFO(get_logger(), "Cancellation was successful. Stopping the robot.");
+          action_server_->terminate_all();
+          publishZeroVelocity();
+          return;
+        } else {
+          RCLCPP_INFO_THROTTLE(
+            get_logger(), *get_clock(), 1000, "Waiting for the controller to finish cancellation");
+        }
       }
 
       // Don't compute a trajectory until costmap is valid (after clear costmap)
--- a/nav2_core/include/nav2_core/controller.hpp
+++ b/nav2_core/include/nav2_core/controller.hpp
@@ -117,6 +117,16 @@ public:
     nav2_core::GoalChecker * goal_checker) = 0;
 
   /**
+   * @brief Cancel the current control action
+   * @return True if the cancellation was successful. If false is returned, computeVelocityCommands
+   * will be called until cancel returns true.
+   */
+  virtual bool cancel()
+  {
+    return true;
+  }
+
+  /**
    * @brief Limits the maximum linear speed of the robot.
    * @param speed_limit expressed in absolute value (in m/s)
    * or in percentage from maximum robot speed.
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/parameter_handler.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/parameter_handler.hpp
@@ -53,6 +53,7 @@ struct Parameters
   double curvature_lookahead_dist;
   bool use_rotate_to_heading;
   double max_angular_accel;
+  double cancel_deceleration;
   double rotate_to_heading_min_angle;
   bool allow_reversing;
   double max_robot_pose_search_dist;
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -86,7 +86,7 @@ ParameterHandler::ParameterHandler(
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".max_angular_accel", rclcpp::ParameterValue(3.2));
   declare_parameter_if_not_declared(
-    node, plugin_name_ + ".cancel_deceleration", rclcpp::ParameterValue(1.0));
+    node, plugin_name_ + ".cancel_deceleration", rclcpp::ParameterValue(3.2));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".allow_reversing", rclcpp::ParameterValue(false));
   declare_parameter_if_not_declared(
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -86,6 +86,8 @@ ParameterHandler::ParameterHandler(
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".max_angular_accel", rclcpp::ParameterValue(3.2));
   declare_parameter_if_not_declared(
+    node, plugin_name_ + ".cancel_deceleration", rclcpp::ParameterValue(1.0));
+  declare_parameter_if_not_declared(
     node, plugin_name_ + ".allow_reversing", rclcpp::ParameterValue(false));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".max_robot_pose_search_dist",
@@ -151,6 +153,7 @@ ParameterHandler::ParameterHandler(
   node->get_parameter(
     plugin_name_ + ".rotate_to_heading_min_angle", params_.rotate_to_heading_min_angle);
   node->get_parameter(plugin_name_ + ".max_angular_accel", params_.max_angular_accel);
+  node->get_parameter(plugin_name_ + ".cancel_deceleration", params_.cancel_deceleration);
   node->get_parameter(plugin_name_ + ".allow_reversing", params_.allow_reversing);
   node->get_parameter(
     plugin_name_ + ".max_robot_pose_search_dist",
@@ -237,6 +240,8 @@ ParameterHandler::dynamicParametersCallback(
         params_.regulated_linear_scaling_min_speed = parameter.as_double();
       } else if (name == plugin_name_ + ".max_angular_accel") {
         params_.max_angular_accel = parameter.as_double();
+      } else if (name == plugin_name_ + ".cancel_deceleration") {
+        params_.cancel_deceleration = parameter.as_double();
       } else if (name == plugin_name_ + ".rotate_to_heading_min_angle") {
         params_.rotate_to_heading_min_angle = parameter.as_double();
       }
--- a/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
@@ -238,6 +238,23 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
       collision_checker_->costAtPose(pose.pose.position.x, pose.pose.position.y), transformed_plan,
       linear_vel, x_vel_sign);
 
+    if (cancelling_) {
+      const double & dt = control_duration_;
+      linear_vel = speed.linear.x - x_vel_sign * dt * params_->cancel_deceleration;
+
+      if (x_vel_sign > 0) {
+        if (linear_vel <= 0) {
+          linear_vel = 0;
+          finished_cancelling_ = true;
+        }
+      } else {
+        if (linear_vel >= 0) {
+          linear_vel = 0;
+          finished_cancelling_ = true;
+        }
+      }
+    }
+
     // Apply curvature to angular velocity after constraining linear velocity
     angular_vel = linear_vel * regulation_curvature;
   }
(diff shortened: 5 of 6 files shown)


nav2_controller/src/controller_server.cpp
```diff
@@ -451,17 +451,25 @@ void ControllerServer::computeControl()
...
-        publishZeroVelocity();
-        return;
+        if (controllers_[current_controller_]->cancel()) {
+          RCLCPP_INFO(get_logger(), "Cancellation was successful. Stopping the robot.");
+          action_server_->terminate_all();
+          publishZeroVelocity();
+          return;
+        } else {
+          if (!cancelling) {
+            RCLCPP_INFO(get_logger(), "Goal was canceled. Trying to stop the robot.");
```
SteveMacenski, line 469: If you make this a throttle log, then you don't need to have this `if(not var) {var = !var}` logic and put this all in 1 line. Having a regular 1s reminder about what's happening still would be useful I think too

nav2_controller/src/controller_server.cpp
```diff
@@ -451,17 +451,25 @@ void ControllerServer::computeControl()
 
     last_valid_cmd_time_ = now();
     rclcpp::WallRate loop_rate(controller_frequency_);
+    bool cancelling = false;
```
SteveMacenski, line 454: I think having a cancelling timeout would be good so that if a controller is not implemented well, we can still stop it eventually. Default to like 5s but you can increase it for vehicle

SteveMacenski:
I think having the cancel state projected back to the `computeVelocityCommands` as an argument would be important. Right now, I don't know how the controller would know when `cancel` stopped and a new request came in otherwise. So I think `cancelling` should be given to it.

Or possibly even have separate `computeCancelCommands()` API to use when `cancelling = true`. That may be a more ergonomics thing: I imagine that when you have a controller that is handling normal vs cancel state, you're basically having an if/else for 2 very different methods either way. Those would probably render into separate functions, so we could just call separate functions.

Rayman:
@SteveMacenski which controller is the easiest to modify with this new API? I was thinking about the pure pursuit controller but that one doesn't have decelleration limits so I don't know how fast to brake. How is the acceleration determined?

SteveMacenski:
Which ever one you like - I think DWB is the only one with explicit decel/accel parameters right now. MPPI will have it before EOY, but not its current state. RPP is a pure geometric approach, so there's not really any dynamic modeling (right, its just following a carrot). We have some constraints on speed based on some heuristic functions, but that's also not dynamics based, its situational.

SteveMacenski:
@Rayman anything I need to review? Or still working on it :smile:

Rayman:
Yes sorry I'm still working on it.

I've modified the RPP controller to add a cancel_decelleration parameter that shows how to use the new API. Is this in line with the design of the controller? I thought that one was the easiest to modify. Most controllers don't have acceleration limits, only the DWB. But that one is quite complicated with the critics so I didn't know how to modify it to support cancellation.

I've also adressed one of your review comments. Last thing to do is add some unit tests probably.

image

nav2_controller/src/controller_server.cpp
```diff
@@ -459,10 +459,16 @@ void ControllerServer::computeControl()
...
-        return;
+        if (controllers_[current_controller_]->cancel()) {
+          RCLCPP_INFO(get_logger(), "Cancellation was successful. Stopping the robot.");
+          action_server_->terminate_all();
+          publishZeroVelocity();
+          return;
+        } else {
+          RCLCPP_INFO_THROTTLE(
+            get_logger(),
+            *get_clock(), 1000, "Goal was canceled. Trying to stop the robot.");
```
SteveMacenski, line 470:
```suggestion
            *get_clock(), 1000, "Goal was canceled. Requesting controller to stop robot gracefully.");
```

nav2_core/include/nav2_core/controller.hpp
```diff
@@ -116,6 +116,16 @@ class Controller
     const geometry_msgs::msg::Twist & velocity,
     nav2_core::GoalChecker * goal_checker) = 0;
 
+  /**
+   * @brief Cancel the current control action
+   * @return True if the cancellation was successful. If false is returned, computeVelocityCommands
+   * will be called until cancel returns true.
+   */
+  virtual bool cancel()
```
SteveMacenski, line 124: Just make sure to add to the Nav2 tutorial on Controllers in the doc updates

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/parameter_handler.hpp
```diff
@@ -53,6 +53,7 @@ struct Parameters
   double curvature_lookahead_dist;
   bool use_rotate_to_heading;
   double max_angular_accel;
+  double cancel_decelleration;
```
SteveMacenski, line 56: `deceleration`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -238,10 +238,37 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+      }
+
+      RCLCPP_INFO(logger_, "cancelling to %f", linear_vel);
+    }
+
     // Apply curvature to angular velocity after constraining linear velocity
     angular_vel = linear_vel * regulation_curvature;
   }
 
+
```
SteveMacenski, line 271: Remove extra line

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -238,10 +238,37 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
+          finished_cancelling_ = true;
+        }
+      } else {
+        if (linear_vel > 0) {
+          linear_vel = 0;
+          finished_cancelling_ = true;
+        }
+      }
+
+      RCLCPP_INFO(logger_, "cancelling to %f", linear_vel);
```
SteveMacenski, line 264: Remove log

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -445,6 +478,11 @@ void RegulatedPurePursuitController::setSpeedLimit(
   }
 }
 
+void RegulatedPurePursuitController::reset()
+{
+  cancelling_ = false;
```
SteveMacenski, line 483: Reset `finished_cancelling_`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -238,10 +238,37 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
       collision_checker_->costAtPose(pose.pose.position.x, pose.pose.position.y), transformed_plan,
       linear_vel, x_vel_sign);
 
+    if (cancelling_) {
+      const double & dt = control_duration_;
+      auto linear_delta = speed.linear.x - linear_vel;
```
SteveMacenski, line 243: I'm not totally grokking this logic. Why not find the `speed.linear.x - dt * params_->cancel_decelleration` (or `+` if reversing)?

I'm not sure why you need to compute `speed.linear.x - linear_vel` first at all and the later max/min feels like that is always quashed to the `dt  * decel` value

SteveMacenski (changes requested):

SteveMacenski:
Generally looks good to me!

nav2_core/include/nav2_core/controller.hpp
Rayman: https://github.com/ros-planning/navigation.ros.org/pull/535

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Rayman: The original logic I had was taking into account that the RPPC could already break and it would make sure to break at least that hard. For example when slowing down at the end of paths or during cusp points. I

I've implemented you suggestion to simplify the code.

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Rayman: Done!

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Rayman: Done

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
Rayman: Done

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/parameter_handler.hpp
Rayman: Done

Rayman:
I don't understand the CI failure. I didn't edit any `package.xml` so why is it complaining about dependency issues?

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -85,6 +85,8 @@ ParameterHandler::ParameterHandler(
     node, plugin_name_ + ".rotate_to_heading_min_angle", rclcpp::ParameterValue(0.785));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".max_angular_accel", rclcpp::ParameterValue(3.2));
+  declare_parameter_if_not_declared(
+    node, plugin_name_ + ".cancel_deceleration", rclcpp::ParameterValue(1.0));
```
SteveMacenski, line 89:
```suggestion
    node, plugin_name_ + ".cancel_deceleration", rclcpp::ParameterValue(3.2));
```

Also, this param needs to be added to the RPP configuration guide

SteveMacenski (approved):

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
Rayman: I've added this to the documentation PR

Rayman:
I've tested this only in simulation for now. I'll test this on a robot in the next coming weeks but I need some more parts to be ready for that. I'll send this PR also to a collegue to review

nav2_controller/src/controller_server.cpp
```diff
@@ -458,10 +458,17 @@ void ControllerServer::computeControl()
...
+        if (controllers_[current_controller_]->cancel()) {
+          RCLCPP_INFO(get_logger(), "Cancellation was successful. Stopping the robot.");
+          action_server_->terminate_all();
+          publishZeroVelocity();
+          return;
+        } else {
+          RCLCPP_INFO_THROTTLE(
+            get_logger(),
+            *get_clock(), 1000,
+            "Goal was canceled. Requesting controller to stop robot gracefully.");
```
Timple, line 470: Nitpick: this is an implementation detail. "Waiting for the controller to finish cancellation"

Timple (commented):
Looks good to me! 

This now works in simulation for the pure pursuit and our own controller. So for me that's good enough :slightly_smiling_face:. I don't think the physical interfaces of the robot will influence this behavior. But I'll leave that up to @SteveMacenski

nav2_controller/src/controller_server.cpp
Timple: On the other hand, `publishZeroVelocity();` enforces a stop, so this is the only option :smile:. You can leave it.

SteveMacenski: Sure, @Rayman can you make that string update? Then I can merge

SteveMacenski:
Sim is good enough for me - yal know what you're doing. Feel free to implement other controllers if you like :wink:

nav2_controller/src/controller_server.cpp
Rayman: Done

SteveMacenski (approved):

SteveMacenski:
Thanks @Rayman !

doisyg:
Nice work, I haven't tested it but reading on the changes, 2 questions:
1. In the way it is implemented in RPP, I believe it is now a mandatory behavior to use the `cancel_deceleration`. Any way to making it optional, for instance when one wants to rely on the deceleration configured and applied in a velocity smoother node (and without having to manage another decel limit parameter somewhere else?
2. Is it applied when preempting the controller server action ? (in my current use-case I would not want that, but I can think of use cases where it would be desirable)

SteveMacenski:
> I believe it is now a mandatory behavior to use the cancel_deceleration

You could set the deceleration to something insanely high to in effect disable, but this is a good point, it should be an option. We  can make a param pretty easily that'll bypass this if needed. 

> Is it applied when preempting the controller server action

Nope, only on cancel:
 
```
      if (action_server_->is_cancel_requested()) {
        if (controllers_[current_controller_]->cancel()) {
          ...
        } ...
      }
```

doisyg:
All good short term then!

PR #4140, opened 2024-02-23
RPP Dexory tweaks

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Similar to https://github.com/ros-planning/navigation2/pull/3788#issuecomment-1958276641 |

---

This is me finally contributing back after rebasing and testing with latest main the RPP tweaks that we have been using for months in production.
Warning, we only use RPP for linear paths, so caution when generalizing the usage of the added feature.

- New behavior controlled by a new parameter `interpolate_curvature_after_goal` (default false) that interpolates a carrot after the goal in order to maintain a constant curvature lookahead distance. This is to avoid instabilities at the end of the path on the generation of the angular speed. The carrot used for the linear speed computation stays the same.
  - Interpolation is based on the orientation of the vector formed by the last 2 poses of the path. Hence paths of length 1 are rejected when `interpolate_curvature_after_goal` is true. It required to modify a bit the path handler logic to be sure to keep at least 2 poses on the pruned plan.
  - It can be used only when `use_fixed_curvature_lookahead: true`
  - Publish `curvature_lookahead_point` similarly to `lookahead_point` for visualisation

https://github.com/ros-planning/navigation2/assets/15727892/5ade17f1-f0a4-4688-b960-0e9e7dff5e12

- Fix conflict between `use_rotate_to_heading` and `allow_reversing` (added `x_vel_sign` parameter to `shouldRotateToPath` for that). So `use_rotate_to_heading` can now be used backward.

These 2 tweaks are a bit intertwined, hence contributing them together. This may be partially redundant with https://github.com/ros-planning/navigation2/pull/3788 @james-ward 

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/CMakeLists.txt
+++ b/nav2_regulated_pure_pursuit_controller/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(nav2_regulated_pure_pursuit_controller)
 
 find_package(ament_cmake REQUIRED)
+find_package(nav2_amcl REQUIRED)
 find_package(nav2_common REQUIRED)
 find_package(nav2_core REQUIRED)
 find_package(nav2_costmap_2d REQUIRED)
@@ -26,6 +27,7 @@ set(dependencies
   nav2_costmap_2d
   pluginlib
   nav_msgs
+  nav2_amcl
   nav2_util
   nav2_core
   tf2
--- a/nav2_regulated_pure_pursuit_controller/README.md
+++ b/nav2_regulated_pure_pursuit_controller/README.md
@@ -90,6 +90,7 @@ Note: The maximum allowed time to collision is thresholded by the lookahead poin
 | `rotate_to_heading_min_angle` | The difference in the path orientation and the starting robot orientation to trigger a rotate in place, if `use_rotate_to_heading` is enabled. | 
 | `max_angular_accel` | Maximum allowable angular acceleration while rotating to heading, if enabled | 
 | `max_robot_pose_search_dist` | Maximum integrated distance along the path to bound the search for the closest pose to the robot. This is set by default to the maximum costmap extent, so it shouldn't be set manually unless there are loops within the local costmap. | 
+| `interpolate_curvature_after_goal` | Needs use_fixed_curvature_lookahead to be true. Interpolate a carrot after the goal dedicated to the curvate calculation (to avoid oscilaltions at the end of the path) | 
 
 Example fully-described XML with default parameter values:
 
@@ -137,6 +138,7 @@ controller_server:
       rotate_to_heading_min_angle: 0.785
       max_angular_accel: 3.2
       max_robot_pose_search_dist: 10.0
+      interpolate_curvature_after_goal: false
       cost_scaling_dist: 0.3
       cost_scaling_gain: 1.0
       inflation_cost_scaling_factor: 3.0
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -168,7 +168,7 @@ ParameterHandler::ParameterHandler(
   if (!params_.use_fixed_curvature_lookahead && params_.interpolate_curvature_after_goal) {
     RCLCPP_WARN(
       logger_, "For interpolate_curvature_after_goal to be set to true, "
-      "use_fixed_curvature_lookahead should be true, it is currently set to false");
+      "use_fixed_curvature_lookahead should be true, it is currently set to false. Disabling.");
     params_.interpolate_curvature_after_goal = false;
   }
   node->get_parameter(
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -91,6 +91,9 @@ ParameterHandler::ParameterHandler(
     node, plugin_name_ + ".max_robot_pose_search_dist",
     rclcpp::ParameterValue(costmap_size_x / 2.0));
   declare_parameter_if_not_declared(
+    node, plugin_name_ + ".interpolate_curvature_after_goal",
+    rclcpp::ParameterValue(false));
+  declare_parameter_if_not_declared(
     node, plugin_name_ + ".use_collision_detection",
     rclcpp::ParameterValue(true));
 
@@ -160,6 +163,15 @@ ParameterHandler::ParameterHandler(
   }
 
   node->get_parameter(
+    plugin_name_ + ".interpolate_curvature_after_goal",
+    params_.interpolate_curvature_after_goal);
+  if (!params_.use_fixed_curvature_lookahead && params_.interpolate_curvature_after_goal) {
+    RCLCPP_WARN(
+      logger_, "For interpolate_curvature_after_goal to be set to true, "
+      "use_fixed_curvature_lookahead should be true, it is currently set to false");
+    params_.interpolate_curvature_after_goal = false;
+  }
+  node->get_parameter(
     plugin_name_ + ".use_collision_detection",
     params_.use_collision_detection);
 
@@ -168,16 +180,6 @@ ParameterHandler::ParameterHandler(
       logger_, "The value inflation_cost_scaling_factor is incorrectly set, "
       "it should be >0. Disabling cost regulated linear velocity scaling.");
     params_.use_cost_regulated_linear_velocity_scaling = false;
-  }
-
-  /** Possible to drive in reverse direction if and only if
-   "use_rotate_to_heading" parameter is set to false **/
-
-  if (params_.use_rotate_to_heading && params_.allow_reversing) {
-    RCLCPP_WARN(
-      logger_, "Disabling reversing. Both use_rotate_to_heading and allow_reversing "
-      "parameter cannot be set to true. By default setting use_rotate_to_heading true");
-    params_.allow_reversing = false;
   }
 
   dyn_params_handler_ = node->add_on_set_parameters_callback(
@@ -250,12 +252,6 @@ ParameterHandler::dynamicParametersCallback(
       } else if (name == plugin_name_ + ".use_collision_detection") {
         params_.use_collision_detection = parameter.as_bool();
       } else if (name == plugin_name_ + ".use_rotate_to_heading") {
-        if (parameter.as_bool() && params_.allow_reversing) {
-          RCLCPP_WARN(
-            logger_, "Both use_rotate_to_heading and allow_reversing "
-            "parameter cannot be set to true. Rejecting parameter update.");
-          continue;
-        }
         params_.use_rotate_to_heading = parameter.as_bool();
       } else if (name == plugin_name_ + ".allow_reversing") {
         if (params_.use_rotate_to_heading && parameter.as_bool()) {
--- a/nav2_regulated_pure_pursuit_controller/src/path_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/path_handler.cpp
@@ -47,10 +47,15 @@ double PathHandler::getCostmapMaxExtent() const
 
 nav_msgs::msg::Path PathHandler::transformGlobalPlan(
   const geometry_msgs::msg::PoseStamped & pose,
-  double max_robot_pose_search_dist)
+  double max_robot_pose_search_dist,
+  bool reject_unit_path)
 {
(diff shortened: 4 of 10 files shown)


james-ward:
> @james-ward can you look at this and confirm its a suitable alternative? If so, I can fix up a small number of pendantic things I'd like to change here (remove AMCL dependency...) and merge it. Or if there are elements of your PR you think need to still be here, let me know and I can make sure to review them both as a pair

LGTM. This is essentially what I had - the only difference is that I didn't force the path to always contain at least two points, as I wasn't sure if this would break some other behaviour (ie maybe the pruning of already passed points was necessary for correct behaviour of something else.) If that is not the case, this PR is good for me.

I can test on our platforms if the two of you wish, but I can't do that until later in the week.

james-ward:
The only other thing is that you might want to adapt the new test I wrote on my branch for the actual act of projecting off the end of the path?
https://github.com/james-ward/navigation2/commit/e9dd766077bd395a4160011ed787bdd34c306b46

nav2_regulated_pure_pursuit_controller/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(nav2_regulated_pure_pursuit_controller)
 
 find_package(ament_cmake REQUIRED)
+find_package(nav2_amcl REQUIRED)
```
SteveMacenski, line 5: TODO Me: remove AMCL dep and use `angles`

nav2_regulated_pure_pursuit_controller/README.md
```diff
@@ -90,6 +90,7 @@ Note: The maximum allowed time to collision is thresholded by the lookahead poin
 | `rotate_to_heading_min_angle` | The difference in the path orientation and the starting robot orientation to trigger a rotate in place, if `use_rotate_to_heading` is enabled. | 
 | `max_angular_accel` | Maximum allowable angular acceleration while rotating to heading, if enabled | 
 | `max_robot_pose_search_dist` | Maximum integrated distance along the path to bound the search for the closest pose to the robot. This is set by default to the maximum costmap extent, so it shouldn't be set manually unless there are loops within the local costmap. | 
+| `interpolate_curvature_after_goal` | Needs use_fixed_curvature_lookahead to be true. Interpolate a carrot after the goal dedicated to the curvate calculation (to avoid oscilaltions at the end of the path) |
```
SteveMacenski, line 93: Question for @doisyg @james-ward : Do you feel there's any reason to make this a parameterization or always be in play?

james-ward: I think it should always be in play - for two reasons.
First, it preserves the user's intent when they set the minimum carrot distance.
Second, it only engages in the last segment and fixes bad control behaviour - I don't see a downside to having it running.

SteveMacenski: @doisyg if you agree, you can remove the param and make it the always-on behavior

doisyg:
> Great! @doisyg can you add that and rebase on main, that should make CI fully turn over

Yes! @james-ward just the test of that commit https://github.com/james-ward/navigation2/commit/e9dd766077bd395a4160011ed787bdd34c306b46, or all the added tests in your branch ?

nav2_regulated_pure_pursuit_controller/README.md
doisyg: On that one, I would be a bit conservative and keep it as a parameter for a while, then changing it later to default true, and finally then removing it. A bit like with the former `use_interpolation`.
I am only very confident for straight path.

james-ward:
> > Great! @doisyg can you add that and rebase on main, that should make CI fully turn over
> 
> Yes! @james-ward just the test of that commit [james-ward@e9dd766](https://github.com/james-ward/navigation2/commit/e9dd766077bd395a4160011ed787bdd34c306b46), or all the added tests in your branch ?

I think only that commit is relevant. The others were for internal functions that I created.

nav2_regulated_pure_pursuit_controller/README.md
SteveMacenski: @james-ward can you give this a whirl and let us know what you think? He's testing this on straight-line paths in warehouses, so they don't end on curves (they use other algorithms for free space controlling). If so, we can simplify this a bit

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -159,6 +162,15 @@ ParameterHandler::ParameterHandler(
     params_.max_robot_pose_search_dist = std::numeric_limits<double>::max();
   }
 
+  node->get_parameter(
+    plugin_name_ + ".interpolate_curvature_after_goal",
+    params_.interpolate_curvature_after_goal);
+  if (!params_.use_fixed_curvature_lookahead && params_.interpolate_curvature_after_goal) {
+    RCLCPP_WARN(
+      logger_, "For interpolate_curvature_after_goal to be set to true, "
+      "use_fixed_curvature_lookahead should be true, it is currently set to false");
```
SteveMacenski, line 171:
```suggestion
      "use_fixed_curvature_lookahead should be true, it is currently set to false. Disabling.");
```

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -246,10 +259,15 @@ geometry_msgs::msg::TwistStamped RegulatedPurePursuitController::computeVelocity
...
 
 bool RegulatedPurePursuitController::shouldRotateToPath(
-  const geometry_msgs::msg::PoseStamped & carrot_pose, double & angle_to_path)
+  const geometry_msgs::msg::PoseStamped & carrot_pose, double & angle_to_path,
+  double & x_vel_sign)
 {
   // Whether we should rotate robot to rough path heading
   angle_to_path = atan2(carrot_pose.pose.position.y, carrot_pose.pose.position.x);
+  // In case we are reversing
+  if (x_vel_sign < 0) {
```
SteveMacenski, line 268:
```suggestion
  if (x_vel_sign < 0.0) {
```

SteveMacenski:
I took a look, beyond the parameter remark, I think this is good to go. Once merged, I'll remove the AMCL bits myself. 

Without parameter:
- Just need you to do the migration guide entry (add that gif!)

With parameter
- Without parameter instructions and ...
- Param guide entry for the RPP param

doisyg:
Rebased on main and added / adapted your test @james-ward (testing full quadrant and no single pose path as we reject them now).
If all good, will do the documentation asap

nav2_regulated_pure_pursuit_controller/CMakeLists.txt
doisyg: oh yeah forgot about than one and your eyes when you saw that we are still using the good old amcl angles utils

doisyg:
Docs: https://github.com/ros-planning/navigation.ros.org/pull/530

SteveMacenski:
LGTM pending doc fix and if @james-ward responds to https://github.com/ros-planning/navigation2/pull/4140#discussion_r1506302298

SteveMacenski (approved):

james-ward:
> LGTM pending doc fix and if @james-ward responds to [#4140 (comment)](https://github.com/ros-planning/navigation2/pull/4140#discussion_r1506302298)

I'll give it a run this afternoon.

james-ward:
I'm afraid I wasn't able to test this. We've had to run my patch up until now to make our platforms drivable, and the mainline has diverged enough that this wasn't just a simple update, keeping the other plugins and configs the same. I could not get the platform to move at all.

I will have another look on Monday.

SteveMacenski:
Thanks, I really appreciate your time here. Making sure all the stakeholders are happy and not going to come back in 3 weeks because we made an error saves us all frustration :smile:

james-ward:
This PR does have the effect of projecting the carrot, but it is not smooth. There are big jumps in the projected position. So it isn't keeping the look ahead distance constant as is desired.
[carrot_short.webm](https://github.com/ros-planning/navigation2/assets/1538432/ac16d81f-7540-4b32-9bf1-3c53decb71eb)

This is because the current robot position is assumed to be the second last pose in the path, not its actual position. So the carrot will jump ahead by the spacing of the poses in the path. For big spacings this will result in jumpy carrot movement.

doisyg:
Hi @james-ward, fair enough, we have been using paths/plans with small spacing, hence it works smoothly for us.

`auto current_robot_pose_it = transformed_plan.poses.begin();` is maybe not the best name as it is the first pose of the pruned plan, not the real robot pose. It is used here only to compute the curvature carrot distance to interpolate after the goal. 
We don't want to use the real robot pose for that in order to have a stable carrot, but the closest path pose, which should be the first pose of the pruned plan. However, as you pointed, the accuracy of using the first pose of the pruned plan is limited by the plan pose density.

An easy workaround would be to use a denser plan. A full solution will be to interpolate between the pose of the plan.  Maybe as a separate PR on top of this one ? WDYT ?

james-ward:
The robot position is (0,0) because the `PathHandler` keeps translating and pruning the path so that it is relative to the robot.

So before the last segment, calling `circleSegmentIntersection` gets the point on the current segment that is exactly the look ahead distance away from the current robot position.

But this PR introduces different behaviour in the last segment. It finds the length of the last segment, subtracts it from the look ahead distance and then projects the carrot past the segment end by the remaining amount. For a robot at/very near the goal pose, the actual look ahead will end up being the desired lookahead minus the last segment length.

I think that because you force the controller to not drop the second last pose in the path even when the robot has passed it, you could just call the `circleSegmentIntersection` method on the last segment and it would work. (I think the maths doesn't actually distinguish between finite segments and infinite lines). But failing that, do what I did in my proposed solution and project the segment an additional distance equal to the look ahead and then do the intersection. This would ensure that the extended segment is longer than the look ahead and an intersection with the circle is guaranteed.

james-ward:
@doisyg I have made the changes necessary to keep the carrot at the correct look ahead distance. It clearly wasn't correct before as the tests were looking for distances that were different from the look ahead distance.

My commit is here:
https://github.com/james-ward/navigation2/commit/1a8b5b300998c0370c3fb398cf5a1413545ae91a

SteveMacenski:
Do we suppose that closes the loop on this feature?

james-ward:
> Do we suppose that closes the loop on this feature? I'm defaulting to whatever you both come to an agreement on for this one - I trust your opinions 😄

It does from my point of view. I have been running that latest patch since last week and it is good for us.

SteveMacenski:
@doisyg I know you're in Florida, so take your time, but let me know and push the update if appropriate! :smile:

doisyg:
Thanks a lot @james-ward @SteveMacenski for you patience, and sorry that my iterations are a bit slow.
Pushed your commit, fixed lint and tested in simulation, looks good to go.

doisyg:
and rebased on `main`

SteveMacenski:
@james-ward we all happy with this now? Can you give it one last whirl around? Then we can merge!

james-ward:
> @james-ward we all happy with this now? Can you give it one last whirl around? Then we can merge!

I'll try it this morning. Stand by...

james-ward:
Have confirmed it is good on our platform.

Send it!!

nav2_regulated_pure_pursuit_controller/src/path_handler.cpp
```diff
@@ -47,12 +47,17 @@ double PathHandler::getCostmapMaxExtent() const
...
   const geometry_msgs::msg::PoseStamped & pose,
-  double max_robot_pose_search_dist)
+  double max_robot_pose_search_dist,
+  bool reject_unit_path)
 {
   if (global_plan_.poses.empty()) {
     throw nav2_core::InvalidPath("Received plan with zero length");
   }
 
+  if (reject_unit_path && global_plan_.poses.size() == 1) {
```
SteveMacenski, line 57: when is `reject_unit_path` ever true?

SteveMacenski (commented):
@doisyg I don't see when `reject_unit_path` is ever used. Can this be removed?

Otherwise LGTM and I can come through afterwards on replacing the AMCL dependency

nav2_regulated_pure_pursuit_controller/src/path_handler.cpp
doisyg: When `params_->interpolate_curvature_after_goal` is `true`
https://github.com/doisyg/navigation2/blob/994351e100c50f23c3886cf2375bd342e0fbad9f/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L179-L181

Issue #4141, opened 2024-02-23
SmacHybrid planner can crash to segmentation fault


jonipol:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble (latest ros:humble container)
- Version or commit hash:
  - a3cdbbf7d0a9be9ec50f876ab365a45eff2214d9 + minimal changes to make the rolling version of nav2 to work on Humble (no changes to planner(s) code)
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
This part I am still uncertain. The issue keep happening seemingly randomly during navigation. Have not been able to pinpoint the exact reason/case as I do not yet fully understand what is going on in the SmacHybridPlanner. Sometimes the segmentation fault happens quite soon after starting navigating. Sometimes it takes many many navigations. (Had my simulation giving consecutive goals for 95 minutes before I got the latest seg fault for the log to include in this issue.)

#### Expected behavior
Planner keeps running :grin: 

#### Actual behavior
Planner server crashes to a segmentation fault.
<details>
<summary>Log</summary>

//     Logs I have added
[planner_server-2] D: 1.22173 is smaller than sqrt(2)
[planner_server-2] From x: 603.254 From y: 180.122 From yaw: 5.67232
[planner_server-2] To x: 604 To y: 180 To yaw: 0
//     My logs end
[planner_server-2] Stack trace (most recent call last) in thread 1430:
[planner_server-2] #31   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7ffaaf4aaee7, in 
[...]

PR #4147, opened 2024-02-26
Prevent a possible segmentation fault (#4141)

jonipol:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4141|
| Robotic platform tested on | Gazebo simulation of a Robauta robot |

---

- Prevent a possible segmentation fault in analytic_expansion at smac_planner

Defensive approach as Steve recommended in his comment on the issue. Added check if the iterator == [ vector.end(), as dereferencing ](https://en.cppreference.com/w/cpp/container/vector/end) that can lead into undefined behavior (appeared as segmentation fault in my case). Also added an early return for the case in which the vector would end up being empty, thus leading into the check previously mentioned.

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/analytic_expansion.cpp
+++ b/nav2_smac_planner/src/analytic_expansion.cpp
@@ -189,6 +189,10 @@ typename AnalyticExpansion<NodeT>::AnalyticExpansionNodes AnalyticExpansion<Node
 
   // A move of sqrt(2) is guaranteed to be in a new cell
   static const float sqrt_2 = std::sqrt(2.0f);
+  if (d < sqrt_2) {
+    return AnalyticExpansionNodes();
+  }
+
   unsigned int num_intervals = static_cast<unsigned int>(std::floor(d / sqrt_2));
 
   AnalyticExpansionNodes possible_nodes;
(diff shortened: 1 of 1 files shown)


nav2_smac_planner/src/analytic_expansion.cpp
```diff
@@ -189,6 +189,10 @@ typename AnalyticExpansion<NodeT>::AnalyticExpansionNodes AnalyticExpansion<Node
 
   // A move of sqrt(2) is guaranteed to be in a new cell
   static const float sqrt_2 = std::sqrt(2.0f);
+  if (d < sqrt_2) {
```
SteveMacenski, line 192: Add it to the `if` statement above 

```
if (d > _search_info.analytic_expansion_max_length || d < sqrt_2) {
```

SteveMacenski: ^ that would require moving the sqrt_2 definition above as well

SteveMacenski (changes requested):
LGTM, just a small detail for branching

nav2_smac_planner/src/analytic_expansion.cpp
jonipol: Thanks for feedback!

SteveMacenski:
Mergify backporting failed, will need to be done manually

Issue #4139, opened 2024-02-23
Stop planner if the goal is cancelled


bektaskemal:
#### Feature description
The current implementation of `PlannerServer` invokes the `getPlan` function in a blocking manner. This means that even if a goal is cancelled, the planner computation continues until completion, leading to potential delays, especially with slower planners like Hybrid A*. 

#### Implementation considerations
Potential implementations could be either running planner in an async way and keep checking for `isCancelRequested` or having a cancel service. Both requires `GlobalPlanner` to have a cancel interface.

SteveMacenski:
The planners generally shouldn't be all that long running of processes (Smac Hybrid-A* might be ~400ms in semi-normal large scale use-cases, but I suppose depending on parameterization it could increase more) where we felt that this might be an issue. But its true that especially if you're BOY-planner, then that planner could be, 5-10-20-N seconds long and we might want to kill that.

We do that right now in the Smac Planner and others with planning timeouts. The planner action API contains a timeout that if we exceed, the planner exits. Its different than a cancelation for sure, but helps a portion of this issue to make sure we don't plan indefinitely. 

I'd be very open to an architectural update that would allow for the in-run preemption of planning algorithms based on cancellations.

> Potential implementations could be either running planner in an async way and keep checking for isCancelRequested

That makes sense, basically pass in a function to the planning plugin's main request function who is a lambda that returns the bool value of the action server's is cancel requested API. We can check this on looping over the search or to other internal libraries that can take an external exit condition. I took a quick glance at a couple of the planners and I see where that could be used:

- https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/a_star.cpp#L287-L295 either in the while condition or the if statement below it. Worth benchmarking performance changes in both locations to gauge it since perhaps running it every iteration in the `while` would be overkill. This would cover all Smac Planners all at once
- Similarly in Theta* https://github.com/ros-planning/navigation2/blob/main/nav2_theta_star_planner/src/theta_star.cpp#L60

NavFn is a little trickier, but should also be possible in the same concept. This is something you'd be willing to take the lead on implementing and opening a PR? This seems like a good contribution for sure!

PR #4148, opened 2024-02-26
Stop planner if the goal is cancelled

bektaskemal:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [4139](https://github.com/ros-planning/navigation2/issues/4139) |
| Robotic platform tested on | Gazebo TB3 simulation |

---

* Stops the planner execution if the goal is cancelled

---

#### For Maintainers: 

### DIFF
--- a/nav2_navfn_planner/src/navfn.cpp
+++ b/nav2_navfn_planner/src/navfn.cpp
@@ -44,6 +44,7 @@
 #include "nav2_navfn_planner/navfn.hpp"
 
 #include <algorithm>
+#include "nav2_core/planner_exceptions.hpp"
 #include "rclcpp/rclcpp.hpp"
 
 namespace nav2_navfn_planner
@@ -293,12 +294,12 @@ NavFn::setCostmap(const COSTTYPE * cmap, bool isROS, bool allow_unknown)
 }
 
 bool
-NavFn::calcNavFnDijkstra(bool atStart)
+NavFn::calcNavFnDijkstra(std::function<bool()> cancelChecker, bool atStart)
 {
   setupNavFn(true);
 
   // calculate the nav fn and path
-  return propNavFnDijkstra(std::max(nx * ny / 20, nx + ny), atStart);
+  return propNavFnDijkstra(std::max(nx * ny / 20, nx + ny), cancelChecker, atStart);
 }
 
 
@@ -307,12 +308,12 @@ NavFn::calcNavFnDijkstra(bool atStart)
 //
 
 bool
-NavFn::calcNavFnAstar()
+NavFn::calcNavFnAstar(std::function<bool()> cancelChecker)
 {
   setupNavFn(true);
 
   // calculate the nav fn and path
-  return propNavFnAstar(std::max(nx * ny / 20, nx + ny));
+  return propNavFnAstar(std::max(nx * ny / 20, nx + ny), cancelChecker);
 }
 
 //
@@ -571,7 +572,7 @@ NavFn::updateCellAstar(int n)
 //
 
 bool
-NavFn::propNavFnDijkstra(int cycles, bool atStart)
+NavFn::propNavFnDijkstra(int cycles, std::function<bool()> cancelChecker, bool atStart)
 {
   int nwv = 0;  // max priority block size
   int nc = 0;  // number of cells put into priority blocks
@@ -581,6 +582,10 @@ NavFn::propNavFnDijkstra(int cycles, bool atStart)
   int startCell = start[1] * nx + start[0];
 
   for (; cycle < cycles; cycle++) {  // go for this many cycles, unless interrupted
+    if (cycle % cancel_check_interval == 0 && cancelChecker()) {
+      throw nav2_core::PlannerCancelled("Planner was cancelled");
+    }
+
     if (curPe == 0 && nextPe == 0) {  // priority blocks empty
       break;
     }
@@ -652,7 +657,7 @@ NavFn::propNavFnDijkstra(int cycles, bool atStart)
 //
 
 bool
-NavFn::propNavFnAstar(int cycles)
+NavFn::propNavFnAstar(int cycles, std::function<bool()> cancelChecker)
 {
   int nwv = 0;  // max priority block size
   int nc = 0;  // number of cells put into priority blocks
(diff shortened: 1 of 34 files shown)


nav2_smac_planner/src/a_star.cpp
```diff
@@ -294,6 +295,11 @@ bool AStarAlgorithm<NodeT>::createPath(
       }
     }
 
+    // Check for cancel every Nth iteration
+    if (iterations % _cancel_check_interval == 0 && cancel_checker()) {
```
SteveMacenski, line 299: Is 500 vs 5000 a big enough difference to not have it in the same timing interval branch (and possibly rename to be more generic)?

nav2_navfn_planner/src/navfn.cpp
```diff
@@ -667,6 +672,10 @@ NavFn::propNavFnAstar(int cycles)
 
   // do main cycle
   for (; cycle < cycles; cycle++) {  // go for this many cycles, unless interrupted
+    if (cycle % cancel_check_interval == 0 && cancelChecker()) {
+      throw nav2_core::PlannerCancelled("Planner was cancelled");
```
SteveMacenski, line 676: I don't see where this exception is caught specially to handle this condition now that its created

SteveMacenski: It should be in the catch list and if caught, do what we do in the cancel function: https://github.com/ros-planning/navigation2/blob/9e0e1f6b785d89cebd876dfbe1fbed7bea8c644f/nav2_planner/src/planner_server.cpp#L298

SteveMacenski (changes requested):
Overall, LGTM!

nav2_navfn_planner/src/navfn.cpp
```diff
@@ -667,6 +672,10 @@ NavFn::propNavFnAstar(int cycles)
 
   // do main cycle
   for (; cycle < cycles; cycle++) {  // go for this many cycles, unless interrupted
+    if (cycle % cancel_check_interval == 0 && cancelChecker()) {
```
SteveMacenski, line 675: Where is cancel_check_interval set and parameterized?

nav2_theta_star_planner/src/theta_star.cpp
```diff
@@ -60,6 +61,11 @@ bool ThetaStar::generatePath(std::vector<coordsW> & raw_path)
   while (!queue_.empty()) {
     nodes_opened++;
 
+    if (nodes_opened % cancel_check_interval_ && cancel_checker()) {
```
SteveMacenski, line 64:
```suggestion
    if (nodes_opened % cancel_check_interval_ == 0 && cancel_checker()) {
```

nav2_navfn_planner/src/navfn.cpp
bektaskemal: I parametrized it now for other planners but it wasn't so clear how nicely do it for NavFn, so I made it constant for now.

nav2_navfn_planner/src/navfn.cpp
bektaskemal: Done.

nav2_smac_planner/src/a_star.cpp
bektaskemal: I am not very sure about numbers so I am open to suggestions for default values, but I made it a parameter now so seperate branch is needed.

nav2_navfn_planner/src/navfn.cpp
SteveMacenski: Where is that constant set? As far as I see, its just instantiated (so 0) but never set

nav2_smac_planner/src/a_star.cpp
SteveMacenski: Lets not then, use the same parameter branch to check for all potential preemption conditions like cancel / timeout

nav2_smac_planner/src/smac_planner_lattice.cpp
```diff
@@ -504,15 +509,18 @@ SmacPlannerLattice::dynamicParametersCallback(std::vector<rclcpp::Parameter> par
...
-      }
-    } else if (name == _name + ".max_on_approach_iterations") {
-      reinit_a_star = true;
-      _max_on_approach_iterations = parameter.as_int();
-      if (_max_on_approach_iterations <= 0) {
-        RCLCPP_INFO(
-          _logger, "On approach iteration selected as <= 0, "
-          "disabling tolerance and on approach iterations.");
-        _max_on_approach_iterations = std::numeric_limits<int>::max();
+      } else if (name == _name + ".max_on_approach_iterations") {
```
SteveMacenski, line 512: Appears to be misformatting, I'm sure CI is complaining about linting

SteveMacenski (changes requested):
Overall, looks very good thank you for this great contribution!

A couple of small things to finish it up
- Documentation updates for new parameter for all appropriate algorithms. Add to Readmes (if they have a param list) & navigation.ros.org's configuration guide entries for the algorithms to add their params
- Migration guide entry on navigation.ros.org on Planner plugin API + new param and how to use 
- I'm noticing alot of code cov warnings on lack of test coverage on this PRs diffs. We need things to be 90% tested, it seems like the dynamic parameter changes weren't added to their tests to exercise those lines, nor is cancel actually called on any of the algorithms as part of test to make sure that it does actually return early -- thus also that exception itself / planner server handling isn't exercised either

nav2_smac_planner/src/smac_planner_lattice.cpp
bektaskemal: I believe there was a mistake here and this line wasn't inside PARAMETER_INTEGER branch, so I fixed it.

SteveMacenski (approved):
LGTM, just one set of important lines that aren't covered, the actual catching of the exception and checking that the correct state happens in the planner server. 

We have some tests that trigger the others in https://github.com/ros-planning/navigation2/tree/main/nav2_system_tests/src/error_codes. I think you can just add the new plugin for cancel and check that its terminated correctly as a cancel state

Sorry to be a stickler on this one - these are just important systems to make sure are tested and have near full coverage in case of regression.

bektaskemal:
> We have some tests that trigger the others in https://github.com/ros-planning/navigation2/tree/main/nav2_system_tests/src/error_codes. I think you can just add the new plugin for cancel and check that its terminated correctly as a cancel state

If planner returns PlannerCancelled without a cancel request, the result status was ABORTED since we call terminate_all. So I tested it with an actual cancel request. 
https://github.com/ros-planning/navigation2/pull/4148/files#diff-7d5b834e9aba5d3d110b02c975dc4c997b854ce7e92bdeb03d68108fa3071df1R120-R124

SteveMacenski (approved):

SteveMacenski:
Great - a couple of tests failed - I don't think they're related to you - but rerunning CI just to check. I expect to be able to merge this before the end of today

PR #4171, opened 2024-03-08
nav2_controller: add loop rate log

ARK3r:
Following the format [here](https://github.com/ros-planning/navigation2/blob/c9d9b5c698d1f5aff060806481d18ba7e9b18cb2/nav2_planner/src/planner_server.cpp#L444) added loop rate logging to `nav2_controller/controller_server.cpp`.


SteveMacenski (approved):
No beef with the content of this PR, good contribution! Please fix the build and linting issues though before it can be merged.

nav2_controller/src/controller_server.cpp
```diff
@@ -479,10 +481,13 @@ void ControllerServer::computeControl()
...
       }
 
+      auto cycle_duration = this->now() - start_time;
+
       if (!loop_rate.sleep()) {
         RCLCPP_WARN(
-          get_logger(), "Control loop missed its desired rate of %.4fHz",
-          controller_frequency_);
+          get_logger(),
+          "Control loop missed its desired rate of %.4f Hz.  Current loop rate is %.4f Hz",
```
SteveMacenski, line 489:
```suggestion
          "Control loop missed its desired rate of %.4f Hz. Current loop rate is %.4f Hz.",
```

nav2_controller/src/controller_server.cpp
```diff
@@ -479,10 +481,13 @@ void ControllerServer::computeControl()
         break;
       }
 
+      auto cycle_duration = this->now() - start_time;
+
```
SteveMacenski, line 485:
```suggestion
```

nav2_controller/src/controller_server.cpp
```diff
@@ -419,6 +419,8 @@ void ControllerServer::computeControl()
 {
   std::lock_guard<std::mutex> lock(dynamic_params_lock_);
 
+  auto start_time = this->now();
+
```
SteveMacenski, line 423:
```suggestion
```

SteveMacenski:
@ARK3r this was reverted in https://github.com/ros-planning/navigation2/pull/4210 this fundamentally did not work. You set the start time to the start of the control cycle rather than at that current computational iteration.

In the future **please test** your PRs before submitting them.  This was a really silly mistake that could have been immediately caught if you ever attempted to use that log.

ARK3r:
Hi @SteveMacenski, this was a rookie mistake and at the time I simply wasn't thinking it through. I'll be sure to not let this happen again.

SteveMacenski:
If you want to resubmit another PR to add this feature (that works :wink: ) We're still open to that feature!

PR #4174, opened 2024-03-08
45% performance improvement in MPPI

SteveMacenski:
From 12.7ms on average to 6.80ms on average for 10000 cycles - with full footprint checking enabled

---

Validate final run-time and system performance of changes  to make sure still high functioning with changes
- [ ] Cost - OK avoiding obstacles and dynamic obstacles (@BriceRenaudeau ?)
- [x] Path Align - OK aligning effectively with the path and alongside other critics
- [x] Path Angle - OK checking only end in a broad set of situations?
- [x] High level Behavior - OK (retune?)

---

- [x] Increase default trajectory_point_step to 4 for path align
- [x] Document + default for trajectory_point_step for cost critic
- [x] Move out legacy Path Align
- [x] docs, lint https://github.com/ros-planning/navigation.ros.org/pull/533
- [x] Tests update


nav2_mppi_controller/src/critics/path_align_critic.cpp
```diff
@@ -111,13 +113,12 @@ void PathAlignCritic::score(CriticData & data)
...
       // let the obstacle critic take over in this region due to dynamic obstacles
-      if ((*data.path_pts_valid)[path_pt]) {
+      if (path_pts_valid[path_pt]) {
         dx = P_x(path_pt) - Tx;
         dy = P_y(path_pt) - Ty;
         num_samples += 1.0f;
         if (use_path_orientations_) {
-          const auto T_yaw = xt::view(data.trajectories.yaws, t, xt::all());
-          dyaw = angles::shortest_angular_distance(P_yaw(path_pt), T_yaw(p));
+          dyaw = angles::shortest_angular_distance(P_yaw(path_pt), data.trajectories.yaws(t, p));
```
SteveMacenski, line 121: FIX: need T_yaw for skipping by the trajectory point step

SteveMacenski:
TODO apply to all critics

```
  if (power_ > 1u) {
    data.costs += xt::pow(std::move(cost) * weight_, power_);
  } else {
    data.costs += std::move(cost) * weight_;
  }
```

Makes 10x faster if we don't need to use power

nav2_costmap_2d/src/footprint_collision_checker.cpp
```diff
@@ -127,13 +127,14 @@ void FootprintCollisionChecker<CostmapT>::setCostmap(CostmapT costmap)
...
-  double x, double y, double theta, const Footprint footprint)
+  double x, double y, double theta, const Footprint & footprint)
 {
   double cos_th = cos(theta);
   double sin_th = sin(theta);
   Footprint oriented_footprint;
+  oriented_footprint.reserve(footprint.size());
+  geometry_msgs::msg::Point new_pt;
   for (unsigned int i = 0; i < footprint.size(); ++i) {
-    geometry_msgs::msg::Point new_pt;
```
Timple, line 136: The `.reserve` makes sense. But I was under the impression moving `new_pt` out of the loop only introduced potential scoping issues. The compiler would optimize this away.

Happy to learn though.

nav2_mppi_controller/src/critics/path_follow_critic.cpp
```diff
@@ -63,11 +63,11 @@ void PathFollowCritic::score(CriticData & data)
...
-    xt::pow(last_x - path_x, 2) +
-    xt::pow(last_y - path_y, 2));
-
-  data.costs += xt::pow(weight_ * std::move(dists), power_);
+  if (power_ > 1u) {
+    data.costs += xt::pow(weight_ * std::move(xt::hypot(last_x - path_x, last_y - path_y)), power_);
+  } else {
+    data.costs += weight_ * std::move(xt::hypot(last_x - path_x, last_y - path_y));
+  }
 }
```
alanxuefei, line 71: Performance is improved through inline functions, reduced xtensor usage, object declaration optimization, and strategic use of vector.reserve().

Just a minor suggestion: std::move might not be necessary here. The code may be refactored as follows:
```
auto distance = xt::hypot(last_x - path_x, last_y - path_y);
auto weightedDistance = weight_ * distance;
if (power_ > 1u) {
    data.costs += xt::pow(weightedDistance, power_);
} else {
    data.costs += weightedDistance;
}

```

nav2_costmap_2d/src/footprint_collision_checker.cpp
SteveMacenski: That change did not have measurable changes, I previously had everything inlined into the `oriented_footprint` with `emplace_back` looking to avoid a copy. That didn't to anything so I aired on the side of reverting changes if they didn't help. This could be reverted back as well

nav2_mppi_controller/src/critics/path_follow_critic.cpp
SteveMacenski: The `std::move()` tends to increase stability of the runtime across the board. It removes some jitter in the run-time which I think is a nice-to-have where possible. This critic is not overly expensive, so the jitter of just this critic is not massive, but its a general pattern everywhere where it does make a positive impact. 

Your comment on cleaning up the branch for dealing with the `pow` is good - it was my intention to go back and clean that up to be less repetitive but I forgot to actually do it. I'll make that change assuming it doesn't negatively impact performance.

nav2_costmap_2d/src/footprint_collision_checker.cpp
SteveMacenski: Honestly, none of the changes here really made much of a difference. The collision checker is still a nearly totally unexplored area of improvements. The cost critic still takes between 0.4-0.7 ms (which is 10-100x more than most others) of which that time is primarily spent in the collision checking process. 

I intended to spend more time to really dig into that, but I found other changes dropped the ~2ms from before down to the 0.4-0.7 range which I felt was "good enough". I think there's more room to squeeze out performance with a specialized MPPI collision checker object with precomputation / caching / operating on `float`s.

nav2_mppi_controller/src/critics/path_follow_critic.cpp
SteveMacenski: Some of the biggest improvements weren't those. 

- Removing any and all `double` into `float`s had a huge impact 
- Removing the `pow` when not needed was another big improvement
- Rewriting the sin/cos for the trajectory roll outs saved 1ms
- Rewriting the pass through motion model saved 0.6ms
- Rewriting the Cost critic and inlining functions (among some other nickel and diming) saved 1.2ms
- Some util function rewrites saved in total a bit more than 0.5ms (I don't have exact numbers on this one, I worked on different util functions in different contexts)

jjd9 (commented):
Hello, I did not find any glaring issues, but I found some small things I thought were worth pointing out. Thanks!

nav2_mppi_controller/src/critics/path_follow_critic.cpp
alanxuefei: Thank you for sharing the tips on improving performance; It's really useful. It's enlightening to learn that std::move can remove jitter during runtime.

std::move casts an l-value to an r-value.

 I can understand the use of std::move in other places.
However, the following code xt::hypot(last_x - path_x, last_y - path_y) already returns an r-value. Do we need to use std::move to cast it again?

`data.costs += weight_ * std::move(xt::hypot(last_x - path_x, last_y - path_y));`

nav2_mppi_controller/include/nav2_mppi_controller/tools/utils.hpp
```diff
@@ -49,6 +49,9 @@
 #include "builtin_interfaces/msg/time.hpp"
 #include "nav2_mppi_controller/critic_data.hpp"
 
+#define M_PIF 3.141592653589793238462643383279502884e+00F
+#define M_PIF_2 1.5707963267948966e+00F
```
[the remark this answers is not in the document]
SteveMacenski: Yes, they are `float` allowing for fully `float` operations

aosmw:
Hi Steve, 

Do you know about this "pitfall"?  https://xtensor.readthedocs.io/en/latest/pitfall.html#random-numbers-not-consistent
Using a random number function from xtensor actually returns a lazy generator. That means, accessing the same element of a random number generator does not give the same random number if called twice.

Is there a speed advantage if the xt::random::randn operations in NoiseGenerator::generateNoisedControls are wrapped in xt::eval?

regards
Mikew

PR #4181, opened 2024-03-14
Nacho/cleanup get transform util

nachovizzo:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Simulatin |

---

The following pattern is widely used in ROS applications, and (far from ideal) I'd like to extend the `nav2_utils::getTransform` to also spit `TransformStamped` messages. I know, these util functions should be in the tf library but here seemed like a lower-hanging fruit to bite:
```cpp
try {
  // Obtaining the transform to get data from source to target frame
  return tf_buffer->lookupTransform(
    target_frame_id, source_frame_id,
    tf2::TimePointZero, transform_timeout);
} catch (tf2::TransformException & e) {
  RCLCPP_ERROR(
    rclcpp::get_logger("getTransform"),
    "Failed to get \"%s\"->\"%s\" frame transform: %s",
    source_frame_id.c_str(), target_frame_id.c_str(), e.what());
  return std::nullopt;
}
```
Some comments:

* Open discussion PR. Not necessary to pretend that 100% gets merged
* This PR introduces a **hard dependency in C++17** on the `nav2_util` pkg. I'm happy to express this in the build system but I'm not sure if this is a common practice in nav2. Do you think this might bring a problem
* I tried to be specific commit-wise. So the changes can be reviewed in the commits and not just the entire diff
* I renamed some bits to follow more the traditional names from `tf2` to avoid confusing users
* I introduced a BREAKIN change by swapping the `soruce` and `frame` id from the `getTransform` function. I'm not sure why it was "inverted" (compared to tf2) in the first place. But I found it very confusing to read this:
```cpp
// tf2
geometry_msgs::msg::TransformStamped
lookupTransform(const std::string & target_frame, const std::string & source_frame, ...)

// nav2
bool getTransform(const std::string & source_frame_id,   const std::string & target_frame_id, ...)

/// Function call
lookupTransform(target, source)
```
* I know it's not common in ROS, but I'm trying to introduce a silly pattern that I think it's much better than returning booleans on functions, and forcing the user to provide an output argument:
Before:
```cpp
OutputType out_variable;
bool res = FunctionCall(/*const*/ arg1,/*const*/ arg2, ..., /* non-const */ out_variable);
if(res) {
 // use out_variable
} else {
return;
}
```

For a more modern and easier to read + safer approach:
```cpp
auto out_variable = FunctionCall(/*const*/ arg1,/*const*/ arg2);
if(out_variable.has_value()) {
 // use out_variable
} else {
return;
}
```

This improves the code (on my opinion) from:
```cpp
 tf2::Transform tf_transform;
  if (base_shift_correction_) {
    // Obtaining the transform to get data from source frame and time where it was received
    // to the base frame and current time
    if (
      !nav2_util::getTransform(
        data_->header.frame_id, data_->header.stamp,
        base_frame_id_, curr_time, global_frame_id_,
        transform_tolerance_, tf_buffer_, tf_transform))
    {
      return false;
    }
  } else {
    // Obtaining the transform to get data from source frame to base frame without time shift
    // considered. Less accurate but much more faster option not dependent on state estimation
    // frames.
    if (
      !nav2_util::getTransform(
        data_->header.frame_id, base_frame_id_,
        transform_tolerance_, tf_buffer_, tf_transform))
    {
      return false;
    }
  }
```
, to:
```cpp
  const auto tf_transform = [&]() -> std::optional<tf2::Transform> {
      // Obtaining the transform to get data from source frame and time where it was received to the
      // base frame and current time
      if (base_shift_correction_) {
        return nav2_util::getTransform(
          base_frame_id_, curr_time, data_->header.frame_id,
          data_->header.stamp, global_frame_id_, transform_timeout_,
          tf_buffer_);
      }

      // Obtaining the transform to get data from source frame to base frame without time shift
      // considered. Less accurate but much more faster option not dependent on state estimation
      // frames.
      return nav2_util::getTransform(
        base_frame_id_, data_->header.frame_id, transform_timeout_,
        tf_buffer_);
    }();

  if (!tf_transform.has_value()) {return false;}

```

- Make sure we aren't using the "bare" pattern without calling `nav2_utils::getTransform`

#### For Maintainers: 


SteveMacenski:
> This PR introduces a hard dependency in C++17 on the nav2_util pkg. I'm happy to express this in the build system but I'm not sure if this is a common practice in nav2. Do you think this might bring a problem

This is fine!

nav2_collision_monitor/src/pointcloud.cpp
```diff
@@ -78,30 +79,25 @@ bool PointCloud::getData(
...
-    // frames.
-    if (
-      !nav2_util::getTransform(
-        data_->header.frame_id, base_frame_id_,
-        transform_tolerance_, tf_buffer_, tf_transform))
-    {
-      return false;
-    }
-  }
+  const auto tf_transform = [&]() -> std::optional<tf2::Transform> {
```
SteveMacenski, line 82: I don't see the value in this pattern

SteveMacenski (changes requested):
Heed the linter

nav2_collision_monitor/include/nav2_collision_monitor/circle.hpp
```diff
@@ -38,14 +38,14 @@ class Circle : public Polygon
...
-   * @param transform_tolerance Transform tolerance
+   * @param transform_timeout How long to block before failing
    */
   Circle(
     const nav2_util::LifecycleNode::WeakPtr & node,
     const std::string & polygon_name,
     const std::shared_ptr<tf2_ros::Buffer> tf_buffer,
     const std::string & base_frame_id,
-    const tf2::Duration & transform_tolerance);
+    const tf2::Duration & transform_timeout);
```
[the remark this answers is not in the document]
nachovizzo: I'm happy to avoid this change. Sadly I was not there to complain earlier :P it's just the name, so it doesn't bother me that much.

Should I revert this bit?

nav2_collision_monitor/src/pointcloud.cpp
nachovizzo: The only real benefit is that *I* can read it better. Performance wise it's kind of the same. The problem of pushing this into the util, is that it will always spit an optional value, so the caller always needs to check whether there is a meaningful value there or not

nav2_collision_monitor/include/nav2_collision_monitor/circle.hpp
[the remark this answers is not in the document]
nachovizzo: Ok, I already reverted it

nav2_collision_monitor/src/pointcloud.cpp
SteveMacenski: I think I'm correct in saying that the PC2/Range/Scan lines are identical. That attribute by itself makes me think that if we're making change, it should be abstracted into a function in the collision monitor's `Source` for them all to share use of. Especially considering `Source` already has the base/global frame, TF buffer, and timeout parameter locally so it would only need to pass in the current stamp/frame of interest. 

At that point, I might argue that `bool myMethodName(args..., tf2::Transform & output_transform)` would be stylistically more Nav2, but I'm not that picky. If you want to use `optional` instead, power to you, but that function should probably be abstracted if its just the same lines repeated 3x as a lambda.

nav2_util/src/robot_utils.cpp
```diff
@@ -135,12 +121,60 @@ bool getTransform(
...
-    return false;
+    return std::nullopt;
   }
+}
 
-  // Convert TransformStamped to TF2 transform
-  tf2::fromMsg(transform.transform, tf2_transform);
-  return true;
+template<>
+std::optional<tf2::Transform> getTransform(
```
SteveMacenski, line 129: I would not - though - support this. In my opinion, consistency is important in large projects so folks understand what they're getting. If you're changing any of these to use `std::optional`, you must change them all and update them globally across the project

SteveMacenski (changes requested):

nav2_util/src/robot_utils.cpp
nachovizzo: Steve, would you mind clarifying the meaning of `this` (not the pointer :stuck_out_tongue: ) and "all`?

I do agree that constituency is important, but one needs to start somewhere anyway. It's hard to draw the line. If you are mentioning replacing all functions that return booleans instead of optional, then I'd agree that's the way to go, but that would be an immense PR and I didn't go that way since I first wanted to openly discuss what would be the preferred option.

Once I have directions from your side I'm happy to go full in, but I want to avoid spending hours changing the entire API to realize later that was a waste of time ;)

nav2_collision_monitor/src/pointcloud.cpp
nachovizzo: I also spotted this but went more conservative and tried not to make that many changes *besides renaming*. If you would like me to shrink this code, I'm happy to!

SteveMacenski: Please :-)

nav2_util/src/robot_utils.cpp
SteveMacenski: This meaning to use optionals for the output of `getTransform` in `nav2_util`, given other methods in `robot_utils` are transforming data and not using that - which creates an inconsistent experience with users using the functions just with different input arguments.

So, its all or none. Either add your `getTransform` functions using the current styling or change all of the current versions and ones similar to it to use `std::optional` and update their local code usage. 

My preference would be that you just add the 1-2 new functions you need using the `bool` return, but if you did update the PR to make all of them use `std::optional` and update their in-line uses, I couldn't argue with that. I'll say that I try to keep the C++14 and newer feature uses to a minimum to keep it as approachable to the largest audience as possible. Sometimes its unavoidable and less approachable to _not_ use them, but this feels like a case that we're introducing C++17 features for the sake of it, rather than it being genuinely necessary. 

But, that's obviously just my single humble opinion. I wouldn't want to block you from contributing in the future over things like this, so if you want it I wouldn't block you from doing it as long as we maintain consistency. It just wouldn't be my recommendation.

nachovizzo: > I wouldn't want to block you from contributing in the future over things like this

Worry not Steve! I opened this PR to get a deeper understanding of how I can start contributing in the future without forcing you to review changes that are completely off of what's the vision for nav2.

As a summary of our discussion, this is what I plan to do now:

- [ ] merge quatri-plicated calls to the utility function to the `Source` class (discussion above)
- [ ] Remove the proposal about `std::optional`. Was just an approximation, I don't mind discarding it, I had fun doing it anyway. The main reason for me is that if we are in C++14, it's not worth introducing a hard dependency on 17. I can imagine an angry nacho trying to build this for a platform that's still not supporting c++17, and when understanding WHY we need 17 going bananas
- [ ] Use the more nav2/ROS 2 style of spitting booleans. I still don't like this at all... But I can't refactor the entire ROS 2 code base at this point.
- [ ] Add 2 extra util functions (THAT ARE NOT USED IN NAV2) but will extremely likely benefit users from the `nav2_util` package

Once you give me the OK I'll go ahead and make those changes.

PS: I still prefer this style of opening a PR and discussing rather than trying to write what I want to do in issues in terrible English. I prefer the code speak for me :P and I will never be bother to rollback a change I did

SteveMacenski: Go ahead! I don’t love how that Tf set of calls is written either, but abstracting into a method would make it a lot less ugly. 

BTW: modern ROS 2 is on C++17, so you’re fine there. If there were going to have problems, it would be far before now.

SteveMacenski:
@nachovizzo waddup here? :100:

nachovizzo:
> @nachovizzo waddup here? 💯

Done!

nav2_collision_monitor/include/nav2_collision_monitor/source.hpp
```diff
@@ -24,9 +24,9 @@
 #include "tf2/time.h"
 #include "tf2_ros/buffer.h"
 
-#include "nav2_util/lifecycle_node.hpp"
-
 #include "nav2_collision_monitor/types.hpp"
+#include "nav2_util/lifecycle_node.hpp"
+#include "std_msgs/msg//header.hpp"
```
SteveMacenski, line 29:
```suggestion
#include "std_msgs/msg//header.hpp"
```

`//`?

SteveMacenski (approved):

SteveMacenski:
2 things:
- Pull in main, CI should run again
- Do the DCO bits 

Otherwise, love it, super clean, I love abstracting out nonsense

nav2_collision_monitor/include/nav2_collision_monitor/source.hpp
nachovizzo: Well, no idea how this end up here!

nachovizzo:
@SteveMacenski done with this, I pulled `main` but the CI still fails to run :cry:

SteveMacenski:
It ran (i.e. built) there are some new test failures that popped up while our CI was down. I just haven't had a chance to look at them yet but not caused by this PR (probably; or at least in whole)

Issue #4179, opened 2024-03-14
[SmacPlanner] The footprint is not dynamic


BriceRenaudeau:
## Bug report

When the footprint of the robot is changed using the costmap topic, the planner doesn't update the collision_checker so it can plan in obstacles. It can be seen in the `/planned_footprints` (they are using the costmap footprint)

![smac_footprint_collision](https://github.com/ros-planning/navigation2/assets/48433002/6f65dc17-181a-4048-9e77-43bb8a86e81a)

 The footprint is not a dynamic parameter of the planner, and it's a good point, It must retrieve it from the costmap.

**Required Info:**

- Operating System:
  - Ubuntu22.04
- ROS2 Version:
  - Iron
- Version or commit hash:
  - source - main
- DDS implementation:
  - cyclone

#### Steps to reproduce issue
1. Start the navigation
2. Change the footprint of the costmap using the topic
3. The costmap is updated
4. The collision_checker of the planner is not updated

#### Expected behavior

The planner updates the collision_checker and the path doesn't go into obstacles

#### Actual behavior

The planner doesn't update the collision_checker and the path can go into obstacles

#### Additional information

Note : If you change the value of the parameter `angle_quantization_bins`, the collision_checker is updated and the path is good.

Implementation possibilities:
- Easy and dirty: on every `createPlan()`,  update collision_checker before setting the a_start. (tested :heavy_check_mark: )
  - Strangely to work perfectly, you need to recreate the collision_checker, not just set the footprint;
- Hard and clean :  Make collision_checker aware of the costmap changes.

SteveMacenski:
We set it on initialization https://github.com/ros-planning/navigation2/blob/09ba08b7f45004df92f77fa683ea33d8afa2acce/nav2_smac_planner/src/smac_planner_lattice.cpp#L188

I suppose we could check if its changed in the main planning callback -- if so send it the updated footprint. I wouldn't want to generically do it every iteration since we precompute orientations that would slow down the planner to do when not necessary. That should be a 10 line PR, can you open it :-)

BriceRenaudeau:
The footprint change check is already done in the `setFootprint`.

I added it to the main `create_plan` juste before setting it to the `_a_star->setCollisionChecker(&_collision_checker);`
Strangely, It doesn't work well. I had to create a new _collision_checker before.

SteveMacenski:
[...]

PR #4193, opened 2024-03-18
[SmacPlanner] Update footprint if changed

BriceRenaudeau:
This PR updates the footprint (if different) used in the smac planner at every `create_plan` call.

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4179  |
| Robotic platform tested on | Our Robot |

---

- Clear `oriented_footprints_` the fix the update issue
- Call `_collision_checker.setFootprint()` in create_plan method 

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/smac_planner_hybrid.cpp
+++ b/nav2_smac_planner/src/smac_planner_hybrid.cpp
@@ -338,6 +338,10 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
   }
 
   // Set collision checker and costmap information
+  _collision_checker.setFootprint(
+    _costmap_ros->getRobotFootprint(),
+    _costmap_ros->getUseRadius(),
+    findCircumscribedCost(_costmap_ros));
   _a_star->setCollisionChecker(&_collision_checker);
 
   // Set starting point, in A* bin search coordinates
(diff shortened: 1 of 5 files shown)


nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -338,6 +338,10 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
   }
 
   // Set collision checker and costmap information
+  _collision_checker.setFootprint(
```
SteveMacenski, line 341: Can you set to update only when changed? Recomputing each time with footprint precomputing will add unnecessary latency.

SteveMacenski (changes requested):

nav2_smac_planner/src/smac_planner_hybrid.cpp
BriceRenaudeau: The footprint change check is already done in this function:
https://github.com/ros-planning/navigation2/blob/914d881f6e154d8d7bb1975892d4642eaf4c7366/nav2_smac_planner/src/collision_checker.cpp#L51-L67

SteveMacenski:
That's not on you, that's the weird state CI is in due to the Rolling Migration to 24.04 that's happening right now and isn't completed. Things are in a weird state.

SteveMacenski (approved):

Issue #4177, opened 2024-03-12
Validate messages in subscription callbacks for critical error in population


GoesM:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu22.04
- ROS2 Version:
  -  humble
- Version or commit hash:
  -  the latest
- DDS implementation:
  -  defaulted

#### Steps to reproduce issue

Launch the navigation2 normally, as following steps:
```sh
#!/bin/bash
export ASAN_OPTIONS=halt_on_error=0:new_delete_type_mismatch=0:detect_leaks=0:log_pah=asan
source install/setup.bash
export TURTLEBOT3_MODEL=waffle
export GAZEBO_MODEL_PATH=$GAZEBO_MODEL_PATH:/opt/ros/humble/share/turtlebot3_gazebo/models
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=True use_rviz:=False use_composition:=False 
```

Curious about how nav2 face to topic-interception, i keep sending demaged `/map` msg onto topic `/map`, which is like this:
```sh
ros2 topic pub /map nav_msgs/msg/OccupancyGrid "
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: map
info:
  map_load_time:
    sec: 0
    nanosec: 0
  resolution: 0.05000000074505806
  width: 384
  height: 384
  origin:
    position:
      x: -10.0
      y: -10.0
      z: 0.0
    orientation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
[...]

PR #4206, opened 2024-03-22
provide message validation check API

GoesM:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4177  |
| Primary OS tested on | Ubuntu  |
| Robotic platform tested on | gazebo  |

---

- Add message-validation check in `nav2_util` pkg
- Add check in `nav2_amcl` for `map` message to avoid **heap-buffer-overflow** bug mentioned in #4177 

- new file named `validate_message.hpp`, which provide a structure of message-validation-check in `nav2_util`
- add a validation check for map message of `nav2_amcl`

---

#### For Maintainers: 


nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+#include "nav_msgs/msg/occupancy_grid.hpp"
+#include "nav_msgs/msg/odometry.hpp"
+
+
+namespace nav2_util
+{
+
+const int NSEC_PER_SEC = 1e9;  // 1 second = 1e9 nanosecond
+
+bool validateMsg(const builtin_interfaces::msg::Time_<std::allocator<void>> & msg)
```
SteveMacenski, line 28:
```suggestion
bool validateMsg(const builtin_interfaces::msg::Time & msg)
```

^ Why not this? For `Header` as well

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+{
+  // check item
```
SteveMacenski, line 30: Rather, `if (msg.sec ==0 && msg.nanosec == 0)`, else it should be valid.

Nanosec is a `uint32`, so it cannot be negative. Seconds can be, but I think that actually is valid in some odd cases

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+  if (msg.sec < 0) {return false;}                                      // invalid second-stamp
+  if (msg.nanosec < 0 || msg.nanosec >= NSEC_PER_SEC) {return false;}   // invalid nanosec-stamp
+  return true;
+}
+
+bool validateMsg(const std_msgs::msg::Header_<std::allocator<void>> & msg)
+{
+  //  check sub-type
+  if (!validateMsg(msg.stamp)) {return false;}
+  // msg.frame_id :  @todo any check for it ?
```
SteveMacenski, line 40: Check if frame id is an empty string

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+  return true;
+}
+
+bool validateMsg(const geometry_msgs::msg::Point & msg)
+{
+  // @todo any check for following item ?
```
SteveMacenski, line 46: NaN or Inf

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+  msg.x;
+  msg.y;
+  msg.z;
+  return true;
+}
+bool validateMsg(const geometry_msgs::msg::Quaternion & msg)
+{
+  // @todo any check for following item ?
+  msg.x;
```
SteveMacenski, line 55: NaN or Inf for any values

It could also be checked that its normalized: https://math.stackexchange.com/questions/1703466/normalizing-a-quaternion

If its not normalized within some `eta` value, then reject

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+bool validateMsg(const geometry_msgs::msg::Pose & msg)
+{
+  // check sub-type
+  if (!validateMsg(msg.position)) {return false;}
+  if (!validateMsg(msg.orientation)) {return false;}
+  return true;
+}
+
+
+// 验证地图元信息的函数
```
SteveMacenski, line 71: Lets translate that :wink: `Function to verify map meta information`

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+bool validateMsg(const nav_msgs::msg::MapMetaData & msg)
+{
+  // check sub-type
+  if (!validateMsg(msg.map_load_time)) {return false;}
```
SteveMacenski, line 75: I don't think `map_load_time` is critical enough for us to reject an update because that isn't properly formatted, in general

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+  if (!validateMsg(msg.origin)) {return false;}
+
+  // @todo  any check for following item ?
```
SteveMacenski, line 78: Check non-zero I suppose

SteveMacenski (changes requested):
Looks good for a first go! I think it has some linting issues and needs to add unit tests for each of the functions, but overall with a few trivial changes this looks like a good starting point!

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: > which default is (0,0)

Since I noticed it, so I think `if (msg.sec ==0 && msg.nanosec == 0)` it still should be valid ?

> Nanosec is a uint32, so it cannot be negative. Seconds can be

well, if negative seconds is allowed, I'm going to change this function as following ? 

```cpp
bool validateMsg(const builtin_interfaces::msg::Time & msg)
{
    // check item 
    // negative second is allowed 
    if (msg.nanosec >= NSEC_PER_SEC) return false;  // invalid nanosec-stamp
    return true; 
}


```

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: I was concerned about the possibility of a default empty string for the frame_id. That's why I didn't include that check to ensure the generality of this function's functionality.

So can we be certain that empty frame_id must be definitely invalid ?

nav2_util/include/nav2_util/validate_messages.hpp
SteveMacenski: When would `msg.sec ==0 && msg.nanosec == 0` be valid?

nav2_util/include/nav2_util/validate_messages.hpp
SteveMacenski: That is the default, which is why we need to check that it contains some actual valid information 

Empty frame is definitely invalid for the majority of things that include a header. There are some special cases to that, but not anywhere that we're talking about using this

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: I catched a map message by `ros2 topic echo /map` during launching nav2, and its header is:

```yaml
# by: ros2 topic echo /map
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: map
...
```

SteveMacenski: I'd argue that's invalid and we should fix that as part of this PR :-) You've uncovered a problem by checking the validity of messages, that's good!

GoesM: I see. It seems that we need to add a timestamp-mechanism to the publisher of the map. I'll update this PR over the weekend or possibly later, trying to address all of today's issues in one go. :-)

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: I'm not entirely sure if a precision error tolerance like epsilon = 1e-6 is appropriate; I'll test it locally later.

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: well , I had a try to find how time-stamp of a map-message is set by `nav2_map_server` as following:

```cpp
  // Since loadMapFromFile() does not belong to any node, publishing in a system time.
  rclcpp::Clock clock(RCL_SYSTEM_TIME);
  msg.info.map_load_time = clock.now();
  msg.header.frame_id = "map";
  msg.header.stamp = clock.now();
```

HOWEVER, I `ros2 topic echo /map` during launching nav2 for many times, while the time-stamp of message I got is still 
```yaml
# by: ros2 topic echo /map
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: map
...
```

It's odd to me, I'd have more check in code later.

GoesM:
I've updated some conditions in validateMsg(), but I've encountered a few details that need addressing. I'll take some time to figure out how to resolve them. However, I'll go ahead and update my current code to share my progress here~  ^_^

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1390,6 +1391,10 @@ void
 AmclNode::mapReceived(const nav_msgs::msg::OccupancyGrid::SharedPtr msg)
 {
   RCLCPP_DEBUG(get_logger(), "AmclNode: A new map was received.");
+  if (!nav2_util::validateMsg(*msg)) {
```
SteveMacenski, line 1394: You can put this in the Static Layer's callback too :-)

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,142 @@
...
+  if (!validateMsg(msg.orientation)) {return false;}
+  return true;
+}
+
+
+// Function to verify map meta information
+bool validateMsg(const nav_msgs::msg::MapMetaData & msg)
+{
+  // check sub-type
+  msg.map_load_time;   // no neccessary to check it
```
SteveMacenski, line 112: Can remove this, no need to call it

nav2_amcl/src/amcl_node.cpp
GoesM: you mean such msg-type used in `costmap` ?

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: of course.

nav2_util/include/nav2_util/validate_messages.hpp
GoesM: It' so strange that I can't understand what happened. The time-stamp of message I catch by command is always 
```yaml
sec:0
nanosec:0
```
Fortunately, this issue does not affect the usage and operation of navigation2. Perhaps there is some mechanism in place that I am not familiar with, ensuring this aspect remains unaffected.

SteveMacenski: I don't understand. I assume there needs to be some updates to missing timestamps like you mentioned? That + running all the tests to make sure nothing else breaks + fixing the linting issues is the last before merging

GoesM:
# About the **time-stamp** of map message 

### current code for updating the time-stamp

https://github.com/ros-planning/navigation2/blob/0be2f25782413677fedc58ea090c6ecfa8b9a6b8/nav2_map_server/src/map_server/map_server.cpp#L243-L248

in this way, we would **NOT** update the stamp successfully. 

Though I'm not very clear about the reson , but we can prove it by following inserted code:

```cpp
//goes insert
  updateMsgHeader();
  std::cout <<
    "[GOES|DEBUG] [map_io]: msg.header.stamp.sec: " << msg_.header.stamp.sec << std::endl
        << "msg_.header.stamp.nanosec" << msg_.header.stamp.nanosec << std::endl; 
  std::cout <<
    "[GOES|DEBUG] [map_io]: msg_.info.map_load_time.sec : " << msg_.info.map_load_time.sec << std::endl
        << "msg_.header.msg_.info.map_load_time.nanosec" << msg_.info.map_load_time.nanosec << std::endl; 
// insert end
```

then , we could catch the log as following:
```
[map_server-6] [GOES|DEBUG] [map_io]: msg.header.stamp.sec: 0
[map_server-6] msg_.header.stamp.nanosec0
[map_server-6] [GOES|DEBUG] [map_io]: msg_.info.map_load_time.sec : 0
[map_server-6] msg_.header.msg_.info.map_load_time.nanosec0
```


### suggested way to update the time-stamp

I've tried to update the stamp in another way, like following:
```cpp
void MapServer::updateMsgHeader()
{
  rclcpp::Clock clock(RCL_SYSTEM_TIME);
  msg_.info.map_load_time = clock.now();
  msg_.header.frame_id = frame_id_;
  msg_.header.stamp = clock.now();
}
```

then , we could catch the log as following, which shows a successful result.
```
[map_server-6] [GOES|DEBUG] [map_io]: msg.header.stamp.sec: 1712391915
[map_server-6] msg_.header.stamp.nanosec28685368
[map_server-6] [GOES|DEBUG] [map_io]: msg_.info.map_load_time.sec : 1712391915
[map_server-6] msg_.header.msg_.info.map_load_time.nanosec28684872
```

SteveMacenski:
That's not a solution unfortunately since we don't always run in ROS (i.e. simulation) time. The clock of the node should be set to the right time source. Does the timestamp not get populated from the service call as well or just on the `on_configure` first go?

GoesM:
I fix some typo this time.

# the reason why time-stamp of message from map-server is (zero,zero)

After some experiments, it seems clear for me that updateMsgHeader as following would bring non-zero time after `on_activate()` finished.
Because the `node->now()` would not start its time until it's activated. 
```
void MapServer::updateMsgHeader() 
 { 
   msg_.info.map_load_time = now(); 
   msg_.header.frame_id = frame_id_; 
   msg_.header.stamp = now(); 
 } 
```

SO, here's still the trouble that :

map_server only send map-message once during `on_activate()` but NOT after activated, so that the time-stamp would be zero and rejected by our `validation_message` check, then nav2_amcl would not start its work.
(how map_server send map-message shown as following:

https://github.com/ros-planning/navigation2/blob/677e6accb3aea9380678da7d623b3387578d3549/nav2_map_server/src/map_server/map_server.cpp#L128-L144

Here's some different ways I come up with to solve the trouble:
- send the map-message first time after `on_activate()` (for example let `nav2_amcl` call the service "/nav2_map_server/load_map" provide by nav2_map_server)
- remove validation check for **zero-time-stamp**
- using the following way to update header:
```cpp
void MapServer::updateMsgHeader()
{
  rclcpp::Clock clock(RCL_SYSTEM_TIME);
  msg_.info.map_load_time = clock.now();
  msg_.header.frame_id = frame_id_;
  msg_.header.stamp = clock.now();
}
```

I'm not very sure if there are any better methods available, or which one among these methods is most suitable. I should primarily rely on your advice regarding which direction to proceed with modifications. : )

SteveMacenski:
> remove validation check for zero-time-stamp

Lets do that.

Also please pull in main or rebase so CI can run. you also have a couple linting problems that the Actions job identified. After that, this should be good to merge!

nav2_util/include/nav2_util/validate_messages.hpp
```diff
@@ -0,0 +1,142 @@
...
+bool validateMsg(const geometry_msgs::msg::Quaternion & msg)
+{
+  //  check sub-type
+  if (!validateMsg(msg.x)) {return false;}
+  if (!validateMsg(msg.y)) {return false;}
+  if (!validateMsg(msg.z)) {return false;}
+  if (!validateMsg(msg.w)) {return false;}
+
+  // logic check
+  // 1> the quaternion should be normalized:
```
SteveMacenski, line 89: No need for the comments -- checking the normalization is the same as any distance calculation `sqrt(componentA*componentA + componentB*componentA + ...)`

GoesM:
Oops, I noticed that the CI test failed due to the spin_behavior_test, which doesn't seem to be caused by the validation_message.

SteveMacenski:
Dont worry about the spin test - its flaky. You do though have a conflict you need to resolve before we can merge

GoesM:
I think my rebase did not work correctly and I believe it's more wise for me to open another PR.

PR #4209, opened 2024-03-22
Enable reloading BT xml file with same name

huemerj:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Inhouse AGV & Gazebo |

---

- Current state: In the current implementation of the `BtActionServer`, a new BT is not loaded if its xml name is the same as the current one. This means that if you want to change and then test a BT xml file (while keeping the same filename), you need to restart the action server each time the file changes. 

- This PR: Adds a boolean parameter `always_reload_bt_xml` (defaults to `false`) to decide if a new BT should always be loaded, even if the xml names are the same. This allows keeping the action server running while changing/developing the xml file.
 

- Might be necessary to add the new parameter `always_reload_bt_xml` to the Behavior-Tree Navigator docs.

---

- Would be nice to backport this PR to Humble/Iron.

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
@@ -90,10 +90,11 @@ public:
   /**
    * @brief Replace current BT with another one
    * @param bt_xml_filename The file containing the new BT, uses default filename if empty
+   * @param overwrite If true, the new BT will be reloaded even if the filename is the same
    * @return bool true if the resulting BT correspond to the one in bt_xml_filename. false
    * if something went wrong, and previous BT is maintained
    */
-  bool loadBehaviorTree(const std::string & bt_xml_filename = "");
+  bool loadBehaviorTree(const std::string & bt_xml_filename = "", bool overwrite = false);
 
   /**
    * @brief Getter function for BT Blackboard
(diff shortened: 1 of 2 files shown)


SteveMacenski (changes requested):
Rename to `always_reload_bt_xml` and add to the Nav2 configuration / migration guides and this LGTM

huemerj:
> Rename to `always_reload_bt_xml` and add to the Nav2 configuration / migration guides and this LGTM
> 
> Please heed the DCO bot (sign commits `-s`)

Done. https://github.com/ros-planning/navigation.ros.org/pull/534

SteveMacenski (approved):

SteveMacenski:
@huemerj please heed the DCO bot, all commits need to be signed, then I can merge this. If you click on the CI job, it'll give you instructions about how to adjust your previous commits 

I'm going to merge this but I'd like to note for future onlookers of this: This is useful for the context @huemerj brings it up in. If you're using this because you have a BT that isn't properly resetting its state with your custom BT nodes, so using this to hard reset everything, I highly suggest fixing your underlying issues which could be causing other issues on the blackboard which could possibly persist in some cases. This may seem like an easy solution, but I suggest spending a little bit of time to look into the proper solution, especially in robotics where small mistakes in odd cases can cause real headaches.

huemerj:
@SteveMacenski Would be nice if you could please backport this to Humble/Iron.

SteveMacenski:
Next release cycle it will be

christophfroehlich:
> Next release cycle it will be

@SteveMacenski May I ping you again and ask for a backport?

SteveMacenski:
If it wasn't backported (which it looks like it could have been?) it was due to huge merge conflicts on cherry picking. If you'd like this in iron, please submit a PR against `iron` making these changes manually and I can merge that and have it released in the next sync. 

Sorry about the manual-ness, the automated tools must have failed badly enough that I judged it not possible at the time.

PR #4212, opened 2024-03-24
Add uint suffix

MikeWrock:
The space_nav repo uses slightly stricter compiler flags that causes the modified lines to result in a docker build failure:

https://github.com/space-ros/docker/actions/runs/8406248510/job/23019919497

Fixing this upstream allows up to stay current with upstream repos


MikeWrock:
updating to target main

Issue #4185, opened 2024-03-15
Remove RPP Dependency on AMCL


SteveMacenski:
Per https://github.com/ros-planning/navigation2/pull/4140

Angles should be able to do the trick instead: https://github.com/ros/angles

PR #4222, opened 2024-03-27
Remove AMCL dependency from RPP

jncfa:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4185 |

---

Replaces the `nav2_amcl::angleutils` usage in RPP with `angles` and drops the dependency to `nav2_amcl`, which in this case was being used just for one angle normalization call.

---

#### For Maintainers: 


SteveMacenski (approved):
Perfect, thank you! A much better solution :smile:

Issue #4199, opened 2024-03-19
[collision monitor] Dynamic radius for circle polygon


anaelle-sw:
#### Feature description
The `collision_monitor` polygons are dynamic, by subscribing to a topic on which the list of polygon's points are published.
The circle polygon could be dynamic as well. 

#### Implementation considerations
In order to keep a design similar to the other polygons, the circle polygon could be updated by subscribing to a topic on which the radius would be published. 
The topic can be defined by using the same parameter than for the other polygons: `<polygon_name>.polygon_sub_topic` (the subscription type would be defined by the polygon type).
 In order to not break the current usage, if no topic is defined as a parameter, the parameter `<polygon_name>.radius` will be used.
 I can open a PR with an implementation of this solution.

SteveMacenski:
> The circle polygon could be dynamic as well.

All I need to hear, please open a PR!

PR #4226, opened 2024-03-28
[Collision monitor] Dynamic radius for circle polygon

anaelle-sw:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4199 |
| Robotic platform tested on | RobOcc's robot |

---

* New parameter `<polygon_name>.radius_sub_topic`, for circle type polygon
* New float-type subscription to  `<polygon_name>.radius_sub_topic`, for circle type polygon. If a new radius message is received, the circle radius is updated

* Add new parameter `<polygon_name>.radius_sub_topic` to default configs, documentation page, migration guide

---

#### For Maintainers: 


nav2_collision_monitor/src/polygon.cpp
```diff
@@ -450,6 +438,25 @@ bool Polygon::getParameters(
...
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+
+  if (!polygon_sub_topic.empty()) {
+    RCLCPP_INFO(
+      logger_,
+      "[%s]: Subscribing on %s topic for polygon",
+      polygon_name_.c_str(), polygon_sub_topic.c_str());
```
SteveMacenski, line 452: Can't you use the same `polygon_sub_topic` and when this is a polygon do this operation with the `PolygonStamped` type and when using circle use the `Float32` type? That would remove the need for some or all of the changes in  `Circle::getParameters`

nav2_collision_monitor/src/circle.cpp
```diff
@@ -84,25 +84,88 @@ bool Circle::getParameters(
...
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+
+  if (!polygon_sub_topic.empty()) {
+    RCLCPP_INFO(
+      logger_,
+      "[%s]: Subscribing on %s topic for polygon",
+      polygon_name_.c_str(), polygon_sub_topic.c_str());
+    rclcpp::QoS polygon_qos = rclcpp::SystemDefaultsQoS().transient_local();
```
SteveMacenski, line 136: Ditto on QoS

SteveMacenski (changes requested):

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: We can use inheritance to our benefit here :-)

SteveMacenski:
@anaelle-sw only major missing thing I see is tests. You should be able to mimic the tests written for the dynamic polygons for the circles (and a little simpler)

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -450,6 +438,25 @@ bool Polygon::getParameters(
...
+    rclcpp::QoS polygon_qos = rclcpp::SystemDefaultsQoS().transient_local();
```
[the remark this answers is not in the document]
anaelle-sw: Oops, good catch! I fixed it

nav2_collision_monitor/src/polygon.cpp
anaelle-sw: I agree it's a better design to keep same parameter `polygon_sub_topic` for both polygon and circle types! 

I moved declaration and init of parameters `polygon_sub_topic` and `footprint_topic` to function `Polygon::getCommonParameters()`. So that in function `Circle::getParameters` there is only the declaration and init of parameter `radius`.

nav2_collision_monitor/src/circle.cpp
anaelle-sw: Fixed

anaelle-sw:
> only major missing thing I see is tests.
Ok, I will add some tests

> Given these changes requested, also the docs should be updated in line with not having a new parameter + explaining the use in the existing parameter for a Circle vs Polygon type.
I updated the docs with the changes you suggested

SteveMacenski:
Rebase / pull in main to get CI to turn over again :smile: Docs LGTM.

nav2_collision_monitor/src/circle.cpp
```diff
@@ -80,12 +80,14 @@ bool Circle::getParameters(
     throw std::runtime_error{"Failed to lock node"};
   }
 
-  if (!getCommonParameters(polygon_pub_topic)) {
+  // Clear the polygon subscription topic. It will be set later, if necessary.
+  polygon_sub_topic.clear();
```
SteveMacenski, line 84: Why clear?

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -372,6 +363,28 @@ bool Polygon::getCommonParameters(std::string & polygon_pub_topic)
         node, polygon_name_ + ".polygon_pub_topic", rclcpp::ParameterValue(polygon_name_));
       polygon_pub_topic = node->get_parameter(polygon_name_ + ".polygon_pub_topic").as_string();
     }
+
+    if (action_type_ == STOP || action_type_ == SLOWDOWN || action_type_ == LIMIT ||
```
SteveMacenski, line 367: `if (action_type_ != APPROACH) {} else {}` is more concise

SteveMacenski (changes requested):

nav2_collision_monitor/src/polygon.cpp
anaelle-sw: I kept the test as it was before, but you're right, I changed it.

SteveMacenski:
Please pull in main or rebase so that CI can turn over, its been fixed for about a week

nav2_collision_monitor/src/circle.cpp
```diff
@@ -94,15 +96,77 @@ bool Circle::getParameters(
...
+      "[%s]: Polygon circle radius is not defined. Using dynamic subscription instead.",
+      polygon_name_.c_str());
+  }
+
+  if (polygon_sub_topic.empty()) {
     RCLCPP_ERROR(
       logger_,
-      "[%s]: Error while getting circle parameters: %s",
-      polygon_name_.c_str(), ex.what());
+      "[%s]: Error while getting polygon parameters: static radius and sub topic both not defined",
```
SteveMacenski, line 114:
```suggestion
      "[%s]: Error while getting circle parameters: static radius and sub topic both not defined",
```
This is still sp. to circle, no?

SteveMacenski (approved):

anaelle-sw:
Some of the tests failed, so few changes were required:
* Initializing the variable `radius_squared_` to `-1` (will be updated either if the parameter `radius` is declared, or either when receiving a new message on radius topic)
* Updating the member function `isShapeSet()` for `Circle` class, so that it will return `true` only if `radius_squared_` is not set to `-1` 
* In member function `getParameters`, first try to get static polygon definition param (`points` or `radius`), and then call function `getCommonParameters`
* In member function `getCommonParameters`, declare topics param (`polygon_sub_topic` or `footprint_topic`), only if function `isShapeSet()` returns `false` (so only if static polygon definition param (`points` or `radius`) was not declared first)

anaelle-sw:
I fixed the tests failures by: 

* Not setting a default value for parameter `polygon_sub_topic`. So if the user doesn't set it,  an exception will be raised (and catch) and the parameter won't be declared.

* In case that the `points` parameter value is not pairs of number, clearing the polygon's points. In this case, the member function `getPolygonFromString` was already returning `false`, but if the polygon generation had already started, the partially-built polygon was kept as it was. Calling `polygon.clear()` prevents from using this partial polygon, and function `isShapeSet()` would return `false`. 
Since I had to implement next bullet point too, this one could actually be reverted and the tests shouldn't fail. But I thought that may be it wouldn't bad to prevent usage of a partially built polygon (from parameter defined with format error).


* Finally, last failure source was the velocity polygon. During configuration, it gets a list of points for each of its sub polygon, but the variable `polygon_` is actually not set until receiving a new velocity message. So the function `isShapeSet()` would return `false` until then, and when calling `getCommonParameters` the parameter `polygon_sub_topic` or `footprint_topic` would be declared, which is not expected. 
My first idea was to apply the sub polygon for null velocity range right after getting all points from all sub polygons. But I think it might induce unwanted behaviors, since the robot may not be actually stopped during configuration (it's not my case, but I rather not assume this). 
The second idea, which I implemented, is to add a new boolean argument `use_dynamic_sub` to member function `getCommonParameters`. When this argument is set to `false`, the parameters `polygon_sub_topic` and `footprint_topic`  won't be declared. For velocity polygons, this argument is always set to `false`. For circles, this argument is set to `true` only if `radius` parameter is not  declared. For polygons,  this argument is set to `true` only if `points` parameter is not declared or is malformed.

anaelle-sw:
I have also resolved the conflict of the docs PR

nav2_collision_monitor/src/circle.cpp
```diff
@@ -70,6 +70,15 @@ int Circle::getPointsInside(const std::vector<Point> & points) const
   return num;
 }
 
+bool Circle::isShapeSet()
+{
+  if (radius_squared_ == -1.0) {
```
SteveMacenski, line 75: Could default to Nan? Just need to check on first use everywhere if `isnan()`

SteveMacenski (approved):

anaelle-sw:
> I see that you added `use_dynamic_sub` instead of clearing the string and checking if its empty - not that I need you to justify it, but why wasn't checking empty string sufficient?

If static polygon is used (which means parameter `points` is properly declared), the parameters `polygon_sub_topic` and `footprint_topic` are supposed to not be declared (or at least, this is the expected behaviour in tests, may be this can change...?). So with current design, there are two solutions:
- undeclared parameters `polygon_sub_topic` and `footprint_topic`  if static polygon is used. I tried this, but dynamically typed parameters cannot be undeclared, so this doesn't build
- use an new argument (`use_dynamic_sub`) to not declare parameters  `polygon_sub_topic` and `footprint_topic` if not necessary

Or we could change design. For instance, declaring 
`polygon_sub_topic` and `footprint_topic` in `getParameters` (instead of `getCommonParameters `), or considering the sub topic parameters can be declared in case the static polygon is used.

kaichie:
Overall the changes looks good to me and seems to follow the existing pattern. Happy to see we can now dynamically update the Circle polygon! :+1: :star2:

SteveMacenski:
OK! Any reason I shouldn't merge this on Monday @anaelle-sw ?

anaelle-sw:
> OK! Any reason I shouldn't merge this on Monday @anaelle-sw ?

I'll just rebase branch on main and it should be ready to be merged

Issue #4197, opened 2024-03-19
[collision monitor] Select the observation sources used with each polygon


anaelle-sw:
#### Feature description
Currently, the `collision_monitor` associates all the observation sources to all polygons. 

In some cases, we need to apply only some of the observation sources, for some of the polygons.

#### Implementation considerations 
A simple solution consists in launching multiple `collision_monitor` nodes, and split the observation sources (or polygons) definitions between these nodes. But it would require to define multiple times some of the polygons (or some of the observations sources). Plus, having multiple `collision_monitor` nodes it not necessarily a good design since it could induce latency in the velocity command.

A nicer solution consists in defining a new parameter `<polygon_name>.observation_sources`, for each polygon. This parameter would be a list of string, containing some (or all) of the names of the defined observation sources. Only the observations sources which names are listed in `<polygon_name>.observation_sources` will be used to check collision with the polygon `<polygon_name>`. The parameter `observation_sources` still exists and contains the names of all observation sources, and the sources definition is unchanged. If the list `<polygon_name>.observation_sources` is empty, all sources would be used (to no break previous usage). I can open an PR with an implementation of this solution.

SteveMacenski:
> In some cases, we need to apply only some of the observation sources, for some of the polygons.

Is it that you want to toggle certain sources for certain polygons, or is it more that you want to disable certain sources when the robot is in certain states (i.e. a box blocking this sensor on a forklift)?  Currently, each source can be disabled with a dynamic parameter which you don't want to use that particular source at that moment. 

Doing what you suggest is possible, but I think its going to be more ugly than you think, given how we pass just a `data` vector into the polygons that is quashed from all sources. We'd have to adjust the input to the polygon's functions to being individual data source vectors with their associated names rather than a vector of point values that have no attributes about their ownership (or, I suppose a vector of points with their name associations per-point). Not rocket science, but a little more involved since we need to propagate that source-ownership information to the polygons to be able to check with that parameter. 

Can you motivate this feature a bit? I'm having a hard time imagining when this would be helpful, with the exception of optimizing performance by only checking certain sources over less number of polygons -- but my intuition is that is going to be negligible improvement to the full system -- but happy to be proved wrong

anaelle-sw:
In this case, we don't really need to toggle sources while the node is up. Actually, we need to define at launch time which source will be checked with which polygon, for each polygon. Here is our use case for more context:

[...]

PR #4227, opened 2024-03-28
[collision monitor] Select the observation sources used with each polygon

anaelle-sw:
## Basic Info
| ------ | ----------- |
| Ticket(s) this addresses   | #4197 |
| Robotic platform tested on | RobOcc's robot |

---

New string vector parameter `<polygon_name>.sources_names`. Only the sources which names are specified in parameter are used to check collision with current polygon. If the parameter is not set, all the observation sources are used.

Add new parameter `<polygon_name>.sources_names` to default configs, documentation page, and migration guide

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/test/collision_monitor_node_test.cpp
+++ b/nav2_collision_monitor/test/collision_monitor_node_test.cpp
@@ -143,7 +143,8 @@ public:
   void setCommonParameters();
   void addPolygon(
     const std::string & polygon_name, const PolygonType type,
-    const double size, const std::string & at);
+    const double size, const std::string & at,
+    const std::vector<std::string> & sources_names = std::vector<std::string>());
   void addPolygonVelocitySubPolygon(
     const std::string & polygon_name, const std::string & sub_polygon_name,
     const double linear_min, const double linear_max,
@@ -309,7 +310,8 @@ void Tester::setCommonParameters()
 
 void Tester::addPolygon(
   const std::string & polygon_name, const PolygonType type,
-  const double size, const std::string & at)
+  const double size, const std::string & at,
+  const std::vector<std::string> & sources_names)
 {
   if (type == POLYGON) {
     cm_->declare_parameter(
@@ -408,6 +410,13 @@ void Tester::addPolygon(
     polygon_name + ".polygon_pub_topic", rclcpp::ParameterValue(polygon_name));
   cm_->set_parameter(
     rclcpp::Parameter(polygon_name + ".polygon_pub_topic", polygon_name));
+
+  if (!sources_names.empty()) {
+    cm_->declare_parameter(
+      polygon_name + ".sources_names", rclcpp::ParameterValue(sources_names));
+    cm_->set_parameter(
+      rclcpp::Parameter(polygon_name + ".sources_names", sources_names));
+  }
 }
 
 void Tester::addPolygonVelocitySubPolygon(
@@ -1499,9 +1508,10 @@ TEST_F(Tester, testCollisionPointsMarkers)
   // Share TF
   sendTransforms(curr_time);
 
+  // No source published, empty marker array published
   publishCmdVel(0.5, 0.2, 0.1);
   ASSERT_TRUE(waitCollisionPointsMarker(500ms));
-  ASSERT_EQ(collision_points_marker_msg_->markers[0].points.size(), 0u);
+  ASSERT_EQ(collision_points_marker_msg_->markers.size(), 0u);
 
   publishScan(0.5, curr_time);
   ASSERT_TRUE(waitData(0.5, 500ms, curr_time));
@@ -1575,6 +1585,48 @@ TEST_F(Tester, testVelocityPolygonStop)
   ASSERT_TRUE(waitActionState(500ms));
   ASSERT_EQ(action_state_->action_type, STOP);
   ASSERT_EQ(action_state_->polygon_name, "VelocityPoylgon");
+
+  // Stop Collision Monitor node
+  cm_->stop();
+}
+
+TEST_F(Tester, testSourceAssociatedToPolygon)
+{
+  // Set Collision Monitor parameters:
+  // - 2 sources (scan and range)
+  // - 1 stop polygon associated to range source
+  // - 1 slowdown polygon (associated with all sources by default)
+  setCommonParameters();
+  addSource(SCAN_NAME, SCAN);
+  addSource(RANGE_NAME, RANGE);
+  std::vector<std::string> range_only_sources_names = {RANGE_NAME};
+  std::vector<std::string> all_sources_names = {SCAN_NAME, RANGE_NAME};
+  addPolygon("StopOnRangeSource", POLYGON, 1.0, "stop", range_only_sources_names);
+  addPolygon("SlowdownOnAllSources", POLYGON, 1.0, "slowdown");
+  setVectors({"StopOnRangeSource", "SlowdownOnAllSources"}, {SCAN_NAME, RANGE_NAME});
+
+  // Start Collision Monitor node
+  cm_->start();
+
+  // Share TF
+  rclcpp::Time curr_time = cm_->now();
+  sendTransforms(curr_time);
+
+  // Publish sources so that :
+  // - scan obstacle is in polygons
+  // - range obstacle is far away from polygons
+  publishScan(0.5, curr_time);
+  publishRange(4.5, curr_time);
+  ASSERT_TRUE(waitData(0.5, 500ms, curr_time));
+
+  // Publish cmd vel
+  publishCmdVel(0.5, 0.0, 0.0);
+  ASSERT_TRUE(waitCmdVel(500ms));
+
+  // Since the stop polygon is only checking range source, slowdown action should be applied
+  ASSERT_TRUE(waitActionState(500ms));
+  ASSERT_EQ(action_state_->action_type, SLOWDOWN);
+  ASSERT_EQ(action_state_->polygon_name, "SlowdownOnAllSources");
 
   // Stop Collision Monitor node
   cm_->stop();
--- a/nav2_collision_monitor/test/polygons_test.cpp
+++ b/nav2_collision_monitor/test/polygons_test.cpp
@@ -46,6 +46,7 @@ static const char POLYGON_SUB_TOPIC[]{"polygon_sub"};
 static const char POLYGON_PUB_TOPIC[]{"polygon_pub"};
 static const char POLYGON_NAME[]{"TestPolygon"};
 static const char CIRCLE_NAME[]{"TestCircle"};
+static const char OBSERVATION_SOURCE_NAME[]{"source"};
 static const std::vector<double> SQUARE_POLYGON {
   0.5, 0.5, 0.5, -0.5, -0.5, -0.5, -0.5, 0.5};
 static const std::vector<double> ARBITRARY_POLYGON {
@@ -235,7 +236,11 @@ public:
 
 protected:
   // Working with parameters
-  void setCommonParameters(const std::string & polygon_name, const std::string & action_type);
+  void setCommonParameters(
+    const std::string & polygon_name, const std::string & action_type,
+    const std::vector<std::string> & observation_sources =
+    std::vector<std::string>({OBSERVATION_SOURCE_NAME}),
+    const std::vector<std::string> & sources_names = std::vector<std::string>());
   void setPolygonParameters(
     const char * points,
     const bool is_static);
@@ -289,7 +294,10 @@ Tester::~Tester()
   tf_buffer_.reset();
 }
 
-void Tester::setCommonParameters(const std::string & polygon_name, const std::string & action_type)
+void Tester::setCommonParameters(
+  const std::string & polygon_name, const std::string & action_type,
+  const std::vector<std::string> & observation_sources,
+  const std::vector<std::string> & sources_names)
 {
   test_node_->declare_parameter(
     polygon_name + ".action_type", rclcpp::ParameterValue(action_type));
@@ -336,6 +344,18 @@ void Tester::setCommonParameters(const std::string & polygon_name, const std::st
     polygon_name + ".polygon_pub_topic", rclcpp::ParameterValue(POLYGON_PUB_TOPIC));
   test_node_->set_parameter(
     rclcpp::Parameter(polygon_name + ".polygon_pub_topic", POLYGON_PUB_TOPIC));
+
+  test_node_->declare_parameter(
+    "observation_sources", rclcpp::ParameterValue(observation_sources));
+  test_node_->set_parameter(
+    rclcpp::Parameter("observation_sources", observation_sources));
+
+  if (!sources_names.empty()) {
+    test_node_->declare_parameter(
+      polygon_name + ".sources_names", rclcpp::ParameterValue(sources_names));
+    test_node_->set_parameter(
+      rclcpp::Parameter(polygon_name + ".sources_names", sources_names));
+  }
 }
 
 void Tester::setPolygonParameters(
@@ -863,7 +883,7 @@ TEST_F(Tester, testPolygonGetCollisionTime)
   nav2_collision_monitor::Velocity vel{0.5, 0.0, 0.0};  // 0.5 m/s forward movement
   // Two points 0.2 m ahead the footprint (0.5 m)
   std::unordered_map<std::string, std::vector<nav2_collision_monitor::Point>> points_map;
-  points_map.insert({"source", {{0.7, -0.01}, {0.7, 0.01}}});
+  points_map.insert({OBSERVATION_SOURCE_NAME, {{0.7, -0.01}, {0.7, 0.01}}});
   // Collision is expected to be ~= 0.2 m / 0.5 m/s seconds
   EXPECT_NEAR(polygon_->getCollisionTime(points_map, vel), 0.4, SIMULATION_TIME_STEP);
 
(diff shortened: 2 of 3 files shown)


nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -405,7 +405,14 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in, const std_msgs::msg:
   }
 
   // Points array collected from different data sources in a robot base frame
-  std::vector<Point> collision_points;
```
SteveMacenski, line 408: This variable is no longer defined but appears to still be used. Does this compile / work?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -405,7 +405,14 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in, const std_msgs::msg:
...
+  std::unordered_map<std::string, std::vector<Point>> sources_collision_points_map;
+
+  // Fill collision_points array from different data sources
+  for (std::shared_ptr<Source> source : sources_) {
```
SteveMacenski, line 411: The source data population happens below, I don't think you should be adding a new duplicate but adjusting the old one, no? 

If you create the map element, you can pass in the reference to that vector to `getData()` so that it populates the map in-line rather than copying it twice (once from the source, another to the map).

```
  // Fill collision_points array from different data sources
  for (std::shared_ptr<Source> source : sources_) {
    if (source->getEnabled()) {
      if (!source->getData(curr_time, collision_points) &&  // <-- Right here populate an element of the map rather than the vector (that is no longer defined)
        source->getSourceTimeout().seconds() != 0.0)
      {
        action_polygon = nullptr;
        robot_action.polygon_name = "invalid source";
        robot_action.action_type = STOP;
        robot_action.req_vel.x = 0.0;
        robot_action.req_vel.y = 0.0;
        break;
      }
    }
  }
```

SteveMacenski: It seems like this for loop should be deleted and the block of code below needs to be adjusted

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -468,6 +475,28 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in, const std_msgs::msg:
     // Update polygon coordinates
     polygon->updatePolygon(cmd_vel_in);
 
+    // Get collision points for sources associated to polygon
+    std::vector<std::string> sources_names = polygon->getSourcesNames();
```
SteveMacenski, line 479: It seems to me to be more sensible to pass in a reference to the map of collision points to `isPointsInside` function for each of the polygons and let each of the polygons use its own internal `getSourcesNames` when iterating through the map to decide if it wants to use it or not.

In that case, this block of code would be removed and we just pass in the map of collision points rather than the vector of collision points to the `processXYZ()` functions

SteveMacenski (changes requested):
I think we  can do this more efficiently and straight forward
- Create the unordered map
- Adjust the current `  for (std::shared_ptr<Source> source : sources_) {` loop to contain (1) creating an entry for a source name key with an empty vector as the value. The return of insert will give you an iterator to that pair
- Pass into `getData` the reference to the value (vector) to populate. That eliminates an entire data copy
- Adjust the `processStopSlowdownLimit` / `processApproach` / etc functions to take in this map
- Pass the map to the `isPointInside`
- When iterating through the sources, use the sources names to decide if you want to use that one or not

This eliminates any extra looping and a bunch of data copies (and not parallel populating a map and a vector of data)

SteveMacenski:
@anaelle-sw Hi! Any update here?

anaelle-sw:
Sorry, I haven't had the time to work on this properly lately. I will rebase the branch on main and begin to apply to changes you requested by the end of the week.

SteveMacenski:
Thanks @anaelle-sw  :-)

anaelle-sw:
The MR is now rebased on current main branch. I have also applied your optimization suggestions. But I have not tested yet the behaviour with these changes.

I entirely agree with most of the optimizations you asked for, but actually I have a slightly better solution about function `getPointsInside()`:

You suggested to pass the entire map (which associates the sources names to their vector of points) to function `getPointsInside()`, instead of a simple array of points. So, in this function, we should have a "for" loop on the map in order to choose to use or not each source (depending on the polygon config). This clearly optimizes the processing of points for Stop, Slowdown, and Limit polygons (function `processStopSlowdownLimit()` is calling only **once** function `getPointsInside`).

But in the case of Approach polygons, the function `processApproach()` calls `getCollisionTime()`, which can call **twice** `getPointsInside()`. This means we get to go through the "for" loop on the entire map twice. To avoid this, I have simply moved this "for" loop to `getCollisionTime()` function. This way, 1 - we make sure to go through the map only once for each polygon, and 2 - it avoids useless computation (i.e. points frame transformation made in `getCollisionTime()`) when the sources are actually not used for current polygon. 

To do so, I have kept the already existing `getPointsInside()` function, which takes a vector of points as argument. This function is called by `getCollisionTime` the same way as it was before. But I have also overloaded this `getPointsInside()`  function to be able to take the entire map as argument, which is more convenient to call from function `processStopSlowdownLimit()`, in the cas of the Stop/Slowdown/Limit polygons.

SteveMacenski:
It does not appear to be building properly, but I love this idea and looks good to me!

SteveMacenski:
Many of the collision monitor tests are now failing!

SteveMacenski:
Hi @anaelle-sw how can I help push this along?

SteveMacenski:
@anaelle-sw Please rebase after I merge https://github.com/ros-navigation/navigation2/pull/4642 to fix the build issue

anaelle-sw:
Hi @SteveMacenski , sorry it took me so long to work on this and fix the tests. It should be good to be reviewed now, the branch is rebased, and DCO, linting and  collision monitor tests are fixed.

I also add 3 new tests for polygons:
* if no specific sources are declared for polygon, all observation sources should be used
* if not declared source is associated to polygon, configuration should fail
* if  specific sources are declared for polygon, only these observation sources will be used

And a test for collision monitor node:
Two polygons are declared, with the same size. The first one uses all observation sources (range + scan) and has a "slowdown" action type. The second one uses only the range source and has a "stop" action type.
If the range source publishes an obstacle far away from the polygons, while the scan source publishes an obstacle inside the polygons, only the "slowdown" action type coming from the first polygon should be applied.

nav2_collision_monitor/src/collision_detector_node.cpp
```diff
@@ -168,17 +168,17 @@ bool CollisionDetector::getParameters()
...
-  }
-
   if (!configureSources(
       base_frame_id, odom_frame_id, transform_tolerance, source_timeout,
       base_shift_correction))
   {
     return false;
   }
 
+  if (!configurePolygons(base_frame_id, transform_tolerance)) {
```
SteveMacenski, line 178: Question: Why reordered?

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -412,17 +412,20 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in, const std_msgs::msg:
...
   // By default - there is no action
   Action robot_action{DO_NOTHING, cmd_vel_in, ""};
   // Polygon causing robot action (if any)
   std::shared_ptr<Polygon> action_polygon;
 
-  // Fill collision_points array from different data sources
+  // Fill collision points array from different data sources
   for (std::shared_ptr<Source> source : sources_) {
+    std::vector<Point> collision_points;
+    auto iter = sources_collision_points_map.insert({source->getSourceName(), collision_points});
```
SteveMacenski, line 425:
```suggestion
    auto iter = sources_collision_points_map.insert({source->getSourceName(), std::vector<Point>()});
```

I think this would work, no? Removes intermediate variable that wouldn't need to be stored in scope then

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -434,34 +437,34 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in, const std_msgs::msg:
...
-      p.z = 0.0;
-      marker.points.push_back(p);
+    if (collision_points_marker_pub_->get_subscription_count() > 0) {
+      // visualize collision points with markers
+      auto marker_array = std::make_unique<visualization_msgs::msg::MarkerArray>();
+      visualization_msgs::msg::Marker marker;
+      marker.header.frame_id = get_parameter("base_frame_id").as_string();
+      marker.header.stamp = rclcpp::Time(0, 0);
+      marker.ns = "collision_points";
+      marker.id = 0;
```
SteveMacenski, line 448: Maybe have the ID set to the position of the source in `sources_`? That way, a new one isn't overrided by the last one & you can see which is coming from which index. Can you verify that you can see all `N`? I think right now, only the last would really appear in rviz because the previous `N-1`s would be immediately overridden within nanoseconds

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -434,34 +437,34 @@ void CollisionMonitor::process(const Velocity & cmd_vel_in, const std_msgs::msg:
...
-      p.y = point.y;
-      p.z = 0.0;
-      marker.points.push_back(p);
+    if (collision_points_marker_pub_->get_subscription_count() > 0) {
+      // visualize collision points with markers
+      auto marker_array = std::make_unique<visualization_msgs::msg::MarkerArray>();
+      visualization_msgs::msg::Marker marker;
+      marker.header.frame_id = get_parameter("base_frame_id").as_string();
+      marker.header.stamp = rclcpp::Time(0, 0);
+      marker.ns = "collision_points";
```
SteveMacenski, line 447: Or perhaps have the `ns` be `"collision_points_" + source->getName()` so that they're namespaced separately? I forget how rviz handles `ns` versus `id` field for visualizing multiple messages

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -229,19 +234,55 @@ int Polygon::getPointsInside(const std::vector<Point> & points) const
...
 }
 
+int Polygon::getPointsInside(
+  const std::unordered_map<std::string,
+  std::vector<Point>> & sources_collision_points_map) const
+{
+  int num = 0;
+  std::vector<std::string> polygon_sources_names = getSourcesNames();
+
+  for (const auto & iter : sources_collision_points_map) {
```
SteveMacenski, line 244: If you have some map of all potential sources, this is either larger than or equal to the size of `getSourcesNames()`. So, it is more efficient to search through each `getSourcesNames()` rather than the map, since that'll be less than or equal to the same size.

So, this could be rewritten more efficiently and concisely as 

```
for (const auto & source_name : polygon_sources_names) {
  if (const auto & collision_points = sources_collision_points_map.find(source_name) != sources_collision_points_map.end()) {
    num += getPointsInside(collision_points);
  }
}
```

 as well below in `getCollisionTime()`

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -392,6 +433,30 @@ bool Polygon::getCommonParameters(
           node->get_parameter(polygon_name_ + ".footprint_topic").as_string();
       }
     }
+
+    // By default, use all observation sources for polygon
+    const std::vector<std::string> default_observation_sources;
+    nav2_util::declare_parameter_if_not_declared(
+      node, "observation_sources", rclcpp::ParameterValue(default_observation_sources));
```
SteveMacenski, line 440: By the time this is called, shouldn't we know that observiation_sources is already declared? We could be able to get just `get_parameter` it, no?

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -392,6 +433,30 @@ bool Polygon::getCommonParameters(
...
+    // Check the observation sources configured for polygon are defined
+    for (auto source_name : sources_names_) {
+      if (std::find(observation_sources.begin(), observation_sources.end(), source_name) ==
+        observation_sources.end())
+      {
+        RCLCPP_ERROR_STREAM(
+          logger_,
+          "Observation source [" << source_name <<
+            "] configured for polygon [" << getName() <<
+            "] is not defined!");
```
SteveMacenski, line 456:
```suggestion
            "] is not defined as one of the node's observation_source!");
```

SteveMacenski (changes requested):
Small nitpick, but overall looks good to me!

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: ^ mostly just for clarify for users about how to fix

nav2_collision_monitor/src/collision_monitor_node.cpp
anaelle-sw: You're right, it's updated

nav2_collision_monitor/src/collision_monitor_node.cpp
anaelle-sw: The behavior in rviz wasn't good with any of the solutions (using `ns` or `id`). The makers seemed to be correctly published but they were displayed with a very slow rate in rviz. The delay between the sensors actual data display and the markers update was something like a 10sec... I think publishing a new message for each source on a single topic cannot work and is not a nice design.

A solution is to actually use the `visualization_msgs::msg::MarkerArray` type. If we add a marker to the `MarkerArray` message for each source, and publish this array only once when calling `process()`, markers display works great in rviz, with good update rate. Plus, if we add the source's name in `marker.ns`, we can easily select the sources we actually want to check in rviz "Display" panel.

5_sources_rviz

Here is an example for 5 sources (a scan in dark blue and 4 ranges). I only added the different colors for this test, this change is not on PR branch.

nav2_collision_monitor/src/polygon.cpp
anaelle-sw: You're right, it's updated for both `getPointsInside()` and `getCollisionTime()` functions

nav2_collision_monitor/src/polygon.cpp
anaelle-sw: Yes, the parameter `observation_sources` is already declared when node is getting polygons' parameters.

The problem is the tests. In `polygons_test.cpp`, we don't use any collision monitor node but only the polygons. So not declaring parameter `observation_sources` here with a default value results in many tests failures. 

If we consider that the polygon as an independent structure, so that its methods `configure()` , `getParameters()` and `getCommonParameter()` can be successfully called without needing any collision monitor node, then we need to keep this line. Since the polygons are tested independently, I though this was the more reasonable.

But if we don't want to risk any double declaration of param `observation_sources`, we can remove this line from `polygon.cpp` and add it to  `polygons_test.cpp` so that the parameter will be declared first, only for these tests

nav2_collision_monitor/src/polygon.cpp
anaelle-sw: Done :)

nav2_collision_monitor/src/polygon.cpp
SteveMacenski: I don't mind its re-declaration, however the original is a required input (i.e. only specifies type, no default value, it must be specified) 

```
      node, "observation_sources", rclcpp::PARAMETER_STRING_ARRAY);
```

Here, you say the default value is none, which is not the same expression. We'd need them to be the same if we're going to re-declare in case this module is used independently so that they act the same. If you do that, then you'd still need to declare or set the value of `observation_sources` in the tests before running the polygon work so that it could `get_parameter` something without throwing the unset-parameter exception. 

So, I think you can declare again, but it should be  `PARAMETER_STRING_ARRAY`, and then the tests need to still declare or set it to some value

SteveMacenski:
Otherwise, LGTM, but you have a test failing so I can't see the coverage metrics to see if this covers the bulk of the important changes

nav2_collision_monitor/src/collision_monitor_node.cpp
anaelle-sw: The solution I told you about is already commit on this PR branch. So currently this is the structure:
```c++
process()  {

 ... // some init

  auto marker_array = std::make_unique<visualization_msgs::msg::MarkerArray>();
  for (std::shared_ptr<Source> source : sources_) {

   ... // checking source is valid
   

    if (collision_points_marker_pub_->get_subscription_count() > 0) {
      visualization_msgs::msg::Marker marker;
      marker.ns = "collision_points_" + source->getSourceName();

      ... // filling marker data for current source
      
      marker_array->markers.push_back(marker);
    } // end of check on marker topic subscription count
  } // end of source loop

  // publishing one marker array message containing all sources (on marker in array per source)
  if (collision_points_marker_pub_->get_subscription_count() > 0) {
    collision_points_marker_pub_->publish(std::move(marker_array));
  }

  ... // checking polygons and actions, publishing velocity and polygons
} // end of process function
```

It allows the results on the rviz screenshot with normal update rate, but the color of all markers in array is red

nav2_collision_monitor/src/polygon.cpp
anaelle-sw: >  declare again, but it should be PARAMETER_STRING_ARRAY, and then the tests need to still declare or set it to some value

It's done

SteveMacenski (approved):

SteveMacenski:
LGTM! Any reason this is still a draft? I could merge this in I think!

anaelle-sw:
> Any reason this is still a draft? 

No, I just forgot to update it. It should be good now

PR #4231, opened 2024-04-01
replace layered_costmap_->isRolling() with map_frame_ == global_frame_ in static layer

alanxuefei:
The costmap frame can be directly compared with the static layer frame, instead of inferring their relationship based on `isRolling()`.

Typically, the static layer always utilizes the `map` frame. The global costmap (isRolling == true) uses the `map` frame, whereas the local costmap (isRolling == false) employs the `odom` frame. For a local costmap, the map cells are transformed from the map frame to the odom frame.

Rather than using `isRolling` to distinguish between the global and local costmaps, we can directly check if `map_frame_` equals `global_frame_`. This not only enhances readability but also permits the static layer in the global costmap to use a different frame from the costmap. For instance, the static layer frame could be `map_docker`, while the global costmap continues to use `map`. The transformation would be from `map_docker` to `map`.

 

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/static_layer.cpp
+++ b/nav2_costmap_2d/plugins/static_layer.cpp
@@ -387,15 +387,15 @@ StaticLayer::updateCosts(
     return;
   }
 
-  if (!layered_costmap_->isRolling()) {
-    // if not rolling, the layered costmap (master_grid) has same coordinates as this layer
+  if (map_frame_ == global_frame_) {
+    // When map_frame_ equals global_frame_, the layered costmap (master_grid) has same coordinates as this layer
     if (!use_maximum_) {
       updateWithTrueOverwrite(master_grid, min_i, min_j, max_i, max_j);
     } else {
       updateWithMax(master_grid, min_i, min_j, max_i, max_j);
     }
   } else {
-    // If rolling window, the master_grid is unlikely to have same coordinates as this layer
+    // When map_frame_ does not equal global_frame_, the coordinates in this layer must be transformed to align with those in the master_grid.
     unsigned int mx, my;
     double wx, wy;
     // Might even be in a different frame


SteveMacenski:
This is an incorrect assumption. You can and do have situations that you have a rolling cost map in the map frame or other global static frame.  For example, for rolling window global planning in hyper massive spaces.

alanxuefei:
If it is not rolling, it is assumed that map_frame_ is equal to global_frame_. 
If it is rolling, it is assumed that map_frame_ is equal to global_frame_. 
Why don't we directly compare map_frame_ to global_frame_ to see if they are equal.
 ```
 if (!layered_costmap_->isRolling()) {  // change to map_frame_ == global_frame_
    // if not rolling, the layered costmap (master_grid) has same coordinates as this layer
    if (!use_maximum_) {
      updateWithTrueOverwrite(master_grid, min_i, min_j, max_i, max_j);
    } else {
      updateWithMax(master_grid, min_i, min_j, max_i, max_j);
    }
  } else {
    // If rolling window, the master_grid is unlikely to have same coordinates as this layer
    unsigned int mx, my;
    double wx, wy;
    // Might even be in a different frame
    geometry_msgs::msg::TransformStamped transform;
    try {
      transform = tf_->lookupTransform(
        map_frame_, global_frame_, tf2::TimePointZero,
        transform_tolerance_);
    } catch (tf2::TransformException & ex) {
      RCLCPP_ERROR(logger_, "StaticLayer: %s", ex.what());
      return;
    }
    // Copy map data given proper transformations
    tf2::Transform tf2_transform;
    tf2::fromMsg(transform.transform, tf2_transform);

    for (int i = min_i; i < max_i; ++i) {
      for (int j = min_j; j < max_j; ++j) {
        // Convert master_grid coordinates (i,j) into global_frame_(wx,wy) coordinates
        layered_costmap_->getCostmap()->mapToWorld(i, j, wx, wy);
        // Transform from global_frame_ to map_frame_
        tf2::Vector3 p(wx, wy, 0);
        p = tf2_transform * p;
        // Set master_grid with cell from map
        if (worldToMap(p.x(), p.y(), mx, my)) {
          if (!use_maximum_) {
            master_grid.setCost(i, j, getCost(mx, my));
          } else {
            master_grid.setCost(i, j, std::max(getCost(mx, my), master_grid.getCost(i, j)));
          }
        }
      }
    }
  }
  current_ = true;
}
```

**Comment by alanxuefei** [comment]
I found one reason to use boolean instead of string comparison. 
Performing a string comparison (which is what comparing frame IDs would entail) every cycle could be less efficient than checking a boolean state.

However, we can compare two frames during initialization.

**Comment by alanxuefei** [comment]
@SteveMacenski 
Thanks for sharing ideas about the rolling global costmap.

I will build a global costmap rolling over massive static map. Both are in map frame. 
In this case, each cell will run 0 transform which is unnecessary. 
Once the demo is done, I will update this PR,

**Comment by SteveMacenski** [comment]
What happens when the global frame is a rolling frame and the map's frame is also a rolling frame?

**Comment by alanxuefei** [comment]
If the map frame (layer) is equal to the global frame (costmap), (let's say odom or map), the transform from global_frame_ to map_frame_ is unnecessary.  This could lead to CPU savings.

```
 if (!layered_costmap_->isRolling()) {  // propose to use map_frame_ == global_frame_
   // if not rolling, the layered costmap (master_grid) has same coordinates as this layer
   if (!use_maximum_) {
     updateWithTrueOverwrite(master_grid, min_i, min_j, max_i, max_j);
   } else {
     updateWithMax(master_grid, min_i, min_j, max_i, max_j);
   }
 } else {
   // If rolling window, the master_grid is unlikely to have same coordinates as this layer
   unsigned int mx, my;
   double wx, wy;
   // Might even be in a different frame
   geometry_msgs::msg::TransformStamped transform;
   try {
     transform = tf_->lookupTransform(                                      // the transform codes
       map_frame_, global_frame_, tf2::TimePointZero,
       transform_tolerance_);
   } catch (tf2::TransformException & ex) {
     RCLCPP_ERROR(logger_, "StaticLayer: %s", ex.what());
     return;
   }
   // Copy map data given proper transformations
   tf2::Transform tf2_transform;
   tf2::fromMsg(transform.transform, tf2_transform);

   for (int i = min_i; i < max_i; ++i) {
     for (int j = min_j; j < max_j; ++j) {
       // Convert master_grid coordinates (i,j) into global_frame_(wx,wy) coordinates
       layered_costmap_->getCostmap()->mapToWorld(i, j, wx, wy);
       // Transform from global_frame_ to map_frame_
       tf2::Vector3 p(wx, wy, 0);
       p = tf2_transform * p;
       // Set master_grid with cell from map
       if (worldToMap(p.x(), p.y(), mx, my)) {
         if (!use_maximum_) {
           master_grid.setCost(i, j, getCost(mx, my));
         } else {
           master_grid.setCost(i, j, std::max(getCost(mx, my), master_grid.getCost(i, j)));
         }
       }
     }
   }
 }
 current_ = true;
}
```

**Comment by SteveMacenski** [comment]
There are real use-cases when both frames might not be static and do require this kind of transform when rolling. I don't think this is wise to do without thinking through all of the potential edge cases and benchmarking that it makes a manifest difference in performance to justify the complexity it would introduce to understandability

**Comment by alanxuefei** [comment]
Let's look at this another way: When should we use the code below? The condition is if the two frames are different, even if they are not static.

```
transform = tf_->lookupTransform(                                     
       map_frame_, global_frame_, tf2::TimePointZero,
       transform_tolerance_);
```

I agree that the static layer has been functioning well for years. It's better not to alter code that is already working.

SteveMacenski:
When the map might (1) not be aligned perfectly with the previous iteration, (2) the current map due to moving reference frames, (3) the user is specifically stating that this costmap should roll within a fixed window regardless which might have aliasing due to different orientations w.r.t. the staticly set map due to yaw rotations.

PR #4254, opened 2024-04-10
Backport #4136 to iron: Add API to gracefully cancel a controller

Timple:
Backport of #4136 to iron.

Feel free to close if this is too much of an api/abi break.

Only the implementation is backported, not the pure pursuit modifications as they modify behavior.

---

#### For Maintainers: 


SteveMacenski:
Unfortunately this breaks API stability :( But otherwise I would have loved to, this is a really useful feature. Thankfully, Jazzy's only a month away!

Issue #4249, opened 2024-04-09
[SMAC planner] Start and goal are moved to closest cell


BriceRenaudeau:
## Bug report

When creating a plan, the start and the goal are moved to the middle of the closest cell.

![image](https://github.com/ros-planning/navigation2/assets/48433002/f07b53d9-a1c2-41b7-b512-fe4ab0a89a88)
- start = base_footprint
- goal = red arrow
- grid = 5cm/5cm = costmap resolution

**Required Info:**

- Operating System:
  - Ubuntu
- ROS2 Version:
  - Iron -->
- Version or commit hash:
  - source
- DDS implementation:
  - cyclon

#### Steps to reproduce issue
Plan a path not in the middle of a cell.

#### Expected behavior
The start and the goal are at the requested place.

#### Actual behavior
The start and the goal are moved to the middle of the closest cell

#### Additional information

This is due to the process used to add start and goal poses:
(smac_planner_hybrid.cpp)
``` c++
unsigned int mx, my;
costmap->worldToMap(start.pose.position.x, start.pose.position.y, mx, my)
_a_star->setStart(mx, my, orientation_bin_id);
```
But `setStart` turns it back to float:
https://github.com/ros-planning/navigation2/blob/f815e8bd8e3826603e97749bb31a3145268de023/nav2_smac_planner/src/a_star.cpp#L145-L157

We can just compute the `worldToMap` manually to keep it as a float and modify `setStart` input types.

BriceRenaudeau:
A quick test and it works:
![image](https://github.com/ros-planning/navigation2/assets/48433002/21d4e8b0-ac6e-443a-905a-40e4101c07ea)

SteveMacenski:
This is definitely true and is detailed in this ticket: https://github.com/ros-planning/navigation2/issues/3955

I even highlight a solution in this comment: https://github.com/ros-planning/navigation2/issues/3955#issuecomment-1808461774 if but it seem that you're already so inclined and implemented it. Open a PR and you can quash another ticket too :-)

BriceRenaudeau:
I saw that issue last year but the title confused me and I didn't follow your reply.
I have some time to make a PR.
I will go for what you described;
> * Create a `worldToMapContinous` to compute the decimals for the world to map conversion.
> * Change the `setGoal` to use floats as inputs and `static_cast` to unsigned int's for the `getIndex` method.
> * Do the same with setStart as well

Issue #3955, opened 2023-11-10
SmacPlannerHybrid does not respect tolerance


aniket11bh:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - humble
- Version or commit hash:
  - ros-humble-navigation2	1.1.12-1jammy.20231005.224358
- DDS implementation:
  - cycloneDDS

#### Steps to reproduce issue
- Give a set of poses, and use smac planner hybrid for generating the path.
- In the attached figure, the grid size (resolution) is 1m. 

#### Expected behavior
SmacPlannerHybrid should generate path within tolerance, if it is unable to generate within tolerance it should fail.

#### Actual behavior
Planner is generating paths, but it not generating paths within given tolerance.

#### Additional information
```yaml
planner_server:
  ros__parameters:
    planner_plugins: ["GridBased"]
    use_sim_time: True

    GridBased:
      plugin: "nav2_smac_planner/SmacPlannerHybrid"
      tolerance: 0.01                      # tolerance for planning if unable to reach exact pose, in meters
      downsample_costmap: false           # whether or not to downsample the map
      downsampling_factor: 1              # multiplier for the resolution of the costmap layer (e.g. 2 on a 5cm costmap would be 10cm)
      allow_unknown: false                # allow traveling in unknown space
      max_iterations: 1000000             # maximum total iterations to search for before failing (in case unreachable), set to -1 to disable
      max_on_approach_iterations: 1000    # maximum number of iterations to attempt to reach goal once in tolerance
      max_planning_time: 5.0              # max time in s for planner to plan, smooth, and upsample. Will scale maximum smoothing and upsampling times based on remaining time after planning.
      motion_model_for_search: "DUBIN"    # For Hybrid Dubin, Redds-Shepp

      minimum_turning_radius: 10.0        # For Hybrid/Lattice nodes: minimum turning radius in m of path / vehicle
      analytic_expansion_max_length: 30.0    # For Hybrid/Lattice nodes: The maximum length of the analytic expansion to be considered valid to prevent unsafe shortcutting (in meters). This should be scaled with minimum turning radius and be no less than 4-5x the minimum radius
      analytic_expansion_ratio: 3.5       # For Hybrid/Lattice nodes: The ratio to attempt analytic expansions during search for final approach.
      angle_quantization_bins: 72         # For Hybrid nodes: Number of angle bins for search, must be 1 for 2D node (no angle search)
[...]

PR #4255, opened 2024-04-10
Set start and goal as float

BriceRenaudeau:
* Create a `worldToMapContinous` to compute the decimals for the world to map conversion.
* Change the `setGoal` to use floats as inputs and `static_cast` to unsigned int's for the `getIndex` method.
* Do the same with setStart as well

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3955 and #4249 |
| Robotic platform tested on | sim |

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d.cpp
+++ b/nav2_costmap_2d/src/costmap_2d.cpp
@@ -302,8 +302,8 @@ bool Costmap2D::worldToMapContinuous(double wx, double wy, float & mx, float & m
     return false;
   }
 
-  mx = (wx - origin_x_) / resolution_ + 0.5;
-  my = (wy - origin_y_) / resolution_ + 0.5;
+  mx =  static_cast<float>((wx - origin_x_) / resolution_) + 0.5f;
+  my =  static_cast<float>((wy - origin_y_) / resolution_) + 0.5f;
 
   if (mx < size_x_ && my < size_y_) {
     return true;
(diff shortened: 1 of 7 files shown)


BriceRenaudeau:
CI failure not related.

nav2_costmap_2d/src/costmap_2d.cpp
```diff
@@ -296,6 +296,21 @@ bool Costmap2D::worldToMap(double wx, double wy, unsigned int & mx, unsigned int
   return false;
 }
 
+bool Costmap2D::worldToMapContinuous(double wx, double wy, float & mx, float & my) const
+{
+  if (wx < origin_x_ || wy < origin_y_) {
+    return false;
+  }
+
+  mx = (wx - origin_x_) / resolution_ + 0.5;
```
SteveMacenski, line 305: Cast the origin values to float before running this operation. I've been burned alot recently in MPPI/Smac with mixing types around and getting unexpected outcomes or massive performance hits

SteveMacenski: `+ 0.5f`

SteveMacenski:
@BriceRenaudeau the first 2 failures seem related to this PR. I've never before seen those tests fail and they're targeting Smac

BriceRenaudeau:
> @BriceRenaudeau the first 2 failures seem related to this PR. I've never before seen those tests fail and they're targeting Smac

Le planner returns: "Start occupied" on the verySmallPath tests.

nav2_costmap_2d/src/costmap_2d.cpp
BriceRenaudeau: Wz, origin_x and resolution are double, I made the static cast on the result.

SteveMacenski:
Wonderful!

PR #4258, opened 2024-04-11
Add expanding the ~/ to the full home dir of user in the path to the map yaml. 

Wiktor-99:
Add expanding the **~/** to the full home dir of user in the path to the map yaml. 

- Add implementation.
- Add unit tests.


SteveMacenski (approved):
Please fix the linter errors and otherwise LGTM

PR #4267, opened 2024-04-18
map_server: update to yaml-cpp 0.8.0

981213:
yaml-cpp 0.8.0 requires explicit linking against yaml-cpp::yaml-cpp.
This fixes building breakage on rolling due to https://github.com/ros2/yaml_cpp_vendor/pull/48

---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Not yet. It shouldn't affect anything pratically. |

---

* Fix building on rolling due to recent yaml-cpp upgrade

None.

---

None.

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/CMakeLists.txt
+++ b/nav2_map_server/CMakeLists.txt
@@ -11,6 +11,7 @@ find_package(rclcpp_components REQUIRED)
 find_package(nav_msgs REQUIRED)
 find_package(nav2_msgs REQUIRED)
 find_package(yaml_cpp_vendor REQUIRED)
+find_package(yaml-cpp REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(tf2 REQUIRED)
 find_package(nav2_util REQUIRED)
@@ -118,7 +119,8 @@ target_include_directories(${map_io_library_name} SYSTEM PRIVATE
   ${GRAPHICSMAGICKCPP_INCLUDE_DIRS})
 
 target_link_libraries(${map_io_library_name}
-  ${GRAPHICSMAGICKCPP_LIBRARIES})
+  ${GRAPHICSMAGICKCPP_LIBRARIES}
+  yaml-cpp::yaml-cpp)
 
 if(WIN32)
   target_compile_definitions(${map_io_library_name} PRIVATE
@@ -160,5 +162,5 @@ ament_export_libraries(
   ${library_name}
   ${map_io_library_name}
 )
-ament_export_dependencies(${map_io_dependencies} ${map_server_dependencies})
+ament_export_dependencies(${map_io_dependencies} ${map_server_dependencies} yaml-cpp)
 ament_package()


981213:
The CI failed to build because it doesn't have the latest yaml_cpp_vendor package.

SteveMacenski:
@981213 should this PR wait then until that lands?

nav2_map_server/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ find_package(rclcpp_components REQUIRED)
 find_package(nav_msgs REQUIRED)
 find_package(nav2_msgs REQUIRED)
 find_package(yaml_cpp_vendor REQUIRED)
+find_package(yaml-cpp REQUIRED)
```
SteveMacenski, line 14: This should not be necessary, this is the role of the vendor package.

nav2_map_server/CMakeLists.txt
```diff
@@ -118,7 +119,8 @@ target_include_directories(${map_io_library_name} SYSTEM PRIVATE
   ${GRAPHICSMAGICKCPP_INCLUDE_DIRS})
 
 target_link_libraries(${map_io_library_name}
-  ${GRAPHICSMAGICKCPP_LIBRARIES})
+  ${GRAPHICSMAGICKCPP_LIBRARIES}
+  yaml-cpp::yaml-cpp)
```
SteveMacenski, line 123: Ditto and below

SteveMacenski (changes requested):
Please follow up, I don't think this is an appropriate PR since we have the vendor. If the vendor needs to be updated, then update that and it'll be reflected here after a sync. Why are these changes needed here instead?

981213:
> should this PR wait then until that lands?

yaml_cpp_vendor 9.0.0 has been released into rolling, and that's how I noticed this problem. This is caused by the _vendor package update and independent of OS.

>  Why are these changes needed here instead - that breaks the vendorization?

This PR is created according to
https://github.com/ros2/yaml_cpp_vendor/pull/48#issuecomment-2046048236
and following:
https://github.com/ros-perception/image_common/pull/305
https://github.com/ros2/rviz/pull/1183
https://github.com/ros2/rosbag2/pull/1605

The solution provided upstream is to change all these downstream.

981213:
I agree that this change is a bit strange. Let me ask in that yaml-cpp update PR.

SteveMacenski (approved):

SteveMacenski:
Build has a number of complaints along the lines of 

```
CMake Error at CMakeLists.txt:32 (add_executable):
  Target "map_server" links to target "yaml-cpp::yaml-cpp" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

```

981213:
Oh, yaml_cpp_vendor 9.0 isn't available as binary package for jammy:
```
sudo apt-cache madison ros-rolling-yaml-cpp-vendor
ros-rolling-yaml-cpp-vendor | 8.3.1-1jammy.20240124.152723 | http://mirrors.ustc.edu.cn/ros2/ubuntu jammy/main amd64 Packages
```
```
sudo apt-cache madison ros-rolling-yaml-cpp-vendor
ros-rolling-yaml-cpp-vendor | 9.0.0-1noble.20240416.181051 | http://packages.ros.org/ros2/ubuntu noble/main amd64 Packages
```
This PR should be blocked until the switching to noble then.

SteveMacenski:
OK! I'll tag it that way with the 24.04 PR, thanks for the help and proactivity!

SteveMacenski:
Note to self that this is also a problem with Robot Localization to fix

tonynajjar:
I ran into the same issue and sunk 2 hours to fix it in a similar way,if only I found this PR beforehand :sweat_smile: Anyway, I think this should now be ready to be merged

SteveMacenski:
https://app.circleci.com/pipelines/github/ros-planning/navigation2/11466/workflows/31511139-b085-4f4b-ad71-5573452fa375/jobs/35332/steps

CI tells me this shouldn't happen until the 24.04 migration

tonynajjar:
Ah I tried it with the rolling docker images and there it was needed to work. I guess CI is not using latest rolling yet

SteveMacenski:
@981213 thanks for the contribution here - @tonynajjar is working on the full 24.04 migration which this commit was added to in https://github.com/ros-navigation/navigation2/pull/4298. So, I'm closing this as that PR supersedes with some other necessary changes. I want to thank you for your help here and responses! It is greatly appreciated - it all takes a village :-)

PR #4272, opened 2024-04-22
[Humble] Fixing 4230

SteveMacenski:


corot:
Does the fix have any impact on behavior?  Cannot identify any

SteveMacenski:
Yes, it restarts to iterate from the start of each trajectory in the batch. Before it started at the furthest from the last sample which is bad since each trajectory candidate should be scored independently.

PR #4275, opened 2024-04-24
Collision monitor: Add polygon source

SteveMacenski:
Continuation of https://github.com/ros-planning/navigation2/pull/3885


nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,200 @@
...
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+  auto curr_time = node->now();
+
+  // check if older similar polygon exists already and replace it with the new one
+  for (auto & polygon_stamped : data_) {
+    if (msg->header.frame_id == polygon_stamped.header.frame_id &&
+      rclcpp::Time(msg->header.stamp) >= rclcpp::Time(polygon_stamped.header.stamp) &&
+      areSamePolygons(*msg, polygon_stamped))
```
tonynajjar, line 184: I think the frame_ids don't necessarily have to be the same for the polygons to be considered the same and hence be replaced. It would be weird but if a user decides to start publishing the polygon with the same id with a different frame_id... why not?

For the time we could maybe apply the same reasoning but it's a bit more delicate because it's used for removing the polygons by timeout.

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  getCommonParameters(source_topic);
+
+  nav2_util::declare_parameter_if_not_declared(
+    node, source_name_ + ".sampling_distance", rclcpp::ParameterValue(0.1));
+  sampling_distance_ = node->get_parameter(source_name_ + ".sampling_distance").as_double();
+  nav2_util::declare_parameter_if_not_declared(
+    node, source_name_ + ".polygon_similarity_threshold", rclcpp::ParameterValue(2.0));
+  polygon_similarity_threshold_ =
+    node->get_parameter(source_name_ + ".polygon_similarity_threshold").as_double();
```
tonynajjar, line 169: polygon_similarity_threshold is not used anymore

nav2_collision_monitor/src/polygon_source.cpp
```diff
@@ -0,0 +1,200 @@
...
+    tf2::doTransform(polygon_stamped, poly_out, tf);
+    convertPolygonStampedToPoints(poly_out, data);
+  }
+  return true;
+}
+
+void PolygonSource::convertPolygonStampedToPoints(
+  const geometry_msgs::msg::PolygonStamped & polygon, std::vector<Point> & data) const
+{
+  /*  Full function generated by GPT */
```
tonynajjar, line 124: I don't know what you want to do about it but just a reminder because I know you feel a certain way about AI-generated code

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: Fixed

nav2_collision_monitor/src/polygon_source.cpp
SteveMacenski: Fixed

Issue #4277, opened 2024-04-25
footprint_clearing_enabled for more (or all) layers


tonynajjar:
Spinoff from https://github.com/ros-planning/navigation2/issues/4271#issuecomment-2073467118

#### Feature description
`footprint_clearing_enabled` is only a parameter of the obstacle layer. it's also in the voxel layer but doesn't seem to be used. I think it would make sense if this parameter was also implemented for the static layer. Going further, we can also argue for it to be at top-level costmap such that it applies to all layers/

SteveMacenski:
>  it's also in the voxel layer but doesn't seem to be used

The voxel layer derives from the obstacle layer and they share the same `updateFootprint` function which uses the parameter, so the voxel layer should work :-) Do you see it not working or is that based on code review?

Since its a param, I have no issue with it being applied to the static layer as well! :smile: 

Please submit a PR and that should be easy to merge

tonynajjar:
> Do you see it not working or is that based on code review?

I'm wrong then, was purely code review (after I found 'clearing_layers' not used so I assumed there is more 😬)

> I have no issue with it being applied to the static layer as well

What do you think about a param to apply it to the whole costmap instead of by layer? I haven't checked if that's easily feasible but I assume so

SteveMacenski:
I think we should leave it on a layer by layer basis for granularity of controls. I suppose we could offer both though like we have for some settings. That would be fine!

PR #4282, opened 2024-04-26
Add footprint clearing for static layer

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-planning/navigation2/issues/4277 |
| Robotic platform tested on | custom gazebo simulation |

---

* I added footprint clearing to static layer similarly to obstacle_layer

image
image

* Document new parameter

---

* Adding parameter to more layers

#### For Maintainers: 


nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -430,6 +450,9 @@ StaticLayer::updateCosts(
       }
     }
   }
+  if (footprint_clearing_enabled_) {
+    setConvexPolygonCost(transformed_footprint_, nav2_costmap_2d::FREE_SPACE);
+  }
```
tonynajjar, line 455: I don't know much about the costmap code, is there a difference to placing this snippet here vs before the setting of the costs just above?

SteveMacenski: I think there is probably correct - else it could be overrided by the update

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -132,6 +132,7 @@ StaticLayer::getParameters()
   declareParameter("map_subscribe_transient_local", rclcpp::ParameterValue(true));
   declareParameter("transform_tolerance", rclcpp::ParameterValue(0.0));
   declareParameter("map_topic", rclcpp::ParameterValue(""));
+  declareParameter("footprint_clearing_enabled", rclcpp::ParameterValue(true));
```
SteveMacenski, line 135:
```suggestion
  declareParameter("footprint_clearing_enabled", rclcpp::ParameterValue(false));
```

Lets not change default behavior - but LGTM with documentation update to include param + migration guide

SteveMacenski (approved):

tonynajjar:
Okay, what do you think about adding a small padding to the footprint clearing? Maybe as a parameter as well?

SteveMacenski:
> However when I add some clearing padding it works

I think its wise to understand why that's happening and if that's the right solution. But if we know why and this is the right solution, I don't have a problem with it

tonynajjar:
I found another small bug but I can't really find the cause, maybe you have an idea. I found that the static layer occasionally flickers, i.e. without moving the footprint, the static layer sometimes "unclears". It happens about once every 10-20 seconds and lasts less than a second, if that tells you anything.

tonynajjar:
Update: I figured out where the flickering came from. It was because of the conversation we had [here](https://github.com/open-navigation/navigation2/pull/4282#discussion_r1580855166). Turns out it has to be before the chunk where the costs are updated. I don't know why for sure but it's like that in obstacle, voxel and stvl layers

SteveMacenski (approved):

PR #4284, opened 2024-04-28
[Smac Planner] Fix node index overflow for huge map

doisyg:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Dexory Simulation |

---

For huge maps and using a high value for the parameter `angle_quantization_bins`, it is possible to overflow the node index which is using a unsigned 32 bit integer:
https://github.com/ros-planning/navigation2/blob/9f6b0dd3fb61508b1e042eff6deaca3d21d99550/nav2_smac_planner/src/a_star.cpp#L290
For instance for a map of 10813x6743 pixels, and using `angle_quantization_bins: 64`, that gives us 4,666,371,776 which is above 4,294,967,295 (utin32 max).
I tried to only use 64 bit for the index, and not the map x/y size, but that is to discuss as we should then be careful of explicit casting (I got bitten). Hence the draft. And because I tested it only for the hybrid planner.

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
@@ -55,7 +55,7 @@ public:
   typedef typename NodeT::Coordinates Coordinates;
   typedef typename NodeT::CoordinateVector CoordinateVector;
   typedef typename NodeVector::iterator NeighborIterator;
-  typedef std::function<bool (const unsigned int &, NodeT * &)> NodeGetter;
+  typedef std::function<bool (const uint64_t &, NodeT * &)> NodeGetter;
 
   /**
    * @struct nav2_smac_planner::NodeComparator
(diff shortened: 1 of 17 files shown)


SteveMacenski:
Neat, that is a nice catch!

SteveMacenski:
@doisyg there are a couple of updates needed (actually surprised this works without them), but one of the biggest flags I wanted to mention is that the costmap itself has the get index functions as `unsigned int`. So, while for the Smac Planner we have the `* 72` or what have you dimensionality problem, you might also need to be aware the limit is ~10M m2 for costmap size. I don't think you're even remotely close to that, but something to keep an eye on. That would cause a myriad of issues across the stack in general, though should be fixable by the same kind of PR here

SteveMacenski:
Also see https://github.com/open-navigation/navigation2/pull/4295/commits/a6f1f9e035f7f3db5a4615e83215273a033b000f these are the lines I disagree with, they're changing conversions for handling of the costmap, not the Graph that was exceeding it. The costmap APIs will downshift these into `unsigned int` anyway per its API

Issue #4281, opened 2024-04-26
RotationShimController does not rotate when goal is closer than `forward_sampling_distance`


gennartan:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04
- ROS2 Version: Iron
- Version or commit hash: ros-iron-nav2-rotation-shim-controller	1.2.6-1jammy.20240213.190815
- DDS implementation: /

#### Steps to reproduce issue
1- Send a goal to the controller closer than `forward_sampling_distance`
2- Done, the robot does not rotate towards the goal. 

#### Expected behavior
If the path is smaller than `forward_sampling_distance`, the robot should rotate towards the last point of the path.

#### Actual behavior
If the new goal is not in front of the robot, the robot motion will be as described in [nav2  documentation](https://navigation.ros.org/tuning/index.html#rotate-in-place-behavior). _The result is an awkward, stuttering, or whipping around behavior_  often resulting in a goal failure.

#### Additional information

* I also tried to reduce the `forward_sampling_distance` to the `xy_goal_tolerance` to ensure that the goal is validated when it is too close and avoid this akward behavior. This also failed because the requested orientation when starting navigation toward a new goal was too random.
* I am ready to solve this problem by creating a PR. But I first wanted to be sure that this could be a wanted behavior of this controller.

SteveMacenski:
Thanks for the note! Can you highlight what you think we should do in this situation instead? If the goal is that close, it seems reasonable to me for the primarily controller to be immediately called instead of attempting to rotate. You could also decrease the forward sampling distance to be small, but you wouldn't want it to be super small (< 0.2m) else that angle may not represent the true intent of the path future (or if very very small, the atan2 function blowing up).

gennartan:
I was thinking of something like this: [1]. We are already using this in my current project. Without it, there are some paths that systematically fails.

As I see it, using my commit [1], there are three cases:

* ´goal_dist < xy_goal_tolerance´: the goal will be validated by the goal checker if in the correct orientation. If not in the correct orientation, the robot will first rotate towards the goal, then rotate to the goal heading. This could be a bit akward since it would perform 2 different rotations. (Maybe later, a new feature could rotate the robot directly toward the goal heading when the goal is closer than the goal checker tolerance).
[...]

PR #4290, opened 2024-05-01
[RotationShimController] rotate also on short paths

gennartan:
When the path is shorter than ´forward_sampling_distance´ the rotatitonShimController would directly give control to the primary controller to navigate to the goal. This would lead to the exact behavior that was tried to be fixed by the rotationShimController: "The result is an awkward, stuttering, or whipping around behavior" [1]. It often resulted in navigation failure.

In this case, the controller should try to rotate towards the last point of the path, so that the primary controller can have a better starting orientation.

[1] https://navigation.ros.org/tuning/index.html#rotate-in-place-behavior

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4281  |
| Primary OS tested on | Ubuntu 20.04 / 22.04 |
| Robotic platform tested on | Gazebo simulation - custom hardware |

---

* Let the rotationShimController rotate the robot when the path length is shorter than ´forward_sampling_distance´ (this allow the robot to have the correct orientation when the requested goal is really close)

* Need to update the ´forward_sampling_distance´ description of https://navigation.ros.org/tutorials/docs/using_shim_controller.html --> add "if there is no point further than forward_sampling_distance, rotates toward the goal"
* Same gos for https://navigation.ros.org/configuration/packages/configuring-rotation-shim-controller.html

#### For Maintainers: 

### DIFF
--- a/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
+++ b/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
@@ -195,10 +195,7 @@ geometry_msgs::msg::PoseStamped RotationShimController::getSampledPathPt()
     }
   }
 
-  throw nav2_core::ControllerException(
-          std::string(
-            "Unable to find a sampling point at least %0.2f from the robot,"
-            "passing off to primary controller plugin.", forward_sampling_distance_));
+  return current_path_.poses.back();
 }
 
 geometry_msgs::msg::Pose


nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -195,10 +195,7 @@ geometry_msgs::msg::PoseStamped RotationShimController::getSampledPathPt()
     }
   }
 
-  throw nav2_core::ControllerException(
-          std::string(
-            "Unable to find a sampling point at least %0.2f from the robot,"
-            "passing off to primary controller plugin.", forward_sampling_distance_));
+  return current_path_.poses[current_path_.poses.size() - 1];
```
SteveMacenski, line 198:
```suggestion
  return current_path_.poses.back();
```

SteveMacenski (changes requested):
Would it be sensible in L179 to consider increasing the minimum size of path to 3 or 4?

gennartan:
> Would it be sensible in L179 to consider increasing the minimum size of path to 3 or 4?

I think it makes sense to keep to 2.
If the path resolution is low, two points can still be far from each other in a path.  As an example, my current project was using a path resolution of 0.2m. And we don't want the path resolution to change the behavior of this node.

SteveMacenski:
Sorry about the toggled CI, I was testing some changes to the CI bots for the org change. Once CI comes back happy I can merge

SteveMacenski:
@gennartan, a bug was identified by @NikolasE  in https://github.com/ros-navigation/navigation2/commit/94d3a9d8610d367088ad48fa066d378f31844601

```
double angular_distance_to_heading = std::abs(goal_yaw - pose_yaw);
```

---

I think this will fail if goal_yaw is -pi and pose_yaw pi. Robot is already aligned, but I assume this could trigger a full rotation on the spot.

This could help:
https://github.com/ros/angles/blob/54ad72b0ce473a2e1d98da447898bf2fcd36519f/angles/include/angles/angles.h#L104C72-L104C73

PR #4293, opened 2024-05-03
ruff rule F601 - Dictionary key literal `'use_composition'` repeated

cclauss:
% `ruff check`
```
Error: nav2_system_tests/src/system/test_wrong_init_pose_launch.py:117:21: F601 Dictionary key literal `'use_composition'` repeated
```
% ` ruff rule F601`
# multi-value-repeated-key-literal (F601)

Derived from the **Pyflakes** linter.

Fix is sometimes available.

## What it does
Checks for dictionary literals that associate multiple values with the same key.

## Why is this bad?
Dictionary keys should be unique. If a key is associated with multiple values, the earlier values will be overwritten. Including multiple values for the same key in a dictionary literal is likely a mistake.

## Example
```python
foo = {
    "bar": 1,
    "baz": 2,
    "baz": 3,
}
foo["baz"]  # 3
```

Use instead:
```python
foo = {
    "bar": 1,
    "baz": 2,
}
foo["baz"]  # 2
```

## References
- [Python documentation: Dictionaries](https://docs.python.org/3/tutorial/datastructures.html#dictionaries)

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_system_tests/src/system/test_wrong_init_pose_launch.py
+++ b/nav2_system_tests/src/system/test_wrong_init_pose_launch.py
@@ -114,7 +114,6 @@ def generate_launch_description():
                     'bt_xml_file': bt_navigator_xml,
                     'use_composition': 'False',
                     'autostart': 'True',
-                    'use_composition': 'False',
                 }.items(),
             ),
         ]


SteveMacenski (approved):
Thanks :-)

PR #4295, opened 2024-05-03
Continuation of #4284

SteveMacenski:
See https://github.com/open-navigation/navigation2/pull/4284


SteveMacenski:
@doisyg also see https://github.com/open-navigation/navigation2/commit/a6f1f9e035f7f3db5a4615e83215273a033b000f these are the lines I disagree with, they're changing conversions for handling of the costmap, not the Graph that was exceeding it. The costmap APIs will downshift these into unsigned int anyway per its API

doisyg:
Thanks for following up, makes sense! 
(and checked with my big map)

PR #4298, opened 2024-05-04
Move Nav2 CI to 24.04 / Rolling

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

Various fixes for Nav2 Rolling on Noble

---

#### For Maintainers: 


.devcontainer/devcontainer.json
```diff
@@ -3,10 +3,11 @@
     "build": {
         "dockerfile": "../Dockerfile",
         "context": "..",
-        "target": "visualizer",
+        "target": "dever",
```
tonynajjar, line 6: I think this would make more sense as a default as I imagine most people just want a docker image of Nav2 built from source

.devcontainer/devcontainer.json
```diff
@@ -3,10 +3,11 @@
...
         "cacheFrom": "ghcr.io/ros-planning/navigation2:main"
     },
     "runArgs": [
+        "--name=nav2"
```
tonynajjar, line 10: makes the process of rebuilding and accessing the container quicker if we don't have to always check the name of the container

Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN apt-get update && \
       ros-$ROS_DISTRO-rmw-fastrtps-cpp \
       ros-$ROS_DISTRO-rmw-connextdds \
       ros-$ROS_DISTRO-rmw-cyclonedds-cpp \
-    && pip3 install \
+    && pip3 install --break-system-packages \
```
tonynajjar, line 60: because of https://peps.python.org/pep-0668/. Maybe the cleaner way is to use venv

nav2_mppi_controller/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if(COMPILER_SUPPORTS_FMA)
 endif()
 
 # If building one the same hardware to be deployed on, try `-march=native`!
-add_compile_options(-O3 -finline-limit=10000000 -ffp-contract=fast -ffast-math -mtune=generic)
```
SteveMacenski, line 68: Readd

nav2_common/cmake/nav2_package.cmake
```diff
@@ -37,7 +37,7 @@ macro(nav2_package)
   endif()
 
   if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference)
```
SteveMacenski, line 40: Readd

Dockerfile
SteveMacenski: ? what happens if we don't include this?

tonynajjar: This: https://github.com/open-navigation/navigation2/actions/runs/8858102523/job/24326368742

nav2_common/cmake/nav2_package.cmake
tonynajjar: Removing -Wnull-dereference temporarily because of https://stackoverflow.com/questions/77436264/potential-null-deference-warning

tonynajjar: Removing Wshadow temporarily because of https://github.com/BehaviorTree/BehaviorTree.CPP/issues/811

nav2_system_tests/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(nav2_system_tests)
 
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 23)
+endif()
```
tonynajjar, line 6: added because of:

```
opt/overlay_ws/src/navigation2/nav2_system_tests/src/behaviors/spin/test_spin_behavior_node.cpp:74:3: error: ‘f32’ or ‘F32’ suffix on floating constant only available with ‘-std=c++2b’ or ‘-std=gnu++2b’ [-Werror=pedantic]
   74 |   if (std::getenv("MAKE_FAKE_COSTMAP") != NULL && abs(target_yaw) > M_PI_2f32) {
```

.devcontainer/devcontainer.json
ruffsl: That's fine. With foxglove now closed source, I'll have to get around to cleaning out the old visualizer stages anyhow.

.devcontainer/devcontainer.json
ruffsl: That fine. Just so long as we don't assume the container name anywhere else.
One shouldn't grow accustomed to hardcoding names of an ephemeral containers.

Dockerfile
ruffsl: Yea, I encountered this as well. it's the simplests workaround until colcon org upstreams my cache extension.
They really want folks to sign a waiver agreement now before handing over a :foot: :gun: .

nav2_map_server/CMakeLists.txt
```diff
@@ -160,5 +162,5 @@ ament_export_libraries(
   ${library_name}
   ${map_io_library_name}
 )
-ament_export_dependencies(${map_io_dependencies} ${map_server_dependencies})
+ament_export_dependencies(${map_io_dependencies} ${map_server_dependencies} yaml-cpp)
```
ruffsl, line 165: Nice. Why didn't this bite us before?

Dockerfile
```diff
@@ -101,6 +101,8 @@ RUN . $UNDERLAY_WS/install/setup.sh && \
       --from-paths src \
       --skip-keys " \
         slam_toolbox \
+        turtlebot3_gazebo \
+        gazebo_ros_pkgs \
```
ruffsl, line 105: Shouldn't you add these here too?

https://github.com/open-navigation/navigation2/blob/d0a65d5afbab3b758408b79ebf5cd344b3377b00/.circleci/config.yml#L110-L112

ruffsl (commented):
I'll let @SteveMacenski comment on the CMake diffs, but only a few comments on the Docker/CI bits.

nav2_map_server/CMakeLists.txt
SteveMacenski: The vendor structure changed recently :(

SteveMacenski:
@tonynajjar my only beef left here is the cmake changes - I don't want to remove important flags for our project due to dependency issues. Do you know how many files-ish (not exact; is it a lot or a little) have a problem due to this? If small, we can do what we did with other dependencies (example):

```
// turn off the specific warning. Can also use "-Wall"
#pragma GCC diagnostic ignored "-Wunused-but-set-variable"

#include <boost/uuid/uuid.hpp>
#include <boost/uuid/uuid_generators.hpp>
#include <boost/uuid/uuid_io.hpp>
#include <boost/lexical_cast.hpp>

// turn the warnings back on
#pragma GCC diagnostic pop
```

Though, I'm not sure why `-finline-limit=10000000` was removed from MPPI?

nav2_mppi_controller/CMakeLists.txt
tonynajjar: This causes: 

```
/usr/include/c++/13/bits/stl_algobase.h:437:30: error: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ forming offset 8 is out of the bounds [0, 8] of object ‘shape’ with type ‘index_type’ {aka ‘std::array<long unsigned int, 1>’} [-Werror=array-bounds=]
  437 |             __builtin_memmove(__result, __first, sizeof(_Tp) * _Num);
```

Ideally the compile options that are set in nav2_package() should not apply to external packages. Reading around it looks like this is possible but I'm at my limit of Cmake knowledge to understand how to make this happen

SteveMacenski: `target_compile_options` perhaps for that one then?

nav2_system_tests/src/behavior_tree/test_behavior_tree_node.cpp
```diff
@@ -79,10 +79,10 @@ class BehaviorTreeHandler
...
     }
 
-    auto xml_string = std::string(
-      std::istreambuf_iterator<char>(xml_file),
-      std::istreambuf_iterator<char>());
-
+    std::stringstream buffer;
+    buffer << xml_file.rdbuf();
+    xml_file.close();
+    std::string xml_string = buffer.str();
```
tonynajjar, line 85: This achieves the same and avoids the [error](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105580)

SteveMacenski:
At last review, this looks good to me. @ruffsl any thoughts? 

Otherwise, CI is failing to build (still looks to be on Jammy?) but no more code-related concerns. I'll probably file a few tickets when we merge as reminders but nothing that I'm going to concern you with

nav2_mppi_controller/CMakeLists.txt
aosmw: target_link_libraries provides a way  to suppress warnings from third party headers SYSTEM keyword https://stackoverflow.com/a/70169343 

ament_target_dependencies also provides this option.
https://github.com/ament/ament_cmake/blob/fdbf4574d7ccc67c29d63f906ddbd88017eb9ecc/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L51

Introduced 4 years ago so might do the job you want.
https://github.com/ament/ament_cmake/commit/10a6ceb471a717113289e2486eec513db26e85d6

tonynajjar:
>  CI is failing to build (still looks to be on Jammy?)

yes because `ghcr.io/ros-planning/navigation2:main` is still on jammy, since the [Update CI image](https://github.com/ros-navigation/navigation2/actions/workflows/update_ci_image.yaml) has been failing for a while, probably since ubuntu noble migration. @ruffsl any idea why [it fails](https://github.com/ros-navigation/navigation2/actions/runs/9107994592/job/25037900242)?

>  I'll probably file a few tickets when we merge as reminders but nothing that I'm going to concern you with

Yep, didn't plan on leaving simple TODOs in the code

ruffsl:
> @ruffsl any idea why [it fails](https://github.com/ros-navigation/navigation2/actions/runs/9107994592/job/25037900242)?

Something funky is going on. Looking at the latest workflow run for the `Update CI Image`, why is it trying to login to GHCR using my own user name? Did something screwy happen with the GitHub org/repo transfer?

- https://github.com/ros-navigation/navigation2/actions/workflows/update_ci_image.yaml
- https://github.com/ros-navigation/navigation2/actions/runs/9124355843
- https://github.com/ros-navigation/navigation2/actions/runs/9124355843/job/25088397479#step:2:18
- https://github.com/ros-navigation/navigation2/actions/runs/9124355843/workflow#L39-L40

```
Starting job container
  /usr/bin/docker --config /home/runner/work/_temp/.docker_b1f940ca-3298-4b86-adc9-d03dea1cd05a login ghcr.io -u ruffsl --password-stdin
```

For example, for that same workflow run linked above, my username is also associated with the trigger commit on main:

> Triggered via schedule 11 hours ago
> @ruffsl [ bf291d7](https://github.com/ros-navigation/navigation2/commit/bf291d704dbb1c2572575dd0620b12f816407233) [main](https://github.com/ros-navigation/navigation2)

But when clicking at the commit link, @SteveMacenski is the author of that commit, as I haven't pushed in a while.

---

At first I had suspected that perhaps the context `${{ github.repository_owner }}` now somehow resolves to ruffsl, E.g:

- https://github.com/ros-navigation/navigation2/actions/runs/9124355843/workflow#L73

But that thankfully doesn't seem to be the case, as seen from the next workflow job `Rebuild CI Image`:

- https://github.com/ros-navigation/navigation2/actions/runs/9124355843/job/25088411537#step:4:4

This github org/repo has no Actions secrets and variables, so not sure how else the docker login user could get altered.

ruffsl:
Oh no, we left behind our CI images at the old `ros-planning` GitHub org, and our workflow is also just as confused:

- https://github.com/orgs/ros-planning/packages/container/package/navigation2
- https://github.com/orgs/ros-navigation/packages

>   /usr/bin/docker --config /home/runner/work/_temp/.docker_fec47f9d-77ff-4c82-92d2-c6f945d07d45 pull ghcr.io/ros-navigation/navigation2:main
> https://github.com/ros-navigation/navigation2/actions/runs/9107994592/job/25037900242#step:2:19

Oh well, I'm unsure if GitHub packages even can be migrated, but we should probably delete them to avoid use confusion.

I'll build the images for the `builder` target stage locally, then create and push them to a new container registry repo for our new GitHub org. That should hopefully bootstrap everything.

ruffsl:
@SteveMacenski , could I ask you to change the package visibility for this new container registry to public?
- https://github.com/orgs/ros-navigation/packages/container/navigation2/settings

Then CircleCI, and any end users, could download the image.
Once that's public, then just retrigger the CircleCI for this PR and it should rev right back up.

I'm going to refactor the CI for GitHub action towards the end of this month, so not to concerned with the funky stuff above.

SteveMacenski:
I think it should be done? Please let me know if I missed something

ruffsl:
@tonynajjar , looks like the CI is now working but the build is still failing:

> Summary: 34 packages finished [42min 33s]
  1 package failed: nav2_mppi_controller
  1 package aborted: nav2_smac_planner
  3 packages had stderr output: nav2_common nav2_mppi_controller nav2_simple_commander
  2 packages not processed
> https://app.circleci.com/pipelines/github/ros-navigation/navigation2/11638/workflows/a432e259-88ed-4d15-89e1-36b6a335661f/jobs/35773?invite=true#step-117-51302_41

It's the weekend so my eyes glaze over from the wall of gcc errors, but it odd in that I had this branch building locally just fine, albeit a different Dockerfile I'm refactoring. Did you use any Cmake flags to ignore any warnings when building locally?

tonynajjar:
Very odd, that's the same error that I get when building nav2_mppi_controller on main, it's supposed to be solved with the latest changes though... Anyway this lead me to do a better fix and with that one looks like the build succeeds!

tonynajjar:
@SteveMacenski system tests obviously fail because of gazebo missing. Would you want to disable system tests needing gazebo and merge this or wait for the gazebo transition to be done first?

SteveMacenski:
I was hoping we wouldn't be this close to the wire on the new gazebo migration, but I guess we are where we are. 

Comment them out but add a ticket which points to the commit where they are commented out to add back in once we have a migration to a newer simulator. We should make sure these are not forgotten.

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -20,6 +20,9 @@ find_package(nav2_util REQUIRED)
 
 nav2_package()
 
+# TODO: Delete after https://github.com/BehaviorTree/BehaviorTree.CPP/issues/811 is released
```
SteveMacenski, line 23: File ticket with this commit as a reminder to check in on it, we don't do TODOs in this repo, they're forgotten about easily. Leave the comment though in-line (just not TODO).

SteveMacenski (changes requested):
One pedantic process thing, but otherwise LGTM. Edit: DCO sign off too, if you'd be so kind :-) 

I'm planning this afternoon to dogfood my own tutorial https://docs.nav2.org/tutorials/docs/docker_dev.html with Noble/Jazzy to build and take a look at all of this on my 22.04 machine

SteveMacenski:
By the way, I see many more test failures than those shown above when I run this locally on the `jazzy-desktop-full` image. Reproduce with (change the builders to your liking, I like to limit it when running in the background it doesn't thrash my machine while doing other foreground work).

```
sudo docker run -it --net=host -v .:/jazzy_ws  osrf/ros:jazzy-desktop-full

sudo apt update
rosdep update
rosdep install --from-paths src --ignore-src -y -r --skip-keys "slam_toolbox gazebo_ros_pkgs turtlebot3_gazebo"
colcon build --parallel-workers 2
colcon test --parallel-workers 2
```
---

Failures 

```
    [ RUN      ] PointPortTest.test_wrong_syntax
    unknown file: Failure
    C++ exception with description "The port with name "test" and value "1.0;2.0;3.0;4.0" can not be converted to geometry_msgs::msg::Point_<std::allocator<void> >" thrown in the test body.
    
    [  FAILED  ] PointPortTest.test_wrong_syntax (0 ms)

    [ RUN      ] QuaternionPortTest.test_wrong_syntax
    unknown file: Failure
    C++ exception with description "The port with name "test" and value "1.0;2.0;3.0;4.0;5.0" can not be converted to geometry_msgs::msg::Quaternion_<std::allocator<void> >" thrown in the test body.
    
    [  FAILED  ] QuaternionPortTest.test_wrong_syntax (0 ms)
```

```
    [----------] 1 test from DistanceTraveledConditionTestFixture
    [NON-XML-CHAR-0x1B][37;1m2024-05-20 22:14:22.815 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_TRANSPORT_SHM[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mFailed to create segment be0cc4e20bd7f9b2: boost::interprocess_exception::library_error[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mcompute_per_allocation_extra_size[NON-XML-CHAR-0x1B][m
    [NON-XML-CHAR-0x1B][37;1m2024-05-20 22:14:22.816 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_TRANSPORT_SHM[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mFailed to create Shared Memory Manager for domain fastrtps: boost::interprocess_exception::library_error[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mcreate[NON-XML-CHAR-0x1B][m
    [NON-XML-CHAR-0x1B][37;1m2024-05-20 22:14:22.816 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_PARTICIPANT[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mUnable to Register SHM Transport. SHM Transport is not supported in the current platform.[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mRTPSParticipantImpl[NON-XML-CHAR-0x1B][m
    [WARN] [1716243262.828360158] []: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
    [INFO] [1716243262.928507625] [test_behavior_tree_fixture]: Transforms are available now!
    [ RUN      ] DistanceTraveledConditionTestFixture.test_behavior
    -- run_test.py: return code -11
    -- run_test.py: generate result file '/jazzy_ws/build/nav2_behavior_tree/test_results/nav2_behavior_tree/test_condition_distance_traveled.gtest.xml' with failed test
    -- run_test.py: verify result file '/jazzy_ws/build/nav2_behavior_tree/test_results/nav2_behavior_tree/test_condition_distance_traveled.gtest.xml'
  >>>
- test_condition_time_expired
  <<< failure message
```

^ this repeats for `TimeExpiredConditionTestFixture`, `PathExpiringTimerConditionTestFixture`, `GoalUpdatedConditionTestFixture`, `GloballyUpdatedGoalConditionTestFixture`, `InitialPoseReceivedConditionTestFixture`, `DistanceControllerTestFixture`, `SpeedControllerTestFixture`, `RateControllerTestFixture`, `GoalUpdaterTestFixture`, `GoalUpdatedControllerTestFixture`, 


```
    [ RUN      ] GoalUpdaterTestFixture.test_get_latest_goal_update
    /jazzy_ws/src/navigation2/nav2_behavior_tree/test/plugins/decorator/test_goal_updater_node.cpp:192: Failure
      updated_goal
        Which is: 96-byte object <4A-CB 4B-66 0F-96 0B-0C B8-7A F6-30 FE-7F 00-00 00-00 00-00 00-00 00-00 00-74 65-00 90-58 00-00 10-7C F6-30 FE-7F 00-00 00-00 00-00 00-00 F0-3F 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 F0-3F>
      goal_to_update_2
        Which is: 96-byte object <4A-CB 4B-66 C2-9A 0B-0C 58-7A F6-30 FE-7F 00-00 00-00 00-00 00-00 00-00 00-E8 5E-4B 90-58 00-00 C0-7B F6-30 FE-7F 00-00 00-00 00-00 00-00 08-40 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 F0-3F>
    
    [  FAILED  ] GoalUpdaterTestFixture.test_get_latest_goal_update (0 ms)
```

```
    [ RUN      ] Tester.testCollisionPointsMarkers
    [NON-XML-CHAR-0x1B][37;1m2024-05-20 22:15:42.154 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_TRANSPORT_SHM[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mFailed to create segment 1cc5915a972d89d9: boost::interprocess_exception::library_error[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mcompute_per_allocation_extra_size[NON-XML-CHAR-0x1B][m
    [NON-XML-CHAR-0x1B][37;1m2024-05-20 22:15:42.154 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_TRANSPORT_SHM[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mFailed to create Shared Memory Manager for domain fastrtps: boost::interprocess_exception::library_error[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mcreate[NON-XML-CHAR-0x1B][m
    [NON-XML-CHAR-0x1B][37;1m2024-05-20 22:15:42.154 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_PARTICIPANT[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mUnable to Register SHM Transport. SHM Transport is not supported in the current platform.[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mRTPSParticipantImpl[NON-XML-CHAR-0x1B][m
    [INFO] [1716243342.157957346] [collision_monitor]: 
    	collision_monitor lifecycle node launched. 
    	Waiting on external lifecycle transitions to activate
    	See https://design.ros2.org/articles/node_lifecycle.html for more information.
    [INFO] [1716243342.160239857] [collision_monitor]: Configuring
    [INFO] [1716243342.161840263] [collision_monitor]: [Scan]: Creating Scan
    [INFO] [1716243342.163109480] [collision_monitor]: Activating
    [INFO] [1716243342.163130186] [collision_monitor]: Creating bond (collision_monitor) to lifecycle manager.
    [WARN] [1716243342.164861030] [collision_monitor]: Robot to stop due to invalid source. Either due to data not published yet, or to lack of new data received within the sensor timeout, or if impossible to transform data to base frame
    /jazzy_ws/src/navigation2/nav2_collision_monitor/test/collision_monitor_node_test.cpp:1509: Failure
    Expected: (collision_points_marker_msg_->markers[0].points.size()) != (0u), actual: 0 vs 0
    
    [INFO] [1716243342.209206392] [collision_monitor]: [Scan]: Destroying Scan
    [INFO] [1716243342.210921746] [collision_monitor]: Destroying
    [  FAILED  ] Tester.testCollisionPointsMarkers (71 ms)
```

The following failed with all tests passing, but issues in shutdown like 
```
    Warning: class_loader.ClassLoader: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded.
             at line 127 in ./src/class_loader.cpp
    [INFO] [1716243365.746525522] [costmap]: Running Nav2 LifecycleNode rcl preshutdown (costmap)
    [INFO] [1716243365.746611543] [costmap]: Destroying bond (costmap) to lifecycle manager.
    Failed to delete datawriter, at ./src/publisher.cpp:45 during '__function__'
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    -- run_test.py: return code -11
    -- run_test.py: inject classname prefix into gtest result file '/jazzy_ws/build/nav2_smac_planner/test_results/nav2_smac_planner/test_nodehybrid.gtest.xml'
    -- run_test.py: verify result file '/jazzy_ws/build/nav2_smac_planner/test_results/nav2_smac_planner/test_nodehybrid.gtest.xml'
  >>>
```

`NodeHybridTest.test_node_hybrid`, `AStarTest.test_a_star_2d`, `SmacTest.test_smac_se2`, `SmacTest.test_smac_lattice`, `SmootherTest.test_full_smoother`, `WPTest.test_dynamic_parameters`


```
    [ RUN      ] ActionTest.test_simple_action_activation_cycling
    [INFO] [1716243188.077976554] [fibonacci_server_node]: Deactivating
    [WARN] [1716243188.078035280] [fibonacci_server_node]: [fibonacci] [ActionServer] Requested to deactivate server but goal is still executing. Should check if action server is running before deactivating.
  >>>
build/nav2_util/test_results/nav2_util/test_actions.gtest.xml: 1 test, 1 error, 0 failures, 0 skipped
- nav2_util test_actions.gtest.missing_result
  <<< error message
    The test did not generate a result file.
  >>>
build/nav2_waypoint_follower/Testing/2024052
```

---

The linting failures can be fixed applying this patch attached: [mypatch.zip](https://github.com/ros-navigation/navigation2/files/15382271/mypatch.zip)

SteveMacenski:
I started looking into some of them, for instance the waypoint follower test can be made to pass by adding the following to the end of the `test_dynamic_parameters` test

```
  follower->deactivate();
  follower->cleanup();
  follower.reset();
```

Even though these should automatically be called on destruction which was tested to work prior to the 24.04 migration https://github.com/ros-navigation/navigation2/blob/main/nav2_util/src/lifecycle_node.cpp#L77-L106

This error is new to us, so maybe its just straight up crashing at the RMW layer by running the rcl preshutdown procedure?

```
    [INFO] [1716245596.048208666] [waypoint_follower]: Running Nav2 LifecycleNode rcl preshutdown (waypoint_follower)
    Failed to publish log message to rosout: cannot publish data, at ./src/rmw_publish.cpp:66, at ./src/rcl/publisher.c:284
    [INFO] [1716245596.048288331] [waypoint_follower]: Destroying bond (waypoint_follower) to lifecycle manager.
    Failed to publish log message to rosout: cannot publish data, at ./src/rmw_publish.cpp:66, at ./src/rcl/publisher.c:284
```

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -57,7 +57,7 @@ inline BT::NodeStatus GoalUpdater::tick()
 
   getInput("input_goal", goal);
 
-  callback_group_executor_.spin_some();
```
tonynajjar, line 60: This fixes the goal_updater test. The problem is that for some reason with `spin_some()`, even if there is theoretically some work in the pipeline, the callbacks are not called. Interestingly enough, the test also succeeds if we call `spin_some` twice in a row

```
```

I've also check that this fails in other tests like `test_is_battery_charging` but the overall test succeeds coincidentally because FAILURE is expected whether or not the subscription callback is called.

I don't think this is a recent issue, I've seen this in the past. I'd propose in this PR to just fix what is obviously broken and create a separate ticket to check whether or not `spin_some()` triggers the callbacks

SteveMacenski: Worth noting to @clalancette

tonynajjar:
Looks like `ament_uncrustify` from CI does not agree with the one I'm running locally (rolling image) and there is no common ground. Something to look into

SteveMacenski:
This is the action test failure:

```
   [----------] Global test environment tear-down
    [==========] 7 tests from 1 test suite ran. (5276 ms total)
    [  PASSED  ] 7 tests.
    [WARN] [1716327862.626602630] [fibonacci_server_node]: [fibonacci] [ActionServer] Current goal was not completed successfully.
    [WARN] [1716327862.626609212] [fibonacci_server_node]: [fibonacci] [ActionServer] Client requested to cancel the goal. Cancelling.
    
    >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
    This error state is being overwritten:
    
      'feedback publisher is invalid, at ./src/rcl_action/action_server.c:940'
    
    with this new error message:
    
      'cannot publish data, at ./src/rmw_publish.cpp:66'
    
    rcutils_reset_error() should be called after error handling to avoid this.
    <<<
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    
    >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
    This error state is being overwritten:
    
      'Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262'
    
    with this new error message:
    
      'cannot publish data, at ./src/rmw_publish.cpp:66'
    
    rcutils_reset_error() should be called after error handling to avoid this.
    <<<
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    
    >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
    This error state is being overwritten:
    
      'Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262'
    
    with this new error message:
    
      'cannot publish data, at ./src/rmw_publish.cpp:66'
    
    rcutils_reset_error() should be called after error handling to avoid this.
    <<<
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    
    >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
    This error state is being overwritten:
    
      'Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262'
    
    with this new error message:
    
      'cannot publish data, at ./src/rmw_publish.cpp:66'
    
    rcutils_reset_error() should be called after error handling to avoid this.
    <<<
    Failed to delete datawriter, at ./src/publisher.cpp:45 during '__function__'
    Failed to delete datawriter, at ./src/publisher.cpp:45 during '__function__'
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    [ERROR] [1716327862.627785482] [fibonacci_server_node.rclcpp]: Error in destruction of rcl subscription handle: Failed to delete datareader, at ./src/subscription.cpp:52, at ./src/rcl/subscription.c:200
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    [ERROR] [1716327862.627801920] [fibonacci_server_node.rclcpp]: Error in destruction of rcl subscription handle: Failed to delete datareader, at ./src/subscription.cpp:52, at ./src/rcl/subscription.c:200
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    [ERROR] [1716327862.627815406] [fibonacci_server_node.rclcpp]: Error in destruction of rcl subscription handle: Failed to delete datareader, at ./src/subscription.cpp:52, at ./src/rcl/subscription.c:200
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    [ERROR] [1716327862.627862083] [fibonacci_server_node.rclcpp]: Error in destruction of rcl subscription handle: Failed to delete datareader, at ./src/subscription.cpp:52, at ./src/rcl/subscription.c:200
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    [ERROR] [1716327862.627889188] [fibonacci_server_node.rclcpp]: Error in destruction of rcl service handle: Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    [ERROR] [1716327862.627910522] [fibonacci_server_node.rclcpp]: Error in destruction of rcl service handle: Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    [ERROR] [1716327862.627930226] [fibonacci_server_node.rclcpp]: Error in destruction of rcl service handle: Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    [ERROR] [1716327862.627949664] [fibonacci_server_node.rclcpp]: Error in destruction of rcl service handle: Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    [ERROR] [1716327862.627970082] [fibonacci_server_node.rclcpp]: Error in destruction of rcl service handle: Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    [ERROR] [1716327862.627990988] [fibonacci_server_node.rclcpp]: Error in destruction of rcl service handle: Fail in delete datawriter, at ./src/rmw_service.cpp:104, at ./src/rcl/service.c:262
    Failed to delete datawriter, at ./src/publisher.cpp:45 during '__function__'
    [ERROR] [1716327862.628012608] [fibonacci_server_node.rclcpp]: Error in destruction of rcl publisher handle: cannot publish data, at ./src/rmw_publish.cpp:66, at ./src/rcl/publisher.c:201
    -- run_test.py: return code -11
    -- run_test.py: inject classname prefix into gtest result file '/jazzy_ws/build/nav2_util/test_results/nav2_util/test_actions.gtest.xml'
    -- run_test.py: verify result file '/jazzy_ws/build/nav2_util/test_results/nav2_util/test_actions.gtest.xml'
  >>>

```

The actions passed during this run, but when bringing down the file, that's when all the issues occur. That is similar to the issues with `NodeHybridTest.test_node_hybrid`, `AStarTest.test_a_star_2d`, `SmacTest.test_smac_se2`, `SmacTest.test_smac_lattice`, `SmootherTest.test_full_smoother`

SteveMacenski:
@tonynajjar merge in changes from https://github.com/ros-navigation/navigation2/tree/BR2 into your branch. This addresses Actions and the BT failures

facontidavide:
> Behavior tree node failures RESOLVED - BT.CPP bug/regression causing crash if input configs don't contain all possible ports for the tests (CC @facontidavide: even if that was an intentional change, I expect that a crash isn't the expected behavior rather an exception or log. Though seems like maybe its a regression?)

Yes, I should fix it and have am exception instead.

SteveMacenski:
With the recent pushes, the remaining are the test bringdown failures which are just `N` instances of presumably the same problem. All the tests exhibit the Fsat-DDS shared memory log errors, but not blocking our CI motion to 24.04. Nor is the fact that there may be some issues (?) on bring down upstream once we find a workaround that gets us moving. But these probably should be followed up upon to those appropriate project leads for their awareness (which I've started).

Will be looking into the last thing this afternoon.

SteveMacenski:
OK, fixed. It looks like another instance of the destruction not working like described in the WPF comment above. To my unsophisticated eye, it looks like some of the context / RMW items aren't still active during the pre-shutdown callbacks, causing exceptions / internal ROS 2 things to crash / deadlock. When we fully set things down so that our pre-shutdown callback doesn't have to do any work or is avoided altogether by destroying the object before the rclcpp shutdown, then we avoid the problem. I think this last, short one does a good job at highlighting this:

```
    [INFO] [1716413808.442739807] [costmap]: Running Nav2 LifecycleNode rcl preshutdown (costmap)
    [INFO] [1716413808.442783529] [costmap]: Destroying bond (costmap) to lifecycle manager.
    Failed to delete datawriter, at ./src/publisher.cpp:45 during '__function__'
    cannot publish data, at ./src/rmw_publish.cpp:66 during '__function__'
    Fail in delete datareader, at ./src/rmw_service.cpp:89 during '__function__'
    -- run_test.py: return code -11
```

SteveMacenski:
@tonynajjar do you see the `nav2_costmap_2d.DynParamTestNode` test failure locally? I do not but CI is reporting it :(

tonynajjar:
> @tonynajjar do you see the `nav2_costmap_2d.DynParamTestNode` test failure locally? I do not but CI is reporting it :(

Just tried on rolling locally 10 times and I didn't get it

SteveMacenski:
Maybe break the cache with `v23`? Or the docker image https://github.com/ros-navigation/navigation2/blob/f671d38f73f543cb5c5adc6b0ff24b32c0ccc227/.circleci/config.yml#L488 ?

Dockerfile
```diff
@@ -96,11 +96,15 @@ ARG OVERLAY_WS
...
 COPY --from=cacher /tmp/$OVERLAY_WS ./
+
+# Remove/Replace turtlebot3_gazebo and gazebo_ros_pkgs from --skip-keys after https://github.com/ros-navigation/navigation2/pull/3634
 RUN . $UNDERLAY_WS/install/setup.sh && \
     apt-get update && rosdep install -q -y \
       --from-paths src \
       --skip-keys " \
         slam_toolbox \
+        turtlebot3_gazebo \
+        gazebo_ros_pkgs \
```
SteveMacenski, line 107: `gazebo_ros_pkgs` can be removed here and in the CircleCI file

.circleci/config.yml
```diff
@@ -109,6 +105,8 @@ _commands:
                   --ignore-src \
                   --skip-keys " \
                     slam_toolbox \
+                    turtlebot3_gazebo \
+                    gazebo_ros_pkgs \
```
SteveMacenski, line 109:
```suggestion
```

Dockerfile
SteveMacenski, line 107:
```suggestion
```

SteveMacenski (approved):

Issue #4110, opened 2024-02-09
[for help] BT node Is_path_longer_on_approach doesn't work


kevinpretell:
## Bug report

I'm practicing on this nav2 simulation for outdoor navigation without problems.
https://navigation.ros.org/tutorials/docs/navigation2_with_gps.html

To manage obstacles I found this BT node which is very interesting, if the robot sees a dynamic obstacle it stops until it goes away.
https://navigation.ros.org/behavior_trees/trees/nav_to_pose_and_pause_near_goal_obstacle.html

I then replaced my default nav_to_pose.xml behavior with the one from the demo above by placing it in the config folder.

But nav2 seems to ignore the new beaviour xml by following its path without stopping at the obstacle.

I ask here because I saw that no one is talking around about this BT node.
Many thanks in advance

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - ROS2 Iron
 
- DDS implementation:
  - default dds

#### Steps to reproduce issue
Replace the default bt: navigate_to_pose_w_replanning_and_recovery.xml
with the bt: navigate_to_pose_w_replanning_goal_patience_and_recovery.xml

PathLongerOnApproach setting:
`<PathLongerOnApproach path="{path}" prox_len="6.0" length_factor="1.2">`

start the gazebo simulation:
ros2 launch nav2_gps_waypoint_follower_demo gps_waypoint_follower.launch.py use_rviz:=True

From the gazebo I add an obstacle in front of the moving robot.

#### Expected behavior
the robot should stop and wait for the obstacle to go away before following the new path

#### Actual behavior
the robot does not stop when a longer path is recalculated waiting for the obstacle to go away, and takes the longer path

#### Additional information
Maybe I need to change something else in the Nav2 params file?
here is my bt navigator params:
```
bt_navigator:
  ros__parameters:
    global_frame: map
    robot_base_frame: base_link
    odom_topic: /odom
    bt_loop_duration: 10
    default_server_timeout: 20
    default_nav_to_pose_bt_xml: "/home/kevin/iron_ws/src/nav2_gps_waypoint_follower_demo/config/nav_to_pose.xml"
    default_nav_through_poses_bt_xml: "/home/kevin/iron_ws/src/nav2_gps_waypoint_follower_demo/config/nav_through_pose.xml"

    # navigators: ["navigate_to_pose", "navigate_through_poses"]
    # navigate_to_pose:
[...]

PR #4300, opened 2024-05-06
fixes the issue when the path longer on approach decorator goes to idle state

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4110 |
| Primary OS tested on |(Ubuntu|
| Robotic platform tested on | Neobotix ROX |

---

* Crux of the issue seems that the status of the node is set to IDLE. Last year, when I had implemented, the status ended up to success and went IDLE only when the entire BT was IDLE. This IDLE state was blocking the logic of the intended decorator going forward. I however have found a fix, that uses the goals of the old path and the new path, to avoid the blocking.

---

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Thanks!!

PR #4302, opened 2024-05-06
Create service to get cost of a cell in the costmap

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

Draft PR created for https://github.com/open-navigation/navigation2/issues/4271

* Added getCost service

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -249,7 +249,10 @@ Costmap2DROS::on_configure(const rclcpp_lifecycle::State & /*state*/)
     makeFootprintFromString(footprint_, new_footprint);
     setRobotFootprint(new_footprint);
   }
-
+  // Create server to get the cost at a given point
+  get_cost_service_ = create_service<nav2_msgs::srv::GetCost>(
+    "get_cost_" + getName(),
+    std::bind(&Costmap2DROS::getCostCallback, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
   // Add cleaning service
   clear_costmap_service_ = std::make_unique<ClearCostmapService>(shared_from_this(), *this);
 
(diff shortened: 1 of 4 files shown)


nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -259,6 +262,18 @@ Costmap2DROS::on_configure(const rclcpp_lifecycle::State & /*state*/)
...
 }
 
+void Costmap2DROS::getCostCallback(
+  const std::shared_ptr<rmw_request_id_t> /*request_header*/,
+  const std::shared_ptr<nav2_msgs::srv::GetCost::Request> request,
+  const std::shared_ptr<nav2_msgs::srv::GetCost::Response> response)
+{
+  RCLCPP_INFO(
+    get_logger(), "Received request to get cost at point (%f, %f)", request->x, request->y);
+  auto cost = static_cast<float>(layered_costmap_->getCostmap()->getCost(
```
SteveMacenski, line 272: I think some bounds checking needs to happen that the request is within the costmap's size

nav2_msgs/srv/GetCost.srv
```diff
@@ -0,0 +1,6 @@
+# Get costmap cost at given point
+
+float32 x
+float32 y
+---
+uint8 cost
```
SteveMacenski, line 6: You cast the output to a float in the cpp code but its a uint?

SteveMacenski (changes requested):

SteveMacenski:
I'm not entirely sure this is a good thing to expose to users. I understand you want it to debug the current problem, but this will be crazy inefficient if misused in applications. I think perhaps this should just be a debugging tool for the current work unless you feel strongly about it otherwise

tonynajjar:
I don't feel strongly about it but can you elaborate on "but this will be crazy inefficient if misused in applications"? 
I was hoping this could also be the base for an rviz plugin with which you just click a point on the costmap and it gives you the cost

nav2_msgs/srv/GetCost.srv
tonynajjar: good point, I just copied it from [here](https://github.com/open-navigation/navigation2/blob/d0a65d5afbab3b758408b79ebf5cd344b3377b00/nav2_smac_planner/src/collision_checker.cpp#L114.) why is it cast there?

SteveMacenski: Because footprint cost is a float.

SteveMacenski:
I suppose if there's an rviz thing for it in nav2's config, that's fine with me

SteveMacenski:
It would have to be the truncated occupancy grid value, because that’s what rviz displays.

BriceRenaudeau:
Is it possible to add a bool in the .srv to get costmap cost at pose or footprint cost at pose?

SteveMacenski:
> Is it possible to add a bool in the .srv to get costmap cost at pose or footprint cost at pose?

The request would also require the `theta` then too

SteveMacenski:
Any update here @tonynajjar?

tonynajjar:
Yeah I was wondering what to do with it. I don't need it anymore personally but if enough people want it I guess we can keep it in the backlog as a low-prio thing to implement?

SteveMacenski:
I guess you could close this PR, open a ticket as low priority / good first issue with a description: service to get cell/footprint cost + rviz panel for selecting and using it. You could link this branch as a starting point

PR #4344, opened 2024-05-15
Fix the bug that behavior tree plugin PathLongerOnApproach will not work while using smac planner.

### LINKED ISSUE #4326: behavior tree plugin PathLongerOnApproach bug report
## Bug report

**Required Info:**

- Operating System:
  -Ubuntu 22.04 
  - 
- ROS2 Version:
  -Humble 
  - 
- Version or commit hash:
  -1.1.14 
  - 
- DDS implementation:
  -CycloneDDS 
  - 

#### Steps to reproduce issue

```
bool PathLongerOnApproach::isPathUpdated(
  nav_msgs::msg::Path & new_path,
  nav_msgs::msg::Path & old_path)
{
  return new_path != old_path && old_path.poses.size() != 0 &&
         new_path.poses.size() != 0 &&
         old_path.poses.back() == new_path.poses.back();
}
```

#### Expected behavior
Return true if path is update
#### Actual behavior
return false because in this line:
`         old_path.poses.back() == new_path.poses.back();
`
It compare two different path's PoseStamped.But The PoseStamped's comparison contain time's compare and position's compare:
```
  // comparison operators
  bool operator==(const PoseStamped_ & other) const
  {
    if (this->header != other.header) {
      return false;
    }
    if (this->pose != other.pose) {
      return false;
    }
    return true;
  }
```
Old path's last point's position is same as the new path's last point's position. Meanwhile, their timestamped is obviously different as i observed. So that even if the new path has updated. It will still return false.Which cause PathLongerOnApproach plugin out of action.
By the way. I tested it in the situation that `use_sim_time : false`

StetroF:
See https://github.com/ros-navigation/navigation2/issues/4326.
Fix the bug that isPathUpdated function will return false even if the new path is update for the reason that it not only compare the position between new path and old path's last pose.But also compare these two poses 's timestamped which is obviously different.

---

## Basic Info

| ------ | ----------- |

---

### DIFF
--- a/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
+++ b/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
@@ -36,7 +36,7 @@ bool PathLongerOnApproach::isPathUpdated(
 {
   return new_path != old_path && old_path.poses.size() != 0 &&
          new_path.poses.size() != 0 &&
-         old_path.poses.back() == new_path.poses.back();
+         old_path.poses.back().pose == new_path.poses.back().pose;
 }
 
 bool PathLongerOnApproach::isRobotInGoalProximity(


tonynajjar (approved):
Looks reasonable!

SteveMacenski (approved):

PR #4348, opened 2024-05-16
Remove flaky spin test that needs to be rewritten

SteveMacenski:
https://github.com/ros-navigation/navigation2/issues/4349 tracking its reintroduction. CC @tonynajjar


tonynajjar (approved):
Lovely

SteveMacenski:
All 12 green!

Issue #4237, opened 2024-04-03
[MPPI]: Replace xtensor with Eigen Experiment


guoxxiong:

SteveMacenski:
Can you back up that claim or provide any other detail?

guoxxiong:
Because my hardware does not support xsimd acceleration, and xtensor does not support gcc-7, while other programs rely on gcc-7, so I want to know whether the performance will be weakened by using Eigen instead of xtensor for the tensor of 2000x50.

SteveMacenski:
That's not sufficient to prevent us from looking into it as a potential course of action.

guoxxiong:
For a 2000*60 array, run following code, it outputs:
dot_product_xt: 240000
xtensor dot product took 0.00535805 seconds.
dot_product_eigen: 240000
Eigen dot product took 0.000759429 seconds.

Code:
#include <iostream>
#include <chrono>
#include <xtensor/xarray.hpp>
#include "/usr/include/eigen3/Eigen/Dense"

constexpr int rows = 2000;
constexpr int cols = 60;

int main()
{

    xt::xarray<double> A_xt({rows, cols}, 1.0);
    xt::xarray<double> B_xt({rows, cols}, 2.0);
    Eigen::MatrixXd A_eigen(rows, cols);
    Eigen::MatrixXd B_eigen(rows, cols);
    A_eigen.setOnes();
    B_eigen= B_eigen.setOnes() * 2.0;


    double dot_product_xt = 0.0;
    double dot_product_eigen = 0.0;


    auto start_xtensor = std::chrono::steady_clock::now();
    dot_product_xt = xt::sum(A_xt * B_xt)();
    std::cout << "dot_product_xt: " << dot_product_xt << std::endl;
    auto end_xtensor = std::chrono::steady_clock::now();
    std::chrono::duration<double> elapsed_xtensor = end_xtensor - start_xtensor;
    std::cout << "xtensor dot product took " << elapsed_xtensor.count() << " seconds." << std::endl;


    auto start_eigen = std::chrono::steady_clock::now();
    dot_product_eigen = (A_eigen.array() * B_eigen.array()).sum();
    std::cout << "dot_product_eigen: " << dot_product_eigen << std::endl;
    auto end_eigen = std::chrono::steady_clock::now();
    std::chrono::duration<double> elapsed_eigen = end_eigen - start_eigen;
[...]

PR #4352, opened 2024-05-17
Add acceleration constraints in the MPPI controller

Ayush1285:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #3351 |
| Robotic platform tested on | gazebo simulation of tb3 |

---

* Updating MPPI controller to consider acceleration and deceleration constraints.

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
@@ -28,6 +28,11 @@ struct ControlConstraints
   float vx_min;
   float vy;
   float wz;
+  float ax_max;
+  float ax_min;
+  float ay_max;
+  float ay_min;
+  float az;
 };
 
 /**
@@ -39,20 +44,6 @@ struct SamplingStd
   float vx;
   float vy;
   float wz;
-};
-
-/**
- * @struct mppi::models::AccelConstraints
- * @brief Constraints on acceleration and deceleration
- */
-struct AccelConstraints
-{
-  float ax_max;
-  float ax_min;
-  float ay_max;
-  float ay_min;
-  float alphaz_max;
-  float alphaz_min;
 };
 
 }  // namespace mppi::models
--- a/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
@@ -31,8 +31,7 @@ struct ControlConstraints
   float ax_max;
   float ax_min;
   float ay_max;
-  float ay_min;
-  float az;
+  float az_max;
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
@@ -49,6 +49,10 @@ struct AccelConstraints
 {
   float ax_max;
   float ax_min;
+  float ay_max;
+  float ay_min;
+  float alphaz_max;
+  float alphaz_min;
 };
 
 }  // namespace mppi::models
--- a/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
@@ -41,6 +41,16 @@ struct SamplingStd
   float wz;
 };
 
+/**
+ * @struct mppi::models::AccelConstraints
+ * @brief Constraints on acceleration and deceleration
+ */
+struct AccelConstraints
+{
+  float ax_max;
+  float ax_min;
+};
+
 }  // namespace mppi::models
 
 #endif  // NAV2_MPPI_CONTROLLER__MODELS__CONSTRAINTS_HPP_
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -86,25 +86,27 @@ public:
     float max_delta_vx = model_dt_ * control_constraints_.ax_max;
     float min_delta_vx = model_dt_ * control_constraints_.ax_min;
     float max_delta_vy = model_dt_ * control_constraints_.ay_max;
-    float min_delta_vy = model_dt_ * control_constraints_.ay_min;
-    float max_delta_wz = model_dt_ * control_constraints_.az;
+    float max_delta_wz = model_dt_ * control_constraints_.az_max;
+    float vx_last = state.vx(0, 0);
+    float vy_last = state.vy(0, 0);
+    float wz_last = state.wz(0, 0);
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        float& vx_curr = state.vx(i, j - 1);
         float& cvx_curr = state.cvx(i, j - 1);        
-        cvx_curr = std::clamp(cvx_curr, vx_curr + min_delta_vx, vx_curr + max_delta_vx);
+        cvx_curr = std::clamp(cvx_curr, vx_last + min_delta_vx, vx_last + max_delta_vx);
         state.vx(i, j) = cvx_curr;
+        vx_last  = cvx_curr;
 
-        float& wz_curr = state.wz(i, j - 1);
         float& cwz_curr = state.cwz(i, j - 1);
-        cwz_curr = std::clamp(cwz_curr, wz_curr - max_delta_wz, wz_curr + max_delta_wz);
+        cwz_curr = std::clamp(cwz_curr, wz_last - max_delta_wz, wz_last + max_delta_wz);
         state.wz(i, j) = cwz_curr;
+        wz_last = cwz_curr;
 
         if (is_holo) {
-          float& vy_curr = state.vy(i, j - 1);
           float& cvy_curr = state.cvy(i, j - 1);
-          cvy_curr = std::clamp(cvy_curr, vy_curr + min_delta_vy, vy_curr + max_delta_vy);
+          cvy_curr = std::clamp(cvy_curr, vy_last - max_delta_vy, vy_last + max_delta_vy);
           state.vy(i, j) = cvy_curr;
+          vy_last = cvy_curr;
         }
       }
     }
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -55,22 +55,14 @@ public:
   virtual ~MotionModel() = default;
 
   /**
-    * @brief Configure motion model on bringup and load required parameters
-    * @param name Name of plugin
-    * @param param_handler Parameter handler object
+    * @brief Initialize motion model on bringup and set required variables
+    * @param control_constraints Constraints on control
+    * @param model_dt duration of a time step
     */
-  void on_configure(
-    const std::string & name,
-    ParametersHandler * param_handler)
+  void initialize(const models::ControlConstraints & control_constraints, float model_dt)
   {
-    auto getParam = param_handler->getParamGetter(name);
-    getParam(accel_constraints_.ax_max, "ax_max", 2.0);
-    getParam(accel_constraints_.ax_min, "ax_min", -2.0);
-    getParam(accel_constraints_.ay_max, "ay_max", 0.0);
-    getParam(accel_constraints_.ay_min, "ay_min", -0.0);
-    getParam(accel_constraints_.alphaz_max, "alphaz_max", 3.0);
-    getParam(accel_constraints_.alphaz_min, "alphaz_min", -3.0);
-    getParam(model_dt_, "model_dt", 0.05);
+    control_constraints_ = control_constraints;
+    model_dt_ = model_dt;
   }
 
   /**
@@ -91,22 +83,28 @@ public:
     // }
 
     const bool is_holo = isHolonomic();
+    float max_delta_vx = model_dt_ * control_constraints_.ax_max;
+    float min_delta_vx = model_dt_ * control_constraints_.ax_min;
+    float max_delta_vy = model_dt_ * control_constraints_.ay_max;
+    float min_delta_vy = model_dt_ * control_constraints_.ay_min;
+    float max_delta_wz = model_dt_ * control_constraints_.az;
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        float max_feasible_vx = state.vx(i, j - 1) + model_dt_ * accel_constraints_.ax_max;
-        float min_feasible_vx = state.vx(i, j - 1) + model_dt_ * accel_constraints_.ax_min;
-        state.cvx(i, j - 1) = std::clamp(state.cvx(i, j - 1), min_feasible_vx, max_feasible_vx);
-        state.vx(i, j) = state.cvx(i, j - 1);
+        float& vx_curr = state.vx(i, j - 1);
+        float& cvx_curr = state.cvx(i, j - 1);        
+        cvx_curr = std::clamp(cvx_curr, vx_curr + min_delta_vx, vx_curr + max_delta_vx);
+        state.vx(i, j) = cvx_curr;
 
-        float max_feasible_wz = state.wz(i, j - 1) + model_dt_ * accel_constraints_.alphaz_max;
-        float min_feasible_wz = state.wz(i, j - 1) + model_dt_ * accel_constraints_.alphaz_min;
-        state.cwz(i, j - 1) = std::clamp(state.cwz(i, j - 1), min_feasible_wz, max_feasible_wz);
-        state.wz(i, j) = state.cwz(i, j - 1);
+        float& wz_curr = state.wz(i, j - 1);
+        float& cwz_curr = state.cwz(i, j - 1);
+        cwz_curr = std::clamp(cwz_curr, wz_curr - max_delta_wz, wz_curr + max_delta_wz);
+        state.wz(i, j) = cwz_curr;
+
         if (is_holo) {
-          float max_feasible_vy = state.vy(i, j - 1) + model_dt_ * accel_constraints_.ay_max;
-          float min_feasible_vy = state.vy(i, j - 1) + model_dt_ * accel_constraints_.ay_min;
-          state.cvy(i, j - 1) = std::clamp(state.cvy(i, j - 1), min_feasible_vy, max_feasible_vy);
-          state.vy(i, j) = state.cvy(i, j - 1);
+          float& vy_curr = state.vy(i, j - 1);
+          float& cvy_curr = state.cvy(i, j - 1);
+          cvy_curr = std::clamp(cvy_curr, vy_curr + min_delta_vy, vy_curr + max_delta_vy);
+          state.vy(i, j) = cvy_curr;
         }
       }
     }
@@ -126,7 +124,7 @@ public:
 
 protected:
   float model_dt_{0.0};
-  models::AccelConstraints accel_constraints_{0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
+  models::ControlConstraints control_constraints_{0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -87,10 +87,10 @@ public:
     float min_delta_vx = model_dt_ * control_constraints_.ax_min;
     float max_delta_vy = model_dt_ * control_constraints_.ay_max;
     float max_delta_wz = model_dt_ * control_constraints_.az_max;
-    float vx_last = state.vx(0, 0);
-    float vy_last = state.vy(0, 0);
-    float wz_last = state.wz(0, 0);
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
+      float vx_last = state.vx(i, 0);
+      float vy_last = state.vy(i, 0);
+      float wz_last = state.wz(i, 0);
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
         float & cvx_curr = state.cvx(i, j - 1);
         cvx_curr = std::clamp(cvx_curr, vx_last + min_delta_vx, vx_last + max_delta_vx);
@@ -127,7 +127,7 @@ public:
 protected:
   float model_dt_{0.0};
   models::ControlConstraints control_constraints_{0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
-   0.0f};
+    0.0f};
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -127,7 +127,7 @@ public:
 protected:
   float model_dt_{0.0};
   models::ControlConstraints control_constraints_{0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
-  0.0f};
+   0.0f};
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -92,18 +92,18 @@ public:
     float wz_last = state.wz(0, 0);
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        float& cvx_curr = state.cvx(i, j - 1);        
+        float & cvx_curr = state.cvx(i, j - 1);
         cvx_curr = std::clamp(cvx_curr, vx_last + min_delta_vx, vx_last + max_delta_vx);
         state.vx(i, j) = cvx_curr;
-        vx_last  = cvx_curr;
+        vx_last = cvx_curr;
 
-        float& cwz_curr = state.cwz(i, j - 1);
+        float & cwz_curr = state.cwz(i, j - 1);
         cwz_curr = std::clamp(cwz_curr, wz_last - max_delta_wz, wz_last + max_delta_wz);
         state.wz(i, j) = cwz_curr;
         wz_last = cwz_curr;
 
         if (is_holo) {
-          float& cvy_curr = state.cvy(i, j - 1);
+          float & cvy_curr = state.cvy(i, j - 1);
           cvy_curr = std::clamp(cvy_curr, vy_last - max_delta_vy, vy_last + max_delta_vy);
           state.vy(i, j) = cvy_curr;
           vy_last = cvy_curr;
@@ -126,7 +126,8 @@ public:
 
 protected:
   float model_dt_{0.0};
-  models::ControlConstraints control_constraints_{0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
+  models::ControlConstraints control_constraints_{0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
+  0.0f};
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -19,6 +19,7 @@
 
 #include "nav2_mppi_controller/models/control_sequence.hpp"
 #include "nav2_mppi_controller/models/state.hpp"
+#include "nav2_mppi_controller/models/constraints.hpp"
 
 // xtensor creates warnings that needs to be ignored as we are building with -Werror
 #pragma GCC diagnostic push
@@ -53,6 +54,21 @@ public:
   virtual ~MotionModel() = default;
 
   /**
+    * @brief Configure motion model on bringup and load required parameters
+    * @param name Name of plugin
+    * @param param_handler Parameter handler object
+    */
+  void on_configure(
+    const std::string & name,
+    ParametersHandler * param_handler)
+  {
+    auto getParam = param_handler->getParamGetter(name);
+    getParam(accl_constraints_.ax_max, "ax_max", 2.0);
+    getParam(accl_constraints_.ax_min, "ax_min", -2.0);
+    getParam(model_dt_, "model_dt", 0.05);
+  }
+
+  /**
    * @brief With input velocities, find the vehicle's output velocities
    * @param state Contains control velocities to use to populate vehicle velocities
    */
@@ -72,6 +88,9 @@ public:
     const bool is_holo = isHolonomic();
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
+        float max_feasible_vx = state.vx(i, j-1) + model_dt_*accl_constraints_.ax_max;
+        float min_feasible_vx = state.vx(i, j-1) + model_dt_*accl_constraints_.ax_min;
+        state.cvx(i, j - 1) = std::clamp(state.cvx(i, j - 1), min_feasible_vx, max_feasible_vx);
         state.vx(i, j) = state.cvx(i, j - 1);
         state.wz(i, j) = state.cwz(i, j - 1);
         if (is_holo) {
@@ -92,6 +111,10 @@ public:
    * @param control_sequence Control sequence to apply constraints to
    */
   virtual void applyConstraints(models::ControlSequence & /*control_sequence*/) {}
+
+protected:
+  float model_dt_{0.0};
+  models::AccelConstraints accl_constraints_{0.0, 0.0};
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -16,6 +16,7 @@
 #define NAV2_MPPI_CONTROLLER__MOTION_MODELS_HPP_
 
 #include <cstdint>
+#include <string>
 
 #include "nav2_mppi_controller/models/control_sequence.hpp"
 #include "nav2_mppi_controller/models/state.hpp"
@@ -63,8 +64,12 @@ public:
     ParametersHandler * param_handler)
   {
     auto getParam = param_handler->getParamGetter(name);
-    getParam(accl_constraints_.ax_max, "ax_max", 2.0);
-    getParam(accl_constraints_.ax_min, "ax_min", -2.0);
+    getParam(accel_constraints_.ax_max, "ax_max", 2.0);
+    getParam(accel_constraints_.ax_min, "ax_min", -2.0);
+    getParam(accel_constraints_.ay_max, "ay_max", 0.0);
+    getParam(accel_constraints_.ay_min, "ay_min", -0.0);
+    getParam(accel_constraints_.alphaz_max, "alphaz_max", 3.0);
+    getParam(accel_constraints_.alphaz_min, "alphaz_min", -3.0);
     getParam(model_dt_, "model_dt", 0.05);
   }
 
@@ -88,12 +93,19 @@ public:
     const bool is_holo = isHolonomic();
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        float max_feasible_vx = state.vx(i, j-1) + model_dt_*accl_constraints_.ax_max;
-        float min_feasible_vx = state.vx(i, j-1) + model_dt_*accl_constraints_.ax_min;
+        float max_feasible_vx = state.vx(i, j - 1) + model_dt_ * accel_constraints_.ax_max;
+        float min_feasible_vx = state.vx(i, j - 1) + model_dt_ * accel_constraints_.ax_min;
         state.cvx(i, j - 1) = std::clamp(state.cvx(i, j - 1), min_feasible_vx, max_feasible_vx);
         state.vx(i, j) = state.cvx(i, j - 1);
+
+        float max_feasible_wz = state.wz(i, j - 1) + model_dt_ * accel_constraints_.alphaz_max;
+        float min_feasible_wz = state.wz(i, j - 1) + model_dt_ * accel_constraints_.alphaz_min;
+        state.cwz(i, j - 1) = std::clamp(state.cwz(i, j - 1), min_feasible_wz, max_feasible_wz);
         state.wz(i, j) = state.cwz(i, j - 1);
         if (is_holo) {
+          float max_feasible_vy = state.vy(i, j - 1) + model_dt_ * accel_constraints_.ay_max;
+          float min_feasible_vy = state.vy(i, j - 1) + model_dt_ * accel_constraints_.ay_min;
+          state.cvy(i, j - 1) = std::clamp(state.cvy(i, j - 1), min_feasible_vy, max_feasible_vy);
           state.vy(i, j) = state.cvy(i, j - 1);
         }
       }
@@ -114,7 +126,7 @@ public:
 
 protected:
   float model_dt_{0.0};
-  models::AccelConstraints accl_constraints_{0.0, 0.0};
+  models::AccelConstraints accel_constraints_{0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
 };
 
 /**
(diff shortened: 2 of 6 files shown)


nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
```diff
@@ -41,6 +41,16 @@ struct SamplingStd
   float wz;
 };
 
+/**
+ * @struct mppi::models::AccelConstraints
+ * @brief Constraints on acceleration and deceleration
+ */
+struct AccelConstraints
```
SteveMacenski, line 48: I think we should include all the axes :-)

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -52,6 +53,21 @@ class MotionModel
     */
   virtual ~MotionModel() = default;
 
+  /**
+    * @brief Configure motion model on bringup and load required parameters
+    * @param name Name of plugin
+    * @param param_handler Parameter handler object
+    */
+  void on_configure(
```
SteveMacenski, line 61: As you mentioned in your comment, the constraint can be violated in the softmax. So, the acceleration constraints should be `get-ed` in the same place and stored alongside the other kino-dynamic constraints -- then passed into the `predict` function to use. That way we keep this all consistent and we have access to the acceleration constraints in the optimizer as well to clip in `applyControlSequenceConstraints` - which as you point out seems necessary

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -72,6 +88,9 @@ class MotionModel
     const bool is_holo = isHolonomic();
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
+        float max_feasible_vx = state.vx(i, j-1) + model_dt_*accl_constraints_.ax_max;
```
SteveMacenski, line 91: I'm curious what the runtime change of this action is. Can you benchmark the performance changes after adding in the wz / vy acceleration constraints

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -92,6 +111,10 @@ class MotionModel
    * @param control_sequence Control sequence to apply constraints to
    */
   virtual void applyConstraints(models::ControlSequence & /*control_sequence*/) {}
+
+protected:
+  float model_dt_{0.0};
+  models::AccelConstraints accl_constraints_{0.0, 0.0};
```
SteveMacenski, line 117: Misspell, just to fix when moved to the other file :-)

SteveMacenski (changes requested):
Really good first stab!

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: I agree with the consistency part that we get all the kino-dynamic constraints stored in the same place. But Optimizer has no use of acceleration constraints, since both `predict` and `applyControlSequenceConstraints` (which 
 calls `motion_model_->applyConstraints`) functions are part of MotionModel class. So, do we need to store it in Optimizer class?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: I benchmarked the performance changes by running the simulation using default tb3_simulation_launch.py.
 
MPPI parameters:
`time_steps: 56
 model_dt: 0.05
 batch_size: 2000
 ax_max: 3.5
 ax_min: -3.0
 ay_max: 0.0
 ay_min: -0.0
 thetaz_max: 3.0
 thetaz_min: -3.0`

**Without acceleration constraints:**
`[component_container_isolated-5] Total duration in ms = 10931
[component_container_isolated-5] No. of loops = 927
[component_container_isolated-5] Average time in ms = 11.7918
`
**With acceleration constraints:**
`[component_container_isolated-5] Total duration in ms = 12858
[component_container_isolated-5] No. of loops = 1001
[component_container_isolated-5] Average time in ms = 12.8452
`
The increment in the execution time of the control loop was around 12.84 - 11.79  = 1.05 ms averaging over 1000 loops.

Ayush1285: Benchmark results:

Running ./optimizer_benchmark
Run on (12 X 4100 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 9216 KiB (x1)
Load Average: 0.77, 1.97, 1.46

Iterations: 2
Batch size: 1000
Time steps: 50
Path points: 150

** With acceleration constraints for all axes**
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
BM_DiffDrivePointFootprint   4.87 ms         4.87 ms          115
BM_DiffDrive                         5.03 ms         5.03 ms          123
BM_Omni                              5.29 ms         5.29 ms          126
BM_Ackermann                    4.91 ms         4.91 ms          130

** Without acceleration constraints **
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
BM_DiffDrivePointFootprint   4.28 ms         4.28 ms          149
BM_DiffDrive                         4.58 ms         4.58 ms          142
BM_Omni                              4.79 ms         4.79 ms          118
BM_Ackermann                    5.59 ms         5.59 ms          100

nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
```diff
@@ -41,6 +41,20 @@ struct SamplingStd
...
+{
+  float ax_max;
```
SteveMacenski, line 50: Q: Can't these be added to `ControlConstraints` above? I think they count as control constraints

nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
```diff
@@ -41,6 +41,20 @@ struct SamplingStd
...
+  float ax_min;
+  float ay_max;
+  float ay_min;
+  float alphaz_max;
```
SteveMacenski, line 54:
```suggestion
  float az_max;
```

There's no reason to have angular velocity min and max. Its just one value - since its rotating left/right. There's no reason a vehicle would want to accelerate differently left than right (hence you see in ControlConstraints only 1 value for `wz`  - no min/max

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -52,6 +54,25 @@ class MotionModel
...
+    * @brief Configure motion model on bringup and load required parameters
+    * @param name Name of plugin
+    * @param param_handler Parameter handler object
+    */
+  void on_configure(
+    const std::string & name,
+    ParametersHandler * param_handler)
+  {
+    auto getParam = param_handler->getParamGetter(name);
+    getParam(accel_constraints_.ax_max, "ax_max", 2.0);
```
SteveMacenski, line 67: If added to ControlConstraints, then this can/should be removed to that location where the other ControlConstraints are populated. The motion model constructor can take in a pointer to the constraints object (so that values contained can be updated). Each motion model derived class can then decide if it wants to use it, but they all have access to it.

SteveMacenski (changes requested):

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: applyControlSequenceConstraints also clips the velocities (https://github.com/ros-navigation/navigation2/blob/bf291d704dbb1c2572575dd0620b12f816407233/nav2_mppi_controller/src/optimizer.cpp#L243-L244) shouldn't we also be clipping the accelerations?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: Can you average a 3-5 runs together to get an average value? I want to get a good sense of this. But overall seems not terrible. 

Please also:
- Precompute all the `model_dt_ * accel_constraints_.ax_max`'s (and similar) before the loop and use those variables. Given the size of the tensor, removing those multiplications from the loop can speed things up
- `state.wz(i, j - 1)` when you access things, store them as references (`float & wz_curr = state.wz(i, j - 1)`) and use those. When you do those `state.wz(i, j - 1)`, that's a lookup that adds compute time. Look up each once and use the stored value reference from then on.

Ayush1285: yes sure. I'll modify the code to avoid lookups. Just a quick question, why we aren't publishing the generated local trajectory(from final control sequences) to visualize it in RViz?

nav2_mppi_controller/include/nav2_mppi_controller/models/constraints.hpp
Ayush1285: Yes, that would be better.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: We are! Its done in another file though :-) Check out [the ](https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/src/trajectory_visualizer.cpp)

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: I think either would be sensible - but I prefer the way I suggested since that's in line with how other things are done for consistency. I think of the `model`'s apply constraints to be unique things to that motion model, and not things we always want to do

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: Instead of passing the constraints in the constructor, we can define a new function initialize(). Because if we pass it through the constructor, we will need to modify default constructors of DiffDriveMotionModel and OmniMotionModel. Similarly, explicit constructor of AckermannMotionModel takes parameter_handler as the parameter, we will need to modify that also. what is your suggestion this?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: makes sense.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: Without acceleration constraints:
| Benchmark | run 1 | run 2 | run 3 | run 4 | run 5 |  Average |
| --------------- | ------- | ------- | ------- | ------- | ------- | ------------ |
| BM_DiffDrivePointFootprint  | 4.75 ms | 4.21 ms | 4.23 ms | 4.84 ms | 4.31 ms  | 4.468 ms |
| BM_DiffDrive | 4.72 ms | 4.67 ms | 4.67 ms | 4.87 ms | 4.61 ms | 4.708 ms |
| BM_Omni  | 5.47 ms | 4.89 ms | 4.98 ms | 5.45 ms | 5.02 ms | 5.162 ms |
| BM_Ackermann | 4.45 ms | 5.35 ms | 5.34 ms | 4.52 ms | 5.37 ms | 5.005 ms

With acceleration constraints for all axes:
| Benchmark | run 1 | run 2 | run 3 | run 4 | run 5 |  Average | %change|
| --------------- | ------- | ------- | ------- | ------- | ------- | ------------ | --------------- |
| BM_DiffDrivePointFootprint  | 4.81 ms | 4.94 ms | 4.86 ms | 4.81 ms | 4.92 ms  | 4.868 ms | 8.95% |
| BM_DiffDrive | 5.00 ms | 5.07 ms | 5.05 ms | 5.10 ms | 5.09 ms | 5.062 ms | 7.51% |
| BM_Omni  | 5.36 ms | 5.24 ms | 5.37 ms | 5.43 ms | 5.24 ms | 5.328 ms | 3.21% |
| BM_Ackermann | 4.90 ms | 4.94 ms | 4.94 ms | 5.02 ms | 4.90 ms | 4.940 ms | -1.29% |

Ayush1285: @SteveMacenski any comments/suggestions on the benchmark results? I've also tried this to check the performance `for(unsigned int j = 1; j != state.vx.shape(1); j++) {
      auto vx_curr = xt::view(state.vx, xt::all(), j - 1);
      auto cvx_curr = xt::view(state.cvx, xt::all(), j - 1);
      xt::noalias(cvx_curr) = xt::clip(cvx_curr, vx_curr + min_delta_vx, vx_curr + max_delta_vx);
      xt::noalias(xt::view(state.vx, xt::all(), j)) = cvx_curr;

      auto wz_curr = xt::view(state.wz, xt::all(), j - 1);
      auto cwz_curr = xt::view(state.cwz, xt::all(), j - 1);
      xt::noalias(cwz_curr) = xt::clip(cwz_curr, wz_curr - max_delta_wz, wz_curr + max_delta_wz);
      xt::noalias(xt::view(state.wz, xt::all(), j)) = cwz_curr;

      if(is_holo) {
        auto vy_curr = xt::view(state.vy, xt::all(), j - 1);
        auto cvy_curr = xt::view(state.cvy, xt::all(), j - 1);
        xt::noalias(cvy_curr) = xt::clip(cvy_curr, vy_curr + min_delta_vy, vy_curr + max_delta_vy);
        xt::noalias(xt::view(state.vy, xt::all(), j)) = cvy_curr;
      }  
    }`
but it was much slower. If we somehow parallelize the computation for rows(1000-2000 rows), could that reduce the compute time significantly?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -70,12 +83,28 @@ class MotionModel
...
+    float min_delta_vx = model_dt_ * control_constraints_.ax_min;
+    float max_delta_vy = model_dt_ * control_constraints_.ay_max;
+    float min_delta_vy = model_dt_ * control_constraints_.ay_min;
+    float max_delta_wz = model_dt_ * control_constraints_.az;
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        state.vx(i, j) = state.cvx(i, j - 1);
-        state.wz(i, j) = state.cwz(i, j - 1);
+        float& vx_curr = state.vx(i, j - 1);
+        float& cvx_curr = state.cvx(i, j - 1);
```
SteveMacenski, line 94: I think you can store the `cvx_last` for `j - 1` so that you don't do redundant lookups 

If you have the loop doing: get last vx, get last cvx, clip, set vx. Then that clipped vx for that loop iteration is next loop's `j - 1`! That would reduce us back down to having no additional look ups which will probably fix the performance issue. 

You just need to make sure you set the initial values in the first for loop of the nested for look to set the conditions properly.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: See https://github.com/ros-navigation/navigation2/pull/4352#discussion_r1612242099, I think that should get us back down to sub ~1% increase hopefully

Ayush1285: benchmarks with a batch size of 2000, as results were more stable compared to a batch size of 1000.

Without acceleration constraints:
| Benchmark |  Average |
| --------------- | ---------- |
| BM_DiffDrivePointFootprint  | 11.22 ms |
| BM_DiffDrive |  12.06 ms |
| BM_Omni  | 14.66 ms |
| BM_Ackermann | 12.54 ms |

With acceleration constraints for all axes:
| Benchmark | Average |
| --------------- | ----------- |
| BM_DiffDrivePointFootprint  | 12.8 ms |
| BM_DiffDrive | 11.92 ms |
| BM_Omni  | 14.1 ms |
| BM_Ackermann |  12.36 ms |

After reducing all the redundant lookups, surprisingly compute time with acceleration constraints becomes lesser than the compute time without acceleration constraints. (Except for the 1st row, DiffDrive with Footprint).

osama-z-salah:
I have tried your PR, but it seems like there is an issue with achieving the maximum linear velocity (vx_max). Can you confirm if you have tested this with the robot, and did you observe any similar behavior?

Ayush1285:
> @Ayush1285 I have tried your PR, but it seems like there is an issue with achieving the maximum linear velocity (vx_max). Can you confirm if you have tested this with the robot, and did you observe any similar behavior?
I haven't tested the behaviour of maximum linear velocity. Ideally we shouldn't be having this problem. Can you please tell me what were the parameters you used? So that I can also test using same parameters.

Ayush1285:
> @Ayush1285 I have tried your PR, but it seems like there is an issue with achieving the maximum linear velocity (vx_max). Can you confirm if you have tested this with the robot, and did you observe any similar behavior?

@osama-z-salah  I tried with following parameters: vx_max = 0.5 m/s, ax_max = 2.0 m/s^2
Plot of vx with acceleration constraints  |  Plot of vx without acceleration constraints
:---------------------------------------------------:|:---------------------------------------------------------:
Screenshot from 2024-05-28 02-01-21 | 
Screenshot from 2024-05-28 02-06-26

nav2_mppi_controller/src/optimizer.cpp
```diff
@@ -77,6 +77,10 @@ void Optimizer::getParams()
   getParam(s.base_constraints.vx_min, "vx_min", -0.35f);
   getParam(s.base_constraints.vy, "vy_max", 0.5f);
   getParam(s.base_constraints.wz, "wz_max", 1.9f);
+  getParam(s.base_constraints.ax_max, "ax_max", 2.0f);
```
SteveMacenski, line 80: We should set these in the default Nav2 config file now that MPPI is default. Probably set it to something high so that it doesn't impact performance but can be bounded by the user when they want to.

SteveMacenski: We also need to add this to our migration guide that we added acceleration constraints to MPPI + to the MPPI's configuration guide on docs.nav2.org

SteveMacenski: Do me a favor as well and check on the signs of the acceleration and at least log a warning if the acceleration signs are wrong (i.e. ax_min being positive, for example) and set them to be rationally correct

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -70,12 +83,30 @@ class MotionModel
...
+    float vy_last = state.vy(0, 0);
+    float wz_last = state.wz(0, 0);
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        state.vx(i, j) = state.cvx(i, j - 1);
-        state.wz(i, j) = state.cwz(i, j - 1);
+        float & cvx_curr = state.cvx(i, j - 1);
+        cvx_curr = std::clamp(cvx_curr, vx_last + min_delta_vx, vx_last + max_delta_vx);
+        state.vx(i, j) = cvx_curr;
+        vx_last = cvx_curr;
```
SteveMacenski, line 98: The vx need to be rest between the nested for loops, not just before the loop. Otherwise we start with `vx_last` for sample `N+1` as the last vx for sample `N` rather than being the first sample of `N+1`

SteveMacenski (changes requested):

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: I think those are within the range of measurement noise - this is perfectly good, thanks for working with me to get this compute minimized :-)

SteveMacenski:
> It goes near to the max linear velocity with acceleration constraints also.

But it doesn't reach it. I've seen this before when I once tried to limit the velocities in the motion model to its maximum. I think what I determined at the time was that we clipped samples at the maximum speed so when we score + find the next optimal, it'll never achieve the maximum speed due to the soft-max averaging. This is why the constraint critic was introduced instead so we punish velocities out of bounds but don't bound the velocities being scored themselves. 

I think may require us to add the acceleration constraints to the ConstraintCritic - perhaps instead of doing it in the motion model. Or maybe spend some time thinking about why the acceleration is being limited artificially. Since this is on acceleration, I wouldn't have expected this issue to occur and this is where the MPPI technique applies the dynamics models to roll out control velocities into actual velocity candidates. 

With that said, I also pointed out an actual computational bug in my last review, so its possible this issue just goes away after that's fixed. I'll also take a look over the code again with this in mind to see if I don't see something else. 

---

Linting:

```
Diff with 5 lines
--- include/nav2_mppi_controller/motion_models.hpp
+++ include/nav2_mppi_controller/motion_models.hpp.uncrustify
@@ -130 +130 @@
-   0.0f};
+    0.0f};
```

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -70,12 +83,30 @@ class MotionModel
...
+    float max_delta_wz = model_dt_ * control_constraints_.az_max;
+    float vx_last = state.vx(0, 0);
+    float vy_last = state.vy(0, 0);
+    float wz_last = state.wz(0, 0);
     for (unsigned int i = 0; i != state.vx.shape(0); i++) {
       for (unsigned int j = 1; j != state.vx.shape(1); j++) {
-        state.vx(i, j) = state.cvx(i, j - 1);
-        state.wz(i, j) = state.cwz(i, j - 1);
+        float & cvx_curr = state.cvx(i, j - 1);
+        cvx_curr = std::clamp(cvx_curr, vx_last + min_delta_vx, vx_last + max_delta_vx);
```
SteveMacenski, line 96: Oh. You're modifying the underlying `cvx` asset to be in the dynamic constraints. Change all the `float &` to a `float` so they you don't modify `cvx` and see if that fixes the issue. You shouldn't modify the underlying control requested, just the velocity output that the control represents after applying the motion model

SteveMacenski (changes requested):

Ayush1285:
> > It goes near to the max linear velocity with acceleration constraints also.
> 
> But it doesn't reach it. I've seen this before when I once tried to limit the velocities in the motion model to its maximum. I think what I determined at the time was that we clipped samples at the maximum speed so when we score + find the next optimal, it'll never achieve the maximum speed due to the soft-max averaging. This is why the constraint critic was introduced instead so we punish velocities out of bounds but don't bound the velocities being scored themselves.
> 
> I think may require us to add the acceleration constraints to the ConstraintCritic - perhaps instead of doing it in the motion model. Or maybe spend some time thinking about why the acceleration is being limited artificially. Since this is on acceleration, I wouldn't have expected this issue to occur and this is where the MPPI technique applies the dynamics models to roll out control velocities into actual velocity candidates.
> 
> With that said, I also pointed out an actual computational bug in my last review, so its possible this issue just goes away after that's fixed. I'll also take a look over the code again with this in mind to see if I don't see something else. That bug though would cause some issues so it wouldn't surprise me if that's just it.
> 
> Linting:
> 
> ```
> Diff with 5 lines
> --- include/nav2_mppi_controller/motion_models.hpp
> +++ include/nav2_mppi_controller/motion_models.hpp.uncrustify
> @@ -130 +130 @@
> -   0.0f};
> +    0.0f};
> ```

The problem was due to the above bugs only, thanks for catching them :)
Screenshot from 2024-05-29 00-51-44
Now it goes up to maximum velocity.

Screenshot from 2024-05-29 00-52-49
This is the plot of acceleration, it remains within specified constraints(2.0 m/s^2) for all iterations. Except 1, where it spikes to 2.5

SteveMacenski:
@Ayush1285 to fix that, we just add to the `applyControlConstraints` which I think you haven't done yet. That should square all this up for merging with the other comments above :-)

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: But in our case, the control output is the desired velocity, and `control_sequence` is calculated by performing softmax on `cvx`. So, if we don't constrain `cvx`, then it may result in acceleration crossing the bounds.

Ayush1285: for that, we will need to clamp the final control sequence as you said. But I'm not sure what will be the effect of applying hard constraints on the final control sequence.

SteveMacenski: Well first - is this what fixes the acceleration problem? Or was it the other things that fixed it.

> So, if we don't constrain cvx, then it may result in out of bounds acceleration.

Ah yes. This is back to an old question I had: [should these be `cvx` or be `vx`?](https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/src/optimizer.cpp#L378-L382) Previously it didn't matter since they were the same, but now they're not. I postulate that if the dynamic model applies is predictive (i.e. if we apply _this_ control, _this_ velocity occurs) then it should be `cvx`. But, if the dynamic model is applying restrictions, then I think it should be `vx` so we are using the restricted velocities.

> we will need to clamp the final control sequence as you said

We should do that irregardless. However, I think if we just clamp while using `cvx` that could create differences between the scored and optimal softmaxed trajectories. We should attempt to minimize or eliminate that difference as much as possible. Relying on that clamp should be dealing with small rounding errors, not major differentiations. 

So options...

- If changing the `float &` to `float` isn't what fixed your issues, then revert that? But I suspect that **is** what fixed your issues
- Change the softmax application to the velocities, not the controls
- Keep using the controls for the softmax but  add the acceleration constraints to the `ConstraintCritic` and score breaches negatively. That will minimize the breaches like we do with velocities so that when we do hit the `applyControlSequenceConstraints`, we're applying it to only rounding error levels. We have a recipe for that with the  velocities that works. 

It may be worth trying a couple of these to see what works the best

Ayush1285: > If changing the float & to float isn't what fixed your issues, then revert that? But I suspect that is what fixed your issues

It was due to that bug in setting vx_last. I've tried running with float& again, it reaches maximum velocity.
Do we have existing full integration tests for navigation through poses or waypoint following?

SteveMacenski: > It was due to that bug in setting vx_last. 

Oh, ok... so it goes full speed again? Umm.. I think that works then!

Ayush1285: > Oh, ok... so it goes full speed again? Umm.. I think that works then! Can you show me the plots like you did before?

Yes sure!
Screenshot from 2024-05-30 00-36-17

This is the global plan on which I'm testing
Screenshot from 2024-05-28 01-59-52

SteveMacenski: Oh ok! Well that works then I think!

SteveMacenski:
I believe documentation on docs.nav2.org is the only thing left? The configuration + migration guide

Ayush1285:
PR for the docs update: https://github.com/ros-navigation/docs.nav2.org/pull/562

nav2_mppi_controller/src/optimizer.cpp
```diff
@@ -77,11 +77,23 @@ void Optimizer::getParams()
...
+  getParam(s.base_constraints.ax_min, "ax_min", -3.0f);
+  getParam(s.base_constraints.ay_max, "ay_max", 3.0f);
+  getParam(s.base_constraints.az_max, "az_max", 3.5f);
   getParam(s.sampling_std.vx, "vx_std", 0.2f);
   getParam(s.sampling_std.vy, "vy_std", 0.2f);
   getParam(s.sampling_std.wz, "wz_std", 0.4f);
   getParam(s.retry_attempt_limit, "retry_attempt_limit", 1);
 
+  s.base_constraints.ax_max = std::abs(s.base_constraints.ax_max);
+  if (s.base_constraints.ax_min > 0) {
```
SteveMacenski, line 90:
```suggestion
  if (s.base_constraints.ax_min > 0.0) {
```

SteveMacenski (approved):

SteveMacenski:
Nitpick and I can merge (you have your branch setup so I can't push it)

RBT22:
What do you think about the practical usability limits of this approach?
Similarly to my previous attempts implementing acceleration limits to MPPI, I experienced unpredictable behaviors. Lowering the acceleration limits causes the robot (TB3, Iron) to sway, stutter, or completely disregard it's path. This behavior can be observed from around <1.0 m/s^2.

Setting accelerations to 0.5 m/s^2 resulted in the following for me:

Ayush1285:
> What do you think about the practical usability limits of this approach? Similarly to my previous attempts implementing acceleration limits to MPPI, I experienced unpredictable behaviors. Lowering the acceleration limits causes the robot (TB3, Iron) to sway, stutter, or completely disregard it's path. This behavior can be observed from around <1.0 m/s^2.
> 
> Setting accelerations to 0.5 m/s^2 resulted in the following for me:  [  ](https://private-user-images.githubusercontent.com/43723477/335610504-f795d4d6-dae9-401f-afb6-cfbc4edb4cc3.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTcxNzcyMTEsIm5iZiI6MTcxNzE3NjkxMSwicGF0aCI6Ii80MzcyMzQ3Ny8zMzU2MTA1MDQtZjc5NWQ0ZDYtZGFlOS00MDFmLWFmYjYtY2ZiYzRlZGI0Y2MzLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MzElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTMxVDE3MzUxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWFmZTkxMmUzZWMxMWI4OGU4ZTUxODYxNWExNTc5MWNkYTEzOWE2MTc2NDExYWM4ZmRkNWE0Y2E5N2IxYWI3OWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.jBrF5tHhl-vdRKbFs7XPUw1HMzE1HCI_Q3kphRVF-jE) [ ](https://private-user-images.githubusercontent.com/43723477/335610504-f795d4d6-dae9-401f-afb6-cfbc4edb4cc3.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTcxNzcyMTEsIm5iZiI6MTcxNzE3NjkxMSwicGF0aCI6Ii80MzcyMzQ3Ny8zMzU2MTA1MDQtZjc5NWQ0ZDYtZGFlOS00MDFmLWFmYjYtY2ZiYzRlZGI0Y2MzLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MzElMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTMxVDE3MzUxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWFmZTkxMmUzZWMxMWI4OGU4ZTUxODYxNWExNTc5MWNkYTEzOWE2MTc2NDExYWM4ZmRkNWE0Y2E5N2IxYWI3OWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.jBrF5tHhl-vdRKbFs7XPUw1HMzE1HCI_Q3kphRVF-jE)

Above 2 m/s^2 should be fine I guess, I've also faced similar issue with lower acceleration bounds.

Ayush1285:
@SteveMacenski what do you think about this stalling behaviour of bot at low acceleration bounds? It occurs even at 1.5 m/s^2.
[Screencast from 05-31-2024 11:10:57 PM.webm](https://github.com/ros-navigation/navigation2/assets/64409716/010d9468-27cf-4270-95c5-543fefe292de)

SteveMacenski:
What do those acceleration graphs look like in RQT? 

> Lowering the acceleration limits causes the robot (TB3, Iron) to sway, stutter, or completely disregard it's path. This behavior can be observed from around <1.0 m/s^2.

Wrt your videos, did you try retuning? It seems like the robot just doesn't want to leave the zero-costed space, which makes sense to me with low constraints. If you remove the obstacle critic (just for experimentation), does it go back to normal? It may be that we just need to tweak the cost/obstacle critic defaults. 

---

I'm noting the the [TB3 params in DWB](https://github.com/ros-navigation/navigation2/blob/humble/nav2_bringup/params/nav2_params.yaml#L153-L155) (also used by linobot) were set as 2.5 m/s2 and 3.2 rad/s2 which are above the threshold for concern? I'm doing a quick survey and see [joystick teleop for fetch is 1.0m/s2](https://github.com/ZebraDevs/fetch_ros/blob/melodic-devel/fetch_teleop/src/joystick_teleop.cpp#L108), [clearpath husky is 3.0m/s2](https://answers.ros.org/question/335763/what-is-the-max-linear-acceleration-of-husky-a200/). 

That's a sampling of robots big, small; indoor and outdoor. I think in truly low acceleration applications that this is perhaps not the silver bullet solution, but I don't think many robots require sub-1m/s2 acceleration constraints so as long as this works for the "normal" range of values, I think we're fine (?). For really weird, specialized situations, there are going to be more optimal algorithms to help you, so I think super-low acceleration counts as part of that "weird, specialized" class of robot systems which not many fall into. 

Imagine you had a robot with 0.1 m/s2 acceleration constraints. If the path tracker handled that perfectly, that's such a slow acceleration profile I think you would **strictly need** to have a planner that takes into account the dynamic constraints into its formulation because you couldn't just make a turn, you'd have to slow down waaaay before that turn to turn super duper slowly, then ramp back up. So between 0.1 and 1 m/s2 (where we handle well I'm told above), where do we draw that line where its no longer an effective general acceleration profile for path tracking requiring some specialized development to make plausible?

Ayush1285:
> But regardless, I think the issue your videos show is related to tuning it seems? Something like 0.5m/s2 should probably work (especially at low maximum speeds). 

yeah, after removing the Cost critic, it works even at 0.5 m/s^2.

SteveMacenski:
> yeah, after removing the Cost critic, it works even at 0.5 m/s^2.
> yes, it works fine if I remove the Cost Critic.

Seems then just like a tuning thing, I'm not worried about it then! 

I've asked another maintainer to review and give his feedback since this is something they will plan to use as well. Besides that, anything else blocking merging?

Ayush1285:
> > yeah, after removing the Cost critic, it works even at 0.5 m/s^2.
> > yes, it works fine if I remove the Cost Critic.
> 
> Seems then just like a tuning thing, I'm not worried about it then!
> 
> I've asked another maintainer to review and give his feedback since this is something they will plan to use as well. Besides that, anything else blocking merging?

I've just updated the test for the positive value of ax_min constraints. I don't think anything else is remaining. I just wanted to run system tests to be sure.

doisyg:
Great to see work is being conducted on this, thank you all. I had a first look.
First a bit of context. My robotic platform is a massive 600 kg one with specific kinematic constraints: 0.25m/s2 for acceleration and 0.5m/s2 for deceleration on the linear axis. And both 1.2rad/s2 of accel/deccel on rotation (Z). And it is fully bidirectional.

I did not properly dig on the implementation, but from the parameters available:
- ax_min / ax_max
- ay_max
- az_max

it looks like this PR intent is to deal only with acceleration, not deceleration (or I will assume to see more parameters for ay and az). 
If the intend here is to have different linear accelerations limits when moving forward and backward, then you should check that `ax_min` doesn't become the deceleration limit when moving forward, and symmetrically  that `ax_max` doesn't become the deceleration limit when moving backward. (I believe that's the case here).

However, I would be much in favor in dealing with decceleration properly, in a similar way as it is done for the velocity smoother, but be careful, it is easy to mix up when also dealing with bi-directionality. See for https://github.com/ros-navigation/navigation2/pull/3529 instance.

Finally I gave it a spin without setting the new parameters, to check that it was not breaking existing tuning. It looks okay. I did not test yet the behavior when setting the parameters as it is unclear who should be set / expected for my bi-directional platform when different accel and deccel constraints.

Ayush1285:
> If the intend here is to have different linear accelerations limits when moving forward and backward, then you should check that ax_min doesn't become the deceleration limit when moving forward, and symmetrically that ax_max doesn't become the deceleration limit when moving backward.

I think the naming of the parameters may be confusing. They are not the limits of acceleration while moving forward and backward. `ax_max` is maximum acceleration along the x-axis, and `ax_min` is maximum deceleration(braking power).

SteveMacenski:
@doisyg I think @Ayush1285 did a good job explaining. Would you want different constraints for reversing acceleration/deceleration than forward direction? I can see the sense in setting different velocity bounds for robots with a directional preference, but it didn't occur to me someone might want higher order things to also have different limits for forward vs reverse.

doisyg:
Thank you both for explaining!
I don't think we need different constraints for forward and backward X movement (at least I don't). But definitely different accel and deccel constraints. So all good.
I don't need different accel/deccel for rotation, but I could see someone needing it. We don't necessarily need to address that now, but just noting.

Another point to note, this PR will not help smoothing the transition between speeds generated by MPPI and the 0.0 twist sent by the controller server once the goal checker kicks in. But that's totally normal (for me this is handled by the velocity smoother). I will make an experiment later to check if this work can actually help achieving better goal accuracy by taking into account the decel constraint of the robot (and with smaller goal checker thresholds), which would be amazing.

If you are all happy, despite my limited testing, this is good to go. I mainly tested that it doesn't break the current default behavior. About the general impact of adding appropriately tuned kinematic constraints, that will need more continuous testing on my end, but I can assess after merging and a couple of days/weeks of live run.

SteveMacenski:
Sounds good!

SteveMacenski:
Thanks @Ayush1285 for this awesome work, it is much appreciated and will shortly be in use by a number of companies worldwide! This has been a big ask for MPPI! Any interest in other MPPI improvement projects (https://github.com/ros-navigation/navigation2/issues/3351) or other things in Nav2 I can help you through?

Ayush1285:
> Thanks @Ayush1285 for this awesome work, it is much appreciated and will shortly be in use by a number of companies worldwide! This has been a big ask for MPPI! Any interest in other MPPI improvement projects (#3351) or other things in Nav2 I can help you through?

Thanks Steve for your insights on performance improvement and other reviews. 
Further, I am interested in trying out Eigen for MPPI and potential optimization of collision checker object. I can work on these two items.

SteveMacenski:
Fantastic! There's been a chat on Eigen in https://github.com/ros-navigation/navigation2/issues/4237 if you hadn't already seen that is good to reference. I think that's the most bang-for-your-buck thing to do so I'd recommend doing that first (the collision checker object also important, but I think Eigen could be a 20-30% improvement, whereas collision checker 10-20%. Still high, but this is higher)

A user said they had good experiences once actually using the Eigen tensor library and the blog post (https://romanpoya.medium.com/a-look-at-the-performance-of-expression-templates-in-c-eigen-vs-blaze-vs-fastor-vs-armadillo-vs-2474ed38d982) makes me feel very optimistic it can help. I've profiled MPPI in detail many times and a non-trivial amount of time is spent now fully internal to xtensor views and strides. If Eigen can do that even 5% faster, that's ~2-3% improvements in the performance. And those metrics claim it could be up to 8x faster. So we could potentially double performance overnight conceptually (though I'd settle for much less :laughing: )

Ayush1285:
> Fantastic! There's been a chat on Eigen in #4237 if you hadn't already seen that is good to reference. I think that's the most bang-for-your-buck thing to do so I'd recommend doing that first (the collision checker object also important, but I think Eigen could be a 20-30% improvement, whereas collision checker 10-20%. Still high, but this is higher)
> 
> A user said they had good experiences once actually using the Eigen tensor library and the blog post (https://romanpoya.medium.com/a-look-at-the-performance-of-expression-templates-in-c-eigen-vs-blaze-vs-fastor-vs-armadillo-vs-2474ed38d982) makes me feel very optimistic it can help. I've profiled MPPI in detail many times and a non-trivial amount of time is spent now fully internal to xtensor views and strides. If Eigen can do that even 5% faster, that's ~2-3% improvements in the performance. And those metrics claim it could be up to 8x faster. So we could potentially double performance overnight conceptually (though I'd settle for much less 😆 )

Thanks!! I'll check the above links and start on them in my leisure time.
> though I'd settle for much less 😆

That's good to know :laughing:

Ayush1285:
@SteveMacenski  Where can we continue the discussion on Eigen for MPPI?  I've tried the first draft(in my fork repo), which is not optimised currently.

SteveMacenski:
@Ayush1285 wow, that was quick and sounds awesome! I think this thread is the right spot! https://github.com/ros-navigation/navigation2/issues/4237

Issue #4345, opened 2024-05-15
New "Configure" transition for the LifecycleManager


jonipol:
Hello!

## Feature request

The current implementation of the LifecycleManager only allows to transition all the nodes from _unconfigured_ to _active_ which causes some issues in certain cases at startup for the system we are working on. If one would add the transition from _unconfigured_ to _inactive_ as "CONFIGURE" would you approve such contribution? Or is there some problem with that which I have not been able to see?

#### Feature description

Add a CONFIGURE transition to the LifecycleManager which transitions all the managed nodes from _unconfigured_ to _inactive_ state.

#### Implementation considerations

Requires:
- Adding a new command to _ManageLifecycleNodes.srv_ eg. `uint8 CONFIGURE = 5`
- Handle the case in the lifecycle_manager.cpp

SteveMacenski:
Seems reasonable enough - take a look at the lifecycle manager and lifecycle manager client, I think that you'll see its easy that we can exposure the full spectrum of lifecycle states in addition to the "composed" transitions that we do for a 1-button startup. I think that would be a great contribution and I would be happy to review / merge that PR :smile: 

You'd just have to call the manage nodes service 2x for bring up / shutdown using the rclcpp lifecycle transition types rather than our "startup" and "shutdown". 

Just make sure to expose them all so that you can bring up and shut down through the appropriate state transitions -- not just configure, but also activate, deactivate, etc.

SteveMacenski:
PR to be merged imminently

PR #4371, opened 2024-05-27
Add configure and cleanup transitions to lifecycle manager and client

jonipol:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4345  |
| Robotic platform tested on | Gazebo simulation, Leo Rover, Robauta Sampo |

---

* Added CONFIGURE and CLEANUP commands for lifecycle_manager and lifecycle_manager_client

I was not able to find any documentation regarding to the available lifecycle transitions. If some exists, would need to that,

---

Might be nice to have a test which makes sure that calling shutdown when the nodes are in inactive state does the shutdown properly. I did some testing locally and seems to be working as expected. I tried to add such test into the test_lifecycle_manager.cpp but a problem there is that the shutdown cannot be called multiple times in the same test file, The test environment is not reset between tests since the lifecycle_manager is span up with the launch file. I did not want to write another test file (+ maybe a new launch file) for that 3 line test. Also attempted to change the way the test is ran to have the control over the lifecycle_manager with test fixture but was not successful with that.

#### For Maintainers: 

### DIFF
--- a/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
+++ b/nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
@@ -105,6 +105,16 @@ protected:
    */
   bool startup();
   /**
+   * @brief Configures the managed nodes.
+   * @return true or false
+   */
+  bool configure();
+  /**
+   * @brief Cleanups the managed nodes
+   * @return true or false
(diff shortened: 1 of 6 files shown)


SteveMacenski (approved):
Easiest PR I've been able to merge all month, thanks! Its very clear and well implemented with the file's conventions!

PR #4372, opened 2024-05-28
Fix flaky smoother test

SteveMacenski:
Removes tolerance so it should now work deterministically if I understand the issue correctly


SteveMacenski:
This is ultimately caused by the new analytic expansion technique checking until the end point. Thus, the last point and the goal node can be within floating point error of each other which causes an additional segment to be added when overlapping. This last duplicate point is removed in the planner plugins by checking if there is close overlap but that is not performed in the test. Thus, we can accept either 2 or 3 and either is valid as long as the planner plugins are properly interpreting that duplicate result

PR #4389, opened 2024-05-31
Persist bash history in devcontainer

tonynajjar:
Persist the bash history of devcontainer in a local volume

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -34,6 +34,11 @@
             "source": "overlay-${devcontainerId}",
             "target": "/opt/overlay_ws",
             "type": "volume"
+        },
+        {
+            "source": "nav2-bashhistory",
+            "target": "/commandhistory",
+            "type": "volume"
         }
     ],
     "features": {
--- a/Dockerfile
+++ b/Dockerfile
@@ -162,6 +162,11 @@ FROM caddy:builder AS caddyer
 RUN xcaddy build \
     --with github.com/caddyserver/replace-response
 
+# Save bash history
+# https://code.visualstudio.com/remote/advancedcontainers/persist-bash-history
+RUN SNIPPET="export PROMPT_COMMAND='history -a' && export HISTFILE=/commandhistory/.nav2_bash_history" \
+&& echo "$SNIPPET" >> "/root/.bashrc"
+
 # multi-stage for visualizing
 FROM dever AS visualizer
 


SteveMacenski (approved):
I leave final approval to @ruffsl

tonynajjar:
@SteveMacenski what's up with the tests now :sob:

SteveMacenski:
> @SteveMacenski what's up with the tests now 😭 Why is this showing up now

Welcome to my life as a Nav2 maintainer :upside_down_face: :laughing:

tonynajjar:
@SteveMacenski I added one more thing: auto-sourcing of overlay if it exists, otherwise source underlay

tonynajjar:
In my own projects I generate docker images for every PR to be able to easily "checkout" a PR (code + environment). I guess we don't do that here right? It would have helped me test this PR better. Is this an idea you'd be open to? If you're worried about CI time, it can be triggered manually

nav2_bashrc
```diff
@@ -0,0 +1,14 @@
+OVERLAY_DIR=/opt/overlay_ws
```
SteveMacenski, line 1: This should be in `/tools` not in the root of the repository please

tonynajjar: Ah nice, I was looking for a better place

SteveMacenski: `/tools` is our catch-all for random developer tools

tools/nav2_bashrc
```diff
@@ -0,0 +1,14 @@
...
+# https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html#source-the-overlay
+OVERLAY=$OVERLAY_DIR/install/setup.bash
+UNDERLAY=/opt/ros/${ROS_DISTRO}/setup.bash
+if [ -f "$OVERLAY" ]; then
+    source $OVERLAY
+else
+    source $UNDERLAY
+fi
+
+# https://code.visualstudio.com/remote/advancedcontainers/persist-bash-history
```
ruffsl, line 11: I see the vscode docs recommends to bake this into the Dockerfile, but I think that's just adds unnecessary opportunities to bust the docker layer cache when it doesn't have to. Please add such user session config to the `postCreateCommand` where we already manage things like .bashrc.

https://github.com/ros-navigation/navigation2/blob/d302f1ad53e0af1d55ddc94214ade7e5f9010f44/.devcontainer/post-create-command.sh#L10-L11  

- https://containers.dev/implementors/spec/#post-container-creation

tools/nav2_bashrc
```diff
@@ -0,0 +1,14 @@
+OVERLAY_DIR=/opt/overlay_ws
+
+# https://docs.ros.org/en/rolling/Tutorials/Workspace/Creating-A-Workspace.html#source-the-overlay
+OVERLAY=$OVERLAY_DIR/install/setup.bash
+UNDERLAY=/opt/ros/${ROS_DISTRO}/setup.bash
+if [ -f "$OVERLAY" ]; then
```
ruffsl, line 5: Don't think we should add this kind of logic to a one off script. Let's add an alias to the .bashrc instead, as we still may need to open a shell without automatically sourcing a workspace.

.devcontainer/devcontainer.json
```diff
@@ -34,6 +34,11 @@
             "source": "overlay-${devcontainerId}",
             "target": "/opt/overlay_ws",
             "type": "volume"
+        },
+        {
+            "source": "nav2-bashhistory",
+            "target": "/commandhistory",
```
ruffsl, line 40: When I overhaul the user account setup for the devcontainer, we can track such dot files easier than using root level paths.

ruffsl (changes requested):

tonynajjar:
@ruffsl if you can incorporate the idea of a persistent bash history in your PR I'm happy to close this.

tools/nav2_bashrc
tonynajjar: > as we still may need to open a shell without automatically sourcing a workspace

I'm curious to know which use cases would be affected by the auto-sourcing of the workspace? 

Adding an alias to the bashrc doesn't add much more value to manually doing it (once it's done once and in the bash history, it's easy to pull up).

ruffsl:
> if you can incorporate the idea of a persistent bash history in your PR

Feel free to give it a try, I generalized it for any dotfile under the default home directory inside the devcontainer:
- https://github.com/ros-navigation/navigation2/pull/4392
- https://github.com/ros-navigation/navigation2/pull/4392/commits/667a8ff2a13abe7d60438acef1e4d141efa99e97

I prefer to use fish shell or sometime zsh, so this work for those use cases too.

tonynajjar:
Will give it a try as soon as I can!

Issue #4364, opened 2024-05-23
[collision_monitor] Add temporal axis to min_points behavior


gennartan:
#### Feature description

The current behavior of `min_points` parameter is to apply a speed reduction (or whatever action is required by the collision monitor) when the last valid measurement point is triggers this behavior. In case of noisy measurement, one single measurement could cause the robot to slow down quickly, causing jerky motion.

The idea to add a temporal axis to this parameter would be to require a minimum of X positive consecutive measurement to trigger the collision monitor to slow / stop the robot. And also a minimum of Y negative consecutive measurement to continue normally.

#### Implementation considerations

Another idea could be to filter the output of the measurement source. But this would add imprecision and timing in the actual measurement. I am thinking of of robot driving on grass. Simply depending on wind conditions, ultrasonic sensors could detect leaves or higher weeds occasionally even with correctly calibrated / positioned sensors. So I think that adding this feature directly on the collision monitor could give the best results.

SteveMacenski:
> In case of noisy measurement, one single measurement could cause the robot to slow down quickly, causing jerky motion.

I suppose historically the kinds of sensors used for collision monitoring shouldn't be super noisy to cause such issues. However, given this isn't a safety-rated solution and it is useful for robots that may **only** have noisy sensors that are unable to be finely filtered due to computational / BOM considerations, I think this is a very good suggestion.

> Another idea could be to filter the output of the measurement source.

I think that should be done by the user if that's the right answer so they can filter it as appropriate for their sensor modality / environment / application. Else, this project could have embedded a ton of unique filters that should probably just stand alone. As you mention with your grass example, thats would require a pretty unique filter, so its best that there's a node between the data & the collision monitor to perform that filtering pre-collision-monitoring.  

So, I like adding the temporal element! And the filter bit is definitely still an option (maybe we can document that potential pipeline design?) in addition to temporal for those that that is a good solution. 

Are you open to contributing that?

gennartan:
Yes I am open to contributing that ! I'll get started soon

Shakthi-Bala:
Hi, I’d be happy to take this issue on. Please let me know if it’s okay to proceed. Thanks!

SteveMacenski:
Please!

Shakthi-Bala:
Thanks for the response, I will update this week

mini-1235:
@Shakthi-Bala any updates?

Neomelt:
Hi maintainers, a quick update on #4364.

[...]

PR #4390, opened 2024-06-01
Temporal min points

gennartan:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (#4364) |
| Robotic platform tested on | Tested using colcon test only|

---

* Add 2 new parmeters `min_consecutive_in` and `min_consecutive_out` that adds a temporal axis to the `min_points` parameter.

* Added 2 new parameters, so need to add that to default configs and documentation page

---

* The `min_consecutive_out` parameter does not work for the `PolygonApproach`  type because the collision time is computed with the latest sensor measurements. So whenever a measurement is outside the range, the collision time will be `-1` and we will not be able to keep the robot at a slower speed. So it would be required to remember the previous `collision_time` and take an appropriate decision (keep the worst, keep the best, ... ?) 

#### For Maintainers: 


SteveMacenski:
Make sure to pull in main so that CI works for the Rotation Shim fix

nav2_collision_monitor/src/collision_monitor_node.cpp
```diff
@@ -571,7 +573,8 @@ bool CollisionMonitor::processApproach(
 
   // Obtain time before a collision
   const double collision_time = polygon->getCollisionTime(collision_points, velocity);
-  if (collision_time >= 0.0) {
+  polygon->newDecisionPoint(collision_time >= 0.0);
```
SteveMacenski, line 576: I don't think this is sensible for the approach polygon, no?

nav2_collision_monitor/src/polygon.cpp
```diff
@@ -229,6 +230,28 @@ int Polygon::getPointsInside(const std::vector<Point> & points) const
   return num;
 }
 
+void Polygon::newDecisionPoint(bool is_in)
+{
+  if (is_in && current_consecutive_in_ < min_consecutive_in_) {
+    current_consecutive_in_++;
+    current_consecutive_out_ = 0;
+    if (current_consecutive_in_ >= min_consecutive_in_) {
+      current_decision_ = true;
```
SteveMacenski, line 239: I think the current state needs to be reset on `activate` so that you can activate / deactivate and have it work without stale measurements

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
```diff
@@ -145,6 +145,19 @@ class Polygon
    */
   virtual int getPointsInside(const std::vector<Point> & points) const;
 
+  /**
+   * @brief Add a new decision point to count the consecutive number of
+   * time enough point are in the polygon. Updates the current decision value
+   * based on previous measurements.
+   */
+  void newDecisionPoint(bool is_in);
```
SteveMacenski, line 153: On naming: this is a little ambiguous as to the what decision it is. Can you come up with some alternatives?

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
```diff
@@ -251,6 +264,10 @@ class Polygon
   ActionType action_type_;
   /// @brief Minimum number of data readings within a zone to trigger the action
   int min_points_;
+  /// @brief Minimum number of consecutive data with minimum min_points to trigger the action
+  int min_consecutive_in_;
```
SteveMacenski, line 268: On naming: this a little ambiguous as a variable / parameter name. Can you come up with some alternatives?

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
```diff
@@ -251,6 +264,10 @@ class Polygon
...
+  /// @brief Minimum number of consecutive data without minimum min_points to stop the action
+  int min_consecutive_out_;
```
SteveMacenski, line 270: Thoughts about a duration rather than a number of consecutive cycles? I would think that having the system blocked for > 0.5 seconds or something is more intuitive than 5 cycles, or what have you. I think that would be more intuitive to me, but open to leaving as is if others prefer this (@tonynajjar ?)

SteveMacenski (changes requested):

nav2_collision_monitor/include/nav2_collision_monitor/polygon.hpp
tonynajjar: I agree that specifying a duration would be more intuitive especially because the CM and CD's run on a rate different than the rate of the measurements. For CD it's a configurable rate and for CM it's the rate of the cmd_vel. The "meaning" of `min_consecutive_out_` is dependent on the difference between the CM/CD rate and the measurement rate, if that makes sense.

TLDR: I would vote for duration

SteveMacenski: @gennartan what do you think?

gennartan: Specifying a duration is a better idea, I will update the PR soon and come back with better names for the variables

SteveMacenski:
@gennartan I see that you pushed some updates, should I take a look and review or just incremental updates? 

CI failed ... oddly, might be worth rebasing your PR, but I retriggered CI to see if it was just a fluke. Note the linting test failures from `ament_cpplint` and `ament_uncrustify`, those are important

gennartan:
Hi, just incremental updates for now, I still need to add some testing to it. I will check why the tests are failling and fix that.

Do you know if there is a method to create tests with precise timing to simulate the clock in ROS ?

SteveMacenski:
Ok!

> Do you know if there is a method to create tests with precise timing to simulate the clock in ROS ?

I think an illustrative example of what you’re looking to do might help me give a suggestion. I’m not sure that should be required here, but maybe we’re thinking about different definitions of ‘precise’. Is using the clock’s stamp insufficiently accurate?

gennartan:
> I think an illustrative example of what you’re looking to do might help me give a suggestion. I’m not sure that should be required here, but maybe we’re thinking about different definitions of ‘precise’. Is using the clock’s stamp insufficiently accurate?

My question is more about the clock control in the test. If I want to test if the polygon action is correctly applied after 10s of positive detection, I don't want to wait for 10 actual seconds on the system running the tests using a sleep function, I want to be able to change the clock vlaue to simulate those 10 seconds. 

Do you know if such a function exists ?

SteveMacenski:
> I don't want to wait for 10 actual seconds on the system running the tests using a sleep function, I want to be able to change the clock value to simulate those 10 seconds.

You could set simulation time and have your own `/clock` publisher, but that seems like overkill to bake into the test. I suggest making some of the times smaller values like 2-3 seconds instead. Keep the test times down whereever reasonably possible, but don't go overboard.

SteveMacenski:
Hi @gennartan how can I help push this along? :smile:

gennartan:
Hi, sorry, I am a bit blocked :smile: 
My main concern is that I don't want that feature to cause bugs, and I recently had a few bugs with time jumps with my current system causing some nodes to bug silently.

I will soon push my latest updates of the code (I fixed the tests that were failing), but I think I still need a bit more work on this to be sure not to create bugs

SteveMacenski:
Ok, I appreciate your caution!

SteveMacenski:
Ok! thanks for letting us know!

PR #4393, opened 2024-06-03
Revamp nav2_util CMakeLists.txt to use modern idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | related to #4357  |

---

1.  Switches to using target_link_libraries everywhere. This gives us finer-grained control over what dependencies are exported to downstream as public, or private.  In the particular case of nav2_util, this actually doesn't matter *too* much, but it will help for other packages.
2.  Moves the include directory down one level to include/${PROJECT_NAME}, which is best practice in ROS 2 since Humble.
3.  Makes sure to export nav2_util as a CMake target, so downstream users of it can use that target.
4.  Moves the base_footprint_publisher.hpp header file into the src directory, as it isn't functionality that an external project could use.

One thing of note here is that to make these changes, I used https://github.com/clalancette/ros2_pkg_dep_tool to at least guide which dependencies were needed.  I didn't actually apply all of the changes suggested by the tool, as it would require quite a bit of changes to the `#include` lines, but at least it was useful in guiding what dependencies were actually needed here.

None should be *required*, particularly for this first PR.  Once the whole set has gone in, we should probably update the docs to suggest people link against Navigation2 using the CMake targets and `target_link_libraries`, rather than `ament_target_dependencies`.

---

Assuming the basic idea here is accepted, all of the rest of the packages in Navigation2 should undergo a similar treatment.  See https://github.com/ros-navigation/navigation2/pull/4357 for a prototype of what that would look like, though significant changes need to be made there.

#### For Maintainers: 

### DIFF
--- a/nav2_util/src/CMakeLists.txt
+++ b/nav2_util/src/CMakeLists.txt
@@ -10,34 +10,41 @@ add_library(${library_name} SHARED
   odometry_utils.cpp
   array_parser.cpp
 )
-
-ament_target_dependencies(${library_name}
-  rclcpp
-  nav2_msgs
-  tf2
-  tf2_ros
-  nav_msgs
-  geometry_msgs
-  lifecycle_msgs
-  rclcpp_lifecycle
-  tf2_geometry_msgs
-  bondcpp
+target_include_directories(${library_name}
+  PUBLIC
+  "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
+target_link_libraries(${library_name} PUBLIC
+  bondcpp::bondcpp
+  ${geometry_msgs_TARGETS}
+  ${lifecycle_msgs_TARGETS}
+  ${nav2_msgs_TARGETS}
+  ${nav_msgs_TARGETS}
+  ${rcl_interfaces_TARGETS}
+  rclcpp::rclcpp
+  rclcpp_action::rclcpp_action
+  rclcpp_lifecycle::rclcpp_lifecycle
+  tf2_ros::tf2_ros
+  tf2::tf2
+  ${tf2_geometry_msgs_TARGETS}
+)
+target_link_libraries(${library_name} PRIVATE
+  ${bond_TARGETS}
 )
 
 add_executable(lifecycle_bringup
   lifecycle_bringup_commandline.cpp
 )
-target_link_libraries(lifecycle_bringup ${library_name})
+target_link_libraries(lifecycle_bringup PRIVATE ${library_name} rclcpp::rclcpp)
 
 add_executable(base_footprint_publisher
   base_footprint_publisher.cpp
 )
-target_link_libraries(base_footprint_publisher ${library_name})
-
-find_package(Boost REQUIRED COMPONENTS program_options)
+target_link_libraries(base_footprint_publisher PRIVATE ${library_name} rclcpp::rclcpp ${tf2_msgs_TARGETS})
 
 install(TARGETS
   ${library_name}
+  EXPORT ${library_name}
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
--- a/nav2_util/CMakeLists.txt
+++ b/nav2_util/CMakeLists.txt
@@ -2,54 +2,36 @@ cmake_minimum_required(VERSION 3.5)
 project(nav2_util)
 
 find_package(ament_cmake REQUIRED)
+find_package(bond REQUIRED)
+find_package(bondcpp REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(lifecycle_msgs REQUIRED)
 find_package(nav2_common REQUIRED)
 find_package(nav2_msgs REQUIRED)
-find_package(tf2_ros REQUIRED)
-find_package(tf2 REQUIRED)
-find_package(tf2_geometry_msgs REQUIRED)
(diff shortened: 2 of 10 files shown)


clalancette:
CI is failing to build because we need a release of bondcpp into Rolling.  I'll take a look at getting that done.

nav2_util/src/CMakeLists.txt
```diff
@@ -10,34 +10,41 @@ add_library(${library_name} SHARED
...
-  geometry_msgs
-  lifecycle_msgs
-  rclcpp_lifecycle
-  tf2_geometry_msgs
-  bondcpp
+target_include_directories(${library_name}
+  PUBLIC
+  "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
+target_link_libraries(${library_name} PUBLIC
```
SteveMacenski, line 17: Question: when do you know when to do `geometry_msgs_TARGETS` vs setting a library like `tf2_ros::tf2_ros`?

nav2_util/src/CMakeLists.txt
```diff
@@ -10,34 +10,41 @@ add_library(${library_name} SHARED
...
+  ${rcl_interfaces_TARGETS}
+  rclcpp::rclcpp
+  rclcpp_action::rclcpp_action
+  rclcpp_lifecycle::rclcpp_lifecycle
+  tf2_ros::tf2_ros
+  tf2::tf2
+  ${tf2_geometry_msgs_TARGETS}
+)
+target_link_libraries(${library_name} PRIVATE
+  ${bond_TARGETS}
```
SteveMacenski, line 32: Question: why is this one & rclcpp below private? When do we do public vs private?

SteveMacenski (approved):
Questions, but no issues once bond changes turn over

nav2_util/src/CMakeLists.txt
clalancette: Right, good question.

In general, you should always prefer to use the target name, like `tf2_ros::tf2_ros`.  That is the most CMake-like thing to do.

In the current Rolling code, the exception is for messages/services/actions/interfaces packages.  There, there are multiple targets that need to be linked against for things to work, so we need to use `${geometry_msgs_TARGETS}`.  We hope to eventually fix that so that there is a "single" target for message packages, but we aren't there yet.

nav2_util/src/CMakeLists.txt
clalancette: Yeah, good question.

So the difference between PUBLIC and PRIVATE is that PUBLIC dependencies are exported to downstream consumers, while PRIVATE ones are not.  Marking dependencies as PRIVATE (where they can be) is part of what helps speed up compilation, since the compiler doesn't have to go searching through additional directories for headers and libraries.

As for when to use each, there are separate rules for executables and libraries:
1.  No downstream can ever compile or link against an executable, so all dependencies can be PRIVATE.  This honestly doesn't matter too much, as no downstream will even attempt to link, but making it PRIVATE is just cleaner.
2. For libraries, a dependency is PRIVATE if it is only used during compilation of *this* package.  So in a standard ROS package layout, if something is only ever referenced in the `src` directory, then it can be PRIVATE.  Anything that is referenced in the `include` directory needs to be PUBLIC, because in order for a downstream package to compile, it also needs to have access to those includes.

As a concrete example for 2, in `nav2_util` the `bond` package (which is really a message package, despite the name) is only ever referenced in https://github.com/ros-navigation/navigation2/blob/03ee50a629425bdd2f24463474732b7d9bf6fe95/nav2_util/src/lifecycle_node.cpp , so it can be PRIVATE.  On the other hand, `geometry_msgs` is referenced in https://github.com/ros-navigation/navigation2/blob/03ee50a629425bdd2f24463474732b7d9bf6fe95/nav2_util/include/nav2_util/geometry_utils.hpp (among other places), so it needs to be PUBLIC (a library that depends on `nav2_util` needs to know where to find the `geometry_msgs` headers so it can compile).

Does this make sense?

clalancette: > * So what if the compiled library is used in another package (i.e. lifecycle node is then used in other packages as the base of the action servers)? Seem unintuitive to me that the include in the cpp vs hpp would make that much of a difference, but 👍

I'll preface this by saying I don't completely know how this works, so the below is a bit of speculation.

Thinking about it, if you have a symbol that is only ever internal to a library, then no downstream can ever make (direct) reference to it.  Thus, the downstream packages don't have to know how to resolve that upstream symbol in order to compile/link.  They *do* have to know how to resolve that symbol at runtime, but I believe that is handled either via RPATH/RUNPATH, or via LD_LIBRARY_PATH.

> But if that was in the `lifecycle_node.hpp` file, then it would need to be public, correct?

Yes, exactly.

> So.. is it worth me at some point doing a review of the stack and moving as much to cpp files to make the dependencies private?

Moving as much as possible into the `.cpp` files will reduce the number of dependencies you need to export to downstream, so yes, I think it is worth doing.

SteveMacenski:
Let me know when bond’s all good for this to turn over. Consider it approved and we can move onto the next one(s) while we’re waiting!

clalancette:
We've had a rolling sync now, so in theory CI should pass here.  If you could kick it off again, we'll see!

SteveMacenski:
Builds now! But a weird test failed that seems unrelated but also not known as being flaky. I'm rerunning it now and I'll check in the morning. If it happens again, I'll take care of it so you can move on.

PR #4395, opened 2024-06-04
[RPP] add is_rotating_to_heading feedback

doisyg:
---

## Basic Info

| ------ | ----------- |

---

A draft first to discuss, I would be interested in the feedback of RPP users. cc @james-ward 

This PR adds the publisher topic `is_rotating_to_heading` in order to feedback to an external system if RPP is currently rotating to align with the path or final pose. 
A situation that happens either if `shouldRotateToGoalHeading(carrot_pose)` or `shouldRotateToPath(rotate_to_path_carrot_pose, angle_to_heading, x_vel_sign)` is true.

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
+++ b/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
@@ -27,6 +27,7 @@
 #include "pluginlib/class_loader.hpp"
 #include "pluginlib/class_list_macros.hpp"
 #include "geometry_msgs/msg/pose2_d.hpp"
+#include "std_msgs/msg/bool.hpp"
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

SteveMacenski:
No issue with the implementation, though curious on the need. Feel free to merge when you're satisfied

doisyg:
> No issue with the implementation, though curious on the need. Feel free to merge when you're satisfied

Scanning application where I want to start scanning only after RPP is aligned with the path I want to scan. Hence I need the feedback

I will wait a bit to see if @james-ward has any feedback, and then we can merge!

SteveMacenski:
Do you want it actively or passively (i.e. need event based topics or query based)? A parameter or service could also work. Also I was thinking that a diagnostic message with that instead could be useful. 

Any way, happy with this PR, just some thoughts

doisyg:
Active, and published in sync with the Twist is perfect for me. So as it is currently (I could use diagnostic too but a Bool topic seems simpler to produce and consume to me).
Leaving it open for 24h and then I will merge

Issue #4379, opened 2024-05-30
18 bugs (UAF) in `nav2_amcl` by setting dynamic parameters


GoesM:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu22.04
- ROS2 Version:
  -  humble
- Version or commit hash:
  -  the latest
- DDS implementation:
  -  defaulted

#### Steps to reproduce issue

Launch the navigation2 normally, as following steps:
```sh
#!/bin/bash
export ASAN_OPTIONS=halt_on_error=0:new_delete_type_mismatch=0:detect_leaks=0:log_pah=asan
source install/setup.bash
export TURTLEBOT3_MODEL=waffle
export GAZEBO_MODEL_PATH=$GAZEBO_MODEL_PATH:/opt/ros/humble/share/turtlebot3_gazebo/models
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=True use_rviz:=False use_composition:=False 
```

Learning about how the dynamic-parameter works , I had a try on it . 
```sh
ros2 param set /amcl z_rand 0.5
```

[notice] whatever the value be (0.5 or anyother double value43) , the UAF occurs all the time. 

#### Expected behavior
no crash occurs

#### Actual behavior
The ASAN reporting a **heap-user-after-free** bug to me as following, and the `nav2_amcl` stop its work.

```
=================================================================
==145868==ERROR: AddressSanitizer: heap-use-after-free on address 0x6170000aaab0 at pc 0x730fff32b5a4 bp 0x7fff30eba470 sp 0x7fff30eba468
READ of size 8 at 0x6170000aaab0 thread T0
    #0 0x730fff32b5a3 in message_filters::Signal1<sensor_msgs::msg::LaserScan_<std::allocator<void> > >::removeCallback(std::shared_ptr<message_filters::CallbackHelper1<sensor_msgs::msg::LaserScan_<std::allocator<void> > > > const&) (/home/***/nav2_humble/install/nav2_amcl/lib/libamcl_core.so+0x52b5a3) (BuildId: e96a3822dfb41386b5409b544e4bc74088297548)
[...]

PR #4397, opened 2024-06-05
[amcl] reset laser_scan_filter before reinit 

GoesM:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4379 |
| Primary OS tested on |  Ubuntu |

---

- reset `laser_scan_filter_` before reinit it
- avoid UAF bugs mentioned in #4379 

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/src/amcl_node.cpp
+++ b/nav2_amcl/src/amcl_node.cpp
@@ -1367,6 +1367,7 @@ AmclNode::dynamicParametersCallback(
     lasers_update_.clear();
     frame_to_laser_.clear();
     laser_scan_connection_.disconnect();
+    laser_scan_filter_.reset();
     laser_scan_sub_.reset();
 
     initMessageFilters();


SteveMacenski:
Does this actually fix the issue? Isn't the laser scanner still able to process scans while parameters are changing? I thought you were going to reset it at the start of the callback and always reset the message filter

GoesM:
I‘m so sorry for any confusion caused by my previous description. 

The following statement may be more clear: 


After such dynamic parameters change,  because of the lack of `scan_filter_.reset()` before `initMessageFilters()` , the `scan_filter_` is still running even after `scan_sub_.reset()`. 

BUT based on the definition of `scan_filter_`, it would access the freed pointer `scan_sub_`:

https://github.com/ros-navigation/navigation2/blob/7dc76da7cf166e9a1b9345c5d3e91544b1ca696b/nav2_amcl/src/amcl_node.cpp#L1515-L1519 

so that , the use-after-free bug occurs. 

***
So what I did was add this missing code.

SteveMacenski:
Got it! It may have been that you described it right and I just jumped the gun (I've been known to do that) thinking I knew where you were going with that. Thanks!

PR #4418, opened 2024-06-11
Remove start clear

SteveMacenski:
We found that clearing the start made other orientations in the same pose also be "free" even though those might be in critical collision at other orientations. We know the start is "free" because it has to be definitionally since the robot is there. 

This removes the clearing to fix that issue. Conveniently, we have defensive programming elsewhere that this makes it the only required change. Within the planner plugins themselves we already interpret 1 iteration as being start occupied and thus no valid potential path solution could be found, so we keep that contextual failure mode if there's no way out from the start. 

Addresses https://github.com/ros-navigation/navigation2/issues/4271

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
@@ -250,12 +250,6 @@ protected:
   inline void populateExpansionsLog(
     const NodePtr & node, std::vector<std::tuple<float, float, float>> * expansions_log);
 
-  /**
-   * @brief Clear Start
-   */
-  void clearStart();
-
-
   bool _traverse_unknown;
   int _max_iterations;
   int _max_on_approach_iterations;
--- a/nav2_smac_planner/src/a_star.cpp
+++ b/nav2_smac_planner/src/a_star.cpp
@@ -250,8 +250,6 @@ bool AStarAlgorithm<NodeT>::areInputsValid()
   }
 
   // Note: We do not check the if the start is valid because it is cleared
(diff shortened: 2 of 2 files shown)


tonynajjar:
Regression test?

tonynajjar:
@SteveMacenski would it make sense to bring back the[ starting pose collision check ](https://github.com/ros-navigation/navigation2/pull/3980/files#diff-ea65b17d27ef944be98d5e39fdb7275f0d2733805daefec99b0c606e3a9d6c84L238) now that this is removed?

SteveMacenski:
Intuitively, if the robot is at its start pose, doesn't that mean its not in collision by definition that it is currently occupying that space?If its critically in collision (not just against an obstacle but transported into another obstacle) that it is _really_ not able to plan then my comment in the description of this revert PR still makes sense to me. 

I would think this is good behavior you want so that your robot can recover itself from bad situations where it ends up potentially up against or very close to other obstacles for whatever reason.

If not, then I could support adding it back in as a parameterized feature (default off, since I think most people would want that default off to have the easiest recovery path possible).

tonynajjar:
> I would think this is good behavior you want so that your robot can recover itself from bad situations where it ends up potentially up against or very close to other obstacles for whatever reason

I can see this point. My concern is that in such scenarios, by pure visual interpolation, I don't trust that that it can go from pose 1 (fully horizontal inside the lethal zone) to pose 2 without colliding with the lethal space at the border of the pose 1. 
Does SMAC collision-checks only the the blue footprints or also some invisible poses in between?

image

PR #4424, opened 2024-06-12
[nav2_costmap_2s]: free correctly the memory in the InflationLayer

facontidavide:
## Basic Info

| ------ | ----------- |

---

When using very large maps ( we are talking about maps in the order of 200 Mb), we noticed that the `InflationLayer` would consume a VERY large amount of memory.

After investigation, we realized that:

- memory is never freed
- the total amount of memory consumed can be reduced (no spikes)

See the difference before and after:

### After

The impact in terms of performance (CPU usage) should be negligible

None

---

None

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/inflation_layer.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/inflation_layer.hpp
@@ -60,18 +60,16 @@ class CellData
 public:
   /**
    * @brief  Constructor for a CellData objects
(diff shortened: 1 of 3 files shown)


doisyg (approved):
Tested at Dexory, great memory gains proportional to the number of inflated cells. In other words bigger gains with big maps and small resolution.
Will wait to see if @SteveMacenski has a say and then we can merge.
@BriceRenaudeau fyi

SteveMacenski (approved):
Neat, no objections here!

PR #4426, opened 2024-06-12
[AMCL] reduce the amount of memory used

facontidavide:
---

## Basic Info

| ------ | ----------- |

---

This is a memory optimization that should not affect realtime perfomance, but will reduce considerably the amount of memory used by AMCL.

In our large map (200 Mb) this is the memory usage **before** the change:

And **after** the changes:

The main changes are:

1. Changes in `map_cell_t`: reduction from 12 bytes to 5 bytes
2. Add a new parameter to reduce the memory used by `free_space_indices` using 1/4th of the points (equally spaced)

Describe the new parameter "freespace_downsampling"

---

To be discussed

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/src/amcl_node.cpp
+++ b/nav2_amcl/src/amcl_node.cpp
@@ -227,6 +227,11 @@ AmclNode::AmclNode(const rclcpp::NodeOptions & options)
   add_parameter(
     "first_map_only", rclcpp::ParameterValue(false),
     "Set this to true, when you want to load a new map published from the map_server");
+
+  add_parameter(
+    "freespace_downsampling", rclcpp::ParameterValue(
+      false),
+    "If true, downsample the free space used by the Pose Generator. This will save memory in VERY large maps");
 }
 
 AmclNode::~AmclNode()
@@ -404,7 +409,7 @@ AmclNode::checkElapsedTime(std::chrono::seconds check_interval, rclcpp::Time las
 }
 
 #if NEW_UNIFORM_SAMPLING
-std::vector<std::pair<int, int>> AmclNode::free_space_indices;
+std::vector<AmclNode::Point2D> AmclNode::free_space_indices;
 #endif
 
 bool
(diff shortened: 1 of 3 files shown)


doisyg:
Noting that it correspond to a new (smaller) maximum side size of 3.2 km for AMCL maps at a standard resolution of 0.05m/px. I think that it is fair and that we don't have any use cases of nav2 with maps that big. 
I expect other part of the system to fail before a map side .
wdyt @SteveMacenski and @tonynajjar ?

facontidavide:
UPDATE: sorry I don't understand why this PR include files outside **nav2_amcl**. I will clean it up

nav2_amcl/src/amcl_node.cpp
```diff
@@ -1106,6 +1111,7 @@ AmclNode::initParameters()
   get_parameter("always_reset_initial_pose", always_reset_initial_pose_);
   get_parameter("scan_topic", scan_topic_);
   get_parameter("map_topic", map_topic_);
+  get_parameter("freespace_downsampling", freespace_downsampling_);
```
SteveMacenski, line 1114: Perhaps set this as a configuratable downsampling step? i.e. 2,3,4,5,... might be nice to experiment with yourself.

SteveMacenski (approved):
Otherwise LGTM. AMCL changes usually scare me since its been untouched for so long, but this is minor and I don't foresee any issues. 

`double->float` there's a reason MPPI/Smac both use floats for everything, I found massive improvements in performance on it alone, but I haven't had the cycles to try to apply that to the full stack to see how much more it could be reduced (though, those are 2 of the heaviest operations outside of the costmap).

Party on! What's next for memory consumption? :sunglasses:

SteveMacenski:
`freespace_downsampling_` needs to be a class member, that should get CI unblocked

nav2_amcl/src/amcl_node.cpp
facontidavide: We have a comically large map and 1/th is already enough to make us happy

SteveMacenski:
Still a draft, should I merge this in?

facontidavide:
if it is good for you, it is good for me. I will prepare tomorrow a PR for the documentation https://docs.nav2.org/configuration/packages/configuring-amcl.html?highlight=amcl

facontidavide:
> Party on! What's next for memory consumption? 😎

:smile: sparse costmaps ??? Not yet... not yet...

SteveMacenski:
Please do update the docs for this param. You can put it into the Iron->Jazzy file since Nav2 hasn't forked for Jazzy yet (still working on it..)

> 😄 sparse costmaps ??? Not yet... not yet...

Well before you go nutty there, lets discuss that. I have a broader plan for costmaps in the medium term future :wink:

facontidavide:
> Well before you go nutty there, lets discuss that. I have a broader plan for costmaps in the medium term future
Keep me in the loop!

PR #4427, opened 2024-06-13
Small changes

xianglunkai:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_docking/opennav_docking/src/simple_charging_dock.cpp
+++ b/nav2_docking/opennav_docking/src/simple_charging_dock.cpp
@@ -284,19 +284,23 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
 {
   double velocity = 0.0;
   double effort = 0.0;
+  size_t match_count = 0;  
   for (size_t i = 0; i < state->name.size(); ++i) {
     for (auto & name : stall_joint_names_) {
       if (state->name[i] == name) {
         // Tracking this joint
         velocity += abs(state->velocity[i]);
         effort += abs(state->effort[i]);
+        ++match_count;
       }
     }
   }
 
   // Take average
-  effort /= stall_joint_names_.size();
-  velocity /= stall_joint_names_.size();
+  if (match_count > 0) {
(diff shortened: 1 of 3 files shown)


nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -284,19 +284,23 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
         effort += abs(state->effort[i]);
+        ++match_count;
       }
     }
   }
 
   // Take average
-  effort /= stall_joint_names_.size();
-  velocity /= stall_joint_names_.size();
+  if (match_count > 0) {
```
SteveMacenski, line 300: Is this just to avoid a divide by zero issue? If so, we could just have a `bool` instead which would be more intuitive unless we're going to use the `count` rather than the `size` (or check that `count == size` or something to check we have all the stuff needed)

SteveMacenski:
Please sign off with DCO! If you click on the job, it'll tell you how to ammend your commit for it (`-s`)

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
mikeferguson: If this is really targeting a divide by zero, we probably should update the configure() to not create the subscriber if stall_joint_names_ is empty() - it already prints an error, but there is no reason to have the subscriber run the callback if there are no names to match

SteveMacenski:
@xianglunkai can you update the PR as we asked in the review so we can merge this? :smile:

lharikrishnan1993:
The recommended change seems to have been already done in https://github.com/ros-navigation/navigation2/pull/4406 at [lines](https://github.com/ros-navigation/navigation2/pull/4406/files#diff-acfc4215f306528e573ee3dd0e311b35492816104ccba2ab3a1ef3d976a7c1fdR127-R129).

Can we close this PR?
@SteveMacenski

SteveMacenski:
Thanks for the note - done :-)

PR #4429, opened 2024-06-13
Added dynamic reconfigure and export controller in docking

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Clara and Morphia robots |
| Does this PR contain AI generated software? | Nop |

---

1. I added the callback for dynamic reconfigure to tune the docking controller in runtime.
2. The docking controller is a library so can be used for other packages.

I don't think so.

---

#### For Maintainers: 

### DIFF
--- a/nav2_docking/opennav_docking/src/controller.cpp
+++ b/nav2_docking/opennav_docking/src/controller.cpp
@@ -60,6 +60,7 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
 bool Controller::computeVelocityCommand(
   const geometry_msgs::msg::Pose & pose, geometry_msgs::msg::Twist & cmd, bool backward)
 {
+  std::lock_guard<std::mutex> lock(dynamic_params_lock_);
   cmd = control_law_->calculateRegularVelocity(pose, backward);
   return true;
 }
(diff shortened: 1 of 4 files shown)


ajtudela:
How are we going to manage this type of changes with respect to the repository in opennav, do I apply the same changes there or leave the other unchanged?

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -61,4 +64,44 @@ bool Controller::computeVelocityCommand(
   return true;
 }
 
+rcl_interfaces::msg::SetParametersResult
+Controller::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
+{
+  std::lock_guard<std::mutex> lock(dynamic_params_lock_);
```
SteveMacenski, line 70: Where's the other lock so that we're locking something?

SteveMacenski (changes requested):

nav2_docking/opennav_docking/src/controller.cpp
ajtudela: ups

SteveMacenski (approved):

Issue #4415, opened 2024-06-11
Nav2 Docking: The docking_server does not work if you do not define instances of docks using the DockDatabase


jcarlosgm30:
Hi @SteveMacenski ,

As far as I understood from nav2_docking README.md, it is not necessary to know the dock's localization in advance if you set the pose (and use_dock_id to false) in the docking request:

> The docking action can either operate on a dock in the DockDatabase or from a dock specified in the docking request. This second option is useful for testing or when dock's locales are not necessarily known in advance. If use_dock_id = true, it uses the dock_id field to specify which dock in the database to use. Else, you must populate the dock_pose and dock_type fields.
> 
> Extracted from: https://github.com/ros-navigation/navigation2/tree/main/nav2_docking#docking-action.

However, if you have not defined the dock's instances using the DockDatabase, the docking server is going to fail when it initializes:

https://github.com/ros-navigation/navigation2/blob/main/nav2_docking/opennav_docking/src/docking_server.cpp#L91

I am not sure if this is a bug or I didn't understand well the documentation.

BR,
José Carlos.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04 LTS
- ROS2 Version:
  - ROS2 Humble
- Version or commit hash:
  - 7737a2fafa27c18efef4689582799bdce6aeb1df
- DDS implementation:
  - Eclipse Cyclone DDS

#### Steps to reproduce issue
1. Do not set the dock instances in the .yaml parameter file to be taken by DockDatabase
2. Run the docking_server.
3. An error appears: 
`[opennav_docking-2] [ERROR] [1718100300.110008781] [docking_server]: Dock database filepath nor dock parameters set. Unable to perform docking actions.`

#### Expected behavior
Allows to use the docking_server using directly the action request, without the need to define the dock instances using the DockDatabase

#### Actual behavior
It is necessary to define at least one instance of the dock in the .yaml parameter file to be taken by the DockDatabase.

#### Additional information

SteveMacenski:
[...]

PR #4442, opened 2024-06-17
[Docking] Feature: support for not needing to set dock instances

jcarlosgm30:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4415 |
| Robotic platform tested on | Gazebo Simulation |

---

The changes made are as follows:
* [Support for not needing to set `dock_instances`](https://github.com/ros-navigation/navigation2/commit/15b085167ae4ee7682c636c7987a66a9b94488e1)
* [Improve some log messages in `dock_database.cpp`](https://github.com/ros-navigation/navigation2/commit/24337173b228c22a963be3ae5b98760adf1e6f18)
* [Comment the dock instances out from `nav2_params`](https://github.com/ros-navigation/navigation2/commit/ec9c012e8b4dd1e62d8a0aeaac3b106f3280a984)

These changes enable the system to work in the following use cases:
* Setting the dock instances in the parameters or database yaml file.
* Not setting the dock instances in these files but using the dock pose in the action request or the reload database service.

Also, the reload database service is available in any case where at least one dock plugin has been defined.

* The documentation may need to be updated to reflect the "Not setting ..." use case.

---

* The unit tests may need to be updated to address new changes

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/params/nav2_params.yaml
+++ b/nav2_bringup/params/nav2_params.yaml
@@ -407,11 +407,12 @@ docking_server:
       filter_coef: 0.1
 
     # Dock instances
-    docks: ['home_dock']  # Input your docks here
-    home_dock:
-      type: 'simple_charging_dock'
-      frame: map
-      pose: [0.0, 0.0, 0.0]
+    # The following example illustrates how you should configure the docks instances.
+    # docks: ['home_dock']  # Input your docks here
+    # home_dock:
+    #   type: 'simple_charging_dock'
+    #   frame: map
+    #   pose: [0.0, 0.0, 0.0]
 
     controller:
       k_phi: 3.0
--- a/nav2_bringup/params/nav2_params.yaml
+++ b/nav2_bringup/params/nav2_params.yaml
@@ -407,7 +407,7 @@ docking_server:
       filter_coef: 0.1
 
     # Dock instances
-    # The following example illustrates how you should configure the docks instances.
+    # The following example illustrates configuring dock instances.
     # docks: ['home_dock']  # Input your docks here
     # home_dock:
     #   type: 'simple_charging_dock'
--- a/nav2_docking/opennav_docking/src/dock_database.cpp
+++ b/nav2_docking/opennav_docking/src/dock_database.cpp
@@ -82,12 +82,20 @@ void DockDatabase::reloadDbCb(
   const std::shared_ptr<nav2_msgs::srv::ReloadDockDatabase::Request> request,
   std::shared_ptr<nav2_msgs::srv::ReloadDockDatabase::Response> response)
 {
+  auto node = node_.lock();
   DockMap dock_instances;
   if (utils::parseDockFile(request->filepath, node_.lock(), dock_instances)) {
     dock_instances_ = dock_instances;
     response->success = true;
+    RCLCPP_INFO(
+      node->get_logger(),
+      "Dock database reloaded from file %s.", request->filepath.c_str());
     return;
   }
+
+  RCLCPP_ERROR(
+    node->get_logger(),
+    "Dock database reload failed from file %s.", request->filepath.c_str());
   response->success = false;
 }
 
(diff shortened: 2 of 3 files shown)


nav2_bringup/params/nav2_params.yaml
```diff
@@ -407,11 +407,12 @@ docking_server:
       filter_coef: 0.1
 
     # Dock instances
-    docks: ['home_dock']  # Input your docks here
-    home_dock:
-      type: 'simple_charging_dock'
-      frame: map
-      pose: [0.0, 0.0, 0.0]
+    # The following example illustrates how you should configure the docks instances.
```
SteveMacenski, line 410:
```suggestion
    # The following example illustrates configuring dock instances.
```

nav2_docking/opennav_docking/src/dock_database.cpp
```diff
@@ -71,12 +82,20 @@ void DockDatabase::reloadDbCb(
...
   if (utils::parseDockFile(request->filepath, node_.lock(), dock_instances)) {
     dock_instances_ = dock_instances;
     response->success = true;
+    RCLCPP_INFO(
+      node->get_logger(),
+      "Dock database reloaded from file %s.", request->filepath.c_str());
     return;
   }
+
+  RCLCPP_ERROR(
```
SteveMacenski, line 96: Remove. Failure messages are already logged by `parseDockFile`

nav2_docking/opennav_docking/src/dock_database.cpp
```diff
@@ -71,12 +82,20 @@ void DockDatabase::reloadDbCb(
   const std::shared_ptr<nav2_msgs::srv::ReloadDockDatabase::Request> request,
   std::shared_ptr<nav2_msgs::srv::ReloadDockDatabase::Response> response)
 {
+  auto node = node_.lock();
   DockMap dock_instances;
   if (utils::parseDockFile(request->filepath, node_.lock(), dock_instances)) {
```
SteveMacenski, line 87: Update to use this new `node`

nav2_docking/opennav_docking/src/dock_database.cpp
```diff
@@ -99,6 +118,17 @@ Dock * DockDatabase::findDock(const std::string & dock_id)
 
 Dock * DockDatabase::findDockInstance(const std::string & dock_id)
 {
+  if (dock_instances_.empty()) {
```
SteveMacenski, line 121: This is not necessary, please remove. If the database is empty, `find` results in the map's `end` which returns nullptr already.

SteveMacenski (changes requested):
Generally looks good! Just some small cleanup!

nav2_docking/opennav_docking/src/dock_database.cpp
jcarlosgm30: I only add it to be more verbose for the user

nav2_docking/opennav_docking/src/dock_database.cpp
jcarlosgm30: done

nav2_bringup/params/nav2_params.yaml
jcarlosgm30: done

nav2_docking/opennav_docking/src/dock_database.cpp
jcarlosgm30: done

nav2_docking/opennav_docking/src/dock_database.cpp
jcarlosgm30: done

SteveMacenski (approved):

jcarlosgm30:
Yes, that test fails because there is no dock plugin initialized so the `/reload_database` service will not be created.

The old logic was that the `/reload_database` service was only created when `getDockPlugins` and `getDockInstances` returned false. As we discussed, the new logic creates the `/reload_database` service whenever at least a dock plugin is initialized. 

So I think it will be enough to initialize a dock plugin in the test.

SteveMacenski (approved):

SteveMacenski:
Fabulous, thank you!

PR #4449, opened 2024-06-18
Add cout bt logs

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_bt_navigator/src/bt_navigator.cpp
+++ b/nav2_bt_navigator/src/bt_navigator.cpp
@@ -125,7 +125,7 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
       navigators_.push_back(class_loader_.createUniqueInstance(navigator_type));
       if (!navigators_.back()->on_configure(
           node, plugin_lib_names, feedback_utils,
-          &plugin_muxer_, odom_smoother_))
+          &plugin_muxer_, odom_smoother_, logger_cout_))
       {
         return nav2_util::CallbackReturn::FAILURE;
       }
--- a/nav2_core/include/nav2_core/behavior_tree_navigator.hpp
+++ b/nav2_core/include/nav2_core/behavior_tree_navigator.hpp
@@ -128,7 +128,8 @@ public:
     const std::vector<std::string> & plugin_lib_names,
     const FeedbackUtils & feedback_utils,
     nav2_core::NavigatorMuxer * plugin_muxer,
-    std::shared_ptr<nav2_util::OdomSmoother> odom_smoother) = 0;
+    std::shared_ptr<nav2_util::OdomSmoother> odom_smoother,
+    std::shared_ptr<BT::StdCoutLogger> logger_cout_) = 0;
 
   /**
    * @brief Activation of the navigator's backend BT and actions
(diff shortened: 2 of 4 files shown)


tonynajjar:
I don't have the time currently to do what it takes to get this merged in but I wanted to open a draft PR with the idea otherwise my branch would just slowly disappear

nav2_core/include/nav2_core/behavior_tree_navigator.hpp
```diff
@@ -128,7 +128,8 @@ class NavigatorBase
     const std::vector<std::string> & plugin_lib_names,
     const FeedbackUtils & feedback_utils,
     nav2_core::NavigatorMuxer * plugin_muxer,
-    std::shared_ptr<nav2_util::OdomSmoother> odom_smoother) = 0;
+    std::shared_ptr<nav2_util::OdomSmoother> odom_smoother,
+    std::shared_ptr<BT::StdCoutLogger> logger_cout_) = 0;
```
SteveMacenski, line 132: Why not the abstract logger so folks could pass in any logger they want (or have option in the higher level node to select the logger type they prefer)?

nav2_bt_navigator/src/bt_navigator.cpp
```diff
@@ -125,7 +125,7 @@ BtNavigator::on_configure(const rclcpp_lifecycle::State & /*state*/)
       navigators_.push_back(class_loader_.createUniqueInstance(navigator_type));
       if (!navigators_.back()->on_configure(
           node, plugin_lib_names, feedback_utils,
-          &plugin_muxer_, odom_smoother_))
+          &plugin_muxer_, odom_smoother_, logger_cout_))
```
SteveMacenski, line 128: Where is `logger_cout_` made? Shouldn't this be created in the BT Navigator to be passed onto the navigator base to share? I'm not sure why we're passing in a nullptr and its unused in the BT Navigator level unless (1) we're creating it here for all N instances to share or (2) we're keeping the logger object here to use somewhere.

(1) would mean we should be instantiating it above and then passing it in to share
(2) would mean we can remove modifying the `on_configure` API / storing the logger object at this high level, because its unused and breaks ABI (since its unused here anyway)

SteveMacenski (changes requested):

nav2_bt_navigator/src/bt_navigator.cpp
tonynajjar: yeah I know its weird to pass a nullptr but I got to this point to statisfy:

- There can only be one instance of [StdCoutLogger](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/a574beaa39c25649ca42f5d529a6033e3d9f4327/src/loggers/bt_cout_logger.cpp#L12) so it has to be shared across all navigator plugins
- the logger can only be created with a `tree_` which is only available `BehaviorTreeNavigator` and not in `BTNavigator`

It's possible that I missed another obvious way of implementing though

SteveMacenski: @facontidavide this is the second time that the singleton nature of the loggers has caused us problems in Nav2. We actually removed live Groot monitoring because of it. 

Would it be possible to allow multiple loggers in BT.CPP to allow for this?

facontidavide: First, it is ok to remove the singleton constraint. Let me work on it.

facontidavide: Done: https://github.com/BehaviorTree/BehaviorTree.CPP/commit/ced041bc3c00f450a83ee8bfc62471511bcc575d

facontidavide: Note that creating your own logger is trivial and boring. StdCout is only 30 lines of code. Are you sure that you don't want to create your own, maybe one that use `RCLCPP_INFO` instead of `std::cout` ?

SteveMacenski: I leave that to @tonynajjar , but I agree that doing an RCLCPP log would be better so it enters the `.ros/logs` entries. That would probably need to live here I guess since ROS shouldn't bleed into BT.CPP.

facontidavide: > That would probably need to live here I guess since ROS shouldn't bleed into BT.CPP.

exactly!

tonynajjar:
What I need is already [there](https://github.com/ros-navigation/navigation2/blob/23a48154bef1be3b063c7566698a2fdf4f497f66/nav2_behavior_tree/include/nav2_behavior_tree/ros_topic_logger.hpp#L77) :man_facepalming: Closing this PR. Let me know if you see more potential to it

PR #4456, opened 2024-06-20
Call `gz sim` directly without `shell=True` for a clean exit

azeey:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | System tests on Gazebo simulation  |

---

* https://github.com/ros-navigation/navigation2/pull/4440#issuecomment-2178985584 mentions that Gazebo needs to be SIGKILL'd. This is due to https://github.com/ros2/launch/issues/757 where the SIGINT signal doesn't propagate to Gazebo when `shell=True`. A simple fix is to execute gz-sim directly instead of using the launch files from ros_gz. 

TODO:
- ~If we're okay with this chance, I can update the rest of the launch files.~

#### For Maintainers: 


SteveMacenski:
Testing this right now with the full suite of tests and working well. I think we'll keep the python3 util in place so that optionality exists (its still generally useful, even outside of GZ) 

Closing, new PR with the full suite ready. Thanks for the help @azeey!

Issue #4405, opened 2024-06-08
Nav2 Docking: Option for Non-Charging Dock Plugin Types


SteveMacenski:
If you want to "dock" with a non-charging type, we can easily add in a new plugin type for a non-charging docking maneuver. For example: docking with a convey belt or another robot.

It should be straight forward: have a non-charging dock plugin type and adjust the logic that if we're a non-charging dock to bypass certain parts of the process. We just rely more heavily on user-defined plugin for the specific situation to understand when "done" is _done_ (though the tolerance is a pretty good and reliable metric as we found in docking even without the joint state motor efforts)

redvinaa:
Summing up #44:
How to implement a scenario where a robot has to dock both forward and backward (for example, forward to a table / conveyor belt, and backward to actually charge)?
An option is to run two docking_server instances.

SteveMacenski:
^ Or have the option for reversing to be a feature of the dock not of the system like we have today. Food for thought.

SteveMacenski:
In progress - should have a PR opened this week

SteveMacenski:
https://github.com/ros-navigation/navigation2/pull/4627 implements

PR #4458, opened 2024-06-21
Dock panel

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Morphia robot / Gazebo |

---
New panel for docking / undocking action, similar to Nav2 panel:
* Goal status and feedback.
* Error codes.
* Docking action its 1000% configurable: dock id, pose and type; and flag for nav to stage.

docking panel

[panel_docking.webm](https://github.com/ros-navigation/navigation2/assets/4681902/d8f61647-fdf4-4567-bd1d-6965cf670545)

---

#### For Maintainers: 


SteveMacenski:
Very cool, I love this! Can we get this in the Nav2 default rviz config? Maybe on the right side of the window since the left side is getting pretty crowded. Or I suppose I wouldn't be opposed to rearranging the panel setup in rviz to be less "default" for a Nav2 specific layout. Perhaps:
- Move the displays panel (where you toggle/pick stuff to visualize) to the right top, with the views on the right bottom
- Move the Nav2 main panel to the top left, Nav2 docking center left, Nav2 controller selector bottom left

But just a suggestion! If we're going to have more of these plugins, its worth optimizing for them now!

nav2_rviz_plugins/src/docking_panel.cpp
```diff
@@ -0,0 +1,485 @@
...
+    goal_msg.dock_pose.header.frame_id = "map";
+    goal_msg.dock_pose.header.stamp = client_node_->now();
+    goal_msg.dock_pose.pose.position.x = dock_pose_x_->text().toDouble();
+    goal_msg.dock_pose.pose.position.y = dock_pose_y_->text().toDouble();
+    goal_msg.dock_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
+      dock_pose_yaw_->text().toDouble());
+    goal_msg.dock_type = combo_box->currentText().toStdString();
+
+    RCLCPP_INFO(
+      client_node_->get_logger(), "DockRobot will be called using dock pose: (%f, %f)",
```
SteveMacenski, line 244: Add the type here in the log

nav2_rviz_plugins/src/docking_panel.cpp
```diff
@@ -0,0 +1,485 @@
...
+      RCLCPP_ERROR(client_node_->get_logger(), "Dock pose is empty.");
+      return;
+    }
+    goal_msg.dock_pose.header.frame_id = "map";
+    goal_msg.dock_pose.header.stamp = client_node_->now();
+    goal_msg.dock_pose.pose.position.x = dock_pose_x_->text().toDouble();
+    goal_msg.dock_pose.pose.position.y = dock_pose_y_->text().toDouble();
+    goal_msg.dock_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
+      dock_pose_yaw_->text().toDouble());
+    goal_msg.dock_type = combo_box->currentText().toStdString();
```
SteveMacenski, line 241: How is the combo box populated with dock plugins?

ajtudela: The combobox is filled using the same ``pluginLoader`` function as the Selector: https://github.com/ros-navigation/navigation2/pull/4458/files#diff-e29dbc8bae4e148962b55e5b26da6a96d5dfca9fdb8223724abca2c463f3dd11R37

ajtudela:
I personally use a config similar to this: 
rviz

ajtudela:
So fixing the alignment for the docking panel, also fixed the alignment for the Nav2 panel!

rviz2

nav2_rviz_plugins/src/docking_panel.cpp
ajtudela: Done

SteveMacenski:
Great! Let me know when you want me to look at it again!

Can you try a layout where all the Nav2 stuff is on one side? It would be nice if it were grouped together!

ajtudela:
> Great! Let me know when you want me to look at it again! This is really quick turn around on this tool and its unquestionably valuable to have, thanks!
> 
> Can you try a layout where all the Nav2 stuff is on one side? It would be nice if it were grouped together!

You can review it now. I've fixed the log and the layout

SteveMacenski:
Can you also push the updated rviz config?

ajtudela:
> 
> Can you try a layout where all the Nav2 stuff is on one side? It would be nice if it were grouped together!

The only option is to merge the panels in "tabs". You can see them at the bottom.
Captura desde 2024-06-21 20-25-52

SteveMacenski:
> The only option is to merge the panels in "tabs". You can see them at the bottom.

Can't you move the top Displays panel to the right hand side?

ajtudela:
> > The only option is to merge the panels in "tabs". You can see them at the bottom.
> 
> Can't you move the top Displays panel to the right hand side where the Views are?

After a bit of struggle, if have this:
Captura desde 2024-06-21 20-39-34

SteveMacenski:
Eh ok, you've convinced me about https://github.com/ros-navigation/navigation2/pull/4458#issuecomment-2183217347 can you push that one? I'll test and merge

ajtudela:
Done!

SteveMacenski:
Weird... basically all the tests failed. Trying again!

ajtudela:
> Weird... basically all the tests failed. Trying again!

Yeah, I thought CI was failing again. I made a rebase and now the tests passed.

SteveMacenski (approved):

SteveMacenski:
Thanks! This looks great! Amazing tool to have now for users!

SteveMacenski:
@ajtudela I'm running into a problem while using the panel. It doesn't appear that the terminal state (i.e. action completed) is updating the state machine properly. When docking, it works fine but then never exits the "active action" state (i.e. keeps the dock/undock button greyed out and the cancel button active) even though the action was long completed. That prevents us from being able to undock or redock

ajtudela:
It's weird, I saw the code of the new non-charging dock type and it doesn't change the code of the docking server. The status of the buttons and text only depends on the goal status actions. Maybe these callbacks are not called when the action is completed?

https://github.com/ros-navigation/navigation2/blob/c2e4aabdab5a4d3ebbf011c9e28d00ce3e27a4c3/nav2_rviz_plugins/src/docking_panel.cpp#L152-L181

I'll check it out as soon as I get back to the lab.

ajtudela:
> @ajtudela I'm running into a problem while using the panel developing #4405 on `main`. It doesn't appear that the terminal state (i.e. action completed) is updating the state machine properly. When docking, it works fine but then never exits the "active action" state (i.e. keeps the dock/undock button greyed out and the cancel button active) even though the action was long completed. That prevents us from being able to undock or redock

I've tried several times to perform dock/undock actions using the new non-charging dock successfully using the dock panel . Can you give me more details about your system so I can try to reproduce the bug?

SteveMacenski:
That is weird, because I saw this consistently. I used the `main` branch and the provided `nav2_bringup` launch files so the server bringup / branch should be the same between us (nothing custom). Then using the rviz tool which interacts with the docking server fine (i.e. it docks) but then the rviz panel state seems to miss the results or something

PR #4480, opened 2024-06-24
Bumping to 1.3.0 for jazzy release

SteveMacenski:


chrisib:
I just pulled this, but I'm seeing this error when I try to build on Jazzy:

```
CMake Error at src/CMakeLists.txt:17 (target_link_libraries):
  Target "nav2_util_core" links to:

    bondcpp::bondcpp

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

I've installed all the dependencies with `rosdep`, including `ros-jazzy-bondcpp`.

Before opening a github issue I figured I'd just ask if this was a known problem at the moment.

SteveMacenski:
You will need to clean your workspace and rebuild to remove some old caches.

chrisib:
I just ran
```
cd ~/colcon_ws
rm -rf build/ log/ install/
colcon build
```
but I'm still seeing the same error.

SteveMacenski:
@clalancette are the updates you made for build times not backported to Jazzy or take some time for a future sync?

chrisib:
If it's helpful I can create an issue for this, rather than it being a discussion inside the merge request itself.

SteveMacenski:
If this is indeed a Jazzy problem (which I suspect it is now), it'll come up in the build farm as I'm working to release Jazzy binaries so I'll cross that bridge and handle it shortly as I get to a point that nav2_utils tries to build (right now dealing with some another annoying interdependency issues)

SteveMacenski:
I see that now, I’ll handle it tomorrow

clalancette:
> @clalancette can the updates you made for bondcpp 4.0.1 be added to Jazzy? They only appear to be released on rolling

Yes, that's true.  I can do a release for it now.

chrisib:
> I see that now, I’ll handle it tomorrow

I'm kind of relieved you're seeing it too and that it's not just some weird bug on my end.

Just in case it makes any difference, I'm building it for ARM on a Raspberry Pi 4 with Ubuntu Server 24.04 + Jazzy.  Happy to provide any other information that might be helpful getting this resolved and/or pulling any test branches.

SteveMacenski:
@chrisib bondcpp was updated. If you build that from source for now, you should be fine, but once Jazzy syncs up, it'll work without that (and you can use Nav2 binaries with it!)

DeccanLin:
> I just ran
> 
> ```
> cd ~/colcon_ws
> rm -rf build/ log/ install/
> colcon build
> ```
> 
> but I'm still seeing the same error:
> 
> ```
> --- stderr: nav2_util
> CMake Error at src/CMakeLists.txt:17 (target_link_libraries):
>   Target "nav2_util_core" links to:
> 
>     bondcpp::bondcpp
> 
>   but the target was not found.  Possible reasons include:
> 
>     * There is a typo in the target name.
>     * A find_package call is missing for an IMPORTED target.
>     * An ALIAS target is missing.
> 
> CMake Generate step failed.  Build files cannot be regenerated correctly.
> ---
> Failed   <<< nav2_util [20.0s, exited with code 1]
> ```

The issue still persists in version 1.3.1

chrisib:
@DeccanLin I had to clone `bondcpp` into my workspace (https://github.com/ros/bond_core) to compile nav2 for Jazzy. Once `bond` was in the `src` folder everything seems to have built correctly.

I suspect once the next sync of `.deb` packages from OSRF is released this should (hopefully) no longer be required.

PR #4485, opened 2024-06-24
Revamp nav2_behavior_tree CMakeLists.txt to use modern idioms.

clalancette:
## Basic Info

| ------ | ----------- |

---

This commit does a number of things:

1.  Switches to using target_link_libraries everywhere. This gives us finer-grained control over what dependencies are exported to downstream as public, or private.  In the particular case of nav2_behavior_tree, this actually doesn't matter *too* much, but it will help for other packages.
2.  Moves the include directory down one level to include/${PROJECT_NAME}, which is best practice in ROS 2 since Humble.
3.  Makes sure to export nav2_behavior_tree as a CMake target, so downstream users of it can use that target.
4.  Removes the use of boost.  To do this, we had to introduce our own version of the "split_string" method.
5.  Moves the test_action_server.hpp file into the main include directory.  This is slightly odd, but because downstream packages (opennav_docking_bt) depend on this header file to compile their own tests, it is technically part of the public interface.

None needed.

---

The rest of the packages in this repository need a similar treatment (PRs will be upcoming).

#### For Maintainers: 


clalancette:
This one needs a release of behaviortree.cpp to be done so we can get the exported target there.  I'll contact the maintainers tomorrow and see if we can get a release of that.

SteveMacenski:
Are these changes backportable to Jazzy or only rolling?

clalancette:
> Are these changes backportable to Jazzy or only rolling?

They are backportable to Jazzy, assuming that https://github.com/BehaviorTree/BehaviorTree.CPP/pull/826 is released into Jazzy (it has not yet been, as far as I can tell).

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -229,39 +240,59 @@ set(GENERATED_DIR ${CMAKE_CURRENT_BINARY_DIR}/gen)
 configure_file(plugins_list.hpp.in ${GENERATED_DIR}/plugins_list.hpp)
 
 add_executable(generate_nav2_tree_nodes_xml src/generate_nav2_tree_nodes_xml.cpp)
-ament_target_dependencies(generate_nav2_tree_nodes_xml ${dependencies})
+target_link_libraries(generate_nav2_tree_nodes_xml PRIVATE
+  behaviortree_cpp::behaviortree_cpp
```
SteveMacenski, line 244: This also requires Nav2 utils as of https://github.com/ros-navigation/navigation2/pull/4492

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -229,39 +240,59 @@ set(GENERATED_DIR ${CMAKE_CURRENT_BINARY_DIR}/gen)
...
 ament_export_include_directories(
-  include
+  include/${PROJECT_NAME}
 )
 
-ament_export_libraries(
-  ${library_name}
-  ${plugin_libs}
-)
+ament_export_libraries(${library_name})
```
SteveMacenski, line 277: Shouldn't we also export the plugins? Someone could plausibly use existing plugins as a base class and override methods for custom implementations. It would surprise me if _someone_ wasn't already doing that.

nav2_behavior_tree/src/generate_nav2_tree_nodes_xml.cpp
```diff
@@ -24,12 +23,31 @@
...
+  }
+}
+
 int main()
 {
   BT::BehaviorTreeFactory factory;
 
   std::vector<std::string> plugins_list;
-  boost::split(plugins_list, nav2::details::BT_BUILTIN_PLUGINS, boost::is_any_of(";"));
+  split_string(plugins_list, nav2::details::BT_BUILTIN_PLUGINS, ";");
```
SteveMacenski, line 50: See https://github.com/ros-navigation/navigation2/pull/4492: I replace with an existing nav2_util as part of fixing up some Jazzy release issues w.r.t. boost removal

SteveMacenski (changes requested):

SteveMacenski:
@clalancette thanks for all this work, especially this package + Smac Planner + MPPI + System Tests are the largest time hogs so even this alone probably cuts off minutes. It would be great if during these PRs you gave a ballpark of time saved for that package and/or stack. Nothing formal, just something to acknowledge your efforts for. 

I was curious to know what capacity you're working on this. Is this something of your own volition to help me / the community / yourself if doing stuff on the rare occasion that Nav2 needs assistance -- or is this something Intrinsic needs bc you're looking at starting to use it?

nav2_behavior_tree/CMakeLists.txt
clalancette: Ah, OK.  I wasn't sure about that, so I did not export them.  But I absolutely can, I'll look into doing that.

nav2_behavior_tree/src/generate_nav2_tree_nodes_xml.cpp
clalancette: Got it, will go looking for this function.

clalancette:
> Is this something of your own volition to help me / the community / yourself if doing stuff on the rare occasion that Nav2 needs assistance

It is me doing it on my own volition.  I've been using Navigation2 as a test bed for rmw_zenoh work, and I noticed that the compile times were high, so I decided to look into it.

> It would be great if during these PRs you gave a ballpark of time saved for that package and/or stack. Nothing formal, just something to acknowledge your efforts for.

I can give some numbers as I open PRs, but this is unfortunately a case where we won't see a lot of savings until we get the bulk of it done.  That is, most of savings are going to be when the downstream packages can rely on the exported targets, rather than the LIBRARIES/INCLUDE_DIRS.

One other thing I'll mention is that I started this journey on the `iron` branch, and there, this series made a *huge* difference (as stated in https://github.com/ros-navigation/navigation2/pull/4357, I went from 22 minutes to 9 minutes locally).  As I've been porting to the `main` branch, I've been seeing less of a difference.  I'm not sure if that is because `main` is just better to start with, or if it's because I haven't yet ported everything over yet.

nav2_behavior_tree/src/generate_nav2_tree_nodes_xml.cpp
clalancette: This should be fixed in the latest push now.

nav2_behavior_tree/CMakeLists.txt
clalancette: This is fixed in the latest push.

nav2_behavior_tree/CMakeLists.txt
clalancette: This is fixed in the latest push.

SteveMacenski:
Got it! Your help is much appreciated :saluting_face: 

> as a test bed for rmw_zenoh work

I had some threads with Angelo on this but haven't heard back.. I need to ping them. I wanted to add into our build matrix to run Zenoh once the binaries are available like we have to run Fast/Cyclone on our nightlies. That'll show issues if they exist. Even now, Fast and Cyclone fail on different things. 

> One other thing I'll mention is that I started this journey on the iron branch, and there, this series made a huge difference (as stated in https://github.com/ros-navigation/navigation2/pull/4357, I went from 22 minutes to 9 minutes locally). As I've been porting to the main branch, I've been seeing less of a difference. I'm not sure if that is because main is just better to start with, or if it's because I haven't yet ported everything over yet.

Must be - Iron is pretty close or essentially the same as `main`. Probably just not ported yet!

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -213,12 +206,30 @@ add_library(nav2_goal_updated_controller_bt_node SHARED plugins/decorator/goal_u
...
+    ${std_msgs_TARGETS}
+    ${std_srvs_TARGETS}
+  )
   target_compile_definitions(${bt_plugin} PRIVATE BT_PLUGIN_EXPORT)
 endforeach()
 
-install(TARGETS ${library_name}
-                ${plugin_libs}
+install(TARGETS ${library_name} ${plugin_libs}
+  EXPORT ${library_name}
```
SteveMacenski, line 232: plugins?

nav2_behavior_tree/CMakeLists.txt
```diff
@@ -229,39 +240,54 @@ set(GENERATED_DIR ${CMAKE_CURRENT_BINARY_DIR}/gen)
...
 ament_export_include_directories(
-  include
+  include/${PROJECT_NAME}
 )
 
-ament_export_libraries(
-  ${library_name}
-  ${plugin_libs}
-)
+ament_export_libraries(${library_name})
```
SteveMacenski, line 272: plugins?

nav2_behavior_tree/CMakeLists.txt
clalancette: Bah, sorry.

nav2_behavior_tree/CMakeLists.txt
SteveMacenski: I don't think the export targets is needed - or is it?

nav2_behavior_tree/CMakeLists.txt
clalancette: Actually, no.  This is correct as-is (it helps to sleep on it).

If you look at the documentation for [install](https://cmake.org/cmake/help/latest/command/install.html#signatures), it takes multiple targets, but only a single EXPORT name.  All of the targets will be associated with that name.

Thus, downstream packages that wish to depend on this package can do so with a combination of the EXPORT name and the library they are interested in.  For instance, this will EXPORT a library called `nav2_behavior_tree::nav2_behavior_tree`, which is the library that contains the `behavior_tree_engine`.  But it will also export `nav2_behavior_tree::nav2_compute_path_to_pose_action_bt_node`, which is one of the behavior tree nodes listed in `plugin_libs`.  And it will export the rest as well.

So this should do the right thing for all of the downstream consumers that wish to use the CMake target names.

nav2_behavior_tree/CMakeLists.txt
clalancette: This shouldn't be needed, as a single EXPORT name is used here.  See my comment in https://github.com/ros-navigation/navigation2/pull/4485#discussion_r1655001068.

clalancette:
> > It would be great if during these PRs you gave a ballpark of time saved for that package and/or stack. Nothing formal, just something to acknowledge your efforts for.
> 
> I can give some numbers as I open PRs

As a not very scientific number, locally I went from a compile time for all of navigation2 of 10 minutes 32 seconds on `main` to 10 minutes 19 seconds with this PR.

nav2_behavior_tree/CMakeLists.txt
SteveMacenski: OK. I would think having the plugins separated from the BT engine might be good since the folks using each probably have different tasks in mind which use only one or the other.

I'm not going to die on a hill about this, but seems like separating them would be good so that consumers of `behavior_tree_engine` aren't getting like 30+ and growing plugins. For instance, the BT Navigator or other's custom navigator packages based on the BT Navigator's plugin interface

SteveMacenski:
I think just waiting on the BT stuff to sync (and perhaps https://github.com/ros-navigation/navigation2/pull/4485#discussion_r1655183962?)? Otherwise, let me know when we can retrigger CI for this to go on!

clalancette:
> I think just waiting on the BT stuff to sync (and perhaps [#4485 (comment)](https://github.com/ros-navigation/navigation2/pull/4485#discussion_r1655183962)?)? Otherwise, let me know when we can retrigger CI for this to go on!

Yep, exactly.  Once those are synced out, we can rerun CI on this and it should pass.

SteveMacenski:
OK, if there's others you can open while we wait, happy to have a few of these open in parallel. Whatever works for you :-)

nav2_behavior_tree/CMakeLists.txt
clalancette: > I'm not going to die on a hill about this, but seems like separating them would be good so that consumers of `behavior_tree_engine` aren't getting like 30+ and growing plugins. For instance, the BT Navigator or other's custom navigator packages based on the BT Navigator's plugin interface

So that's the thing; they won't get all of those plugins by default.  The way that this works is that the EXPORT name acts as kind of a namespace, and the thing after the `::` is the individual library.  So linking against `nav2_behavior_tree::nav2_behavior_tree` will *only* link against `behavior_tree_engine`.  Linking against `nav2_behavior_tree::nav2_compute_path_to_pose_action_bt_node` will *only* link against that particular BT node.  And so on.

All of that said, if you prefer we can split this out to have a different namespace for the plugins.  It won't make a functional difference to downstream projects, but it may be clearer.  Just let me know if you prefer that, and I can make the change here.

SteveMacenski: Ohhh, nice! Got it, that makes sense. I'll look over this again with that in mind, I didn't realize that.

SteveMacenski (approved):

clalancette:
All right, so this now builds against the latest Rolling.  Unfortunately, it looks like the `test_goal_updater_node` test is failing now, but that is also failing on the `main` branch for me locally.  So I think this is blocked until that is fixed, then we can consider merging this.

SteveMacenski:
> Unfortunately, it looks like the test_goal_updater_node test is failing now, but that is also failing on the main branch for me locally. So I think this is blocked until that is fixed, then we can consider merging this.

Its failing in our CI too, I'm not sure why that popped up out of no where, but that's not related to this PR. We can merge this now I think and no worse for ware. Are you OK with that?

clalancette:
> Its failing in our CI too for the last ~week or so, I'm not sure why that popped up out of no where, but that's not related to this PR. We can merge this now I think and no worse for ware. Are you OK with that?

Yep, that sounds good to me.  Thanks!

PR #4486, opened 2024-06-25
Add a BT condition to check robot's vicinity with an AI model (currently ChatGPT)

### LINKED ISSUE #4484: How to implement path validity checking with ML models (GPT, LLM, etc)
Clearly an ML model could be quite useful in detecting a mud puddle, a cord, or a pit of hot lava which a robot shouldn't drive into, whereas Lidar and point clouds cannot distinguish any of those obstacles. Here's a quick example from ChatGPT. It identifies the cord obstacle correctly, then the lack of an obstacle. Many roboticists over the years would have killed to have functionality like this. (I think there are lots of other applications for ML when it comes to navigation, too.)

Screenshot from 2024-06-24 18-13-51

Related to PR #4483, it was suggested to move this check into the path validity server.

There are some details to work out. I think only a few meters in front of the robot should be checked for validity. It is not clear how to translate image to physical dimension. This will be more of a "fuzzy check." It would also be nice to allow different ML models to be swapped in/out.

However, I've worked with the ChatGPT API before and I think it will be quite easy, from that perspective.

AndyZe:
Note: I leave for vacation on 6/26, back on 6/30. So I won't be responsive during that time. I think that's good, it will give you time to noodle over how to handle the AI aspect. I'll write some tests when I get back.

## PR requirements:

- requires an OpenAI account. I believe even a free account will work; I pay $20/month.
- you need to set an environment variable like so. Get your key here: https://help.openai.com/en/articles/4936850-where-do-i-find-my-openai-api-key

`export OPENAI_API_KEY="..."`

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Fixes #4484  |
| Robotic platform tested on | TODO, RealSense D415 |

---

* Sends a request out to ChatGPT to look for anything blocking forward progress in the robot's vicinity
* Uses one camera image as input
* Useful for detecting things like cords on the ground, cracks, pools of water, small animals, hot lava ... etc... that Lidar and pointclouds won't
* It's a fuzzy check. It's not parameterized. Basically it uses the AI model's "common sense" 

chatgpt

pool

* Need to update the BT plugin tutorials

---

* I think there will be a need to call multiple AI models in the near future. It won't just be ChatGPT. Might want to think about a separate library/interface for calling/prompting models

#### For Maintainers:

### DIFF
--- a/nav2_behavior_tree/plugins/condition/ml_vicinity_condition.cpp
+++ b/nav2_behavior_tree/plugins/condition/ml_vicinity_condition.cpp
@@ -1,0 +1,133 @@
+// Copyright (c) 2024 Andy Zelenak
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "nav2_behavior_tree/plugins/condition/ml_vicinity_condition.hpp"
+#include <chrono>
+#include <cstdlib>
+#include <string>
+#include <cv_bridge/cv_bridge.hpp>
+#include <curl/curl.h>
+#include <nlohmann/json.hpp>
+
+namespace nav2_behavior_tree
+{
+
+MLVicinityCondition::MLVicinityCondition(
+  const std::string & condition_name,
+  const BT::NodeConfiguration & conf)
+: BT::ConditionNode(condition_name, conf)
+{
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+
+  server_timeout_ = config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
+  getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+  std::string image_topic;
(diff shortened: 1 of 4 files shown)


nav2_behavior_tree/plugins/condition/ml_vicinity_condition.cpp
```diff
@@ -0,0 +1,121 @@
...
+BT::NodeStatus MLVicinityCondition::tick()
+{
+  if (promptAIModel())
+  {
+    return BT::NodeStatus::SUCCESS;
+  }
+  return BT::NodeStatus::FAILURE;
+}
+
+bool MLVicinityCondition::promptAIModel()
```
AndyZe, line 49: i figured i'd cram everything related to ChatGPT into this one function so it can be swapped out easily, although it's not the most efficient way

SteveMacenski:
I'm going to again ask that this is closed and to have a fuller conversation before asking for reviews. See ticket you filed #4484

AndyZe:
Why so grumpy, Steve? I don't want it to be like that.

If you let me target Humble branch, I would be able to develop locally much easier. Rolling is an inconvenience.

SteveMacenski:
> Why so grumpy, Steve? I don't want it to be like that.

Sorry, not grumpy, this is boilerplate text, I get students frequently that submit broken PRs and trigger dozens of jobs to get it to build since they're not attempting to build locally and it just sucks up all of our build time and makes things hard to review.

In general though, I expect that PRs are in a complete I-want-a-review-to-merge state. If we're still talking more 'how should we do this' and 'does something like this belong in Nav2' kind of stuff, a ticket for discussion with perhaps a link a branch that is in progress to poke around is better process IMO.  Carpet bombing maintainers with emails and notifications about broken builds in a PR doesn't look good :laughing:

PR #4506, opened 2024-07-01
Fix map to world coordinate conversion

HovorunBh:
---

## Basic Info

| ------ | ----------- |

---

For reproducing the issue see the test I added 

We noticed that ComputePathThroughPoses with SmacPlannerHybrid when asked to generate path through poses that are on the same line (in this example: [{x: 2, y: 0}, {x: 4, y: 0}, {x: 6, y: 0}]) generates path like this: 
Screenshot from 2024-07-01 13-33-38

It is assumed that end of path1 is the start of path2, but when end of path1 is converted from map to world, and then back to map it ends up in a different costmap cell.

When we convert world to map coordinates we [add 0.5](https://github.com/ros-navigation/navigation2/blob/30f028d1bf5a4cb3ba429061823135ba6b394ca9/nav2_costmap_2d/src/costmap_2d.cpp#L299)

Therefore when we convert map to world I think we should subtract 0.5. Otherwise we end up in the next costmap cell
With this change path looks like more straight: 
Screenshot from 2024-07-01 13-35-33

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/utils.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/utils.hpp
@@ -46,9 +46,9 @@ inline geometry_msgs::msg::Pose getWorldCoords(
 {
   geometry_msgs::msg::Pose msg;
   msg.position.x =
-    static_cast<float>(costmap->getOriginX()) + (mx + 0.5) * costmap->getResolution();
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

SteveMacenski:
Thanks! All good points, well explained, and with visual examples!

tonynajjar:
Awesome find and PR Bohdan, let's backport the fix from Nav2 directly and close this PR?

PR #4507, opened 2024-07-02
fix(nav2_costmap_2d): make obstacle layer not current on enabled toggle

bektaskemal:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/issues/4465 |
| Robotic platform tested on | Turtlebot Sim|

---

- Made obstacle layer consistent with other layers and make current false after enabled is toggled. So isCurrent function of costmap will return false and require update.

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -307,8 +307,11 @@ ObstacleLayer::dynamicParametersCallback(
         max_obstacle_height_ = parameter.as_double();
       }
     } else if (param_type == ParameterType::PARAMETER_BOOL) {
-      if (param_name == name_ + "." + "enabled") {
+      if (param_name == name_ + "." + "enabled" && enabled_ != parameter.as_bool()) {
         enabled_ = parameter.as_bool();
+        if (enabled_) {
+          current_ = false;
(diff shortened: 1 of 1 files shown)


nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -307,8 +307,11 @@ ObstacleLayer::dynamicParametersCallback(
         max_obstacle_height_ = parameter.as_double();
       }
     } else if (param_type == ParameterType::PARAMETER_BOOL) {
-      if (param_name == name_ + "." + "enabled") {
+      if (param_name == name_ + "." + "enabled" && enabled_ != parameter.as_bool()) {
```
SteveMacenski, line 310: Remove, we need to do all parameter updates, this would not allow you to enable it.

SteveMacenski (changes requested):
Did you test this to get the behavior we discussed?

SteveMacenski (approved):

bektaskemal:
> Did you test this to get the behavior we discussed?

Yes, it works as we discussed.

SteveMacenski:
Thanks for the discussion / contribution!

PR #4527, opened 2024-07-08
Fix backward motion for graceful controller

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Gazebo classic |

---

- Fix computation of linear and angular velocities to allow correct backward motion.
- It can be backported to other branches.

---

#### For Maintainers: 

### DIFF
--- a/nav2_graceful_controller/src/smooth_control_law.cpp
+++ b/nav2_graceful_controller/src/smooth_control_law.cpp
@@ -43,8 +43,7 @@ void SmoothControlLaw::setSlowdownRadius(double slowdown_radius)
 }
 
 void SmoothControlLaw::setSpeedLimit(
-  const double v_linear_min, const double v_linear_max,
-  const double v_angular_max)
+  const double v_linear_min, const double v_linear_max, const double v_angular_max)
 {
   v_linear_min_ = v_linear_min;
   v_linear_max_ = v_linear_max;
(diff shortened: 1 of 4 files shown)


SteveMacenski (approved):
@mikeferguson can you please review this as well? I want to make sure I'm not missing anything subtle. This is with respect to https://github.com/open-navigation/opennav_docking/issues/47

nav2_graceful_controller/src/smooth_control_law.cpp
```diff
@@ -59,13 +58,15 @@ geometry_msgs::msg::Twist SmoothControlLaw::calculateRegularVelocity(
...
+  // Invert the curvature if the robot is moving backwards
+  curvature = backward ? -curvature : curvature;
 
   // Adjust the linear velocity as a function of the path curvature to
   // slowdown the controller as it approaches its target
   double v = v_linear_max_ / (1.0 + beta_ * std::pow(fabs(curvature), lambda_));
 
   // Slowdown when the robot is near the target to remove singularity
-  v = std::min(v_linear_max_ * (ego_coords.r / slowdown_radius_), v);
+  v = std::min(v_linear_max_ * fabs(ego_coords.r / slowdown_radius_), v);
```
SteveMacenski, line 69: isn't `r` always positive now? Shouldn't `(ego_coords.r / slowdown_radius_)` always be positive?

SteveMacenski (changes requested):

nav2_graceful_controller/src/smooth_control_law.cpp
ajtudela: yeap, that is not necessary anymore.

ajtudela:
The CI failed but its not related to this PR.

SteveMacenski (approved):

HappySamuel:
Hi

Will this PR being merged into humble version as well?
As i am currently still using humble version, but found that humble version not yet being updated with the latest fix.

SteveMacenski:
I believe it already is in https://github.com/open-navigation/opennav_docking -- but if that's wrong, please open a PR and happy to merge!

PR #4543, opened 2024-07-18
Add min_distance_to_obstacle parameter to RPP

doisyg:
---

## Basic Info

| ------ | ----------- |

---

This PR adds the parameter `min_distance_to_obstacle` to stop in front of an obstacle based on distance and not only time to collision

---

#### For Maintainers: 


nav2_regulated_pure_pursuit_controller/src/collision_checker.cpp
```diff
@@ -90,16 +90,21 @@ bool CollisionChecker::isCollisionImminent(
...
   curr_pose.y = robot_pose.pose.position.y;
   curr_pose.theta = tf2::getYaw(robot_pose.pose.orientation);
+  double accumulated_distance = 0.0;
 
   // only forward simulate within time requested
   int i = 1;
-  while (i * projection_time < params_->max_allowed_time_to_collision_up_to_carrot) {
+  while (i * projection_time < params_->max_allowed_time_to_collision_up_to_carrot ||
+    (accumulated_distance < params_->min_distance_to_obstacle &&
+    projection_time * linear_vel > 0.01))
```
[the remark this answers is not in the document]
SteveMacenski: I suppose maybe I'm also missing the detail: do you want these both conditions to need to be satisfied to stop, or just one or other other?

If, for instance, you want both, which I think you do, we could instead do something like:

`max_allowed_time_to_collision_check = std::max(params_->max_allowed_time_to_collision_up_to_carrot, params_->min_distance_to_obstacle * linear_vel);` and use `max_allowed_time_to_collision_check` in the `while` loop condition instead of the parameter. I think that would make this intent more clear.

SteveMacenski:
@doisyg what's the story here?

nav2_regulated_pure_pursuit_controller/src/collision_checker.cpp
doisyg: Agreed, using now:

```
    max_allowed_time_to_collision_check = std::max(
        params_->max_allowed_time_to_collision_up_to_carrot,
        params_->min_distance_to_obstacle / std::max(std::abs(linear_vel), 0.01)
```

nav2_regulated_pure_pursuit_controller/src/collision_checker.cpp
```diff
@@ -92,8 +92,15 @@ bool CollisionChecker::isCollisionImminent(
   curr_pose.theta = tf2::getYaw(robot_pose.pose.orientation);
 
   // only forward simulate within time requested
+  double max_allowed_time_to_collision_check = params_->max_allowed_time_to_collision_up_to_carrot;
+  if (params_->min_distance_to_obstacle > 0.0) {
+    max_allowed_time_to_collision_check = std::max(
+        params_->max_allowed_time_to_collision_up_to_carrot,
+        params_->min_distance_to_obstacle / std::max(std::abs(linear_vel), 0.01)
```
SteveMacenski: 0.01? Should this be something like `min_approach_linear_velocity`?

doisyg: Indeed, thanks !

PR #4557, opened 2024-07-24
Switch nav2_map_server to use modern CMake idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Follow up to #4357  |

---

This commit does a number of things:

1. Switches to using target_link_libraries in nav2_map_server, which gives us finer-grained control over what dependencies are exported to downstream as public or private.
2. Moves the include directory down one level to include/${PROJECT_NAME}, which is best practice in ROS 2 since Humble.
3. Makes sure to export nav2_map_server as a CMake target, so downstream users of it can use that target.
4. Adds in `lifecycle_msgs` as a dependency.
5. Possibly most controversially, changes one test.  While running the tests I noted some odd output from `test_map_saver_cli.cpp`.  In particular, it was complaining that `__node:=map_saver_test_node` was deprecated.  That is true, but adding in `--ros-args --remap` didn't work, and that is because a space was missing from the front of that.  With that change now in place, I also had to update the return value from the command, since it now returns 256 instead of 65280.

None needed

---

The rest of the packages in this repository need a similar treatment (PRs will be upcoming).

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/CMakeLists.txt
+++ b/nav2_map_server/CMakeLists.txt
@@ -4,24 +4,27 @@ project(nav2_map_server)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake_modules)
 
 find_package(ament_cmake REQUIRED)
-find_package(nav2_common REQUIRED)
-find_package(rclcpp REQUIRED)
-find_package(rclcpp_lifecycle REQUIRED)
-find_package(rclcpp_components REQUIRED)
+find_package(GRAPHICSMAGICKCPP REQUIRED)
+find_package(lifecycle_msgs REQUIRED)
 find_package(nav_msgs REQUIRED)
+find_package(nav2_common REQUIRED)
 find_package(nav2_msgs REQUIRED)
-find_package(yaml_cpp_vendor REQUIRED)
+find_package(nav2_util REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rclcpp_components REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(tf2 REQUIRED)
-find_package(nav2_util REQUIRED)
-find_package(GRAPHICSMAGICKCPP REQUIRED)
+find_package(yaml_cpp_vendor REQUIRED)
 find_package(yaml-cpp REQUIRED)
 
 nav2_package()
 
-include_directories(include)
-
 set(map_server_executable map_server)
+
+set(library_name ${map_server_executable}_core)
+
+set(map_io_library_name map_io)
 
 set(map_saver_cli_executable map_saver_cli)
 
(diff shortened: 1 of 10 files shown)


nav2_map_server/CMakeLists.txt
```diff
@@ -162,5 +170,7 @@ ament_export_libraries(
   ${library_name}
   ${map_io_library_name}
 )
-ament_export_dependencies(${map_io_dependencies} ${map_server_dependencies} yaml-cpp)
+ament_export_dependencies(nav_msgs nav2_msgs nav2_util rclcpp rclcpp_lifecycle)
```
SteveMacenski, line 173: yaml-cpp? I think a user added that because they needed that exported 

tf2, rclcpp_components?

nav2_map_server/test/map_saver_cli/test_map_saver_cli.cpp
```diff
@@ -92,9 +92,9 @@ TEST(MapSaverCLI, CLITest)
     std::string(
     "ros2 run nav2_map_server map_saver_cli "
     "-t map_failure --occ 100 --free 2 --mode trinary --fmt png -f ") + file_path +
-    std::string("--ros-args __node:=map_saver_test_node");
+    std::string(" --ros-args --remap __node:=map_saver_test_node");
   return_code = system(command.c_str());
-  EXPECT_EQ(return_code, 65280);
+  EXPECT_EQ(return_code, 256);
```
SteveMacenski, line 97: No note assuming CI test is happy, but noting this for myself to check

nav2_map_server/CMakeLists.txt
clalancette: `yaml-cpp`, `tf2`, and `rclcpp_components` do not need to be exported, because they are PRIVATE dependencies.  That is, downstream consumers do not need to know about the dependency in order to successfully build/link.

nav2_map_server/test/map_saver_cli/test_map_saver_cli.cpp
clalancette: Ah, it looks like there are bunch of failing external symbols in the tests.  That means that something is wrong with this PR.  I'll take a closer look and see if I can reproduce locally.

clalancette: So I can't reproduce, and looking at the code again there is nothing obviously wrong jumping out at me.

I'm not that familiar with CircleCI, but in the job below it never seems to run the "Setup Workspace" or "Build Workspace" steps that are listed in https://github.com/ros-navigation/navigation2/blob/main/.circleci/config.yml .  Looking at those logs would be helpful for me to try and reproduce.

@SteveMacenski Do you know why those steps aren't present in that run, and/or how I might get access to those logs?

SteveMacenski: The `release_build` job below does the building of the main software stack (`system_build` does the system tests separately). The job that's failing (`release_test`) is only doing the `colcon test` bits on those built and cached artifacts from the builds.

You can find the build artifacts here: https://app.circleci.com/pipelines/github/ros-navigation/navigation2/12291/workflows/ec12a613-f542-4479-9c2b-640744e3fbaf/jobs/37306/artifacts
You can find the test artifacts here: https://app.circleci.com/pipelines/github/ros-navigation/navigation2/12291/workflows/ec12a613-f542-4479-9c2b-640744e3fbaf/jobs/37308/artifacts

I'm also retriggering the build from the start to validate its not a fluke, but it does seem like its a problem

SteveMacenski: Something else worth trying is to update `v26` to `v27` in the 3x places in the circle config file to break the cache which may contain the previous works and rebuild from scratch.

clalancette: Ah, thanks for pointing out where it was built.  Following that, I was able to see that the CircleCI job uses a number of mixins, and compiling locally with those mixins I can reproduce.  Now to see what is going on here.

clalancette: All right, I found the issue.  [66e300b](https://github.com/ros-navigation/navigation2/pull/4557/commits/66e300b43bf5b1a3a5246bd591d3593d211d1a66) should fix it.

clalancette:
> Blaming `Executor::spin_some_impl(std::chrono::nanoseconds max_duration, bool exhaustive)`, I see there's been alot of poking around there recently in the last few months, and a PR to fix some other regression in May, which makes me think maybe there's another regression lurking in that corner.

Yes, there has been a lot of poking at that recently.

Something that is new-ish here is https://github.com/ros2/rclcpp/pull/2517, which was merged back on May 2nd, but may or may not have been released until recently (it is a little hard to tell exactly when something was merged).  If you go back to rclcpp version 28.2.0, does that make a difference?

SteveMacenski:
I tested and no double-spin is required for this test with 28.2.0 (i.e. patch in https://github.com/ros-navigation/navigation2/pull/4558 to fix our CI is not required). So that's new. 

Note that we've seen some 'warm up spins' required since the initial Jazzy migration to 24.04 https://github.com/ros-navigation/navigation2/pull/4298 (late May) which introduced a double spin  or 'while not received -> spin some' logic. Rolling's release on April 30 had 28.2.0 https://discourse.ros.org/t/new-packages-for-ros-2-rolling-ridley-2024-04-30/37485 so we had also seen it then in some other cases - but perhaps those were fixed by the later changes and no longer required

SteveMacenski:
Your changes fixed it (minus the test we're discussing)

PR #4558, opened 2024-07-24
Fix goal updater for CI

SteveMacenski:


SteveMacenski:
@clalancette should I go ahead and merge this or do you want me to hold off for an rclcpp discussion?

clalancette:
> @clalancette should I go ahead and merge this or do you want me to hold off for an rclcpp discussion?

So I think what we should do here is to open an issue against rclcpp with all of the data we currently have (a self-contained reproducer would be ideal, but not required).  Once that issue is open, I would then go ahead and merge this workaround for now.

@SteveMacenski Does that make sense to you?  Do you mind doing that?

SteveMacenski:
@clalancette done: https://github.com/ros2/rclcpp/issues/2589 

I would like to  merge this however to get our CI green, but its easy enough to revert for testing the failure case as part of the development for a fix

tonynajjar:
@SteveMacenski I noticed that this was not backported to Jazzy branch. Shouldn't it be?

PR #4564, opened 2024-07-25
Switch nav2_amcl to use modern CMake idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Followup to #4357  |

---

This PR updates nav2_amcl to use modern CMake idioms.

This change is a little unusual in that it rearranges the hierarchy of CMakeLists.txt files.  This is because the individual libraries (like map_lib) are used in some of the other libraries, and CMake was complaining about that. So I flattened the whole thing out and now build all of the libraries at the top-level.

This PR also contains a couple of fixes to previous packages where I failed to properly export the include directory.

Finally, this PR goes ahead and removes quite a bit of dead code from this package.  This has been dead for quite a while, so I think it makes sense to remove it at this point.

None needed.

---

Changing the rest of the packages in Navigation2 to use modern CMake idioms (PRs will be upcoming).

#### For Maintainers: 


nav2_amcl/include/nav2_amcl/pf/pf_pdf.hpp
```diff
@@ -30,9 +30,6 @@
 
 #include "nav2_amcl/pf/pf_vector.hpp"
 
-// #include <gsl/gsl_rng.h>
```
SteveMacenski, line 33: Why add?

nav2_amcl/include/nav2_amcl/pf/pf_vector.hpp
```diff
@@ -51,38 +51,15 @@ typedef struct
 // Return a zero vector
 pf_vector_t pf_vector_zero(void);
 
-// Check for NAN or INF in any component
```
SteveMacenski, line 54: Why add?

nav2_amcl/src/pf/pf.c
```diff
@@ -236,16 +236,6 @@ int pf_update_converged(pf_t * pf)
   return 1;
 }
 
-// Update the filter with some new action
```
SteveMacenski, line 239: Why add? (for all new commented out code, won't mention again)

nav2_map_server/CMakeLists.txt
```diff
@@ -166,7 +166,7 @@ if(BUILD_TESTING)
   add_subdirectory(test)
 endif()
 
-ament_export_include_directories(include)
+ament_export_include_directories("include/${PROJECT_NAME}")
```
SteveMacenski, line 169: Intentional change here (ditto with voxel grid)? Can you explain why we export this way (other refactor packages require too?)

nav2_amcl/include/nav2_amcl/pf/pf_pdf.hpp
clalancette: Basically because this code has been commented out since 2018; I think it is time to remove it :).  Anyone trying to resurrect it will probably have to reimplement a bunch of stuff anyway, so I don't particularly see a need to keep it around anymore.

That said, if you'd like to discuss it separately I'm happy to put it into a separate PR.

nav2_map_server/CMakeLists.txt
clalancette: Right, this was a bug in my two previous PRs.  I fixed them here because it is somewhat important to fix.

What this fixes is the CMake variable called `nav2_map_server_INCLUDE_DIRS`.  If any downstream consumers were using that variable, then it would point to the wrong subdirectory currently.  That is, the header files are installed to `/path/to/nav2_map_server/include/nav2_map_server/nav2_map_server`, but without this fix `nav2_map_server_INCLUDE_DIRS` points to `/path/to/nav2_map_server/include`.  That means that a `#include "nav2_map_server/foo.hpp"` wouldn't work currently, because it would have to be `#include "nav2_map_server/nav2_map_server/foo.hpp"`.

We didn't notice this in the previous PRs because `ament` prefers to use targets when they are available, so it hardly ever falls back to `nav2_map_server_INCLUDE_DIRS`.  But it is possible that there are downstream projects that do use it, so it may as well be correct.

nav2_amcl/include/nav2_amcl/pf/pf_pdf.hpp
SteveMacenski: I suppose I'd rather keep all the code that contained AMCL intact for whoever might come along and care to really redevelop it. I can't gauge from a glance how useful / important any of those utilities or old methods are, so I'd prefer to keep them commented out than deleted so that there's an explicit historical record of it. AMCL and Costmap2D are two "special" packages in my book that get a bit of different treatment for code treatment since there's alot of history there, much of which I know, but I wasn't there and I don't know when things like this might be back to be helpful in the future.

clalancette: All right, fair enough.  I've just rebased onto the latest and taken out the changes that removed the dead code.

SteveMacenski (approved):

PR #4572, opened 2024-07-29
Fix backward moving while docking

iradionov:
## Basic Info

| Robotic platform tested on | simulation |

---

This [commit](https://github.com/ros-navigation/navigation2/commit/f73fa2499b54bfb346fa3dcdd7a5a200ba349752) ruins the backward docking process because the phi angle is rotated (+M_PI) and a robot tries to approach with wrong direction.  

#### For Maintainers: 

### DIFF
--- a/nav2_docking/opennav_docking/src/docking_server.cpp
+++ b/nav2_docking/opennav_docking/src/docking_server.cpp
@@ -417,16 +417,12 @@ bool DockingServer::approachDock(Dock * dock, geometry_msgs::msg::PoseStamped & 
 
     // Make sure that the target pose is pointing at the robot when moving backwards
     // This is to ensure that the robot doesn't try to dock from the wrong side
-    if (dock_backwards_) {
-      target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
-        tf2::getYaw(target_pose.pose.orientation) + M_PI);
-    }
 
     // The control law can get jittery when close to the end when atan2's can explode.
     // Thus, we backward project the controller's target pose a little bit after the
     // dock so that the robot never gets to the end of the spiral before its in contact
     // with the dock to stop the docking procedure.
-    const double backward_projection = 0.25;
+    const double backward_projection = dock_backwards_ ? -0.25 : 0.25;
     const double yaw = tf2::getYaw(target_pose.pose.orientation);
     target_pose.pose.position.x += cos(yaw) * backward_projection;
     target_pose.pose.position.y += sin(yaw) * backward_projection;


ajtudela:
Hi, 

This is already fixed in the main branch. Your dock detector must publish a pose pointing in the direction of the motion, regardless of whether the robot is moving forward or backward.

image
Don't look the robot, just the orientation of the dock.


Please, check this issue for more info: https://github.com/open-navigation/opennav_docking/issues/47

ajtudela:
@SteveMacenski Two days left!! 😃 

Yes, the direction of the dock arrow must be the same for both forward and backwards. (If the docking is backwards, the controller will reverse it internally, so the trajectory is generated correctly).

I agree, it should be clear from the documentation that the orientation of the dock has to be the same, regardless of whether the docking is forward or backwards.

iradionov:
Hi, guys!
Thank you for the response and the details. It seems that it's not only me have a misunderstanding in terms of goal direction.
We tried this algorithm in simple cases (moving on a straight line) with a "wrong" direction and it worked somehow. Yesterday I tried this old version on complex cases and yes, the goal direction should be rotated when movement is backward. 
For us, this point wasn't clear and in our case (a not small Ackermann-type robot) a more natural way is looking at the docking process as a parking. So the goal direction should be matched with the car's direction at this point.
Something like this
forward
backward

What is the reason why we should invert the goal direction for backward moving?

SteveMacenski:
I don't understand the question fully or what I'm looking at in terms of the images, the frames are unlabeled and not described. They also look both like adequate solutions so I don't see 'the problem' and 'the solution' curves. 

> the goal direction should be rotated when movement is backward

Precision here would be helpful. *You* shouldn't need to rotate anything specially for reversing. It sounds like your dock's detected orientation need to be rotated but that would have been required for forward or backward. It was just the wrong orientation - but to be fair - I don't document what that should be anywhere, so this is a very useful conversation for a gap in docs that I should fill.  Then, inline in the software, we handle any inversions required without your need for modification to make reversing work properly. 

Is that correct? If so, I think that makes this PR not needed - anything else we need to discuss?

> What is the reason why we should invert the goal direction for backward moving?

Precision would be helpful, I don't know if you mean your dock's detected orientation or the inversion actually in the software.

iradionov:
As I understand @ajtudela, on this image 
image
the left arrow should be rotated; this is a wrong setup. Am I right?

ajtudela:
> the arrow (docking pose) should be rotated; this is a wrong setup. Am I right? If so, this PR is not actual. But for our team, this representation seems more natural for a car-like robot. Thanks for your replies.

Sorry for the late replay. You shouldn't have to rotate anything for backwards or forward docking, because the pose of the dock doesn't depend on the movement of the robot, it's always the same. 

The server, internally, will reverse the dock orientation if the movement is backwards, but you only have to worry about publishing the pose correctly. 

I'm not sure if this answers your questions, please let me know.

SteveMacenski:
@iradionov to be clear, are you saying with the current state of main, setting the orientation of the dock as suggested, is still having a problem for you?

iradionov:
It's not a problem. I understand this idea, but as mentioned ahead this representation seems more natural for our team since we work with a car-like robot. 
Thanks for your time.

PR #4590, opened 2024-08-01
Switch nav2_mppi_controller to modern CMake idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Follow up to #4357  |

---

Change nav2_mppi_controller to use modern CMake idioms:
1. Use target_compile_options where useful.
2. Change from ament_target_dependences() to target_link_libraries()
3. Move the header files down one subdirectory, which is best practice since Humble.
4. Export the CMake target so that downstream libraries can use it.
5. Clean up the set of dependencies.

None needed.

---

This is part of a larger series of PRs to switch Navigation2 to modern CMake idioms.  There will be follow-up PRs converting more of the packages.

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/CMakeLists.txt
+++ b/nav2_mppi_controller/CMakeLists.txt
@@ -8,36 +8,24 @@ set(XTENSOR_USE_TBB 0)
 set(XTENSOR_USE_OPENMP 0)
 set(XTENSOR_USE_XSIMD 1)
 
-# set(XTENSOR_DEFAULT_LAYOUT column_major)  # row_major, column_major
-# set(XTENSOR_DEFAULT_TRAVERSAL row_major)  # row_major, column_major
-
 find_package(ament_cmake REQUIRED)
+find_package(angles REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(nav2_common REQUIRED)
+find_package(nav2_core REQUIRED)
+find_package(nav2_costmap_2d REQUIRED)
+find_package(nav2_util REQUIRED)
+find_package(nav_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(visualization_msgs REQUIRED)
 find_package(xsimd REQUIRED)
 find_package(xtensor REQUIRED)
-
-include_directories(
-  include
-)
-
-set(dependencies_pkgs
-  rclcpp
-  nav2_common
-  pluginlib
-  tf2
-  geometry_msgs
-  visualization_msgs
-  nav_msgs
-  nav2_core
-  nav2_costmap_2d
-  nav2_util
-  tf2_geometry_msgs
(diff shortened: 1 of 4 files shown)


nav2_mppi_controller/CMakeLists.txt
```diff
@@ -65,63 +53,126 @@ if(COMPILER_SUPPORTS_FMA)
...
+  tf2::tf2
+  tf2_geometry_msgs::tf2_geometry_msgs
+  tf2_ros::tf2_ros
+  ${visualization_msgs_TARGETS}
+  xtensor
+  xtensor::optimize
+  xtensor::use_xsimd
 )
 
 add_library(mppi_critics SHARED
```
SteveMacenski, line 92: Morbidly curious - why the reorder?

nav2_mppi_controller/CMakeLists.txt
```diff
@@ -65,63 +53,126 @@ if(COMPILER_SUPPORTS_FMA)
...
-  src/trajectory_visualizer.cpp
-  src/path_handler.cpp
-  src/parameters_handler.cpp
   src/noise_generator.cpp
+  src/optimizer.cpp
+  src/parameters_handler.cpp
+  src/path_handler.cpp
+  src/trajectory_visualizer.cpp
+)
+target_compile_options(mppi_controller PUBLIC -fconcepts -O3 -finline-limit=10000000 -ffp-contract=fast -ffast-math -mtune=generic)
```
SteveMacenski, line 66: Added `-fconcepts`? We should be on C++17 which doesn't need that for the `auto` issue

clalancette: `-fconcepts` was already there, from the "old" line 97.  That said, I can remove it if you think we don't need it anymore.

nav2_mppi_controller/CMakeLists.txt
clalancette: Just for alphabetical order.  In my opinion, it just makes it easier when adding new files in the future; there is an obvious place for it to go .  But it isn't required, so if you want me to put it back like it was I can.

SteveMacenski: No, not a problem, just curiosity

SteveMacenski:
Do you have a list you're working off of? Might be nice to track status in the ticket so we know what's done / coming 

I really appreciate your work here. What are the compiling times down to now? MPPI and the behavior tree packages were two of the longest packages

clalancette:
> Do you have a list you're working off of? Might be nice to track status in the ticket so we know what's done / coming

We are pretty far along.  I'll put a comment in the other ticket, but the packages that are left are:

* [ ] nav2_navfn_planner
* [ ] nav2_planner
* [ ] nav2_regulated_pure_pursuit_controller
* [ ] nav2_rotation_shim_controller
* [ ] nav2_rviz_plugins
* [ ] nav2_system_tests
* [ ] nav2_smoother
* [ ] opennav_docking
* [ ] nav2_waypoint_follower
* [ ] nav2_smac_planner
* [ ] nav2_graceful_controller
* [ ] dwb_plugins
* [ ] dwb_critics
* [ ] dwb_core
* [ ] costmap_queue

> I really appreciate your work here. What are the compiling times down to now?

Right now, we are saving about 30 seconds from before my work.  I expect to get a bit more than that, particular when we reach nav2_system_tests, but we'll see as we get there.

Issue #4304, opened 2024-05-07
Navigate through poses terminates pre-maturely when incidentally passing over final poses


thegindi:
## Bug report

 
**Required Info:**
 
- Operating System:
  -Ubuntu 22.04.3
- ROS2 Version:
  - Humble Binaries
- Version or commit hash:
  - ros-humble-navigation2   1.1.12-1jammy.20240126.081240
- DDS implementation:
  - not sure
 
#### Steps to reproduce issue
Follow steps in https://navigation.ros.org/getting_started/index.html to run default TB3 simulation
```
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False
```
1. set 2D pose estimate
2. Switch to waypoint / Nav through poses mode
3. Set 3 waypoints where waypoint 1 and 3 are essentially the same and waypoint 2 is distinct

 https://github.com/open-navigation/navigation2/assets/146736538/65447a7a-4f3f-4958-85df-fce30000bace

#### Expected behavior
the turtlebot would travel to the waypoints in order.
#### Actual behavior
The turtlebot thinks it has achieved its goal as soon as it reaches waypoint 1 (which also happens to be very similar to the final waypoint)
#### Additional information
This issue has already been raised before https://github.com/open-navigation/navigation2/issues/2723 however I believe there was some miscommunication. https://github.com/open-navigation/navigation2/issues/2622 This issue was believed to have solved the problem yet they are different issues.
 
I believe what is happening here is that the simpleGoalChecker is only checking the robots current pose against the final pose in the path. Hence, if the robot happens to run over this final pose while it is still halfway through completing its mission, it will simply think it has reached the end.
 
I have currently implemented a solution which implements a new BT node "removePassedPoses" (this node essentially replicates the "removePassedGoals" plugin except it will remove poses from the path as the robot progresses.  
 
Then adding the following snippet in in controller_server::isGoalReached()
```
if(path.poses.size()<3){
    return goal_checkers_[current_goal_checker_]->isGoalReached(
    pose.pose, transformed_end_pose.pose,
    velocity);
  }else{
    return false;
  }
```
 

----

SteveMacenski:
Those tickets were mostly about intersecting paths, not intersecting goals, hence the additional element of the goal checker. 

[...]

PR #4593, opened 2024-08-04
Add current path to GoalChecker interface

ottojo:
The controller server uses a configurable goal checker to determine if the robot has completed its current path as defined by the global plan. Current goal checkers compare the goal (end of path) pose and twist to the current robot state, but that is not sufficient in all cases.

In some use cases, the path will visit the goal multiple times, and the goal might coincide with the starting position. It is still desired that the robot follows the entire path, instead of immediately ending navigation once the goal pose is reached.

This PR adds a parameter to the GoalChecker interface to inform the goal checker of the current path, which enables building more sophisticated goal checkers that (for example) take progress along the path into account.

We already use such a goal checker internally, which just subscribes to the global plan via the appropriate ROS topic, but i think this here is the cleaner solution (and also avoids race conditions of checking against an old path in the goal checker...)

I don't have a nice testing setup and a goal checker using this interface yet (other than our own robot and the mentioned goal checker), which is why I mark this as draft for now.

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4304 |
| Robotic platform tested on | (WIP) |

---

* [x] Extend `GoalChecker::isGoalReached` interface with argument for current path
* [x] Add this argument to the in-tree goal checker implementations
* [ ] Add a goal checker which checks for path-completion
* [ ] Define a test-scenario with global plan that returns to start pose

* [ ] Migration guide for out-of-tree goal checkers

---

* Implementing more sophisticated goal checkers

#### For Maintainers:

### DIFF
--- a/nav2_controller/include/nav2_controller/plugins/simple_goal_checker.hpp
+++ b/nav2_controller/include/nav2_controller/plugins/simple_goal_checker.hpp
@@ -66,7 +66,7 @@ public:
   void reset() override;
   bool isGoalReached(
     const geometry_msgs::msg::Pose & query_pose, const geometry_msgs::msg::Pose & goal_pose,
-    const geometry_msgs::msg::Twist & velocity) override;
+    const geometry_msgs::msg::Twist & velocity, const nav_msgs::msg::Path & current_path) override;
   bool getTolerances(
     geometry_msgs::msg::Pose & pose_tolerance,
     geometry_msgs::msg::Twist & vel_tolerance) override;
--- a/nav2_controller/include/nav2_controller/plugins/stopped_goal_checker.hpp
+++ b/nav2_controller/include/nav2_controller/plugins/stopped_goal_checker.hpp
@@ -61,7 +61,7 @@ public:
     const std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros) override;
   bool isGoalReached(
     const geometry_msgs::msg::Pose & query_pose, const geometry_msgs::msg::Pose & goal_pose,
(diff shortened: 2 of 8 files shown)


SteveMacenski:
> We already use such a goal checker internally, which just subscribes to the global plan via the appropriate ROS topic, but i think this here is the cleaner solution (and also avoids race conditions of checking against an old path in the goal checker...)

Agreed!

> I don't have a nice testing setup and a goal checker using this interface yet (other than our own robot and the mentioned goal checker), which is why I mark this as draft for now.

I see you're marking this as a solution to the nav through poses ticket. I see this as a good API-change step needed to implement the goal checker specifics for that and I approve of this intermediate step. I think we do need to adjust the goal checker(s) or make a new one (probably just update the existing?) to fully close the ticket and actually use this API change. 

So, I like where this is going - no objections and I look forward to the next step!

SteveMacenski:
@ottojo following up here - have you had a chance to work on this?

josephduchesne:
I'm also interested in this, and can potentially help out if it's blocked.

Not having this is a real headache for me, and I'd rather not duplicate @ottojo's patch solution (and make yet another proprietary goal checker that subscribes to the current path).

SteveMacenski:
@josephduchesne I think we'd appreciate the help! This has been blocked / not updated in some time. Taking this and bringing it to the end would be great :-)

ottojo:
@josephduchesne Please feel free to continue this, i don't currently have the time to continue this.

SteveMacenski:
Closing in favor of #4789 
Thanks @ottojo for getting this conversation started!

PR #4594, opened 2024-08-05
add use_namespace_for_frame_id

finani:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | mvsim simulation |

---

### Problems

* I wanted to simulate multiple agents in mvsim simulator with nav2.
* Each robot must have a different topic name and frame id.
* But even if I set use_namespace and namespace, each frame ids are the same.

### Requirements

* If I want to simulate multiple agents using nav2 on a single network such as simulator, I need to separate frame ids like topic names.
* There should be no side effects if new parameter is not set in yaml file.

### Solution

* I added a new parameter "use_namespace_for_frame_id".

### Result

* You can control whether the frame ids follow the topic names.

| use_namespace | namespace | use_namespace_for_frame_id | topic name | frame id |
| ---------------------- | ---------------- | ---------------------------------------- | --------------- | ----------- |
| false                    | ""                 | false (or empty line)                  | odom          | odom     |
| false                    | "r1"              | false (or empty line)                  | odom          | odom     |
| true                     | ""                 | false (or empty line)                  | odom          | odom     |
| true                     | "r1"              | false (or empty line)                  | r1/odom      | odom     |
| false                    | ""                 | true                                           | odom          | odom     |
| false                    | "r1"              | true                                           | odom          | odom     |
| true                     | ""                  | true                                           | odom          | odom     |
| **true**                | **"r1"**         | **true**                                     | **r1/odom** | **r1/odom** |

---

* I think there might be some optimizations to be made from STL vector
* I see alot of redundancy in this package, we might want to add a function `bool XYZ()` to reduce clutter
* I tested on a differential drive robot, but there might be issues turning near corners on an omnidirectional platform

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -350,6 +350,7 @@ protected:
    * @brief Get ROS parameters for node
    */
   void initParameters();
+  bool use_namespace_for_frame_id_;
   double alpha1_;
   double alpha2_;
   double alpha3_;
--- a/nav2_amcl/src/amcl_node.cpp
+++ b/nav2_amcl/src/amcl_node.cpp
@@ -61,6 +61,10 @@ AmclNode::AmclNode(const rclcpp::NodeOptions & options)
   RCLCPP_INFO(get_logger(), "Creating");
 
   add_parameter(
+    "use_namespace_for_frame_id", rclcpp::ParameterValue(false),
+    "Set this to True to use namespace for frame id");
+
+  add_parameter(
     "alpha1", rclcpp::ParameterValue(0.2),
     "This is the alpha1 parameter", "These are additional constraints for alpha1");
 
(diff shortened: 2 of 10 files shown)


SteveMacenski:
@finani we generally do this by remapping the TF streams to different topics for multirobot situations so that it more closely aligns with how real multi-robot distributed hardware would be setup: https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/launch/navigation_launch.py#L61

That makes the remapping of the frame names within the `/tf` topic not necessary since we have individual `r1/tf` and `r2/tf` topics that each listen to as part of its transformation tree and `TransformListener`s inside of each of the modules. 

I think that generally meets your needs, no?

finani:
You have a point.
I considered about a central monitoring system to gathering all tf topics.
However, I also think set frame id independently is not an appropriate solution based on "It more closely aligns with how real multi-robot distributed hardware would be setup".
Thank you for your comments.
I will close this PR.

SteveMacenski:
Thanks for the contribution! I look forward to others in the future as you run into bugs or features!

PR #4595, opened 2024-08-05
Remove in collision goals

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/issues/4587 |

---

* Added BT node that removes goals if the footprint cost is below a configurable cost threshold
* Fixed getCostCallback not handling out-of-bounds if use_footprint is true

---

#### For Maintainers: 


nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -0,0 +1,71 @@
...
+   * @brief Function to read parameters and initialize class variables
+   */
+  void initialize();
+
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<Goals>("input_goals", "Original goals to remove from"),
+      BT::InputPort<std::string>("costmap",
+          "Which costmap to use for checking collision. Choices: local, global, both"),
```
tonynajjar, line 49: Is this alright or would you have it specified differently?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -0,0 +1,71 @@
...
+      BT::InputPort<std::string>("cost_threshold",
+          "Cost threshold for considering a goal in collision"),
```
tonynajjar, line 51: I started out without this then thought it would make sense to have the cost_threshold configurable to remove waypoints that are too close to obstacles. Maybe I should change the name of the node though

nav2_rviz_plugins/src/costmap_cost_tool.cpp
```diff
@@ -89,6 +89,7 @@ void CostmapCostTool::callCostService(float x, float y)
   auto request = std::make_shared<nav2_msgs::srv::GetCost::Request>();
   request->x = x;
   request->y = y;
+  // request->use_footprint = true;
```
SteveMacenski, line 92: thanks, you can remove that line, it shouldn't be there

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+inline BT::NodeStatus RemoveInCollisionGoals::tick()
+{
+  setStatus(BT::NodeStatus::RUNNING);
+
+  if (!initialized_) {
+    initialize();
+  }
+
+  getInput("costmap", which_costmap_);
```
SteveMacenski, line 54: Do you think you'd change costmaps to use mid-BT execution for a given instance of this node? If not, we can getInput that in the `initialize()` method and only create the global or local costmap service instead of both

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,135 @@
...
+    return BT::NodeStatus::SUCCESS;
+  }
+
+  Goals valid_goal_poses;
+  for (const auto & goal : goal_poses) {
+    auto request = std::make_shared<nav2_msgs::srv::GetCost::Request>();
+    request->x = goal.pose.position.x;
+    request->y = goal.pose.position.y;
+    request->theta = tf2::getYaw(goal.pose.orientation);
+    request->use_footprint = true;
```
SteveMacenski, line 71: Might be worth an input  port option

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,135 @@
...
+  which_costmap_("both"),
+  cost_threshold_(253)
+{}
+
+void RemoveInCollisionGoals::initialize()
+{
+  node_ = config().blackboard->get<rclcpp::Node::SharedPtr>("node");
+  get_global_cost_client_ = node_->create_client<nav2_msgs::srv::GetCost>(
+    "global_costmap/get_cost_global_costmap");
+  get_local_cost_client_ = node_->create_client<nav2_msgs::srv::GetCost>(
```
SteveMacenski, line 41: We have a simple service client in Nav2 utils that you may want to use instead, it reduces alot of the boilerplate duplication and largely use it across nav2's codebase

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
SteveMacenski: I see there's a 'both' option - do you think you'd want / need that for a single given instance of this BT node?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
SteveMacenski: I agree to keep it, but we do need to consider that LETHAL is the 'max' for footprint checking and INFLATED is 'max' when doing center cost checking. We should also consider a logical default when not specified (and the type `std::string` is probably incorrect :-) )

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
SteveMacenski: Works for me, but maybe make it the actual `costmap_topic` to use instead - defaulted to global

SteveMacenski (changes requested):
Usual bits on tests and docs, but overall looks like a good initial prototype!

Thought: Adjust the `GetCost` service to a `GetCosts` service so that you don't call it repeatedly - populate 1 request with N goals and get their costs back in bulk. That's probably useful for a _number_ of reasons

nav2_rviz_plugins/src/costmap_cost_tool.cpp
tonynajjar: So I actually added this line to get the footprint cost using the rviz tool. I guess there is no easy way to specify from rviz whether to get footprint or point cost so I see 3 options:

1- Get and print both
2- Leave it commented out for users to manually comment out if needed
3- Have 2 different rviz buttons

I tend towards option 1. You?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
tonynajjar: you mean the costmap service name right? e.g. `/global_costmap/get_cost_global_costmap`

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: > I see there's a 'both' option - do you think you'd want / need that for a single given instance of this BT node?

I'm changing it so that it only checks on one service and in the BT we can chain them together if we want to check on both costmaps

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: > and largely use it across nav2's codebase

I guess you mean [this](https://github.com/ros-navigation/navigation2/blob/main/nav2_util/include/nav2_util/service_client.hpp). It makes me realize that something in my implementation (and using the service_client) is not ideal -> if the service takes too long it would block the BT ticking. 

This is why [this logic](https://github.com/ros-navigation/navigation2/blob/9e90417dd57b27c99c29b849321ad90205ca5523/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L205) exists in "pure" BT service nodes.

Do you think for BT nodes that call a service we should reimplement similar logic that would return RUNNING if the service call takes too long? I would think so

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
tonynajjar: > and INFLATED is 'max' when doing center cost checking

Can you explain this to me? Why isn't the max here also LETHAL?

nav2_rviz_plugins/src/costmap_cost_tool.cpp
SteveMacenski: > Get and print both

How do you specify orientation? I'm not sure that's overly useful without orientation for non-circular footprints. I think you'd need to specify an arrow like the initial pose or 2D goal tools

tonynajjar: ah right so we would need a separate tool for that. Ok forget it, out of scope of this PR

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -43,6 +43,7 @@ void RemovePassedGoals::initialize()
 
   robot_base_frame_ = BT::deconflictPortAndParamFrame<std::string>(
     node, "robot_base_frame", this);
+  initialized_ = true;
```
tonynajjar, line 46: not related to PR but it was missing

tonynajjar:
I graduated the PR from draft status. Docs and some manual testing are are missing, you can do a second round of review though

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,79 @@
...
+  getInput("cost_threshold", cost_threshold_);
+
+  getInput("input_goals", input_goals_);
+
+  if (input_goals_.empty()) {
+    setOutput("output_goals", input_goals_);
+    should_send_request_ = false;
+    return;
+  }
+  request_ = std::make_shared<nav2_msgs::srv::GetCosts::Request>();
```
tonynajjar, line 48: The `request_` is never recreated/reset in bt_service_node. It would make sense to add that there don't you think?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -0,0 +1,75 @@
...
+  void on_tick() override;
+
+  BT::NodeStatus on_completion(std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
+  override;
+
+  static BT::PortsList providedPorts()
+  {
+    return providedBasicPorts({
+        BT::InputPort<Goals>("input_goals", "Original goals to remove from"),
+        BT::InputPort<std::string>("costmap_cost_service",
```
SteveMacenski, line 57: Set the default to global costmap's?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -0,0 +1,75 @@
...
+          "Service to get cost from costmap"),
+        BT::InputPort<double>("cost_threshold",
```
SteveMacenski, line 59: Q: Should we set defaults for use footprint and cost threshold? If we do one, we should do both since they're related

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
+
+namespace nav2_behavior_tree
+{
+
+RemoveInCollisionGoals::RemoveInCollisionGoals(
+  const std::string & service_node_name,
+  const BT::NodeConfiguration & conf)
+: BtServiceNode<nav2_msgs::srv::GetCosts>(service_node_name, conf),
+  use_footprint_(true),
```
SteveMacenski, line 31: If setting here like this, then we should set these as BT port defaults as well. I think default `true` is reasonable

In this case, the cost threshold should be 254 for footprint checking, not 253 (inflated)

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
SteveMacenski: I think for this case since the goal vector length can change this is good!

SteveMacenski (changes requested):
Otherwise LGTM, just doc updates:
- Update migration guide that its GetCosts now
- Add the new BT node to the plugins page, migration guide, configuration guide

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
SteveMacenski: yeah!

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
SteveMacenski: If using the footprint, the edges of the footprint can be in the inflated space since that's the edge-cost of the footprint, not the center cost of a point check.

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: Sorry I wasn't clear enough, I meant to recreate the request in the BTServiceNode parent instead of for every child. Seems like a generic enough thing to do to me.

SteveMacenski: Oh, I think some rely on it being the same by populating it once in initialization -- but if that's not the case, happy to have that change too. Some API change could come and make that break.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -0,0 +1,72 @@
...
+   */
+  void on_tick() override;
+
+  BT::NodeStatus on_completion(std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
+  override;
+
+  static BT::PortsList providedPorts()
+  {
+    return providedBasicPorts({
+        BT::InputPort<Goals>("input_goals", "Original goals to remove from"),
```
SteveMacenski, line 56: Where did the costmap port go? Readd it and make it's default the global costmap

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
+
+namespace nav2_behavior_tree
+{
+
+RemoveInCollisionGoals::RemoveInCollisionGoals(
+  const std::string & service_node_name,
+  const BT::NodeConfiguration & conf)
+: BtServiceNode<nav2_msgs::srv::GetCosts>(service_node_name, conf,
+    "/global_costmap/get_cost_global_costmap")
```
SteveMacenski, line 31: why hard coded now?

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: As far as I understand that's the way to do it with BtServiceNode. This is not hardcoded, it can still be overridden, see [here](https://github.com/ros-navigation/navigation2/blob/746bb880a775201fbdc4cca38a873cc3d0d9ab40/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L72). I must admit I didn't test overriding it

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
tonynajjar: it's already [here](https://github.com/ros-navigation/navigation2/blob/746bb880a775201fbdc4cca38a873cc3d0d9ab40/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L115)

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
SteveMacenski: Oh, yes, that works!

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -0,0 +1,77 @@
...
+    request_->poses.push_back(pose);
+  }
+}
+
+BT::NodeStatus RemoveInCollisionGoals::on_completion(
+  std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
+{
+  Goals valid_goal_poses;
+  for (size_t i = 0; i < response->costs.size(); ++i) {
+    if (response->costs[i] <= cost_threshold_) {
```
SteveMacenski, line 63: I think less then, not equal to

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: definitely, thanks

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: It's only used [here](https://github.com/ros-navigation/navigation2/blob/746bb880a775201fbdc4cca38a873cc3d0d9ab40/nav2_behavior_tree/plugins/action/clear_costmap_service.cpp#L44) in Nav2 but some people might rely on it in their private plugins. I'm happy either way, the benefit of keeping it as is, is at least **to have the option** to initialize once.

SteveMacenski: I'm ambivalent unless someone had a compelling argument either way.

SteveMacenski (approved):

PR #4600, opened 2024-08-06
Switch nav2_rviz_plugins to modern CMake idioms.

clalancette:
In addition, make some fixes to the code to more appropriately hold on to the Node object (there were warnings from the compiler without these fixes).

---

## Basic Info

| ------ | ----------- |

---

Change nav2_rviz_plugins to use modern CMake idioms:
1.  Make it a "nav2_package".  This ended up adding in additional warnings, which then required some code changes to fix.  Thus, there are some additional code changes in here to more appropriately hold onto weak pointers.
2.  Switch from ament_target_dependencies to target_link_libraries.
3.  Export the target so downstream CMake consumers can use it.
4.  Push the include directory down one level, which is best practice since Humble.

None needed.

---

This is part of a larger series to switch Navigation2 to use modern CMake idioms.  There will be follow-up PRs converting more of the packages.

#### For Maintainers: 

### DIFF
--- a/nav2_rviz_plugins/src/nav2_panel.cpp
+++ b/nav2_rviz_plugins/src/nav2_panel.cpp
@@ -29,6 +29,7 @@
 
 #include "nav2_rviz_plugins/goal_common.hpp"
 #include "nav2_rviz_plugins/utils.hpp"
+#include "rclcpp/rclcpp.hpp"
 #include "rviz_common/display_context.hpp"
 #include "rviz_common/load_resource.hpp"
 #include "ament_index_cpp/get_package_share_directory.hpp"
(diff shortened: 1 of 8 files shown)


SteveMacenski (approved):

nav2_rviz_plugins/src/nav2_panel.cpp
```diff
@@ -777,7 +778,12 @@ void Nav2Panel::handleGoalSaver()
 void
 Nav2Panel::onInitialize()
 {
-  auto node = getDisplayContext()->getRosNodeAbstraction().lock()->get_raw_node();
+  node_ptr_ = getDisplayContext()->getRosNodeAbstraction().lock();
+  if (node_ptr_ == nullptr) {
+    // The node no longer exists, so just don't initialize
```
SteveMacenski, line 783: For all: Some kind of log, even if `cout`, that this is the case would be nice (for all instances)

clalancette: Good call.  Fixed in 864a450.

SteveMacenski (approved):

PR #4602, opened 2024-08-07
Fix behavior server

VladyslavHrynchak200204:
---

## Basic Info

| ------ | ----------- |

---

**Purpose:**

This PR fix a problem with TimedBehavior. It occurs when we run first DriveOnHeading -> cancel first DriveOnHeading -> immediately run the second  DriveOnHeading. This way we accept new goal -> requested preemption -> terminate_all goals because of cancel first DriveOnHeading -> current goal from second DriveOnHeading is empty -> second  DriveOnHeading fails.

image

**Changes:**

We swapped two checks in TimedBehavior. First we check if we have preemption requested and then check if we request to cancel. In this case our current goal won’t be empty and DriveOnHeading won’t fail.

**Before:**

image

We don`t see log "Running drive_on_heading"  from the second DriveOnHeading because it fails

**After:**

image

#### For Maintainers: 

### DIFF
--- a/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
@@ -237,15 +237,6 @@ protected:
 
     while (rclcpp::ok()) {
       elasped_time_ = clock_->now() - start_time;
-      if (action_server_->is_cancel_requested()) {
-        RCLCPP_INFO(logger_, "Canceling %s", behavior_name_.c_str());
-        stopRobot();
-        result->total_elapsed_time = elasped_time_;
-        onActionCompletion(result);
-        action_server_->terminate_all(result);
-        return;
-      }
-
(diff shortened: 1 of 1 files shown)


VladyslavHrynchak200204:
Hello @SteveMacenski ! Could you check this PR?

nav2_behaviors/include/nav2_behaviors/timed_behavior.hpp
```diff
@@ -259,6 +250,15 @@ class TimedBehavior : public nav2_core::Behavior
         return;
       }
 
+      if (action_server_->is_cancel_requested()) {
```
SteveMacenski, line 253: Can you check the other Nav2 action servers and see if any others should be flipped as well?

VladyslavHrynchak200204: Everything else looks fine.

SteveMacenski (approved):

PR #4611, opened 2024-08-09
Switch dwb_critics to modern CMake idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |

---

Update dwb_critics to use modern CMake idioms:
1.  Switch from ament_target_dependencies to target_link_libraries.
2.  Export a CMake target so downstream packages can use the libraries here.
3.  Push the include directories down one level, which is best practice since Humble.

None needed.

---

This is part of a larger series updating Navigation2 to modern CMake idioms.  There are approximately 6 packages left after this one.  However, note that I will be taking a break from this work for one or two weeks, as I am on vacation, and still need to do the conversion for 4 more of the packages.

#### For Maintainers: 


SteveMacenski (approved):
Looks good to me, but rerunning CI. It failed on a test I've never seen fail before

Issue #4616, opened 2024-08-12
loopback_sim: provide realistic scan data


SteveMacenski:
https://github.com/nobleo/mobile_robot_simulator/blob/master/src/laser_simulator.cpp takes in the map and creates a laser scan based on it and the robot's pose. We should look to do that in `nav2_loopback_sim` rather than max range

PR #4617, opened 2024-08-13
Added laser data from map in nav2_loopback_sim

JatinPatil2003:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4616 |
| Primary OS tested on | Humble |

---

* Publishes realistic laser data according to the map

---

#### For Maintainers: 

### DIFF
--- a/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
+++ b/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
@@ -18,6 +18,9 @@ import math
 from geometry_msgs.msg import PoseWithCovarianceStamped, Twist
 from geometry_msgs.msg import Quaternion, TransformStamped, Vector3
 from nav_msgs.msg import Odometry
+from nav_msgs.srv import GetMap
+from tf2_ros import TransformListener, Buffer
+import tf_transformations
 import rclpy
 from rclpy.duration import Duration
 from rclpy.node import Node
@@ -26,8 +29,14 @@ from sensor_msgs.msg import LaserScan
 from tf2_ros import TransformBroadcaster
 import tf_transformations
 
-from .utils import addYawToQuat, matrixToTransform, transformStampedToMatrix
-
+from .utils import (
+    addYawToQuat, 
+    matrixToTransform, 
+    transformStampedToMatrix, 
+    get_world2map_coordinates, 
+    get_map2world_coordinates, 
+    get_map_occupancy
+)
 
 """
 This is a loopback simulator that replaces a physics simulator to create a
@@ -48,6 +57,7 @@ class LoopbackSimulator(Node):
         self.initial_pose = None
         self.timer = None
         self.setupTimer = None
+        self.map = None
 
         self.declare_parameter('update_duration', 0.01)
         self.update_dur = self.get_parameter('update_duration').get_parameter_value().double_value
@@ -88,6 +98,14 @@ class LoopbackSimulator(Node):
         self.scan_pub = self.create_publisher(LaserScan, 'scan', sensor_qos)
 
         self.setupTimer = self.create_timer(0.1, self.setupTimerCallback)
+        
+        self.map_client = self.create_client(GetMap, '/map_server/map')
+
+        self.tf_buffer = Buffer()
+        self.tf_listener = TransformListener(self.tf_buffer, self)
+
+        self.get_map()
+
         self.info('Loopback simulator initialized')
 
     def setupTimerCallback(self):
@@ -122,6 +140,7 @@ class LoopbackSimulator(Node):
                 self.setupTimer.destroy()
                 self.setupTimer = None
             self.timer = self.create_timer(self.update_dur, self.timerCallback)
+            self.timer_laser = self.create_timer(0.1, self.publishLaserScan)
             return
 
         self.initial_pose = msg.pose.pose
@@ -165,23 +184,24 @@ class LoopbackSimulator(Node):
 
         self.publishTransforms(self.t_map_to_odom, self.t_odom_to_base_link)
         self.publishOdometry(self.t_odom_to_base_link)
-        self.publishLaserScan()
 
     def publishLaserScan(self):
         # Publish a bogus laser scan for collision monitor
-        scan = LaserScan()
-        # scan.header.stamp = (self.get_clock().now()).to_msg()
-        scan.header.frame_id = self.scan_frame_id
-        scan.angle_min = -math.pi
-        scan.angle_max = math.pi
-        scan.angle_increment = 0.005817705996  # 0.333 degrees
-        scan.time_increment = 0.0
-        scan.scan_time = 0.1
-        scan.range_min = 0.1
-        scan.range_max = 100.0
-        num_samples = int((scan.angle_max - scan.angle_min) / scan.angle_increment)
-        scan.ranges = [scan.range_max - 0.01] * num_samples
-        self.scan_pub.publish(scan)
+        self.scan_msg = LaserScan()
+        self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
+        self.scan_msg.header.frame_id = self.scan_frame_id
+        self.scan_msg.angle_min = -math.pi
+        self.scan_msg.angle_max = math.pi
+        self.scan_msg.angle_increment = 0.0261799 # 1.5 degrees
+        self.scan_msg.time_increment = 0.0
+        self.scan_msg.scan_time = 0.1
+        self.scan_msg.range_min = 0.05
+        self.scan_msg.range_max = 100.0
+        num_samples = int((self.scan_msg.angle_max - self.scan_msg.angle_min) / self.scan_msg.angle_increment)
+        self.scan_msg.ranges = [0.0] * num_samples
+        x, y, theta = self.get_laser_pose()
+        self.get_laserscan(num_samples, x, y, theta)
+        self.scan_pub.publish(self.scan_msg)
 
     def publishTransforms(self, map_to_odom, odom_to_base_link):
         map_to_odom.header.stamp = \
@@ -208,7 +228,138 @@ class LoopbackSimulator(Node):
     def debug(self, msg):
         self.get_logger().debug(msg)
         return
+    
+    def get_map(self):
+        request = GetMap.Request()
+        if self.map_client.wait_for_service(timeout_sec=1.0):
+            # Request to get map
+            future = self.map_client.call_async(request)
+            rclpy.spin_until_future_complete(self, future)
+            if future.result() is not None:
+                self.map = future.result().map
+                self.get_logger().info(f'Received map data, {self.map.info.width}X{self.map.info.height}, resolution:={self.map.info.resolution}')
+            else:
+                self.get_logger().error('Failed to call map service')
+        else:
+            self.get_logger().error('Map service not available')
 
+    def get_laser_pose(self):
+        try:
+            if self.initial_pose is None:
+                return 0.0, 0.0, 0.0
+            
+            if self.map is None:
+                self.get_map()
+                return 0.0, 0.0, 0.0
+            
+            # Wait for transform and lookup
+            now = rclpy.time.Time()
+            transform = self.tf_buffer.lookup_transform(
+                'map', self.scan_frame_id, now, rclpy.duration.Duration(seconds=1.0))
+
+            # Extract pose information
+            x = transform.transform.translation.x
+            y = transform.transform.translation.y
+            rotation = transform.transform.rotation
+            theta = tf_transformations.euler_from_quaternion([
+                rotation.x,
+                rotation.y,
+                rotation.z,
+                rotation.w
+            ])[2]
+
+            return x, y, theta
+
+        except Exception as ex:
+            self.get_logger().error('Transform lookup failed: %s' % str(ex))
+            return 0.0, 0.0, 0.0
+
+    def get_laserscan(self, num_samples, x, y, theta):
+        for i in range(int(num_samples)):
+            if self.map is None:
+                self.scan_msg.ranges = [self.scan_msg.range_max] * num_samples
+                break
+            else:
+                curr_angle = theta + self.scan_msg.angle_min + i * self.scan_msg.angle_increment
+                self.scan_msg.ranges[i] = self.find_map_range(x, y, curr_angle)
+
+    def find_map_range(self, x, y, theta):
+        # Using "A Faster Voxel Traversal Algorithm for Ray Tracing" by Amanatides & Woo
+        # ======== Initialization Phase ========
+        origin = [x, y]  # u
+        dir = [math.cos(theta), math.sin(theta)]  # v
+        
+        start_x, start_y = get_world2map_coordinates(x, y, self.map)
+        
+        if start_x < 0 or start_y < 0 or start_x >= self.map.info.width or start_y >= self.map.info.height:
+            # Outside the map, find entry point
+            delta_x = abs(self.map.info.origin.position.x - x)
+            delta_y = abs(self.map.info.origin.position.y - y)
+            intersect_x = x + (dir[0] * delta_x)
+            intersect_y = y + (dir[1] * delta_y)
+            start_x, start_y = get_world2map_coordinates(intersect_x, intersect_y, self.map)
+        
+        current = [start_x, start_y]  # X, Y
+        step = [0, 0]  # stepX, stepY
+        tMax = [0.0, 0.0]  # tMaxX, tMaxY
+        tDelta = [0.0, 0.0]  # tDeltaX, tDeltaY
+        
+        voxel_border = [0.0, 0.0]
+        voxel_border[0], voxel_border[1] = get_map2world_coordinates(current[0], current[1], self.map)
+        voxel_border[0] -= 0.5 * self.map.info.resolution  # This is the lower left voxel corner
+        voxel_border[1] -= 0.5 * self.map.info.resolution  # This is the lower left voxel corner
+        
+        for i in range(2):  # For each dimension (x, y)
+            # Determine step direction
+            if dir[i] > 0.0:
+                step[i] = 1
+            elif dir[i] < 0.0:
+                step[i] = -1
+            else:
+                step[i] = 0
+            
+            # Determine tMax, tDelta
+            if step[i] != 0:
+                if step[i] == 1:
+                    voxel_border[i] += step[i] * self.map.info.resolution
+                
+                # tMax - voxel boundary crossing
+                tMax[i] = (voxel_border[i] - origin[i]) / dir[i]
+                # tDelta - distance along ray so that vertical/horizontal component equals one voxel
+                tDelta[i] = self.map.info.resolution / abs(dir[i])
+            else:
+                tMax[i] = float('inf')
+                tDelta[i] = float('inf')
+        
+        # ======== Incremental Traversal ========
+        while True:
+            # Advance
+            dim = 0 if tMax[0] < tMax[1] else 1
+            
+            # Advance one voxel
+            current[dim] += step[dim]
+            tMax[dim] += tDelta[dim]
+            
+            # Check map inbounds
+            if current[0] < 0 or current[0] >= self.map.info.width or current[1] < 0 or current[1] >= self.map.info.height:
+                # self.get_logger().info("Here-1")
+                return self.scan_msg.range_max
+            
+            # Determine current range
+            current_range = math.sqrt((current[0] - start_x) ** 2 + (current[1] - start_y) ** 2) * self.map.info.resolution
+            
+            # Are we at max range?
+            if current_range > self.scan_msg.range_max:
+                # self.get_logger().info("Here-1")
+                return self.scan_msg.range_max
+            else:
+                occ = get_map_occupancy(current[0], current[1], self.map)
+                if occ >= 60:  # Current cell is occupied
+                    # Is range less than min range
+                    if current_range < self.scan_msg.range_min:
+                        continue
+
+                    return current_range
 
 def main():
     rclpy.init()
--- a/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
+++ b/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
@@ -206,7 +206,7 @@ class LoopbackSimulator(Node):
     def publishLaserScan(self):
         # Publish a bogus laser scan for collision monitor
         self.scan_msg = LaserScan()
-        # self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
+        self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
         self.scan_msg.header.frame_id = self.scan_frame_id
         self.scan_msg.angle_min = -math.pi
         self.scan_msg.angle_max = math.pi
--- a/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
+++ b/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
@@ -240,11 +240,15 @@ class LoopbackSimulator(Node):
                 self.map = future.result().map
                 self.get_logger().info('Laser scan will be populated using map data')
             else:
-                self.get_logger().warn('Failed to get map, '
-                    'Laser scan will be populated using max range')
+                self.get_logger().warn(
+                    'Failed to get map, '
+                    'Laser scan will be populated using max range'
+                )
         else:
-            self.get_logger().warn('Failed to get map, '
-                'Laser scan will be populated using max range')
+            self.get_logger().warn(
+                'Failed to get map, '
+                'Laser scan will be populated using max range'
+            )
 
     def getLaserPose(self):
         if self.initial_pose is None:
--- a/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
+++ b/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
@@ -72,6 +72,12 @@ class LoopbackSimulator(Node):
         self.declare_parameter('scan_frame_id', 'base_scan')
         self.scan_frame_id = self.get_parameter('scan_frame_id').get_parameter_value().string_value
 
+        self.declare_parameter('publish_map_odom_tf', False)
+        self.publish_map_odom_tf = self.get_parameter('publish_map_odom_tf').get_parameter_value().bool_value
+
+        self.declare_parameter('scan_publish_rate', 10.0)
+        self.scan_publish_rate = self.get_parameter('scan_publish_rate').get_parameter_value().double_value
+
         self.t_map_to_odom = TransformStamped()
         self.t_map_to_odom.header.frame_id = self.map_frame_id
         self.t_map_to_odom.child_frame_id = self.odom_frame_id
@@ -138,7 +144,7 @@ class LoopbackSimulator(Node):
                 self.setupTimer.destroy()
                 self.setupTimer = None
             self.timer = self.create_timer(self.update_dur, self.timerCallback)
-            self.timer_laser = self.create_timer(0.1, self.publishLaserScan)
+            self.timer_laser = self.create_timer((1/self.scan_publish_rate), self.publishLaserScan)
             return
 
         self.initial_pose = msg.pose.pose
@@ -186,7 +192,7 @@ class LoopbackSimulator(Node):
     def publishLaserScan(self):
         # Publish a bogus laser scan for collision monitor
         self.scan_msg = LaserScan()
-        # self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
+        self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
         self.scan_msg.header.frame_id = self.scan_frame_id
         self.scan_msg.angle_min = -math.pi
         self.scan_msg.angle_max = math.pi
@@ -208,7 +214,8 @@ class LoopbackSimulator(Node):
         map_to_odom.header.stamp = \
             (self.get_clock().now() + Duration(seconds=self.update_dur)).to_msg()
         odom_to_base_link.header.stamp = self.get_clock().now().to_msg()
-        self.tf_broadcaster.sendTransform(map_to_odom)
+        if self.publish_map_odom_tf:
+            self.tf_broadcaster.sendTransform(map_to_odom)
         self.tf_broadcaster.sendTransform(odom_to_base_link)
 
     def publishOdometry(self, odom_to_base_link):
@@ -251,9 +258,6 @@ class LoopbackSimulator(Node):
             )
 
     def getLaserPose(self):
-        if self.initial_pose is None:
-            return 0.0, 0.0, 0.0
-
         try:
             # Wait for transform and lookup
             now = rclpy.time.Time()
--- a/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
+++ b/nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
@@ -73,10 +73,12 @@ class LoopbackSimulator(Node):
         self.scan_frame_id = self.get_parameter('scan_frame_id').get_parameter_value().string_value
 
         self.declare_parameter('publish_map_odom_tf', False)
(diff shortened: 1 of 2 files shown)


nav2_loopback_sim/nav2_loopback_sim/utils.py
```diff
@@ -63,3 +65,21 @@ def matrixToTransform(matrix):
     transform.rotation.z = quaternion[2]
     transform.rotation.w = quaternion[3]
     return transform
+
+
+def get_world2map_coordinates(world_x, world_y, map_msg):
```
SteveMacenski, line 70: Use the bound checks in https://github.com/ros-navigation/navigation2/blob/3f311124e356dbfa33f35e23caa5474a57f99ab7/nav2_costmap_2d/src/costmap_2d.cpp#L284-L297

nav2_loopback_sim/nav2_loopback_sim/utils.py
```diff
@@ -63,3 +65,21 @@ def matrixToTransform(matrix):
...
+    map_x = int(math.floor((world_x - map_msg.info.origin.position.x) / map_msg.info.resolution))
+    map_y = int(math.floor((world_y - map_msg.info.origin.position.y) / map_msg.info.resolution))
+    return map_x, map_y
+
+
+def get_map2world_coordinates(map_x, map_y, map_msg):
```
SteveMacenski, line 76: This appears to me to incorrect? https://github.com/ros-navigation/navigation2/blob/3f311124e356dbfa33f35e23caa5474a57f99ab7/nav2_costmap_2d/src/costmap_2d.cpp#L278-L282

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -18,16 +18,23 @@
...
 from sensor_msgs.msg import LaserScan
-from tf2_ros import TransformBroadcaster
+from tf2_ros import Buffer, TransformBroadcaster, TransformListener
 import tf_transformations
 
-from .utils import addYawToQuat, matrixToTransform, transformStampedToMatrix
-
+from .utils import (
+    addYawToQuat,
+    get_map2world_coordinates,
```
SteveMacenski, line 32: Rename them to be inline with the camelCase naming convention please :-)

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -88,6 +96,14 @@ def __init__(self):
         self.scan_pub = self.create_publisher(LaserScan, 'scan', sensor_qos)
 
         self.setupTimer = self.create_timer(0.1, self.setupTimerCallback)
+
+        self.map_client = self.create_client(GetMap, '/map_server/map')
+
+        self.tf_buffer = Buffer()
+        self.tf_listener = TransformListener(self.tf_buffer, self)
+
+        self.get_map()
```
SteveMacenski, line 105: For all: use camelcase for naming things `getMap()`

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,23 +182,27 @@ def timerCallback(self):
...
-        scan.angle_increment = 0.005817705996  # 0.333 degrees
-        scan.time_increment = 0.0
-        scan.scan_time = 0.1
-        scan.range_min = 0.1
-        scan.range_max = 100.0
-        num_samples = int((scan.angle_max - scan.angle_min) / scan.angle_increment)
-        scan.ranges = [scan.range_max - 0.01] * num_samples
-        self.scan_pub.publish(scan)
+        self.scan_msg = LaserScan()
+        self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```
SteveMacenski, line 189:
```suggestion
        # self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+                return 0.0, 0.0, 0.0
+
+            if self.map is None:
+                self.get_map()
+                return 0.0, 0.0, 0.0
+
+            # Wait for transform and lookup
+            now = rclpy.time.Time()
+            transform = self.tf_buffer.lookup_transform(
+                'map', self.scan_frame_id, now, rclpy.duration.Duration(seconds=1.0))
```
SteveMacenski, line 260:
```suggestion
                'map', self.scan_frame_id, now, rclpy.duration.Duration(seconds=0.1))
```

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+    def find_map_range(self, x, y, theta):
+        # Using "A Faster Voxel Traversal Algorithm for Ray Tracing" by Amanatides & Woo
+        # ======== Initialization Phase ========
+        origin = [x, y]  # u
+        direction = [math.cos(theta), math.sin(theta)]  # v
+
+        start_x, start_y = get_world2map_coordinates(x, y, self.map)
+
+        if (start_x < 0 or start_y < 0 or
+                start_x >= self.map.info.width or start_y >= self.map.info.height):
```
SteveMacenski, line 297: I think if the robot is off the map, we can just set it to max value -- but I'm not totally grokking this. can you explain?

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
 
+    def get_map(self):
+        request = GetMap.Request()
+        if self.map_client.wait_for_service(timeout_sec=1.0):
+            # Request to get map
+            future = self.map_client.call_async(request)
+            rclpy.spin_until_future_complete(self, future)
+            if future.result() is not None:
+                self.map = future.result().map
+                self.get_logger().info(
```
SteveMacenski, line 241: I think this log should mention that the laser scan will be populated

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+                    f'Received map data, {self.map.info.width}X{self.map.info.height}')
+            else:
+                self.get_logger().error('Failed to call map service')
```
SteveMacenski, line 244: These should fail by saying that the max range will be used instead

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
         self.get_logger().debug(msg)
         return
 
+    def get_map(self):
+        request = GetMap.Request()
+        if self.map_client.wait_for_service(timeout_sec=1.0):
```
SteveMacenski, line 235:
```suggestion
        if self.map_client.wait_for_service(timeout_sec=5.0):
```

if we're only trying once, we should try hard

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+        else:
+            self.get_logger().error('Map service not available')
+
+    def get_laser_pose(self):
+        try:
+            if self.initial_pose is None:
+                return 0.0, 0.0, 0.0
+
+            if self.map is None:
+                self.get_map()
```
SteveMacenski, line 254: Remove this, I'd try once on startup and after that just use max. This adds alot of latency

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+
+            return x, y, theta
+
+        except Exception as ex:
+            self.get_logger().error('Transform lookup failed: %s' % str(ex))
+            return 0.0, 0.0, 0.0
+
+    def get_laserscan(self, num_samples, x, y, theta):
+        for i in range(int(num_samples)):
+            if self.map is None:
```
SteveMacenski, line 281: or initial pose not set

SteveMacenski, line 281: This `if` shouldn't be in the loop. It should be checked at the start (with the initial pose addition) and then returned immediately with the max range minus 0.1

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+
+        current = [start_x, start_y]  # X, Y
+        step = [0, 0]  # stepX, stepY
+        tMax = [0.0, 0.0]  # tMaxX, tMaxY
+        tDelta = [0.0, 0.0]  # tDeltaX, tDeltaY
+
+        voxel_border = [0.0, 0.0]
+        voxel_border[0], voxel_border[1] = get_map2world_coordinates(
+            current[0], current[1], self.map)
+        voxel_border[0] -= 0.5 * self.map.info.resolution  # This is the lower left voxel corner
```
SteveMacenski, line 313: I'm not really grokking the voxel_border stuff from L313 to L337. Can you explain? I think the worldToMap and mapToWorld conversions should be handling the `0.5`'s fine itself. I don't get the other stuff

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,143 @@ def debug(self, msg):
...
+    def get_laserscan(self, num_samples, x, y, theta):
+        for i in range(int(num_samples)):
+            if self.map is None:
+                self.scan_msg.ranges = [self.scan_msg.range_max] * num_samples
+                break
+            else:
+                curr_angle = theta + self.scan_msg.angle_min + i * self.scan_msg.angle_increment
+                self.scan_msg.ranges[i] = self.find_map_range(x, y, curr_angle)
+
+    def find_map_range(self, x, y, theta):
```
SteveMacenski, line 288: This function has some analogs to https://github.com/ros-navigation/navigation2/blob/main/nav2_simple_commander/nav2_simple_commander/line_iterator.py. Perhaps this can be used instead to reduce code duplication and potential errors for ray marching? Example use here: https://github.com/ros-navigation/navigation2/blob/main/nav2_simple_commander/nav2_simple_commander/footprint_collision_checker.py#L113-L126 you could march from start -> max range and keep going until you hit an occupied cell, max range, or leave the map

SteveMacenski (changes requested):
Thanks for taking this on! Largely looks great but a few bugs / nits that are good to have resolved that I think makes it even better :-)

nav2_loopback_sim/nav2_loopback_sim/utils.py
SteveMacenski: We have reference implementations: https://github.com/ros-navigation/navigation2/blob/main/nav2_simple_commander/nav2_simple_commander/costmap_2d.py#L143-L186

nav2_loopback_sim/nav2_loopback_sim/utils.py
JatinPatil2003: Updated

nav2_loopback_sim/nav2_loopback_sim/utils.py
JatinPatil2003: Both were same but I have changed it to be same as reference

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
JatinPatil2003: As `worldToMap` is updated to check bounds, I have removed this condition

Originally it was implemented [here](https://github.com/nobleo/mobile_robot_simulator/blob/master/src/laser_simulator.cpp) which will keep populating laser scan data even when the robot gets out of the map

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
JatinPatil2003: I have tried to keep it same as original [reference](https://github.com/nobleo/mobile_robot_simulator/blob/master/src/laser_simulator.cpp)

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
JatinPatil2003: okay i'll change logic using `LineIterator` to find laser ranges

JatinPatil2003:
Converting PR to draft for implementation of `LineIterator`

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,133 @@ def debug(self, msg):
...
 
+    def getMap(self):
+        request = GetMap.Request()
+        if self.map_client.wait_for_service(timeout_sec=5.0):
+            # Request to get map
+            future = self.map_client.call_async(request)
+            rclpy.spin_until_future_complete(self, future)
+            if future.result() is not None:
+                self.map = future.result().map
+                self.get_logger().info('Laser scan will be populated')
```
SteveMacenski, line 241:
```suggestion
                self.get_logger().info('Laser scan will be populated using map data')
```

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -88,8 +96,16 @@ def __init__(self):
...
         self.setupTimer = self.create_timer(0.1, self.setupTimerCallback)
+
+        self.map_client = self.create_client(GetMap, '/map_server/map')
+
+        self.tf_buffer = Buffer()
+        self.tf_listener = TransformListener(self.tf_buffer, self)
+
         self.info('Loopback simulator initialized')
 
+        self.getMap()
```
SteveMacenski, line 107: Move before the initialized log

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,133 @@ def debug(self, msg):
...
+        request = GetMap.Request()
+        if self.map_client.wait_for_service(timeout_sec=5.0):
+            # Request to get map
+            future = self.map_client.call_async(request)
+            rclpy.spin_until_future_complete(self, future)
+            if future.result() is not None:
+                self.map = future.result().map
+                self.get_logger().info('Laser scan will be populated')
+            else:
+                self.get_logger().warn('Failed to get map')
```
SteveMacenski, line 243: All part of 1 log (ditto below)

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,133 @@ def debug(self, msg):
...
+                self.get_logger().info('Laser scan will be populated using max range')
+        else:
+            self.get_logger().warn('Failed to get map')
+            self.get_logger().info('Laser scan will be populated using max range')
+
+    def getLaserPose(self):
+        try:
+            if self.initial_pose is None:
+                return 0.0, 0.0, 0.0
```
SteveMacenski, line 252: `None`s? (as well in the exception handling) That would make it the same as the map to world conversion methods error handling

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +230,133 @@ def debug(self, msg):
...
+
+        mx, my = worldToMap(x, y, self.map)
+
+        if not mx and not my:
+            self.scan_msg.ranges = [self.scan_msg.range_max - 0.01] * num_samples
+            return
+
+        for i in range(int(num_samples)):
+            curr_angle = theta + self.scan_msg.angle_min + i * self.scan_msg.angle_increment
+            self.scan_msg.ranges[i] = self.findMapRange(mx, my, x, y, curr_angle)
```
SteveMacenski, line 289: Should this be `ranges.append()` instead?

SteveMacenski:
Thanks! Looks so much better already! :-)

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
JatinPatil2003: both can be done actually I have initialized `self.scan_msg.ranges = [0.0] * num_samples` already and in earlier commits it was not returning anything when the scan range is less than min range do by default it was `0.0`

but later I have changed it to return `0.0`, So now `.append()` will also work

but let me know should I change it to `.append()`

JatinPatil2003:
> Thanks! Looks so much better already! :-)

@SteveMacenski Should I change PR status then?

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
SteveMacenski: Oh ok! That works, I missed that

SteveMacenski:
I assume you tested this and see a reasonable laser scan that Nav2 can use? Just checking before I do my final pass through that 'this has shown to work' :-)

JatinPatil2003:
https://github.com/user-attachments/assets/99a60754-9fe5-469e-bfdf-044b5f6246b2

This is working of nav2_loopback_sim, laser scan with nav2

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -122,6 +138,7 @@ def initialPoseCallback(self, msg):
                 self.setupTimer.destroy()
                 self.setupTimer = None
             self.timer = self.create_timer(self.update_dur, self.timerCallback)
+            self.timer_laser = self.create_timer(0.1, self.publishLaserScan)
```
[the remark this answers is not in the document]
JatinPatil2003: I did try publishing laser scan on same timer but as ray tracing is more instruction heavy it is taking more time to populate laser scan and due to that odometry and tf performance is getting affected 

maybe if 2 timers are not optimal solution then making a separate node to publish laser scan makes sense

JatinPatil2003:
> I generally approve except for the `findMapRange` stuff which should use the line iterator
> 
> > This is working of nav2_loopback_sim, laser scan with nav2
> 
> Generally looks good! I think something is possibly going on (?) since the laser scan appears to be flickering between 2 solutions, even when the robot is still. These should perfectly line up since there's no source of error. If there's no error/noise and the robot's TF doesn't move, this should be completely deterministic. I'm wondering if that's from something in the ray tracing that the line iterator should resolve or the call to TF. You use TF lookups to get the map->laser frame, but we do have map->odom and odom->base_link transforms in the simulator exactly. I suppose if you looked up at startup base_link->laser frame and cached that, then it should be 100% aligned (if that's the error cause)

Yea it's not due to ray tracing, while testing with nav it tested with nav2_bringup so amcl is also publishing another tf

I think a parameter for publishing map to odom will be the solution for this

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,23 +182,27 @@ def timerCallback(self):
...
-        scan.angle_increment = 0.005817705996  # 0.333 degrees
-        scan.time_increment = 0.0
-        scan.scan_time = 0.1
-        scan.range_min = 0.1
-        scan.range_max = 100.0
-        num_samples = int((scan.angle_max - scan.angle_min) / scan.angle_increment)
-        scan.ranges = [scan.range_max - 0.01] * num_samples
-        self.scan_pub.publish(scan)
+        self.scan_msg = LaserScan()
+        # self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```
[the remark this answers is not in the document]
JatinPatil2003: As I'm using different timers for scan publishing stamp will be needed

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -64,6 +72,14 @@ def __init__(self):
         self.declare_parameter('scan_frame_id', 'base_scan')
         self.scan_frame_id = self.get_parameter('scan_frame_id').get_parameter_value().string_value
 
+        self.declare_parameter('publish_map_odom_tf', False)
```
SteveMacenski, line 75: I don't understand this parameter. Why? 

At the very least, it should be default to true

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -64,6 +72,14 @@ def __init__(self):
...
+        self.publish_map_odom_tf = self.get_parameter(
+            'publish_map_odom_tf').get_parameter_value().bool_value
+
+        self.declare_parameter('scan_publish_rate', 10.0)
```
SteveMacenski, line 79: Make it the duration like we have for the other update parameter

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -122,6 +146,7 @@ def initialPoseCallback(self, msg):
                 self.setupTimer.destroy()
                 self.setupTimer = None
             self.timer = self.create_timer(self.update_dur, self.timerCallback)
+            self.timer_laser = self.create_timer((1/self.scan_publish_rate), self.publishLaserScan)
```
SteveMacenski, line 149: Then don't need to invert

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
SteveMacenski: Please do not resolve conversations that you did not complete.

SteveMacenski (changes requested):
Did you try the line iterator method? Perhaps you didn't push this to `origin` so I could see

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
SteveMacenski: > I think a parameter for publishing map to odom will be the solution for this

I don't understand the take away. I'm saying that we shouldn't use TF buffer lookups. We have `self.t_map_to_odom` and `self.t_odom_to_base_link`. If trying to find the laser pose, we only need to find on initialization the `self.t_base_link_to_laser_scanner` and we have all the exact transforms needed to find the laser scanner link in the map frame without involving TF

JatinPatil2003: In latest commit I used the cached tf but still the same problem of flickering is there, when this parameter is set to False then it work well with nav2(no flickering)

**Both tested with LineIterator and with using cached tf**

https://github.com/user-attachments/assets/84a90c31-5bb2-49b2-8487-f619b665864b


https://github.com/user-attachments/assets/1f2aa291-0389-46ff-a17e-6c70b325b63e

JatinPatil2003:
I think the current ray tracking algorithm is working well as compared to LineIterator
This case is only when robot is near obstacle but works well when not too close

**This image is using LineIterator**
image

**This Image is using current ray tracking algorithm**
image

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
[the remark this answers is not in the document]
SteveMacenski: OK - we can leave as is

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
SteveMacenski: Is that happening before your changes? You're also showign a particle cloud, so I don't know what you mean by true/false because they're both using AMCL. This replaces AMCL when set to true -- maybe that's why you have flickering. You're running both simultaneously.

Try using the provided loopback launch files I provide in `nav2_bringup` - which does not have AMCL.

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -115,13 +156,15 @@ def initialPoseCallback(self, msg):
             self.t_odom_to_base_link.transform.translation = Vector3()
             self.t_odom_to_base_link.transform.rotation = Quaternion()
             self.publishTransforms(self.t_map_to_odom, self.t_odom_to_base_link)
+            self.mat_map_to_odom = transformStampedToMatrix(self.t_map_to_odom)
```
SteveMacenski, line 159: These are non-static, do not store, get when needed. We don't want to have duplicate information we're trying to keep consistent.

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,29 +208,35 @@ def timerCallback(self):
 
         self.publishTransforms(self.t_map_to_odom, self.t_odom_to_base_link)
         self.publishOdometry(self.t_odom_to_base_link)
-        self.publishLaserScan()
+        self.mat_map_to_odom = transformStampedToMatrix(self.t_map_to_odom)
```
SteveMacenski, line 211: These are non-static, do not store, get when needed. We don't want to have duplicate information we're trying to keep consistent.

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,29 +208,35 @@ def timerCallback(self):
...
+        self.mat_odom_to_base = transformStampedToMatrix(self.t_odom_to_base_link)
```
SteveMacenski, line 212: These are non-static, do not store, get when needed. We don't want to have duplicate information we're trying to keep consistent.

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +258,164 @@ def debug(self, msg):
...
+                )
+        else:
+            self.get_logger().warn(
+                'Failed to get map, '
+                'Laser scan will be populated using max range'
+            )
+
+    def getLaserPose(self):
+        mat_map_to_laser = tf_transformations.concatenate_matrices(
+            self.mat_map_to_odom,
```
SteveMacenski, line 283: Transform here when needed

SteveMacenski (changes requested):

SteveMacenski:
> This image is using LineIterator

I like this in concept. I'm a little unclear why they seem to "stack up" to so much thickness of the wall -- shouldn't it have hit one of the previous cells before being basically on the other side of the wall's thickness? We do this in the [collision footprint checker](https://github.com/ros-navigation/navigation2/blob/main/nav2_simple_commander/nav2_simple_commander/footprint_collision_checker.py#L113) in cell coordinates with 0.5 cell steps -- maybe that's worth a try instead of using world coordinates? `line_iterator = LineIterator(worldToMap(float(x0)), worldToMapfloat(y0)), worldToMap(x1), worldToMap(y1), 0.5)` 

> This Image is using current ray tracking algorithm

I also like this in concept, but the arcing is a bit unnatural and I think would cause some odd behaviors with some localization algorithms.

JatinPatil2003:
> This image is using LineIterator

This has been fixed. Now it is working perfectly

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -77,10 +73,6 @@ def __init__(self):
         self.declare_parameter('scan_frame_id', 'base_scan')
         self.scan_frame_id = self.get_parameter('scan_frame_id').get_parameter_value().string_value
 
-        self.declare_parameter('publish_map_odom_tf', True)
```
SteveMacenski: Why is this removed? Please readd

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,23 +198,26 @@ def timerCallback(self):
...
-        scan.angle_increment = 0.005817705996  # 0.333 degrees
-        scan.time_increment = 0.0
-        scan.scan_time = 0.1
-        scan.range_min = 0.1
-        scan.range_max = 100.0
-        num_samples = int((scan.angle_max - scan.angle_min) / scan.angle_increment)
-        scan.ranges = [scan.range_max - 0.01] * num_samples
-        self.scan_pub.publish(scan)
+        self.scan_msg = LaserScan()
+        self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```
SteveMacenski, line 205:
```suggestion
        # self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```

Please leave this as commented out. I'll test after this is approved to see if readding that still works in the launch file

SteveMacenski: Note to self:
- Test localization
- Test SLAM Toolbox
- Test with TB3/4 launch files

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -209,6 +245,89 @@ def debug(self, msg):
...
+        mx0, my0 = worldToMap(x0, y0, self.map)
+
+        if not 0 < mx0 < self.map.info.width or not 0 < my0 < self.map.info.height:
+            # outside map
+            self.scan_msg.ranges = [self.scan_msg.range_max - 0.1] * num_samples
+            return
+
+        for i in range(num_samples):
+            curr_angle = theta + self.scan_msg.angle_min + i * self.scan_msg.angle_increment
+            x1 = x0 + self.scan_msg.range_max * math.cos(curr_angle)
```
SteveMacenski, line 311: `range_max` should be in grid coordinates ( divided by resolution)

SteveMacenski (approved):

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
JatinPatil2003: I don't get it x0 and x1 are world coordinates so basically distance between (x0, y0) and (x1, y1) should be range_max, which should be in meters

let me know if I'm wrong here

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,29 +202,33 @@ def timerCallback(self):
...
-        scan.angle_increment = 0.005817705996  # 0.333 degrees
-        scan.time_increment = 0.0
-        scan.scan_time = 0.1
-        scan.range_min = 0.1
-        scan.range_max = 100.0
-        num_samples = int((scan.angle_max - scan.angle_min) / scan.angle_increment)
-        scan.ranges = [scan.range_max - 0.01] * num_samples
-        self.scan_pub.publish(scan)
+        self.scan_msg = LaserScan()
+        # self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```
SteveMacenski, line 209: OK - I played with this and this can be uncommented out. I'm not sure why I was having issues before.

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
```diff
@@ -165,29 +202,33 @@ def timerCallback(self):
...
+        self.scan_msg.header.frame_id = self.scan_frame_id
+        self.scan_msg.angle_min = -math.pi
+        self.scan_msg.angle_max = math.pi
+        # 1.5 degrees
+        self.scan_msg.angle_increment = 0.0261799
+        self.scan_msg.time_increment = 0.0
+        self.scan_msg.scan_time = 0.1
+        self.scan_msg.range_min = 0.05
+        self.scan_msg.range_max = 100.0
```
SteveMacenski, line 218:
```suggestion
        self.scan_msg.range_max = 30.0
```

nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py
SteveMacenski, line 209:
```suggestion
        self.scan_msg.header.stamp = (self.get_clock().now()).to_msg()
```

SteveMacenski:
Thanks @JatinPatil2003 !

PR #4620, opened 2024-08-14
Added promixity BT node and BT tree

Jakubach:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | (Gazebo simulation of custom robot) |

---

BT Node that return if the robot is in the goal proximity
PR related to the Steve advice on robotics stack:
https://robotics.stackexchange.com/questions/112576/maneuvers-on-paths-end-point-with-navigation2-smac-lattice-planner/112577?noredirect=1#comment48615_112577

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp
@@ -1,0 +1,81 @@
+// Copyright (c) 2024 Jakub Chudziński
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__IS_GOAL_NEARBY_CONDITION_HPP_
+#define NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__IS_GOAL_NEARBY_CONDITION_HPP_
+
+#include <string>
+#include "nav_msgs/msg/path.hpp"
+
+#include "rclcpp/rclcpp.hpp"
+#include "behaviortree_cpp/condition_node.h"
+
+
+namespace nav2_behavior_tree
+{
+
+/**
+ * @brief A BT::ConditionNode that returns SUCCESS when the IsGoalNearby
+ * service returns true and FAILURE otherwise
+ */
+class IsGoalNearbyCondition : public BT::ConditionNode
+{
+public:
+  /**
+   * @brief A constructor for nav2_behavior_tree::IsGoalNearbyCondition
+   * @param condition_name Name for the XML tag for this node
+   * @param conf BT node configuration
+   */
+  IsGoalNearbyCondition(
+    const std::string & condition_name,
+    const BT::NodeConfiguration & conf);
+
+  IsGoalNearbyCondition() = delete;
+
+  /**
+   * @brief The main override required by a BT action
+   * @return BT::NodeStatus Status of tick execution
+   */
+  BT::NodeStatus tick() override;
+
+  /**
+   * @brief Creates list of BT ports
+   * @return BT::PortsList Containing node-specific ports
+   */
+  static BT::PortsList providedPorts()
+  {
+    return {
+      BT::InputPort<nav_msgs::msg::Path>("path", "Planned Path"),     
+      BT::InputPort<double>(
+        "proximity_threshold", 3.0,
+        "Proximity length (m) of the remaining path considered as a nearby"),
+    };
+  }
+
+private:
+        
+    /**
+     * @brief Checks if the robot is in the goal proximity
+     * @param goal_path current planned path to the goal
+     * @param prox_thr proximity length (m) of the remaining path considered as a nearby
+     * @return whether the robot is in the goal proximity
+     */
+    bool isRobotInGoalProximity(
+        const nav_msgs::msg::Path& goal_path,
+        const double& prox_thr);
+};
+
+}  // namespace nav2_behavior_tree
+
+#endif  // NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__IS_GOAL_NEARBY_CONDITION_HPP_
--- a/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.cpp
+++ b/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.cpp
@@ -1,0 +1,42 @@
+#include "nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp"
+#include "nav2_util/geometry_utils.hpp"
+
+namespace nav2_behavior_tree
+{
+
+IsGoalNearbyCondition::IsGoalNearbyCondition(
+  const std::string & condition_name,
+  const BT::NodeConfiguration & conf)
+: BT::ConditionNode(condition_name, conf)
+{
+}
+
+bool IsGoalNearbyCondition::isRobotInGoalProximity(
+  const nav_msgs::msg::Path& goal_path,
+  const double& prox_thr)
+{
+  return nav2_util::geometry_utils::calculate_path_length(goal_path, 0) < prox_thr;
+}
+
+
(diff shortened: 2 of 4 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp
```diff
@@ -0,0 +1,81 @@
...
+     * @return whether the robot is in the goal proximity
+     */
+    bool isRobotInGoalProximity(
+        const nav_msgs::msg::Path& goal_path,
+        const double& prox_thr);
+};
+
+}  // namespace nav2_behavior_tree
+
+#endif  // NAV2_BEHAVIOR_TREE__PLUGINS__CONDITION__IS_GOAL_NEARBY_CONDITION_HPP_
```
SteveMacenski, line 81: EOF line

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+
+namespace nav2_behavior_tree
+{
+
+/**
+ * @brief A BT::ConditionNode that returns SUCCESS when the IsGoalNearby
+ * service returns true and FAILURE otherwise
+ */
+class IsGoalNearbyCondition : public BT::ConditionNode
```
SteveMacenski, line 32: This needs test coverage

nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.cpp
```diff
@@ -0,0 +1,42 @@
...
+    }
+    return BT::NodeStatus::FAILURE;
+}
+
+}  // namespace nav2_behavior_tree
+
+#include "behaviortree_cpp/bt_factory.h"
+BT_REGISTER_NODES(factory)
+{
+  factory.registerNodeType<nav2_behavior_tree::IsGoalNearbyCondition>("IsGoalNearby");
```
SteveMacenski, line 41: Need to add this to the [nav2 node index](https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/nav2_tree_nodes.xml)

nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.cpp
```diff
@@ -0,0 +1,42 @@
...
+  const BT::NodeConfiguration & conf)
+: BT::ConditionNode(condition_name, conf)
+{
+}
+
+bool IsGoalNearbyCondition::isRobotInGoalProximity(
+  const nav_msgs::msg::Path& goal_path,
+  const double& prox_thr)
+{
+  return nav2_util::geometry_utils::calculate_path_length(goal_path, 0) < prox_thr;
```
SteveMacenski, line 18: That's checking the last path length, but not the robot's proximity. If you replan every 10 seconds (or only on events) then this wouldn't tell you much about the robot's proximity to the goal based on the last path marker.

I think this either needs to 
- find the path's closest point like we do in the Controller plugins [1] which means we need to track the last path index to know where to search starting from on each call up to some maximum distance 
- just be based on a distance check from the robot's current pose -- which has problems if the path have overlapping segments from Navigate Through Poses. 

[1] https://github.com/ros-navigation/navigation2/blob/main/nav2_regulated_pure_pursuit_controller/src/path_handler.cpp#L61-L79

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_path_if_not_in_goal_proximity.xml
```diff
@@ -0,0 +1,46 @@
...
+            <ClearEntireCostmap name="ClearGlobalCostmap-Subtree" service_name="global_costmap/clear_entirely_global_costmap"/>
+          </Sequence>
+          <Spin name="SpinRecovery" spin_dist="1.57"/>
+          <Wait name="WaitRecovery" wait_duration="5"/>
+          <BackUp name="BackUpRecovery" backup_dist="0.30" backup_speed="0.05"/>
+        </RoundRobin>
+      </ReactiveFallback>
+    </RecoveryNode>
+  </BehaviorTree>
+</root>
```
SteveMacenski, line 46: EOF line

nav2_bt_navigator/behavior_trees/navigate_w_recovery_and_replanning_path_if_not_in_goal_proximity.xml
```diff
@@ -0,0 +1,46 @@
...
+        <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath" topic_name="controller_selector"/>
+        <PlannerSelector selected_planner="{selected_planner}" default_planner="GridBased" topic_name="planner_selector"/>
+        <RateController hz="1.0" name="RateControllerComputePathToPose">
+          <RecoveryNode number_of_retries="1" name="RecoveryComputePathToPose">
+            <Fallback name="FallbackComputePathToPose">
+              <ReactiveSequence name="CheckIfNewPathNeeded">
+                <Inverter>
+                  <GlobalUpdatedGoal/>
+                </Inverter>
+                <IsGoalNearby path="{path}" proximity_threshold="2.0"/>
```
SteveMacenski, line 19: Did you test this?

SteveMacenski (changes requested):
Otherwise, needs a PR to add this new BT node to the migration guide and a configuration guide page to the BT XML nodes for this node. Looks good to start with!

nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.cpp
Jakubach: I am trying to implement the first idea but I am not sure if I've understood it correctly. It's should return a value to the closest point in the path (in case of straight line it would be a next point on the line, and in case of complex path not necessarily)?

```
BT::NodeStatus IsGoalNearbyCondition::tick()
{
    nav_msgs::msg::Path path;
    double prox_thr;
    getInput("path", path);
    getInput("proximity_threshold", prox_thr);

    geometry_msgs::msg::PoseStamped pose; // robot pose in map frame
    nav2_util::getCurrentPose(
    pose, *tf_buffer_,
    "map", "base_link", 0.05);
    
    geometry_msgs::msg::PoseStamped robot_pose; // robot_pose in path frame
    if (!transformPose(path.header.frame_id, pose, robot_pose)) {
    }

    nav2_util::geometry_utils::first_after_integrated_distance(
    path.poses.begin(), path.poses.end(), max_robot_pose_search_dist_);


    // First find the closest pose on the path to the robot
    // bounded by when the path turns around (if it does) so we don't get a pose from a later
    // portion of the path
    auto closest_pose_it =
    nav2_util::geometry_utils::min_by(
    path.poses.begin() + last_closest_index_, closest_pose_upper_bound,
    [&robot_pose](const geometry_msgs::msg::PoseStamped & ps) {
      return nav2_util::geometry_utils::euclidean_distance(robot_pose, ps);
    });

    last_closest_index_ = std::distance(path.poses.begin(), closest_pose_it);

    double distance_to_closest_point = nav2_util::geometry_utils::euclidean_distance(robot_pose, *closest_pose_it);
    RCLCPP_INFO(node_->get_logger(), "Distance to closest point: %f", distance_to_closest_point);

    if(distance_to_closest_point < prox_thr){
    }
}
```

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp
Jakubach: What should I correct in this case?

SteveMacenski:
@Jakubach following up here - have you had a chance to work on this?

Jakubach:
> @Jakubach following up here - have you had a chance to work on this?

@SteveMacenski I had another job but will back to this probably this month or November.

SteveMacenski:
Great, thank you!

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_goal_nearby_condition.hpp
SteveMacenski: Add an extra line at the end of the file so that there's 81 lines instead of 81

mini-1235:
Are there any progress on this? 

Why don't we just check the distance from "Goal" to "Current Robot Pose" instead of using "Planned Path" here, I think that also solves the issue(?) Then we can copy + paste from `GoalReached` and add an input distance that can be configured by the user, we will also check if the path is valid/inavlid, if valid we can stop replanning.

SteveMacenski:
Woof, that conflict hit alot. Can you clean this up and make the last few changes so we can get this in this year? :smile:

Jakubach:
@SteveMacenski I just moved to the new PR to clean it up: https://github.com/ros-navigation/navigation2/pull/5843.
I will finish it tomorrow hopefully and after that ping you there for review

PR #4622, opened 2024-08-17
fixing path longer on approach

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4110  |
| Robotic platform tested on | ROX |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
+++ b/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
@@ -36,6 +36,7 @@ bool PathLongerOnApproach::isPathUpdated(
 {
   return old_path.poses.size() != 0 &&
          new_path.poses.size() != 0 &&
+         new_path.poses.size() != old_path.poses.size() &&
          old_path.poses.back().pose.position == new_path.poses.back().pose.position;
 }
 
--- a/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
+++ b/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
@@ -35,7 +35,7 @@ bool PathLongerOnApproach::isPathUpdated(
   nav_msgs::msg::Path & old_path)
 {
   return old_path.poses.size() != 0 &&
-         new_path.poses.size() != 0 && new_path.poses.size() > old_path.poses.size() &&
+         new_path.poses.size() != 0 &&
          old_path.poses.back().pose.position == new_path.poses.back().pose.position;
 }
 
--- a/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
+++ b/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
@@ -34,9 +34,9 @@ bool PathLongerOnApproach::isPathUpdated(
   nav_msgs::msg::Path & new_path,
   nav_msgs::msg::Path & old_path)
 {
-  return new_path != old_path && old_path.poses.size() != 0 &&
-         new_path.poses.size() != 0 &&
-         old_path.poses.back().pose == new_path.poses.back().pose;
+  return old_path.poses.size() != 0 &&
+         new_path.poses.size() != 0 && new_path.poses.size() > old_path.poses.size() &&
+         old_path.poses.back().pose.position == new_path.poses.back().pose.position;
 }
 
 bool PathLongerOnApproach::isRobotInGoalProximity(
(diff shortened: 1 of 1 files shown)


padhupradheep:
almost done!! I want to test it once again in fresh eyes before getting this in. 

Sorry, it got longer than expected.

SteveMacenski:
I know how that goes sometimes :wink:

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
```diff
@@ -34,9 +34,9 @@ bool PathLongerOnApproach::isPathUpdated(
   nav_msgs::msg::Path & new_path,
   nav_msgs::msg::Path & old_path)
 {
-  return new_path != old_path && old_path.poses.size() != 0 &&
-         new_path.poses.size() != 0 &&
-         old_path.poses.back().pose == new_path.poses.back().pose;
+  return old_path.poses.size() != 0 &&
```
SteveMacenski, line 37: A new path could be shorter though?

Wouldn't it be something like 

```
new_path != old_path && !old_path.poses.empty() && !new_path,poses.empty()
```

Shouldn't the `new_path != old_path` iterate through the vector and check that all contents are the same? I don't see the reason for checking `back().pose.position` or `back().pose` at all

padhupradheep: `new_path != old_path` would be only valid for the NavFn planner, for the smac planners there are headers attached to each positions of the plan, which eventually leads to failure when we do that above said comparison.  That is the reason why I had to stick to the `back().pose`

SteveMacenski: Ah yes, I remember this now, thanks. 

```
  return old_path.poses.size() != 0 &&
         new_path.poses.size() != 0 && new_path.poses.size() > old_path.poses.size() &&
         old_path.poses.back().pose.position == new_path.poses.back().pose.position;
```

Why the ` new_path.poses.size() > old_path.poses.size()`? Can't a new path be shorter? Should this be a `==`?

padhupradheep: >Can't a new path be shorter?

Good question! Sure, the path can be shorter. It was just a design decision, where the intention was to short circuit the if condition [here](https://github.com/ros-navigation/navigation2/blob/1d745ec7e7821f5e2ed066d7b70935c2ecf12016/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp#L76), as soon as the updated path is shorter, since the bigger picture of this node is to check if the new path is longer than the older one. 

Thus in this scenario, the if condition fails and continues with the execution as soon as the first condition is evaluated without the need for the evaluation of the other conditions.

SteveMacenski: Vacation away! 

> it was just a design decision, where the intention was to short circuit the if condition [here](https://github.com/ros-navigation/navigation2/blob/1d745ec7e7821f5e2ed066d7b70935c2ecf12016/nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp#L76),

Lets make sure each method does what a user would expect it to do by name, so I don't think we should short circuit it. Life is long, intentions are forgotten, and code is copied - so keeping things decoupled and meeting the intent of the method is good for maintainability!

padhupradheep:
I'm back! Will finish it EoD

padhupradheep:
Gtg now

SteveMacenski:
@padhupradheep this PR needs to be rebased to build - there was a rolling API change

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
```diff
@@ -34,9 +34,9 @@ bool PathLongerOnApproach::isPathUpdated(
   nav_msgs::msg::Path & new_path,
   nav_msgs::msg::Path & old_path)
 {
-  return new_path != old_path && old_path.poses.size() != 0 &&
+  return old_path.poses.size() != 0 &&
          new_path.poses.size() != 0 &&
-         old_path.poses.back().pose == new_path.poses.back().pose;
+         old_path.poses.back().pose.position == new_path.poses.back().pose.position;
```
SteveMacenski, line 39: Wait, what happened to checking the sizes are the same `new_path.poses.size() == old_path.poses.size()`?

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/decorator/path_longer_on_approach.cpp
padhupradheep: Oh yeahh.. you mentioned that in you previous comments. 

The function checks `isPathUpdated`, having the equality means then the function returns true, when the path is not updated. Therefore it should be `!=`. Will update it soon..

padhupradheep: Sorry it's taking longer than expected. I'd be on it soon 🥲

SteveMacenski (approved):

PR #4624, opened 2024-08-19
dynamic parameter handling of observation_sources for obstacle layer

dylan-gonzalez:
## Basic Info

| ------ | ----------- |

Dynamic parameter handling for `observation_sources` in the obstacle layer.

This functionality can also be extended to the voxel_layer.

`observation_sources` should probably be identified in the docs as a dynamic parameter.

I made this a Draft PR as there is probably some optimisation to be done here, and it might be best to not remove sources from the buffers entirely but keep them there, with their subscribers simply unsubscribed? Would like to get some feedback on my implementation here

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -333,7 +333,7 @@ ObstacleLayer::dynamicParametersCallback(
         std::vector<std::string> matched_sources;
         std::ostringstream new_sources_stream;
 
-        //find sources that already have been defined
+        //find all the sources that have already been set.
         while (ss >> source) {
           while (ss_old >> source_old) {
             if (source == source_old) {
@@ -359,7 +359,7 @@ ObstacleLayer::dynamicParametersCallback(
         std::int16_t index = 0;
         std::stringstream ss_old_final(observation_sources_);
 
-        // find the subscribers that weren't mentioned and remove them
+        // find and remove the sources that were not passed into the parameter update
         while (ss_old_final >> source_old) {
           if (std::find(matched_sources.begin(), matched_sources.end(), source_old) == matched_sources.end()) {
             // The observation source needs to be removed
@@ -378,6 +378,7 @@ ObstacleLayer::dynamicParametersCallback(
           }
         }
 
+        //initializse any new sources
         auto start_index = observation_subscribers_.size() - 1;
         if (!new_sources.empty()) {
           initializeObservationSources(new_sources);
@@ -387,7 +388,6 @@ ObstacleLayer::dynamicParametersCallback(
           }
         }
 
-        //replace with new sources
         observation_sources_ = parameter.as_string();
       }
     }
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -71,10 +71,6 @@ ObstacleLayer::~ObstacleLayer()
 void ObstacleLayer::onInitialize()
 {
   bool track_unknown_space;
-  double transform_tolerance;
-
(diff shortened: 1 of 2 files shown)


SteveMacenski:
@dylan-gonzalez heed the bot + make sure to sign off on your commits with DCO.

SteveMacenski:
How many sources do you have going into a particular layer? typically, I would think that either (1) you have separate layers for separate applications so you can disable/enable layers that contain the sensors that you want used in different situations so that its more semantically meaningful or (2) have the inputs to the costmap layer be themselves enabled/disabled with muxing. We could also enable dynamically setting each source for each layer too, but just wanted to mention some other options that already exist.

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -313,7 +321,77 @@ ObstacleLayer::dynamicParametersCallback(
...
+
+        // find and remove the sources that were not passed into the parameter update
+        while (ss_old_final >> source_old) {
+          if (std::find(matched_sources.begin(), matched_sources.end(), source_old) == matched_sources.end()) {
+            // The observation source needs to be removed
+            RCLCPP_INFO(
+              logger_,
+              "Removing observation source: %s", source_old.c_str());
+
+            // assume that the order of the sources in observation_sources_ is the same as in the vectors
```
SteveMacenski, line 370: Make sure these are in the right order to not cause segfaults or accessing incorrect memory. Also make sure to unsubscriber or otherwise reset the state before destroying them to also prevent other potential race conditions

dylan-gonzalez:
> How many sources do you have going into a particular layer?

My sensors are external to the robot and are defined along a set path, so for the local costmap I'd only want to use information from sensors nearest to the robot. Let's say I have 10 sensors along the path, and only want to use 2 at any point in time for the local costmap's obstacle layer

> have the inputs to the costmap layer be themselves enabled/disabled with muxing

This seems like a good option, I hadn't thought of it. And it's probably quicker than having to reinitialise subscribers all the time, etc with the dynamic parameter handling route. Looks like ros2 topic tools has a mux node, but seems to only allow a single output topic (as a normal multiplexer should). I might look into adding support for a multi-output multiplexer.

Thanks for the feedback @SteveMacenski

SteveMacenski:
> My sensors are external to the robot and are defined along a set path, so for the local costmap I'd only want to use information from sensors nearest to the robot.

Neat! I'll take you at face value that this is the right answer for how to handle it. 

> Looks like ros2 topic tools has a mux node, but seems to only allow a single output topic (as a normal multiplexer should). I might look into adding support for a multi-output multiplexer.

I'd do `N` muxers (one for each source), whereas you have some simple program that decides which sources to use at a given time and send the OK signal to those and the STOP signal to the rest. No matter what, some program has to decide which streams are OK/NOT, so the difference between a single muxing node to handle `N` vs `N` muxing nodes is pretty inconsequential, I think. There are solid arguments either way and unless you were very particular, I'd go the path of least resistance. One throws the updates-for-a-new-environment into a config file for 1 node vs the other which throws it into a launch file to spin up `N` appropriate nodes in a loop. 

Though, its worth noting that the obstacle layer should handle `N` sources in a single topic OK, even if its not what I would advice from a debuggability / tracability perspective. I believe each measurement reading's frame is stored for transformation in the buffer, so they should each be handled correctly (should you want to do that, against my advice). 

---

Should we leave this open / work on it? Or do you think those alternatives are better for your system?

dylan-gonzalez:
I'll take the path of least resistance - `N` muxers. Seems most reasonable and simple to implement.

The only 'bad' thing I noticed about using all `N` sources with the obstacle layer is that it would complain that some sensors were outside the bounds of the local costmap (which makes sense), but I guess that's the sort of thing you're talking about in terms of debuggability.

I'll close this, thanks again.

PR #4634, opened 2024-08-20
Switch nav2_smac_planner to modern CMake idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |

---

Switch nav2_smac_planner to use modern CMake idioms:
1. Switch from ament_target_dependencies to target_link_libraries
2. Don't compile the common files in this package more than once.  Instead, compile those into a library, then link that library into all of the other ones.
3.  Export a CMake target so downstream packages can use it.
4.  Push the include directories down one level, which is best practice since Humble.

None needed.

---

This is part of a larger series to switch Navigation2 to modern CMake idioms.  After this PR, there 3 PRs to go.

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/CMakeLists.txt
+++ b/nav2_smac_planner/CMakeLists.txt
@@ -4,29 +4,23 @@ project(nav2_smac_planner)
 set(CMAKE_BUILD_TYPE Release) #Debug, Release
 
 find_package(ament_cmake REQUIRED)
-find_package(nav2_common REQUIRED)
-find_package(rclcpp REQUIRED)
-find_package(rclcpp_action REQUIRED)
-find_package(rclcpp_lifecycle REQUIRED)
-find_package(std_msgs REQUIRED)
-find_package(visualization_msgs REQUIRED)
-find_package(nav2_util REQUIRED)
-find_package(nav2_core REQUIRED)
-find_package(nav2_msgs REQUIRED)
-find_package(nav_msgs REQUIRED)
-find_package(geometry_msgs REQUIRED)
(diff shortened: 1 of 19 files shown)


nav2_smac_planner/CMakeLists.txt
```diff
@@ -4,137 +4,168 @@ project(nav2_smac_planner)
...
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 17)
-endif()
+find_package(pluginlib REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
+find_package(rcl_interfaces REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(visualization_msgs REQUIRED)
```
SteveMacenski, line 24: `nav2_package()`?

clalancette: Good point.

When I turn that on locally, I get a host of new compiler warnings (which are then errors, because of `-Werror`).  Would you like me to submit a separate PR where we *only* turn on `nav2_package()` and fix the warnings, or do you want me to just fix them all in this PR?

SteveMacenski: A new PR would be fine. Also, if you'd prefer, you could file a ticket with those errors and we can kick that can for another time entirely. I don't think that's critically in your path of what you're aiming to do, so I don't want to pile on - unless you'd like to :-)

PR #4635, opened 2024-08-21
Fixes for nav2_smac_planner to build with -Werror.

clalancette:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Follow-up to #4634  |

---

This is a follow-up PR to #4634 where we change `nav2_smac_planner` to use `nav2_package`.  This turns on additional compiler warnings, along with `-Werror`, so it requires some fixes to go along with it.

I want to highlight one controversial change in here.  The [UNKNOWN constant in constants.hpp](https://github.com/ros-navigation/navigation2/commit/03577dc1486ef6c426370252a708fc0fdf6a9b36#diff-47d128ffcb4f25caaf908b5962cefca6558de3139d5a9a2ea03f88e04dcdfadeR62) was causing shadowing problems with `UNKNOWN` embedded inside of `enum class` elsewhere in the code.  So I renamed that constant (along with the rest) to `UNKNOWN_COST`.  This breaks the API, but resolves the error.

It will possibly be necessary to update the documentation to mention the API break in constants.hpp.

---

None needed.

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/analytic_expansion.cpp
+++ b/nav2_smac_planner/src/analytic_expansion.cpp
@@ -337,44 +337,29 @@ void AnalyticExpansion<NodeT>::cleanNode(const NodePtr & /*expanded_nodes*/)
 template<>
 typename AnalyticExpansion<Node2D>::AnalyticExpansionNodes AnalyticExpansion<Node2D>::
 getAnalyticPath(
-  const NodePtr & node,
-  const NodePtr & goal,
-  const NodeGetter & node_getter,
-  const ompl::base::StateSpacePtr & state_space)
+  const NodePtr &,
+  const NodePtr &,
+  const NodeGetter &,
+  const ompl::base::StateSpacePtr &)
 {
-  (void)node;
-  (void)goal;
-  (void)node_getter;
-  (void)state_space;
-
   return AnalyticExpansionNodes();
 }
 
 template<>
 typename AnalyticExpansion<Node2D>::NodePtr AnalyticExpansion<Node2D>::setAnalyticPath(
-  const NodePtr & node,
-  const NodePtr & goal_node,
-  const AnalyticExpansionNodes & expanded_nodes)
+  const NodePtr &,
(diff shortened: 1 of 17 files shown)


nav2_smac_planner/src/analytic_expansion.cpp
```diff
@@ -360,6 +369,12 @@ typename AnalyticExpansion<Node2D>::NodePtr AnalyticExpansion<Node2D>::tryAnalyt
   const NodeGetter & getter, int & analytic_iterations,
```
SteveMacenski, line 369: By Nav2 styling, we typically do `const NodePtr & /*current_node*/` in the method signatures to avoid the `(void)current_node;`'s inline. Can you update for here and other instances? 

I'd also be fine with simply a `const NodePtr &` with no variable as well given these are passthroughs that will never be modified to use them

SteveMacenski:
Christ. This kind of amateur shit I have no excuse for. I don't know how `nav2_package()` got overlooked for this package. A few of these probably created real bugs. This was a really good nitpick note to uncover.

nav2_smac_planner/src/analytic_expansion.cpp
clalancette: > By Nav2 styling, we typically do `const NodePtr & /*current_node*/` in the method signatures to avoid the `(void)current_node;`'s inline. Can you update for here and other instances?

That's fine.  I should have checked the style elsewhere first.  Will make the change.

SteveMacenski: I'm only really worried about nav2_smac_planner/src/analytic_expansion.cpp & nav2_smac_planner/src/node_hybrid.cpp instances. I'm alot more relaxed on unit tests.

clalancette: Fixed in 4379bbe.

SteveMacenski (approved):

PR #4638, opened 2024-08-21
Switch nav2_system_tests to modern CMake idioms.

clalancette:
---

## Basic Info

| ------ | ----------- |

---

Update nav2_system_tests to use modern CMake idioms:
1.  Make everything in this package a `test_depend`, since this is all tests.
2.  Along with the previous point, move everything in this package under the BUILD_TESTING conditional in CMake.
3.  Switch from ament_target_dependencies to target_link_libraries everywhere.
4.  Remove all references to Boost.

None needed.

---

This is part of a larger series updating Navigation2 to use modern CMake idioms.  After this PR, there are 2 PRs left to open.

#### For Maintainers: 

### DIFF
--- a/nav2_system_tests/CMakeLists.txt
+++ b/nav2_system_tests/CMakeLists.txt
@@ -1,114 +1,110 @@
 cmake_minimum_required(VERSION 3.5)
 project(nav2_system_tests)
 
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 23)
-endif()
-
 find_package(ament_cmake REQUIRED)
 find_package(nav2_common REQUIRED)
-find_package(rclcpp REQUIRED)
-find_package(std_msgs REQUIRED)
-find_package(nav2_util REQUIRED)
-find_package(nav2_map_server REQUIRED)
-find_package(nav2_behavior_tree REQUIRED)
-find_package(nav2_msgs REQUIRED)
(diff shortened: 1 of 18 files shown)


clalancette:
I should also mention; with this in, my compile time savings locally is about 1 minute.  This is substantially less than I originally saw in #4357; I suspect that there is some other bug on the `iron` branch that makes that one really slow.  Regardless, I think we accomplished a lot with this series in terms of cleanup, so I'm happy we did it.

I'm also going to close #4357 at this point, as with this PR we accomplish everything that one set out to, and more.

SteveMacenski (approved):

SteveMacenski:
I also noticed that the WPF wasn't updated https://github.com/ros-navigation/navigation2/blob/main/nav2_waypoint_follower/CMakeLists.txt - I don't think that's probably the difference maker in compile times, but would you mind doing the last one so they're all homogeneous?

clalancette:
> I also noticed that the WPF wasn't updated https://github.com/ros-navigation/navigation2/blob/main/nav2_waypoint_follower/CMakeLists.txt - I don't think that's probably the difference maker in compile times, but would you mind doing the last one so they're all homogeneous?

Yeah, that's one of the last 2 PRs I have (the other is just some minor cleanup here and there).  I left it to last because it also requires an update to `robot_localization`, which I haven't opened the PR for yet either.  I'll do that soon.

PR #4640, opened 2024-08-22
Publish optimal trajectory as a Path message

alyquantillion:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Gazebo simulation of a custom robot |

---

* I publish the optimal trajectory as a `nav_msgs::msg::Path` message. Needed if we want to calculate the velocity commands ourselves.

None

---

None

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/controller.cpp
+++ b/nav2_mppi_controller/src/controller.cpp
@@ -45,8 +45,6 @@ void MPPIController::configure(
     parent_, name_,
     costmap_ros_->getGlobalFrameID(), parameters_handler_.get());
 
-  optimal_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("local_plan", 1);
-
   RCLCPP_INFO(logger_, "Configured MPPI Controller: %s", name_.c_str());
 }
 
@@ -62,14 +60,12 @@ void MPPIController::activate()
 {
   trajectory_visualizer_.on_activate();
   parameters_handler_->start();
-  optimal_path_pub_->on_activate();
   RCLCPP_INFO(logger_, "Activated MPPI Controller: %s", name_.c_str());
 }
 
 void MPPIController::deactivate()
 {
   trajectory_visualizer_.on_deactivate();
-  optimal_path_pub_->on_deactivate();
   RCLCPP_INFO(logger_, "Deactivated MPPI Controller: %s", name_.c_str());
 }
 
@@ -102,22 +98,18 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   RCLCPP_INFO(logger_, "Control loop execution time: %ld [ms]", duration);
 #endif
 
-  if (optimal_path_pub_->get_subscription_count() > 0) {
-    publish_optimal_path(optimizer_.getOptimizedTrajectory());
-  }
-
   if (visualize_) {
-    visualize(std::move(transformed_plan));
+    visualize(std::move(transformed_plan), cmd.header.stamp);
   }
 
   return cmd;
 }
 
-void MPPIController::visualize(nav_msgs::msg::Path transformed_plan)
+void MPPIController::visualize(nav_msgs::msg::Path transformed_plan, const builtin_interfaces::msg::Time & cmd_stamp)
 {
   trajectory_visualizer_.add(optimizer_.getGeneratedTrajectories(), "Candidate Trajectories");
   trajectory_visualizer_.add(optimizer_.getOptimizedTrajectory(), "Optimal Trajectory");
-  trajectory_visualizer_.visualize(std::move(transformed_plan));
+  trajectory_visualizer_.visualize(std::move(transformed_plan), cmd_stamp);
 }
 
 void MPPIController::setPlan(const nav_msgs::msg::Path & path)
@@ -128,36 +120,6 @@ void MPPIController::setPlan(const nav_msgs::msg::Path & path)
 void MPPIController::setSpeedLimit(const double & speed_limit, const bool & percentage)
 {
   optimizer_.setSpeedLimit(speed_limit, percentage);
-}
-
-void MPPIController::publish_optimal_path(const xt::xtensor<float, 2> & optimal_traj)
-{
-  auto & size = optimal_traj.shape()[0];
-  if (!size) {
-    return;
-  }
-
-  nav_msgs::msg::Path optimal_path;
-  optimal_path.header.stamp = clock_->now();
-  optimal_path.header.frame_id = (costmap_ros_->getGlobalFrameID()).c_str();
-
-  for (size_t i = 0; i < size; i++) {
-    // create new pose for the path
-    geometry_msgs::msg::PoseStamped pose_stamped;
-    pose_stamped.header.frame_id = (costmap_ros_->getGlobalFrameID()).c_str();
-
-    // position & orientation
-    pose_stamped.pose = utils::createPose(optimal_traj(i, 0), optimal_traj(i, 1), 0.);
-
-    tf2::Quaternion quaternion_tf2;
-    quaternion_tf2.setRPY(0., 0., optimal_traj(i, 2));
-    auto quaternion = tf2::toMsg(quaternion_tf2);
-    pose_stamped.pose.orientation = quaternion;
-
-    // add pose to the path
-    optimal_path.poses.push_back(pose_stamped);
-  }
-  optimal_path_pub_->publish(optimal_path);
 }
 
 }  // namespace nav2_mppi_controller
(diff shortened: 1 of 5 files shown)


nav2_mppi_controller/src/controller.cpp
```diff
@@ -122,6 +130,36 @@ void MPPIController::setSpeedLimit(const double & speed_limit, const bool & perc
   optimizer_.setSpeedLimit(speed_limit, percentage);
 }
 
+void MPPIController::publish_optimal_path(const xt::xtensor<float, 2> & optimal_traj)
+{
+  auto & size = optimal_traj.shape()[0];
+  if (!size) {
```
SteveMacenski, line 136: This is an int, check the size explicitly, don't use boolean operations please

nav2_mppi_controller/src/controller.cpp
```diff
@@ -122,6 +130,36 @@ void MPPIController::setSpeedLimit(const double & speed_limit, const bool & perc
...
+    return;
+  }
+
+  nav_msgs::msg::Path optimal_path;
+  optimal_path.header.stamp = clock_->now();
```
SteveMacenski, line 141: Use the stamp from the TwistStamped returned so that they are properly synchronized

nav2_mppi_controller/src/controller.cpp
```diff
@@ -122,6 +130,36 @@ void MPPIController::setSpeedLimit(const double & speed_limit, const bool & perc
...
+  optimal_path.header.frame_id = (costmap_ros_->getGlobalFrameID()).c_str();
```
SteveMacenski, line 142:
```suggestion
  optimal_path.header.frame_id = costmap_ros_->getGlobalFrameID();
```

nav2_mppi_controller/src/controller.cpp
```diff
@@ -122,6 +130,36 @@ void MPPIController::setSpeedLimit(const double & speed_limit, const bool & perc
...
+
+  for (size_t i = 0; i < size; i++) {
+    // create new pose for the path
+    geometry_msgs::msg::PoseStamped pose_stamped;
+    pose_stamped.header.frame_id = (costmap_ros_->getGlobalFrameID()).c_str();
```
SteveMacenski, line 147:
```suggestion
    pose_stamped.header.frame_id = costmap_ros_->getGlobalFrameID();
```

nav2_mppi_controller/src/controller.cpp
```diff
@@ -122,6 +130,36 @@ void MPPIController::setSpeedLimit(const double & speed_limit, const bool & perc
...
+
+    // position & orientation
+    pose_stamped.pose = utils::createPose(optimal_traj(i, 0), optimal_traj(i, 1), 0.);
```
SteveMacenski, line 150:
```suggestion
    pose_stamped.pose = utils::createPose(optimal_traj(i, 0), optimal_traj(i, 1), 0.0);
```

nav2_mppi_controller/src/controller.cpp
```diff
@@ -122,6 +130,36 @@ void MPPIController::setSpeedLimit(const double & speed_limit, const bool & perc
...
+
+    tf2::Quaternion quaternion_tf2;
+    quaternion_tf2.setRPY(0., 0., optimal_traj(i, 2));
+    auto quaternion = tf2::toMsg(quaternion_tf2);
+    pose_stamped.pose.orientation = quaternion;
```
SteveMacenski, line 155: Can simply be `pose_stamped.pose.orientation = tf2::toMsg(quaternion_tf2);`, don't need intermediate variable

nav2_mppi_controller/src/controller.cpp
```diff
@@ -98,6 +102,10 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   RCLCPP_INFO(logger_, "Control loop execution time: %ld [ms]", duration);
 #endif
 
+  if (optimal_path_pub_->get_subscription_count() > 0) {
+    publish_optimal_path(optimizer_.getOptimizedTrajectory());
```
SteveMacenski, line 106: `getOptimizedTrajectory` involves populating the return each call https://github.com/ros-navigation/navigation2/blob/1c56c7a532a8df9783c742961e341542e1e609e6/nav2_mppi_controller/src/optimizer.cpp#L369-L384. Instead, you should include this publisher into the `trajectory visualizer` class, and publish like you do here, in the `visualize()` member so that (a) you call this function once and (b) so that all the visualization work is in its namesake class

SteveMacenski (changes requested):

SteveMacenski:
This also requires test coverage

nav2_mppi_controller/src/trajectory_visualizer.cpp
```diff
@@ -28,6 +28,7 @@ void TrajectoryVisualizer::on_configure(
   trajectories_publisher_ =
     node->create_publisher<visualization_msgs::msg::MarkerArray>("/trajectories", 1);
   transformed_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("transformed_global_plan", 1);
+  optimal_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("/local_plan", 1);
```
SteveMacenski, line 31:
```suggestion
  optimal_path_pub_ = node->create_publisher<nav_msgs::msg::Path>("optimal_trajectory", 1);
```

nav2_mppi_controller/src/trajectory_visualizer.cpp
```diff
@@ -81,6 +85,8 @@ void TrajectoryVisualizer::add(
   for (size_t i = 0; i < size; i++) {
     add_marker(i);
   }
+
+  optimal_traj_ = std::make_unique<xt::xtensor<float, 2>>(trajectory);
```
SteveMacenski, line 89: You could populate the path msg here like we do with `points_`

nav2_mppi_controller/src/trajectory_visualizer.cpp
```diff
@@ -111,14 +117,21 @@ void TrajectoryVisualizer::reset()
...
+void TrajectoryVisualizer::visualize(
+  const nav_msgs::msg::Path & plan,
+  const builtin_interfaces::msg::Time & cmd_stamp)
 {
   if (trajectories_publisher_->get_subscription_count() > 0) {
     trajectories_publisher_->publish(std::move(points_));
   }
 
+  if (optimal_path_pub_->get_subscription_count() > 0) {
+    publish_optimal_path(cmd_stamp);
```
SteveMacenski, line 132: Then this just becomes `optimal_path_pub_->publish(optimal_path);`

nav2_mppi_controller/src/trajectory_visualizer.cpp
SteveMacenski: It could still be another function for populating it, but call it here

SteveMacenski (changes requested):
LGTM, though its not compiling (that needs to be fixed, obviously!). One more change for making it consistent with the existing code and then we're good to go

SteveMacenski:
@alyquantillion please rebase or pull in main so that CI turns over. There was an API update in an upstream library that needs to be propagated here to build

nav2_mppi_controller/src/trajectory_visualizer.cpp
```diff
@@ -81,6 +87,8 @@ void TrajectoryVisualizer::add(
   for (size_t i = 0; i < size; i++) {
     add_marker(i);
```
SteveMacenski, line 88: The `for` loop is here, you can add in `populate_optimal_path` with `i` so we don't iterate twice

nav2_mppi_controller/src/trajectory_visualizer.cpp
```diff
@@ -127,4 +140,33 @@ void TrajectoryVisualizer::visualize(const nav_msgs::msg::Path & plan)
...
+  optimal_path_->header.stamp = cmd_stamp;
+  optimal_path_->header.frame_id = frame_id_;
+
+  for (size_t i = 0; i < size; i++) {
+    // create new pose for the path
+    geometry_msgs::msg::PoseStamped pose_stamped;
+    pose_stamped.header.frame_id = frame_id_;
+
+    // position & orientation
+    pose_stamped.pose = utils::createPose(optimal_traj(i, 0), optimal_traj(i, 1), 0.0);
```
SteveMacenski, line 161: This is already done in `add_marker` - would it be good to move the population of the path into that lambda? That would reduce the code duplication and looping over the data structure twice

alyquantillion: Indeed. `populate_optimal_path` does not exist anymore, everything is directly done in `add` and the lambda

SteveMacenski (approved):

SteveMacenski:
Thanks for iterating, looks perfect to me!

PR #4643, opened 2024-08-23
Various small CMake fixes.

clalancette:
---

## Basic Info

| ------ | ----------- |

---

Just some minor fixes to various CMake problems found throughout the repository as I was reviewing things in following up for #4357.

None needed.

---

None needed.

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/CMakeLists.txt
+++ b/nav2_smac_planner/CMakeLists.txt
@@ -1,8 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 project(nav2_smac_planner)
 
-set(CMAKE_BUILD_TYPE Release) #Debug, Release
-
 find_package(ament_cmake REQUIRED)
 find_package(ament_index_cpp REQUIRED)
 find_package(angles REQUIRED)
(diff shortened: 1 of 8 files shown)


nav2_smac_planner/CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 project(nav2_smac_planner)
 
-set(CMAKE_BUILD_TYPE Release) #Debug, Release
```
SteveMacenski, line 4: We need this for the optimization that it provides, this package in particular is release-necessity for good performance

clalancette: Yes, but I believe that `nav2_common` already does this, in a nicer way:

https://github.com/ros-navigation/navigation2/blob/6bf9d68af4c485c11947114e0e91ea5b0b5efc27/nav2_common/cmake/nav2_package.cmake#L21-L28

SteveMacenski: I think about it a bit differently, but perhaps you're right I shouldn't make multiple gates for folks to get through and know about to compile with debug if they want to

Issue #4618, opened 2024-08-14
The default_server_timeout parameter in bt_navigator is always overridden by the server_timeout port in each BT action node.


alanxuefei:
## Bug report

The `default_server_timeout` parameter in `bt_navigator` is always overridden by the `server_timeout` port in each BT action node. 

As a result, the `default_server_timeout` never takes effect because the `server_timeout` in each BT action node is mandatory and supersedes the default value.

Please refer to
https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp#L185

The code

```
    server_timeout_ = config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
    getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
```

may be changed as follows:

```cpp
if (!getInput<std::chrono::milliseconds>("server_timeout", server_timeout_)) {
    server_timeout_ = config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
}
```
   

**Required Info:**

- **Operating System:**
  - Ubuntu 22.04
- **ROS2 Version:**
  - Humble binaries
- **Version or commit hash:**
  - Main
- **DDS implementation:**
  - Fast-RTPS

#### Steps to reproduce issue

1. Create a Behavior Tree (BT) action node in a Nav2 configuration file.
2. Set a default `server_timeout` value in the corresponding YAML file.
3. Execute the BT with the action node.
4. Observe that the default value of `server_timeout` from the YAML file is not applied.

```cpp
// Code in bt_action_node.hpp:
server_timeout_ = config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
```

#### Expected behavior

The default value of `server_timeout` specified in the YAML file should be applied unless overridden by a specific BT action node configuration.

#### Actual behavior

The default value of `server_timeout` from the YAML file is never applied because `server_timeout_` is required to be explicitly set for each BT action node, which overrides the default.

#### Additional information

This bug prevents the default `server_timeout` value specified in the YAML configuration from being used in BT action nodes. As a result, users must manually set `server_timeout` for each node, leading to potential inconsistencies and additional configuration overhead.
```

SteveMacenski:
[...]

PR #4649, opened 2024-08-25
Fix: Ensure `server_timeout` Respects Default Values Using `getInputPortOrBlackboard()`

alanxuefei:
## Basic Info

| Info | Details |
| ------ | ----------- |
| **Ticket(s) this addresses** | #4618 |
| **Primary OS tested on** | Ubuntu 22.04 |
| **Robotic platform tested on** | Gazebo simulation, Turtlebot3 hardware |
| **Does this PR contain AI generated software?** | No |

---

- Updated `bt_action_node`, `bt_cancel_action_node`, and `bt_service_node` to use `getInputPortOrBlackboard()` for initializing `server_timeout`.
- Ensured that the `server_timeout` parameter now respects the default value from the blackboard if not explicitly set in the BT action node.
- This change addresses the issue where the `default_server_timeout` was always overridden, leading to more consistent behavior and reducing configuration overhead.

- No new parameters were added, so no updates to default configs are necessary.

---

- Further testing on different robotic platforms to ensure compatibility.
- Potential optimizations in the BT nodes to enhance performance.
- Documentation updates in case of additional related changes in the future.

---

#### For Maintainers: 
- [ ] Check that any new plugins are added to the plugins page
- [ ] If BT Node, Additionally: add to BT's XML index of nodes for Groot, BT package's readme table, and BT library lists

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -58,9 +58,7 @@ public:
     // Get the required items from the blackboard
     auto bt_loop_duration =
       config().blackboard->template get<std::chrono::milliseconds>("bt_loop_duration");
-    server_timeout_ =
-      config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
-    getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    getInputOrBlackboard("server_timeout", server_timeout_);
     wait_for_service_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("wait_for_service_timeout");
 
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_cancel_action_node.hpp
@@ -56,9 +56,7 @@ public:
     callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
 
     // Get the required items from the blackboard
-    server_timeout_ =
-      config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
-    getInput<std::chrono::milliseconds>("server_timeout", server_timeout_);
+    getInputOrBlackboard("server_timeout", server_timeout_);
     wait_for_service_timeout_ =
       config().blackboard->template get<std::chrono::milliseconds>("wait_for_service_timeout");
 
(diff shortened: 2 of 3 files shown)


SteveMacenski (approved):

SteveMacenski:
@alanxuefei isn't this still true?

> We'd have the change all the BT nodes not to have defaults in that case then so that getInput would return false when unset to use the global timeout.

alanxuefei:
> @alanxuefei isn't this still true?
> 
> > We'd have the change all the BT nodes not to have defaults in that case then so that getInput would return false when unset to use the global timeout.

Yes, the statement is correct. 
 
In the default behavior trees (e.g., [navigate_to_pose_w_replanning_goal_patience_and_recovery.xml](https://github.com/ros-navigation/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_goal_patience_and_recovery.xml)), there is no `server_timeout` parameter.

After second analysis, the original code actually works:
```
// Check ROS2 parameter.
server_timeout_ = config().blackboard->template get<std::chrono::milliseconds>("server_timeout");
// Check BT port (if the BT port is unset, server_timeout_ remains unchanged).    
```

However, the new code is more concise and efficient (inline):
```
template<typename T> inline
bool getInputPortOrBlackboard(
  const BT::TreeNode & bt_node,
  const BT::Blackboard & blackboard,
  const std::string & param_name,
  T & value)
{
  if (bt_node.getInput<T>(param_name, value)) {       //check BT port first
    return true;
  }
  if (blackboard.get<T>(param_name, value)) {           // check ROS2 param
    return true;
  }
  return false;
}
```

SteveMacenski:
So this doesn't change the behavior? I agree this is cleaner so happy to merge either way, but what sparked this PR / work if you weren't running into a problem (and do you still have a problem)?

alanxuefei:
> So this doesn't change the behavior? I agree this is cleaner so happy to merge either way, but what sparked this PR / work if you weren't running into a problem (and do you still have a problem)?

We mistakenly thought that once a BT port is declared in a BT node, it must also be included in the XML file. In fact, if the port is not defined in the XML file, `getInput` will return `false` without causing any system errors. For `server_timeout`, we can simply remove it from xml and it will use default value.

SteveMacenski:
So we're good to merge this?

alanxuefei:
Since it is a base function for each BT node, it's good to double-check its functionality. The tests have passed, and the logic has been verified. I believe it's ready to be merged.

Issue #4674, opened 2024-09-12
Option to disable publishZeroVelocity in the controller server


reinzor:
#### Feature description

Option to disable [`publishZeroVelocity` in the controller server](https://github.com/ros-navigation/navigation2/blob/7eb47d84012dd9e25988ce5c9c269280e6ee3dd1/nav2_controller/src/controller_server.cpp#L720).

#### Implementation considerations

For our use case, we would like to start a new path when driving so our end state is not necessarily standing still. We would like to have full control of the command velocity references within our controller. Related issues in move base flex (ros1 implementation that we were using):
  - https://github.com/magazino/move_base_flex/issues/195
  - https://github.com/magazino/move_base_flex/issues/109

Would it be possible to register a new parameter for the `controller_server`, e.g. `publish_zero_velocity` that defaults to `true` and use this parameter in an `if` statement for [this line](https://github.com/ros-navigation/navigation2/blob/7eb47d84012dd9e25988ce5c9c269280e6ee3dd1/nav2_controller/src/controller_server.cpp#L731)?

Thanks!

reinzor:
I think we only do not want to publish the zero velocity. If we name the parameter `publish_zero_velocity` and this would change the reset behavior, that is not really expected as a user. So I would suggest only skipping this actual zero velocity publishing. 

Maybe we should refactor it as follows by creating a new function `onGoalExit` that is called when a goal exits, so the current occurrences of `publishZeroVelocity`. In this function we call optionally `publishZeroVelocity` and we move the reset controller logic to this function as well. Would that be an idea?

In our use case we are sending path segments to the controller where every path segment is a new goal for the controller. The last point of our path segment is not necessarily a zero velocity path point, this is currently assumed by the controller server.

SteveMacenski:
>  a new function onGoalExit

Who would register this - the controller plugins? In that case, we're just kicking the can down the the pipeline and each controller needs to implement that same parameterization for what is a 'system' behavior. I don't dislike the idea of making it user set, but I don't see the path to do so that doesn't involve a new plugin class (which may be overkill for a bool param unless you think we may need to do additional things in the future). 

> In our use case we are sending path segments to the controller where every path segment is a new goal for the controller. The last point of our path segment is not necessarily a zero velocity path point, this is currently assumed by the controller server.

[...]

PR #4675, opened 2024-09-13
feat(controller-server): `publish_zero_velocity` parameter

reinzor:
For optionally publishing a zero velocity command reference on goal exit. Publishing a zero velocity is not desired when we are following consecutive path segments that end with a velocity.

#4674


nav2_controller/src/controller_server.cpp
```diff
@@ -746,6 +748,13 @@ void ControllerServer::publishZeroVelocity()
   velocity.header.frame_id = costmap_ros_->getBaseFrameID();
   velocity.header.stamp = now();
   publishVelocity(velocity);
+}
+
+void ControllerServer::onGoalExit()
+{
+  if (publish_zero_velocity_) {
+    publishZeroVelocity();
+  }
```
chanhhoang99, line 757: I have 1 question about this option.
What would happend if the goal_checker kick in when we have such a option ?
As far as I know when we are in the goal_checker tolerance, controller will publish zero velocity. Does this place an impact on that part ?

nav2_controller/include/nav2_controller/controller_server.hpp
```diff
@@ -176,6 +176,10 @@ class ControllerServer : public nav2_util::LifecycleNode
    * @brief Calls velocity publisher to publish zero velocity
    */
   void publishZeroVelocity();
+  /**
```
SteveMacenski, line 179: Space between dogyxen and last method

nav2_controller/src/controller_server.cpp
```diff
@@ -298,7 +300,7 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
    */
   costmap_ros_->deactivate();
 
-  publishZeroVelocity();
+  onGoalExit();
```
SteveMacenski, line 303: I think on deactivation, `publishZeroVelocity` is the right behavior, no? We don't want our robot still moving at all if this is being shut down

SteveMacenski (changes requested):
Just the usual migration + configuration guide update to introduce the new parameter.

nav2_controller/src/controller_server.cpp
SteveMacenski: > As far as I know when we are in the goal_checker tolerance, controller will publish zero velocity

That is ideally the case, but not always implemented if someone's plugin is not well thought out. This is a protection mechanism. Its better to publish a zero twice than not at all - especially for robots that don't have in their firmware to stop executing a command after a certain timeout. And trust me, more robot platforms out there than you would think will continue to drive indefinitely with an extraneous command than you think (its a little scary).

reinzor: > controller will publish zero velocity

Only if implemented. This is exactly what we're trying to avoid here by making this behavior optional in the controller server.

nav2_controller/include/nav2_controller/controller_server.hpp
reinzor: This is not consistent throughout the header file. I can add this if you want.

nav2_controller/src/controller_server.cpp
reinzor: I altered the PR; but maybe we should publish a zero velocity on the `on_deactivate` method of the velocity publisher.

nav2_controller/src/controller_server.cpp
```diff
@@ -299,6 +301,9 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
   costmap_ros_->deactivate();
 
   publishZeroVelocity();
+  for (auto & controller : controllers_) {
```
SteveMacenski, line 304: They have already been deactivated in the lines above, this is not necessary

nav2_controller/src/controller_server.cpp
```diff
@@ -513,42 +518,42 @@ void ControllerServer::computeControl()
...
-    publishZeroVelocity();
+    onGoalExit();
     std::shared_ptr<Action::Result> result = std::make_shared<Action::Result>();
     result->error_code = Action::Result::PATIENCE_EXCEEDED;
     action_server_->terminate_current(result);
     return;
   } catch (nav2_core::InvalidPath & e) {
     RCLCPP_ERROR(this->get_logger(), "%s", e.what());
-    publishZeroVelocity();
+    onGoalExit();
```
SteveMacenski, line 556: Missed one in L563

SteveMacenski (approved):
LGTM. Add to the migration guide the change + new parameter and add to the controller server's configuration guide the new parameter for documentation for future lookers

Edit: Sorry that wasn't supposed to be an approval, but the remaining items are trivial

SteveMacenski (approved):

reinzor:
@SteveMacenski can we backport this on Jazzy? I can alter the change so that we maintain ABI compatible.

SteveMacenski:
@reinzor sure! An altered version would need to be done, but should be easy enough

Timple:
This feature has been verified in production :slightly_smiling_face: 

Would a Jazzy bloom release be in order? As the last one was from August.

SteveMacenski:
It’s on my schedule for when I’m back from Europe for ROSCon / ROS-I. Unfortunately my release process involves some testing and builds that I need my full screen setup on (really hard to do it all on a 12” screen).

PR #4676, opened 2024-09-13
Improvements in RemoveInCollisionGoals and adjacent features

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu, |
| Robotic platform tested on | custom gazebo simulation |

---

* Change GetCosts to use PoseStamped instead of Pose2D so that we can use the frame_id to transform the pose in the costmap frame
* Added consider_unknown_as_obstacle to control whether NO_INFORMATION is treated as "in collision" or not
* Make GetCosts service return 255.0 instead of -1.0 if the coords are out of bounds
* fix rviz costmap tool to use the global fixed frame of rviz

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
+++ b/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
@@ -37,6 +37,7 @@ void RemoveInCollisionGoals::on_tick()
   getInput("use_footprint", use_footprint_);
   getInput("cost_threshold", cost_threshold_);
   getInput("input_goals", input_goals_);
+  getInput("consider_unknown_as_obstacle", consider_unknown_as_obstacle_);
 
   if (input_goals_.empty()) {
     setOutput("output_goals", input_goals_);
@@ -47,22 +48,32 @@ void RemoveInCollisionGoals::on_tick()
   request_->use_footprint = use_footprint_;
 
   for (const auto & goal : input_goals_) {
-    geometry_msgs::msg::Pose2D pose;
-    pose.x = goal.pose.position.x;
-    pose.y = goal.pose.position.y;
-    pose.theta = tf2::getYaw(goal.pose.orientation);
-    request_->poses.push_back(pose);
+    request_->poses.push_back(goal);
   }
 }
 
 BT::NodeStatus RemoveInCollisionGoals::on_completion(
   std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
 {
+  for (size_t i = 0; i < input_goals_.size(); ++i) {
+    double yaw = tf2::getYaw(input_goals_[i].pose.orientation);
+    RCLCPP_INFO(
+      node_->get_logger(),
+      "Goal %ld: x: %f, y: %f, yaw: %f, costs: %f", i,
+      input_goals_[i].pose.position.x, input_goals_[i].pose.position.y, yaw, response->costs[i]);
+  }
+
   Goals valid_goal_poses;
   for (size_t i = 0; i < response->costs.size(); ++i) {
-    if (response->costs[i] < cost_threshold_) {
+    if ((response->costs[i] == 255 && !consider_unknown_as_obstacle_) || response->costs[i] < cost_threshold_) {
       valid_goal_poses.push_back(input_goals_[i]);
     }
+  }
+  // Warn if all goals have been removed
+  if (valid_goal_poses.empty()) {
+    RCLCPP_WARN(
+      node_->get_logger(),
+      "All goals are in collision and have been removed from the list");
   }
   setOutput("output_goals", valid_goal_poses);
   return BT::NodeStatus::SUCCESS;
--- a/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
+++ b/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
@@ -65,7 +65,9 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
 
   Goals valid_goal_poses;
   for (size_t i = 0; i < response->costs.size(); ++i) {
-    if ((response->costs[i] == 255 && !consider_unknown_as_obstacle_) || response->costs[i] < cost_threshold_) {
+    if ((response->costs[i] == 255 && !consider_unknown_as_obstacle_) ||
+      response->costs[i] < cost_threshold_)
+    {
       valid_goal_poses.push_back(input_goals_[i]);
     }
   }
--- a/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
+++ b/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
@@ -55,14 +55,6 @@ void RemoveInCollisionGoals::on_tick()
 BT::NodeStatus RemoveInCollisionGoals::on_completion(
   std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
 {
-  for (size_t i = 0; i < input_goals_.size(); ++i) {
-    double yaw = tf2::getYaw(input_goals_[i].pose.orientation);
-    RCLCPP_INFO(
-      node_->get_logger(),
-      "Goal %ld: x: %f, y: %f, yaw: %f, costs: %f", i,
-      input_goals_[i].pose.position.x, input_goals_[i].pose.position.y, yaw, response->costs[i]);
-  }
-
   Goals valid_goal_poses;
   for (size_t i = 0; i < response->costs.size(); ++i) {
     if ((response->costs[i] == 255 && !consider_unknown_as_obstacle_) ||
@@ -71,9 +63,9 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
       valid_goal_poses.push_back(input_goals_[i]);
     }
   }
-  // Warn if all goals have been removed
+  // Inform if all goals have been removed
   if (valid_goal_poses.empty()) {
-    RCLCPP_WARN(
+    RCLCPP_INFO(
       node_->get_logger(),
       "All goals are in collision and have been removed from the list");
   }
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -862,7 +862,8 @@ void Costmap2DROS::getCostsCallback(
           pose_transformed.pose.position.y, yaw, footprint));
     } else {
       RCLCPP_DEBUG(
-        get_logger(), "Received request to get cost at point (%f, %f)", pose_transformed.pose.position.x,
+        get_logger(), "Received request to get cost at point (%f, %f)",
+          pose_transformed.pose.position.x,
         pose_transformed.pose.position.y);
 
       // Get the cost at the map coordinates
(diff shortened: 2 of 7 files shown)


tonynajjar:
There is still one thing I'd like to modify in this PR but would like your opinion because it might affect a lot of things upstream. About the implementation of [footprintCost](https://github.com/ros-navigation/navigation2/blob/b71cccd8c47d4cb472db1c3a509af3821336933a/nav2_costmap_2d/src/footprint_collision_checker.cpp#L48), if the footprint is partially outside of bounds, then the cost is LETHAL_OBSTACLE.
However, in the case of track_unknown_space = false (considered as free space), shouldn't footprintCost also take this into consideration and not return LETHAL_OBSTACLE?

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -47,23 +48,35 @@ void RemoveInCollisionGoals::on_tick()
...
+    request_->poses.push_back(goal);
   }
 }
 
 BT::NodeStatus RemoveInCollisionGoals::on_completion(
   std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
 {
+  for (size_t i = 0; i < input_goals_.size(); ++i) {
+    double yaw = tf2::getYaw(input_goals_[i].pose.orientation);
+    RCLCPP_INFO(
```
SteveMacenski, line 60: Remove log or change to debug

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -47,23 +48,35 @@ void RemoveInCollisionGoals::on_tick()
...
-    if (response->costs[i] < cost_threshold_) {
+    if ((response->costs[i] == 255 && !consider_unknown_as_obstacle_) ||
+      response->costs[i] < cost_threshold_)
+    {
       valid_goal_poses.push_back(input_goals_[i]);
     }
   }
+  // Warn if all goals have been removed
+  if (valid_goal_poses.empty()) {
+    RCLCPP_WARN(
```
SteveMacenski, line 76: is that a warn or an info? That seems like expected behavior

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -835,26 +836,35 @@ void Costmap2DROS::getCostsCallback(
...
-    bool in_bounds = costmap->worldToMap(pose.x, pose.y, mx, my);
+    geometry_msgs::msg::PoseStamped pose_transformed;
+    transformPoseToGlobalFrame(pose, pose_transformed);
+    bool in_bounds = costmap->worldToMap(
+      pose_transformed.pose.position.x,
+      pose_transformed.pose.position.y, mx, my);
 
     if (!in_bounds) {
-      response->costs.push_back(-1.0);
+      response->costs.push_back(costmap->getDefaultValue());
```
SteveMacenski, line 846: Isn't unknown the better option here, if we're off the costmap? I don't think the default value applies where the map doesn't exist. Unknown seems rational to me since we really don't know anything

SteveMacenski (changes requested):
No major notes, just details. This doesn't appear to compile, so once this does its an easy merge

tonynajjar:
@SteveMacenski can you also comment on my question about footprintCost, I'd like to address it in this PR as well

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: Oops that slipped in

SteveMacenski:
Sorry, I missed that.

> if the footprint is partially in free space and partially outside of bounds, then the cost is LETHAL_OBSTACLE. 

If we can't convert world->map because out of bounds, that's an error case. For example, if we have a line segment of the footprint that's 1 meter long. Only the very last point is off the map, but there is a collision along its length at 0.5m from the starting corner. 

In this case, we can't raycast to find the collision (unless we march back along the ray to find when it starts being valid again, and collision check up to then). Imagine now that there was some non-lethal cost `154` along the edge and the corner went out of the map. Should we return unknown or 154? 

I'm open to changing behavior here, but its not clear to me even if we refactored this code to march as long as it was valid for (and then take unknown for the rest), what we'd return if any part of the edge is non-zero. 

> However, in the case of track_unknown_space = false (considered as free space), shouldn't footprintCost also take this into consideration and return FREE_SPACE?

`track_unknown_space` is about unknown as the background for the costmap (ie before receiving sensor measurements, should we consider the cost free (0) or unknown (255) to be not understood until some free space is raycasted out by sensor measurements), not off the costmap. I don't interpret `track_unknown_space` as applicable for off the map itself -- but I suppose that could be a not-so-unreasonable assumption to expand into.

However to do so, we need to answer questions like what I posed in the comment block above & how we handle these situations

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
tonynajjar: yes it's expected, I just found it useful to know when the output was empty to debug why the planner was throwing an error. But you're right, it can be a simple info

tonynajjar:
Okay it's more complicated than I anticipated. I'd like to find a way to solve my immediate problem without too much refactoring then. 
My use case is that in the BT I run RemoveInCollisionGoals both on poses both for the global and the local costmap.
I simply don't want RemoveInCollisionGoals to remove poses that are outside the local costmap (or at the border if using use_footprint)  because these poses are not clearly in collision -> it's unknown. We get more information about the local costmap cost once we get closer to them. 

Any recommendations for solving this without major refactorings? Let me know if I wasn't clear enough

nav2_rviz_plugins/src/costmap_cost_tool.cpp
```diff
@@ -96,21 +96,24 @@ void CostmapCostTool::callCostService(float x, float y)
 {
   // Create request for local costmap
   auto request = std::make_shared<nav2_msgs::srv::GetCosts::Request>();
-  geometry_msgs::msg::Pose2D pose;
-  pose.x = x;
-  pose.y = y;
+  geometry_msgs::msg::PoseStamped pose;
```
SteveMacenski, line 99: stamp?

SteveMacenski (changes requested):

SteveMacenski:
> Okay it's more complicated than I anticipated.

Yeah :/ The system wasn't designed with 'off the map' being a valid or highly considered case. They would have made an OFF_THE_MAP enum type if that was considered. 

> I don't want RemoveInCollisionGoals to remove these poses because they are not for sure in collision -> it's unknown.
For NO_INFORMATION, I added consider_unknown_as_obstacle as a workaround but for the footprint cost LETHAL I can't see a solution -> that's why I was proposing to change it.

So if you're off the map, you return NO_INFORMATION and you have handling for NO_INFORMATION to make that OK in this PR with a parameterization. If its LETHAL, that's still lethal to skip. I don't understand the issue, that seems to do what you want it to do? 

W.r.t. collision checking footprints and part is on the costmap and part is off the costmap that we previously discussed, I suppose that's why this is in the BT to execute on occasion as the task progresses? Once in proximity of that goal, it should be fully on the costmap again to proceed or skip based on its LETHAL or OK known-information state. I think this handles itself due to the rolling costmaps

tonynajjar:
> If its LETHAL, that's still lethal to skip. I don't understand the issue, that seems to do what you want it to do?

I **don't** want to skip this pose, the cost is LETHAL but only because that's what footprintCost returns for footprints at the border of the costmap when in reality the footprint is on FREE and "OFF_THE_MAP" -> I'd rather consider this as non-lethal

> W.r.t. collision checking footprints and part is on the costmap and part is off the costmap that we previously discussed, I suppose that's why this is in the BT to execute on occasion as the task progresses? Once in proximity of that goal, it should be fully on the costmap again to proceed or skip based on its LETHAL or OK known-information state. I think this handles itself due to the rolling costmaps


Exactly, the problem arises when all the close waypoints are removed because of LETHAL (rightfully so because of a real obstacle) and the only "valid" waypoint left is the one at the border of the costmap -> so this one is also removed -> empty list of poses is sent to ComputePathThroughPoses -> planner rightfully throws an error

SteveMacenski:
Is it possible to know before collision checking if the goals are off the map? We could have another `bool` like what this PR proposes in the BT node for how to interpret such cases. Either in the BT node:

```
  for (const auto & goal : input_goals_) {
    request_->poses.push_back(goal);
  }
```

Or in the service itself, we have some off-map designator only at this higher level BT. Perhaps change the message to allow the population of another field about the on-mapped-ness status. I don't love having `costs` and `XYZ`, so if we can combine those, that would be nice. The `costs` are in `float32` and not in the `0-255` that the cost map is stored in, so we could have some value we assign to off map (`-1`)

tonynajjar:
> Is it possible to know before collision checking if the goals are off the map

I see [isOutofBounds](https://github.com/ros-navigation/navigation2/blob/771eca41bb7b6ad26856c434dd79e48a399c4601/nav2_costmap_2d/src/layered_costmap.cpp#L127C22-L127C35) exists but that's only to check if a point is off the map not if a footprint is completely/partially off the map. Is there one? If so, then we can just use it in the service, we already do check that [here](https://github.com/ros-navigation/navigation2/pull/4676/files#diff-d3369b82718255857b744f9289b4e31a34924f72fb40212604b4aa8276b857f2R845) and that works well for point costs.

While that would solve my immediate issue, I don't think it's good design for the service to do its own on-mapped-ness check and return a cost different than what `footprintCostAtPose` and `getCost` would return -> given the same coordinates, a node using the service would get a different cost than another node using the costmap directly

SteveMacenski:
There is not one currently, we'd need to make one and expose it. The `in_bounds` check would handle some of the cases (others obviously not in the center is in the costmap but an edge isn't), though. So I suppose from the API level, we can use that to define how we want to communicate that out-of-mapped-ness, that also works for the point-costs. Then, we need to figure out the tech details of the collision checker object to do the same. At least it decouples the problem a bit. 

A couple of questions:
- How long are these requests and the number of poses within them?
- How far away are some of these goals off the map?
- Is it important that you understand the collision state of poses 2-3-4-5 poses down the line from the current task?

What I'm thinking is that you only feed into this the next `N` rather than all `M` poses, where you can make some guarantee that the next `N` will be within the rolling costmap's bounds based on regular spacing or known distances from the current origin. If you don't need to know the state of pose `N+5` or above, you could simply not check them until within relevant proximity that would promise it to be fully within the costmap's bounds. 

If you plan to mark partially-off-map poses as "OK" with lack of information, I figure this has the same behavioral outcomes with less than a third of the technical difficulty. We just need a new BT node to extract only the first `N` poses or pose within `N` distance of the current robot pose (or have that logic be part of the remove in collision goals BT node, which also seems rational to set a limit of where we want to check)

tonynajjar:
Thanks for the proposals, I can't justify spending much more time on this for now so I'll take it as accepted limitation and will found some hacky solution on my side. This can be merged from my side.

SteveMacenski (approved):

tonynajjar:
I personally don't really mind not having it in the jazzy binaries, I build a long-term fork.

SteveMacenski:
Ok! Just checking since this is one of the first `main` branch things that haven't been able to be backporte to `jazzy` :-)

PR #4684, opened 2024-09-20
fix(simple-action-server): info log instead of warn on cancel

reinzor:
Cancelling a goal is nominal behavior and therefore it should not log warning.


SteveMacenski (approved):
Makes sense to me!

reinzor:
@SteveMacenski can we backport this on Jazzy?

SteveMacenski:
Done

reinzor:
Great! Thanks @SteveMacenski for all your hard work and quick responses!

PR #4687, opened 2024-09-24
feat(controller-server): `publish_zero_velocity` parameter jazzy backport

reinzor:
Backport of https://github.com/ros-navigation/navigation2/pull/4675

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -63,6 +63,7 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
 
   declare_parameter("failure_tolerance", rclcpp::ParameterValue(0.0));
   declare_parameter("use_realtime_priority", rclcpp::ParameterValue(false));
+  declare_parameter("publish_zero_velocity", rclcpp::ParameterValue(true));
 
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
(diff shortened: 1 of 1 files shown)


nav2_controller/src/controller_server.cpp
```diff
@@ -719,16 +730,18 @@ void ControllerServer::publishVelocity(const geometry_msgs::msg::TwistStamped &
...
-  velocity.twist.angular.x = 0;
-  velocity.twist.angular.y = 0;
-  velocity.twist.angular.z = 0;
-  velocity.twist.linear.x = 0;
-  velocity.twist.linear.y = 0;
-  velocity.twist.linear.z = 0;
-  velocity.header.frame_id = costmap_ros_->getBaseFrameID();
-  velocity.header.stamp = now();
-  publishVelocity(velocity);
+  if (get_parameter("publish_zero_velocity").as_bool()) {
```
SteveMacenski, line 733: You can store the variable in the class, that's OK

nav2_controller/src/controller_server.cpp
```diff
@@ -292,7 +293,17 @@ ControllerServer::on_deactivate(const rclcpp_lifecycle::State & /*state*/)
    */
   costmap_ros_->deactivate();
 
-  publishZeroVelocity();
+  geometry_msgs::msg::TwistStamped velocity;
```
SteveMacenski, line 296:
```suggestion
  // Always publish a zero velocity when deactivating the controller server
  geometry_msgs::msg::TwistStamped velocity;
```

nav2_controller/src/controller_server.cpp
reinzor: According to https://www.ros.org/reps/rep-0009.html:

```
You cannot...
  Add new non-static data members, even if they are protected or private
```

Also the `get_parameter` is a map lookup so should be fast.

SteveMacenski: I suppose I haven't enforced that part of ABI, but perhaps I should be -- it would just make direct backporting _any new feature_ virtually impossible since they would likely have some parameter set tied to them. I enforce API stability and behavior stability (i.e. not changing defaults or fundamentally changing the behavior on a module mid-distribution), but I think there's an argument that I should by enforcing stricter ABI as well. Though, its never come up and no one's had an issue. 

This is fine.

SteveMacenski (approved):

PR #4692, opened 2024-09-25
Humble compatibility for watchdog mechanism

Aadi0902:
---

## Basic Info

| ------ | ----------- |

---

This PR adds a per source blocking watchdog mechanism, i.e. stop the robot if a source is not publishing at the expected rate (building on the existing source_timeout mechanism) for **humble**. This feature already exists on main following [PR 3880](https://github.com/ros-navigation/navigation2/pull/3880)

-

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/include/nav2_collision_monitor/pointcloud.hpp
+++ b/nav2_collision_monitor/include/nav2_collision_monitor/pointcloud.hpp
@@ -70,8 +70,9 @@ public:
    * @param curr_time Current node time for data interpolation
    * @param data Array where the data from source to be added.
    * Added data is transformed to base_frame_id_ coordinate system at curr_time.
+   * @return false if an invalid source should block the robot
    */
-  void getData(
+  bool getData(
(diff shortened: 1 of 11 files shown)


SteveMacenski:
This is not able to be backported since the API changes return types. Unfortunately this is why this was not backported originally.

SteveMacenski:
Closing due to aforementioned reason. If you can re-implement the feature in a backportable way (no API changes), I'm happy to review / merge that!

Aadi0902:
> Closing due to aforementioned reason. If you can re-implement the feature in a backportable way (no API changes), I'm happy to review / merge that!

@SteveMacenski  quick question: though the proposed changes do change function definitions, what are the implications of changing return type from void to bool? technically it should still support any use case where the function was used as void right?

SteveMacenski:
If anyone else created derivative classes of those classes, they would stop compiling after a Humble sync. Additionally a struct was changed with similar consequences

PR #4699, opened 2024-10-01
Adding disengagement threshold to rotation shim controller

SteveMacenski:
https://github.com/ros-navigation/navigation2/pull/4690

@sandeepdutta please review! I updated the default for my parameter to be half of the threshold (22.5) instead of a third (15), but you can change that in your own config file if you prefer that behavior. 

For the backport, I'll have to make the default 45 degrees so we don't change the default behavior, but you can again set it to 1/3 (or otherwise) in your configuration file to change that behavior for your application on Humble


SteveMacenski:
Docs PR: https://github.com/ros-navigation/docs.nav2.org/pull/595

sandeepdutta:
Hi Steve,

I have merged your changes & ran it on the robot, when I set the angular_disengage_threshold it works .
It does engage at the higher angle and disengages when the disengage threshold is hit.

Thank you very much for accepting this change, would be great if I can get it on humble so I will not need to
maintain my branch.

Best Regards
Sandeep

SteveMacenski:
Backport bot should take care of this

PR #4706, opened 2024-10-04
fix(nav2_theta_star_planner): Planner Server Crashing When Planning Outside Map Bounds with Theta* on Humble

brayanpa:
This PR resolves the issue discussed in: https://github.com/ros-navigation/navigation2/issues/4694
The planner server was crashing when attempting to plan paths outside the map bounds using the Theta* planner on Humble.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/issues/4694 |
| Primary OS tested on | Ubuntu 20.04.6 LTS |
| Robotic platform tested on | planner playground, Gary Robot hardware |

---

## Description

* This change returns an empty path when either the goal or start coordinates are outside the map bounds, preventing the planner server from crashing.


SteveMacenski (approved):

SteveMacenski:
Thanks!

Issue #4704, opened 2024-10-03
Log discrepancy when setting MPPI parameters dynamically


PlatHum:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Humble
- Version or commit hash:
  - 1.1.16-1jammy.20240830.231501

- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue

Using `ros2 param set` command to set MPPI parameters triggers a NOT FOUND warning, although the command returns 'Set parameter successful'.
```
## General form:
ros2 param set /{$controller_server_node_name} {$mppi_controller_name}.{$parameter} {$value}

##For example:
ros2 param set /controller_server FollowPathMPPI.AckermannConstraints.min_turning_r 1.0
##or
ros2 param set /controller_server FollowPathMPPI.gamma 1.0
##or
 ros2 param set /controller_server FollowPathMPPI.ConstraintCritic.cost_weight 1.0

```

#### Expected behavior
```
### If parameter exists, is dynamic and changed successfully
##On set parameter terminal:
Set parameter successful
##On controller_server terminal:
[controller_server-28] [WARN] [controller_server]: {$Something that indicates success in changing parameter}
[controller_server-28] [INFO] [controller_server]: Optimizer reset
```
```
### If parameter does not exist, or is not dynamic or not changed successfully
##On set parameter terminal:
Setting parameter failed: {$reason_for_failure}
##On controller_server terminal:
[controller_server-28] [WARN] [controller_server]: {$Something that indicates failure in changing parameter}
[controller_server-28] [INFO] [controller_server]: Optimizer reset

```

#### Actual behavior
```
##On set parameter terminal:
Set parameter successful

##On controller_server terminal:
[controller_server-28] [WARN] [controller_server]: Parameter FollowPathMPPI.AckermannConstraints.min_turning_r not found
[controller_server-28] [INFO] [controller_server]: Optimizer reset
```

#### Additional information
Seems like [`parameters_handler`](https://github.com/ros-navigation/navigation2/blob/c280e3c6f7f5a2e2e83c75a85c5ab9e40bf51b7b/nav2_mppi_controller/src/parameters_handler.cpp#L68) always returns success even though a parameter might not be found.

----

SteveMacenski:
[...]

PR #4707, opened 2024-10-04
Fix : Log discrepancy when setting MPPI parameters dynamically

PlatHum:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4704 |

---

* If `ParametersHandler::dynamicParamsCallback` cannot find the dynamically set parameter name in its parameters list, set the `ParametersResult`  output message's `successful` field to `false` instead of always `true`.

---

*  The `rcl_interfaces/msg/SetParametersResult` message returned also has a `reason` field that could be populated with the reason for failure. This could be the name of the parameter not found or the full [already printed warning](https://github.com/PlatHum/navigation2/blob/e34634c97fc140003c83013682fec39fe9f3ec58/nav2_mppi_controller/src/parameters_handler.cpp#L70).

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/parameters_handler.cpp
+++ b/nav2_mppi_controller/src/parameters_handler.cpp
@@ -53,7 +53,7 @@ ParametersHandler::dynamicParamsCallback(
 {
   rcl_interfaces::msg::SetParametersResult result;
   std::lock_guard<std::mutex> lock(parameters_change_mutex_);
-  bool success =true;
+  bool success = true;
 
   for (auto & pre_cb : pre_callbacks_) {
     pre_cb();
--- a/nav2_mppi_controller/src/parameters_handler.cpp
+++ b/nav2_mppi_controller/src/parameters_handler.cpp
@@ -53,6 +53,7 @@ ParametersHandler::dynamicParamsCallback(
 {
   rcl_interfaces::msg::SetParametersResult result;
   std::lock_guard<std::mutex> lock(parameters_change_mutex_);
+  bool success =true;
 
   for (auto & pre_cb : pre_callbacks_) {
     pre_cb();
@@ -67,6 +68,7 @@ ParametersHandler::dynamicParamsCallback(
       callback->second(param);
     } else {
       RCLCPP_WARN(logger_, "Parameter %s not found", param_name.c_str());
+      success = false;
     }
   }
 
@@ -74,7 +76,7 @@ ParametersHandler::dynamicParamsCallback(
     post_cb();
   }
 
-  result.successful = true;
+  result.successful = success;
   return result;
 }
 


SteveMacenski (approved):

SteveMacenski:
@PlatHum can you rebase on `main`? Its failing with a bunch of tests and I'm not sure why. We did update the cache recently so I'm curious if that's why. If that doesn't work and this is still failing several tests, there may be downstream implications of this PR that we didn't consider

PlatHum:
> @PlatHum can you rebase on or pull in `main`? Its failing with a bunch of tests and I'm not sure why. We did update the cache recently so I'm curious if that's why. If that doesn't work and this is still failing several tests, there may be downstream implications of this PR that we didn't consider. I just retriggered the job for a second time, lets see what it does.

I think this reveals a previously unkown/uncaught error with dynamic parameter setting, perhaps? Or at least a problem with the tests performed. Previously, setting parameters always returned success even if the parameter was not set successfully or not found, which made it seem that all was fine.

From the logs of one of the tests as an example:
```
    [controller_server-8] [WARN] [1728127317.967062716] [controller_server]: Parameter controller_server.verbose not found
    [controller_server-8] [INFO] [1728127317.969753939] [controller_server]: Optimizer reset
    [controller_server-8] [ERROR] [1728127317.969881341] [controller_server]: Caught exception in callback for transition 13
    [controller_server-8] [ERROR] [1728127317.969913634] [controller_server]: Original error: parameter 'controller_server.verbose' could not be set: 
    [controller_server-8] [WARN] [1728127317.969939448] [controller_server]: Error occurred while doing error handling.
    [controller_server-8] [FATAL] [1728127317.969949455] [controller_server]: Lifecycle node controller_server does not have error state implemented
    [lifecycle_manager-17] [ERROR] [1728127317.970300630] [lifecycle_manager_navigation]: Failed to change state for node: controller_server
    [lifecycle_manager-17] [ERROR] [1728127317.970356458] [lifecycle_manager_navigation]: Failed to bring up all requested nodes. Aborting bringup.
```
For this one in particular, it seems like `ParametersHandler::dynamicParamsCallback` does not recognize `verbose` as one of its parameters, which [it should](https://github.com/PlatHum/navigation2/blob/07a28e5701f6331a65daadd021f339e353b2df5b/nav2_mppi_controller/src/parameters_handler.cpp#L47).

SteveMacenski:
Please look into this - this is causing not only the individual unit tests to fail, but also the full system tests that use MPPI, so this PR would not allow MPPI to work, it does not appear from the other failures. Have you tested MPPI after making these changes?

SteveMacenski:
@PlatHum would you agree that https://github.com/ros-navigation/navigation2/pull/4711 supersedes?

PlatHum:
> @PlatHum would you agree that #4711 supersedes?

Yea. @aosmw figured out that the "verbose" parameter needed to be declared before the dynamic parameter handler. If it passed the tests, that fixes it. 
Can/Should I close this PR or is it you that takes care of closing?

SteveMacenski:
I can close - but we appreciate your proactivity and help! I hope to merge another one of your PRs soon :smile:

PR #4711, opened 2024-10-07
mppi parameters_handler: Improve verbose handling (#4704)

aosmw:
The "verbose" parameter of the parameters_handler is a special case that needs registration before the
dynamic parameter handler callback is registered.

In verbose mode make the parameter handler info/warn/debug messages more expressive.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #4704 |
| Robotic platform tested on | AOS kelpie |

---

Improve the verbose handling of the mppi parameter_handler class and make it more verbose to
help distinguish between ParameterType::Dynamic and ParameterType::Static parameters.

Key point is to treat the verbose parameter of the mppi parameter_handler class
as a special case and register it *BEFORE* the dynamic parameter handler callback is
registered.

Other parameters may be "ParameterType::Static" on purpose, so a decision is required 
about whether or not to return false from the dynamicParamsCallback is required.

I left it as false as per initial PR. 

---

Make decision about returning false and remove commented out return true.

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
@@ -252,7 +252,7 @@ void ParametersHandler::setParamCallback(
     const rclcpp::Parameter & param, rcl_interfaces::msg::SetParametersResult & result) {
       std::string reason = "Rejected change to static parameter: " + std::to_string(param);
       result.successful = false;
-      if (result.reason.empty()) {
+      if (!result.reason.empty()) {
         result.reason += "\n";
       }
       result.reason += reason;
--- a/nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
@@ -211,7 +211,7 @@ void ParametersHandler::getParam(
 
   if (param_type == ParameterType::Dynamic) {
     if (verbose_) {
-      RCLCPP_INFO(node->get_logger(), "setDynamicParamCallback for %s", name.c_str());
+      RCLCPP_DEBUG(node->get_logger(), "setDynamicParamCallback for %s", name.c_str());
     }
     setDynamicParamCallback(setting, name);
   } else {
--- a/nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
@@ -100,9 +100,11 @@ public:
     * @brief Set a parameter to a dynamic parameter callback
     * @param setting Parameter
     * @param name Name of parameter
+    * @param param_type Type of parameter (dynamic or static)
     */
   template<typename T>
-  void setDynamicParamCallback(T & setting, const std::string & name);
+  void setDynamicParamCallback(
+    T & setting, const std::string & name, ParameterType param_type = ParameterType::Dynamic);
 
   /**
     * @brief Get mutex lock for changing parameters
@@ -208,19 +210,7 @@ void ParametersHandler::getParam(
     node, name, rclcpp::ParameterValue(default_value));
 
   setParam<ParamT>(setting, name, node);
-
-  if (param_type == ParameterType::Dynamic) {
-    if (verbose_) {
-      RCLCPP_DEBUG(node->get_logger(), "setDynamicParamCallback for %s", name.c_str());
-    }
-    setDynamicParamCallback(setting, name);
-  } else {
-    if (verbose_) {
-      RCLCPP_DEBUG(
-        node->get_logger(), "ParameterType::Static therefore no setDynamicParamCallback for %s",
-        name.c_str());
-    }
-  }
+  setDynamicParamCallback(setting, name, param_type);
 }
 
 template<typename ParamT, typename SettingT, typename NodeT>
@@ -233,13 +223,14 @@ void ParametersHandler::setParam(
 }
 
 template<typename T>
-void ParametersHandler::setDynamicParamCallback(T & setting, const std::string & name)
+void ParametersHandler::setDynamicParamCallback(
+  T & setting, const std::string & name, ParameterType param_type)
 {
   if (get_param_callbacks_.find(name) != get_param_callbacks_.end()) {
     return;
   }
 
-  auto callback = [this, &setting, name](const rclcpp::Parameter & param) {
+  auto dynamic_callback = [this, &setting, name](const rclcpp::Parameter & param) {
       setting = as<T>(param);
 
       if (verbose_) {
@@ -247,10 +238,17 @@ void ParametersHandler::setDynamicParamCallback(T & setting, const std::string &
       }
     };
 
-  addDynamicParamCallback(name, callback);
+  auto static_callback = [this, &setting, name](const rclcpp::Parameter & param) {
+      if (verbose_) {
+        RCLCPP_DEBUG(logger_, "Ignoring change to static parameter: %s",
+          std::to_string(param).c_str());
+      }
+    };
 
-  if (verbose_) {
-    RCLCPP_INFO(logger_, "Dynamic Parameter added %s", name.c_str());
+  if (param_type == ParameterType::Dynamic) {
+    addDynamicParamCallback(name, dynamic_callback);
+  } else {
+    addDynamicParamCallback(name, static_callback);
   }
 }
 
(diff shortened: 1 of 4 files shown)


nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
```diff
@@ -210,7 +210,16 @@ void ParametersHandler::getParam(
   setParam<ParamT>(setting, name, node);
 
   if (param_type == ParameterType::Dynamic) {
+    if (verbose_) {
+      RCLCPP_INFO(node->get_logger(), "setDynamicParamCallback for %s", name.c_str());
```
SteveMacenski, line 214: Debug log perhaps?

nav2_mppi_controller/src/parameters_handler.cpp
```diff
@@ -66,15 +69,21 @@ ParametersHandler::dynamicParamsCallback(
     {
       callback->second(param);
     } else {
-      RCLCPP_WARN(logger_, "Parameter %s not found", param_name.c_str());
+      if (verbose_) {
```
SteveMacenski, line 72: I approve of these changes, but I think you can remove all the comments. The log warning content is pretty clear

SteveMacenski (changes requested):
Generally approve, 2 nits and will merge if tests pass!

nav2_mppi_controller/src/parameters_handler.cpp
SteveMacenski: Is there still a question on the second success comment?

nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
aosmw: Ok done.

nav2_mppi_controller/src/parameters_handler.cpp
aosmw: Removed all comments.

Added test to check behavior when trying to get a parameter that has not been declared. ie you get an exception.

Note that to get the log warning content the handler needs to have verbose_ set to true.  Maybe this code path should always warn.

nav2_mppi_controller/src/parameters_handler.cpp
```diff
@@ -53,6 +54,7 @@ ParametersHandler::dynamicParamsCallback(
 {
   rcl_interfaces::msg::SetParametersResult result;
   std::lock_guard<std::mutex> lock(parameters_change_mutex_);
+  bool success;
```
aosmw, line 57: `bool success` needs a default

aosmw: done

nav2_mppi_controller/src/parameters_handler.cpp
aosmw: I went a bit further/deeper and used the ParameterType::Dynamic / Static to improve the log output.

Now attempts to set parameters that are not declared will report an error.
There will be a debug message indicating that as static parameters are being ignored if in verbose mode otherwise the log is silent.

nav2_mppi_controller/include/nav2_mppi_controller/tools/parameters_handler.hpp
```diff
@@ -224,24 +223,32 @@ void ParametersHandler::setParam(
...
+  auto dynamic_callback = [this, &setting, name](const rclcpp::Parameter & param) {
       setting = as<T>(param);
 
       if (verbose_) {
         RCLCPP_INFO(logger_, "Dynamic parameter changed: %s", std::to_string(param).c_str());
       }
     };
 
-  addDynamicParamCallback(name, callback);
+  auto static_callback = [this, &setting, name](const rclcpp::Parameter & param) {
```
SteveMacenski, line 241: If a static callback is created, then if someone attempts to set a static parameter through the dynamic parameter interface, it would `return true` on the change, even though its static and rejected the update. I would think that it should fail to find the callback and thus `return false`, just as if the parameter doesn't exist since this parameter is not able to be adjusted dynamically

SteveMacenski: @aosmw following up on this item

aosmw: @SteveMacenski 

I've taken this further in the latest commit prompted by your previous comment.  I have snipped the commit message here too.

Provide a mechanism to populate an rcl_interfaces::msg::SetParametersResult
with the reasons for unsuccessful parameter setting, so that it may be
propagated back to a set parameter client.

The mechanism provides feedback when attempting to set undefined
parameters, static parameters and could be used to validate dynamic
parameters and provide a reason for rejection.

NOTE: This changes public interface of ParametersHandler class.
s/setDynamicParamCallback/setParamCallback
s/addDynamicParamCallback/addParamCallback which takes
a callback function that is able to populate a
rcl_interfaces::msg::SetParametersResult.

In order to indicate an unsuccessful parameter change and the
reason, callback functions should append a "\n"
to the reason before appending to if it is not empty.

aosmw:
Not ready. Need to fix reason appending.

SteveMacenski:
OK, let me know :-0

nav2_mppi_controller/src/parameters_handler.cpp
```diff
@@ -64,17 +68,23 @@ ParametersHandler::dynamicParamsCallback(
     if (auto callback = get_param_callbacks_.find(param_name);
       callback != get_param_callbacks_.end())
     {
-      callback->second(param);
+      callback->second(param, result);
     } else {
-      RCLCPP_WARN(logger_, "Parameter %s not found", param_name.c_str());
+      result.successful = false;
```
SteveMacenski, line 73: Wouldn't multiple `parameters` in the for loop overwrite the `result.successful = false` if the next one is successful?

aosmw: That is not my read of it.  

The result.successful is initialized to true outside the for loop.

The "contract" (albeit un-enforceable) of the callback is to only touch the  reference to the result passed into it on failure.

SteveMacenski: This is true, sorry

aosmw: No worries.

Question:  Is the "atomically" in "set_parameters_atomically" intended to mean?

1. Change multiple parameters in the same operation (tick)  OR 
2. Only change parameters if all parameters can be set without failure

The current main branch AND this PR (currently) supports option 1.

Better support for option 2  could be offered by passing a reference to the result to the pre_callbacks where validation could occur.

For instance, Rejection of attempts to change a static parameter type should probably be done in a pre_callback.

Then a check for result.successful should be done before looping again and calling the callbacks to actually set the parameters.  These callbacks could also fail, but at that point supporting a rollback is going to add a lot more complexity.

If option 1 is enough then I think is ready for review.

aosmw:
> Not ready. Need to fix reason appending.
Reason appending was fixed.

I also commented on a resolved conversation -  https://github.com/ros-navigation/navigation2/pull/4711#discussion_r1849403363

Other than that its ready for review.

SteveMacenski (commented):
My only nitpick is why we don't just have all updates to `result.reason` have a `/n` at the end so that we don't need these blocks in both locations? 

```
      if (!result.reason.empty()) {
        result.reason += "\n";
      }
```

Otherwise, LGTM

nav2_mppi_controller/src/parameters_handler.cpp
SteveMacenski: Atomically means its all or nothing for multiple parameters, while non-atomically sets up to the failure.

I think option 1 is fine for now. We don't really handle dynamic parameters in the way that the atomic would make a difference. Even if the `rclcpp::get_parameter()` returns the original value due to a failure, the stored `my_param_value_` would be updated in the dynamic parameter callbacks for each case. 

I think to properly handle the dynamic parameter callbacks, we'd need the dynamic parameter callbacks to **only** do parameter validation, then as a post-callback step, re-get all parameter values via `get_parameter`. That would require a bit update across the stack, not just in MPPI, to be treated that way.

We could make that update, but I don't think we need to make that here in this PR. I'm a fan of multiple bitesized PRs versus one very large one.

aosmw:
> My only nitpick is why we don't just have all updates to `result.reason` have a `/n` at the end so that we don't need these blocks in both locations?
> 
> ```
>       if (!result.reason.empty()) {
>         result.reason += "\n";
>       }
> ```
Only to avoid the newline in the log in the case of a single reason.

SteveMacenski (approved):

PR #4712, opened 2024-10-07
map_saver: fix saved raw image being too dark

DylanDeCoeyer-Quimesis:
The map cell value comes from an OccupancyGrid and ranges from 0 to
100. Dividing it by 255 makes the output ratio range from 0 to 0.39. This is clearly an error as the output should range from 0 to 1. The consequence is that the output raw image is darker than expected.

---

## Basic Info

| ------ | ----------- |

---

* Fixed an error in the pixel value when saving a costmap in raw mode

---

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/src/map_io.cpp
+++ b/nav2_map_server/src/map_io.cpp
@@ -501,24 +501,24 @@ void tryWriteMapToFile(
             }
             break;
           case MapMode::Raw:
-          {
-            double cell_cost;
-            if (map_cell >= nav2_util::OCC_GRID_FREE && map_cell <= nav2_util::OCC_GRID_OCCUPIED) {
-              // Linear conversion from OccupancyGrid data range [OCC_GRID_FREE..OCC_GRID_OCCUPIED]
-              // to costmap data range [FREE_SPACE..LETHAL_OBSTACLE]
-              uint8_t lethal_obstacle_cost = 254; // nav2_costmap_2d::LETHAL_OBSTACLE
-              uint8_t free_space_cost = 0; // nav2_costmap_2d::FREE_SPACE
-              cell_cost = static_cast<double>(map_cell) * (lethal_obstacle_cost - free_space_cost)
-                / (nav2_util::OCC_GRID_OCCUPIED - nav2_util::OCC_GRID_FREE);
-            } else {
-              // nav2_util::OCC_GRID_UNKNOWN or other invalid values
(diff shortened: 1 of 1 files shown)


nav2_map_server/src/map_io.cpp
```diff
@@ -505,7 +505,7 @@ void tryWriteMapToFile(
             if (map_cell < 0 || 100 < map_cell) {
               q = MaxRGB;
             } else {
-              q = map_cell / 255.0 * MaxRGB;
+              q = map_cell / 100.0 * MaxRGB;
```
SteveMacenski, line 508: Similar to the other PR, the ROS 1 version seems to use 255 (see `negate`) which has been in use for 10+ years. This seems OK on face value, but I'm a little weary 

https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/image_loader.cpp#L130-L136

SteveMacenski: Though, the `MapMode::Scale` seems to agree with you, so I'm leaning towards merging this. Lets continue to chat in the other PR you submitted and we'll merge these as a pair

Thanks!

DylanDeCoeyer-Quimesis: Hi @SteveMacenski , thanks for your reviews. You are right, the other PR relies on this one. However, the opposite is not true.

In my understanding, this PR fixes a real bug. It could be merged even though discussions are still opened in the other one.

SteveMacenski: I commented on the other thread. Can you update this PR to contain the updates that supersedes https://github.com/ros-navigation/navigation2/pull/4713? Lets just put both of these together, its a simple enough set of code changes to review and are related

DylanDeCoeyer-Quimesis:
After investigation, I finally understood that there was no error in the initial implementation. The values stored in the .pgm are NOT meant to be equivalent to the ones stored in the Costmap2D internal grid, despite sharing the same range [0-255]. The .pgm file is just used as a uint8_t container to store the raw int8_t values of the OccupancyGrid ranging from 0 to 100. The only modified value is -1 (UNKNOWN) which is mapped to 255. So, it's expected to have a "dark" costmap stored in the .pgm file. And the UNKNOWN level is perfectly distinguishable from the others.

My apologies for the time lost.

SteveMacenski:
No worries!

PR #4715, opened 2024-10-09
Simplify namespaced bringups and multirobot sim

luca-della-vedova:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | N/A, see https://github.com/ros-navigation/nav2_minimal_turtlebot_simulation/pull/15 for background |
| Primary OS tested on | Ubuntu 24.04 Jazzy |
| Robotic platform tested on | TB3 (single / multiple) and TB4 sim on Gazebo Harmonic |

---

Reworked namespacing, no more having an rviz config with absolute topics and one for namespaced topics that relies on string substitution, but just a single one that will use the namespace passed to rviz to be resolved, for example:

```
## BEFORE:
# rviz_default_view.rviz, subscribing to laser scan
topic: /scan
# rviz_namespaced_view.rviz, subscribing to laser scan
topic: <robot_namespace>/scan
```

```
## AFTER:
# rviz_default_view.rviz
topic: scan
# When running, just set the rviz namespace, when none is specified it will automatically default to /scan
```

Note that topic names were also hardcoded in the nav2 params, to my understanding most of the duplication in the config files for the multirobot sim was really just to make sure different robots are subscribed to different topics, however now with proper namespace support all the robots can just reuse the same config file without needing to change topics, do string substitution or anything like that.

Not sure, the previous behavior should still work, but now bringing up multiple robots should require significantly less work / config file duplication.

---

There is just one rough edge (and the only code change in this PR) that I had to hack to make this work, which I'm not terribly happy with (but not too sure there is any better way), [here](https://github.com/ros-navigation/navigation2/compare/main...luca-della-vedova:navigation2:luca/namespaced_tb?expand=1#diff-1a89c00160d487e4e59524a966824eb7aab2267fdd5ffbb561be59eab96f9567R192-R202).

The costmap topics are in the namespace `<chosen_namespace>/[local, global]_costmap`, which makes it tricky for users to pass a local topic to subscribe to such as `scan`, since that would be parsed as `<chosen_namespace>/local_costmap/scan` which is clearly not what most cases would do.
I added that "hack" to manually go up one namespace and make sure that the topic name is relative to the chosen namespace, not the costmap local one. It works well but I can't say I'm too happy with it (I was really hoping there would be a `..` operator for topics so I could specify it as `../scan` but alas that seems to not be the case).

#### For Maintainers: 


nav2_bringup/rviz/nav2_default_view.rviz
```diff
@@ -58,7 +58,7 @@ Visualization Manager:
         Durability Policy: Volatile
         History Policy: Keep Last
         Reliability Policy: Reliable
-        Value: /robot_description
```
SteveMacenski, line 61: Q: I believe if you just save an rviz config, this is how it comes out. While I agree with this change, I worry about its maintainability for the future when new panels are added, configs change, and so on and we save a new config and invariably forget to remove the `/`'s  at some point in the long-term future

nav2_bringup/params/nav2_params.yaml
```diff
@@ -42,7 +42,7 @@ bt_navigator:
   ros__parameters:
     global_frame: map
     robot_base_frame: base_link
-    odom_topic: /odom
```
SteveMacenski, line 45: Comment: Regardless, I agree with this. These were added due to back-in-the-day issues with ROS 2 topics needing the full path which has long _long_ not been true

nav2_bringup/params/nav2_params.yaml
```diff
@@ -211,7 +211,7 @@ local_costmap:
         mark_threshold: 0
         observation_sources: scan
         scan:
-          topic: /scan
+          topic: scan
```
SteveMacenski, line 214: Doesn't this try to find the node's relative topic? `global_costmap/global_costmap/scan`?

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -189,6 +189,17 @@ void ObstacleLayer::onInitialize()
     node->get_parameter(name_ + "." + source + "." + "raytrace_min_range", raytrace_min_range);
     node->get_parameter(name_ + "." + source + "." + "raytrace_max_range", raytrace_max_range);
 
+    // If the user passed a relative topic, use it as relative to this node's parent namespace.
```
SteveMacenski, line 192: This seems unintuitive (and require updating in all layers) - but I also recognize the current behavior is also not overly intuitive to users if they don't specify the full path. 

What if the namespace of the topic is different than the namespace of the costmap server? I.e. `lidar1` vs `robot1` because you have `N` lidars of the same make and `N` robots you're simulating. the namespace of this node is usually not related to the namespace of the sensor drivers which are launched independently. 

So, I'd argue either keeping the current behavior or if not `/`, we throw out the namespace entirely and treat it like it was `/` (but then has the multirobot details to consider)

SteveMacenski (changes requested):

nav2_bringup/rviz/nav2_default_view.rviz
luca-della-vedova: I gave this a spin and I don't think so, it seems that the rviz box where you fill the topic is just a text box where actually no validation takes place and you can input whatever you want. Saving the rviz config just updates the field, for example I tried to change it such as:

image

And this showed up in my git diff:

image

nav2_costmap_2d/plugins/obstacle_layer.cpp
luca-della-vedova: > What if the namespace of the topic is different than the namespace of the costmap server? I.e. `lidar1` vs `robot1` because you have `N` lidars of the same make and `N` robots you're simulating. the namespace of this node is usually not related to the namespace of the sensor drivers which are launched independently.

I understand, again there is no perfect solution that I could find. Absolute paths will still work so I would say that if you have your costmap server in a `robot1` namespace but your lidars in `lidar1` namespace you should just set the topic as `/lidar1/scan`.
The current behavior is not really any better, since if you specify `lidar1/scan` you would end up subscribing to `/robot1/[local, global]_costmap/lidar1/scan`.
If you set the namespace of the costmap server to `robot1` and you ask it to listen to the `scan` topic (as a relative path), it sounds more intuitive to me that it would listen to `/robot1/scan`, rather than the current behavior of silently adding a `[local, global]_costmap` namespace inbetween.

Still, this is just a proposal, in the current incantation it can get rid of all the special configurations for multirobot / namespaced robots which is imho the ideal scenario, but of course it introduces tradeoffs such as this. Any other compromise might still move in the right direction, it just won't be 100% there but it's still something, for example:
* Have at least a shared rviz file and don't rely on string substitution
* Change some of the topics such as `/odom` to be relative so users don't need to change too many fields in multirobot params

Happy to remove parts of this if you think they are too controversial

SteveMacenski: Sorry for the delay -- ROSCon / ROS-I and my European travel thereafter was in the way. I'm back at home now.  

> I understand, again there is no perfect solution that I could find. Absolute paths will still work so I would say that if you have your costmap server in a robot1 namespace but your lidars in lidar1 namespace you should just set the topic as /lidar1/scan.

We agree on this - I generally think it wise to specify the full topic for an application. 

Something I'm also considering and maybe this is not such a big deal is that: alot of people use the Nav2 config files as the copy+paste basis for their robot's configurations. So right now we have the absolute path that folks would replace with their absolute path. In exchange for Nav2 software simplicity, I think it pushes some subtle complexity onto the user to realize that **they** shouldn't put `scan`, even though we do here, they should put `/scan`. A few things in ROS are done this way that create some user friction that I generally try to avoid, but also like you say, `/robot1/[local, global]_costmap/lidar1/scan` is really not much better than `/robot1/scan` other than the fact that it says the costmaps to know where its coming from more immediately to fix. 

Perhaps just adding a comment about this in the yaml file entries (+config docs on docs.nav2.org) for the topics would be sufficient to get past this so that its not subtle but immediately visible to the end-user. And a recommendation to use full topic paths if you don't want to worry about it. 

What do you think about that solution? Is that least or more straight forward for a user? If so, I'm OK with this as long as we update all the default plugins.

luca-della-vedova: OK I hope I understood correctly, I created a [migration guide PR](https://github.com/ros-navigation/docs.nav2.org/pull/609) and in https://github.com/ros-navigation/navigation2/pull/4715/commits/55263dd8af48e8657842bfbb7b2a89369575f3f2 I refactored the code that joins the topic to the namespace into a common function and reused it in other places where I found costmap layers to be using topic parameters to subscribe, (seems to be the `range_sensor` and `static_layer`?).
Note that I am _not_ very confident with the change I did to these two layers since I am not 100% sure how to test them

nav2_dwb_controller/dwb_critics/test/obstacle_footprint_test.cpp
```diff
@@ -125,11 +125,11 @@ TEST(ObstacleFootprint, Prepare)
...
   auto node = nav2_util::LifecycleNode::make_shared("costmap_tester");
 
-  auto costmap_ros = std::make_shared<nav2_costmap_2d::Costmap2DROS>("test_global_costmap");
+  std::string ns = "/ns";
+  auto costmap_ros = std::make_shared<nav2_costmap_2d::Costmap2DROS>("test_global_costmap", ns);
   costmap_ros->configure();
 
   std::string name = "name";
-  std::string ns = "ns";
   critic->initialize(node, name, ns, costmap_ros);
```
luca-della-vedova, line 133: This test started failing, I'm not sure the fact that two nodes in different namespaces could communicate effectively was a feature or a bug, but now they need to have the same namespace. I added a convenience method for building a `Costmap2DROS` object while only specifying the parent_namespace to make this a bit less verbose, but I can revert that if needed

nav2_bringup/params/nav2_params.yaml
SteveMacenski, line 214: For here + the global costmap below, add in the example in a comment so that its clear inline here the behavior and recommendation to use the global `/`

```
   * For example:
   *   * User chosen namespace is `tb4`.
   *   * User chosen topic is `scan`.
   *   * Topic will be remapped to `/tb4/scan` without `local_costmap`.
   *   * Use global costmap `/scan` if you do not wish the node namespace to apply
 ```

(I updated it for task)

File: nav2_costmap_2d/include/nav2_costmap_2d/costmap_layer.hpp [hunk 6]
```diff
@@ -192,6 +194,23 @@ class CostmapLayer : public Layer, public Costmap2D
...
+   * If the topic has an absolute path, it is returned instead.
+   *
+   * This is necessary for user defined relative topics to work as expected since costmap layers
+   * add a an additional `costmap_name` namespace to the topic.
+   * For example:
+   *   * User chosen namespace is `tb4`.
+   *   * User chosen topic is `scan`.
+   *   * Costmap node namespace will be `/tb4/global_costmap`.
+   *   * Without this function, the topic would be `/tb4/global_costmap/scan`.
+   *   * With this function, topic will be remapped to `/tb4/scan`.
```
SteveMacenski, line 208: /+ 

```
   * Use global costmap `/scan` if you do not wish the node namespace to apply
```

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
```diff
@@ -89,6 +89,18 @@ class Costmap2DROS : public nav2_util::LifecycleNode
...
   explicit Costmap2DROS(const std::string & name, const bool & use_sim_time = false);
 
+  /**
+   * @brief  Constructor for the wrapper, the node will
+   * be placed in a namespace equal to the parent_namespace and the node's name
+   * @param name Name of the costmap ROS node
+   * @param name Absolute namespace of the node hosting the costmap node
+   * @param use_sim_time Whether to use simulation or real time
+   */
+  explicit Costmap2DROS(
```
SteveMacenski, line 99: Can we remove this? We have alot of constructors already and if this is just for a test, we can use the one with the parent / local namespaces and just leave the local namespace = name

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -189,6 +189,7 @@ void ObstacleLayer::onInitialize()
     node->get_parameter(name_ + "." + source + "." + "raytrace_min_range", raytrace_min_range);
     node->get_parameter(name_ + "." + source + "." + "raytrace_max_range", raytrace_max_range);
 
+    topic = joinWithParentNamespace(topic);
```
SteveMacenski, line 192: STEVE: Update STVL/NPVL

SteveMacenski (changes requested):

nav2_costmap_2d/plugins/obstacle_layer.cpp
SteveMacenski: Correct! I think we just need that note added to the configuration guide for each of the `topic` parameters in the costmap layers (I mentioned in the comments below + in the docs PR). The migration guide is also required, so that's good as well thanks! 

Range is another layer like Obstacle/Voxel. If you found that those work, I wouldn't question Range.

Static is worth testing. You could change the namespaces and make sure it all works still. If you launch the Cloned / Unique multi-robot launch files does it still work? Does the single robot without namespace and with namespace work? Then I think you're fine.

luca-della-vedova: OK I've done the following to test it:

Apply this patch to make the topic local:

```
lucadv@noble:~/ionic_ws/src/nav2/navigation2/nav2_costmap_2d/plugins$ git diff
diff --git a/nav2_bringup/params/nav2_params.yaml b/nav2_bringup/params/nav2_params.yaml
index 6bd51ac1..650e33d1 100644
--- a/nav2_bringup/params/nav2_params.yaml
+++ b/nav2_bringup/params/nav2_params.yaml
@@ -229,6 +229,7 @@ local_costmap:
       static_layer:
         plugin: "nav2_costmap_2d::StaticLayer"
         map_subscribe_transient_local: True
+        map_topic: map
 
 global_costmap:
@@ -263,6 +264,7 @@ global_costmap:
           obstacle_max_range: 2.5
           obstacle_min_range: 0.0
       static_layer:
+        map_topic: map
         plugin: "nav2_costmap_2d::StaticLayer"
         map_subscribe_transient_local: True
       inflation_layer:
```

Run the multirobot bringup with and without the fix for the parent namespace.
Without the fix (I assume the fact that there is no cost around the map means it's not working):

Screenshot from 2024-11-21 10-00-17

I also get the following error:

```
[component_container_isolated-8] [WARN] [1732155528.401152235] [robot1.global_costmap.global_costmap]: Can't update static costmap layer, no map received
```

With the fix:

Screenshot from 2024-11-21 10-01-21

So I guess it's doing the right thing?
I've done the same for a unique tb4 simulation both with and without a namespace and got the same result. This is expected since even if we push no namespace by default the node will have a `/global_costmap` namespace.

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_ros.hpp
luca-della-vedova: Reverted in https://github.com/ros-navigation/navigation2/pull/4715/commits/ac5d861dbf44f0c80bd624c4d55722e120c12c28

nav2_costmap_2d/include/nav2_costmap_2d/costmap_layer.hpp
luca-della-vedova: https://github.com/ros-navigation/navigation2/pull/4715/commits/ac5d861dbf44f0c80bd624c4d55722e120c12c28 (changed `Use global costmap` to `Use global topic` since `/scan` is not really a costmap?)

nav2_bringup/params/nav2_params.yaml
luca-della-vedova: https://github.com/ros-navigation/navigation2/pull/4715/commits/ac5d861dbf44f0c80bd624c4d55722e120c12c28

luca-della-vedova:
I realized I forgot to properly cleanup the `use_namespace` parameter throughout the codebase since it's not used anymore so I did it in https://github.com/ros-navigation/navigation2/pull/4715/commits/ba08fe2d42e3e3fe8141e03907b56f6e8483a42d.

nav2_bringup/launch/bringup_launch.py
```diff
@@ -58,16 +57,6 @@ def generate_launch_description():
...
-    # '<robot_namespace>' keyword shall be replaced by 'namespace' launch argument
-    # in config file 'nav2_multirobot_params.yaml' as a default & example.
-    # User defined config file should contain '<robot_namespace>' keyword for the replacements.
-    params_file = ReplaceString(
-        source_file=params_file,
-        replacements={'<robot_namespace>': ('/', namespace)},
-        condition=IfCondition(use_namespace),
-    )
-
     configured_params = ParameterFile(
```
luca-della-vedova, line 60: I can't quite tell if this is still necessary or not now that the above was removed

SteveMacenski: The `ParameterFile` bit? Yes, that is necessary

nav2_costmap_2d/plugins/obstacle_layer.cpp
SteveMacenski: I'm not totally clear why that's needed. If we don't set map topic in the layer, it should use the costmap's global version - which is `map` by default [1]. The local/global costmap's main parameters don't change, right? We're only adjusting the layer's param handling [2]. 

I'm not disagreeing with the change, but I'd poke that API a bit and see why its not working now and make sure that's a sane behavior. If so, add in your patch to make it a local parameter to the layers. 

Or, more correctly, shouldn't we fix the fact that the global `map_topic` in the costmap itself isn't working? Should that also have your `joinWithParentNamespace` change? 

[1] https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_costmap_2d/src/costmap_2d_ros.cpp#L73
[2] https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_costmap_2d/plugins/static_layer.cpp#L152-L157

SteveMacenski (approved):
Beyond this topic on `map_topic`, I approve all other changes and happy to merge once resolved :-)

nav2_costmap_2d/plugins/obstacle_layer.cpp
luca-della-vedova: Apologies if this will be a long reply but as you can probably see by know it is a complicated topic with a lot of corner cases.
I admit that my first approach was really a "test and see" and not really digging super deep into the codebase, but I did now and some more explanations:

The situation is arguably a bit confusing, the topic will subscribe to the content of the `map_topic_` variable. If the user specifies a `map_topic` parameter that will be the content of `map_topic_`. Otherwise, it will default to the default value of `map_topic`.

What makes it even more confusing, is that the `Costmap2DROS` node can be built in a variety of ways which affect the parameter declaration. I did the good ol' printing method and found out that during a bringup the costmap nodes are created through a constructor different from the one you linked,
[this one](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_costmap_2d/src/costmap_2d_ros.cpp#L78-L106), specifically in the [controller_server](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_controller/src/controller_server.cpp#L71-L73) and [planner_server](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_planner/src/planner_server.cpp#L66-L68).
Now it's interesting to see how that constructor solves the namespace problem and why it works out of the box:

* The input parameters are `parent_namespace` and `local_namespace`.
* The node is launched with a namespace that is the result of `nav2_util::add_namespaces(parent_namespace, local_namespace)`
* The [subscription itself](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_costmap_2d/src/costmap_2d_ros.cpp#L102-L105) fully ignores the `local_namespace` and only uses the `parent_namespace` to create an absolute topic and store it in the parameter value. I would say this is pretty much what I'm doing downstream, except I don't have the luxury of knowing both the `local_namespace` and `parent_namespace`.

The problem however, is that logic will only be applied if a global `map_topic` is not defined since it is contained in the "default value logic" of parameter declaration. Once we define a `map_topic` we will go back to square one, as you correctly identified.
Now if I "just" use `joinWithParentNamespace` also with global topics this particular use case will be fixed, since the default value concatenation will make the topic path absolute, which will result in a noop with the function call, and specifying the parameter value will mark it as relative, and be parsed with the parent, and I've done that in https://github.com/ros-navigation/navigation2/pull/4715/commits/745af36add45b98f4c75be1fb1a5e37a95a1416c

However, food for thought, what happens if a users was to pass a `local_namespace` that contains a `/`, such as `global_costmap/my_costmap`? Note that right now this is only possible programmatically, i.e. by changing [this call](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_controller/src/controller_server.cpp#L71-L73). The answer is "this implementation would break down, since it hardcodes "go up one level" and the information of how many of the levels of namespacing are global and how many are local is lost.
As a more robust and possibly cleaner alternative I just came up with a private parameter approach, where the node declares a `__parent_namespace` parameter so we know exactly what's the parent namespace and we can concatenate topics accordingly, without having to hardcode anything. The cost of this of course is another parameter (I believe there is no proper concept of private parameters other than adding underscores like in Python?)
https://github.com/ros-navigation/navigation2/pull/4715/commits/f66d9a97bf053bd4ea0d579632225f069c5edfd6

Again apologies for the long iteration and really thanks a lot for the thorough review, I'm learning a lot and I feel this is going to a good place!

luca-della-vedova: It seems https://github.com/ros-navigation/navigation2/commit/0c958dc994f5e618fef2471c0851a2bc3b19b1d5 adds a dependency inversion on `CostmapLayer` that will not work unless it is part of `Costmap2DROS`, hence reverted it https://github.com/ros-navigation/navigation2/pull/4715/commits/f8dda14c2204b381faf69653e13e77c73dd4fbd7

SteveMacenski: > [this one](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_costmap_2d/src/costmap_2d_ros.cpp#L78-L106), specifically in the [controller_server](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_controller/src/controller_server.cpp#L71-L73) and [planner_server](https://github.com/ros-navigation/navigation2/blob/90a6c8d80096469a22e1a56e5447d58fb8e66ef6/nav2_planner/src/planner_server.cpp#L66-L68).

:skull: :facepalm:  The way Intel handled the namespacing for multi-robot in costmap 2D since its an internal node (usually) of the task servers is consistently a problem. I totally forgot they included some special handling for `map_topic` -- that in my opinion should just be outright removed if there's a better proposal you have. Don't take keeping that in place as a requirement. Feel free to make any adjustments in the constructor you see fit, that is by no means the way I would like it. 

> Now if I "just" use joinWithParentNamespace also with global topics this particular use case will be fixed, since the default value concatenation will make the topic path absolute, which will result in a noop with the function call, and specifying the parameter value will mark it as relative, and be parsed with the parent, and I've done that in https://github.com/ros-navigation/navigation2/commit/745af36add45b98f4c75be1fb1a5e37a95a1416c

That seems very rational to me. Even better yet, we could simply remove the global `map_topic` from existing. I checked against ROS 1 and its only there for historical reasons. We could just fully depreciate it and it **must** be set **only** in the static layer - which is the only place its ever been used since the beginning of ROS 1 Navigation (it just used to be that the static layer was built-in). 

I think that would resolve all of your later "food for thought" comments. We'd just need to remove `map_topic` from the docs for the `Costmap2DROS` object & migration guide entry you wrote mention that it needs to be moved until the static layer now, if using another topic

SteveMacenski: This looks good to me - no issues. The local namespace appears to only be used for the `map_topic` anyway, so that's wonderful redundant information to remove. 

Get the migration update and LGTM to merge the pair, thanks so much!

luca-della-vedova: Done! https://github.com/ros-navigation/docs.nav2.org/pull/609/commits/835860d0f996e67142bc2fea272d488ef09877fb

SteveMacenski (approved):
Small docs update, then good to merge. I really appreciate your time and patience, I agree this is a really nice simplification and I'm sure every time I don't need to update `N` files with new parameter changes, I'll think of this!

Martin-Oehler:
Can this be merged on the Jazzy branch as well so it's in the next release? It seems like these changes are only on main.

SteveMacenski:
It cannot be, it breaks the current behavior for users that might be relying on the current mapping schema for their systems. We wouldn't want to break thousands of robots in an update not processing sensor data, now would we? :laughing:

PR #4720, opened 2024-10-12
[map_io] Replace std logs by rclcpp logs

doisyg:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Dexory's bot |

---

Replaced all the "std::cout" calls in map_io.cpp by their RCLCPP_(DEBUG/INFO/WARN/ERROR) equivalent.
We call a lot `loadMapYaml` externaly in our application and the verbose debug logs were polluting our log console.

---

#### For Maintainers: 


nav2_map_server/src/map_io.cpp
```diff
@@ -202,8 +207,9 @@ void loadMapFromFile(
   Magick::InitializeMagick(nullptr);
   nav_msgs::msg::OccupancyGrid msg;
 
-  std::cout << "[INFO] [map_io]: Loading image_file: " <<
-    load_parameters.image_file_name << std::endl;
+  RCLCPP_INFO_STREAM(
+    rclcpp::get_logger("map_io"), "Loading image_file: " <<
```
SteveMacenski, line 211: You can reduce the boilerplate by creating a Logger object at the start of the methods that have logging and pass those in instead of recreating each time

SteveMacenski (changes requested):
A couple of details to change, but LGTM

SteveMacenski (approved):

PR #4738, opened 2024-11-01
Remove nav2_loopback_sim dependency on transforms3d.

clalancette:
## Basic Info

| ------ | ----------- |

---

The package never uses it, so don't declare it.

I noticed this because I noticed that recent builds of Navigation2 on Jazzy don't work on RHEL anymore, i.e. https://build.ros2.org/job/Jsrc_el9__navigation2__rhel_9__source/ is disabled.  The reason for that is that `python3-transforms3d` is not packaged on RHEL-9.  However, it turns out that this package doesn't actually use `transforms3d`, so the easiest thing to do here is to remove the dependency.

Note that if accepted, this should also be backported to the `jazzy` branch, and a new bloom-release should be done on the buildfarm for Jazzy.

None needed.

---

Backport to the `jazzy` branch.

#### For Maintainers: 

### DIFF
--- a/nav2_loopback_sim/package.xml
+++ b/nav2_loopback_sim/package.xml
@@ -12,7 +12,6 @@
   <exec_depend>nav_msgs</exec_depend>
   <exec_depend>tf_transformations</exec_depend>
   <exec_depend>tf2_ros</exec_depend>
-  <exec_depend>python3-transforms3d</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>


SteveMacenski (approved):

SteveMacenski:
@clalancette Thanks! Auto-backport in progress. New release will be cut EOM (maybe a bit earlier since I skipped the October once due to ROSCon, ROS-I)

SteveMacenski:
I actually have it on my schedule for Friday, so it'll be sooner than later

PR #4744, opened 2024-11-07
Pass IDLE to on_tick, use that for initialize condition

redvinaa:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |#3988 |
| Robotic platform tested on | – |

---

* Pass IDLE state to BT action nodes (to the on_tick callback), and use that as initialize condition
* This way ports can be updated at every restart (on first start and after each halt)
* See conversation at the bottom of #3988

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -191,14 +191,14 @@ public:
   {
     // first step to be done only at the beginning of the Action
     if (!BT::isStatusActive(status())) {
-      // setting the status to RUNNING to notify the BT Loggers (if any)
-      setStatus(BT::NodeStatus::RUNNING);
-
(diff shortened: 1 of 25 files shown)


SteveMacenski (approved):
I see no issues with this, LGTM once CI passes :-) 

Thanks for bringing this to our attention!

SteveMacenski:
@redvinaa it looks like 1 test is failing (ignore DWB; that's occasionally flaky)

redvinaa:
I have no idea why that particular test failed. It's possible that a rerun would solve it.

SteveMacenski:
I triggered a rebuild, but I don't expect that to change, a unit test like that shouldn't be flaky

redvinaa:
Found it: the subscriber was always recreated in `initialize`, and as in the test we spin first and then tick, the callback was always called with the old inputs. Changed `initialize` to only create new subscriber when the topic has changed (or in the beginning when the sub doesn't exist yet).

SteveMacenski (approved):

PR #4749, opened 2024-11-09
Backward docking without sensors

Jakubach:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [ticket1](https://github.com/ros-navigation/navigation2/issues/4403) [ticket2](https://github.com/open-navigation/opennav_docking/issues/68) |

---

- Fixed an issue with backward docking described in issue
- Added an option for backward docking without continious pose refining (to use only initial detected dock pose when robot is docking backward but it has sensors for detection only on front)
- Added an initial rotation functionality 
- Added a minimal angular velocity parameter which is used for initial rotation
- Updated readme file

---

#### For Maintainers: 


nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -40,6 +40,9 @@ DockingServer::DockingServer(const rclcpp::NodeOptions & options)
   declare_parameter("fixed_frame", "odom");
   declare_parameter("dock_backwards", false);
   declare_parameter("dock_prestaging_tolerance", 0.5);
+  declare_parameter("initial_rotation", true);
+  declare_parameter("initial_rotation_min_angle", 0.5);
+  declare_parameter("dock_backwards_without_sensor", true);
```
SteveMacenski, line 45: I'd make this `backward_blind`, and in all appropriate places to rename / readme

This should be default `false`.

nav2_docking/README.md
```diff
@@ -203,19 +203,23 @@ For debugging purposes, there are several publishers which can be used with RVIZ
 | base_frame        | Robot's base frame for control law   | string |  "base_link"      |
 | fixed_frame        | Fixed frame to use, recommended to be a smooth odometry frame **not** map   | string |  "odom"      |
 | dock_backwards        | Whether the robot is docking with the dock forward or backward in motion | bool |  false      |
+| dock_backwards_without_sensor        | Whether the robot is docking based on initial detection without updates | bool |  false      |
```
SteveMacenski, line 206: Initial forward detections, then dock in reverse

nav2_docking/opennav_docking/include/opennav_docking/controller.hpp
```diff
@@ -59,11 +59,18 @@ class Controller
...
+   * @returns Twist command.
+   */
+  geometry_msgs::msg::Twist rotateToTarget(const double & angle_to_target);
+
 protected:
   std::unique_ptr<nav2_graceful_controller::SmoothControlLaw> control_law_;
 
   double k_phi_, k_delta_, beta_, lambda_;
-  double slowdown_radius_, v_linear_min_, v_linear_max_, v_angular_max_;
+  double slowdown_radius_, v_linear_min_, v_linear_max_, v_angular_max_,v_angular_min_;
```
SteveMacenski, line 73:
```suggestion
  double slowdown_radius_, v_linear_min_, v_linear_max_, v_angular_max_, v_angular_min_;
```

Missing a space

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -744,6 +760,13 @@ DockingServer::dynamicParametersCallback(std::vector<rclcpp::Parameter> paramete
       if (name == "max_retries") {
         max_retries_ = parameter.as_int();
       }
+    } else if(type == ParameterType::PARAMETER_BOOL){
+      if (name == "initial_rotation") {
+        initial_rotation_ = parameter.as_bool();
+      }
+      if (name == "dock_backwards_without_sensor") {
```
SteveMacenski, line 767: Dock backward isn't a dynamic parameter either, if you're going to enable this one, I think it should be both.

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -744,6 +760,13 @@ DockingServer::dynamicParametersCallback(std::vector<rclcpp::Parameter> paramete
       if (name == "max_retries") {
         max_retries_ = parameter.as_int();
       }
+    } else if(type == ParameterType::PARAMETER_BOOL){
```
SteveMacenski, line 763:
```suggestion
    } else if (type == ParameterType::PARAMETER_BOOL) {
```

You seem to have more linting issues - I'm not sure why CI didn't catch them. Please run `ament_cpplint` and `ament_uncrustify`

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -385,6 +391,17 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
   publishDockingFeedback(DockRobot::Feedback::INITIAL_PERCEPTION);
   rclcpp::Rate loop_rate(controller_frequency_);
   auto start = this->now();
+  // If docking without sensors, stop the robot for short time to get a stable and quality detection of docking pose
+  if(dock_backwards_without_sensor_){
```
SteveMacenski, line 395: This shouldn't be necessary. If we're at the initial staging pose, we're already stopped.

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -59,6 +62,9 @@ DockingServer::on_configure(const rclcpp_lifecycle::State & state)
   get_parameter("fixed_frame", fixed_frame_);
   get_parameter("dock_backwards", dock_backwards_);
   get_parameter("dock_prestaging_tolerance", dock_prestaging_tolerance_);
+  get_parameter("initial_rotation", initial_rotation_);
+  get_parameter("initial_rotation_min_angle", initial_rotation_min_angle_);
+  get_parameter("dock_backwards_without_sensor", dock_backwards_without_sensor_);
   RCLCPP_INFO(get_logger(), "Controller frequency set to %.4fHz", controller_frequency_);
```
SteveMacenski, line 68: Check if the dock backward without sensor is set but the dock backwards is not. In that case, return a failure on the configuration since that's an invalid configuration that is impossible to disambiguate

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -437,20 +454,17 @@ bool DockingServer::approachDock(Dock * dock, geometry_msgs::msg::PoseStamped &
...
     auto command = std::make_unique<geometry_msgs::msg::TwistStamped>();
-    command->header.stamp = now();
-    if (!controller_->computeVelocityCommand(target_pose.pose, command->twist, dock_backwards_)) {
+
+    geometry_msgs::msg::PoseStamped robot_pose = getRobotPoseInFrame(target_pose.header.frame_id);
+    const double angle_to_goal = angles::shortest_angular_distance(
+    tf2::getYaw(robot_pose.pose.orientation), atan2(robot_pose.pose.position.y - target_pose.pose.position.y, robot_pose.pose.position.x - target_pose.pose.position.x));
+
+    if(initial_rotation_ && fabs(angle_to_goal) > initial_rotation_min_angle_){
+      command->twist = controller_->rotateToTarget(angle_to_goal);
```
SteveMacenski, line 465: It seems more natural to have the initial rotation bit performed here: https://github.com/ros-navigation/navigation2/blob/14faf822a6595293cb045b064b6cde867d0731d5/nav2_docking/opennav_docking/src/docking_server.cpp#L288

For example:

```
    while (rclcpp::ok()) {
      try {
        // Invert robot orientation if necessary
        if (dock_backward_without_sensor_) {
            rotate180();
        }
        // Approach the dock using control law
        if (approachDock(dock, dock_pose)) {
          // We are docked, wait for charging to begin
          RCLCPP_INFO(
            get_logger(), "Made contact with dock, waiting for charge to start (if applicable).");

```

Then the rest stays the same

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -59,6 +62,9 @@ DockingServer::on_configure(const rclcpp_lifecycle::State & state)
   get_parameter("fixed_frame", fixed_frame_);
   get_parameter("dock_backwards", dock_backwards_);
   get_parameter("dock_prestaging_tolerance", dock_prestaging_tolerance_);
+  get_parameter("initial_rotation", initial_rotation_);
```
SteveMacenski, line 65: I'm not totally grokking the initial rotation / min angle stuff.

If you have `backward_blind` and `dock_backward` then we know we need to do the initial rotation to go from forward to reverse before doing the rest of the procedure. The `initial_rotation` would thus always be true.

Then if we know we're always rotating around, then we know we should be doing a pure 180 turn. I suppose we could also use the dock's detected position to do 180 +/- error so that we align better with it instead of a pure 180 turn though! That would be cool! But we still don't need a `min_angle` param.

Thus, unless I'm missing something, I think we don't need either of these 2 params.

SteveMacenski (changes requested):
I think we can do this a bit better - but I appreciate the first prototype and your contributions!

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: Oh, CI did pick up on them: https://app.circleci.com/pipelines/github/ros-navigation/navigation2/12898/workflows/0f40ab36-340c-4242-befb-e9c0abdfdc04/jobs/38897/tests

nav2_docking/opennav_docking/src/docking_server.cpp
Jakubach: The purpose behind that was to start docking before reaching 180 degree rotation. But that's true that it breaks the concept of initial rotation to dock. I removed it.

nav2_docking/opennav_docking/src/docking_server.cpp
Jakubach: That's the mystery of creating names. Thank you!

nav2_docking/opennav_docking/test/test_controller.cpp
```diff
@@ -63,6 +64,7 @@ TEST(ControllerTests, DynamicParameters) {
       rclcpp::Parameter("controller.v_linear_min", 5.0),
       rclcpp::Parameter("controller.v_linear_max", 6.0),
       rclcpp::Parameter("controller.v_angular_max", 7.0),
+      rclcpp::Parameter("controller.v_angular_min", 2.0),
```
SteveMacenski, line 67: + dock backwards + dock blind checks

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -57,11 +60,13 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
...
-bool Controller::computeVelocityCommand(
-  const geometry_msgs::msg::Pose & pose, geometry_msgs::msg::Twist & cmd, bool backward)
+ bool Controller::computeVelocityCommand(
+    const geometry_msgs::msg::Pose & pose,
+    const geometry_msgs::msg::Pose & robot_pose,
+    geometry_msgs::msg::Twist & cmd, bool backward)
 {
   std::lock_guard<std::mutex> lock(dynamic_params_lock_);
-  cmd = control_law_->calculateRegularVelocity(pose, backward);
+  cmd = control_law_->calculateRegularVelocity(pose, robot_pose, backward);
```
SteveMacenski, line 69: Please explain this for me -- is this related to the other item on reverse issues? I'd like @ajtudela's thoughts on this but I don't have a problem with it. I'm just surprised none of us thought about this if it was that simple

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -59,6 +61,17 @@ DockingServer::on_configure(const rclcpp_lifecycle::State & state)
...
+  get_parameter("backward_blind", backward_blind_);
+  if(backward_blind_ && !dock_backwards_){
+    RCLCPP_ERROR(get_logger(), "backward_blind is enabled when dock_backwards is disabled.");
+    return nav2_util::CallbackReturn::FAILURE;
+  } else{
+    // If you have backward_blind and dock_backward then
+    // we know we need to do the initial rotation to go from forward to reverse
+    // before doing the rest of the procedure. The initial_rotation would thus always be true.
+    initial_rotation_ = true;
```
SteveMacenski, line 73: Does `initial_rotation_` even need to be a parameter? If `backward_blind_ = true` don't we always do the rotation?

SteveMacenski: It also seems unused in the code itself :-)

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -744,6 +774,15 @@ DockingServer::dynamicParametersCallback(std::vector<rclcpp::Parameter> paramete
       if (name == "max_retries") {
         max_retries_ = parameter.as_int();
       }
+    } else if(type == ParameterType::PARAMETER_BOOL){
+      if (name == "dock_backwards") {
+        dock_backwards_ = parameter.as_bool();
+      }
+      if (name == "backward_blind") {
+        backward_blind_ = parameter.as_bool();
+        initial_rotation_ = parameter.as_bool();
```
SteveMacenski, line 783: You shouldn't update other parameters -- remove dock backward / initial rotation setting

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -401,6 +418,27 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
   }
 }
 
+
```
SteveMacenski, line 421: Remove extra white spaces

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -105,4 +112,20 @@ Controller::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
   return result;
 }
 
+geometry_msgs::msg::Twist Controller::rotateToTarget(const double & angle_to_target)
+{
+  geometry_msgs::msg::Twist vel;
+  vel.linear.x = 0.0;
+  vel.angular.z = 0.0;
+  if(angle_to_target > 0) {
+    vel.angular.z = std::clamp(1.0 * angle_to_target * v_angular_max_,
```
SteveMacenski, line 121: `1.0 *` in both expressions seems unnecessary. I think the other one might mean `-1.0`. 

What is the derivation of this rotation code for `angle_to_target * v_angular_max`? Might it be better to steal the rotation code from something like the rotation shim controller / spin behavior to make this consistent?

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -401,6 +418,27 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
...
+
+void DockingServer::rotateToDock(const geometry_msgs::msg::PoseStamped & dock_pose){
+  rclcpp::Rate loop_rate(controller_frequency_);
+  geometry_msgs::msg::PoseStamped robot_pose = getRobotPoseInFrame(dock_pose.header.frame_id);
+  double angle_to_goal;
+  auto command = std::make_unique<geometry_msgs::msg::TwistStamped>();
+  while(rclcpp::ok()){
+    angle_to_goal = angles::shortest_angular_distance(tf2::getYaw(robot_pose.pose.orientation),
```
SteveMacenski, line 429: A question to ask: would we rather (1) rotate purely 180 from the staging pose and have the control law handle any deviation in the staging pose or (2) rotate approximately 180, taking into account any angular offsets present in the staging pose so that its as straight-backward as possible? 

If (1), then we can just take the robot orientation and the target is just PI from that, no need to look at the dock pose.

If (2), then I think we need to look at the dock pose's orientation, not its relative position angle. But, I would agree this is up for debate about what's better. Another area @ajtudela's opinion would be appreciated :-)

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -401,6 +418,27 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
...
+    atan2(robot_pose.pose.position.y - dock_pose.pose.position.y,
+    robot_pose.pose.position.x - dock_pose.pose.position.x));
+    if(fabs(angle_to_goal) > 0.1){
```
SteveMacenski, line 432: This should either be (1) a configurable threshold or (2) wait until its exceeded so its at least that much of a turn

nav2_docking/opennav_docking/src/controller.cpp
SteveMacenski: Odometric error is often large with high acceleration, so it would probably be wise to have acceleration / decelleration in the rotation action such that we minimize slippage -- since especially in this case, we're fully dead reckoning

nav2_docking/opennav_docking/src/docking_server.cpp
Jakubach: I made a mistake here, there should be an opposite sign (<)

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -437,20 +475,12 @@ bool DockingServer::approachDock(Dock * dock, geometry_msgs::msg::PoseStamped &
         tf2::getYaw(target_pose.pose.orientation) + M_PI);
     }
 
-    // The control law can get jittery when close to the end when atan2's can explode.
-    // Thus, we backward project the controller's target pose a little bit after the
-    // dock so that the robot never gets to the end of the spiral before its in contact
-    // with the dock to stop the docking procedure.
-    const double backward_projection = 0.25;
```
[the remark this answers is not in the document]
SteveMacenski: I believe your PR outright removes this bit, not adjust it. I'm on board with the adjustment if @ajtudela is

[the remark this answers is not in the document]
Jakubach: I share your opinion

SteveMacenski:
Let me know when you want me to review again

nav2_docking/opennav_docking/src/controller.cpp
Jakubach: @SteveMacenski How should I take the current velocity inside docking_server in the good way?
shim controller requires it 
https://github.com/ros-navigation/navigation2/blob/f426bdaadaa1a124bea0da5f32961cef6d679ae2/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp#L288

SteveMacenski: You could get it the same way we do in the controller server with an odometry subscriber: https://github.com/ros-navigation/navigation2/blob/f426bdaadaa1a124bea0da5f32961cef6d679ae2/nav2_controller/src/controller_server.cpp#L626

Jakubach: Thank you @SteveMacenski, it's already implemented and I hope I managed it well.

nav2_docking/opennav_docking/src/docking_server.cpp
Jakubach: I called it tolerance in a hurry, but tell me please if it better should be "threshold"

nav2_docking/opennav_docking/src/docking_server.cpp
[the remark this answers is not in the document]
Jakubach: In my simulation it works fine and I hope we will have similar results!

nav2_docking/opennav_docking/src/controller.cpp
Jakubach: I just noticed that from time to time (not in 100% of situations) robot stops during performing an initial rotation. The angular distance is correct but robot the generatated velocities are like 0.003 rad/s
Now I am inspecting why that happens.

nav2_docking/opennav_docking/src/docking_server.cpp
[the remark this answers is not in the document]
ajtudela: It works, but I'd simplify a little bit to make it clearly: project + flip + transform. Somehting like:

```cpp
// The control law can get jittery when close to the end when atan2's can explode.
// Thus, we backward project the controller's target pose a little bit after the
// dock so that the robot never gets to the end of the spiral before its in contact
// with the dock to stop the docking procedure.
const double backward_projection = 0.25;
const double yaw = tf2::getYaw(target_pose.pose.orientation);
target_pose.pose.position.x += cos(yaw) * backward_projection;
target_pose.pose.position.y += sin(yaw) * backward_projection;

// Make sure that the target pose is pointing at the robot when moving backwards
// This is to ensure that the robot doesn't try to dock from the wrong side
if (dock_backwards_) {
  target_pose.pose.orientation = 
    nav2_util::geometry_utils::orientationAroundZAxis(yaw + M_PI);
}

tf2_buffer_->transform(target_pose, target_pose, base_frame_);
```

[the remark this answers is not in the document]
SteveMacenski: You need to flip the orientation before you backward project it, no? That way we always apply the point behind the docked pose, instead of 0.25m in front of the docked pose. ... or was the order incorrect beforehand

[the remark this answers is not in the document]
ajtudela: If you flip the orientation before the backward projection, the sign of the translation is inverted and the pose gets closer to the robot. 

It seems it was in the wrong order beforehand. It's odd that no one notice it. 

A couple of images:

Flip + Project
before

Project + Flip
after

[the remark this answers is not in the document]
Jakubach: I had similar observation. Putting that code before was a fix but I was not sure about that. Now it's confirmed I think.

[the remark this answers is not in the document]
SteveMacenski: Huh, that is interesting that everyone doing backward docking missed this -- sounds good

SteveMacenski:
Drop a comment when you're ready for a re-review :-)

nav2_docking/opennav_docking/src/docking_server.cpp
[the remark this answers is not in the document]
ajtudela: > Huh, that is interesting that everyone doing backward docking missed this -- sounds good

Yeah, the only thing I can think of is that they set the dock_pose much farther than the real dock, so the robot will bump the dock but never reach the target pose.

It would be nice to ask someone with such a robot what their settings are.

SteveMacenski:
There's some test failures / linting issues reported in CI: https://app.circleci.com/pipelines/github/ros-navigation/navigation2/13004/workflows/13cb1e43-ec84-477d-9253-5d4464f09fb2/jobs/39130/tests

Jakubach:
Thank you, I've noticed that there exist a new bug during undock (after backward docking). I will upload a fix.

[Screencast from 11-24-2024 03:54:41 PM.webm](https://github.com/user-attachments/assets/bbb3ec98-b5ce-4e92-9c05-78b19e8af7cb)

ajtudela:
> I've noticed that there exist a new bug during undock (after backward docking). The staging pose is in a wrong pose. I will upload a fix. The question is, should robot after docking go back to the real staging pose or it could go back to the rotated pose (that after initial rotation).
> 
>  Screencast.from.11-24-2024.03.54.41.PM.webm 
> My first try was to add two rotations and that solved the problem. Robot after undocking is in a "after-initial-rotation-staging-pose". But maybe should `getStagingPose` method handle this?
> 
> ```
> void DockingServer::undockRobot()
> {
>   std::lock_guard<std::mutex> lock(*mutex_);
>   action_start_time_ = this->now();
>   rclcpp::Rate loop_rate(controller_frequency_);
> 
>   auto goal = undocking_action_server_->get_current_goal();
>   auto result = std::make_shared<UndockRobot::Result>();
>   result->success = false;
> 
>   if (!undocking_action_server_ || !undocking_action_server_->is_server_active()) {
>     RCLCPP_DEBUG(get_logger(), "Action server unavailable or inactive. Stopping.");
>     return;
>   }
> 
>   if (checkAndWarnIfCancelled(undocking_action_server_, "undock_robot")) {
>     undocking_action_server_->terminate_all(result);
>     return;
>   }
> 
>   getPreemptedGoalIfRequested(goal, undocking_action_server_);
>   auto max_duration = rclcpp::Duration::from_seconds(goal->max_undocking_time);
> 
>   try {
>     // Get dock plugin information from request or docked state, reset state.
>     std::string dock_type = curr_dock_type_;
>     if (!goal->dock_type.empty()) {
>       dock_type = goal->dock_type;
>     }
> 
>     ChargingDock::Ptr dock = dock_db_->findDockPlugin(dock_type);
>     if (!dock) {
>       throw opennav_docking_core::DockNotValid("No dock information to undock from!");
>     }
>     RCLCPP_INFO(
>       get_logger(),
>       "Attempting to undock robot of dock type %s.", dock->getName().c_str());
> 
>     // Check if the robot is docked before proceeding
>     if (/*dock->isCharger() && */(!dock->isDocked() && !dock->isCharging())) {
>       RCLCPP_INFO(get_logger(), "Robot is not in the dock, no need to undock");
>       return;
>     }
> 
>     // Get "dock pose" by finding the robot pose
>     geometry_msgs::msg::PoseStamped dock_pose = getRobotPoseInFrame(fixed_frame_);
>     if (dock_backwards_) {
>       dock_pose.pose.orientation = 
>         nav2_util::geometry_utils::orientationAroundZAxis(
>           tf2::getYaw(dock_pose.pose.orientation) + M_PI);
>     }
> 
>     // Get staging pose (in fixed frame)
>     geometry_msgs::msg::PoseStamped staging_pose =
>       dock->getStagingPose(dock_pose.pose, dock_pose.header.frame_id);
> 
>     if (dock_backwards_) {
>       staging_pose.pose.orientation = 
>         nav2_util::geometry_utils::orientationAroundZAxis(
>           tf2::getYaw(staging_pose.pose.orientation) + M_PI);
>     }
>     
> 
>     // Control robot to staging pose
>     rclcpp::Time loop_start = this->now();
>     while (rclcpp::ok()) {
>       // Stop if we exceed max duration
>       auto timeout = rclcpp::Duration::from_seconds(goal->max_undocking_time);
>       if (this->now() - loop_start > timeout) {
>         throw opennav_docking_core::FailedToControl("Undocking timed out");
>       }
> 
>       // Stop if cancelled/preempted
>       if (checkAndWarnIfCancelled(undocking_action_server_, "undock_robot") ||
>         checkAndWarnIfPreempted(undocking_action_server_, "undock_robot"))
>       {
>         publishZeroVelocity();
>         undocking_action_server_->terminate_all(result);
>         return;
>       }
> 
>       // Don't control the robot until charging is disabled
>       if (!dock->disableCharging()) {
>         loop_rate.sleep();
>         continue;
>       }
> 
>       // Get command to approach staging pose
>       geometry_msgs::msg::Twist command;
>       if (getCommandToPose(
>           command, staging_pose, undock_linear_tolerance_, undock_angular_tolerance_,
>           !dock_backwards_))
>       {
>         RCLCPP_INFO(get_logger(), "Robot has reached staging pose");
>         // Have reached staging_pose
>         vel_publisher_->publish(command);
>         if (dock->hasStoppedCharging()) {
>           RCLCPP_INFO(get_logger(), "Robot has undocked!");
>           result->success = true;
>           curr_dock_type_.clear();
>           publishZeroVelocity();
>           undocking_action_server_->succeeded_current(result);
>           return;
>         }
>         // Haven't stopped charging?
>         throw opennav_docking_core::FailedToControl("Failed to control off dock");
>       }
> 
>       // Publish command and sleep
>       vel_publisher_->publish(command);
>       loop_rate.sleep();
>     }
>   } catch (const tf2::TransformException & e) {
>     RCLCPP_ERROR(get_logger(), "Transform error: %s", e.what());
>     result->error_code = DockRobot::Result::UNKNOWN;
>   } catch (opennav_docking_core::DockNotValid & e) {
>     RCLCPP_ERROR(get_logger(), "%s", e.what());
>     result->error_code = DockRobot::Result::DOCK_NOT_VALID;
>   } catch (opennav_docking_core::FailedToControl & e) {
>     RCLCPP_ERROR(get_logger(), "%s", e.what());
>     result->error_code = DockRobot::Result::FAILED_TO_CONTROL;
>   } catch (opennav_docking_core::DockingException & e) {
>     RCLCPP_ERROR(get_logger(), "%s", e.what());
>     result->error_code = DockRobot::Result::UNKNOWN;
>   } catch (std::exception & e) {
>     RCLCPP_ERROR(get_logger(), "Internal error: %s", e.what());
>     result->error_code = DockRobot::Result::UNKNOWN;
>   }
> 
>   publishZeroVelocity();
>   undocking_action_server_->terminate_current(result);
> }
> ```

Could you describe the bug, please? I don't see it clearly in the video.

Jakubach:
> > I've noticed that there exist a new bug during undock (after backward docking). The staging pose is in a wrong pose. I will upload a fix. The question is, should robot after docking go back to the real staging pose or it could go back to the rotated pose (that after initial rotation).
> 
> Could you describe the bug, please? I don't see it clearly in the video.

If `dock_backwards` is enabled then a 180 degree rotation of dock_pose if performed. Then if you try to undock the robot, a staging pose will be moved by offset in X axis of a dock framed. The moved frame should be in a -X direction if backward docking was performed. If you maximize the video then there is a visible staging pose frame for docking and the incorrect staging pose frame for undocking.

Jakubach:
There is also another thing that should be considerder.
In `getStagingPose` method inside `simple_charging_dock.cpp` file there is a below code:
```
  // If not using detection, set the dock pose as the given dock pose estimate
  if (!use_external_detection_pose_) {
    // This gets called at the start of docking
    // Reset our internally tracked dock pose
    dock_pose_.header.frame_id = frame;
    dock_pose_.pose = pose;
  }
```

When using `use_external_detection_pose_` and also `backward_blind` then that code is not executed before detection of the docking station. So if I firstly make a undocking then the pose and frame_id are not initialized yet and robot doesn't recognize that it it's in docking pose.

Jakubach:
There is one more thing to consider probably.
I made a test where I started from `undockRobot` method and I was getting an error message:
```
[coverage_navigation.py-1] Waiting for 'UndockRobot' action server
[coverage_navigation.py-1] Undocking from dock of type: robot_dock...
[coverage_navigation.py-1] Undocking request was rejected!
```

Is it caused by `backward_blind` parameter with `use_external_detection_pose_`?

The problem unside `undockRobot` method was with checking the condition `dock->isDocked()`. Inside `isDocked` method 
below condition doesn't allow for procedure of undocking (even if I set the frame in my params file similarly to the `nova_carter_demo`).

```
  if (dock_pose_.header.frame_id.empty()) {
    // Dock pose is not yet valid
    return false;
  }
```
I though that the problem is inside `undockRobot` method because `isDocked` was tested before execution of `getStagingPose` method. But that didn't solve problem fully. Inside `getStagingPose` I had also to comment the condition about `use_external_detection_pose_` allowing then to always set frame. It's a quick fix but I am not sure yet how to fix that properly.

SteveMacenski:
> The question is, should robot after docking go back to the real staging pose or it could go back to the rotated pose (that after initial rotation).

I'd say it should go back to the "real" staging pose, so backup and then rotate so its back in the same state as it started in before calling the docking behavior.

ajtudela:
The `getStagingPose` method depends on the type of dock you are using. If you're using the `SimpleChargingDock` class and your dock is backwards, you need to set your `staging_x_offset` to positive and your `staging_yaw_offset` to `Pi`. Otherwise you are telling the docking_server that your staging pose is behind your robot. Not in front of it.

Jakubach:
Thank you. That fixed the second issue. The first issue is still there and when I start simulation with `undockRobot` (provided in previous post) in `nav2_simple_command` then I get a message that robot is not in docking station. The method `isDocked` returns `False` in that if statement:
```
  if (dock_pose_.header.frame_id.empty()) {
    // Dock pose is not yet valid
    RCLCPP_ERROR(node_->get_logger(), "EMPTY");
    return false;
  }
```

Is that still a problem with configuration?

Jakubach:
@SteveMacenski I would like to update work on that PR. I am not sure what should be corrected in the code to make it pass, could you help with pointing the expected changes?

ajtudela:
@Jakubach, we made some changes to the docking server here #5079 and #5097.

If you have time to finish this PR, could you please rebase main to fix CI?

SteveMacenski:
@ajtudela you may want to take it on if the author doesn't respond. Its been a few months and it may be that they have moved onto other projects

ajtudela:
Yeap, if Jakubach doesn't respond within a few days, I'll cherry-pick these commits and go from there.

Jakubach:
@ajtudela Please feel free to continue this because I don't have the time currently time to continue work on this.

ajtudela:
@SteveMacenski since the backward parameter is now inside the docking plugin, should the new ```backward_blind``` parameter also be inside the plugin?

SteveMacenski:
Yes, I think that makes sense. Though, I think that needs to be used somewhere since we need to call the controller to rotate 180 from the detected dock orientation to start reversing. Maybe it is exposed as `bool Dock::RotateToReverseDocking()` (that's a really crappy name, but you get the idea) which is called after the initial detection to know if we want to rotate 180 from the dock detected pose.

SteveMacenski:
https://github.com/ros-navigation/navigation2/pull/5153 to supersede

Issue #4404, opened 2024-06-08
Nav2 Docking: Optional Collision Checking on undocking / docking


SteveMacenski:
Should be straight forward: check the future trajectory from the graceful controller for collisions, ignoring within some tolerance of the dock itself (since we're making or in contact there). We can use the costmap collision topic object from the costmap package analog to the Behavior Server's use.

ajtudela:
Assign me this as it is the same behavior mentioned here: https://github.com/open-navigation/opennav_docking/pull/43. I can do both at the same time.

SteveMacenski:
Lets add this in a separate PR containing just this work though, don't want to have franken-PRs that are too large to review / merge. Smaller bite sized pieces are always better. Then that can be used in your work. Might make sense for this to be an independent module used by the controller or something so that it can be unit-tested and deployed elsewhere as needed easily

SteveMacenski:
Merging imminent, thanks @ajtudela this is a great feature!

PR #4752, opened 2024-11-12
Added collision detection for docking

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4404 |
| Robotic platform tested on | Morphia |

---

* Added optional collision checking on docking / undocking.

---

#### For Maintainers: 

### DIFF
--- a/nav2_docking/opennav_docking/src/controller.cpp
+++ b/nav2_docking/opennav_docking/src/controller.cpp
@@ -64,7 +64,7 @@ Controller::Controller(
   nav2_util::declare_parameter_if_not_declared(
     node, "controller.simulation_time_step", rclcpp::ParameterValue(0.1));
   nav2_util::declare_parameter_if_not_declared(
-    node, "controller.collision_tolerance", rclcpp::ParameterValue(0.30));
+    node, "controller.dock_collision_threshold", rclcpp::ParameterValue(0.3));
 
   node->get_parameter("controller.k_phi", k_phi_);
   node->get_parameter("controller.k_delta", k_delta_);
@@ -91,7 +91,7 @@ Controller::Controller(
     std::string costmap_topic, footprint_topic;
     node->get_parameter("controller.costmap_topic", costmap_topic);
     node->get_parameter("controller.footprint_topic", footprint_topic);
-    node->get_parameter("controller.collision_tolerance", collision_tolerance_);
+    node->get_parameter("controller.dock_collision_threshold", dock_collision_threshold_);
     configureCollisionChecker(node, costmap_topic, footprint_topic, transform_tolerance_);
   }
 
@@ -165,15 +165,15 @@ bool Controller::isTrajectoryCollisionFree(
       nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose) :
       std::hypot(next_pose.pose.position.x, next_pose.pose.position.y);
 
-    // If this distance is greater than the collision_tolerance, check for collisions
-    auto projected_pose = nav_2d_utils::poseToPose2D(local_pose.pose);
+    // If this distance is greater than the dock_collision_threshold, check for collisions
     if (use_collision_detection_ &&
-      dock_collision_distance > collision_tolerance_ &&
-      !collision_checker_->isCollisionFree(projected_pose))
+      dock_collision_distance > dock_collision_threshold_ &&
+      !collision_checker_->isCollisionFree(nav_2d_utils::poseToPose2D(local_pose.pose)))
     {
       RCLCPP_WARN(
         logger_, "Collision detected at pose: (%.2f, %.2f, %.2f) in frame %s",
-        projected_pose.x, projected_pose.y, projected_pose.theta, fixed_frame_.c_str());
+        local_pose.pose.position.x, local_pose.pose.position.y, local_pose.pose.position.z,
+        local_pose.header.frame_id.c_str());
       trajectory_pub_->publish(trajectory);
       return false;
     }
@@ -226,8 +226,8 @@ Controller::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
         projection_time_ = parameter.as_double();
       } else if (name == "controller.simulation_time_step") {
         simulation_time_step_ = parameter.as_double();
-      } else if (name == "controller.collision_tolerance") {
-        collision_tolerance_ = parameter.as_double();
+      } else if (name == "controller.dock_collision_threshold") {
+        dock_collision_threshold_ = parameter.as_double();
       }
 
       // Update the smooth control law with the new params
--- a/nav2_docking/opennav_docking/src/controller.cpp
+++ b/nav2_docking/opennav_docking/src/controller.cpp
@@ -31,6 +31,7 @@ Controller::Controller(
 : tf2_buffer_(tf), fixed_frame_(fixed_frame), base_frame_(base_frame)
 {
   logger_ = node->get_logger();
+  clock_ = node->get_clock();
 
   nav2_util::declare_parameter_if_not_declared(
     node, "controller.k_phi", rclcpp::ParameterValue(3.0));
@@ -96,7 +97,7 @@ Controller::Controller(
   }
 
   trajectory_pub_ =
-    node->create_publisher<nav_msgs::msg::Path>(node->get_name() + std::string("/trajectory"), 1);
+    node->create_publisher<nav_msgs::msg::Path>("docking_trajectory", 1);
 }
 
 Controller::~Controller()
@@ -122,12 +123,25 @@ bool Controller::isTrajectoryCollisionFree(
   // Visualization of the trajectory
   nav_msgs::msg::Path trajectory;
   trajectory.header.frame_id = base_frame_;
+  trajectory.header.stamp = clock_->now();
 
   // First pose
   geometry_msgs::msg::PoseStamped next_pose;
   next_pose.header.frame_id = base_frame_;
   next_pose.pose.orientation.w = 1.0;
   trajectory.poses.push_back(next_pose);
+
+  // Get the transform from base_frame to fixed_frame
+  geometry_msgs::msg::TransformStamped base_to_fixed_transform;
+  try {
+    base_to_fixed_transform =
+      tf2_buffer_->lookupTransform(fixed_frame_, base_frame_, tf2::TimePointZero);
+  } catch (tf2::TransformException & ex) {
+    RCLCPP_ERROR(
+      logger_, "Could not get transform from %s to %s: %s",
+      base_frame_.c_str(), fixed_frame_.c_str(), ex.what());
+    return false;
+  }
 
   // Generate path
   for (double t = 0; t < projection_time_; t += simulation_time_step_) {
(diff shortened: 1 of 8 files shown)


nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -55,16 +80,98 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
...
+    std::string costmap_topic, footprint_topic;
+    node->get_parameter("controller.costmap_topic", costmap_topic);
+    node->get_parameter("controller.footprint_topic", footprint_topic);
+    node->get_parameter("controller.transform_tolerance", transform_tolerance);
+    node->get_parameter("controller.collision_tolerance", collision_tolerance_);
+    configureCollisionChecker(node, costmap_topic, footprint_topic, transform_tolerance);
+  }
+
+  trajectory_pub_ =
+    node->create_publisher<nav_msgs::msg::Path>(node->get_name() + std::string("/trajectory"), 1);
```
SteveMacenski, line 99:
```suggestion
    node->create_publisher<nav_msgs::msg::Path>("docking_trajectory", 1);
```

More in line with conventions

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -92,6 +199,10 @@ Controller::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
         v_angular_max_ = parameter.as_double();
       } else if (name == "controller.slowdown_radius") {
         slowdown_radius_ = parameter.as_double();
+      } else if (name == "controller.projection_time") {
+        projection_time_ = parameter.as_double();
+      } else if (name == "controller.simulation_time_step") {
+        simulation_time_step_ = parameter.as_double();
```
SteveMacenski, line 205: collision_tolerance?

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -39,6 +48,22 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
     node, "controller.v_angular_max", rclcpp::ParameterValue(0.75));
   nav2_util::declare_parameter_if_not_declared(
     node, "controller.slowdown_radius", rclcpp::ParameterValue(0.25));
+  nav2_util::declare_parameter_if_not_declared(
+    node, "controller.use_collision_detection", rclcpp::ParameterValue(true));
```
SteveMacenski, line 52: Q: Should this be default on or off?

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -55,16 +80,98 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
...
+    // Transform pose from base_frame into fixed_frame
+    geometry_msgs::msg::PoseStamped local_pose = next_pose;
+    local_pose.header.stamp = rclcpp::Time(0);
+    tf2_buffer_->transform(local_pose, local_pose, fixed_frame_);
+
+    // Check for collisions at the projected pose
+    auto projected_pose = nav_2d_utils::poseToPose2D(local_pose.pose);
+    double distance = nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose);
+    if (use_collision_detection_ &&
+      !collision_checker_->isCollisionFree(projected_pose) && distance > collision_tolerance_)
```
SteveMacenski, line 150: I don't understand the role of `collision_tolerance_` - can you explain this `0.3` for both the docking and undocking case?

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -55,16 +80,98 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
...
+    // Apply velocities to calculate next pose
+    next_pose.pose = control_law_->calculateNextPose(
+      simulation_time_step_, target_pose, next_pose.pose, backward);
+
+    // Add the pose to the trajectory for visualization
+    trajectory.poses.push_back(next_pose);
+
+    // Transform pose from base_frame into fixed_frame
+    geometry_msgs::msg::PoseStamped local_pose = next_pose;
+    local_pose.header.stamp = rclcpp::Time(0);
```
SteveMacenski, line 143: You should use `now()` from the start of the loop rather than `0` for better time synchronization. Store the the current time in L125 and use that for this step so its consistent

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -55,16 +80,98 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
...
+      !collision_checker_->isCollisionFree(projected_pose) && distance > collision_tolerance_)
+    {
+      RCLCPP_INFO(
+        logger_, "Collision detected at pose: (%.2f, %.2f, %.2f) in frame %s",
+        projected_pose.x, projected_pose.y, projected_pose.theta, fixed_frame_.c_str());
+      return false;
+    }
+  }
+
+  trajectory_pub_->publish(trajectory);
```
SteveMacenski, line 159: Also publish if in collision to see the collision stopping correctly

SteveMacenski (changes requested):
Only addition I would like to see (which I think is the intent of the `collision_tolerance_` param) is a parameter for a tolerance to the dock or from the dock to ignore collisions. If we're checking within `N` distance of the dock (or proxy of distance along the trajectory control law), ignore collisions since we're literally going to collide with the dock.

So while docking, its at the end (and at the start of the trajectory for undocking) 

Otherwise, this looks very sensible!

nav2_docking/opennav_docking/src/controller.cpp
SteveMacenski: ^ might be why CI is having a test fail

nav2_docking/opennav_docking/src/controller.cpp
ajtudela: Good point, I think the only reason I didn't add it was because collision detection is not toggleable at runtime, but yes, if it's enabled, the *collision_tolerance* could be set as a dynamic parameter.

nav2_docking/opennav_docking/src/controller.cpp
ajtudela: Good question. I set default to true because I think it's important not to collide with objects. 😅

Which brings me to the next question regarding testing. any advice on how to test collision_checker_->isCollisionFree()?

SteveMacenski: That should be already tested by the costmap topic collision checker's own tests, so you can assume internally it works for the purpose of these tests. If you want to test that the docking works with collision checking (probably wise), you'd need to publish a local costmap / footprint in the test so that the collision checker object can obtain it. You could do something like dock 1x with higher but valid cost, then dock 1x with some legal points in the way between the staging and dock pose, then dock 1x with lethal at the dock (which would be the normal use case since it would pick up the dock itself)

nav2_docking/opennav_docking/src/controller.cpp
SteveMacenski: Does this properly handle the dock (end of trajectory stop checking) vs undock (start of trajectory don't check) cases? For undocking, we want to check the full trajectory to the prestaging pose, except the start segment

SteveMacenski: I think this is what I remarked upon in my review main comment -- I think this is implementing what I think is missing, just needs some tweaks :-) 

I'd rename it too to `dock_collision_distance` --> to be clear this is the distance to dock collision to stop checking.

SteveMacenski: I think this is the last major concern!

ajtudela: Currently, it skip the last part of the trajectory for docking and undocking. Although for undocking it should check the whole trajectory up to the staging pose.

I made a quick video yesterday morning. It detects the obstacle, back to the staging pose. When the obstacle is removed, dock.
[docking_obstacle_bueno.webm](https://github.com/user-attachments/assets/95a44185-45ab-4a72-8aed-0dd3c537ccb7)

SteveMacenski: > Currently, it skips the last part of the trajectory for docking and undocking. Although for undocking it should check the whole trajectory from the dock up to the staging pose.

I think there are different ways it should be handled:

Docking: From current pose --> Dock minus some tolerance near the dock (bc we're making contact)

Undocking: From current pose minus some tolerance (bc we're currently in contact) --> Undocking pose

I think the handling for docking vs undocking is inverted. Docking we ignore the *ending* bit for collision checking. Undocking we ignore the *starting* bit for collision checking, since we start presumably in collision with the dock.

Though -- I suppose we could parameterize whether we ignore anything. Presumably some infrastructure docking might not make contact, so we can collision check the full thing

SteveMacenski:
@ajtudela do you want me to review again?

ajtudela:
> @ajtudela do you want me to review again? I see updates but wasn't sure if we came to agreement on the trajectory checking for dock/undock (or if what I'm asking for isn't sensible)

Yes, please. I came to the same conclusion as you when doing the test: skip the last part of the trajectory when docking and the first part when undocking.

It's implemented in the latest commit, and I've added new tests to check for collisions in the 4 scenarios: dock/undock forward/backward. 

If you can review it, I'll fix the python test and your possible changes tomorrow.

Thanks!

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -39,6 +49,22 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
     node, "controller.v_angular_max", rclcpp::ParameterValue(0.75));
   nav2_util::declare_parameter_if_not_declared(
     node, "controller.slowdown_radius", rclcpp::ParameterValue(0.25));
+  nav2_util::declare_parameter_if_not_declared(
+    node, "controller.use_collision_detection", rclcpp::ParameterValue(true));
+  nav2_util::declare_parameter_if_not_declared(
```
SteveMacenski, line 54: Need to add this to the migration guide + add the new parameters to the configuration guide + add new parameters to the `nav2_params.yaml` for example. The `collision_tolerance` especially should be described with some care about what it represents. I think this should _probably_ be renamed to something like `dock_collision_checking_tolerance` or `dock_collision_distance` or something more clear

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -55,16 +81,120 @@ Controller::Controller(const rclcpp_lifecycle::LifecycleNode::SharedPtr & node)
...
+    // This avoids false positives when the robot is at the dock
+    double dock_collision_distance = is_docking ?
+      nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose) :
+      std::hypot(next_pose.pose.position.x, next_pose.pose.position.y);
+
+    // If this distance is greater than the collision_tolerance, check for collisions
+    auto projected_pose = nav_2d_utils::poseToPose2D(local_pose.pose);
+    if (use_collision_detection_ &&
+      dock_collision_distance > collision_tolerance_ &&
+      !collision_checker_->isCollisionFree(projected_pose))
```
SteveMacenski, line 172:
```suggestion
      !collision_checker_->isCollisionFree(nav_2d_utils::poseToPose2D(local_pose.pose)))
```

then you can remove L169

Super nit pick, only since you're going to go back in to update the tests

SteveMacenski (changes requested):
Otherwise, pretty much ready to rumble!

ajtudela:
Update dock here: https://github.com/ros-navigation/docs.nav2.org/pull/610

ajtudela:
Can be backported to humble?

SteveMacenski (approved):
That's a nice solution to the docking/undocking problem - very elegant

ajtudela:
> That's a nice solution to the docking/undocking problem - very elegant
> 
> > Can be backported to humble?
> 
> I'm going to say 'yes'. It technically breaks the controller's constructor & `computeVelocityCommand`, but that's not a public facing object and this is an important update that Humble users will want over the next ~3 years of LTS support. This is still a pretty new package, so I'm willing to make the exception here

Great. I'll make another PR in the other repo.

SteveMacenski:
Thanks, this is great work!

Issue #4736, opened 2024-10-29
Use BT.CPP recommended sleep for event-driven trees?


tonynajjar:
#### Feature description

I would be surprised if I'm the first one to bring up this discussion but I didn't find it anywhere. Why aren't we using the [recommended](https://www.behaviortree.dev/docs/tutorial-basics/tutorial_04_sequence#event-driven-trees) `Tree::sleep()` from BT.CPP instead of the [ROS `rclcpp::WallRate `](https://github.com/ros-navigation/navigation2/blob/44a69ccde93514ed1a962b7f18cb101ec671d93a/nav2_behavior_tree/src/behavior_tree_engine.cpp#L52) to leverage event-driven trees?

I guess an obvious reason would be the better integration of `rclcpp::WallRate` with simulation time. But shouldn't we thrive to have all the benefits that both offer?

In a couple of instances we use [emitWakeUpSignal](https://github.com/ros-navigation/navigation2/pull/4128/files#diff-8e4ad45054e440b04bf332cf96752ac6ae3d5b236829ccd2b593570c64a80c62R381), but afaik these should be useless without `Tree::sleep()` right @facontidavide?

SteveMacenski:
Sorry for the delay, ROSCon, ROS-I, and so forth kept me from a computer. 

> Why aren't we using the [recommended](https://www.behaviortree.dev/docs/tutorial-basics/tutorial_04_sequence#event-driven-trees) Tree::sleep()

Because it didn't exist in BT.CPP when we created that bit of software :-) My immediate thought is that I'm perfectly happy moving to the `Tree::sleep()` API. We use the `WallRate` for this operation, so simulation time isn't even considered for the BT ticking.

After looking at the BT.CPP implementation, I don't see any downsides to that change. @tonynajjar want to submit a PR?

facontidavide:
Exactly! `Tree::sleep()` was created to potentially lower the latency, compared to a regular "sleep".

`Tree::sleep()` is basically a a [std::condition_variable::wait_for](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for) under the hood.

This allows, as @tonynajjar point out, to wake uyp immediately when ANY node invokes `emitWakeUpSignal`.

Note that in general I suggest people to tick at 1 KHz using `Tree::sleep()`. Overhead should be minimal.

[...]

PR #4761, opened 2024-11-22
Use BT.CPP Tree::sleep

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4736 |
| Primary OS tested on | Ubuntu, |
| Robotic platform tested on | Custom simulation |

---

* Replaced BT's engine `rclcpp::WallRate::sleep` with `Tree::sleep()` to leverage event-driven trees

---

#### For Maintainers: 


tonynajjar:
I guess it's as simple as that. I created the PR to see already if CI is happy. Will run my stack with this change and see if I notice anything weird.

nav2_behavior_tree/src/behavior_tree_engine.cpp
```diff
@@ -64,13 +63,7 @@ BehaviorTreeEngine::run(
 
       onLoop();
 
-      if (!loopRate.sleep()) {
-        RCLCPP_DEBUG_THROTTLE(
-          rclcpp::get_logger("BehaviorTreeEngine"),
-          *clock_, 1000,
-          "Behavior Tree tick rate %0.2f was exceeded!",
-          1.0 / (loopRate.period().count() * 1.0e-9));
-      }
```
tonynajjar, line 73: We are losing the feature of knowing if we are not reaching the defined tick rate though :/
On the other hand,  with some modification in BT.cpp, we could get from `Tree::sleep`, whether we interrupted the sleep.

SteveMacenski (approved):

nav2_behavior_tree/src/behavior_tree_engine.cpp
SteveMacenski: software implementation-wise, I'm approving this (obviously :wink: )

It does hit me strange yes that we're not trying to smooth out rates. What's nice about rclcpp::Rate is that it'll only sleep enough to maintain a loop rate. This'll sleep for an amount of time regardless of the time it took since the last loop. I think that's less a debugging as much as functionally important. If you say run at 100hz, you don't want to find it running at 92hz or something just due to the fixed sleep being `0.01s` not considering the BT's execution duration itself. The **logging** of it I agree is mostly for debugging, but I figure if we're going to do this, may as well do that (but its not mission critical)

So to get the best behavior, it seems like we should have a `BT::LoopRate` object perhaps that mimic's ROS' version (we can leave sim time out, just wall time) so that it tracks the rate and calls `sleep()` on the `timeout - loop_time` to maintain a fixed loop rate, less interrupted. 

Since that's pretty standalone, that could go back into BT.CPP easily I think, but we could stage it in Nav2 to get it out sooner than later. Then, open a PR with BT.CPP for its review cycle / merge / release that we could migrate to once out in binaries.

nav2_util/src/bt_loop_rate.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include "nav2_util/bt_loop_rate.hpp"
+
+namespace nav2_util
+{
+
+LoopRate::LoopRate(const rclcpp::Duration & period, BT::Tree * tree)
+: clock_(std::make_shared<rclcpp::Clock>(RCL_STEADY_TIME)), period_(period),
+  last_interval_(clock_->now()), tree_(tree)
+{}
```
tonynajjar, line 23: I would have liked to inherit from WallRate to reuse some code but its attributes are private so I can't use them anyway

nav2_util/include/nav2_util/bt_loop_rate.hpp
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+#include "rclcpp/rclcpp.hpp"
+#include "behaviortree_cpp/bt_factory.h"
+#include "behaviortree_cpp/behavior_tree.h"
+
+namespace nav2_util
+{
+
+class LoopRate
```
SteveMacenski, line 25: `BTLoopRate` --> move to the `nav2_behavior_tree` package. I don't want the BT.CPP dependencies to stretch too low considering the utils are used by other non-Nav2 packages & we have a nice separation of Nav2 BT utilities in the `nav2_behavior_tree` package already.

The naming to make more clear, but also if you move to the `nav2_behavior_tree::LoopRate` namespace, we can drop the `BT` prefix since the namespace indicates that this is a behavior tree specific object.

SteveMacenski (changes requested):
LGTM, lets just move this object to `nav2_behavior_tree` where we have our other BT utilities and base functionality

SteveMacenski (approved):

SteveMacenski:
@tonynajjar looks like CI doesn't like the linty-freshness of the updates :laughing: 

otherwise, will merge.

@facontidavide: This seems like a useful util in BT.CPP - would you accept a PR with it?

tonynajjar:
CI is still looking for `bt_loop_rate.hpp` although it was renamed. I guess some cache issue. Is it safe to merge anyway?

SteveMacenski:
@tonynajjar bump the v28 to v29 (there are 3x) in this file: https://github.com/ros-navigation/navigation2/blob/main/.circleci/config.yml

We need to fix it first since this will persist to `main` if not resolved here first. Yes, looks like a usual caching issue when moving files around. It doesn't like that :laughing:

Issue #4757, opened 2024-11-19
Nav2 Regulated Pure Pursuit set with Ackermann model skips driving maneuvers given by path planner


andreacelani:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04.5 
- ROS2 Version:
  - Humble 
- Version or commit hash:
  - 1.1.16 
- DDS implementation:
  - rmw_fastrtps_cpp 

#### Issue Description:

Nav2 Regulated Pure Pursuit set with Ackermann model skips driving maneuvers given by path planner. It jumps to the farthest point in the lookahaed radius skipping fundamental maneuvers to follow global path .

#### Steps to reproduce issue

1- Configure an Ackermann steering vehicle with the Nav2 stack, kinematically feasible planner, like Smac Planner, and RPP controller. In my case the robot model is an Ackermann based robot of a real agricultural tractor. Wheel base around 2 meters, minimum turning radius 4.2 meters, track 1.6 meters rear and 1.4 meters front.

2- Plan a path without re-planning (one single plan).

3- Observe the vehicle's behavior while executing the planned path, particularly during the reverse segments.

Here are the parameters of controller.
Note: these settings turn on the `use_regulated_linear_velocity_scaling` and the `min_lookahead_dist` has been set to very low value. The aim is to show that even if the controller could search with very limited horizon it will skip large portion of planned path.

```
controller_server:
  ros__parameters:
    use_sim_time: True
    controller_frequency: 10.0
    min_x_velocity_threshold: 0.01
    min_y_velocity_threshold: 0.5
    min_theta_velocity_threshold: 0.01
    progress_checker_plugins: ["progress_checker"]  # progress_checker_plugin: "progress_checker" For Humble and older
    goal_checker_plugins: ["goal_checker"]
    controller_plugins: ["FollowPath"]

    progress_checker:
      plugin: "nav2_controller::SimpleProgressChecker"
      required_movement_radius: 0.5
      movement_time_allowance: 10.0
    goal_checker:
      plugin: "nav2_controller::SimpleGoalChecker"
      xy_goal_tolerance: 0.1
      yaw_goal_tolerance: 0.05
      stateful: True
    FollowPath:
      plugin: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
      desired_linear_vel: 0.25
      lookahead_dist: 5.0
      min_lookahead_dist: 0.1
      max_lookahead_dist: 5.0
[...]

PR #4763, opened 2024-11-24
Better management of RPP path pruning when cusp points are detected

andreacelani:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | ([#4757](https://github.com/ros-navigation/navigation2/issues/4757)) |
| Primary OS tested on | (Ubuntu 22.04.5) |
| Robotic platform tested on | (custom Gazebo simulation with Ackermann model) |

---

the contribution make RPP controller follows better the path given by planner. There is no skip of path points even with tight maneuvers.
This apply to all models (diff drive, ackermann and omni)

[fixed RPP.webm](https://github.com/user-attachments/assets/d03b93ab-bf32-4313-9834-a6fe0f74e31c)

Added the following parameters. They are the same found in MPPI controller
- `inversion_xy_tolerance`
- `inversion_yaw_tolerance`

the parameters are useful to terminate an inversion and start the following the next portion of the path

---

original RPP coding style has been follow (one single file cpp file). MPPI has more structured organization with a dedicated path_handler.cpp file 
-->

#### For Maintainers: 


nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -244,20 +245,118 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
+    * It takes and returns radians.
+    * @param angles Angles to normalize
+    * @return normalized angles
+    */
+  template<typename T>
+  auto normalize_angles(const T & angles)
+  {
+    auto && theta = fmod(angles + M_PI, 2.0 * M_PI);
+    if (theta < 0)
+      theta += 360;
```
SteveMacenski, line 260: Isn't this in radians?

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
```diff
@@ -244,20 +245,118 @@ class RegulatedPurePursuitController : public nav2_core::Controller
...
+    *
+    * The result
+    * would always be -pi <= result <= pi.  Adding the result
+    * to "from" will always get you an equivelent angle to "to".
+    * @param from Start angle
+    * @param to End angle
+    * @return Shortest distance between angles
+    */
+  template<typename F, typename T>
+  auto shortest_angular_distance(
```
SteveMacenski, line 279: I believe both of these functions can be used from `angles` rather than reimplemented https://github.com/ros/angles

nav2_regulated_pure_pursuit_controller/test/test_regulated_pp.cpp
```diff
@@ -166,38 +160,6 @@ TEST(RegulatedPurePursuitTest, createCarrotMsg)
   EXPECT_EQ(rtn->point.z, 0.01);
 }
 
-TEST(RegulatedPurePursuitTest, findVelocitySignChange)
```
SteveMacenski, line 169: make sure to add tests for your new methods (`isWithinInversionTolerances` `prunePlan` `removePosesAfterFirstInversion` `findFirstPathInversion`)

SteveMacenski (changes requested):
Heed the CI bots - we need this in `main` before we can have it in `humble` so that all new distributions and current releases can have this update (i.e. `jazzy` which I assume you'll also want eventually :wink: ). Also on linting and DCO sign offs for quality control. 

I did a good sweep at it and generally look good to me other than missing test coverage / linting issues. I'll do a detailed sweep after the `main` version of this PR is opened, but generally seems good. I assume you tested this fix?

SteveMacenski:
@andreacelani any update?

andreacelani:
@SteveMacenski sorry I had not time, heavy period at work! :sweat_smile:  I'm moving to main and add tests to new methods

SteveMacenski:
Thank you @andreacelani :heart: 

Just checking in :-)

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
andreacelani: yes radians, I translated a function from degree to radians but it looks like I forgot to update `theta += 360` to `theta += 2.0 * M_PI`

nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp
andreacelani: I suppose so. I follow MPPI controller utils.hpp implementation in humble. I see Nav2 Rolling switched to `angles` dependency

andreacelani:
@SteveMacenski I trying to switch to Rolling but my ROS2 is humble. Any advice/workaround to build the main (rolling) while in humble? Or should I install everything in a VM? Thank you

SteveMacenski:
Use a docker container, typically. OSRF has Jazzy / Rolling docker images that you can mount your workspace into. I have some docs on it if you need help setting up https://docs.nav2.org/tutorials/docs/docker_dev.html

SteveMacenski:
@andreacelani I'm hoping this is something you can get done this month, are you able to?

andreacelani:
I hope so, I'm really slowly but I'm working on it. Now I'm fighting with PoseStampedArray message that you can find in Nav2 rolling but not in Ros2 rolling docker image. I'll get out of it 😅

SteveMacenski:
What I did was clone geometry2 into my workspace, then `colcon build --packages-up-to geometry_msgs --parallel-workers 1` and then source the workspace, finally `colcon build` the full space and you should be good to go (its annoying, but getting that built and sourced first makes later things use the workspace version of the package on the next re-build)

mikeferguson:
I found I had to delete the build/install folders after adding common_interfaces to the workspace - my hypothesis is that there was some left over Cmake generated files that were still pointing to /opt/ros for geometry_msgs stuff (kept getting errors with no header found for PoseStampedArray). Once I cleaned out the build/install folders, build ran fine (even doing it all at once, with unlimited workers)

SteveMacenski:
Indeed, I would also expect that, sorry for not mentioning that :frowning:

mini-1235:
I have been working on this issue the past few days to push https://github.com/ros-navigation/navigation2/pull/5446 forward, I think I am able to see a before-and-after difference here.

Should I:
- Open a new PR targeting main to add this in RPP?
- Add this feature in #5446 directly?

SteveMacenski:
If the PRs aren't directly attached to one another, a different PR would be good to keep things small and sharp. Thanks!

Are there other controllers that should have it updated too? I think MPPI has this already, so DWB, Graceful?

mini-1235:
> If the PRs aren't directly attached to one another, a different PR would be good to keep things small and sharp. Thanks!

To add this feature in RPP, we will need to add a couple parameters. Like MPPI, I think we can refactor the parameter handler in RPP, and pass it to the path handler? Or even move the parameter handler to nav2_util, so we can have a nav2_util::ParameterHandler object in every controller?

> Are there other controllers that should have it updated too? I think MPPI has this already, so DWB, Graceful?

I think so. But, to be honest, I haven’t worked with DWB or Graceful before, so I would appreciate your advice on how best to approach those

SteveMacenski:
@mini-1235 yeah that sounds good! Lets start with RPP but if we find that they are _exactly_ the same, then abstracting them might be useful, especially if we look at DWB and Graceful and find that we could use them there too.

> I haven’t worked with DWB or Graceful before, so I would appreciate your advice on how best to approach those

I'd look at their respective path handlers (https://github.com/ros-navigation/navigation2/blob/main/nav2_graceful_controller/src/path_handler.cpp) and see if this would mold well into it or its doing anything unique / unique inputs or outputs that need to be considered. I'm guessing the answer is 'no'

SteveMacenski:
Closing this for now due to lack of response from the original author. Some work is happening in Main that includes this feature so it is being addressed another way. If the author would like to implement the review comments, happy to re-review and continue from there!

PR #4764, opened 2024-11-25
Add IsStoppedBTNode

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 
- [x] Check that any new parameters added are updated in docs.nav2.org


tonynajjar:
@SteveMacenski are you open to this contribution? Still a draft but almost there

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_stopped_condition.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+  std::string topic_name_;
+  std::atomic<bool> is_stopped_;
+  double velocity_threshold_;
+  std::chrono::milliseconds time_stopped_threshold_;
+  rclcpp::Time stopped_stamp_;
+  std::mutex mutex_;
+
+  // Listen to odometry
+  rclcpp::Subscription<nav_msgs::msg::Odometry>::SharedPtr odom_sub_;
```
SteveMacenski, line 97: `OdomSmoother` nav2 util instead? I believe this is handed to the navigators as well on the BT blackboard so you don't need duplicate subscriptions. You can use the one already at play https://github.com/ros-navigation/navigation2/blob/main/nav2_bt_navigator/src/bt_navigator.cpp#L126-L129

SteveMacenski (changes requested):
Groot (https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/nav2_tree_nodes.xml) and doc updates are needed (the usual bits!).

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_stopped_condition.hpp
tonynajjar: Hmm, for such a BT node I don't know if I want my odom to be smoothed though. If the velocity is 0, I want to know it as soon as possible. As for fluctuations, there is the `time_stopped_threshold` that accounts for that.

Side-note: I got inspired from IsStuckCondition which also doesn't use the OdomSmoother, is that an oversight?

nav2_util/include/nav2_util/odometry_utils.hpp
```diff
@@ -68,13 +68,44 @@ class OdomSmoother
    * @brief Get twist msg from smoother
    * @return twist Twist msg
    */
-  inline geometry_msgs::msg::Twist getTwist() {return vel_smooth_.twist;}
+  inline geometry_msgs::msg::Twist getTwist()
+  {
+    std::lock_guard<std::mutex> lock(odom_mutex_);
```
tonynajjar, line 73: that's pretty important right? `vel_smooth_` is not atomically updated

nav2_behavior_tree/plugins/condition/is_stopped_condition.cpp
```diff
@@ -0,0 +1,78 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+#include <chrono>
+
+#include "nav2_behavior_tree/plugins/condition/is_stopped_condition.hpp"
```
SteveMacenski, line 18: This file is missing BT unit testing

nav2_util/include/nav2_util/odometry_utils.hpp
SteveMacenski: Yeah - this is pretty old, back in the Intel + Steve days, so I think there was some R&D thinking and not production thinking back then

SteveMacenski (changes requested):
On terminal conditions (success, fail) shouldn't we update `stopped_stamp_`?

tonynajjar:
> On terminal conditions (success, fail) shouldn't we update stopped_stamp_?

With "update" I assume you mean resetting to `rclcpp::Time(0, 0, RCL_ROS_TIME)`? For FAILURE  we do. For doing it at SUCCESS, I think it's a tradeoff:

- reset at SUCCESS:  Not stateful: Next time it's ticked, it will wait again for `duration_stopped`, which could be undesired or unexpected behavior. But maybe stateful nodes is an anti-pattern so this behavior is actually expected? 
- no reset at SUCCESS: Stateful: Next time it's ticked, it won't wait "since it's been stopped for long enough already". However there could be this bug:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 1. BT node returns SUCCESS after it was stopped for long enough
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 2. robot moves and stops again before the BT node is ticked again
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 3. BT node returns SUCCESS because the moving in between was not considered

All things considered, I think you're right that resetting at success would be the better option

SteveMacenski:
Agreed, I was thinking something too like `node_->get_clock()->now() - stopped_stamp_ > rclcpp::Duration(duration_stopped_)` will always be true if we left the branch of the BT that contains this node and then come back to it after some time, if we don't reset the time after terminal conditions. Then, its a single-shot of the velocity without using your duration piece.

With that said, I think `it will wait again for duration_stopped, which could be undesired or unexpected behavior` can be largely mitigated if you have your waiting duration to be sub-100ms -- about the perception of a person to change. 100ms though should give you some readings to work based off of so any single erroneous measurement doesn't get you. 

---

Also see https://github.com/ros-navigation/docs.nav2.org/pull/612#issuecomment-2504507376 (I know looking at comments after a PR is merged gets lost in the general "merge" notification, for me at least :wink: ).

SteveMacenski (approved):
LGTM - let me know if there's anything else here from that conversation of its good to go

Though, I see a couple of lines that seem easy to cover in unit testing that are not covered: 
- `twist.header.stamp = node_->get_clock()->now();`
- Each of the `!received_odom_` condition's internals

https://app.codecov.io/gh/ros-navigation/navigation2/pull/4764?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=ros-navigation

tonynajjar:
@SteveMacenski it's theoretically ready from my side but I'll stress test it even more in the coming week so we can wait a bit before merging

SteveMacenski:
Ok, please do! Let me know when you're ready

SteveMacenski (approved):

PR #4767, opened 2024-11-26
Improvements in getCostsCallback

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -837,14 +837,11 @@ void Costmap2DROS::getCostsCallback(
...
     geometry_msgs::msg::PoseStamped pose_transformed;
-    transformPoseToGlobalFrame(pose, pose_transformed);
-    bool in_bounds = costmap->worldToMap(
-      pose_transformed.pose.position.x,
-      pose_transformed.pose.position.y, mx, my);
-
-    if (!in_bounds) {
-      response->costs.push_back(NO_INFORMATION);
-      continue;
+    if (!transformPoseToGlobalFrame(pose, pose_transformed)) {
```
tonynajjar, line 840: I've had this fail because of outdated timestamp.

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -866,6 +863,14 @@ void Costmap2DROS::getCostsCallback(
           pose_transformed.pose.position.x,
         pose_transformed.pose.position.y);
 
+      bool in_bounds = costmap->worldToMap(
```
tonynajjar, line 866: moved this because `footprintCostAtPose` does its own on-map check

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -866,6 +863,14 @@ void Costmap2DROS::getCostsCallback(
...
+        pose_transformed.pose.position.x,
+        pose_transformed.pose.position.y, mx, my);
+
+      if (!in_bounds) {
+        response->costs.push_back(LETHAL_OBSTACLE);
```
tonynajjar, line 871: In `footprintCostAtPose` if it's out-of-map, LETHAL_OBSTACLE is returned so I changed it from NO_INFORMATION to match. (In my fork I changed all of them to NO_INFORMATION but that's for my use case, I guess it's debatable what it should be)

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -55,6 +55,13 @@ void RemoveInCollisionGoals::on_tick()
 BT::NodeStatus RemoveInCollisionGoals::on_completion(
   std::shared_ptr<nav2_msgs::srv::GetCosts::Response> response)
 {
+  if (response->costs.size() != input_goals_.size()) {
```
tonynajjar, line 58: handling of the error introduced below

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -837,14 +837,11 @@ void Costmap2DROS::getCostsCallback(
...
-    transformPoseToGlobalFrame(pose, pose_transformed);
-    bool in_bounds = costmap->worldToMap(
-      pose_transformed.pose.position.x,
-      pose_transformed.pose.position.y, mx, my);
-
-    if (!in_bounds) {
-      response->costs.push_back(NO_INFORMATION);
-      continue;
+    if (!transformPoseToGlobalFrame(pose, pose_transformed)) {
+      response->costs.clear();
```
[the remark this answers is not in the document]
SteveMacenski: For the service level `success`? Yeah, that would make sense. 

Can't we just set these costs to `UNKNOWN`? That would essentially be the behavior before and if we can't transform it, it does seem like an "UNKNOWN" case. I could also see it being a total failure mode since not being able to TF transform one goal probably indicates some problems elsewhere.

nav2_costmap_2d/src/costmap_2d_ros.cpp
SteveMacenski: Good catch. Sure, lets keep this.

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
SteveMacenski: ^ `success` added would be good as you mentioned instead :-)

SteveMacenski (changes requested):

SteveMacenski:
Otherwise LGTM - please fix DCO :-)

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -857,6 +856,14 @@ void Costmap2DROS::getCostsCallback(
           pose_transformed.pose.position.x,
         pose_transformed.pose.position.y);
 
+      bool in_bounds = costmap->worldToMap(
+        pose_transformed.pose.position.x,
+        pose_transformed.pose.position.y, mx, my);
+
+      if (!in_bounds) {
+        response->costs.push_back(LETHAL_OBSTACLE);
```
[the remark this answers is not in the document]
tonynajjar: Sure why not, thanks.

SteveMacenski (approved):

PR #4771, opened 2024-11-28
Adapt GoalUpdater to update goals as well

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


tonynajjar:
@SteveMacenski as usual, still a draft but what do you think of the idea? Do you have a preference for a standalone node? `GoalUpdatedCondition` checks both `goal` and `goals` in the same node.

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -74,16 +84,29 @@ inline BT::NodeStatus GoalUpdater::tick()
     }
   }
 
+  if (!last_goals_received_.empty()) {
+    goals = last_goals_received_;
```
SteveMacenski, line 88: Codecov complains about no test coverage :-)

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -74,16 +84,29 @@ inline BT::NodeStatus GoalUpdater::tick()
...
+
   setOutput("output_goal", goal);
+  setOutput("output_goals", goals);
   return child_node_->executeTick();
 }
 
 void
 GoalUpdater::callback_updated_goal(const geometry_msgs::msg::PoseStamped::SharedPtr msg)
 {
+  std::lock_guard<std::mutex> lock(mutex_);
```
SteveMacenski, line 99: No mutex is needed since we spin in the `tick` function -- we know its not executed in parallel to access

nav2_msgs/msg/PosesStamped.msg
```diff
@@ -0,0 +1 @@
+geometry_msgs/PoseStamped[] poses
```
SteveMacenski, line 1: Is this really the most ergonomic way to handle this?

SteveMacenski (changes requested):
I know its a conversational draft, so obviously the main block in `tick` still needs the `goals` version. 

I'm not opposed to having this BT node be goal(s). I'm less certain about the introduction of a new `PosesStamped` message type used only for this application. I imagine the context of use would be for `NavigateThroughPoses` or similar, so shouldn't those be updated as well to use that [1] (among possibly others)?

Also, food for thought, this seems like a good `geometry_msgs` improvement rather than a `nav2_msgs` improvement to add `PosesStamped`. Maybe we should file a ticket over there and ask @tfoote if he's open to that contribution

[1]  https://github.com/ros-navigation/navigation2/blob/main/nav2_msgs/action/NavigateThroughPoses.action#L3

nav2_msgs/msg/PosesStamped.msg
tonynajjar: Probably not, that's one main point I wanted to have discussed. The option I see are:

1- Leave PosesStamped.msg (or maybe the better-named PoseStampedArray.msg) and adapt where necessary (e.g. NavigateThroughPoses.action)
2- Create PoseStampedArray.msg in geometry_msgs
3- (ab)use nav_msgs/Path.msg and not create any new message type

From Nav2's perspective I guess #2 would be convenient but I don't know if @tfoote wants to open the door to creating Arrays from primitive data types (although there is already [one](https://github.com/ros2/common_interfaces/blob/rolling/geometry_msgs/msg/PoseArray.msg))

@SteveMacenski we probably need an intermediary solution to not block this PR for too long. Would you rather #1 (without any committal adaptation to  NavigateThroughPoses.action) or #3? Or do you see another option?

SteveMacenski: I think there shouldn't be much if any objections to Option 2 -- can you file a ticket in `geometry_msgs` with a PR that implements and tag me?

tonynajjar: Done https://github.com/ros2/common_interfaces/pull/262/files

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -73,8 +83,22 @@ inline BT::NodeStatus GoalUpdater::tick()
         last_goal_received_time.seconds(), goal_time.seconds());
     }
   }
-
   setOutput("output_goal", goal);
+
+  if (last_goals_received_.header.stamp != rclcpp::Time(0) && !last_goals_received_.poses.empty()) {
+    auto last_goals_received_time = rclcpp::Time(last_goals_received_.header.stamp);
```
tonynajjar, line 89: What timestamp-checking logic would you have here? In this implementation I'm just checking the top-level one. We could also check all the individual goals timestamps and reject the update if at least one is outdated

SteveMacenski: I think that we should check if any are newer. That would indicate that a new message has come in that was not previously known.

I read this check as looking at the topic's goal and the input port's goal and selecting the topic's if its newer (and a topic update is present). So, I think the analog here would be to check the goals list's header.

I know in the `geometry_msgs/PoseStampedArray` PR we discussed removing it because we couldn't think of a good reason to include it -- I think _this_ is the good reason to include it. You could have that message type defining a set of goals so the goals themselves have their own stamps as dynamic constraints, yet the global header is referring to the time the goals themselves were computed. 

Sorry to flip flop :(

tonynajjar: I did not get what you are proposing to compare exactly? Is it if any of the topic PoseStamped are more recent than any of the input PoseStamped? Or if any of the topic PoseStamped are more recent than the stamp of input PoseStampedArray?

If we are keeping the top-level header with the meaning "the time the goals themselves were computed", then I don't see why we would not just keep it as is (compare current and incoming top-level stamps).

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -73,8 +83,27 @@ inline BT::NodeStatus GoalUpdater::tick()
         last_goal_received_time.seconds(), goal_time.seconds());
     }
   }
-
   setOutput("output_goal", goal);
+
+  if (last_goals_received_.header.stamp != rclcpp::Time(0) && !last_goals_received_.poses.empty()) {
```
tonynajjar, line 88: I have a question about the intended behavior when `last_goal_received_.header.stamp == rclcpp::Time(0)`. When the topic-received goal has no timestamp, do we want to use that goal or reject it? Currently we don't use it but also don't show any warning. Looks like unintended behavior because if it has a timestamp but it's too old, we don't use it but we do warn

tonynajjar:
Obviously waiting on https://github.com/ros2/common_interfaces/pull/262/files but you can check the rest meanwhile

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
SteveMacenski: I'm saying that I think we should do the same comparison as the singular goal case; Check the global header only to make sure its a newer computed goal than the previous one. Don't look at the vector of PoseStamped stamps. I think we're on the same page :-)

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
SteveMacenski: If the timestamp isn't implemented for an incoming goal topic update, we skip it entirely. I think a log would be fantastic, yes!

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -49,32 +52,66 @@ GoalUpdater::GoalUpdater(
...
         "current goal (%f). Ignoring the received goal.",
         last_goal_received_time.seconds(), goal_time.seconds());
+      setOutput("output_goal", goal);
+    }
+  }
+
+  if (last_goals_received_.poses.empty()) {
+    setOutput("output_goals", goals);
+  } else if (last_goals_received_.header.stamp == rclcpp::Time(0)) {
+    setOutput("output_goals", last_goals_received_.poses);
```
SteveMacenski, line 94: So this would then return `goals` instead with an error log. Maybe also just reset the last goals received to an empty message so it only logs once

SteveMacenski: (assuming we agree that a `rclcpp::Time(0)` should be ignored)

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -49,32 +52,70 @@ GoalUpdater::GoalUpdater(
...
+  if (last_goals_received_.poses.empty()) {
+    setOutput("output_goals", goals);
+  } else if (last_goals_received_.header.stamp == rclcpp::Time(0)) {
+    RCLCPP_WARN(
+      node_->get_logger(), "The received goals array has no timestamp. Ignoring.");
+    setOutput("output_goals", goals);
+  } else {
+    auto last_goals_received_time = rclcpp::Time(last_goals_received_.header.stamp);
+    rclcpp::Time most_recent_goal_time = rclcpp::Time(0, 0, node_->get_clock()->get_clock_type());
+    for (const auto & g : goals) {
```
SteveMacenski, line 102: We use the `last_goals_received_.header.stamp` for the topic, why don't we use `goals.header.stamp` as the analog since these are the same message types with the same parent header? Doing it asymmetrically seems weird. 

Otherwise, this looks good to me. I'll follow up again in the `geometry_msgs` PR to see if we can get that in. If not, we can go with a nav2-msgs version for a temporary solution

SteveMacenski (changes requested):

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
SteveMacenski: > I would rather not do that in this PR.

I agree. We should do that after the actual `geometry_msgs` option is merged so we don't fliplash people twice.

SteveMacenski:
Usual bits on docs and updating the groot xml index for the new input/output ports

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
tonynajjar: So to be clear, this PR is blocked both by the `geometry_msgs` option AND by a PR that switches all other users of `std::vector<geometry_msgs::msg::PoseStamped>` to `PoseStampedArray`.

SteveMacenski: OK - we could merge this now with `nav2_msgs` version and update geometry_msgs once that's merged - and then do the rest of the switches to use it. But also happy to wait if that's what you prefer

tools/underlay.repos
```diff
@@ -35,3 +35,7 @@ repositories:
     type: git
     url:  https://github.com/ros-navigation/nav2_minimal_turtlebot_simulation.git
     version: 091b5ff12436890a54de6325df3573ae110e912b
+  ros/common_interfaces:
```
tonynajjar, line 38: just for CI testing

tonynajjar:
I merged https://github.com/ros-navigation/navigation2/pull/4791 into this branch to be able to test in CI. Let's come back to reviewing this PR what's the first one is merged for more clarity

SteveMacenski:
They are basically identical - we could close this one and just use that one!

SteveMacenski:
OK sounds good - that will make this diff much smaller and more concise to the change srequired

SteveMacenski:
I think this can now be rebased for only the changes required to the sp. BT nodes

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -49,39 +51,87 @@ GoalUpdater::GoalUpdater(
...
   // Spin multiple times due to rclcpp regression in Jazzy requiring a 'warm up' spin
   callback_group_executor_.spin_all(std::chrono::milliseconds(1));
   callback_group_executor_.spin_all(std::chrono::milliseconds(49));
 
-  if (last_goal_received_.header.stamp != rclcpp::Time(0)) {
-    auto last_goal_received_time = rclcpp::Time(last_goal_received_.header.stamp);
-    auto goal_time = rclcpp::Time(goal.header.stamp);
-    if (last_goal_received_time > goal_time) {
-      goal = last_goal_received_;
+  if (last_goal_received_set_) {
```
tonynajjar, line 73: I added these flags so that `last_goal_received_.header.stamp == rclcpp::Time(0)` is not valid  because it was not yet set

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -49,39 +51,87 @@ GoalUpdater::GoalUpdater(
...
+      RCLCPP_WARN(
+          node_->get_logger(), "The received goal has no timestamp. Ignoring.");
+      setOutput("output_goal", goal);
     } else {
+      auto last_goal_received_time = rclcpp::Time(last_goal_received_.header.stamp);
+      auto goal_time = rclcpp::Time(goal.header.stamp);
+      if (last_goal_received_time >= goal_time) {
+        setOutput("output_goal", last_goal_received_);
+      } else {
+        RCLCPP_WARN(
```
SteveMacenski, line 85: INFO not warning, I think

nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp
```diff
@@ -49,39 +51,87 @@ GoalUpdater::GoalUpdater(
...
-        last_goal_received_time.seconds(), goal_time.seconds());
+          node_->get_logger(), "The received goals array has no timestamp. Ignoring.");
+      setOutput("output_goals", goals);
+    } else {
+      auto last_goals_received_time = rclcpp::Time(last_goals_received_.header.stamp);
+      auto goals_time = rclcpp::Time(goals.header.stamp);
+      if (last_goals_received_time >= goals_time) {
+        setOutput("output_goals", last_goals_received_);
+      } else {
+        RCLCPP_WARN(
```
SteveMacenski, line 109: Ditto

SteveMacenski (changes requested):
Small change and then we can merge! Update the jazzy backport too and I'll do the pair

SteveMacenski (approved):

PR #4772, opened 2024-11-29
Fixes inability to find `yaml-cpp::yaml-cpp` when building nav2_waypoint_follower from source

Mechazo11:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   #4773  |
| Primary OS tested on | Ubuntu 22.04 (will also work in Ubuntu 24.04) |

---
Please see Issue #4773 for more details.

Not required

---

#### For Maintainers: 

### DIFF
--- a/nav2_waypoint_follower/CMakeLists.txt
+++ b/nav2_waypoint_follower/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.8)
 project(nav2_waypoint_follower)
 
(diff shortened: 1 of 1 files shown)


SteveMacenski:
@Mechazo11 I do not see any yaml-cpp usage in the WPF package. I do not believe this is necessary :frowning: Please see the thread in the previous ticket

Mechazo11:
@SteveMacenski  undetstood. Yest I was aware of the suggestion of removing yaml-cpp dependency exclusively in this package. I will try the fix in ```robot_localization``` package. Should the error repeat itself, I will leave a comment here.

PR #4775, opened 2024-12-02
Route Tool Rviz Panel

john-chrosniak:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | NA |
| Robotic platform tested on | Gazebo turlebot simulation and Ackermann steering robot |
| Does this PR contain AI generated software? | Yes and it is marked inline in the code |

---

* I added an Rviz panel to help create route graphs

* Documentation needed to highlight how to use the new Rviz panel

---

* Could add support for bidirectional edges or adding route operations using the tool in future work

#### For Maintainers: 


nav2_route/include/nav2_route/graph_saver.hpp
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2024 John Chrosniak
```
SteveMacenski, line 1: A big thing that we're missing in the graph saver and its associated plugin is unit testing for test coverage to make sure it works. Can you add in some tests to cover > 90% of the lines? 

Don't worry about the rviz plugin / panel, but the `nav2_route` saver and plugin should have testing done

john-chrosniak: I just finished adding in tests for the graph saver and geojson saver plugin. Coverage for the graph saver is at 90% and coverage for the plugin is at 96%. Let me know if there's anything else I can help with!

nav2_rviz_plugins/launch/route_tool.launch.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
SteveMacenski, line 1: I think something in the preamble is formatted wrong, CI is failing that the license is not recognized. Did you copy paste this over and only replace your name? 

Maybe the python3 line needs to be removed?

nav2_route/include/nav2_route/graph_saver.hpp
SteveMacenski: Can you write one more test to cover this case that is missing (just call the function with no `filepath` field when the param is set)?

```
  if (filepath.empty() && !graph_filepath_.empty()) {
    RCLCPP_DEBUG(
      logger_, "The graph filepath was not provided. "
      "Setting to %s", graph_filepath_.c_str());
    filepath = graph_filepath_;
  } 
```

Otherwise, LGTM, thanks for these!

SteveMacenski:
Question before I review on Friday: I noticed that you didn't modify the original server to have a service in the Route Server itself to save the graph to file. If we did that, we'd probably have to create services to also modify the graph by adding / removing a edge / node but then there's less logic that's happening in the rviz program itself that isn't portable to another GUI window. Can you say a little bit about why you did this all inline in the rviz tool? I assume you modify a bunch of stuff inline that is easier with direct access. I just want that in in mind when giving this a critical look  

Also: Does this dumping to file also do the `metadata` for the node semantics? Just checking.

nav2_rviz_plugins/launch/route_tool.launch.py
john-chrosniak: It looks like it may have been the use of single quotes instead of double quotes. I also removed the python3 line just in case.

john-chrosniak:
> Can you say a little bit about why you did this all inline in the rviz tool? 

My assumption was that route graphs are almost always going to be defined ahead of time, so there wasn't a need to support editing/saving graphs on the fly using the route server. The data structures used in the route tool class are only needed to keep track of nodes/edges as new elements are added/edited/removed, so if we're only modifying route graphs through this UI then this logic should be specific to the route tool.
 
> Also: Does this dumping to file also do the `metadata` for the node semantics? Just checking.

Yes both node and edge metadata is saved.

nav2_route/include/nav2_route/graph_saver.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include "nav2_route/types.hpp"
+#include "nav2_route/interfaces/graph_file_saver.hpp"
+#include "rclcpp/node_interfaces/node_parameters_interface.hpp"
+
+namespace nav2_route
+{
+
+/**
+ * @class nav2_route::GraphSaver
+ * @brief An action server to save graph objects to a file
```
SteveMacenski, line 39:
```suggestion
 * @brief An object to save graph objects to a file
```

nav2_route/src/plugins/graph_file_savers/geojson_graph_file_saver.cpp
```diff
@@ -0,0 +1,174 @@
...
+{
+  for (const auto & node : graph) {
+    Json json_feature, json_properties, json_geometry, json_metadata, json_operations;
+    json_geometry["type"] = "Point";
+    json_geometry["coordinates"] = std::vector<float>{node.coords.x, node.coords.y};
+    json_feature["geometry"] = json_geometry;
+    json_properties["id"] = node.nodeid;
+    json_properties["frame"] = node.coords.frame_id;
+    convertMetaDataToJson(node.metadata, json_metadata);
+    if (json_metadata.size()) {
```
SteveMacenski, line 72:
```suggestion
    if (json_metadata.size() > 0) {
```

Just to be more explicit

nav2_route/src/plugins/graph_file_savers/geojson_graph_file_saver.cpp
```diff
@@ -0,0 +1,174 @@
...
+      json_properties["metadata"] = json_metadata;
+    }
+    convertOperationsToJson(node.operations, json_operations);
+    if (json_operations.size()) {
```
SteveMacenski, line 76:
```suggestion
    if (json_operations.size() > 0) {
```

nav2_route/src/plugins/graph_file_savers/geojson_graph_file_saver.cpp
```diff
@@ -0,0 +1,174 @@
...
+  for (const auto & node : graph) {
+    for (const auto & edge : node.neighbors) {
+      Json json_edge, json_properties, json_geometry, json_metadata, json_operations;
+      json_geometry["type"] = "MultiLineString";
+      json_edge["geometry"] = json_geometry;
+      json_properties["id"] = edge.edgeid;
+      json_properties["startid"] = edge.start->nodeid;
+      json_properties["endid"] = edge.end->nodeid;
+      convertMetaDataToJson(edge.metadata, json_metadata);
+      if (json_metadata.size()) {
```
SteveMacenski, line 97:
```suggestion
      if (json_metadata.size() > 0) {
```

nav2_route/src/plugins/graph_file_savers/geojson_graph_file_saver.cpp
```diff
@@ -0,0 +1,174 @@
...
+        json_properties["metadata"] = json_metadata;
+      }
+      convertOperationsToJson(edge.operations, json_operations);
+      if (json_operations.size()) {
```
SteveMacenski, line 101:
```suggestion
      if (json_operations.size() > 0) {
```

nav2_rviz_plugins/src/route_tool.cpp
```diff
@@ -0,0 +1,300 @@
...
+    });
+}
+
+void RouteTool::on_load_button_clicked(void)
+{
+  graph_to_id_map_.clear();
+  edge_to_node_map_.clear();
+  graph_to_incoming_edges_map_.clear();
+  graph_.clear();
+  struct passwd pwd;
```
SteveMacenski, line 74: Its not clear to me what `pwd` `pw` `buf` or `homedir` is used for here. They don't appear to be used anywhere in the scope of this function

nav2_rviz_plugins/src/route_tool.cpp
```diff
@@ -0,0 +1,300 @@
...
+      }
+    }
+  }
+  next_node_id_ = max_node_id + 1;
+  update_route_graph();
+}
+
+void RouteTool::on_save_button_clicked(void)
+{
+  struct passwd pwd;
```
SteveMacenski, line 105: Ditto here

nav2_rviz_plugins/src/route_tool.cpp
```diff
@@ -0,0 +1,300 @@
...
+      auto start_node = &graph_[graph_to_id_map_[edge_to_node_map_[edge_id]]];
+      for (auto itr = start_node->neighbors.begin(); itr != start_node->neighbors.end(); itr++) {
+        if (itr->edgeid == edge_id) {
+          start_node->neighbors.erase(itr);
+          edge_to_node_map_.erase(edge_id);
+          break;
+        }
+      }
+    }
+    if (graph_[graph_to_id_map_[node_id]].nodeid == node_id) {
```
SteveMacenski, line 213: The bits in this `if` statement are a little difficult for me to parse. I think what you're trying to do is handle the fact that a node was deleted from the vector so the various map lookup structures need adjustment. Is that right? 

If so, may I recommend rather than deleting the node from the vector, we simply set the graph's value of that node to something identifiable (just unpopulated constructor or ID of `max`) and when we go to save, before sending the `graph_` to the file saver, we check and delete any nodes that match that identifiable pattern.

SteveMacenski (changes requested):

nav2_rviz_plugins/src/route_tool.cpp
john-chrosniak: This was intended to get the path to the user's home directory to open up when loading/saving graphs. That variable was unused so the tool was instead defaulting to the current working directory. This is probably the preferred option since it will usually be the user's ROS 2 workspace, so I'll go ahead and remove this logic.

nav2_rviz_plugins/src/route_tool.cpp
john-chrosniak: Same comment as above, removing this block.

nav2_rviz_plugins/src/route_tool.cpp
john-chrosniak: Yes I agree this is a cleaner approach, just added in these changes.

SteveMacenski (approved):

john-chrosniak:
Awesome!!! I'm happy to help out with any fixes/maintenance needed as beta testers are brought on!

SteveMacenski:
Great, I appreciate it!

PR #4777, opened 2024-12-03
Make RecoveryNode return Running

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

See https://github.com/ros-navigation/navigation2/issues/4776

We need to think how that might affect users

---

#### For Maintainers: 


SteveMacenski (commented):
No objections to the fix, but if we do this, is the `while` loop even important anymore? We could just do an `if (retry_count_ > number_of_retries_) {halt(); return FAILURE;}` and the rest could be without the additional scope, no? I think this is the only purpose that the `while` loop holds

tonynajjar:
> No objections to the fix, but if we do this, is the `while` loop even important anymore? We could just do an `if (retry_count_ > number_of_retries_) {halt(); return FAILURE;}` and the rest could be without the additional scope, no? I think this is the only purpose that the `while` loop holds

good point, fixed.

tonynajjar:
Looking at the PipelineSequence, it also has something similar [here](https://github.com/angsa-robotics/navigation2/blob/fix-recoverynode/nav2_behavior_tree/plugins/control/pipeline_sequence.cpp#L53). After SUCCESS, shouldn't we be saving a state and return RUNNING?

SteveMacenski (approved):

SteveMacenski:
These must have been written by the same person. I'd merge a PR to do the same thing as here for that as well :-)

tonynajjar:
Follow up ticket: https://github.com/ros-navigation/navigation2/issues/4782

PR #4781, opened 2024-12-04
nav2_costmap_2d plugin container layer

alexanderjyuen:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | gazebo simulation of turtlebot) |

---

* Added a new costmap layer plugin to nav2_costmap_2d called plugin container layer which allows users to specify different submaps within a costmap. For example if one would like to apply a certain set of inflation layer settings to the static layer, and a different set of inflation layer settings to an obstacle layer, in the same overall costmap, one could group those different layers under different plugin container layers
    
* Modified nav2_costmap_2d integrated testing to also run tests for the plugin container layer
    
* included nav2_params_plugin_container_layer.yaml to run turtlebot example
    

* [Configuring Costmap](https://docs.nav2.org/configuration/packages/configuring-costmaps.html) needs to include plugin container layer in the Plugin Parameters section
* A new page has been added to outline how to configure the plugin container layer
* [Navigation Plugins](https://docs.nav2.org/plugins/index.html) need to include plugin container layer in Costmap Layer section
* [Tuning Guide](https://docs.nav2.org/tuning/index.html#costmap2d-plugins) needs to include plugin container layer in the Costmap2D Plugins section

---

* This could be possible extended to be used with costmap filters, although filters don't necessarily rely on each other

#### For Maintainers: 
- [x] Check that any new parameters added are updated in docs.nav2.org


nav2_bringup/params/nav2_params_plugin_container_layer.yaml
```diff
@@ -0,0 +1,456 @@
...
+    ros__parameters:
+      update_frequency: 1.0
+      publish_frequency: 1.0
+      global_frame: map
+      robot_base_frame: base_link
+      robot_radius: 0.22
+      resolution: 0.05
+      track_unknown_space: true
+
+      plugins: ["plugin_container_static_layer", "plugin_container_obstacle_layer"]
```
SteveMacenski, line 239: Remove this file, but put the examples you give here in the Configuration File at the bottom as an example YAML. You can see examples of that at the bottom of the Smac Planner / MPPI controller pages for rst semantics

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * @brief Reset this costmap
+   */
+  virtual void reset();
+  /**
+   * @brief If clearing operations should be processed on this layer or not
+   */
+  virtual bool isClearable()
+  {
+    return true;
```
SteveMacenski, line 111: We don't necessarily know this - we should check the layers and return true if any are clearable

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
```diff
@@ -0,0 +1,137 @@
...
+    std::vector<rclcpp::Parameter> parameters);
+
+private:
+  /// @brief Dynamic parameters handler
+  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr
+    dyn_params_handler_;
+  bool rolling_window_;
+  int combination_method_;
+  std::vector<std::shared_ptr<Layer>> plugins_;
+  std::vector<std::string> default_plugins_;
```
SteveMacenski, line 129: Remove all defaults; they must be fully specified

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
```diff
@@ -0,0 +1,137 @@
...
+  std::vector<std::string> default_types_;
+  std::vector<std::string> plugin_names_;
+  std::vector<std::string> plugin_types_;
+
+  Costmap2D combined_costmap_;
```
SteveMacenski, line 134: A CostmapLayer derived class already contains a costmap as a base class https://github.com/ros-navigation/navigation2/blob/a0c11b441b5faa1924ba31953900e14cc0a055f1/nav2_costmap_2d/include/nav2_costmap_2d/costmap_layer.hpp#L56

https://github.com/ros-navigation/navigation2/blob/a0c11b441b5faa1924ba31953900e14cc0a055f1/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp

SteveMacenski, line 134: A CostmapLayer derived class already contains a costmap as a base class https://github.com/ros-navigation/navigation2/blob/a0c11b441b5faa1924ba31953900e14cc0a055f1/nav2_costmap_2d/include/nav2_costmap_2d/costmap_layer.hpp#L56

https://github.com/ros-navigation/navigation2/blob/a0c11b441b5faa1924ba31953900e14cc0a055f1/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
```diff
@@ -0,0 +1,137 @@
...
+   * @param event ParameterEvent message
+   */
+  rcl_interfaces::msg::SetParametersResult dynamicParametersCallback(
+    std::vector<rclcpp::Parameter> parameters);
+
+private:
+  /// @brief Dynamic parameters handler
+  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr
+    dyn_params_handler_;
+  bool rolling_window_;
```
SteveMacenski, line 126: Appears unused?

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin(); plugin != plugins_.end();
+    ++plugin)
+  {
+    (*plugin)->updateCosts(combined_costmap_, min_i, min_j, max_i, max_j);
+  }
+
+  costmap_ = combined_costmap_.getCharMap();
+
+  switch (combination_method_) {
```
SteveMacenski, line 125: I think you're missing some https://github.com/ros-navigation/navigation2/blob/a0c11b441b5faa1924ba31953900e14cc0a055f1/nav2_costmap_2d/plugins/obstacle_layer.cpp#L552-L564

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,218 @@
...
+namespace nav2_costmap_2d
+{
+PluginContainerLayer::PluginContainerLayer()
+: default_plugins_{},
+  default_types_{
+    "nav2_costmap_2d::StaticLayer",
+    "nav2_costmap_2d::ObstacleLayer",
+    "nav2_costmap_2d::InflationLayer"}
+{
+  costmap_ = NULL;
```
SteveMacenski, line 36: `nullptr` is a more modern way of doing this -- but also these all can be removed since we don't need the separate grid :wink:

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+PluginContainerLayer::~PluginContainerLayer()
+{
+  costmap_ = NULL;
```
SteveMacenski, line 41: See above, so its just going to be removed, but for an "FYI" this is generally unsafe. You want to check `if(costmap_)` so you can `delete` its memory before setting to a `nullptr`

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,218 @@
...
+  plugin_types_.resize(plugin_names_.size());
+
+  for (size_t i = 0; i < plugin_names_.size(); ++i) {
+    plugin_types_[i] = nav2_util::get_plugin_type_param(node, name_ + "." + plugin_names_[i]);
+  }
+
+  // Then load and add the plug-ins to the costmap
+  for (unsigned int i = 0; i < plugin_names_.size(); ++i) {
+    std::shared_ptr<Layer> plugin = plugin_loader_.createSharedInstance(plugin_types_[i]);
+    std::unique_lock<Costmap2D::mutex_t> lock(*(layered_costmap_->getCostmap()->getMutex()));
```
SteveMacenski, line 71: What's with the lock?

SteveMacenski (changes requested):
I didn't call out everything, since I think much of other comments are moot if the `costmap_` disappears so its better to wait for the revised version :-) 

Check out the codecov annotations in the Code UX of the PR to see some gaps in your testing. Luckily, you have examples of alot of other Nav2 code of how to fill them (most are pretty easy pickings)

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: fixed

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: fixed

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
alexanderjyuen: fixed

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
alexanderjyuen: fixed

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
alexanderjyuen: fixed

nav2_bringup/params/nav2_params_plugin_container_layer.yaml
alexanderjyuen: fixed, added to docs

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
alexanderjyuen: fixed in latest commits

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: fixed in latest commits

alexanderjyuen:
> I didn't call out everything, since I think much of other comments are moot if the `costmap_` disappears so its better to wait for the revised version :-)
> 
> Check out the codecov annotations in the Code UX of the PR to see some gaps in your testing. Luckily, you have examples of alot of other Nav2 code of how to fill them (most are pretty easy pickings)

Latest version should have code coverage up to 99%'

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+  }
+}
+
+bool PluginContainerLayer::isClearable()
+{
+  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin(); plugin != plugins_.end();
+    ++plugin)
+  {
+    if(!(*plugin)->isClearable()) {
+      return false;
```
SteveMacenski, line 203: I think this logic needs o be inverted. If any plugin is clearable, then return true. Else, false. 

Right now, if any one thing isn't clearable (i.e. inflation layer) then we don't clear any of it

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+    plugin_types_[i] = nav2_util::get_plugin_type_param(node, name_ + "." + plugin_names_[i]);
+  }
+
+  // Then load and add the plug-ins to the costmap
+  for (unsigned int i = 0; i < plugin_names_.size(); ++i) {
+    std::shared_ptr<Layer> plugin = plugin_loader_.createSharedInstance(plugin_types_[i]);
+    addPlugin(plugin, plugin_names_[i]);
+  }
+
+  default_value_ = nav2_costmap_2d::FREE_SPACE;
```
SteveMacenski, line 78: Unused?

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+  int min_j,
+  int max_i,
+  int max_j)
+{
+  std::lock_guard<Costmap2D::mutex_t> guard(*getMutex());
+  if (!enabled_) {
+    return;
+  }
+
+  resetMap(min_i, min_j, max_i, max_j);
```
SteveMacenski, line 119: Why reset map?

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+}
+
+void PluginContainerLayer::reset()
+{
+  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin(); plugin != plugins_.end();
+    ++plugin)
+  {
+    (*plugin)->reset();
+  }
+
```
SteveMacenski, line 169: Does the internal layer need to have some update for reset? I imagine so. Any other of these functions?

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: Got it, just to clarify, in a situation where one of the layers is clearable and the other ones are not, we'll just go ahead and return that this plugin is clearable.

Are there any safety implications in that scenario? Such as clearing sub layer that is not supposed to be cleared

SteveMacenski: We use this here: https://github.com/ros-navigation/navigation2/blob/add09f60b17f6134416122f8285eba7268d275d7/nav2_costmap_2d/src/clear_costmap_service.cpp#L116-L119

When someone asks for a costmap clear, this is what tells us if we can / should do it. Your implementation would only clear this container's cost-grid, but wouldn't clear its constituent plugin costmap representations 

What we should do is implement the `clearArea` method (ex. https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/a46e4bb47d175872a2681360901a47cd0b014f7c/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp#L954-L968). Within that, clear the costmap datastructure of the layer & also clear out the region from any `isClearable` sub-layer

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: good catch, added resetMap() in here

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: was unnecessary, removed this

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: required for initial matchSize() in initialization, was undefined and populating map with non zero values after initial matchSize()

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: swapped the logic

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
```diff
@@ -0,0 +1,133 @@
...
+  /// @brief Dynamic parameters handler
+  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr
+    dyn_params_handler_;
+
+  nav2_costmap_2d::CombinationMethod combination_method_;
+  std::vector<std::shared_ptr<Layer>> plugins_;
+  std::vector<std::string> plugin_names_;
+  std::vector<std::string> plugin_types_;
+
+  Costmap2D combined_costmap_;
```
SteveMacenski, line 130: This is unused

nav2_costmap_2d/plugins/plugin_container_layer.cpp
SteveMacenski: Should you use the master costmap's value for this? Or, I figure actually the safer thing to do is set it as unknown space, so that the costmap internal layers themselves can set it to FREE if they choose. The free should override unknown, but the "FREE" here could override otherwise what is intended to be tracked as UNKNOWN -- right?

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,233 @@
...
+  plugin_types_.resize(plugin_names_.size());
+
+  for (size_t i = 0; i < plugin_names_.size(); ++i) {
+    plugin_types_[i] = nav2_util::get_plugin_type_param(node, name_ + "." + plugin_names_[i]);
+  }
+
+  // Then load and add the plug-ins to the costmap
+  for (unsigned int i = 0; i < plugin_names_.size(); ++i) {
+    std::shared_ptr<Layer> plugin = plugin_loader_.createSharedInstance(plugin_types_[i]);
```
SteveMacenski, line 74: Why not add the below before this line and remove the previous for loop?

>     plugin_types_[i] = nav2_util::get_plugin_type_param(node, name_ + "." + plugin_names_[i]);

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,233 @@
...
+  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin(); plugin != plugins_.end();
+    ++plugin)
+  {
+    if((*plugin)->isClearable()) {
+      return true;
+    }
+  }
+  return false;
+}
+
```
SteveMacenski, line 206: I think you still need the `clearArea` override so that you can clear not only this layer's submap, but the constituent sub-plugin's costmaps.

Right now, if you `clear_costmaps`, the Container Layer's internal costmap will be reset, but the plugin's wont. So next update cycle you'll get stale data re-populating it

SteveMacenski (changes requested):

SteveMacenski:
Also, your tests are failing :upside_down_face:

nav2_costmap_2d/include/nav2_costmap_2d/plugin_container_layer.hpp
alexanderjyuen: fixed

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: fixed

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: That makes considerably more sense, fixed in latest commit

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,242 @@
...
+void PluginContainerLayer::clearArea(int start_x, int start_y, int end_x, int end_y, bool invert)
+{
+  // TODO(alexander_yuen): Tests for this method
+  CostmapLayer::clearArea(start_x, start_y, end_x, end_y, invert);
+  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin(); plugin != plugins_.end();
+    ++plugin)
+  {
+    auto costmap_layer = std::dynamic_pointer_cast<nav2_costmap_2d::CostmapLayer>(*plugin);
```
alexanderjyuen, line 209: @SteveMacenski, is this the right way to do this?

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+using std::vector;
+
+namespace nav2_costmap_2d
+{
+PluginContainerLayer::PluginContainerLayer()
+{
+}
+
+PluginContainerLayer::~PluginContainerLayer()
```
SteveMacenski, line 33: You can use the default constructor / destructor if you have nothing to do here

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  dyn_params_handler_ = node->add_on_set_parameters_callback(
+    std::bind(
+      &PluginContainerLayer::dynamicParametersCallback,
+      this,
+      std::placeholders::_1));
+
+  plugin_types_.resize(plugin_names_.size());
+
+  for (unsigned int i = 0; i < plugin_names_.size(); ++i) {
```
SteveMacenski, line 67: We need to add a try/catch around https://github.com/ros-navigation/navigation2/blob/684a4d037241674550b756c5014711fa6c13e35e/nav2_costmap_2d/src/costmap_2d_ros.cpp#L180-L182 now if we do this, since this can throw

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,242 @@
...
+    if((*plugin)->isClearable()) {
+      return true;
+    }
+  }
+  return false;
+}
+
+void PluginContainerLayer::clearArea(int start_x, int start_y, int end_x, int end_y, bool invert)
+{
+  // TODO(alexander_yuen): Tests for this method
```
SteveMacenski, line 204: Remove TODO (and test :wink: )

nav2_costmap_2d/plugins/plugin_container_layer.cpp
```diff
@@ -0,0 +1,242 @@
...
+  CostmapLayer::clearArea(start_x, start_y, end_x, end_y, invert);
+  for (vector<std::shared_ptr<Layer>>::iterator plugin = plugins_.begin(); plugin != plugins_.end();
+    ++plugin)
+  {
+    auto costmap_layer = std::dynamic_pointer_cast<nav2_costmap_2d::CostmapLayer>(*plugin);
+    if (costmap_layer != nullptr) {
```
SteveMacenski, line 210:
```suggestion
    if (plugin->isClearable() && costmap_layer != nullptr) {
```

SteveMacenski (changes requested):
Then pretty much done!

On the docs, I still need the updated image for multiple inflation settings (or another example) to show how it can be used to create differing behavior -- but otherwise that is good too!

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: fixed

nav2_costmap_2d/plugins/plugin_container_layer.cpp
alexanderjyuen: done

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -177,9 +177,16 @@ Costmap2DROS::on_configure(const rclcpp_lifecycle::State & /*state*/)
     layered_costmap_->addPlugin(plugin);
 
     // TODO(mjeronimo): instead of get(), use a shared ptr
-    plugin->initialize(
+    try {
+      plugin->initialize(
       layered_costmap_.get(), plugin_names_[i], tf_buffer_.get(),
```
SteveMacenski, line 182: I'm kind of surprised this isn't caught as a linting error, there is no intent on the arguments

SteveMacenski (approved):

nav2_costmap_2d/src/costmap_2d_ros.cpp
alexanderjyuen: Fixed this manually

SteveMacenski (approved):

SteveMacenski:
Thanks! All good and merged the pair!

Issue #4783, opened 2024-12-04
[feature request]: Add parameter to the DriveOnHeading and BackUp actions to disable collision checking


david-wb:
#### Feature description

The title is self-explanatory, but the purpose of this feature is that, in some cases we might want to advance the robot even in the presence of obstacles, e.g. to intentionally push an obstacle. 

Related to [this](https://robotics.stackexchange.com/questions/113902/how-to-choose-controller-when-using-nav2s-driveonheading-commander-api) Robotics StackExchange question.

#### Implementation considerations

I think the feature can be implemented quite simply by adding a new boolean param to the action messages and using the param inside of the DriveOnHeading implementation to short-circuit collision checking when the param is true. 

#### Ownership
I intend to take ownership of this ticket and produce a PR within 1-2 days.

SteveMacenski:
Thanks! I look forward to it :-)

PR #4785, opened 2024-12-05
new param to disable collision checking in DriveOnHeading and BackUp actions

david-wb:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4783 |
| Robotic platform tested on | Gazebo sim of custom robot |

---

Adds a new bool param called `no_collision_checks` to the `DriveOnHeading` and `BackUp` actions. When the param is set to `true`, collision checking is disabled on these actions. This is useful, for example, when you want to push through known obstacles.

Updated the README for the commander API to include the flag in the listed example.

The new param needs to be added to the docs page.
---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/plugins/action/back_up_action.cpp
+++ b/nav2_behavior_tree/plugins/action/back_up_action.cpp
@@ -36,6 +36,8 @@ void nav2_behavior_tree::BackUpAction::initialize()
   getInput("backup_speed", speed);
   double time_allowance;
   getInput("time_allowance", time_allowance);
+  double disable_collision_checking;
+  getInput("disable_collision_checking", disable_collision_checking);
 
   // Populate the input message
   goal_.target.x = dist;
@@ -43,6 +45,7 @@ void nav2_behavior_tree::BackUpAction::initialize()
   goal_.target.z = 0.0;
   goal_.speed = speed;
   goal_.time_allowance = rclcpp::Duration::from_seconds(time_allowance);
+  goal_.disable_collision_checking = disable_collision_checking;
 }
 
 void BackUpAction::on_tick()
--- a/nav2_behavior_tree/plugins/action/back_up_action.cpp
+++ b/nav2_behavior_tree/plugins/action/back_up_action.cpp
@@ -36,8 +36,8 @@ void nav2_behavior_tree::BackUpAction::initialize()
   getInput("backup_speed", speed);
   double time_allowance;
   getInput("time_allowance", time_allowance);
-  double disable_collision_checking;
-  getInput("disable_collision_checking", disable_collision_checking);
+  double no_collision_checks;
+  getInput("no_collision_checks", no_collision_checks);
 
   // Populate the input message
   goal_.target.x = dist;
@@ -45,7 +45,7 @@ void nav2_behavior_tree::BackUpAction::initialize()
   goal_.target.z = 0.0;
   goal_.speed = speed;
   goal_.time_allowance = rclcpp::Duration::from_seconds(time_allowance);
-  goal_.disable_collision_checking = disable_collision_checking;
+  goal_.no_collision_checks = no_collision_checks;
 }
 
 void BackUpAction::on_tick()
--- a/nav2_behavior_tree/plugins/action/back_up_action.cpp
+++ b/nav2_behavior_tree/plugins/action/back_up_action.cpp
@@ -36,7 +36,7 @@ void nav2_behavior_tree::BackUpAction::initialize()
   getInput("backup_speed", speed);
   double time_allowance;
   getInput("time_allowance", time_allowance);
-  double disable_collision_checks;
+  bool disable_collision_checks;
   getInput("disable_collision_checks", disable_collision_checks);
 
   // Populate the input message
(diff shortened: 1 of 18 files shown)


nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -168,6 +169,10 @@ class DriveOnHeading : public TimedBehavior<ActionT>
     const geometry_msgs::msg::Twist & cmd_vel,
     geometry_msgs::msg::Pose2D & pose2d)
   {
+    if (command_no_collision_checks_) {
```
SteveMacenski, line 172: You should use the action goal's value here, not create a new class member variable :-)

nav2_msgs/action/BackUp.action
```diff
@@ -3,6 +3,7 @@
 geometry_msgs/Point target
 float32 speed
 builtin_interfaces/Duration time_allowance
+bool no_collision_checks
```
SteveMacenski, line 6: For both: set default to `False`

Can we also rename it here + elsewhere as `disable_collision_checks`?

SteveMacenski (changes requested):
Otherwise generally looks good to me! The BT node config guide pages need updating with the new port, migration guide to highlight this change

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
david-wb: I tried to follow same pattern used for `command_x_` and `command_speed_`. Otherwise, I will have to pass the action goal to this method. Which do you prefer? Thanks.

SteveMacenski: But you didn't populate it, it seems?

david-wb:
> Otherwise generally looks good to me! The BT node config guide pages need updating with the new port, migration guide to highlight this change

I'm not yet up to speed on the release process. Would this change be published as an update to the nav2 Jazzy package (`ros-jazzy-navigation2` or will it not be available until the next ros2 release? I think I need to update the [jazzy.rst](https://github.com/ros-navigation/docs.nav2.org/blob/master/migration/Jazzy.rst) migration guide, is that right?

david-wb:
Here is the corresponding docs PR: https://github.com/ros-navigation/docs.nav2.org/pull/617

nav2_behavior_tree/plugins/action/drive_on_heading_action.cpp
```diff
@@ -37,13 +37,16 @@ void DriveOnHeadingAction::initialize()
   getInput("speed", speed);
   double time_allowance;
   getInput("time_allowance", time_allowance);
+  double disable_collision_checks;
```
SteveMacenski, line 40: This should be a bool

nav2_behavior_tree/plugins/action/back_up_action.cpp
```diff
@@ -36,13 +36,16 @@ void nav2_behavior_tree::BackUpAction::initialize()
   getInput("backup_speed", speed);
   double time_allowance;
   getInput("time_allowance", time_allowance);
+  double disable_collision_checks;
```
SteveMacenski, line 39: Bool

nav2_simple_commander/README.md
```diff
@@ -23,8 +23,8 @@ New as of September 2023: the simple navigator constructor will accept a `namesp
 | goToPose(pose, behavior_tree='')  | Requests the robot to drive to a pose (`PoseStamped`).                     |
 | followWaypoints(poses)            | Requests the robot to follow a set of waypoints (list of `PoseStamped`). This will execute the specific `TaskExecutor` at each pose.   |
 | followPath(path, controller_id='', goal_checker_id='') | Requests the robot to follow a path from a starting to a goal `PoseStamped`, `nav_msgs/Path`.     |
-| spin(spin_dist=1.57, time_allowance=10)   | Requests the robot to performs an in-place rotation by a given angle.      |
-| backup(backup_dist=0.15, backup_speed=0.025, time_allowance=10) | Requests the robot to back up by a given distance.         |
+| spin(spin_dist=1.57, time_allowance=10, disable_collision_checks=False)   | Requests the robot to performs an in-place rotation by a given angle.      |
```
SteveMacenski, line 26: Please also update here: https://docs.nav2.org/commander_api/index.html

SteveMacenski (changes requested):
Small details - but otherwise LGTM!

nav2_behavior_tree/plugins/action/drive_on_heading_action.cpp
david-wb: Good catch. I'm Surprised the compiler didn't complain.

SteveMacenski (approved):
Waiting on CI + docs PR to be updated to a non-draft state

david-wb:
Looks like a couple `opennav_docking` tests are failing.

SteveMacenski:
I retriggered. I think its just a flaky test. I dont see why this would cause that. Please update the docs PR - I suspect it’ll pass this time 🙂

PR #4789, opened 2024-12-10
Goalchecker path argument and plugin

josephduchesne:
This is a continuation of work started by @ottojo in draft PR https://github.com/ros-navigation/navigation2/pull/4593 

I'll quote their introduction and, expand a bit afterwards, and update any fields that have changed.

> The controller server uses a configurable goal checker to determine if the robot has completed its current path as defined by the global plan. Current goal checkers compare the goal (end of path) pose and twist to the current robot state, but that is not sufficient in all cases.
> 
> In some use cases, the path will visit the goal multiple times, and the goal might coincide with the starting position. It is still desired that the robot follows the entire path, instead of immediately ending navigation once the goal pose is reached.
> 
> This PR adds a parameter to the GoalChecker interface to inform the goal checker of the current path, which enables building more sophisticated goal checkers that (for example) take progress along the path into account.
> 
> We already use such a goal checker internally, which just subscribes to the global plan via the appropriate ROS topic, but i think this here is the cleaner solution (and also avoids race conditions of checking against an old path in the goal checker...)
> 
>  I don't have a nice testing setup and a goal checker using this interface yet (other than our own robot and the mentioned goal checker), which is why I mark this as draft for now.

I've cherry-picked @ottojo 's commit on top of  the latest main branch, then added a new basic goal checker PathCompleteGoalChecker, which is a subclass of SimpleGoalChecker, but has an additional parameter (path_length_tolerance, default=1), and checks that the current path has <= path_length_tolerance waypoints before allowing the normal SimpleGoalChecker completion logic to take its course.

The result is a trivial extension of SimpleGoalChecker that should be immune to the current_pose = goal_pose short-circuit problem/bug. Rather than try to create an elaborate unit test that navigates a course, I stuck to first principles to ensure that the new plugin behaves the same as SimpleGoalChecker with <= path_length_tolerance waypoints, and that it does not complete with > path_length_tolerance waypoints.

## Basic Info
| Info |	Please fill out this column |
| -------| -------------------------------------- |
| Ticket(s) this addresses |	#4304 |
| Robotic platform tested on |	(WIP)|

* [x]  Extend `GoalChecker::isGoalReached` interface with argument for current path
* [x]  Add this argument to the in-tree goal checker implementations
* [x]  Add a goal checker which checks for path-completion
* []  Define a test-scenario with global plan that returns to start pose

* [ ]  Migration guide for out-of-tree goal checkers

* Implementing more sophisticated goal checkers

#### For Maintainers:
* [ ]  Check that any new parameters added are updated in docs.nav2.org
* [ ]  Check that any significant change is added to the migration guide
* [ ]  Check that any new features **OR** changes to existing behaviors are reflected in the tuning guide
* [ ]  Check that any new functions have Doxygen added
* [ ]  Check that any new features have test coverage
* [ ]  Check that any new plugins is added to the plugins page
* [ ]  If BT Node, Additionally: add to BT's XML index of nodes for groot, BT package's readme table, and BT library lists


nav2_controller/CMakeLists.txt
```diff
@@ -161,7 +183,7 @@ endif()
 
 rclcpp_components_register_nodes(${library_name} "nav2_controller::ControllerServer")
 
-install(TARGETS simple_progress_checker pose_progress_checker simple_goal_checker stopped_goal_checker ${library_name}
+install(TARGETS simple_progress_checker pose_progress_checker simple_goal_checker stopped_goal_checker path_complete_goal_checker ${library_name}
```
SteveMacenski, line 186: See CI: linting is complaining about this line

nav2_controller/include/nav2_controller/plugins/path_complete_goal_checker.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "nav2_controller/plugins/simple_goal_checker.hpp"
+
+namespace nav2_controller
+{
+
+/**
+ * @class PathCompleteGoalChecker
+ * @brief Goal Checker plugin that checks position delta, once path is shorter than a threshold.
+ *
+ * A
```
SteveMacenski, line 51: ?

nav2_controller/include/nav2_controller/plugins/path_complete_goal_checker.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+  void reset() override;
+
+  bool isGoalReached(
+    const geometry_msgs::msg::Pose & query_pose, const geometry_msgs::msg::Pose & goal_pose,
+    const geometry_msgs::msg::Twist & velocity, const nav_msgs::msg::Path & current_path) override;
+
+protected:
+  // threshold for path goal
+  int path_length_tolerance_;
```
SteveMacenski, line 72: double

nav2_controller/plugins/path_complete_goal_checker.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "nav2_util/node_utils.hpp"
+
+using rcl_interfaces::msg::ParameterType;
+using std::placeholders::_1;
+
+namespace nav2_controller
+{
+
+PathCompleteGoalChecker::PathCompleteGoalChecker()
+: SimpleGoalChecker(), path_length_tolerance_(1)
```
SteveMacenski, line 46: We should set `path_length_tolerance_` as an actual distance (i.e. in meters) and thus should be a double. I'm thinking default of 1m? `1.0`

nav2_controller/plugins/path_complete_goal_checker.cpp
```diff
@@ -0,0 +1,111 @@
...
+  SimpleGoalChecker::reset();
+}
+
+bool PathCompleteGoalChecker::isGoalReached(
+  const geometry_msgs::msg::Pose & query_pose, const geometry_msgs::msg::Pose & goal_pose,
+  const geometry_msgs::msg::Twist & twist, const nav_msgs::msg::Path & path)
+{
+  // return false if more than path_length_tolerance_ waypoints exist
+  // note: another useful version of this could check path length
+  if (path.poses.size() > (unsigned int)path_length_tolerance_) {
```
SteveMacenski, line 83: Check the distance from the query pose to the goal pose (and double for length)

SteveMacenski: One of these should be your friend https://github.com/ros-navigation/navigation2/blob/main/nav2_util/include/nav2_util/geometry_utils.hpp#L51-L120

SteveMacenski (changes requested):
A couple of followups:
- We should add in a new configuration guide page for this on docs.nav2.org. Also, a migration guide entry for the new feature!
- Is there a sensible reason why we shouldn't make this the default? I think we can update the nav2_params.yaml to always use this now & update the controller server's built-in default as well. This, I believe, is a globally better solution :-) 

Generally though, this looks good to me

nav2_controller/include/nav2_controller/plugins/path_complete_goal_checker.hpp
josephduchesne: It's currently an int, since it checks the number of poses in the path, not the... length. Ok, I see how this is confusing :)

I'll update the PR so that it works the way you expect, using the nav2_utils path length checker, assuming that's not too expensive for the goal checker. I'll have to check the code, since I am a bit worried about someone getting poor performance for their 11km long, 15,000 pose path. On second thought, they'll get poor performance regardless of the goal checker, so it should be ok.

josephduchesne:
> A couple of followups:
> 
> * We should add in a new configuration guide page for this on docs.nav2.org. Also, a migration guide entry for the new feature!

Will do.

> * Is there a sensible reason why we shouldn't make this the default? I think we can update the nav2_params.yaml to always use this now & update the controller server's built-in default as well. This, I believe, is a globally better solution :-)

I think this is a reasonable default. The only advantage that the existing built-in goal checkers have is that they require potentially far less math to check (which could be important on budget educational bots). I only made a new goal checker because that was what the original draft author had in his todo list :) 

I'll update the default in the PR.

josephduchesne:
I keep thinking that there is probably a way to use `nav2_util::geometry_utils::first_after_integrated_distance` rather than add a second short-circuiting checker to nav2_util::geometry_utils, but I don't think first_after_integrated_distance can be used to check if the remaining path exceeds a specific length without adding several other checks and calculations. It's so close!

nav2_controller/include/nav2_controller/plugins/path_complete_goal_checker.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "nav2_controller/plugins/simple_goal_checker.hpp"
+
+namespace nav2_controller
+{
+
+/**
+ * @class PathCompleteGoalChecker
+ * @brief Goal Checker plugin that checks position delta, once path is shorter than a threshold.
+ *
+ *
```
SteveMacenski, line 51: Remove the extra 2 lines :-)

nav2_controller/src/controller_server.cpp
```diff
@@ -41,7 +41,7 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
   default_progress_checker_types_{"nav2_controller::SimpleProgressChecker"},
   goal_checker_loader_("nav2_core", "nav2_core::GoalChecker"),
   default_goal_checker_ids_{"goal_checker"},
-  default_goal_checker_types_{"nav2_controller::SimpleGoalChecker"},
+  default_goal_checker_types_{"nav2_controller::PathCompleteGoalChecker"},
```
SteveMacenski, line 44: In the docs PR, we should update this default

SteveMacenski (changes requested):

nav2_controller/plugins/path_complete_goal_checker.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    plugin_name + ".path_length_tolerance",
+    rclcpp::ParameterValue(path_length_tolerance_));
+
+  node->get_parameter(plugin_name + ".path_length_tolerance", path_length_tolerance_);
+
+  // Replace SimpleGoalChecker's callback for dynamic parameters
+  node->remove_on_set_parameters_callback(dyn_params_handler_.get());
```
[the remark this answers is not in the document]
josephduchesne: Good point, that's simpler. Will do.

nav2_controller/plugins/path_complete_goal_checker.cpp
```diff
@@ -0,0 +1,112 @@
...
+  SimpleGoalChecker::reset();
+}
+
+bool PathCompleteGoalChecker::isGoalReached(
+  const geometry_msgs::msg::Pose & query_pose, const geometry_msgs::msg::Pose & goal_pose,
+  const geometry_msgs::msg::Twist & twist, const nav_msgs::msg::Path & path)
+{
+  using nav2_util::geometry_utils::is_path_longer_than_length;
+
+  if (is_path_longer_than_length(path, path_length_tolerance_)) {
```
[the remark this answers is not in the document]
josephduchesne: I'll see what I can do :)

SteveMacenski:
@josephduchesne can you button up these last few details so we can get this in? :smile:

josephduchesne:
I started to look into the path handler implementations to try to figure out a good approach for handling the path shortening aspect generically as you suggested. I'm not sure if I'll have time in the next few weeks to try to address that.

It hadn't occurred to me that this might be a problem since while I have been working with many different controllers, I'm doing so in the context of a Nav2 BT similar to the "NavigateWithReplanning" example, containing this bit that wraps Bathe controller/planner/path interaction.

```
<RateController hz="1.0">
  <ReactiveSequence>
    <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.45"/>
    <Sequence name="ComputeAndSmoothPath">
      <ComputePathThroughPoses goals="{goals}" path="{path}" planner_id="{selected_planner}" error_code_id="{compute_path_error_code}"/>
      <SmoothPath unsmoothed_path="{path}" smoothed_path="{path}" error_code_id="{smoother_error_code}"/>
    </Sequence>
  </ReactiveSequence>
</RateController>
<FollowPath path="{path}" controller_id="{selected_controller}" error_code_id="{follow_path_error_code}"/>
```

This means that aside from some slight quirks if I use very loosely defined paths + a very opinionated planner, the controller will suddenly cut corners or similar once a path goal is removed that was previously influencing intermediate steps in the planner or controller. That's not a huge problem, but interesting as an aside.

Back to the design at hand.

I think there's a tension around the "current_path_" use as I'm using it, how its used in controler_server.cpp computeAndPublishVelocity(), which also contains the "copy paste" you'd recommend removing (generating current_path, and perhaps ironically possibly having the exact same bug that I'm trying to fix here if the path contains loops, close zig-zags or self-intersections. 

The issue with implementing that in the goal checker is that the goal checker's job currently doesn't include maintaining a stateful global plan, and updating it as the robot moves. That, I think, is the Controller's job, and the current_path_ passed in to the goal checker should be one that includes only present (maybe) and future path segments, not past ones.

We certainly could increase the scope of the goal checkers to be much more stateful to handle this. I'd rather move to trying to enforce a (soft) expectation that the controllers prune their paths.

Either way, I'm curious to hear your thoughts since I'm jumping back into reading this code after a month away from it, and far less familiar with Nav2's implementation on the best of days.

My hope would be either a clear decision that this PR requires a resolution to the global path is never updated issue, or agreement that we can split this off into its own enhancement request, and follow-up branch + PR down the road. I'm still interested in working on that either way, but I'm working on an agricultural robot that's got to be ready for the planting season, so my time is more limited the next few months.

SteveMacenski:
> I think there's a tension around the "current_path_" use as I'm using it, how its used in controler_server.cpp computeAndPublishVelocity(), which also contains the "copy paste" you'd recommend removing (generating current_path, and perhaps ironically possibly having the exact same bug that I'm trying to fix here if the path contains loops, close zig-zags or self-intersections.

See [MPPI](https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/src/path_handler.cpp)/[RPP](https://github.com/ros-navigation/navigation2/blob/d11de56438da34d55175e15cacd72fa236c9fff6/nav2_regulated_pure_pursuit_controller/src/path_handler.cpp#L61-L128) -- we don't search for the closest-point on the path in the full path scope but rather the closest point from a previous iteration's last point within a search window. That way, the window can be set to the maximum a robot can possibly move between iterations (plus margin) and remove the loop finding. 

> The issue with implementing that in the goal checker is that the goal checker's job currently doesn't include maintaining a stateful global plan, and updating it as the robot moves. That, I think, is the Controller's job, and the current_path_ passed in to the goal checker should be one that includes only present (maybe) and future path segments, not past ones.

I don't think that's always realistic, if you only plan at 1 hz but control at 20 hz, you are naturally going to make movement between them (let alone replanning every 5-10-20 seconds). I agree however that this is annoying logic to continuously reimplement, so it might be worth building a controller util that does this that we can deploy here, MPPI, RPP, etc more easily.  The versions in MPPI/RPP are slightly different, however. 

Perhaps an instance of the controller util, hosted by the controller server can be used by the goal checker? And/Or, the controller plugins are given both the global path and the pre-transformed and pruned path to use as their inputs, so we can remove some of the implementation details from each controller into centralized utilities that are computed in the server and then given to the plugins? 

I think this would move the computation of the transformed / pruned plan into the controller when calling [isGoalReached](https://github.com/ros-navigation/navigation2/blob/694a222edf7c765c9691fdd10a8025ed6b4aa8ce/nav2_controller/src/controller_server.cpp#L770-L790) so that the goal checker itself is not maintaining a path / servicing its state changes. Now, its done by the controller and the input to this function is a pre-pruned pre-transformed plan. That plan could also be given to the controllers directly, if need be/desirable. 

So I think the steps here:
- Create a new `nav2_controller` utility for path handling that performs this action based off of MPPI/RPP/Graceful/etc controllers
- Have this utility class composed into the Controller Server and compute the pruned path at some point 
- Use this path for the goal checker path 
- Consider if this utility can replace the versions (partially or fully) within RPP/MPPI/etc if we pass in this path to the controller plugins when we ask it to compute a trajectory. All is best, but I have a feeling ~1 of them might have unique specifics that make it want to handle its own from the raw path. If ~most, then we can pass it in **both** paths to decide how it wants to work with them


My hope would be either a clear decision that this PR requires a resolution to the global path is never updated issue, or agreement that we can split this off into its own enhancement request, and follow-up branch + PR down the road. I'm still interested in working on that either way, but I'm working on an agricultural robot that's got to be ready for the planting season, so my time is more limited the next few months.

SteveMacenski:
@josephduchesne any thoughts / updates?

mamariomiamo:
> Perhaps an instance of the controller util, hosted by the controller server can be used by the goal checker? And/Or, the controller plugins are given both the global path and the pre-transformed and pruned path to use as their inputs, so we can remove some of the implementation details from each controller into centralized utilities that are computed in the server and then given to the plugins?

Thinking along this line, if the **transformed_global_path** (currently computed in MPPI and graceful_controller) is available in the controller server via some getter function, we can make use of it to update the end_pose_ that is used in the goal checker https://github.com/ros-navigation/navigation2/blob/main/nav2_controller/src/controller_server.cpp#L811C16-L811C36.
Since **transformed_global_path** would be the segment of the global path that the local planner is currently following, we can consider the last pose on this path to be the **local_path_goal** and we can pass into the goal checker to check if the robot has reached it.

One downside is that we will be doing unnecessary checks before the **transformed_global_path** reaches the end of the global plan (i.e. **local_path_goal** is the same as the actual final goal).

Also if for some reason **transformed_global_path** stops updating i.e. **local_path_goal** stays the same and the robot actually reaches the local goal, goal checker would wrongly think the robot has reached the goal.

SteveMacenski:
@mamariomiamo any interest in getting involved in taking on this project? This PR has been stalled out for some time and I'm sure @josephduchesne wouldn't mind the help :-)

josephduchesne:
I'd be happy to pass the baton on. I'm weeks out from our new robot release and deployment so I haven't had a chance to come back to push this over the line 😞

mini-1235:
I was working on #4003, and faced the same problem while testing. If no one is working on this, may I take over?
> - Create a new nav2_controller utility for path handling that performs this action based off of MPPI/RPP/Graceful/etc controllers
> - Have this utility class composed into the Controller Server and compute the pruned path at some point
> - Use this path for the goal checker path

Previously, we do this in RPP
```cpp
auto transformed_plan = path_handler_->transformGlobalPlan(
    pose, params_->max_robot_pose_search_dist, params_->interpolate_curvature_after_goal);
```

Now, I moved this into controller_server, before `computeVelocityCommands` is called, and passed the transformed plan as a new argument in `computeVelocityCommands` and `isGoalReached`

Similar to this PR, the new goal checker will check if the transformed_plan's path length, if it is >, then not reached, else follow the simple goal checker logic

Result:
output


> - Consider if this utility can replace the versions (partially or fully) within RPP/MPPI/etc if we pass in this path to the controller plugins when we ask it to compute a trajectory. All is best, but I have a feeling ~1 of them might have unique specifics that make it want to handle its own from the raw path. If ~most, then we can pass it in both paths to decide how it wants to work with them

I looked through the controllers in nav2:
- Graceful: very similar to RPP's logic, except that RPP has a `interpolate_curvature_after_goal` parameter
- DWB: a little bit different, particularly this block https://github.com/ros-navigation/navigation2/blob/935ffb9925275eec72d6daae964bf605db0a6081/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L516-L527
- MPPI: 1. global_plan_up_to_inversion 2. integrated_distance vs euclidean distance 3. considering bound in costmap frame
- Rotation shim: nothing to change

SteveMacenski:
> If no one is working on this, may I take over?

Please! Note that https://github.com/ros-navigation/navigation2/pull/5387 is also somewhat related, in that it provides the utilities (prototypes) that this may want to leverage as well. You may want to read up on that thread and the ticket it is starting to address.

mini-1235:
> Please! Note that #5387 is also somewhat related, in that it provides the utilities (prototypes) that this may want to leverage as well. You may want to read up on that thread and the ticket it is starting to address.

Ok! I will open a new PR after I went through all the conversations

SteveMacenski:
https://github.com/ros-navigation/navigation2/pull/5446 to supersede - thanks @josephduchesne for getting the ball rolling!

PR #4791, opened 2024-12-11
Use PoseStampedArray

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


tonynajjar:
@SteveMacenski as usual draft PR, before I spend more time on it can you confirm this is the direction we want to go in as discussed in https://github.com/ros-navigation/navigation2/pull/4771?

nav2_msgs/srv/GetCosts.srv
```diff
@@ -1,7 +1,7 @@
 # Get costmap costs at given poses
 
 bool use_footprint
-geometry_msgs/PoseStamped[] poses
+nav2_msgs/PoseStampedArray poses
```
SteveMacenski, line 4: Q: Does this one need to be changed? I don't mind it, but I view this more as the `path.poses` list as its usual use-case and I'm not sure I see the need for the header for this particular application

SteveMacenski (changes requested):
Otherwise, no notes. Just docs for the migration guide and probably an FYI notification on Discourse/Slack would be good so folks expect it

nav2_msgs/action/NavigateThroughPoses.action
tonynajjar, line 1: Regarding what you mentioned [here](https://github.com/ros2/common_interfaces/pull/262#issuecomment-2536897202)

> we can have a stamp to gauge the age of the original request

Don't you mean gauge the age of the **goals computation**? To gauge the age of the actual **goal request**, I think we would need to add another top-level `Header`; otherwise you wouldn't be able to do it for `NavigateToPose`.

SteveMacenski: Yes, sorry, I was imprecise.

tonynajjar:
Happy new year! (ping in disguise @clalancette :smile: )

nav2_msgs/srv/GetCosts.srv
tonynajjar: agree

tools/underlay.repos
```diff
@@ -35,3 +35,8 @@ repositories:
     type: git
     url:  https://github.com/ros-navigation/nav2_minimal_turtlebot_simulation.git
     version: 091b5ff12436890a54de6325df3573ae110e912b
+  ros/common_interfaces:
```
tonynajjar, line 38: just for CI testing. Do you know when we can remove this? Are rolling binaries released nightly?

tonynajjar:
> Just docs for the migration guide

https://github.com/ros-navigation/docs.nav2.org/pull/630

>  FYI notification on Discourse/Slack would be good so folks expect it

Will do after the final review

nav2_msgs/action/ComputePathThroughPoses.action
```diff
@@ -1,5 +1,5 @@
 #goal definition
-geometry_msgs/PoseStamped[] goals
+geometry_msgs/PoseStampedArray goals
```
tonynajjar, line 2: what do you think of renaming `goals` to `poses`?

tools/underlay.repos
SteveMacenski: No, they are released monthlyish like other ROS distributions - it'll take a few weeks before we can remove it

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -557,7 +557,7 @@ def _getPathThroughPosesImpl(self, start, goals, planner_id='', use_start=False)
 
         goal_msg = ComputePathThroughPoses.Goal()
         goal_msg.start = start
-        goal_msg.goals = goals
+        goal_msg.goals.poses = goals
```
SteveMacenski, line 560: Should we populate the header stamp/frame?

nav2_system_tests/src/system/nav_through_poses_tester_node.py
```diff
@@ -94,7 +94,7 @@ def runNavigateAction(self, goal_pose: Optional[Pose] = None):
 
         self.goal_pose = goal_pose if goal_pose is not None else self.goal_pose
         goal_msg = NavigateThroughPoses.Goal()
-        goal_msg.poses = [
+        goal_msg.poses.poses = [
```
SteveMacenski, line 97: Should we populate the stamp / frame?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
```diff
@@ -75,7 +75,7 @@ class ComputePathThroughPosesAction
   {
     return providedBasicPorts(
       {
-        BT::InputPort<std::vector<geometry_msgs::msg::PoseStamped>>(
+        BT::InputPort<geometry_msgs::msg::PoseStampedArray>(
```
SteveMacenski, line 78: All the ports need to have their config docs updated for the change in type

SteveMacenski (changes requested):
I'm generally noticing that the header stamp/frame isn't being populated in some places and I think it aught to be if we're going to be using it to check for updates. I believe that all things that populate a `PoseStampedArray` should have its `header` fully populated.

nav2_msgs/action/ComputePathThroughPoses.action
SteveMacenski: I think `goals` is still a good enough name, but more I'm hesitant to change it since its used in so many places and its easy (especially w.r.t. the blackboard / BT XML / ports) to miss a few places or have users of new distributions have subtle breakages.

I'd leave it unless it was a serious bother. I still think navigating through poses' `poses` are themselves `goals`, just intermediate goals

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
tonynajjar: I don't really see anywhere where `std::vector<geometry_msgs::msg::PoseStamped>` was specifically mentioned so that we need to change it. Can you point me more explicitly?

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
tonynajjar: sure but the stamp isn't really loyal to its definition if I set it to 'now'. I would need to pass the time the goals were computed in the function and that seems unnecessary currently

nav2_system_tests/src/system/nav_through_poses_tester_node.py
tonynajjar: same as above

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
SteveMacenski: This is the "simple" commander afterall, its not perfect in any regard :-)

SteveMacenski (approved):

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_through_poses_action.hpp
SteveMacenski: An example: https://docs.nav2.org/configuration/packages/bt-plugins/actions/ComputePathThroughPoses.html

Basically, any of the BT nodes you modified, you should be able to find the `goals` type within the BT XML config docs for that respective BT node

tonynajjar: thanks, done in docs PR

Issue #4792, opened 2024-12-11
SmacPlanner backtrace failure when goal is close to the start pose


DylanDeCoeyer-Quimesis:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Iron source
- Version or commit hash:
  - 1.2.9
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
Request a path computation with the goal located on the same costmap cell as the robot.

#### Expected behavior
The SmacPlanner successfuly returns a path containing either the start pose only, or the start pose and the goal.
#### Actual behavior
The SmacPlanner fails due to a backtrace error originating from https://github.com/ros-navigation/navigation2/blob/1.2.9/nav2_smac_planner/src/node_hybrid.cpp#L704
#### Additional information
In this case, the start pose is considered as "being" the goal by the A* planner (https://github.com/ros-navigation/navigation2/blob/1.2.9/nav2_smac_planner/src/a_star.cpp#L308), so the backtrace is triggered but the first pose has no parent, leading to the error. This use case is not common, but I wouldn't expect the planner to fail.

SteveMacenski:
I think it would be reasonable to bypass all of this around here (https://github.com/ros-navigation/navigation2/blob/1.2.9/nav2_smac_planner/src/smac_planner_hybrid.cpp#L348) by checking if the start cell is the same as the goal cell and returning the single pointed path

Can submit a PR to implement for the 3 smac planners?

PR #4793, opened 2024-12-12
nav2_smac_planner: handle corner case where start and goal are on the same cell

DylanDeCoeyer-Quimesis:
This case was already properly handled in the smac_planner_2d (since PR #2488), but it was still leading to an A* backtrace failure in the smac_planner_hybrid and smac_planner_lattice. Let's harmonize the handling of this case.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4792 |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/smac_planner_hybrid.cpp
+++ b/nav2_smac_planner/src/smac_planner_hybrid.cpp
@@ -359,8 +359,13 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
   _a_star->setCollisionChecker(&_collision_checker);
 
   // Set starting point, in A* bin search coordinates
-  float mx, my;
-  if (!costmap->worldToMapContinuous(start.pose.position.x, start.pose.position.y, mx, my)) {
+  float mx_start, my_start, mx_goal, my_goal;
+  if (!costmap->worldToMapContinuous(
+    start.pose.position.x,
+    start.pose.position.y,
+    mx_start,
+    my_start))
+  {
     throw nav2_core::StartOutsideMapBounds(
             "Start Coordinates of(" + std::to_string(start.pose.position.x) + ", " +
             std::to_string(start.pose.position.y) + ") was outside bounds");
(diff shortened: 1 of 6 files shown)


nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -404,6 +414,20 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
...
 
+  // Corner case of start and goal being on the same cell
+  if (std::floor(mx_start) == std::floor(mx_goal) &&
+    std::floor(my_start) == std::floor(my_goal))
+  {
+    pose.pose = start.pose;
+    // if we have a different start and goal orientation, set the unique path pose to the goal
+    // orientation
+    if (start.pose.orientation != goal.pose.orientation) {
+      pose.pose.orientation = goal.pose.orientation;
```
SteveMacenski, line 425: Why not just always do this? I don't think we even need the if branch

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -404,6 +414,20 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
...
+    }
+    plan.poses.push_back(pose);
+    return plan;
```
SteveMacenski, line 428: Please check on debug visualizations parameter and publish the plan to the topic. Some people strongly rely on that

SteveMacenski (changes requested):
Otherwise, LGTM. Make sure to do these same comments in the State Lattice planner & add into the unit tests a plan where the start and goal are the same bin to exercise the change in unit testing

nav2_smac_planner/src/smac_planner_hybrid.cpp
DylanDeCoeyer-Quimesis: I took inspiration from the structure in smac_planner_2d, but you're right, there is no added value here.

nav2_smac_planner/src/smac_planner_hybrid.cpp
DylanDeCoeyer-Quimesis: I noticed that the plan is published independently from the value of `debug_visualizations`. So I did the same in this case.

nav2_smac_planner/test/test_smac_2d.cpp
```diff
@@ -77,6 +77,15 @@ TEST(SmacTest, test_smac_2d) {
   } catch (...) {
   }
 
+  // corner case where the start and goal are on the same cell
+  goal.pose.position.x = 0.01;
+  goal.pose.position.y = 0.01;
+
+  try {
+    planner_2d->createPlan(start, goal, dummy_cancel_checker);
```
SteveMacenski, line 85: This shouldn't throw, right? We should get a single-point path that we need to check in the return for with `EXPECT_TRUE` on the `size()`

SteveMacenski (changes requested):
Otherwise, LGTM. Test just needs to actually test the result of the exercised bit

SteveMacenski (approved):
Great work, thanks!

Issue #4115, opened 2024-02-15
graceful_controller: implement iterative selection of control point


mikeferguson:
I'm already working on this - just ticketing it for visibility

SteveMacenski:
From the wobbling in https://youtu.be/QHdrfadFhRg I was thinking about how to potentially address it. I believe you mentioned that your aim would be to back out the control point on the path for collision avoidance reasons -- which my understanding would also help with turns in confined environments where going too far ahead isn't possible. I believe that makes the controller wholesale more functional, but doesn't help with the wobbling.

So my initial thoughts last night over a notepad:
- Only recompute the control points on occasion, not every iteration. Let the robot commit to an action for longer and recompute either every `N` cycles, once the robot has driven N meters, or within some N meter radius of the final control point. Possibly with a velocity multiplier so that we can preemptively recompute it sooner when moving faster (such that if a new curve cannot be found, we have enough time to stop safely at the end of the current curve)

While not related to wobbling, something I want to try and thus documenting here: Either use gradient descent or sample different spiral parameters given the anchors of the boundary conditions and use the best by some scoring metric (distance to path, cost-distance metric, etc), My preference would be gradient descent so its not something tuned on the window sizes and such -- and thus falling into less of the traps of DWA.

SteveMacenski:
Smoothing out path before execution  --  changing gains.

ajtudela:
Note for me:
Add a paramer to backward project the goal target pose a little bit (like the docking server) when **final_rotation=false** to avoid final jittering.

PR #4795, opened 2024-12-13
graceful_controller: implement iterative selection of control points

mikeferguson:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4115 |
| Robotic platform tested on | UBR-1 |
| Does this PR contain AI generated software? | No, only real intelligence was used |

---

 * Iteratively selects a valid target pose to use with the control law. This is done by forward simulating the control law towards the target pose and rejecting the target pose if there is a collision detected.
 * Adds collision checking to the "final_rotation". If this fails, we will attempt to roll out further movement via the control law (this is especially important for non-circular robots!). With this in mind, the parameter `final_rotation` has been renamed to `prefer_final_rotation`.
 * Modifies how `final_rotation` is done a bit - this is closer to how the `graceful_controller` did it. When `prefer_final_rotation` is true, we ignore the rotation of the final pose in the path and drive straight towards (and then do an in-place rotation). Previously, this code was simply using `final_rotation` as "do we latch and rotate", which is only half the battle if you aren't using a kinematically aware planner.

 * Replaces `motion_target_dist` parameter with two new parameters: `min_lookahead` and `max_lookahead`.
 * Renames `final_rotation` to `prefer_final_rotation` (to match `graceful_controller` and also indicate that this doesn't actually ensure there is a final rotation, just that the controller will prefer to generate one if it is collision free)
 * Replace boolean `initial_rotation` and double `initial_rotation_min_angle` with `initial_rotation_tolerance` which is a better name for what this parameter does (and avoids needing two parameters to do one thing). This also involved a number of tests updates to ensure that `initial_rotation_tolerance` cannot conflict with `allow_backward`.
 * Adds parameter `v_angular_min_in_place` to make sure that  rotation commands actually work.
 * Fixed a test that was erroneously looking at `twist.angular.x`

These will be done before this converts from DRAFT:

 - [x] This still needs to be tested on robot (I'm simply opening for visibility and to see where we are in terms of code coverage)
 - [x] ~~There is a TODO around returning cmd_vel when TF fails - this seems like the wrong thing to do, need to look at it closer~~ - updated this to throw, similar to no command found
 - [x] ~~Review parameter `max_robot_pose_search_dist` - seems to be a potentially VERY large number?~~ This is consistent with other controllers in nav2
 - [x] Port orientation filter (at least NavfnPlanner doesn't set orientations)
 
#### For Maintainers:

### DIFF
--- a/nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
@@ -151,6 +151,14 @@ protected:
     const std::vector<geometry_msgs::msg::PoseStamped> & poses,
     std::vector<double> & distances);
 
+  /**
+   * @brief Control law requires proper orientations, not all planners provide them
+   * @param path Path to add orientations into
+   * @returns Path with orientations.
+   */
+  std::vector<geometry_msgs::msg::PoseStamped> addOrientations(
+    const std::vector<geometry_msgs::msg::PoseStamped> & path);
+
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::string plugin_name_;
   std::shared_ptr<nav2_costmap_2d::Costmap2DROS> costmap_ros_;
--- a/nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
@@ -108,32 +108,22 @@ public:
 
 protected:
   /**
-   * @brief Get motion target point.
-   * @param motion_target_dist Optimal motion target distance
-   * @param path Current global path
-   * @return Motion target point
-   */
-  geometry_msgs::msg::PoseStamped getMotionTarget(
-    const double & motion_target_dist,
-    const nav_msgs::msg::Path & path);
-
-  /**
    * @brief Simulate trajectory calculating in every step the new velocity command based on
    * a new curvature value and checking for collisions.
    *
-   * @param robot_pose Robot pose
-   * @param motion_target Motion target point
+   * @param motion_target Motion target point (in costmap local frame?)
    * @param costmap_transform Transform between global and local costmap
    * @param trajectory Simulated trajectory
+   * @param cmd_vel Initial command velocity during simulation
    * @param backward Flag to indicate if the robot is moving backward
    * @return true if the trajectory is collision free, false otherwise
    */
   bool simulateTrajectory(
-    const geometry_msgs::msg::PoseStamped & robot_pose,
     const geometry_msgs::msg::PoseStamped & motion_target,
     const geometry_msgs::msg::TransformStamped & costmap_transform,
     nav_msgs::msg::Path & trajectory,
-    const bool & backward);
+    geometry_msgs::msg::TwistStamped & cmd_vel,
+    bool backward);
 
   /**
    * @brief Rotate the robot to face the motion target with maximum angular velocity.
@@ -141,8 +131,7 @@ protected:
    * @param angle_to_target Angle to the motion target
    * @return geometry_msgs::msg::Twist Velocity command
    */
-  geometry_msgs::msg::Twist rotateToTarget(
-    const double & angle_to_target);
+  geometry_msgs::msg::Twist rotateToTarget(double angle_to_target);
 
   /**
    * @brief Checks if the robot is in collision
@@ -152,6 +141,15 @@ protected:
    * @return Whether in collision
    */
   bool inCollision(const double & x, const double & y, const double & theta);
+
+  /**
+   * @brief Compute the distance to each pose in a path
+   * @param poses Poses to compute distances with
+   * @param distances Computed distances
+   */
+  void computeDistanceAlongPath(
+    const std::vector<geometry_msgs::msg::PoseStamped> & poses,
+    std::vector<double> & distances);
 
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::string plugin_name_;
@@ -163,6 +161,7 @@ protected:
   Parameters * params_;
   double goal_dist_tolerance_;
   bool goal_reached_;
+  bool has_new_path_;
 
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> transformed_plan_pub_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> local_plan_pub_;
--- a/nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
@@ -153,11 +153,9 @@ protected:
 
   /**
    * @brief Control law requires proper orientations, not all planners provide them
-   * @param path Path to add orientations into
-   * @returns Path with orientations.
+   * @param path Path to add orientations into, if required
    */
-  std::vector<geometry_msgs::msg::PoseStamped> addOrientations(
-    const std::vector<geometry_msgs::msg::PoseStamped> & path);
+  void validateOrientations(std::vector<geometry_msgs::msg::PoseStamped> & path);
 
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::string plugin_name_;
@@ -169,7 +167,9 @@ protected:
   Parameters * params_;
   double goal_dist_tolerance_;
   bool goal_reached_;
-  bool has_new_path_;
+
+  // True from the time a new path arrives until we have completed an initial rotation
+  bool do_initial_rotation_;
 
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> transformed_plan_pub_;
   std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<nav_msgs::msg::Path>> local_plan_pub_;
--- a/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
@@ -45,6 +45,7 @@ struct Parameters
   double v_linear_max_initial;
   double v_angular_max;
   double v_angular_max_initial;
+  double v_angular_min_in_place;
   double slowdown_radius;
   double initial_rotation_tolerance;
   bool prefer_final_rotation;
--- a/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
@@ -52,7 +52,7 @@ struct Parameters
   bool prefer_final_rotation;
   double rotation_scaling_factor;
   bool allow_backward;
-  double in_place_collision_tolerance;
+  double in_place_collision_resolution;
 };
 
 /**
--- a/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
@@ -51,6 +51,7 @@ struct Parameters
   bool prefer_final_rotation;
   double rotation_scaling_factor;
   bool allow_backward;
+  bool add_orientations;
 };
 
 /**
--- a/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
@@ -33,7 +33,8 @@ namespace nav2_graceful_controller
 struct Parameters
 {
   double transform_tolerance;
-  double motion_target_dist;
+  double min_lookahead;
+  double max_lookahead;
   double max_robot_pose_search_dist;
   double k_phi;
   double k_delta;
@@ -45,9 +46,8 @@ struct Parameters
   double v_angular_max;
   double v_angular_max_initial;
   double slowdown_radius;
-  bool initial_rotation;
-  double initial_rotation_min_angle;
-  bool final_rotation;
+  double initial_rotation_tolerance;
+  bool prefer_final_rotation;
   double rotation_scaling_factor;
   bool allow_backward;
 };
--- a/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
+++ b/nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
@@ -47,11 +47,12 @@ struct Parameters
   double v_angular_max_initial;
   double v_angular_min_in_place;
   double slowdown_radius;
+  bool initial_rotation;
   double initial_rotation_tolerance;
   bool prefer_final_rotation;
   double rotation_scaling_factor;
   bool allow_backward;
-  bool add_orientations;
+  double in_place_collision_tolerance;
 };
 
 /**
--- a/nav2_graceful_controller/src/graceful_controller.cpp
+++ b/nav2_graceful_controller/src/graceful_controller.cpp
@@ -164,7 +164,7 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
     goal_reached_ = true;
     double angle_to_goal = tf2::getYaw(transformed_plan.poses.back().pose.orientation);
     // Check for collisions between our current pose and goal pose
-    size_t num_steps = fabs(angle_to_goal) / params_->in_place_collision_tolerance;
+    size_t num_steps = fabs(angle_to_goal) / params_->in_place_collision_resolution;
     // Need to check at least the end pose
     num_steps = std::max(static_cast<size_t>(1), num_steps);
     bool collision_free = true;
@@ -430,7 +430,7 @@ void GracefulController::validateOrientations(
   if (path.empty()) {return;}
 
   // Check if we actually need to add orientations
-  for (size_t i = 0; i < path.size() - 1; ++i) {
+  for (size_t i = 1; i < path.size() - 1; ++i) {
     if (tf2::getYaw(path[i].pose.orientation) != 0.0) {return;}
   }
 
--- a/nav2_graceful_controller/src/graceful_controller.cpp
+++ b/nav2_graceful_controller/src/graceful_controller.cpp
@@ -136,10 +136,8 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
   auto transformed_plan = path_handler_->transformGlobalPlan(
     pose, params_->max_robot_pose_search_dist);
 
-  // Add proper orientations to plan, if desired
-  if (params_->add_orientations) {
-    transformed_plan.poses = addOrientations(transformed_plan.poses);
-  }
+  // Add proper orientations to plan, if needed
+  validateOrientations(transformed_plan.poses);
 
   // Publish plan for visualization
   transformed_plan_pub_->publish(transformed_plan);
@@ -155,7 +153,7 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
       logger_, "Could not transform %s to %s: %s",
       costmap_ros_->getBaseFrameID().c_str(), costmap_ros_->getGlobalFrameID().c_str(),
       ex.what());
-    throw nav2_core::NoValidControl("Could not transform");
+    throw ex;
   }
 
   // Compute distance to goal as the path's integrated distance to account for path curvatures
@@ -166,10 +164,9 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
     goal_reached_ = true;
     double angle_to_goal = tf2::getYaw(transformed_plan.poses.back().pose.orientation);
     // Check for collisions between our current pose and goal pose
-    size_t num_steps = fabs(angle_to_goal) / 0.1;
+    size_t num_steps = fabs(angle_to_goal) / params_->in_place_collision_tolerance;
     // Need to check at least the end pose
     num_steps = std::max(static_cast<size_t>(1), num_steps);
-    // If we fail to generate an in-place rotation, maybe we need to move along path a bit more
     bool collision_free = true;
     for (size_t i = 1; i <= num_steps; ++i) {
       double step = static_cast<double>(i) / static_cast<double>(num_steps);
@@ -257,7 +254,7 @@ void GracefulController::setPlan(const nav_msgs::msg::Path & path)
 {
   path_handler_->setPlan(path);
   goal_reached_ = false;
-  has_new_path_ = true;
+  do_initial_rotation_ = true;
 }
 
 void GracefulController::setSpeedLimit(
@@ -299,12 +296,12 @@ bool GracefulController::simulateTrajectory(
   next_pose.pose.orientation.w = 1.0;
 
   // Should we simulate rotation initially?
-  bool sim_initial_rotation = has_new_path_ && (params_->initial_rotation_tolerance > 0.0);
+  bool sim_initial_rotation = do_initial_rotation_ && params_->initial_rotation;
   double angle_to_target =
     std::atan2(motion_target.pose.position.y, motion_target.pose.position.x);
   if (fabs(angle_to_target) < params_->initial_rotation_tolerance) {
     sim_initial_rotation = false;
-    has_new_path_ = false;
+    do_initial_rotation_ = false;
   }
 
   double distance = std::numeric_limits<double>::max();
@@ -370,11 +367,8 @@ geometry_msgs::msg::Twist GracefulController::rotateToTarget(double angle_to_tar
   geometry_msgs::msg::Twist vel;
   vel.linear.x = 0.0;
   vel.angular.z = params_->rotation_scaling_factor * angle_to_target * params_->v_angular_max;
-  if (vel.angular.z < 0.0) {
-    vel.angular.z = std::min(vel.angular.z, -params_->v_angular_min_in_place);
-  } else {
-    vel.angular.z = std::max(vel.angular.z, params_->v_angular_min_in_place);
-  }
+  vel.angular.z = std::copysign(1.0, vel.angular.z) * std::min(abs(vel.angular.z),
+      params_->v_angular_min_in_place);
   return vel;
 }
 
@@ -429,31 +423,26 @@ void GracefulController::computeDistanceAlongPath(
   }
 }
 
-std::vector<geometry_msgs::msg::PoseStamped> GracefulController::addOrientations(
-  const std::vector<geometry_msgs::msg::PoseStamped> & path)
+void GracefulController::validateOrientations(
+  std::vector<geometry_msgs::msg::PoseStamped> & path)
 {
-  std::vector<geometry_msgs::msg::PoseStamped> oriented_path;
-  oriented_path.resize(path.size());
-  if (path.empty()) {
-    // This really shouldn't happen
-    return oriented_path;
+  // This really shouldn't happen
+  if (path.empty()) {return;}
+
+  // Check if we actually need to add orientations
+  for (size_t i = 0; i < path.size() - 1; ++i) {
+    if (tf2::getYaw(path[i].pose.orientation) != 0.0) {return;}
   }
 
-  // The last pose will already be oriented since it is our goal
-  oriented_path.back() = path.back();
-
   // For each pose, point at the next one
-  for (size_t i = 0; i < oriented_path.size() - 1; ++i) {
-    // XY will be unchanged
-    oriented_path[i] = path[i];
+  // NOTE: control loop will handle reversing logic
+  for (size_t i = 0; i < path.size() - 1; ++i) {
     // Get relative yaw angle
     double dx = path[i + 1].pose.position.x - path[i].pose.position.x;
     double dy = path[i + 1].pose.position.y - path[i].pose.position.y;
     double yaw = std::atan2(dy, dx);
-    oriented_path[i].pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(yaw);
+    path[i].pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(yaw);
   }
-
-  return oriented_path;
 }
 
 }  // namespace nav2_graceful_controller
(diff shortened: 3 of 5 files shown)


mikeferguson:
A few other things noted when porting to UBR-1:
 * [ ] need to add a "min_in_place_vel_theta" equivalent - most robots can't generate arbitrarily low angular rotations

mikeferguson:
This was the updated config for testing this on the UBR-1 (moving from the version of graceful controller in my repo): https://github.com/mikeferguson/ubr_reloaded/commit/a24ad85822dcfef24bd87d5fb1e9ae11351f779a

mikeferguson:
Note for posterity: I did also test "allow_backward" - and it still works

mikeferguson:
Notes on things that are still different versus my controller (that maybe we'd want to do in the future):

 * In the control law - the `slowdown_radius` is a poor approximation for actual deceleration limits. Really, you want to cap the velocity based on knowing you can slow to a stop by the time you arrive at the pose.
 * The collision checking in the iterative loop assumes you are starting at max speed (not at your current speed). This is probably not that big of a deal, but if you're going very slowly in a tight area, the controller might wobble a bit more.
 * `add_orientations` just adds the orientations but doesn't do any more sophisticated filtering - I'm not actually convinced that additional filtering is really needed though (and, it always had a few odd edge cases - if we were to implement something here, it should really be a bit more sophisticated than what was in my controller).

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -360,6 +414,48 @@ bool GracefulController::inCollision(const double & x, const double & y, const d
...
+  distances[0] = d;
+  // Compute remaining poses
+  for (size_t i = 1; i < poses.size(); ++i) {
+    d += nav2_util::geometry_utils::euclidean_distance(poses[i - 1].pose, poses[i].pose);
+    distances[i] = d;
+  }
+}
+
+std::vector<geometry_msgs::msg::PoseStamped> GracefulController::addOrientations(
+  const std::vector<geometry_msgs::msg::PoseStamped> & path)
```
mikeferguson, line 433: I should probably add a comment here - that we don't do anything special for reversing - because it will be handled up in L227-232

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -199,24 +155,109 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
       logger_, "Could not transform %s to %s: %s",
       costmap_ros_->getBaseFrameID().c_str(), costmap_ros_->getGlobalFrameID().c_str(),
       ex.what());
-    return cmd_vel;
+    throw nav2_core::NoValidControl("Could not transform");
```
SteveMacenski, line 158: You can just re-throw the TF exception `throw ex;`

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -199,24 +155,109 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
-    throw nav2_core::NoValidControl("Collision detected in the trajectory");
+  // Compute distance to goal as the path's integrated distance to account for path curvatures
+  double dist_to_goal = nav2_util::geometry_utils::calculate_path_length(transformed_plan);
+
+  // If we've reached the XY goal tolerance, just rotate
+  if (dist_to_goal < goal_dist_tolerance_ || goal_reached_) {
+    goal_reached_ = true;
+    double angle_to_goal = tf2::getYaw(transformed_plan.poses.back().pose.orientation);
+    // Check for collisions between our current pose and goal pose
+    size_t num_steps = fabs(angle_to_goal) / 0.1;
```
SteveMacenski, line 169: `0.1` should be a parameter for the collision search resolution

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -199,24 +155,109 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
+    // Need to check at least the end pose
+    num_steps = std::max(static_cast<size_t>(1), num_steps);
+    // If we fail to generate an in-place rotation, maybe we need to move along path a bit more
```
SteveMacenski, line 172: Does this comment belong here?

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -199,24 +155,109 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
 
-  // Generate and publish local plan for debugging / visualization
-  nav_msgs::msg::Path local_plan;
-  if (!simulateTrajectory(pose, motion_target, costmap_transform, local_plan, reversing)) {
-    throw nav2_core::NoValidControl("Collision detected in the trajectory");
+  // Compute distance to goal as the path's integrated distance to account for path curvatures
+  double dist_to_goal = nav2_util::geometry_utils::calculate_path_length(transformed_plan);
+
+  // If we've reached the XY goal tolerance, just rotate
+  if (dist_to_goal < goal_dist_tolerance_ || goal_reached_) {
```
SteveMacenski, line 165: Why isn't this considering if we want to rotate to the goal using the `prefer_final_rotation` parameter?

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -316,11 +365,16 @@ bool GracefulController::simulateTrajectory(
   return true;
 }
 
-geometry_msgs::msg::Twist GracefulController::rotateToTarget(const double & angle_to_target)
+geometry_msgs::msg::Twist GracefulController::rotateToTarget(double angle_to_target)
 {
   geometry_msgs::msg::Twist vel;
   vel.linear.x = 0.0;
   vel.angular.z = params_->rotation_scaling_factor * angle_to_target * params_->v_angular_max;
+  if (vel.angular.z < 0.0) {
```
SteveMacenski, line 373: You could remove the branch with 

```
vel.angular.z = std::copysign(1.0, vel.angular.z) * std::min(abs(vel.angular.z), params_->v_angular_min_in_place);
```

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -360,6 +414,48 @@ bool GracefulController::inCollision(const double & x, const double & y, const d
...
+  double d = std::hypot(poses[0].pose.position.x, poses[0].pose.position.y);
+  distances[0] = d;
+  // Compute remaining poses
+  for (size_t i = 1; i < poses.size(); ++i) {
+    d += nav2_util::geometry_utils::euclidean_distance(poses[i - 1].pose, poses[i].pose);
+    distances[i] = d;
+  }
+}
+
+std::vector<geometry_msgs::msg::PoseStamped> GracefulController::addOrientations(
```
SteveMacenski, line 432: I say just always do it - as long as orientations aren't already populated bc its a feasible path. That also removes a parameter :-)

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -243,58 +284,66 @@ void GracefulController::setSpeedLimit(
...
 
-  // First pose
+  // First pose is robot current pose
   geometry_msgs::msg::PoseStamped next_pose;
   next_pose.header.frame_id = costmap_ros_->getBaseFrameID();
   next_pose.pose.orientation.w = 1.0;
-  trajectory.poses.push_back(next_pose);
+
+  // Should we simulate rotation initially?
+  bool sim_initial_rotation = has_new_path_ && (params_->initial_rotation_tolerance > 0.0);
```
SteveMacenski, line 302: I think there should be an explicit enable `bool` rather than checking if the `tol > 0`

SteveMacenski (changes requested):
I'll admit, this was reasonably hard for me to review with the diffs and reordering of the main code, so I'm not 100% confident in not having missed something on an initial look. I'll probably need another complete read through after changes to sanity check.

@ajtudela since this touches the docking code and I know you're thinking about these areas - want to review / test?

SteveMacenski:
I'm curious about your thoughts from the ticket brainstorming https://github.com/ros-navigation/navigation2/issues/4115#issuecomment-1973569942

nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
```diff
@@ -44,12 +45,13 @@ struct Parameters
...
+  double v_angular_min_in_place;
   double slowdown_radius;
-  bool initial_rotation;
-  double initial_rotation_min_angle;
-  bool final_rotation;
+  double initial_rotation_tolerance;
+  bool prefer_final_rotation;
   double rotation_scaling_factor;
   bool allow_backward;
+  bool add_orientations;
```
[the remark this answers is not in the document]
mikeferguson: We need orientations for any pose that ends up being a candidate for the target_pose of the control law (in order to simulate the control law - even if we end up discarding that candidate and moving to another one when the simulation fails) - so we need orientations on far more than just the back of the path.

As for "check if non-zero" - since there are default values for geometry_msgs::msg::Quaternion, we no longer end up with invalid quaternions in ROS 2. Therefore, I'm not sure we want to magically change any orientation that is basically a "zero" yaw angle. For instance, something like the lattice planner uses a set of discrete headings - and so a fairly large number of poses in the path will have "zero" headings, and we probably don't want to re-orient those towards the next pose in the path since that may no longer actually be a valid path if your robot is non-circular.

mikeferguson:
> I'm curious about your thoughts from the ticket brainstorming [#4115 (comment)](https://github.com/ros-navigation/navigation2/issues/4115#issuecomment-1973569942) (or is the wobbling here more or less gone with this method recomputing at the full rate?)

I expect most of the wobbling will be gone since our motion_target can be so much farther away in most cases

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: We disagree on this, but I will switch it back to that style

nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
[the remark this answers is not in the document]
SteveMacenski: What I'm thinking of is to check first if all path points except the start and end `i = 1 -> N-1` and if **all** are `0`, then I think its a safe bet that its not kinematically feasible. I don't think we should change individual ones, but all-or-nothing based on that.

nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
```diff
@@ -163,6 +169,7 @@ class GracefulController : public nav2_core::Controller
   Parameters * params_;
   double goal_dist_tolerance_;
   bool goal_reached_;
+  bool has_new_path_;
```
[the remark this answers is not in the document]
mikeferguson: I think maybe this just needs to be renamed - it's not so much that "we have a new path" - but rather "we are still doing the initial rotation"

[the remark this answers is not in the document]
SteveMacenski: OK! I don't feel overly strongly about this - just felt like a reasonable separation of concerns

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: Or I suppose we could keep it as that as long as the docs make it clear that if its set to `0.0` or ` < 0.0` it'll be interpreted as disabled. We have _some_ parameters like that around. I'd like add in the `-1` case as also disabled too

nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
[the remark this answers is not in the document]
mikeferguson: I think you're actually right that the name is bad though - because it doesn't JUST indicate a new path has been set (you can see we set this flag when a new path arrives, but we don't clear it until we have successfully done the whole initial rotation)

mikeferguson:
> @ajtudela since this touches the docking code and I know you're thinking about these areas - want to review / test?

I don't think this affects the docking code at all, because that is only using the SmoothControlLaw - which is unmodified

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: I reverted to two parameters - which also seems to help the diff look smaller since git is less confused

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: So, there are really two different features at work here:

1) `prefer_final_rotation` is actually the feature down on L212 - when we are close to the goal (less than max lookahead) and prefer final rotation, we will force the target_pose heading to be the "heading to target pose" (imagine robot is at 0,0,0 and the target_pose is 0.1,0,1.57 - if you feed this into the control law, you take a massive sweeping arc to try and approach that 1.57 heading straight on. Instead, we use atan2(0.0, 0.1) and will end up doing no sweeping turn and then just turning in place at the end). I've found this to be probably the most important feature for actually successfully following the majority of paths regardless of the planner than generated them (and especially true for planners that are not kinematics aware).

2) When you've reached the XY goal, if `prefer_final_rotation` is true you probably have a huge heading error - but even if you aren't preferring final rotation you might have a small heading error (usually, we satisfy the heading requirement before we satisfy XY - but, if your localization is a bit unstable, that may not be the case). In either case, the control law gets quite unstable when the XY error is so small, and it can generate huge swooping motions to correct the heading, so you really ALWAYS want to do that final bit of rotation when you are already sitting on top of your goal.

In some ways, this second part is like the `latch_xy_goal_tolerance` behavior that other controllers have - in my controller, I actually had that parameter, and `goal_reached_` (as this code calls it) wouldn't be set to true, but rather to `latch_xy_goal_tolerance`. Regardless, in the existing graceful_controller outside nav2, you always entered this rotation behavior even if latch and prefer were set to false - you just wouldn't have it be sticky unless latch was true.

mikeferguson:
@SteveMacenski all review comments have been addressed - this is ready for another look when you get a chance

ajtudela:
> I'll admit, this was reasonably hard for me to review with the diffs and reordering of the main code, so I'm not 100% confident in not having missed something on an initial look. I'll probably need another complete read through after changes to sanity check.
> 
> @ajtudela since this touches the docking code and I know you're thinking about these areas - want to review / test?

Of course I'll test it on our robots on Monday morning. 

This part doesn't change the controller of the docking but this is exactly the behaviour I need for the new following server. I'll have to figure out how to integrate this part into the server reusing most of the code.

ajtudela:
I tested the improvements with the UBR-1 this morning. Although I don't really like the new order of the "steps" (rotation before movement, control, final rotation, ...) because it's harder (for me) to read, I have to admit that these changes take care of the wobbling much better than before.

Tomorrow I'll test all the combinations for the new parameters to make sure I haven't missed anything.

I recorded a little video if you want to see the changes in motion:
[new_graceful.webm](https://github.com/user-attachments/assets/2ad422cc-65c3-4aa4-9891-e0681666a685)

mikeferguson:
> I have to admit that these changes take care of the wobbling much better than before.

What are you running for the k_phi/k_delta parameters with the controller? I found that the defaults in the controller right now are pretty aggressive on heading correction (which causes both wobbling, and divergence from the path - which appears to be also causing lots of replanning for you).

I went with lower values that were actually the defaults in https://github.com/mikeferguson/graceful_controller:

```yaml
GracefulController:
      plugin: "nav2_graceful_controller::GracefulController"
      v_linear_min: 0.1
      v_linear_max: 1.0
      v_angular_max: 2.8
      v_angular_min_in_place: 0.6
      max_lookahead: 1.25
      initial_rotation: true
      initial_rotation_tolerance: 0.25
      prefer_final_rotation: true
      # The defaults aren't as good as this (more wiggling)
      k_phi: 2.0
      k_delta: 1.0
      beta: 0.4
      lambda: 2.0
      # This isn't quite comparable to using actual acceleration limits
      # (but works well in practice)
      slowdown_radius: 0.75
```

I didn't update these defaults - at first thinking it might impact some people's docking servers - but upon further review, it looks like that isn't the case (the docking server has it's own parameter set). @SteveMacenski thoughts? Should we update the defaults here?

ajtudela:
I used these:

```yaml
    Graceful:
      plugin: nav2_graceful_controller::GracefulController
      transform_tolerance: 0.5
      motion_target_dist: 1.2
      initial_rotation: true
      initial_rotation_min_angle: 0.75
      final_rotation: true
      allow_backward: false
      k_phi: 3.0
      k_delta: 2.0
      beta: 0.4
      lambda: 2.0
      v_linear_min: 0.1
      v_linear_max: 1.0
      v_angular_max: 5.0
      slowdown_radius: 0.5
```

But I agree with you. I had to lower k_phi/k_delta for the docking server. I'd try with these settings tomorrow.

mikeferguson:
> I had to lower k_phi/k_delta for the docking server

I think I might have actually been the one who chose those higher values for the docking server initially - that is an application where you want the robot to come at the charge dock as straight-on as possible, so the values make sense for docking. I think they are too high for general navigation controlling (but it wasn't something I caught back when we originally added the controller here)

mikeferguson:
@SteveMacenski bumping this - wasn't sure if you maybe missed my comment that this was ready for another review

nav2_graceful_controller/src/parameter_handler.cpp
```diff
@@ -52,21 +54,26 @@ ParameterHandler::ParameterHandler(
...
     node, plugin_name_ + ".v_angular_max", rclcpp::ParameterValue(1.0));
+  declare_parameter_if_not_declared(
+    node, plugin_name_ + ".v_angular_min_in_place", rclcpp::ParameterValue(0.25));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".slowdown_radius", rclcpp::ParameterValue(1.5));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".initial_rotation", rclcpp::ParameterValue(true));
   declare_parameter_if_not_declared(
-    node, plugin_name_ + ".initial_rotation_min_angle", rclcpp::ParameterValue(0.75));
+    node, plugin_name_ + ".initial_rotation_tolerance", rclcpp::ParameterValue(0.25));
```
SteveMacenski, line 64: That's pretty tight - we'd stop whenever the angle is > 14 deg off. I don't think that's a sensible default behavior to have the robot stop that much. Is that what's causing all the halt and rotates in the video that @ajtudela posted? 

Maybe the value before of ~43 deg was too high? But I think this is definitely too low

nav2_graceful_controller/src/parameter_handler.cpp
```diff
@@ -52,21 +54,26 @@ ParameterHandler::ParameterHandler(
...
   declare_parameter_if_not_declared(
-    node, plugin_name_ + ".final_rotation", rclcpp::ParameterValue(true));
+    node, plugin_name_ + ".prefer_final_rotation", rclcpp::ParameterValue(true));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".rotation_scaling_factor", rclcpp::ParameterValue(0.5));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".allow_backward", rclcpp::ParameterValue(false));
+  declare_parameter_if_not_declared(
+    node, plugin_name_ + ".in_place_collision_tolerance", rclcpp::ParameterValue(0.1));
```
SteveMacenski, line 72: `in_place_collision_resolution` perhaps? Tolerance feels like a threshold whereas this is a resolution or granularity

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -360,6 +408,43 @@ bool GracefulController::inCollision(const double & x, const double & y, const d
...
+}
+
+void GracefulController::validateOrientations(
+  std::vector<geometry_msgs::msg::PoseStamped> & path)
+{
+  // This really shouldn't happen
+  if (path.empty()) {return;}
+
+  // Check if we actually need to add orientations
+  for (size_t i = 0; i < path.size() - 1; ++i) {
```
SteveMacenski, line 433:
```suggestion
  for (size_t i = 1; i < path.size() - 1; ++i) {
```

I would see paths having the robot's start pose also populated with the goal but others are not. We don't do that in any nav2 planner to-date, but I could easily imagine it

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -199,24 +153,108 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
+      slowdown_pub_->publish(slowdown_marker);
+      // Publish the local plan
+      local_plan.header = transformed_plan.header;
+      local_plan_pub_->publish(local_plan);
+      // Successfully found velocity command
+      return cmd_vel;
+    }
+  }
+
+  throw nav2_core::NoValidControl("Collision detected in trajectory");
```
SteveMacenski, line 250: Or also that there is no path further away than `params_->min_lookahead`. I don't see how that is gracefully handled for short paths (or very long lookaheads)

SteveMacenski (commented):
>  I found that the defaults in the controller right now are pretty aggressive on heading correction (which causes both wobbling, and divergence from the path - which appears to be also causing lots of replanning for you).

> but upon further review, it looks like that isn't the case (the docking server has it's own parameter set). @SteveMacenski thoughts? Should we update the defaults here?

> I think I might have actually been the one who chose those higher values for the docking server initially - that is an application where you want the robot to come at the charge dock as straight-on as possible, so the values make sense for docking. I think they are too high for general navigation controlling (but it wasn't something I caught back when we originally added the controller here)

We can change those, definitely! I want all defaults to have a really nice out of the box experience to showcase algorithm's values without the need of heavy tuning for 'most' applications.

nav2_graceful_controller/src/parameter_handler.cpp
mikeferguson: An initial rotation only happens when a new plan arrives - I suppose if you have continuous replanning running, this would be a problem - but is that the default out of the box?

nav2_graceful_controller/src/parameter_handler.cpp
mikeferguson: Ah yeah, that sounds better - I was struggling to come up with the right name for this

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: Once the goal is < max_lookahead away, we stop enforcing the requirement that target distance > min_lookahead

SteveMacenski: I don't see that in play - maybe I'm missing something. What I think I see is that:
- when `dist_to_target > params_->max_lookahead`, we pass as we're too far
- when `dist_to_goal < params_->max_lookahead` we go through the loop, maybe adjusting the orientation (maybe not)
- When we don't find anything in that regime that works, in `dist_to_target < params_->min_lookahead`, we break the loop and throw an exception

If there's collisions in the full min_lookahead to max_lookahead range and have path options in that range, we throw. This makes sense.

However, if that path itself is shorter than min_lookahead, I don't see any handling. The `goal_dist_tolerance_` and `min_lookahead` are not related or have a requirement on their comparative relationship that would cause a previous rotate-to-heading action to strictly occur. 

Instead, I see on L206 (`double dist_to_target = target_distances[i];`) we find the distance, then its less than the lookahead minimum (`dist_to_target < params_->min_lookahead`) and then `break;`s into the throw.

nav2_graceful_controller/src/parameter_handler.cpp
SteveMacenski: > I suppose if you have continuous replanning running, this would be a problem - but is that the default out of the box?

Yes :-)

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: So, if you don't ever allow an in-place rotation at the end and miss your goal tolerances, you're going to make massive sweeping arcs (well off the original plan) to get to the goal. I can't imagine that is what users of an ackermann robot are going to want.

I can add a parameter to disable ever doing final rotations, but I think my documentation would probably be "if you find yourself wanting this, use a different controller"

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: Maybe I don't understand your question here - what are you expecting for "gracefully handling" short paths?

* L209 - as you noted, ignores targets that are too far away
* L211 - IF dist_to_goal < max_lookahead then apply optionally prefer_final_rotation and try to simulate to the target (if this fails, we iterate back through the loop and try the next closest target)
* L218 - ELSE (AKA if dist_to_goal >= max_lookahead) check if (dist_to_target < min_lookahead) - in this case, we have lots of path left, but everything between min and max_lookahead is not simulatable. I have a `break` here rather than a `continue` because every other target should be closer... and **as I write this, I'm realizing that may not be true if we have a path that doubles back onto itself. Is your comment actually suggesting that we just change the `break` on L220 to a `continue`?**

`min_lookahead` should be less than `max_lookahead` (we can add some checking parameter settings to ensure this - but logically, setting min > max is just not going to work).

`dist_to_target` should also always be <= `dist_to_goal` (which we compute on L160) - because at worst case our target is the end of the path.

`goal_dist_tolerance_` isn't involved in this part of the code - so I'm not sure what that comment is about.

nav2_graceful_controller/src/parameter_handler.cpp
mikeferguson: I'll revert this change then, and add some notes to the documentation

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: > I can add a parameter to disable ever doing final rotations, but I think my documentation would probably be "if you find yourself wanting this, use a different controller"

Perhaps that is the right answer, but let me bring this up and if you still think that's the play, I'm OK with that.

If you have an ackermann vehicle, then I imagine alot of those users are going to be using the Smac Planners that generate circular minimum-turning-radius curves to kinematically make sure they get to the goal orientation within their constraints. Do you think that the arc this would generate would be that much more extreme than the arc that the minimum turning radius circle would create that the carrot on the path plan would be following?

You have a better intuition on this planner, so I default to your thoughts

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: What I mean is what happens when we have a very short path? If we have a minimum lookahead of 1m (for some large robot) and we've replanned within 0.7m to the goal, but we have a goal tolerance of 0.3m? 

I think right now, we're throw as the distance is < min lookahead and simply fail to control ever. 

> dist_to_target should also always be <= dist_to_goal (which we compute on L160) - because at worst case our target is the end of the path.

OH **dist_to_target** vs **dist_to_goal**, that totally escaped me in reading it.

SteveMacenski: >  I have a break here rather than a continue because every other target should be closer... and as I write this, I'm realizing that may not be true if we have a path that doubles back onto itself. 

I think that should be fine with the looping over itself, since the distances are path integrated distances in both `computeDistanceAlongPath()` and `calculate_path_length()`.

mikeferguson: > I think that should be fine with the looping over itself, since the distances are path integrated distances in both `computeDistanceAlongPath()` and `calculate_path_length()`, no?

Ah yes - OK. that part works.

nav2_graceful_controller/src/graceful_controller.cpp
mikeferguson: Two thoughts on this:

 * I'm not sure there is a way to actually enforce a "minimum turning radius" with this control law. I can't think of a straight forward way to derive k1/k2 based on a minimum turning radius...
 * While this controller may not generate more "extreme" arcs, they certainly aren't kinematically aware (since the derivation of the control law is based on differential drive). I would think if you've gone through the effort of configuring and tuning a kinematically aware planner, you want to rely on that for planning larger motions (rather than letting a "dumber" controller just kinda make stuff up).

It just seems like this is the wrong tool for an Ackerman robot. I've seen it used on diff drive, holonomic, semi-holonomic powered casters (it really shines here because you mostly act like a diff drive and don't make crazy movements), and a walking robot - but not something with harder constraints like Ackerman.

SteveMacenski: OK - works for me. It just wouldn't surprise me if a ticket came in in a few years wanting to use it or a PR making it parameterized, but we can address it if/when that time comes. 

I think if you make the note in the readme / config guide / [navigation algorithms table](https://docs.nav2.org/plugins/index.html#controllers) about which robot types it supports, we're good to go. 

> I'm not sure there is a way to actually enforce a "minimum turning radius" with this control law. I can't think of a straight forward way to derive k1/k2 based on a minimum turning radius...

Just as a note: Pure Pursuit used by the DARPA grand challenge and alot of Nav2 ackermann users is also not kinematically aware. It works 'good enough' even without promises

SteveMacenski:
@ajtudela any thoughts here? This is your work too, so I want to make sure you have a chance to review and give thoughts!

nav2_graceful_controller/src/graceful_controller.cpp
ajtudela: Mm, the control law is derived from the unicycle model because it was intended to be used in a differentially driven wheelchair. So I don't think it was intended to be used in other kinds of vehicles, especially ackermans.

With this in mind, the control law explodes when your dist_to_goal -> 0, so skipping the target heading or just rotating when reaching the goal is necessary. This is the big problem with this controller.

> I think if you make the note in the readme / config guide / [navigation algorithms table](https://docs.nav2.org/plugins/index.html#controllers) about which robot types it supports, we're good to go.

When I wrote the docs for the controller, I only specified it for differential drive: https://docs.nav2.org/plugins/index.html#controllers

ajtudela:
I tested heavily this week and it works pretty well, even in narrow corridors.

I think there's missing coverage in L212-216 (or I'm ready Codecov wrong).

Otherwise, LGTM.

mikeferguson:
@SteveMacenski what else do we need to get this merged?

SteveMacenski (approved):

SteveMacenski:
@mikeferguson a couple of small updates in the docs still needed, but code is good -- thanks for the revisions and help! https://github.com/ros-navigation/docs.nav2.org/pull/624

Issue #4734, opened 2024-10-29
ClearEntireCostmap returns before all costmap layer include valid data


schupf2:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04.5 LTS
- ROS2 Version:
  - Humble binaries
- Version or commit hash:
  - 11.1.13-0jammy
- DDS implementation:
  - Cyclonedds

#### Steps to reproduce issue
When working with a costmap with more then one different layers (in our case, we use `["static_layer", "obstacle_layer", "inflation_layer"]`), we face a problem after clearing the costmap: The the `nav2_msgs::srv::ClearEntireCostmap` service returns after the costmap is cleared. However, at this point no scanner measurements are considered in the obstacle layer. Thus, the space which is not considered in the static layer is marked as free for a short time until the next scanner measurement is included in the obstacle layer.

#### Expected behavior
There should be an option to either enforce waiting on information in all layers being present  before returning from the ClearEntireCostmap service call, or some option to verify the presence of information on all layers.

#### Actual behavior
After the ClearEntireCostmap service call the costmap only shows values form that static layer and no scanner measurements are yet considered in the obstacle layer. Therefore, obstacles in front of the robot are ignored for a brief time.

SteveMacenski:
That is why there is a ‘isCurrent’ API that is checked before planning or control is allowed to make sure there is new data before the cost map is allowed to be used. Why the cost map might be in a state without valid data after a clear, it is not used until the data is again valid for all layers. Does that make sense?

schupf2:
Thank you for the quick answer - that definitely makes sense :) .
However, if one uses the the collision checker directly (and not the whole controller) - the check is not performed and one needs to get a hold of the LayeredCostmap or the Costmap2DROS in order to be able to perform the check. So maybe a service (similar to e.g., ClearEntireCostmap) that waits for isCurrent to evaluate to true, might be a nice feature.

SteveMacenski:
A collision checker though would be used inside of a controller/planner plugin, so if the controller / planner is not being utilized until data is valid yet, then there is no issue :-) There shouldn't be any calls to costmap data structures by the task server's algorithms until `isCurrent` is true

SteveMacenski:
[...]

PR #4800, opened 2024-12-17
Fix isCurrent in obstacle layer and assure CostmapTopicCollisionChecker only uses current costmaps

schupf2:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4734 |
| Robotic platform tested on | -- |

---

* The costmap is only published if it's current, such that the `CostmapTopicCollisionChecker` can be relied on again.
* Fix problems in the obstacle layer, causing it to return `true` on `isCurrent()`, after a reset (even if no data was received).
* For a detailed argumentation, please have a look at the communications in #4734 

* I didn't add documentation so far, but am open to suggestions. Maybe the change regarding the costmap publishing could be added.

---

* The same changes that were made to the obstacle layer are also required for the STVL. I will open a related pull request once this one was merged.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -538,7 +538,7 @@ Costmap2DROS::mapUpdateLoop(double frequency)
         auto map_current = isCurrent();
         if (map_current &&  // only update costmap if it's current
           ((last_publish_ + publish_cycle_ < current_time) ||  // publish_cycle_ is due
-          (current_time < last_publish_)))     // time has moved backwards, probably due to a switch to sim_time // NOLINT
+          (current_time < last_publish_)))  // time has moved backwards, probably due to a switch to sim_time // NOLINT
         {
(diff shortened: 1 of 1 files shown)


schupf2:
> @schupf2, your PR has failed to build. Please check CI outputs and resolve issues. You may need to rebase or pull in `main` due to API changes (or your contribution genuinely fails).

Thank you - I just saw it as well :).
I probably already fixed the problem, but meanwhile I'll change this PR to Draft.

SteveMacenski:
@DLu please review wrt https://github.com/ros-navigation/navigation2/issues/4734's thread

SteveMacenski:
@schupf2 I'm going through older stalled PRs pre-Lyrical release. This seems a valuable contribution, can you highlight what problems remain and if we can keep pushing this in the coming weeks?

Maybe this has actually been resolved by now via other PRs

SteveMacenski:
Actually I think we fixed this already!

PR #4807, opened 2024-12-19
Fix trajectory generation bug in docking controller

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Morphia |

---
This fixes the trajectory generation for collision checking in the docking controller. Previously it used a for loop which iterated until the `projection_time_` was reached regardless of whether the last pose of the trajectory had reached the target. This generated a lot of spurious poses with the same position but different orientations.

This has been replaced, using the same approach as the graceful_controller, with a do-while loop that also checks for this distance.

| Before | After |
| -- | -- |
| before | after |

---

#### For Maintainers: 

### DIFF
--- a/nav2_docking/opennav_docking/src/controller.cpp
+++ b/nav2_docking/opennav_docking/src/controller.cpp
@@ -183,7 +183,7 @@ bool Controller::isTrajectoryCollisionFree(
 
     // Check if we reach the goal
     distance = nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose);
-  }while(distance > 1e-2 && trajectory.poses.size() < max_iter);
+  }while(distance > 1e-2 || trajectory.poses.size() < max_iter);
 
   trajectory_pub_->publish(trajectory);
 
--- a/nav2_docking/opennav_docking/src/controller.cpp
+++ b/nav2_docking/opennav_docking/src/controller.cpp
@@ -145,8 +145,7 @@ bool Controller::isTrajectoryCollisionFree(
 
   // Generate path
   double distance = std::numeric_limits<double>::max();
-  const double resolution = v_linear_max_ * simulation_time_step_;
-  unsigned int max_iter = static_cast<unsigned int>(projection_time_ / simulation_time_step_);
+  unsigned int max_iter = static_cast<unsigned int>(ceil(projection_time_ / simulation_time_step_));
 
   do{
     // Apply velocities to calculate next pose
@@ -184,7 +183,7 @@ bool Controller::isTrajectoryCollisionFree(
 
     // Check if we reach the goal
     distance = nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose);
-  }while(distance > resolution && trajectory.poses.size() < max_iter);
+  }while(distance > 1e-2 && trajectory.poses.size() < max_iter);
 
   trajectory_pub_->publish(trajectory);
 
--- a/nav2_docking/opennav_docking/src/controller.cpp
+++ b/nav2_docking/opennav_docking/src/controller.cpp
@@ -144,7 +144,11 @@ bool Controller::isTrajectoryCollisionFree(
   }
 
   // Generate path
-  for (double t = 0; t < projection_time_; t += simulation_time_step_) {
+  double distance = std::numeric_limits<double>::max();
+  const double resolution = v_linear_max_ * simulation_time_step_;
+  unsigned int max_iter = static_cast<unsigned int>(projection_time_ / simulation_time_step_);
+
+  do{
     // Apply velocities to calculate next pose
     next_pose.pose = control_law_->calculateNextPose(
       simulation_time_step_, target_pose, next_pose.pose, backward);
@@ -177,7 +181,10 @@ bool Controller::isTrajectoryCollisionFree(
       trajectory_pub_->publish(trajectory);
       return false;
     }
-  }
+
+    // Check if we reach the goal
+    distance = nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose);
+  }while(distance > resolution && trajectory.poses.size() < max_iter);
 
   trajectory_pub_->publish(trajectory);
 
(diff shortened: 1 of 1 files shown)


nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -144,7 +144,11 @@ bool Controller::isTrajectoryCollisionFree(
   }
 
   // Generate path
-  for (double t = 0; t < projection_time_; t += simulation_time_step_) {
+  double distance = std::numeric_limits<double>::max();
+  const double resolution = v_linear_max_ * simulation_time_step_;
+  unsigned int max_iter = static_cast<unsigned int>(projection_time_ / simulation_time_step_);
```
SteveMacenski, line 149: I'd make sure to `ceil` the output of the division so that if there are fractional timesteps, we use all. Casting rounds down always.

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -144,7 +144,11 @@ bool Controller::isTrajectoryCollisionFree(
   }
 
   // Generate path
-  for (double t = 0; t < projection_time_; t += simulation_time_step_) {
+  double distance = std::numeric_limits<double>::max();
+  const double resolution = v_linear_max_ * simulation_time_step_;
```
SteveMacenski, line 148: I think this is eq to a docking goal tolerance, which I'm not sure we have in this situation. I would think we want to forward simulate the full trajectory to the theoretical completion, but I can understand the issue that you're bringing up. Wouldn't the `max_iter` solution resolve that for the most part (+/- 1-2 extra checks, but very limited)

ajtudela: Mmm, I think I understand your concern. My first intention was to use the `undock_linear_tolerance` parameter or something similar, but I didn't want to change the function signature or add a new parameter.

Another option that just occurred to me is when we undock, check for collision at the staging pose after the do loop in L188.

SteveMacenski: > undock_linear_tolerance

But this is also used in the docking case as well, not just undocking :-)

ajtudela: In the docking case it doesn't matter because:
* The target pose it's the backward projected pose (dock_pose +0.25) so the dock pose is always checked when `dock_collision_threshold_` = 0.
* Besides that, if `dock_collision_threshold_` > 0 we always skip the last portion.

SteveMacenski: > The target pose it's the backward projected pose (dock_pose +0.25)

But the simulation time and max speed are arbitrary user-defined settings, so that tolerance could possibly be quite high if they select a very coarse simulation time resolution or move at high speeds. While those are likely misconfigurations, they happen 

I don't think what you have here is generally incorrect, but I'm not clearly seeing what it adds in addition to the `max_iter` termination condition. Worst case, we check an extra 2 poses which is pretty inconsequential given the docking server's compute and frequency

mikeferguson: Would it make sense to derive a `resolution` based on the underlying costmap resolution? There comes a point where additional simulations don't get you any different value because you continue to collision check against the same underlying costmap cells

ajtudela: > I don't think what you have here is generally incorrect, but I'm not clearly seeing what it adds in addition to the `max_iter` termination condition. Worst case, we check an extra 2 poses which is pretty inconsequential given the docking server's compute and frequency. So I suppose if it doesn't add alot of benefit (_unless I'm missing something_), it adds some potential for misconfiguration that could be not beneficial

Honestly, it doesn't really add any benefit to docking/undocking. In fact, I noticed this when I used this controller for the following server and saw these extra poses.

> Would it make sense to derive a `resolution` based on the underlying costmap resolution? There comes a point where additional simulations don't get you any different value because you continue to collision check against the same underlying costmap cells

This was also my intention, but since the trajectory is generated for debugging/visualisation purposes without the need to check for collisions, there is no access to the costmap when collision avoidance is disabled.

SteveMacenski: > Would it make sense to derive a resolution based on the underlying costmap resolution? 

I would be perfectly happy with that for the case that we have a costmap. However, for cases where no collision checking is requested, its very possible that no costmap is given at all over the topic or in the broader system 

My suggestion would be to remove this term in general -- this PR already fixes the major problem. I think this is a fine-tuning thing for the last 1-2 poses of what is overall a very light weight process. If this was a global path planner or sampling hundreds of possible trajectories, I'd be more cautious about extra checks, but checking ~100 cells is easy enough, what's +2 more?

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -177,7 +180,10 @@ bool Controller::isTrajectoryCollisionFree(
       trajectory_pub_->publish(trajectory);
       return false;
     }
-  }
+
+    // Check if we reach the goal
+    distance = nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose);
+  }while(distance > 1e-2 && trajectory.poses.size() < max_iter);
```
SteveMacenski, line 186:
```suggestion
  }while(distance > 1e-2 || trajectory.poses.size() < max_iter);
```

Isn't this more correct? We don't need both, but one or the other

SteveMacenski (changes requested):

SteveMacenski (approved):

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -177,7 +180,10 @@ bool Controller::isTrajectoryCollisionFree(
       trajectory_pub_->publish(trajectory);
       return false;
     }
-  }
+
+    // Check if we reach the goal
+    distance = nav2_util::geometry_utils::euclidean_distance(target_pose, next_pose.pose);
+  }while(distance > 1e-2 || trajectory.poses.size() < max_iter);
```
SteveMacenski, line 186:
```suggestion
  }while(distance > 1e-2 && trajectory.poses.size() < max_iter);
```

SteveMacenski: I'm an idiot this morning - you were right with &&. We want to do it while both are true. If any is false, we want to exit :facepal

ajtudela: Yeah, I was writing the truth table and thinking myself "ehh, it was right the first time with the && ...". We need holidays 🤣

PR #4810, opened 2024-12-20
Add acceleration limits to DriveOnHeading and BackUp behaviors

RBT22:
---

## Basic Info

| ------ | ----------- |

---

 Added two params: `acceleration_limit` and `deceleration_limit` for DriveOnHeading and BackUp behaviors.

Added new parameter, so need to add that to default configs and documentation page

---

#### For Maintainers: 

### DIFF
--- a/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
@@ -134,19 +134,19 @@ public:
     cmd_vel->header.frame_id = this->robot_base_frame_;
     cmd_vel->twist.linear.y = 0.0;
     cmd_vel->twist.angular.z = 0.0;
-    if (acceleration_limit_ <= 0.0 || deceleration_limit_ <= 0.0) {
+    if (acceleration_limit_ <= 0.0 || deceleration_limit_ >= 0.0) {
       cmd_vel->twist.linear.x = command_speed_;
     } else {
       double current_speed = last_vel_ == std::numeric_limits<double>::max() ? 0.0 : last_vel_;
 
-      double min_feasible_speed = current_speed - deceleration_limit_ / this->cycle_frequency_;
+      double min_feasible_speed = current_speed + deceleration_limit_ / this->cycle_frequency_;
       double max_feasible_speed = current_speed + acceleration_limit_ / this->cycle_frequency_;
       cmd_vel->twist.linear.x = std::clamp(command_speed_, min_feasible_speed, max_feasible_speed);
 
       // Check if we need to slow down to avoid overshooting
       auto remaining_distance = std::fabs(command_x_) - distance;
       if (command_speed_ > 0.0) {
-        double max_vel_to_stop = std::sqrt(2.0 * deceleration_limit_ * remaining_distance);
+        double max_vel_to_stop = std::sqrt(-2.0 * deceleration_limit_ * remaining_distance);
         if (max_vel_to_stop < cmd_vel->twist.linear.x) {
           cmd_vel->twist.linear.x = max_vel_to_stop;
         }
@@ -250,10 +250,10 @@ protected:
 
     nav2_util::declare_parameter_if_not_declared(
       node, this->behavior_name_ + ".acceleration_limit",
-      rclcpp::ParameterValue(0.0));
+      rclcpp::ParameterValue(2.5));
     nav2_util::declare_parameter_if_not_declared(
       node, this->behavior_name_ + ".deceleration_limit",
-      rclcpp::ParameterValue(0.0));
+      rclcpp::ParameterValue(-2.5));
     node->get_parameter(this->behavior_name_ + ".acceleration_limit", acceleration_limit_);
     node->get_parameter(this->behavior_name_ + ".deceleration_limit", deceleration_limit_);
   }
--- a/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
@@ -266,6 +266,11 @@ protected:
     node->get_parameter(this->behavior_name_ + ".acceleration_limit", acceleration_limit_);
     node->get_parameter(this->behavior_name_ + ".deceleration_limit", deceleration_limit_);
     node->get_parameter(this->behavior_name_ + ".minimum_speed", minimum_speed_);
+    if (acceleration_limit_ < 0.0 || deceleration_limit_ > 0.0) {
+      RCLCPP_ERROR(this->logger_, "DriveOnHeading: acceleration_limit and deceleration_limit must be "
+        "positive and negative respectively");
+      throw std::runtime_error{"Invalid parameter: acceleration_limit or deceleration_limit"};
+    }
   }
 
   typename ActionT::Feedback::SharedPtr feedback_;
--- a/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
@@ -134,35 +134,30 @@ public:
     cmd_vel->header.frame_id = this->robot_base_frame_;
     cmd_vel->twist.linear.y = 0.0;
     cmd_vel->twist.angular.z = 0.0;
-    if (acceleration_limit_ <= 0.0 || deceleration_limit_ >= 0.0) {
-      RCLCPP_INFO_ONCE(this->logger_, "DriveOnHeading: no acceleration or deceleration limits set");
-      cmd_vel->twist.linear.x = command_speed_;
+
+    double current_speed = last_vel_ == std::numeric_limits<double>::max() ? 0.0 : last_vel_;
+    double min_feasible_speed = current_speed + deceleration_limit_ / this->cycle_frequency_;
+    double max_feasible_speed = current_speed + acceleration_limit_ / this->cycle_frequency_;
+    cmd_vel->twist.linear.x = std::clamp(command_speed_, min_feasible_speed, max_feasible_speed);
+
+    // Check if we need to slow down to avoid overshooting
+    auto remaining_distance = std::fabs(command_x_) - distance;
+    bool forward = command_speed_ > 0.0;
+    if (forward) {
+      double max_vel_to_stop = std::sqrt(-2.0 * deceleration_limit_ * remaining_distance);
+      if (max_vel_to_stop < cmd_vel->twist.linear.x) {
+        cmd_vel->twist.linear.x = max_vel_to_stop;
+      }
     } else {
-      double current_speed = last_vel_ == std::numeric_limits<double>::max() ? 0.0 : last_vel_;
-
-      double min_feasible_speed = current_speed + deceleration_limit_ / this->cycle_frequency_;
-      double max_feasible_speed = current_speed + acceleration_limit_ / this->cycle_frequency_;
-      cmd_vel->twist.linear.x = std::clamp(command_speed_, min_feasible_speed, max_feasible_speed);
-
-      // Check if we need to slow down to avoid overshooting
-      auto remaining_distance = std::fabs(command_x_) - distance;
-      bool forward = command_speed_ > 0.0;
-      if (forward) {
-        double max_vel_to_stop = std::sqrt(-2.0 * deceleration_limit_ * remaining_distance);
-        if (max_vel_to_stop < cmd_vel->twist.linear.x) {
-          cmd_vel->twist.linear.x = max_vel_to_stop;
-        }
-      } else {
-        double max_vel_to_stop = -std::sqrt(2.0 * acceleration_limit_ * remaining_distance);
-        if (max_vel_to_stop > cmd_vel->twist.linear.x) {
-          cmd_vel->twist.linear.x = max_vel_to_stop;
-        }
+      double max_vel_to_stop = -std::sqrt(2.0 * acceleration_limit_ * remaining_distance);
+      if (max_vel_to_stop > cmd_vel->twist.linear.x) {
+        cmd_vel->twist.linear.x = max_vel_to_stop;
       }
+    }
 
-      // Ensure we don't go below minimum speed
-      if (std::fabs(cmd_vel->twist.linear.x) < minimum_speed_) {
-        cmd_vel->twist.linear.x = forward ? minimum_speed_ : -minimum_speed_;
-      }
+    // Ensure we don't go below minimum speed
+    if (std::fabs(cmd_vel->twist.linear.x) < minimum_speed_) {
+      cmd_vel->twist.linear.x = forward ? minimum_speed_ : -minimum_speed_;
     }
 
     geometry_msgs::msg::Pose2D pose2d;
@@ -267,7 +262,8 @@ protected:
     node->get_parameter(this->behavior_name_ + ".deceleration_limit", deceleration_limit_);
     node->get_parameter(this->behavior_name_ + ".minimum_speed", minimum_speed_);
     if (acceleration_limit_ < 0.0 || deceleration_limit_ > 0.0) {
-      RCLCPP_ERROR(this->logger_, "DriveOnHeading: acceleration_limit and deceleration_limit must be "
+      RCLCPP_ERROR(this->logger_,
+        "DriveOnHeading: acceleration_limit and deceleration_limit must be "
         "positive and negative respectively");
       throw std::runtime_error{"Invalid parameter: acceleration_limit or deceleration_limit"};
     }
--- a/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
+++ b/nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
@@ -19,6 +19,7 @@
 #include <chrono>
 #include <memory>
 #include <utility>
+#include <limits>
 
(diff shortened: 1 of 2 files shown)


nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -146,6 +175,8 @@ class DriveOnHeading : public TimedBehavior<ActionT>
       return ResultStatus{Status::FAILED, ActionT::Result::COLLISION_AHEAD};
     }
 
+    last_vel_ = std::make_shared<double>(cmd_vel->twist.linear.x);
```
SteveMacenski, line 178: This is weird. If you want to initialize it so you know its valid, you could default it to `numerical_limits<double>::max()` and reset to that / check for it if we have a last value. I don't like this smart pointer stuff though

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -133,7 +134,35 @@ class DriveOnHeading : public TimedBehavior<ActionT>
     cmd_vel->header.frame_id = this->robot_base_frame_;
     cmd_vel->twist.linear.y = 0.0;
     cmd_vel->twist.angular.z = 0.0;
-    cmd_vel->twist.linear.x = command_speed_;
+    if (acceleration_limit_ <= 0.0 && deceleration_limit_ <= 0.0) {
+      cmd_vel->twist.linear.x = command_speed_;
+    } else {
+      double current_speed = last_vel_ ? *last_vel_ : 0.0;
```
SteveMacenski, line 140: I think the behaviors have a odom subscriber to use for the current velocity. 

This actually brings up a whole topic about whether we should base the velocity closed loop (on odom speed) or open looop (on the last command sent, assuming it was executed). We have this also topic in the controllers and velocity smoother. If we're adding in dynamic constraints here, I think its sensible we decide what makes sense here as well

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -133,7 +134,35 @@ class DriveOnHeading : public TimedBehavior<ActionT>
     cmd_vel->header.frame_id = this->robot_base_frame_;
     cmd_vel->twist.linear.y = 0.0;
     cmd_vel->twist.angular.z = 0.0;
-    cmd_vel->twist.linear.x = command_speed_;
+    if (acceleration_limit_ <= 0.0 && deceleration_limit_ <= 0.0) {
```
SteveMacenski, line 137: `||`? I assume we have to set both if we want to use either. That would let you remove the 2x `if` in the other branch as well. ALso we can simply if the `forward &&` to just `forward` and `!forward`

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -133,7 +134,35 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+      if (deceleration_limit_ > 0.0) {
+        min_feasible_speed = current_speed - deceleration_limit_ / this->cycle_frequency_;
+      }
+      if (acceleration_limit_ > 0.0) {
+        max_feasible_speed = current_speed + acceleration_limit_ / this->cycle_frequency_;
+      }
+      cmd_vel->twist.linear.x = std::clamp(command_speed_, min_feasible_speed, max_feasible_speed);
+
+      // Check if we need to slow down to avoid overshooting
+      bool forward = command_speed_ > 0.0 ? true : false;
```
SteveMacenski, line 153:
```suggestion
      bool forward = command_speed_ > 0.0;
```

:wink:

SteveMacenski (commented):
Build is failing - not 100% sure why off hand other than it relates to a method added

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
RBT22: I couldn’t find the general odom subscriber —am I missing something? Should I add one just for this node, or would it make more sense to handle it across all the behaviors?

SteveMacenski: Interesting, there's not one in the nav2_behaviors yet, I could have sworn there was one. 

I think that's a very sensible thing to add to the behavior server's scope, then pass to each of the behavior plugins to use if they see fit https://github.com/ros-navigation/navigation2/blob/main/nav2_behaviors/src/behavior_server.cpp#L132-L137

SteveMacenski: But it also depends if we want to do closed-loop control here or open-loop, assuming that velocities sent are velocities executed. Before going too far down that rabbit hole, do you think open-looped control like what you have now (velocity out = velocity executed) is sensible enough for your needs? I think it very well might be considering the nature of it

SteveMacenski:
Sorry for the delayed review - see comment above

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -214,6 +247,15 @@ class DriveOnHeading : public TimedBehavior<ActionT>
       node,
       "simulate_ahead_time", rclcpp::ParameterValue(2.0));
     node->get_parameter("simulate_ahead_time", simulate_ahead_time_);
+
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".acceleration_limit",
+      rclcpp::ParameterValue(0.0));
```
SteveMacenski, line 253: Better defaults please - see the velocity smoother and use its defaults

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -146,6 +169,8 @@ class DriveOnHeading : public TimedBehavior<ActionT>
       return ResultStatus{Status::FAILED, ActionT::Result::COLLISION_AHEAD};
     }
 
+    last_vel_ = cmd_vel->twist.linear.x;
+
```
SteveMacenski, line 173: remove line L173 space

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -133,7 +134,29 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+      double min_feasible_speed = current_speed - deceleration_limit_ / this->cycle_frequency_;
+      double max_feasible_speed = current_speed + acceleration_limit_ / this->cycle_frequency_;
+      cmd_vel->twist.linear.x = std::clamp(command_speed_, min_feasible_speed, max_feasible_speed);
+
+      // Check if we need to slow down to avoid overshooting
+      auto remaining_distance = std::fabs(command_x_) - distance;
+      if (command_speed_ > 0.0) {
+        double max_vel_to_stop = std::sqrt(2.0 * deceleration_limit_ * remaining_distance);
+        if (max_vel_to_stop < cmd_vel->twist.linear.x) {
+          cmd_vel->twist.linear.x = max_vel_to_stop;
```
SteveMacenski, line 151: I think you may also need to set a minimum velocity parameter. Most robots can't go 1mm/s or have stalling velocities.

SteveMacenski: Speaking of this -- an aside. Do you use the rotation shim controller? Right now, we don't do the deceleration as a function of its target orientation and this would be great to add in there too if you happened to also use it & be open to submitting the PR

SteveMacenski (changes requested):

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
RBT22: Actually, this way works fine for us. I originally tried it with closed-loop control but ended up switching to open-loop, and it’s been working well.

SteveMacenski: OK - we can leave it as open loop then :-) 

See other review comments below

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
RBT22: @SteveMacenski I can add the limits for the shim controller after this, but I’m kinda stuck on the tests—can’t get the subscriber to run its callback. Any tips?

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -214,6 +252,19 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".acceleration_limit",
+      rclcpp::ParameterValue(2.5));
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".deceleration_limit",
+      rclcpp::ParameterValue(-2.5));
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".minimum_speed",
+      rclcpp::ParameterValue(0.01));
```
SteveMacenski, line 264:
```suggestion
      rclcpp::ParameterValue(0.10));
```

More realistic

SteveMacenski: Also for the nav2_params.yaml

SteveMacenski (commented):
Also, configuration guide entries are needed and migration guide to mention the new params

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
SteveMacenski: Is cmd_vel the right topic name? It generally looks right ot me, but I'd look at other subscription-based BT node tests for reference :-)

RBT22: Thanks for the feedback! I looked through the repo but couldn’t find any other subscription-based BT node tests—maybe I missed them? I’ve tried the steps you mentioned, but I’m not sure how to check the topic. Could you give me some guidance or point me to an example?

RBT22:
Hi @SteveMacenski,

I’ve been looking into the testing setup, and I believe I might have misunderstood it initially. If I’m interpreting this correctly now, the tests for the behaviors themselves are part of the nav2_system_tests package. Is that correct?

If so, I’ll need to edit the backup_tester.py and drive_tester.py scripts to incorporate the acceleration checks.

Thanks for your guidance!

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -133,7 +134,35 @@ class DriveOnHeading : public TimedBehavior<ActionT>
     cmd_vel->header.frame_id = this->robot_base_frame_;
     cmd_vel->twist.linear.y = 0.0;
     cmd_vel->twist.angular.z = 0.0;
-    cmd_vel->twist.linear.x = command_speed_;
+    if (acceleration_limit_ <= 0.0 || deceleration_limit_ >= 0.0) {
+      cmd_vel->twist.linear.x = command_speed_;
```
SteveMacenski, line 138: Add an `RCLCPP_INFO_ONCE` that accel/decel limits were not set

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -214,6 +252,19 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+      rclcpp::ParameterValue(2.5));
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".deceleration_limit",
+      rclcpp::ParameterValue(-2.5));
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".minimum_speed",
+      rclcpp::ParameterValue(0.10));
+    node->get_parameter(this->behavior_name_ + ".acceleration_limit", acceleration_limit_);
+    node->get_parameter(this->behavior_name_ + ".deceleration_limit", deceleration_limit_);
+    node->get_parameter(this->behavior_name_ + ".minimum_speed", minimum_speed_);
```
SteveMacenski, line 267: Check after this to force the accel/decel to be positive/negative. That will save users some headaches

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -133,7 +134,35 @@ class DriveOnHeading : public TimedBehavior<ActionT>
     cmd_vel->header.frame_id = this->robot_base_frame_;
     cmd_vel->twist.linear.y = 0.0;
     cmd_vel->twist.angular.z = 0.0;
-    cmd_vel->twist.linear.x = command_speed_;
+    if (acceleration_limit_ <= 0.0 || deceleration_limit_ >= 0.0) {
+      cmd_vel->twist.linear.x = command_speed_;
+    } else {
+      double current_speed = last_vel_ == std::numeric_limits<double>::max() ? 0.0 : last_vel_;
+
+      double min_feasible_speed = current_speed + deceleration_limit_ / this->cycle_frequency_;
```
SteveMacenski, line 142: What if the `current_speed` is negative? If you have asymmetric acceleration and deceleration limits (i.e. `-1.0, 2.5`) then the 'accelerating backward' will be deceleration and `decelerating backward` will be acceleration

SteveMacenski: (we have this issue in MPPI too, so its an easy thing to miss https://github.com/ros-navigation/navigation2/issues/4601)

SteveMacenski:
> If I’m interpreting this correctly now, the tests for the behaviors themselves are part of the nav2_system_tests package. Is that correct?

Yes!

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
RBT22: Thanks for pointing that out! I didn’t miss this, though—if you think about it in physical terms, velocity has direction, so when moving backward, the negative deceleration limit can actually reduce the rate at which speed increases. I tried to reflect that in the docs. But if the convention in the repo is to treat the limits without direction, I’m happy to adjust.

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -214,6 +247,25 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+      node, this->behavior_name_ + ".minimum_speed",
+      rclcpp::ParameterValue(0.10));
+    node->get_parameter(this->behavior_name_ + ".acceleration_limit", acceleration_limit_);
+    node->get_parameter(this->behavior_name_ + ".deceleration_limit", deceleration_limit_);
+    node->get_parameter(this->behavior_name_ + ".minimum_speed", minimum_speed_);
+    if (acceleration_limit_ < 0.0 || deceleration_limit_ > 0.0) {
+      RCLCPP_ERROR(this->logger_,
+        "DriveOnHeading: acceleration_limit and deceleration_limit must be "
+        "positive and negative respectively");
+      throw std::runtime_error{"Invalid parameter: acceleration_limit or deceleration_limit"};
```
SteveMacenski, line 267: If we're going to throw, we need to catch here [1] and return false to the lifecycle transition. I think though you can just fix the sign setting silently, but I'm also OK with this way as well 

[1] https://github.com/ros-navigation/navigation2/blob/6e1c85eba70ed9800b69886c15d73268d60f46e9/nav2_behaviors/src/behavior_server.cpp#L98

nav2_behaviors/include/nav2_behaviors/plugins/drive_on_heading.hpp
```diff
@@ -214,6 +247,25 @@ class DriveOnHeading : public TimedBehavior<ActionT>
...
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".deceleration_limit",
+      rclcpp::ParameterValue(-2.5));
+    nav2_util::declare_parameter_if_not_declared(
+      node, this->behavior_name_ + ".minimum_speed",
+      rclcpp::ParameterValue(0.10));
+    node->get_parameter(this->behavior_name_ + ".acceleration_limit", acceleration_limit_);
+    node->get_parameter(this->behavior_name_ + ".deceleration_limit", deceleration_limit_);
+    node->get_parameter(this->behavior_name_ + ".minimum_speed", minimum_speed_);
+    if (acceleration_limit_ < 0.0 || deceleration_limit_ > 0.0) {
```
SteveMacenski, line 263:
```suggestion
    if (acceleration_limit_ <= 0.0 || deceleration_limit_ >= 0.0) {
```

From your recent changes without supporting non-accel/decel setting, they must be non-zero in order to move at all

SteveMacenski (changes requested):
With these + the documentation updates on docs.nav2.org, I can merge this!

SteveMacenski (approved):

SteveMacenski:
@RBT22 thank you alot for jumping in and implementing this!

Issue #4803, opened 2024-12-18
Pass robot positions to cloned_multi_tb3_simulation_launch.py from another launch file


TannerGilbert:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 24.04
- ROS2 Version:
  - ROS2 Rolling
- Version or commit hash:
  - Current main (Last commit: 27ba98a5d576b9d21f0a721fc330016267676195)
- DDS implementation:
  - rmw_cyclonedds_cpp

#### Steps to reproduce issue

Demo launch file for using cloned_multi_tb3_simulation_launch.py:

```python
import os

from ament_index_python.packages import get_package_share_directory
from launch import LaunchDescription
from launch.actions import IncludeLaunchDescription
from launch.launch_description_sources import PythonLaunchDescriptionSource

def generate_launch_description():
    return LaunchDescription(
        [
            IncludeLaunchDescription(
                PythonLaunchDescriptionSource(
                    os.path.join(get_package_share_directory('nav2_bringup'), "launch", "cloned_multi_tb3_simulation_launch.py")
                ),
                launch_arguments={"robots": "robot1={x: 1.0, y: 1.0, yaw: 1.5707}; robot2={x: 1.0, y: 1.0, yaw: 1.5707}"}.items(),
            )
        ]
    )
```

#### Expected behavior

Robots are defined based on passed launch argument.

#### Actual behavior

The cloned_multi_tb3_simulation_launch.py file currently uses ParseMultiRobotPose which  reads from the command line. This doesn't seem to work with the common ROS2 `launch_argument` notation when calling from another launch file.

A simple workaround would be changing the top level launch file to use `ExecuteProcess` instead of `IncludeLaunchDescription`:
Workaround:
```python
from launch import LaunchDescription
from launch.actions import ExecuteProcess

def generate_launch_description():
    return LaunchDescription(
        [
            ExecuteProcess(
                cmd=['ros2', 'launch', 'nav2_bringup', 'cloned_multi_tb3_simulation_launch.py', "robots:=robot1={x: 0.5, y: 0.5, yaw: 1.5707};", "headless:=False"],
                output='screen',
            ),
        ]
    )
```

My question is should the `cloned_multi_tb3_simulation_launch.py` and `ParseMultiRobotPose` be refactored to allow for the common ROS 2 launch syntax or should it be kept as is?

SteveMacenski:
[...]

PR #4811, opened 2024-12-20
fix: Changed cloned_multi_tb3_simulation_launch.py file to conform with common ROS 2 launch syntax

TannerGilbert:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4803 |
| Primary OS tested on | Ubuntu 24.04, ROS Rolling |
| Robotic platform tested on | Turtlebot3 |

---

- Changed cloned_multi_tb3_simulation_launch.py file to conform with common ROS 2 launch syntax and thereby allow for including it in other launch files as expected.
- Made robots a required launch argument

None

---

I would prefer that the headless and SLAM options are added to this script similar to how it's handled in the [tb3_simulation_launch.py file](https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/launch/tb3_simulation_launch.py)

#### For Maintainers: 


SteveMacenski (changes requested):
I'm not sure this is the best way to handle the problem. I think what makes more sense is to have the parsing `nav2_common` utility be a proper launch substitution https://github.com/ros2/launch_ros/tree/rolling/launch_ros/launch_ros/substitutions. I think that would make this work better and be in the intended launch file's declarative style (rather than trying to get around it using opaque functions). If we're going to try to circumnavigate declarative programming, I think we might be better off using the current styling instead since its more clear to new developers and is equally as circumnavigated

TannerGilbert:
> I'm not sure this is the best way to handle the problem. I think what makes more sense is to have the parsing `nav2_common` utility be a proper launch substitution https://github.com/ros2/launch_ros/tree/rolling/launch_ros/launch_ros/substitutions. I think that would make this work better and be in the intended launch file's declarative style (rather than trying to get around it using opaque functions). If we're going to try to circumnavigate declarative programming, I think we might be better off using the current styling instead since its more clear to new developers and is equally as circumnavigated

I'll have a look at this. Thanks for the feedback

TannerGilbert:
@SteveMacenski sorry for the delay. I had a look at how to write a custom substitution class and tried to refactor the ParseMultiRobotPose accordingly. Still, as a dict is needed in the launchfile to the bringup_cmd_group list I wasn't fully able to remove the OpaqueFunction. Maybe my ROS knowledge is insufficient. I would be glad for any feedback and sorry for the inconvenience.

SteveMacenski:
OK - we can use the opaque function method!

nav2_bringup/launch/cloned_multi_tb3_simulation_launch.py
```diff
@@ -20,23 +20,87 @@
 
 from ament_index_python.packages import get_package_share_directory
 from launch import LaunchDescription
-from launch.actions import (
```
SteveMacenski, line 23: Please revert changes, only change things that you need to within the context of your PR please :-)

nav2_bringup/launch/cloned_multi_tb3_simulation_launch.py
```diff
@@ -20,23 +20,87 @@
...
+
+    # Define commands for launching the navigation instances
+    bringup_cmd_group = []
+    for robot_name in robots_list:
+        init_pose = robots_list[robot_name]
+        group = GroupAction(
+            [
+                LogInfo(
+                    msg=[
+                        'Launching namespace=',
```
SteveMacenski, line 55: This seems like unnecessary line extension - can we collapse this a bit?

nav2_bringup/launch/cloned_multi_tb3_simulation_launch.py
```diff
@@ -200,13 +216,13 @@ def generate_launch_description():
     ld.add_action(start_gazebo_cmd)
     ld.add_action(remove_temp_sdf_file)
 
-    ld.add_action(LogInfo(msg=['number_of_robots=', str(len(robots_list))]))
-
     ld.add_action(
-        LogInfo(condition=IfCondition(log_settings), msg=['map yaml: ', map_yaml_file])
+        LogInfo(condition=IfCondition(log_settings),
```
SteveMacenski, line 220: Ditto here and below

nav2_common/nav2_common/launch/__init__.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from .has_node_params import HasNodeParams
-from .parse_multirobot_pose import ParseMultiRobotPose
```
SteveMacenski, line 16: Why remove?

SteveMacenski (changes requested):

SteveMacenski:
Please rebase / pull in main and that should make CI turn over. TF2 had some API changes we fixed over the weekend

TannerGilbert:
I updated the PR based on the change request. Thanks for all the feedback and patience.

nav2_common/nav2_common/launch/parse_multirobot_pose.py
```diff
@@ -12,66 +12,49 @@
...
 
-    def __init__(self, target_argument: Text):
-        """
-        Parse arguments for multi-robot's pose.
+    Expects input in the format:
+    robots:="robot1={x: 1.0, y: 1.0, yaw: 0.0}; robot2={x: 2.0, y: 2.0, yaw: 1.5707}"
+    """
 
-        for example,
-        `ros2 launch nav2_bringup bringup_multirobot_launch.py
```
SteveMacenski, line 29: Please add back in the instruction that shows example of XY/Yaw and XYZ/RPY so that folks can see how to set both (and that they _could_ set both). That was an intentional choice by the authors

nav2_bringup/launch/cloned_multi_tb3_simulation_launch.py
```diff
@@ -71,6 +132,12 @@ def generate_launch_description():
         description='Full path to world file to load',
     )
 
+    declare_robots_cmd = DeclareLaunchArgument(
+        'robots',
+        default_value='robot1={x: 0.5, y: 0.5, yaw: 0}; robot2={x: -0.5, y: -0.5, yaw: 1.5707}',
```
SteveMacenski, line 137: Ditto here

SteveMacenski (approved):
Very minor changes

Last thing is to add this to our migration guide so people know about these changes! https://docs.nav2.org/migration/Jazzy.html

SteveMacenski (approved):

SteveMacenski:
@TannerGilbert just sanity check me that you tested that the default works now after the updates & that it works when you specify some other robot parameters? We don't have system tests that run the multirobot case, so I just want to have that verified explicitly before I merge.

TannerGilbert:
> @TannerGilbert just sanity check me that you tested that the default works now after the updates & that it works when you specify some other robot parameters? We don't have system tests that run the multirobot case, so I just want to have that verified explicitly before I merge.

Yes I checked after doing the changes that the behaviour didn't change. For a bit more detail when running `ros2 launch nav2_bringup cloned_multi_tb3_simulation_launch.py` one should get the two robots in the default position and in the command line there will be the following output:
```
Launching namespace=robot1 init_pose={'x': 0.5, 'y': 0.5, 'yaw': 0, 'z': 0.0, 'roll': 0.0, 'pitch': 0.0}
Launching namespace=robot2 init_pose={'x': -0.5, 'y': -0.5, 'z': 0, 'roll': 0, 'pitch': 0, 'yaw': 1.5707}
```

When supplying a custom position like `ros2 launch nav2_bringup cloned_multi_tb3_simulation_launch.py robots:="robot1={x: 1.0, y: 1.0, yaw: 1.5707}"` the output changes accordingly.

```
Launching namespace=robot1 init_pose={'x': 1.0, 'y': 1.0, 'yaw': 1.5707, 'z': 0.0, 'roll': 0.0, 'pitch': 0.0}
```

SteveMacenski:
Thanks for the efforts and getting this to  completion! I pride myself that Nav2 is generally a 'best practices' ROS 2 codebase as an example, so its really good to fix this up to be in line with the best practices. Thanks for your time!

Issue #4809, opened 2024-12-19
MPPI GoalCritic bug


redvinaa:
## Bug report

**Required Info:**

- Operating System:
  - 
- ROS2 Version:
  - 
- Version or commit hash:
  - main (16abf9b4) 
- DDS implementation:
  - 

#### Steps to reproduce issue

Publish critic scores and see when each of the critics is active. I've made a branch to do this, but I only have a dirty and forked branch on humble: https://github.com/EnjoyRobotics/navigation2/tree/humble-enjoy-mppi-critic-pub. (I also checked, and this is not fixed on main as of now.) Otherwise it can probably be easily checked by putting some prints in the critics code. In terms of the parameters, `threshold_to_consider` should be ~1-1.5 (assuming local costmap size is 3m), and everything else can stay default.

#### Expected behavior
MPPI GoalCritic should activate only when close to *global* goal.
#### Actual behavior
All critics have access only to the CriticData. In the GoalCritic (specifically the `utils::withinPositionGoalTolerance`), the goal is calculated from the last point of the `CriticData.path`. But this path is already the pruned one (pruning is by integrated distance), and if the path is very wavy, its end can end up close enough to the robot pose to trigger the GoalCritic's `theshold_to_consider` (and also of others', for example of PathFollow). I propose adding the *global* goal pose to the CriticData.

SteveMacenski:
Sure thing! I support that change :-) Can you submit a PR against `main` and `humble`? I can backport to Jazzy from main once merged

PR #4812, opened 2024-12-20
Add goal pose to CriticData

redvinaa:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #4809 ) |
| Robotic platform tested on | custom gazebo simulation, custom hardware |

---

* Added goal pose to MPPI CriticData
* Now all critics can activate based on distance from *global* goal pose instead of pruned

* Bugfix, usage doesn't change

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/controller.cpp
+++ b/nav2_mppi_controller/src/controller.cpp
@@ -92,13 +92,15 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   last_time_called_ = clock_->now();
 
   std::lock_guard<std::mutex> param_lock(*parameters_handler_->getLock());
+  geometry_msgs::msg::Pose goal = path_handler_.getTransformedGoal().pose;
+
   nav_msgs::msg::Path transformed_plan = path_handler_.transformPath(robot_pose);
 
   nav2_costmap_2d::Costmap2D * costmap = costmap_ros_->getCostmap();
   std::unique_lock<nav2_costmap_2d::Costmap2D::mutex_t> costmap_lock(*(costmap->getMutex()));
 
   geometry_msgs::msg::TwistStamped cmd =
-    optimizer_.evalControl(robot_pose, robot_speed, transformed_plan, goal_checker);
+    optimizer_.evalControl(robot_pose, robot_speed, transformed_plan, goal, goal_checker);
 
 #ifdef BENCHMARK_TESTING
   auto end = std::chrono::system_clock::now();
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -119,7 +119,7 @@ void CostCritic::score(CriticData & data)
 
   // If near the goal, don't apply the preferential term since the goal is near obstacles
   bool near_goal = false;
-  if (utils::withinPositionGoalTolerance(near_goal_distance_, data.state.pose.pose, data.path)) {
+  if (utils::withinPositionGoalTolerance(near_goal_distance_, data)) {
     near_goal = true;
   }
 
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -119,7 +119,7 @@ void CostCritic::score(CriticData & data)
 
   // If near the goal, don't apply the preferential term since the goal is near obstacles
   bool near_goal = false;
-  if (utils::withinPositionGoalTolerance(near_goal_distance_, data)) {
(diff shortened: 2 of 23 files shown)


redvinaa:
Please check implementation and when everything is correct, I'll port to rolling

nav2_mppi_controller/src/controller.cpp
```diff
@@ -92,13 +92,15 @@ geometry_msgs::msg::TwistStamped MPPIController::computeVelocityCommands(
   last_time_called_ = clock_->now();
 
   std::lock_guard<std::mutex> param_lock(*parameters_handler_->getLock());
+  geometry_msgs::msg::Pose goal = path_handler_.getTransformedGoal().pose;
```
SteveMacenski, line 95: Shouldn't we get the goal as `transformed_plan.poses.back()`? There's no reason to do another TF operation I don't think and we can find it on L98 below `transformPath()` instead

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -119,7 +119,7 @@ void CostCritic::score(CriticData & data)
 
   // If near the goal, don't apply the preferential term since the goal is near obstacles
   bool near_goal = false;
-  if (utils::withinPositionGoalTolerance(near_goal_distance_, data.state.pose.pose, data.path)) {
```
SteveMacenski, line 122: I much prefer this API so that its explicit about what information that `withinPositionGoalTolerance` is working on instead of handing it everything and then having to reference the implementation to know what data its using. I think you can just update the existing `withinPositionGoalTolerance()` method instead - it seems like a pretty trivial change

SteveMacenski (changes requested):
Generally looks good to me

nav2_mppi_controller/src/controller.cpp
redvinaa: We can't because `transformed_plan` is already pruned.

nav2_mppi_controller/src/critics/cost_critic.cpp
redvinaa: Done!

redvinaa:
Can I have some help with the CI errors? I'm not sure it's because of my changes

SteveMacenski:
@redvinaa the CI errors are because we don't expect PRs directly to existing distributions so its not setup correctly with the caching. You could have build / test issues, but we'd need to to have this opened against `main` to check for them. This PR is opened in the opposite direction we ususally do (main branch -> then backport to humble, if possible)

nav2_mppi_controller/src/path_handler.cpp
```diff
@@ -186,6 +186,20 @@ void PathHandler::prunePlan(nav_msgs::msg::Path & plan, const PathIterator end)
   plan.poses.erase(plan.poses.begin(), end);
 }
 
+geometry_msgs::msg::PoseStamped PathHandler::getTransformedGoal()
+{
+  auto goal = global_plan_.poses.back();
+  goal.header.stamp = rclcpp::Time(0);
+  if (goal.header.frame_id.empty()) {
+    throw std::runtime_error("Goal pose has an empty frame_id");
```
SteveMacenski, line 194: For the `main` version of this PR, make sure to throw https://github.com/ros-navigation/navigation2/blob/main/nav2_core/include/nav2_core/controller_exceptions.hpp#L37-L42 these exceptions insteaad

nav2_mppi_controller/src/path_handler.cpp
```diff
@@ -186,6 +186,20 @@ void PathHandler::prunePlan(nav_msgs::msg::Path & plan, const PathIterator end)
   plan.poses.erase(plan.poses.begin(), end);
 }
 
+geometry_msgs::msg::PoseStamped PathHandler::getTransformedGoal()
```
SteveMacenski, line 189: Pass in the stamp (robot_pose.header.stamp) to use instead of `Time(0)` so that things are synchronized a bit better. If you look at other TF calls in this file, they all reference a meaningful timestamp, not just latest

SteveMacenski (changes requested):
Otherwise LGTM

nav2_mppi_controller/src/path_handler.cpp
redvinaa: Maybe global_plan_.header.stamp is even better no? Because it's related to the path.

SteveMacenski: Sure! Whatever stamp makes the most sense that's used in other places :-)

SteveMacenski (approved):

SteveMacenski (approved):

Issue #4710, opened 2024-10-06
Simple Smoother Aborts Navigation with "No Segments Smoothed" Error When Approaching End of Path


rayferric:
## Bug report

**Required Info:**

- **Operating System**:
  - Ubuntu 24.04.1 LTS
- **ROS2 Version**:
  - Jazzy
- **Version or commit hash**:
  - Jazzy Binaries from 6 Oct 2024 with [this PR](https://github.com/ros-navigation/navigation2/pull/3296)
- **DDS implementation**:
  - Cyclone DDS

#### Steps to reproduce issue

1. Use the following `nav2` parameters:

   ```yaml
   smoother_server:
   ros__parameters:
     costmap_topic: global_costmap/costmap_raw
     footprint_topic: global_costmap/published_footprint
     robot_base_frame: base_footprint
     transform_timeout: 0.1
     smoother_plugins: ["simple_smoother"]
     simple_smoother:
       max_its: 1000
       do_refinement: True
       w_smooth: 0.3
       w_data: 0.2
       tolerance: 1e-10
   ```

2. Set up the behavior tree as follows:

   ```xml
   <Sequence name="ComputeAndSmoothPath">
     <ComputePathToPose goal="{goal}" path="{path}" planner_id="GridBased"/>
     <SmoothPath unsmoothed_path="{path}" smoothed_path="{smooth_path}"/>
   </Sequence>
   ```

3. Execute navigation. The robot will traverse the path and then start approaching the end of it.

#### Expected behavior

The robot smoothly navigates the entire path without aborting, and `SmoothPath` completes as expected.

#### Actual behavior

When the robot approaches the end of the path, the `simple_smoother` aborts the navigation with the following error:

```cpp
if (segments_smoothed == 0) {
  throw nav2_core::FailedToSmoothPath("No segments were smoothed");
}
```

#### Additional information

Logs:
```log
[smoother_server]: Received a path to smooth.
[smoother_server]: No segments were smoothed
[smoother_server]: [smooth_path] [ActionServer] Aborting handle.
[global_costmap.global_costmap]: Received request to clear entirely the global_costmap
[local_costmap.local_costmap]: Received request to clear entirely the local_costmap
```
...repeating endlessly

The exception is thrown even though the smoother could simply return a jagged path. The above exception leads to the whole navigation process being aborted. On the internet, there doesn't seem to be available much information on how to properly setup the SmoothPath node or handle BehaviorTree errors.

[...]

PR #4817, opened 2024-12-24
fix simple smoother failing during final approach

rayferric:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4710 |
| Primary OS tested on | Ubuntu (osrf/ros:jazzy-desktop-full) |
| Robotic platform tested on | my own custom simulated robot  |

---

* modified smoother server's simple_smoother plugin to not throw FailedToSmoothPath when the robot is approaching the end of path

I can alter this behavior if requested. My only requirement for this patch is that no error will be thrown when the path is ending and the last and only segment is too short to be processed, causing `segments_smoothed` to be zero.

#### For Maintainers: 

### DIFF
--- a/nav2_smoother/src/simple_smoother.cpp
+++ b/nav2_smoother/src/simple_smoother.cpp
@@ -65,8 +65,7 @@ bool SimpleSmoother::smooth(
   steady_clock::time_point start = steady_clock::now();
   double time_remaining = max_time.seconds();
 
-  bool success = true, reversing_segment;
-  unsigned int segments_smoothed = 0;
+  bool reversing_segment;
   nav_msgs::msg::Path curr_path_segment;
   curr_path_segment.header = path.header;
 
@@ -88,15 +87,9 @@ bool SimpleSmoother::smooth(
       time_remaining = max_time.seconds() - duration_cast<duration<double>>(now - start).count();
       refinement_ctr_ = 0;
 
-      bool segment_was_smoothed = smoothImpl(
-        curr_path_segment, reversing_segment, costmap.get(), time_remaining);
-
-      if (segment_was_smoothed) {
-        segments_smoothed++;
-      }
-
-      // Smooth path segment naively
-      success = success && segment_was_smoothed;
+      // Attempt to smooth the segment
+      // May throw SmootherTimedOut
+      smoothImpl(curr_path_segment, reversing_segment, costmap.get(), time_remaining);
 
       // Assemble the path changes to the main path
       std::copy(
@@ -106,14 +99,10 @@ bool SimpleSmoother::smooth(
     }
   }
 
-  if (segments_smoothed == 0) {
-    throw nav2_core::FailedToSmoothPath("No segments were smoothed");
-  }
-
-  return success;
+  return true;
 }
 
-bool SimpleSmoother::smoothImpl(
+void SimpleSmoother::smoothImpl(
   nav_msgs::msg::Path & path,
   bool & reversing_segment,
   const nav2_costmap_2d::Costmap2D * costmap,
@@ -142,7 +131,7 @@ bool SimpleSmoother::smoothImpl(
         "Number of iterations has exceeded limit of %i.", max_its_);
       path = last_path;
       updateApproximatePathOrientations(path, reversing_segment);
-      return false;
+      return;
     }
 
     // Make sure still have time left to process
@@ -188,7 +177,7 @@ bool SimpleSmoother::smoothImpl(
           "Returning the last path before the infeasibility was introduced.");
         path = last_path;
         updateApproximatePathOrientations(path, reversing_segment);
-        return false;
+        return;
       }
     }
 
@@ -204,7 +193,6 @@ bool SimpleSmoother::smoothImpl(
 
   updateApproximatePathOrientations(new_path, reversing_segment);
   path = new_path;
-  return true;
 }
 
 double SimpleSmoother::getFieldByDim(
--- a/nav2_smoother/include/nav2_smoother/simple_smoother.hpp
+++ b/nav2_smoother/include/nav2_smoother/simple_smoother.hpp
@@ -92,9 +92,8 @@ protected:
    * @param reversing_segment Return if this is a reversing segment
    * @param costmap Pointer to minimal costmap
    * @param max_time Maximum time to compute, stop early if over limit
-   * @return If smoothing was successful
    */
-  bool smoothImpl(
+  void smoothImpl(
     nav_msgs::msg::Path & path,
     bool & reversing_segment,
     const nav2_costmap_2d::Costmap2D * costmap,
--- a/nav2_smoother/test/test_simple_smoother.cpp
+++ b/nav2_smoother/test/test_simple_smoother.cpp
@@ -246,6 +246,17 @@ TEST(SmootherTest, test_simple_smoother)
   straight_irregular_path.poses[6].pose.orientation = tf2::toMsg(quat2);
   EXPECT_THROW(smoother->smooth(straight_irregular_path, max_time), nav2_core::FailedToSmoothPath);
 
+  // test approach
+  nav_msgs::msg::Path approach_path;
+  approach_path.poses.resize(3);
+  approach_path.poses[0].pose.position.x = 0.5;
+  approach_path.poses[0].pose.position.y = 0.0;
+  approach_path.poses[1].pose.position.x = 0.5;
+  approach_path.poses[1].pose.position.y = 0.1;
+  approach_path.poses[2].pose.position.x = 0.5;
+  approach_path.poses[2].pose.position.y = 0.2;
+  EXPECT_TRUE(smoother->smooth(approach_path, max_time));
+
   // test max iterations
   smoother->setMaxItsToInvalid();
   nav_msgs::msg::Path max_its_path;
--- a/nav2_smoother/test/test_simple_smoother.cpp
+++ b/nav2_smoother/test/test_simple_smoother.cpp
@@ -180,7 +180,7 @@ TEST(SmootherTest, test_simple_smoother)
   EXPECT_NEAR(straight_regular_path.poses[5].pose.position.x, 0.607, 0.01);
   EXPECT_NEAR(straight_regular_path.poses[5].pose.position.y, 0.387, 0.01);
 
-  // Test that collisions are rejected
+  // Test that collisions are disregarded
   nav_msgs::msg::Path collision_path;
   collision_path.poses.resize(11);
   collision_path.poses[0].pose.position.x = 0.0;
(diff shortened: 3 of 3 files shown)


SteveMacenski (approved):
Seems very sensible / a good idea

SteveMacenski:
@rayferric two things:
1. There's a CI test failing from this change. I think you just need to set the size of the path > 1 as an input for an exception that should be thrown
2. Please sign your commits with DCO (see the job failure below)

Then I can merge!

rayferric:
CI failed due to unrelated errors in TF2@rolling:

image

I have pushed an unrelated commit, [use tf2's .hpp includes](https://github.com/ros-navigation/navigation2/pull/4817/commits/f4799fddd231fbdc3f888879986b6b3d49655fb7), in order to see if the CI will go through.

rayferric:
Okay, I've removed the extra commit and rebase-merged the changes from main. Now let's wait until https://github.com/ros-navigation/navigation2/pull/4732 is merged.

SteveMacenski:
Rebase with https://github.com/ros-navigation/navigation2/pull/4823 for CI

nav2_smoother/src/simple_smoother.cpp
```diff
@@ -106,11 +106,17 @@ bool SimpleSmoother::smooth(
     }
   }
 
-  if (segments_smoothed == 0) {
+  if (segments_smoothed == 0 && path_segments.size() > 1) {
     throw nav2_core::FailedToSmoothPath("No segments were smoothed");
   }
 
-  return success;
+  if (!success) {
```
SteveMacenski, line 113: Is this necessary?

rayferric: In order to introduce the least changes necessary, I can remove this modification and modify the tests to use a two-segment path like you suggested.

To explain why those changes were made in the first place, I've noticed that smoother server expects the plugin to return false when the time limit is reached, see:
https://github.com/ros-navigation/navigation2/blob/3763f78d6cf8ed6b4559eb04034891dcf439a41e/nav2_smoother/src/nav2_smoother.cpp#L289-L303

I case of simple smoother, `success` tells whether all calls to `smoothImpl()` returned true:
https://github.com/ros-navigation/navigation2/blob/3763f78d6cf8ed6b4559eb04034891dcf439a41e/nav2_smoother/src/simple_smoother.cpp#L98-L99

And `smoothImpl()` returns true either when either a collision is detected or iteration limit was reached:
https://github.com/ros-navigation/navigation2/blob/3763f78d6cf8ed6b4559eb04034891dcf439a41e/nav2_smoother/src/simple_smoother.cpp#L139-L146
https://github.com/ros-navigation/navigation2/blob/3763f78d6cf8ed6b4559eb04034891dcf439a41e/nav2_smoother/src/simple_smoother.cpp#L184-L192

If time limit is reached, `smoothImpl()` throws an exception, which is out of order since it does not cause smooth() to return false, and instead unwinds the stack:
 https://github.com/ros-navigation/navigation2/blob/3763f78d6cf8ed6b4559eb04034891dcf439a41e/nav2_smoother/src/simple_smoother.cpp#L151-L158
 
My modification would partly reduce this inconsistency by not returning false for reasons unrelated to time limit. This fixes the two failing tests, which expect `FailedToSmoothPath` to be thrown in both cases when 1. collision is found or 2. iteration limit is reached respectively (this for a single-segment path).
 
I could modify the code further to properly handle all cases, by catching the timeout exception in `smooth()` loop and returning false when the exception was caught in any call to `smoothImpl()`.

rayferric:
Here, I rebased with `depreciate_c_headers2`. Now waiting for your decision regarding the test failures

SteveMacenski:
Rebase on main, it was merged in with some other fixes

nav2_smoother/src/simple_smoother.cpp
SteveMacenski: > I could modify the code further to properly handle all return cases, by catching the timeout exception in smooth() loop and returning false when the exception was caught in any call to smoothImpl().

I would prefer that, thank you! These were retrofits on the original software when we added in the contextual error code return types. You've found somewhere we created a gap that would be good to fill! :-)

nav2_smoother/src/simple_smoother.cpp
```diff
@@ -65,8 +65,7 @@ bool SimpleSmoother::smooth(
   steady_clock::time_point start = steady_clock::now();
   double time_remaining = max_time.seconds();
 
-  bool success = true, reversing_segment;
```
SteveMacenski, line 68: Reversing segment is indeed used. The `impl` method has that as a `bool &` which is modified in `updateApproximatePathOrientations`.

nav2_smoother/src/simple_smoother.cpp
```diff
@@ -106,11 +112,14 @@ bool SimpleSmoother::smooth(
...
-  if (segments_smoothed == 0) {
+  // If path is not ending (num_seg > 1), and no segments were smoothed - that's a failure
+  // If some segments were attempted, but none succeeded - also a failure
+  if ((path_segments.size() > 1 || segments_attempted > 0) && segments_smoothed == 0) {
     throw nav2_core::FailedToSmoothPath("No segments were smoothed");
   }
 
-  return success;
+  // no timeout
+  return true;
```
SteveMacenski, line 122: This never checks if `success` like we did before, which is OK, but then we need to compare `segments_attempted == segments_smoothed` to tell if `success` criteria works.

Or, simply leave the success computation and `return success`. Either way

SteveMacenski (changes requested):
I think there are a couple of bugs that need to be addressed, but generally looks good to me once resolved. They're pretty small.

nav2_smoother/src/simple_smoother.cpp
rayferric: `updateApproximatePathOrientations` resets it to false on entry, and the value left there after return is not used outside of this function

nav2_smoother/src/simple_smoother.cpp
rayferric: The server expects `smooth()` to return whether the timeout was reached (and optionally to throw an exception if a fatal failure is encountered):

https://github.com/ros-navigation/navigation2/blob/43d7f7e309e782f1e277191ff414c98b9ac4f51c/nav2_smoother/src/nav2_smoother.cpp#L289-L301

Pure virtual `smooth()`'s docstring says the same thing:

https://github.com/ros-navigation/navigation2/blob/43d7f7e309e782f1e277191ff414c98b9ac4f51c/nav2_core/include/nav2_core/smoother.hpp#L69-L78

SteveMacenski: I think the case of max iterations reached is also breaking a limit where false is returned [1]. Regardless of what the doc strings say, this change introduces a gap. If we smooth everything; we do the right thing. If we smooth nothing; we do the right thing. If we partially smooth, we're not capturing that case in a return or error condition. This is what the previous PR's `success` state measured. If anything, we should update the doc string to be less specific about "time" but rather a "limit" being breached. Time and iteration limits are similar in contextual meaning - we've exceeded application limits  

[1] https://github.com/ros-navigation/navigation2/blob/e1742adc4bf1dd5eb4808f1b29927779de680f62/nav2_smoother/src/simple_smoother.cpp#L149-L154

rayferric:
Hey, in order for us to easily resolve this PR, I've reverted my intrusive changes and fixed the tests so that they pass. Is that alright?

SteveMacenski (approved):

nav2_smoother/src/simple_smoother.cpp
```diff
@@ -106,7 +106,7 @@ bool SimpleSmoother::smooth(
     }
   }
 
-  if (segments_smoothed == 0) {
+  if (segments_smoothed == 0 && path_segments.size() > 1) {
```
SteveMacenski, line 109:
```suggestion
  if (segments_smoothed == 0 && path_segments.size() > 0) {
```

No?

rayferric: Not quite. The original code fixes the basic case of approach by disabling errors when there's just one path segment. If the segment has less than 9 verts, segments_smoothed is 0 and without this condition, the smoothing fails. Though this fix will not work in the case of two segments, if both of them have below 9 vertices, etc.

Suggested fix that would account for the 2+ segment edge case:

I can add back the attempted_segments counter and check if attempted > 1, so that the "no path segments smoothed" would appear whenever at least one segment was long enough to attempt smoothing, but no segments were actually smoothed.

This would also allow me to roll back the changes I made to test cases

SteveMacenski:
I'm going back to the original ticket to remind myself what we were trying to accomplish since we've gone down a few directions. This code has clearly degraded a bit over time and various contributions and refactoring over its lifetime and I want to think about this more abstractly to what we want to do: final path approaches working well and knowing that path segments in various directions can be of arbitrary length.

I think we need to make a few changes here. The returns for `smoothImpl` and when we throw are a bit crazy - as you point out. 

Within `smoothImpl`:
- I think the collision failure case should set `path = last_path`, update the orientations, and return true. If we're mid-execution, then the previous is known to be collision free & that's as good as we can do. If its the first iteration and the path is already in collision, then we can't do much here. If the user cares, then they'd have `goal->check_for_collisions` set in which the finally validated path would then throw the path in collision error [in the main Smoother Server](https://github.com/ros-navigation/navigation2/blob/main/nav2_smoother/src/nav2_smoother.cpp#L306C9-L327)
- For this specific type of smoother, we can return a semi-smoothed path (others if not converged are not going to be valid, like those from the Constrained Smoother). So, for `its >= max_its_` I think we can warn log and return true. 
- For maximum duration, I think we can leave that alone as throwing since that's an actual exit failure

Within `smooth()`:
- Now, all returns from `smoothImpl` are throwing or `return true`, so we don't need to pay much attention to the return bool anymore. I think we can remove the `segments_smoothed` and success logic. And we can change the `smoothImpl` to be a `void` type.
- That would also imply removing the potential to throw `FailedToSmoothPath`, but looking at the git blame for that, it seems haphazardly implemented ([along with the rest of the issues we're removing in this thread](https://github.com/ros-navigation/navigation2/pull/3296))

---

I also reviewed `savitzky_golay_smoother.cpp` to see what changes I wanted to make to be aligned with this to fix other potential messy locations. That one looks good to me and does what I basically just suggested.

rayferric:
Thanks for the review and your well thought out suggestions! Thanks to them, preparing this commit was a breeze. I'm happy that you decided to remove the overly restrictive error handling. As a bonus, I have included a test case for the end-of-path approach scenario, so that my original issue doesn't reoccur.

SteveMacenski (approved):
Sorry that this was such a painful review and iteration process, I promise it'll be better in the future :wink:

PR #4822, opened 2025-01-03
Mppi goal to critic

redvinaa:
Reopen #4812 against main

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/critic_data.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/critic_data.hpp
@@ -38,14 +38,15 @@ namespace mppi
 
 /**
  * @struct mppi::CriticData
- * @brief Data to pass to critics for scoring, including state, trajectories, path, costs, and
- * important parameters to share
+ * @brief Data to pass to critics for scoring, including state, trajectories,
+ * pruned path, global goal, costs, and important parameters to share
  */
 struct CriticData
 {
   const models::State & state;
   const models::Trajectories & trajectories;
   const models::Path & path;
+  const geometry_msgs::msg::Pose & goal;
 
   xt::xtensor<float, 1> & costs;
   float & model_dt;
--- a/nav2_mppi_controller/include/nav2_mppi_controller/optimizer.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/optimizer.hpp
@@ -96,7 +96,7 @@ public:
   geometry_msgs::msg::TwistStamped evalControl(
     const geometry_msgs::msg::PoseStamped & robot_pose,
     const geometry_msgs::msg::Twist & robot_speed, const nav_msgs::msg::Path & plan,
-    nav2_core::GoalChecker * goal_checker);
+    const geometry_msgs::msg::Pose & goal, nav2_core::GoalChecker * goal_checker);
 
   /**
    * @brief Get the trajectories generated in a cycle for visualization
@@ -138,7 +138,8 @@ protected:
   void prepare(
     const geometry_msgs::msg::PoseStamped & robot_pose,
     const geometry_msgs::msg::Twist & robot_speed,
-    const nav_msgs::msg::Path & plan, nav2_core::GoalChecker * goal_checker);
+    const nav_msgs::msg::Path & plan,
+    const geometry_msgs::msg::Pose & goal, nav2_core::GoalChecker * goal_checker);
 
   /**
    * @brief Obtain the main controller's parameters
@@ -256,10 +257,12 @@ protected:
   std::array<mppi::models::Control, 4> control_history_;
   models::Trajectories generated_trajectories_;
   models::Path path_;
+  geometry_msgs::msg::Pose goal_;
   xt::xtensor<float, 1> costs_;
 
-  CriticData critics_data_ =
-  {state_, generated_trajectories_, path_, costs_, settings_.model_dt, false, nullptr, nullptr,
+  CriticData critics_data_ = {
+    state_, generated_trajectories_, path_, goal_,
+    costs_, settings_.model_dt, false, nullptr, nullptr,
     std::nullopt, std::nullopt};  /// Caution, keep references
 
   rclcpp::Logger logger_{rclcpp::get_logger("MPPIController")};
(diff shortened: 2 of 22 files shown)


SteveMacenski:
@redvinaa before you ask, no this CI failure is also not from you. Geometry2 just changed a bunch of headers from `.h` to `.hpp` and the Rolling release with those changes came out last night. https://github.com/ros-navigation/navigation2/pull/4732 resolves, but need to merge that in and you'll need to pull in those changes or rebase. **Then** any CI failures are real :laughing:

SteveMacenski:
Rebase with https://github.com/ros-navigation/navigation2/pull/4823 for CI

SteveMacenski (changes requested):
This breaks a MPPI unit test - please make sure that this works properly. You may need to broadcast a transformation in that test so that the `getTransformedGoal` method can function.

SteveMacenski (approved):
Waiting on CI to turn over - sorry you've hit landmine after landmine in our usually very stable CI. I'm not sure why circle's not cloning but its not your fault. I'm asking our CI guru to investigate :-)

ruffsl:
```shell
Cloning git repository - git@github.com:ros-navigation/navigation2.git
Cloning into '.'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
error running git clone "git@github.com:ros-navigation/navigation2.git": exit status 128
```

Hmm, it failed to even clone the repo. That is strange...

Looks like the CircleCI Deploy key got deleted from GitHub for this GitHub repo.
Did anyone purge this repo's Deploy keys
- https://github.com/ros-navigation/navigation2/settings/keys

Seems to be cloning the repo just fine now!
- https://app.circleci.com/pipelines/github/ros-navigation/navigation2/13343/workflows/4d64d956-d331-4d19-9704-675b22292a11/jobs/39973

SteveMacenski:
@redvinaa you have a test failing:

```
/opt/overlay_ws/src/navigation2/nav2_mppi_controller/test/controller_state_transition_test.cpp:67
Expected: controller->computeVelocityCommands(pose, velocity, {}) doesn't throw an exception.
  Actual: it throws nav2_core::ControllerTFError with description "Unable to transform goal pose into costmap frame".

/opt/overlay_ws/src/navigation2/nav2_mppi_controller/test/controller_state_transition_test.cpp:67
Expected: controller->computeVelocityCommands(pose, velocity, {}) doesn't throw an exception.
  Actual: it throws nav2_core::ControllerTFError with description "Unable to transform goal pose into costmap frame".
```

redvinaa:
I'm not sure what's the best way to go here, because the dummy path is generated with "odom" frame, while the costmap is in "map" frame. @SteveMacenski if you have a better solution than mine, do tell me.

The test now fails because of docking. :D Could you maybe rerun it? It passed on my machine on humble.

SteveMacenski:
> I'm not sure what's the best way to go here, because the dummy path is generated with "odom" frame, while the costmap is in "map" frame. @SteveMacenski if you have a better solution than mine, do tell me.

Fine by me - that test is a really basic exercising of the lifecycle and functions to make sure it can come up, do work, and shut down cleanly. Its not really doing any "work" in that test and each method have their own extensive test cases below. 

> The test now fails because of docking. :D Could you maybe rerun it? It passed on my machine on humble.

Something's going on with that test right now, not your fault. Merging! Thanks for the great contribution, as always :smile:

BriceRenaudeau:
This MR seems to break the `enforce_path_inversion` behavior.
The robot doesn't care for local goals anymore and is more attracted by the final goal.

[Mppi_goal_to_critic_inversion_fail.webm](https://github.com/user-attachments/assets/cf682572-be36-4476-9de1-5ad2062bf118)

SteveMacenski:
@BriceRenaudeau how so, specifically? I'm looking over the PR and I think you're referring to the change in the `GoalCritic`, but that shouldn't have been triggering before approaching the final goal anyway. Doesn't the `PathFollow` critic handle this situation OK? 

Is this happening when the inversion is with in the threshold to consider for the other critics to pass off for on-goal-approach behavior? 

The only 2 changes I see in this PR are:
* For all critics, use the global goal to evaluate bypass conditions like `withinPositionGoalTolerance`
* The `GoalCritic` uses the global goal to score (and also its bypass condition) rather than the end of the pruned path

Which of those (or both) are an issue?

alex-roba:
@SteveMacenski 




> Which of those (or both) are an issue?
I am facing the same issue.
I believe the problem occurs because GoalCritic scores based on the global goal (and its bypass condition) rather than the end of the pruned path. In some cases, inversion can be close to the goal where GoalCritic is active, causing the controller to ignore the inversion and focus only on the final goal pose.

Additionally, with the default parameters, when the robot is outside the GoalCritic threshold, the parameters may not be sufficient to guide the robot toward inversion. However, when GoalCritic considers the pruned path, it helps the robot converge within the inversion tolerances.

BriceRenaudeau:
In our case, the GoalCritic is always active to perform TEB like behavior hence cutting the inversion point.

BriceRenaudeau:
Here is another example of the problem. We want the robot in the opposite direction in a narrow aisle, it should go out of the aisle to make the half turn and then come into the aisle again.
[MPPI_inversion_point_ok.webm](https://github.com/user-attachments/assets/104d8509-e9bb-4b66-9e00-a7a3c3db6f1b)

But with this MR, the robot totally bypasses the path and just moves to the endpoint in the wrong direction.
[MPPI_inversion_point_failed.webm](https://github.com/user-attachments/assets/e6e6f1e3-9128-4f9b-ad8f-b22b278aa4b7)

SteveMacenski:
@BriceRenaudeau this keeps falling off of my radar by accident. Can you file a ticket for this with your context and we can take it from there? There are some Nav2 contributors around that are looking for projects and if we provide a detailed explanation of the issue and possible solutions, I'm sure we can find a contributor that would be willing to work on this to fix the issue. 

I think possibly we should use the 'real' goal unless there's a path inversion and we are caring about the path inversions. If so, use the intermediary goal. We can add in the parameter similar to that we have for the path follower if we should consider inversions / path pose orientations. 

Does that seem reasonable?

My apologies for the delay

PR #4824, opened 2025-01-04
Fixed the problem of static layer not restoring old map values for footprint 

CihatAltiparmak:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/issues/4825 |

---

* I have removed clearing the `map_buffer` just after The method `processMap` is run so we can utilize `map_buffer` to restore the region cleared before.
* I have removed `map_received_in_update_bounds_` because seems we really don't need to. 
* I have left to lock with mutex inside `processMap` method because now we are using lock in parts we use `processMap`

We have added some methods to Costmap2d. So we may need to update Costmap2D API documentation.

---

* We can add some methods to Costmap2d. For example, we can should pass the lambda function setConvexPolygonCost in order to implement dynamically how the cost is set.

#### For Maintainers: 


SteveMacenski (commented):
I didn't carefully review this yet (Saturday afternoon over coffee!), but wouldn't it be possible for us to simply remove the `setConvexPolygonCost` from the internal static layer grid and apply it to the master grid update (`master_grid.setConvexPolygonCost())`)? 

That would apply it more globally however which might not be wanted in some cases (though I find that generally hard to believe _some_ layers would be set for that and others are not, practically speaking).

Alternatively, you could have `setConvexPolygonCost` split into 2 stages. Stage 1 is getting the cells that belong to the footprint 

```
  // we assume the polygon is given in the global_frame...
  // we need to transform it to map coordinates
  std::vector<MapLocation> map_polygon;
  for (unsigned int i = 0; i < polygon.size(); ++i) {
    MapLocation loc;
    if (!worldToMap(polygon[i].x, polygon[i].y, loc.x, loc.y)) {
      // ("Polygon lies outside map bounds, so we can't fill it");
      return false;
    }
    map_polygon.push_back(loc);
  }

  std::vector<MapLocation> polygon_cells;

  // get the cells that fill the polygon
  convexFillCells(map_polygon, polygon_cells);

  // NEW: Now return this!
```

And return those. The second step would be to obtain their values in the loop, store them, and set the new values for FREE. After we do the master grid update, use the originally obtained values to repopulate the static layer. You'd basically just break the `setConvexPolygonCost` into 3 methods: (1) get the cells that belong to a convex shape, (2)  sets its value for a fixed value (for free), and (3) sets its value to using a vector of input of values stored originally (to reset). 

I think that's similar to what you did, but I think this would require alot less changes.

CihatAltiparmak:
Thank you for give a feedback in a light speed. 

> wouldn't it be possible for us to simply remove the setConvexPolygonCost from the internal static layer grid and apply it to the master grid update (master_grid.setConvexPolygonCost()))?

Of course, I can. I just wondered your idea about whether adding new method would be useful for future. For instance, I wanna use setConvexCost in such a way that I can update with max value (`updateWithMax`). I got my answer.

> That would apply it more globally however which might not be wanted in some cases (though I find that generally hard to believe some layers would be set for that and others are not, practically speaking).

I agree with you. I coul not get any chance to take a look at other layers, however I can see where this issue comes from. (#4282 )

SteveMacenski:
Ah, that is pretty recent. Yeah I think doing the second thing I recommended on splitting up the functions would be the best move here

CihatAltiparmak:
Hi @SteveMacenski , If you get a chance, would you take a look at my modifications, I've tried to minimize diffs to ease your work. Btw, I can consider to add the test cases to increase code coverage if you want.

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -138,6 +138,7 @@ StaticLayer::getParameters()
   declareParameter("transform_tolerance", rclcpp::ParameterValue(0.0));
   declareParameter("map_topic", rclcpp::ParameterValue("map"));
   declareParameter("footprint_clearing_enabled", rclcpp::ParameterValue(false));
+  declareParameter("restore_outdated_footprint", rclcpp::ParameterValue(false));
```
SteveMacenski, line 141: Make sure to open a docs PR on docs.nav2.org that adds this parameter to the static layer & in the migration guide that this change is made

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -282,13 +279,9 @@ StaticLayer::incomingMap(const nav_msgs::msg::OccupancyGrid::SharedPtr new_map)
     RCLCPP_ERROR(logger_, "Received map message is malformed. Rejecting.");
     return;
   }
-  if (!map_received_) {
```
SteveMacenski, line 285: Revert

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -220,12 +221,8 @@ StaticLayer::processMap(const nav_msgs::msg::OccupancyGrid & new_map)
       new_map.info.origin.position.x, new_map.info.origin.position.y);
   }
 
-  unsigned int index = 0;
-
-  // we have a new map, update full size of map
-  std::lock_guard<Costmap2D::mutex_t> guard(*getMutex());
```
SteveMacenski, line 226: Revert

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -161,8 +163,7 @@ StaticLayer::getParameters()
 
   // Enforce bounds
   lethal_threshold_ = std::max(std::min(temp_lethal_threshold, 100), 0);
-  map_received_ = false;
```
SteveMacenski, line 164: Revert

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -339,20 +370,20 @@ StaticLayer::updateBounds(
...
 
   // If there is a new available map, load it.
-  if (map_buffer_) {
+  if (has_map_to_process_) {
     processMap(*map_buffer_);
-    map_buffer_ = nullptr;
+    has_map_to_process_ = false;
   }
 
+  updateFootprint(robot_x, robot_y, robot_yaw, min_x, min_y, max_x, max_y);
```
SteveMacenski, line 385: Revert

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -458,6 +501,7 @@ StaticLayer::updateCosts(
       }
     }
   }
+
```
SteveMacenski, line 504: Revret

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -412,7 +448,14 @@ StaticLayer::updateCosts(
   }
 
   if (footprint_clearing_enabled_) {
-    setConvexPolygonCost(transformed_footprint_, nav2_costmap_2d::FREE_SPACE);
+    if (restore_outdated_footprint_) {
```
SteveMacenski, line 451: Wouldn't this need to happen at the end of the function, after we update the master map?

SteveMacenski: I think some of htese functions can live in the costmap_2d object like `setConvexPolygonCost` -- then have `setConvexPolygonCost` itself use some of these to minimize code duplication

SteveMacenski (changes requested):
Please revert changes to the workflows of the node, I think it creates a number of regressions. I think you should only need to adjust the `updateCosts` portion of the implementation.

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: I have removed this parameter, but if you are willing to add it, it can be updated the Costmap2D object itself of `static layer`. Thus, it can be permanent in the master grid. What do you think?

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: Done it. (https://github.com/ros-navigation/navigation2/pull/4824/commits/7b4cee3456c8274f6b55d6ac06fe6cfe3aa43205)

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: Done it (https://github.com/ros-navigation/navigation2/pull/4824/commits/7b4cee3456c8274f6b55d6ac06fe6cfe3aa43205)

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: Done it (https://github.com/ros-navigation/navigation2/pull/4824/commits/7b4cee3456c8274f6b55d6ac06fe6cfe3aa43205)

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: I couldn't find how to revert this. Because, if `updateFootprint` doesn't run continuously, `master_grid` cannot be restored. But I will think about that.

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: Done (https://github.com/ros-navigation/navigation2/pull/4824/commits/7b4cee3456c8274f6b55d6ac06fe6cfe3aa43205)

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: It's really cool idea. I am sory I couldn't get it at the beginning of the convervation. It dramatically decreased the diffs.

nav2_costmap_2d/plugins/static_layer.cpp
SteveMacenski: How it is being restored now?

CihatAltiparmak: If we don't execute `updateFootprint` before if checks, it will updated only when new occupancy map msg is published. This is my concern. In my way, if `footprint_clearing_enabled` is true, the area occupied by robot's footprint is updated per each `updateCost` execution. In current nav2 stack, it is hard to see that. If I am wrong, fix me please.

CihatAltiparmak:
Firstly, my apologies for long delay due to my exams so on. 

> I think what this currently does is clear the footprint on the master grid after update, but that would bypass the set updateWithXYZ() policies that users can set in order to set how they want updates to be performed (i.e. use max value or have it overwrite). This would lose information like having non-zero costs due to sensor measurements, even if removing the static map values, which wouldn't be a solution without introducing a regression.

I have a question here. why do we have to update the areas with max value or overwrite? Doesn't `footprint_clearing_enabled` looses its meaning here? But if you still haven't change the idea, i will try to fix it in no time.

> I also don't understand why updateFootprint as a method was updated. I think the footprint must be transformed into the current frame before touch-ing the points to expand the bounds. I also don't think that updateFootprint should be completed if not updating its bounds. Can/should these be reverted?

I commented on [one of your reviews](https://github.com/ros-navigation/navigation2/pull/4824#discussion_r1931192674), but it saw this later. You are right. Ignore my comments in [that review](https://github.com/ros-navigation/navigation2/pull/4824#discussion_r1931192674).

I will figure out all of them.

nav2_costmap_2d/plugins/static_layer.cpp
CihatAltiparmak: Instead of updating only when new map has been published, what about to check how `updateFootprint` updates the bounds of the `costmap_`? Thus, we can prevent the potential incompabilities between `costmap_` and `master_grid`.

Pseudo is like that. 
```
void
StaticLayer::updateFootprint(
  double robot_x, double robot_y, double robot_yaw,
  double * min_x, double * min_y,
  double * max_x,
  double * max_y)
{
  if (!footprint_clearing_enabled_) {return;}

  for (const auto & point : transformed_footprint_) {
    unsigned int mx, my;
    if (!worldToMap(point.x, point.y, mx, my)) {
      return;
    }

    touch(point.x, point.y, min_x, min_y, max_x, max_y);
  }

  transformFootprint(robot_x, robot_y, robot_yaw, getFootprint(), transformed_footprint_);
}
```

SteveMacenski:
No need for apologies, life happens :wink: 

> I have a question here. why do we have to update the areas with max value or overwrite after clearing the footprint area? Doesn't footprint_clearing_enabled lose its meaning here? But if you still haven't changed your idea, I will try to fix it in no time.

There's a setting `use_max_` that users set for each layer about how to combine the various costmap layer plugins into a single master costmap that's used by the planning/control algorithms. They include updating with max value (i.e. from the existing master costmap, which may have been partially updated by other layers, is this cost greater than the existing one?) or override value (i.e. this value always takes priority, no matter what is already in the master costmap or inserted from other layers before this one).

The behavior that this PR currently implements is overriding of values. The master costmap is always cleared fully under footprint, rather than having the option to combine using the max value that could already exist in the costmap from another layer or a previous update. 

The clear under footprint is specific to each layer, not to the master costmap itself. The combination policy that the user selected the layer to run at should be respected :-)

CihatAltiparmak:
Hello, I tried to apply your suggestion. But I have some concerns. I found out that updateFootprint is only executed when clear costmap entirely service. This bring about the costmap is updated only when no plan is found. Of course, this depends on how behavior tree is implemented. I need to do some brain storming about it. It is hard to see this issue in `Obstacle Layer`. Because, obstacle layer takes into account the obstacles and updates bounds frequently.

I wonder your feedbacks. If I am in the right way, I will continue by adding `restore_cleared_map_region` parameter. Then I will create a PR about this changes.

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -411,8 +411,17 @@ StaticLayer::updateCosts(
     return;
   }
 
+  std::vector<std::pair<MapLocation, unsigned char>> map_region_to_restore;
   if (footprint_clearing_enabled_) {
-    setConvexPolygonCost(transformed_footprint_, nav2_costmap_2d::FREE_SPACE);
+    std::vector<MapLocation> polygon_cells;
+    getCellsOccupiedByPolygon(transformed_footprint_, polygon_cells);
```
SteveMacenski, line 417: I think this getter should also return the costs so that you don't need to then have the next step here be grabbing the costs.

The return could be the vector of pairs (or map locations and costs).

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -458,6 +467,11 @@ StaticLayer::updateCosts(
       }
     }
   }
+
+  // restore the map region occupied by the polygon using cached data
+  for (auto [cell, cost_value] : map_region_to_restore) {
```
SteveMacenski, line 472: I think you can make a costmap_2d method for `setCellsOccupiedByPolygon` instead - which might be nice to use in other places anyway :-)

SteveMacenski (changes requested):
Almost perfect, just a couple of architectural changes to make these more portable for other applications, but then good to go! Thanks for your time and effort prototyping this a few ways!

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -138,6 +138,7 @@ StaticLayer::getParameters()
   declareParameter("transform_tolerance", rclcpp::ParameterValue(0.0));
   declareParameter("map_topic", rclcpp::ParameterValue("map"));
   declareParameter("footprint_clearing_enabled", rclcpp::ParameterValue(false));
+  declareParameter("restore_outdated_map", rclcpp::ParameterValue(false));
```
SteveMacenski, line 141: Lets rename this + the variable to something like `restore_cleared_footprint`. Make sure to also update the migration guide and configuration docs with this.

Lets default to true, I think this is the most reasonable behavior!

nav2_costmap_2d/src/costmap_2d.cpp
```diff
@@ -399,29 +399,59 @@ void Costmap2D::updateOrigin(double new_origin_x, double new_origin_y)
...
       // ("Polygon lies outside map bounds, so we can't fill it");
       return false;
     }
     map_polygon.push_back(loc);
   }
 
-  std::vector<MapLocation> polygon_cells;
-
   // get the cells that fill the polygon
+  std::vector<MapLocation> polygon_cells;
```
SteveMacenski, line 447: Do a `polygon_cells.reserve(100);` to reserve the size to 100 cells, that'll help with resizings as you're pushing back. Larger footprints will still need to resize, but smaller ones shouldn't.

SteveMacenski (changes requested):
Small nitpicks, but otherwise this is perfect and ready!

SteveMacenski:
> I couldn't find a way to keep setConvexPolygonCost method optimized. But if you want, I can revert it(setConvexPolygonCost), which gives rise to keep similar code blocks in the costmap2d class.

In what way? Is this significantly slower now? 

Something you can do is in `getMapRegionOccupiedByPolygon`, you don't need the 2x loops. In the first loop where we populate `map_polygon`, you can directly populate the `polygon_map_region` if you simply `getCost` at that point. 

> Finally, restore_outdated_map depends on footprint_clearing_enabled. Is it strictly necessary to check it in dynamic parameter updates?

That might be good yes. Only set the dynamic param if `footprint_clearing_enabled_ == true`

CihatAltiparmak:
> In what way? Is this significantly slower now?

In the latest implementation, vector consists of pairs instead of only location. so this will increase the storage size of the vector. When we leave `setConvexPolygonCost` as it is, `getMapRegionOccupiedByPolygon` and `setConvexPolygonCost` will have similar code block. That's what i say. But, I did not measure it. It is just my hypothesis.

> Something you can do is in getMapRegionOccupiedByPolygon, you don't need the 2x loops. In the first loop where we populate map_polygon, you can directly populate the polygon_map_region if you simply getCost at that point.

We cannot handle this in the first loop. It's because the first loop only handles the lines of the polygon. To do that, we need to modify `convexFillCells`. Maybe, I can modify `PolygonOutlineCells` class which is passed as action to `bresenham2D` method. On the other hand, I am trying to modify as less method as possible. What is your opinion here?

Btw, I will apply your feedbacks today.

CihatAltiparmak:
Hello, 

I conducted a basic benchmark to measure the effect of our changes. The execution time of my `setConvexPolygonCost` approximatelly equals to 1.34 times slower than previous one. 

The mean of the default implementation's execution time :  ~20000ns
The mean of the new implementation's execution time :  ~27000ns

You can take a look at how i conducted [this benchmark](https://gist.github.com/CihatAltiparmak/03a465b72a1c2ce30897266d9a37df2d#file-static_layer-cpp) if you want.

As you said, I added checks in dynamic parameter update. But I wanna say that setting `restore_cleared_footprint` as true will change the default behaviour of the software stack. It should either inform the users about this change or we should set `restore_cleared_footprint` as false in default.

nav2_costmap_2d/plugins/static_layer.cpp
```diff
@@ -498,6 +508,17 @@ StaticLayer::dynamicParametersCallback(
         current_ = false;
       } else if (param_name == name_ + "." + "footprint_clearing_enabled") {
         footprint_clearing_enabled_ = parameter.as_bool();
+        if (!footprint_clearing_enabled_) {
```
SteveMacenski, line 511: You can remove this log - I think the other one for restore cleared footprint is sufficient

SteveMacenski:
> But I wanna say that setting restore_cleared_footprint as true will change the default behaviour of the software stack

I agree, but this is a good thing. The migration guide is the right place for that https://docs.nav2.org/migration/Jazzy.html. You can add in an entry about the new parameter and default value

SteveMacenski:
>  The execution time of my setConvexPolygonCost approximatelly is 1.34 times slower than previous one.

What if you reserve in polygon_map_region there as well? I'm guessing most of those are the vector resizings since std::vector initializes with size of 0 and each time it resizes, it doubles and copies things over. For 100, that would remove ~6-7 copies.

CihatAltiparmak:
Seems it speeds up when reserved the vector. But the new rate is ~1.31. I had to wait to get rid of the spikes in the benchmark. 

The mean of the new implementation's execution time : ~26500ns

SteveMacenski:
What are you measuring, the static layer update time or `setConvexPolygonCost`? We don't actually use `setConvexPolygonCost` in the static layer anymore (only in the obstacle layer now). It might be good to check where that time is coming from so we understand at least. 

But, it might be worth keeping that version optimized if need be (if measuring `setConvexPolygonCost`).

CihatAltiparmak:
> What are you measuring, the static layer update time or setConvexPolygonCost?

I measured the execution time of the `setConvexPolygonCost`. I have realized that I was benchmarking on balanced mode of the CPU. Instead, I benchmarked on performance mode of CPU. The benchamark results changed significantly. 

The execution time of `setConvexPolygonCost (default implementation): ~4000ns
The execution time of `setConvexPolygonCost (my latest implementation): ~4350ns

Consequently, It increases the computational time by ~8%.

SteveMacenski (approved):

PR #4832, opened 2025-01-08
update padFootprint() function

Nisarg236:
---

## Basic Info

| ------ | ----------- |

---

* Footprint gets distorted when the origin of the footprint is not at the centroid of the footprint.
* Negative values of padding will shrink the footprint

Case 1: (0,0), (0,2), (2,2), (2,0)
image
Case 2: (10,10), (10,12), (12,12) , (12,10)
image

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/footprint.cpp
+++ b/nav2_costmap_2d/src/footprint.cpp
@@ -142,13 +142,40 @@ void transformFootprint(
   }
 }
 
+geometry_msgs::Point calculateCentroid(const std::vector<geometry_msgs::Point>& footprint)
+{
+  geometry_msgs::Point center;
+  double sum_x = 0.0, sum_y = 0.0;
+
+  for (const auto& point : footprint)
+  {
+    sum_x += point.x;
+    sum_y += point.y;
+  }
+
+  if (!footprint.empty()) {
+    center.x = sum_x / footprint.size();
+    center.y = sum_y / footprint.size();
+  }
+
+  return center;
+}
+
 void padFootprint(std::vector<geometry_msgs::msg::Point> & footprint, double padding)
 {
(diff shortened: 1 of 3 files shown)


SteveMacenski:
Please fix build, linting, and DCO issues before we give this a review :-)

nav2_costmap_2d/src/footprint.cpp
```diff
@@ -142,13 +142,40 @@ void transformFootprint(
...
+
+  return center;
+}
+
 void padFootprint(std::vector<geometry_msgs::msg::Point> & footprint, double padding)
 {
-  // pad footprint in place
-  for (unsigned int i = 0; i < footprint.size(); i++) {
-    geometry_msgs::msg::Point & pt = footprint[i];
-    pt.x += sign0(pt.x) * padding;
```
SteveMacenski, line 150: Please point out to me the error in the code here - I don't see any immediately. 

Is the issue the case where you have a coordinate exactly at `0.0`? `0.0` for the footprint would be the center of your robot, since that is the `0.0, 0.0` w.r.t. your base_link or base_footprint frame -- which is almost always the center of your robot or aligned with some axle that wouldn't then be a corner of a robot. That would point to me more an issue in your configuration

SteveMacenski: See https://docs.nav2.org/setup_guides/footprint/setup_footprint.html

SteveMacenski: @Nisarg236 should this be closed?

SteveMacenski:
Closing due to lack of response -- happy to review / reopen but seems the setup guide was ACKed and thereby solved the misconfiguration profile?

PR #4838, opened 2025-01-09
[lifecycle_manager] expose service_timeout

doisyg:
---

## Basic Info

| ------ | ----------- |

---

- This adds a parameter `service_timeout`, default to 5s, to control explicitly the timeout applied to the `change_state` and `get_state` calls in the `lifecycle_manager` node. Which were functionally respectively 5s and 2s (`LifecycleServiceClient` default). This is useful to stabilize bringup in CPU limited systems (or in systems whith looooooots of nodes started at the same time) where a `lifecycle_manager` could failed to transition its managed nodes due to `change_state` or `get_state` timing out.
- Slight refactor of `LifecycleServiceClient` to not crop the timeout to plain seconds in `change_state` and `get_state`. + replacing the 1 `change_state` overload by a default parameter.

Probably: https://docs.nav2.org/configuration/packages/configuring-lifecycle.html

#### For Maintainers: 


SteveMacenski (commented):
I agree with these changes, however I recall trying to add this before to handle one of the few places that there aren't networking timeouts set and it causing issues/instability in the stack. I think related to service calls in ROS 2 with timeouts being alot more flaky to actually get sent/processed (even when that  timeout is like 2000s) than if you sent it without the timeout -- for some weird RMW/DDS reason I'm sure. If memory serves, it made it flaky for the system to actually get brought up some non-trivial percentage of the time, so I was forced to revert it after a ton of user reports. 

I want to see what CI will say but the build was failing. I just re-triggered but if it has those permission issues again, please ping Ruffin to take a look

doisyg:
Summoning @ruffsl ,)

Yes, services in ros2 have a quite a history of failing due to middleware issues, I still have PTSD. I am not ruling it out in our system. At the moment, when bringup all our nodes all together at the same time, we can have the `get_state` of one of our many lifecycle managers randomly  taking more than 5s to get a reply. I am not sure yet if it is just taking more time (optimistic scenario) or failing completely (pessimistic scenario, middleware issue). 

In terms of this PR impact, the change of behavior system wide should be limited: when the new `service_timeout` parameter is not explicitly set, it falls back to a default of 5s. Which was the value applied to `get_change` before this PR, but `change_state` had a default of 2s. So the only functional impact (when no using the new parameter) is that the timeout of the `change_state` call of the lifecycle manager is increased from 2 to 5s.

doisyg:
Test re-run (thanks @ruffsl), now `test_docking_server` failing, related to this PR you think @SteveMacenski ?

SteveMacenski:
The current job run has every system test failing :-) https://app.circleci.com/pipelines/github/ros-navigation/navigation2/13347/workflows/b2a18891-9bba-43c9-9316-dd05b976ee2e/jobs/40015/tests


> we can have the get_state of one of our many lifecycle managers randomly taking more than 5s to get a reply. 

Can we then update only that, exposing the hardcoded timeout as a parameter, without changing the other methods to using a timeout? This PR removes the timeout-less `invoke` version of these methods that we currently rely on.

SteveMacenski:
@doisyg any update here?

doisyg:
Okay, sorry for the late reply. Bear with
To discuss: currently on main, in `LifecycleManager::changeStateForNode`, we have:
- a call to `LifecycleServiceClient::change_state(std::uint8_t transition)`, the version with no timeout argument,  which calls:
  - `!change_state_.wait_for_service(5s)` => hardcoded 5 seconds timeout
  -  `ServiceClient::change_state` `change_state_.invoke(request, response)`, the version with no timeout => no timeout, will wait the answer forever
- a call to `LifecycleServiceClient::get_state()`, which has a default initialization of `timeout` to 2 seconds when no argument is used, and which calls:
  - `!change_state_.wait_for_service(timeout)` => 2 seconds timeout
  -  `LifecycleServiceClient::get_state` `get_state_.invoke(request, timeout)` => 2 seconds timeout
 
 So 3 different way to call services:
 - no timeout (loop forever)
 - a 2 seconds timeout
 - a 5 seconds timeout
 
 which I confused in this PR. My initial problem being that I had at least one of these service timing out when starting a lifecycle manager in challenging CPU usage condition (along with the rest of my huge stack).
https://github.com/ros-navigation/navigation2/blob/4694386872365851c840737515f89e3d409c1d9f/nav2_lifecycle_manager/src/lifecycle_manager.cpp#L244C1-L261

SteveMacenski:
Is what you want to do to expose the hardcoded values instead? 

I agree that I don't like that we have a non-timeout version in use, but it is currently needed due to service reliability issues in rclcpp (but most likely actually from DDS). I think what you want is actually to have the `2s` and `5s` be exposed so that you can change those values (which this PR partially does). 

I think just for the case that we have the non-timing-out calls, those should not be replaced with timeout-able calls. But the current timeout-able calls can have its parameter given. Does that make sense / fix your issue?

Namely, readd 

```
change_state(std::uint8_t transition)
```

and change back to using it in `changeStateForNode`. `get_state` calls to use given set timeout.

doisyg:
> Namely, readd
> 
> ```
> change_state(std::uint8_t transition)
> ```
> 
> and change back to using it in `changeStateForNode`. `get_state` calls to use given parameter set timeout, rather than using the default 2s.


The problem with this, it that there will still be a hardcoded 5s timeout used in `change_state(std::uint8_t transition)` : https://github.com/ros-navigation/navigation2/blob/3e4f0cbc85a69fb6748378f6e0b0c324f5285666/nav2_util/src/lifecycle_service_client.cpp#L85, which is the original cause of my issues.

I cannot find a nice way to solve this while maintaining the legacy behavior and not creating an extra "wait_for_service_timeout".

I just pushed a new proposal which maintains the legacy behavior when "service_timeout" is not set. But when it is set, apply this  "service_timeout" everywhere. Not very clean, but I don't see another way without changing the current default behavior.

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -239,6 +239,8 @@ class LifecycleManager : public rclcpp::Node
   rclcpp::TimerBase::SharedPtr bond_timer_;
   rclcpp::TimerBase::SharedPtr bond_respawn_timer_;
   std::chrono::milliseconds bond_timeout_;
+  std::chrono::milliseconds service_timeout_;
+  bool non_default_timeout;
```
SteveMacenski, line 243:
```suggestion
  bool non_default_timeout_;
```

SteveMacenski: Also, I see a `get_parameter` for this, but no declare

nav2_lifecycle_manager/src/lifecycle_manager.cpp
```diff
@@ -254,11 +276,21 @@ LifecycleManager::changeStateForNode(const std::string & node_name, std::uint8_t
 {
   message(transition_label_map_[transition] + node_name);
 
-  if (!node_map_[node_name]->change_state(transition) ||
-    !(node_map_[node_name]->get_state() == transition_state_map_[transition]))
-  {
-    RCLCPP_ERROR(get_logger(), "Failed to change state for node: %s", node_name.c_str());
-    return false;
+  if (non_default_timeout) {
+    if (!node_map_[node_name]->change_state(transition, service_timeout_) ||
```
SteveMacenski, line 280: All instances of `get_state` I approve, but for `change_state`, I think that should be timeout-less and would still fit the need that your described to increase the timeout from `2s` to something higher for your timeout issue. If no timeout is given, then it spins indefinitely. 

One reason I think for `change_state` that is appropriate is because lifecycle transitions are user-implemented and can take arbitrarily long. For example, the planner server will wait for a map->base_link transform before it finishes activating which is dependent on the user to set an initial pose for AMCL. Making any specific number is a race for users to initialize the pose on startup if not doing programmatically.  

Having a timeout for the `wait_for_service` I think is fine to set as a parameterized value (which I think is your concern?), but without using it for `invoke` which is the concern. 

What we could do instead is to collapse both functions into one:

```
bool LifecycleServiceClient::change_state(
  const uint8_t transition,
  const milliseconds wait_timeout,
  const milliseconds transition_timeout = -1);
```

All instances use `wait_timeout` to the parameterized value, but not for the transition timeout. We use the 2 `invoke()` implementations based on if the transition_timeout is valid:

```
  auto response = change_state_.invoke(request, timeout); 
  bool state = change_state_.invoke(request, response)
```

SteveMacenski: If you agree with that in concept, then I think this legacy handling can be removed. `get_state` always used a timeout and increasing that from 2s by default to 5s is totally fine by me to merge in.

nav2_lifecycle_manager/include/nav2_lifecycle_manager/lifecycle_manager.hpp
doisyg: "service_timeout" declared above https://github.com/ros-navigation/navigation2/blob/1d64f8ae6673b640d992c433de5d0698493a1e79/nav2_lifecycle_manager/src/lifecycle_manager.cpp#L45

But as you suggested below, we can do it without the legacy handling, so I will probably remove "non_default_timeout"

SteveMacenski: OK!

nav2_lifecycle_manager/src/lifecycle_manager.cpp
doisyg: I agree!
Just pushed a new version, though I had to reverted the argument in `change_state` in order to not break the existing using, for instance there: https://github.com/ros-navigation/navigation2/blob/6e58fbe2888cf88d3040edb4b083b8dc1dfd0cd6/nav2_map_server/test/component/test_map_saver_node.cpp#L48

```
  bool change_state(
    const uint8_t transition,  // takes a lifecycle_msgs::msg::Transition id
    const std::chrono::milliseconds transition_timeout = std::chrono::milliseconds(-1),
    const std::chrono::milliseconds wait_for_service_timeout = std::chrono::milliseconds(5000));
```

And in the `LifecycleManager` I am now using the `service_timeout` parameter as both the `wait_for_service_timeout` of `change_state`, and the `timeout` of `get_state`.

doisyg:
Sorry that it is taking me so long to iterate on this one guys. New proposal pushed.

nav2_util/src/lifecycle_service_client.cpp
```diff
@@ -67,33 +67,26 @@ LifecycleServiceClient::LifecycleServiceClient(
...
-  auto response = change_state_.invoke(request, timeout);
-  return response.get();
-}
-
-bool LifecycleServiceClient::change_state(
-  std::uint8_t transition)
-{
-  if (!change_state_.wait_for_service(5s)) {
-    throw std::runtime_error("change_state service is not available!");
+  if (transition_timeout > milliseconds(0)) {
```
SteveMacenski, line 79:
```suggestion
  if (transition_timeout > 0ms) {
```

This should work and look nicer (pendantic)

SteveMacenski (approved):

nav2_util/src/lifecycle_service_client.cpp
SteveMacenski: Looks like your fork changed so I can't merge my own suggestions as a maintainer - please merge this in and I can merge!

SteveMacenski:
@BriceRenaudeau you OK with this?

BriceRenaudeau:
I will try it soon.
For the update of the `managed_nodes_state_`, I may open a PR later.

BriceRenaudeau:
Works fine for me.
@doisyg do you have a specific `service_timeout` that works well with systems with lots of nodes to bringup?

doisyg:
We use 20 s and have no more issues. 
But I would say set at timeout as long as the duration of your launch sequence

Issue #4691, opened 2024-09-25
Investigate adding in explicit IPC for Nav2 nodes again


SteveMacenski:
Previously, we didn't do this since IPC required the use of only the default QoS profiles and we use several others in Nav2 for various things. We should investigate if this challenge has been overcome now and we can use IPC.

Additionally, Nav2 default launches with rviz2 and application nodes that may subscribe to topics outside of the container. In this case, these performance improvements are negated, but I suppose worthwhile to include unless there's some reason _not_ to. I believe that only the topics that are communicating outside of the process break IPC, but worth checking on that to make sure the entire node isn't broken out of IPC for a debug logging topic or something. 

It would be great to get an update from @clalancette on these thoughts -- then execute as it is seen fit.

clalancette:
> It would be great to get an update from @clalancette on these thoughts -- then execute as it is seen fit.

So during the Jazzy development cycle, we ended up merging in https://github.com/ros2/rclcpp/pull/2303 , which added `transient_local` functionality to the intra-process manager.  However, as far as I understand, the intra-process manager still doesn't support the full range of QoS options.

Still, if transient_local functionality is enough for your use-case, you could consider turning it on.  There are 2 different ways to turn it on:
1.  You can pass an `rclcpp::NodeOption` when constructing the node that will create *all* publishers and subscriptions (including the built-in ones) as intra-process.
2.  When you create individual publishers and subscriptions, you can turn on intra-process mode for just that entity.

> Additionally, Nav2 default launches with rviz2 and application nodes that may subscribe to topics outside of the container. In this case, these performance improvements are negated, but I suppose worthwhile to include unless there's some reason _not_ to. I believe that only the topics that are communicating outside of the process break IPC, but worth checking on that to make sure the entire node isn't broken out of IPC for a debug logging topic or something.

The intra-process manager is fairly sophisticated.  If a publisher and subscription on the same topic with the same type are created within the same Context, and intra-process is turned on for both of them, then it will communicate between them using only intra-process.  If any one of those conditions is false, it won't use it.  However, even in the case where intra-process is used, it still *always* creates a DDS entity for the outside world to discover.  That's so that things like `ros2 topic list` still show the topic.  In the case where you have an intra-process publisher, an intra-process subscription, and a non-intra-process subscription, then what happens is that intra-process is still used between the publisher and the intra-process subscription, and a copy is used to send out to the non-intra-process subscription.

[...]

PR #4841, opened 2025-01-10
intraprocess bond::Bond::BondStatusCB use after free

ewak:
The Bond mechanism includes creation of a subscription using a reference to a member function(bondStatusCB) of the Bond class.

This member function operates on member variables.

The lifecycle_node was calling bond_.reset() which releases the memory as far as the lifecycle_node
is concerned but this is not immediately released
from the rclcpp internal mechanisms (especially intraprocess). As a result the bondStatusCB function can called after it has been freed.  This use after free shows up reliably with asan when running the test_bond test.

This change allows the test_bond to suceed by
calling bond_->breakBond() (rather than bond_.reset()) to break the bond rather than expecting it to
be done cleanly by the ~Bond() destructor.

Is it enough is TBC.

Other possibilities might be to get the Bond to
inherit from std::enable_shared_from_this(), as Ros2 Nodes do, so that the pointer to the Bond member function bondStatusCB function remains valid until the subscription is released during destruction.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4691 |
| Robotic platform tested on | none |
| Does this PR contain AI generated software? | no |

---

Call bond_->breakBond() in on_deactivate rather than bond_.reset() to avoid use after free caused by dangling pointer held by rclcpp intraprocess subscription callback mechanism.

none

---

* Consider effect of only using breakBond() rather than reset().  
* Should LifeCycleNode::destroyBond() wait for some timeout for the bond->breakBond() to take affect using waitUntilBroken() ?
* Figure out an improvement to rclcpp intraprocess subscription mechanism to check for memory lifetime of passed in callback function.
* Maybe make changes to bondcpp to have Bond inherit from std::enable_shared_from_this() to extend the lifetime of the subscription callback function.
* Consider creating the bond_ as a shared_ptr rather than a unique_ptr to allow the internal mechanisms of rclcpp to hold onto valid memory until they have a chance to cleanup properly.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_cloud.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_cloud.cpp
@@ -193,15 +193,13 @@ void voxelCallback(const nav2_msgs::msg::VoxelGrid::ConstSharedPtr grid)
   pcl_header.frame_id = frame_id;
   pcl_header.stamp = stamp;
 
-  if (pub_marked->get_subscription_count() > 0)
-  {
+  if (pub_marked->get_subscription_count() > 0) {
     auto cloud = std::make_unique<sensor_msgs::msg::PointCloud2>();
     pointCloud2Helper(cloud, num_marked, pcl_header, g_marked);
     pub_marked->publish(std::move(cloud));
   }
 
-  if (pub_unknown->get_subscription_count() > 0)
-  {
+  if (pub_unknown->get_subscription_count() > 0) {
     auto cloud = std::make_unique<sensor_msgs::msg::PointCloud2>();
     pointCloud2Helper(cloud, num_unknown, pcl_header, g_unknown);
     pub_unknown->publish(std::move(cloud));
--- a/nav2_docking/opennav_docking/test/test_docking_server.py
+++ b/nav2_docking/opennav_docking/test/test_docking_server.py
@@ -17,32 +17,47 @@ import time
 import unittest
 
 from action_msgs.msg import GoalStatus
+from ament_index_python.packages import get_package_prefix
 from geometry_msgs.msg import TransformStamped, Twist, TwistStamped
 from launch import LaunchDescription
+from launch.actions import (
+    LogInfo,
+    SetEnvironmentVariable,
+)
 from launch_ros.actions import Node
 import launch_testing
+import launch_testing.actions
+import launch_testing.asserts
+import launch_testing.markers
+import launch_testing.util
 from nav2_msgs.action import DockRobot, NavigateToPose, UndockRobot
 import pytest
 import rclpy
-import rclpy.time
-import rclpy.duration
 from rclpy.action.client import ActionClient
 from rclpy.action.server import ActionServer
+import rclpy.duration
+import rclpy.time
 from sensor_msgs.msg import BatteryState
-from tf2_ros import TransformBroadcaster, Buffer, TransformListener
+from tf2_ros import Buffer, TransformBroadcaster, TransformListener
 
 
 # This test can be run standalone with:
 # python3 -u -m pytest test_docking_server.py -s
 
 @pytest.mark.rostest
+@launch_testing.markers.keep_alive
 def generate_test_description():
-
-    return LaunchDescription([
-        Node(
+    tmux_gdb_prefix = (
+        'tmux split-window '
+        + get_package_prefix('nav2_bringup')
+        + '/lib/nav2_bringup/gdb_tmux_splitwindow.sh'
+    )
+    logme = LogInfo(msg=f'tmux_gdb_prefix={tmux_gdb_prefix}')
+    docking_server = Node(
             package='opennav_docking',
             executable='opennav_docking',
             name='docking_server',
+            # arguments=['--ros-args', '--log-level', 'debug'],
             parameters=[{'wait_charge_timeout': 1.0,
                          'controller': {
                              'use_collision_detection': False,
@@ -59,17 +74,42 @@ def generate_test_description():
                              'pose': [10.0, 0.0, 0.0]
                          }}],
             output='screen',
-        ),
-        Node(
+        )
+    lifecycle_manager_navigation = Node(
             package='nav2_lifecycle_manager',
             executable='lifecycle_manager',
             name='lifecycle_manager_navigation',
             output='screen',
+            # prefix=tmux_gdb_prefix,
             parameters=[{'autostart': True},
                         {'node_names': ['docking_server']}]
-        ),
+        )
+
+    return LaunchDescription([
+        SetEnvironmentVariable('RCUTILS_LOGGING_BUFFERED_STREAM', '1'),
+        SetEnvironmentVariable('RCUTILS_LOGGING_USE_STDOUT', '1'),
+        logme,
+        docking_server,
+        lifecycle_manager_navigation,
+        # In tests where all of the procs under tests terminate themselves, it's necessary
+        # to add a dummy process not under test to keep the launch alive. launch_test
+        # provides a simple launch action that does this:
+        launch_testing.util.KeepAliveProc(),
         launch_testing.actions.ReadyToTest(),
-    ])
+    ]), locals()
+
+
+@launch_testing.post_shutdown_test()
+class TestPostShutdown(unittest.TestCase):
+
+    def test_action_graceful_shutdown(self,
+                                      proc_info,
+                                      docking_server,
+                                      lifecycle_manager_navigation):
+        """Test graceful shutdown."""
+        launch_testing.asserts.assertExitCodes(proc_info, process=docking_server)
+        launch_testing.asserts.assertExitCodes(proc_info,
+                                               process=lifecycle_manager_navigation)
 
 
 class TestDockingServer(unittest.TestCase):
@@ -146,7 +186,7 @@ class TestDockingServer(unittest.TestCase):
         return result
 
     def check_transform(self, timeout_sec, source_frame='odom', target_frame='base_link'):
-        """Try to look up the transform we're publishing"""
+        """Try to look up the transform we're publishing."""
         if self.transform_available:
             return True
 
@@ -162,8 +202,8 @@ class TestDockingServer(unittest.TestCase):
 
             self.node.get_logger().info('Successfully found transform:')
             self.node.get_logger().info(f'Translation: [{transform.transform.translation.x}, '
-                                 f'{transform.transform.translation.y}, '
-                                 f'{transform.transform.translation.z}]')
+                                        f'{transform.transform.translation.y}, '
+                                        f'{transform.transform.translation.z}]')
             self.transform_available = True
         except Exception as e:
             self.node.get_logger().error(f'Error looking up transform: {str(e)}')
@@ -178,7 +218,6 @@ class TestDockingServer(unittest.TestCase):
         self.tf_buffer = Buffer()
         self.tf_listener = TransformListener(self.tf_buffer, self.node)
         self.transform_available = False
-
 
         # Create a timer to run "control loop" at 20hz
         self.timer = self.node.create_timer(0.05, self.timer_callback)
@@ -222,7 +261,7 @@ class TestDockingServer(unittest.TestCase):
         # Test docking action
         self.action_result = []
         assert self.dock_action_client.wait_for_server(timeout_sec=5.0), \
-          'dock_robot service not available'
+               'dock_robot service not available'
 
         goal = DockRobot.Goal()
         goal.use_dock_id = True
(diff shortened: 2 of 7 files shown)


SteveMacenski:
I thought about that potentially, but that is called in the Bond Destructor already when we reset the shared pointer [1], so that should be done first thing as the object is being destroyed. That was on my list to try to see if it virtually fixed the issue, but I think we should dig a bit deeper to fix `bond` or `rclcpp` itself for whatever is happening so that this issue doesn't come back to bite us in another place or other bond users can do this too. 

> Should LifeCycleNode::destroyBond() wait for some timeout for the bond->breakBond() to take affect using waitUntilBroken() ?

`destroyBond` resets the shared pointer which calls the `~Bond` method, we should have already broken in the destructor and then waited. If it failed after waiting, we should have seen a log from `~Bond`. 

You provide some example fixes for bond / how we use bond (use a shared pointer). These are probably good ways to go that could resolve our issues to move forward quickly, but I think given the issues you report in `rclcpp` below:

>  but this is not immediately released from the rclcpp internal mechanisms (especially intraprocess).
> As a result the bondStatusCB function can called after it has been freed.  This use after free shows up reliably with asan when running the test_bond test.
> Figure out an improvement to rclcpp intraprocess subscription mechanism to check for memory lifetime of passed in callback function.

It sounds like your confirming a bug in `rclcpp` for IPC handling that we should detail in the rclcpp ticket for a more general resolution https://github.com/ros2/rclcpp/issues/2721. We shouldn't be having IPC callbacks triggered after the object containing them is destroyed. Its common to have helper objects composed into your node that have sub/pubs for various path, parameter, visualization, etc handling. Can you update with some detailed results that you found in that ticket for Alberto and Tomoya if there are potential action items for rclcpp IPC handling fixes that this uncovered (if any)?

[1] https://github.com/ros/bond_core/blob/ros2/bondcpp/src/bond.cpp#L163

SteveMacenski:
> Consider creating the bond_ as a shared_ptr rather than a unique_ptr to allow the internal mechanisms of rclcpp to hold onto valid memory until they have a chance to cleanup properly.

By the way, I tried this, it didn't work. The rclcpp maintainers pointed to something they think is related / the cause in this comment https://github.com/ros2/rclcpp/issues/2678#issuecomment-2583472585

ewak:
See https://github.com/ros/bond_core/pull/108 for changes that use enable_shared_from_this to ensure valid lifetime of the member function subscription callback.  I also commented on https://github.com/ros2/rclcpp/issues/2678#issuecomment-2585146308 suggesting that a helper function like createSafeSubscriptionMemFuncCallback() could be added to the rclcpp namespace.  It could have an intermediate home in nav2_utils too.

SteveMacenski:
I'm trying to understand better here the set of options from these PRs / issue tickets. 

1. Here in Nav2, I'm seeing that the issue is resolved by only `breakBond()` rather than destroying the object due to the rclcpp issue. So I'm a little unclear why the `shared_ptr` was migrated, unless the `shared_ptr` + the bondcpp PR resolves the issue. tl;dr: don't we need one or the other, for this limited scope problem?

2. In bondcpp, I see how that could resolve the topic within the thread from `rclcpp`, but wouldn't it be best to resolve it at the `rclcpp` level, since this only fixes bond? 

3. In rclcpp, Why not have something like `createSafeSubscriptionMemFuncCallback` inside of rclcpp that is always used in `create_subscription` and/or adding `stop_callbacks` to `rclcpp::Subscription`'s destructor so that resetting a subscription's shared pointer automatically handles this? I don't see that suggestion considered in the threads, which seems like the easiest, low-impact solution.

ewak:
I'll have a go at explaining what I currently think for each of your points,  

>     1. Here in Nav2, I'm seeing that the issue is resolved by only `breakBond()` rather than destroying the object due to the rclcpp issue. So I'm a little unclear why the `shared_ptr` was migrated, unless the `shared_ptr` + the bondcpp PR resolves the issue. tl;dr: don't we need one or the other, for this limited scope problem?

You suggested earlier, https://github.com/ros-navigation/navigation2/pull/4841#issuecomment-2583869295, and I was concerted that breakBond() is not enough as you also wanted the timers to be stopped. So I reverted that change.

Then for any of the the weak_ptr expiry checking mechanism to work the bond must be created(in nav2) as a shared_ptr(not a unique_ptr) to allow use of shared_from_this().  
 
>     2. In bondcpp, I see how that could resolve the topic within the thread from `rclcpp`, but wouldn't it be best to resolve it at the `rclcpp` level, since this only fixes bond? I assume we're on the same page about that

I agree that it would be best to resolve the issue in rclcpp. With that in mind I explored a way that I thought might do that in a rclcpp PR https://github.com/ros2/rclcpp/pull/2725.  This has prompted some discussion but I think the actual changes I made are likely to be rejected in favor of a `stop_callbacks` mechanism.

>     3. In rclcpp, Why not have something like `createSafeSubscriptionMemFuncCallback` inside of rclcpp that is always used in `create_subscription` and/or adding `stop_callbacks` to `rclcpp::Subscription`'s destructor so that resetting a subscription's shared pointer automatically handles this? I don't see that suggestion considered in the threads, which seems like the easiest, low-impact solution. 

I think the argument against something like rclcpp::createSafeSubscriptionMemFuncCallback() is the onus is put on the user to use it.

IOW, The create_subscription API does not take an argument that can be used as a lifetime pointer to capture with the callback function in an internally constructed lambda. So I don't think it can be done within the current create_subscription API call.

I proposed and coded up an addition of a callback_lifetime weak_ptr to subscription options as an alternative.  It works(tm) within rclcpp but still puts onus on the user to use it. https://github.com/ros2/rclcpp/pull/2725

I am still trying to understand the multi-threaded executor reasoning behind the proposed `subscription.stop_callbacks` mechanism.  I still think bondcpp would need to call it before calling reset() on its shared_ptr to the subscriber.  It's not enough to put it in the `rclcpp::Subscription`'s destructor because that won't fire until all instances of it are released. For instance, Here is a place where the WaitSet can hold a subscription shared_ptr https://github.com/ros2/rclcpp/blob/80768ed14e5ebbeb4fbbff6de42149c9c526409d/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp#L205

> 3.1 Michael's suggestion of demoting everything, everywhere to `weak_ptr` would probably be best, but I don't think much functionally different.

I *think* the demotion of weak_ptr of everything everywhere within rclcpp ONLY handles the actual subscription entity itself.  There is currently not enough information provided in the create_subscription API to handle the lifetime of the registered callback function. (Now just repeating myself I think)

Hope the above helps rather than hinders.

SteveMacenski:
Got it, thank you. I understood about 80% of that but enough to know that you and the folks in that thread know what's going on to an appreciably higher degree than myself. I agree that not putting the onus on the user is _very important_ since that is a huge quality of use / life issue for ROS 2 users (and subtle to explain when it matters).  

> I am still trying to understand the multi-threaded executor reasoning behind the proposed subscription.stop_callbacks mechanism. I still think bondcpp would need to call it before calling reset() on its shared_ptr to the subscriber. It's not enough to put it in the rclcpp::Subscription's destructor because that won't fire until all instances of it are released. For instance, Here is a place where the WaitSet can hold a subscription shared_ptr https://github.com/ros2/rclcpp/blob/80768ed14e5ebbeb4fbbff6de42149c9c526409d/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp#L205

Wouldn't Michael's comment on demoting everything to weak pointers though resolve that issue?

SteveMacenski:
I tried again without avail - still shows virtually every system test failing using this branch + the bond PR

ewak:
@SteveMacenski looking for a sanity check. 

Do you expect a single threaded executor that has a single simple/plain lifecycle node, i.e does nothing but log its various state changes, to exit from its spin() (without requiring Ctrl+C) once the LifecycleNode has been transitioned to the shutdown state?  Eg `ros2 lifecycle set /my_plain_lifecycle_node shutdown' 

It's not doing that for me on rolling right now. Irregardless of use_intraprocess.

To my reading something has to call rclcpp::shutdown() to break the executor out of its spin()

Where I am going with this is...
Should an on_broken() callback be registered to the bond to help actually stop the executor from spin()ing, call rclcpp::shutdown() And hence actually stop and shut down the system tests rather than waiting for them to timeout.

https://github.com/ros/bond_core/blob/dbb995761b4fa81ed6b7bb87c4806e024702c531/bondcpp/include/bondcpp/bond.hpp#L98

SteveMacenski:
> Do you expect a single threaded executor that has a single simple/plain lifecycle node, i.e does nothing but log its various state changes, to exit from its spin() (without requiring Ctrl+C) once the LifecycleNode has been transitioned to the shutdown state? Eg `ros2 lifecycle set /my_plain_lifecycle_node shutdown'

No, the process doesn't end, it it just ready to then be terminated & stops doing work. One can transition up a lifecycle node back up from shutdown to active again (which is a pretty common application). If you do `finalize` though, I'm not sure if that terminates the node, I think not, but that is a final state that cannot be re-transitioned up. Finalized is done-done. 

>  Should an on_broken() callback be registered to the bond to help actually stop the executor from spin()ing, call rclcpp::shutdown() And hence actually stop and shut down the system tests rather than waiting for them to timeout.

I don't think so for the reasons above :-) 

`rclcpp::shutdown` is typically called in the `main()` function after `rclcpp::spin(node);`, to run once its interrupted. But either way, the launch system or system event handler should be the one processing SIGs for shutting down the process. I don't think that's relate to our issue

ewak:
So I went deep on launch_testing (https://github.com/ros2/launch/blob/rolling/launch_testing/launch_testing/legacy/__init__.py) and found out how tests shutdown things.  Long story short they wrap event handling around the test process ie `RegisterEventHandler(OnProcessExit( ... EmitEvent(event=Shutdown(... etc` so me trying to replicate things with ros2 run is not a true representation of how the tests run.

But I have found another fix to bondcpp that allows the tests to run for me. See https://github.com/ros/bond_core/pull/108

SteveMacenski:
@ewak if you uncomment the bond line in our underlay repos file and insert your branch, we can test this in CI as well automatically 

https://github.com/ros-navigation/navigation2/blob/main/tools/underlay.repos#L14

SteveMacenski:
CI seems to pass now with shutdown on all the system tests! I'm going to re-run again 1-2x more just to make sure its consistent, but that's a big step forward.

On the docking changes, mind doing that on the `main` branch as well? We can merge that in and fix that flaky test while we're pending in rclcpp's IPC (which might be some time)

ewak:
> CI seems to pass now with shutdown on all the system tests! I'm going to re-run again 1-2x more just to make sure its consistent, but that's a big step forward.

Woot. I wanted to get a clean build in CI before responding about where to go next on the actual changes to push for in bondcpp and rclcpp.  
 
> On the docking changes, mind doing that on the `main` branch as well? We can merge that in and fix that flaky test while we're pending in rclcpp's IPC (which might be some time)

Yes I will make a clean PR for the docking test flakiness once I am sure its working.

SteveMacenski:
It passed again - trying once more!

ewak:
> > On the docking changes, mind doing that on the `main` branch as well? We can merge that in and fix that flaky test while we're pending in rclcpp's IPC (which might be some time)
> 
> Yes I will make a clean PR for the docking test flakiness once I am sure its working.

See https://github.com/ros-navigation/navigation2/pull/4879

SteveMacenski:
> Woot. I wanted to get a clean build in CI before responding about where to go next on the actual changes to push for in bondcpp and rclcpp.

Let me know how I can help :fire:

mini-1235:
@ewak Thanks for your effort on this! It looks like the issue has already been resolved upstream, and I've opened a PR here https://github.com/ros-navigation/navigation2/pull/5804

hu1097007701:
I reproduced a very similar shutdown race locally under TSan on Navigation2 commit `d8bce5e2f57ae276390d83d8b7697b33198397b9` (ROS 2 Jazzy, Ubuntu 24.04.4, `rmw_fastrtps_cpp`).

The repeated signature is:

- write side:
  - `nav2_util::LifecycleNode::destroyBond()`
  - `bond_.reset()`
  - `operator delete(void*)`
- concurrent side:
  - `bond::Bond::flushPendingCallbacks()`
  - `pthread_mutex_trylock`

Representative local TSan lines were:

- `nav2_util::LifecycleNode::destroyBond()` at `nav2_util/src/lifecycle_node.cpp:152`
- `bond::Bond::flushPendingCallbacks()`

I reproduced it again by replaying a saved Nav2 bringup run that drives the stack into lifecycle reset/shutdown; multiple nodes printed `Destroying bond (...) to lifecycle manager.` before the TSan report.

Important context: my local branch does **not** appear to include the later merge from `#5804`, so this may be confirming an older shutdown path rather than proving a current-`main` regression.

I’m sharing this in case it is still useful for tracking the original race family. If needed, I can also provide the preserved TSan log snippet.

SteveMacenski:
What would be your suggested solution? I can check on main then if this is resolved or not :-) Jazzy is a hair old, so it wouldn't surprise me if this was patched already on Lyrical and rolling

Issue #4808, opened 2024-12-19
[SMAC planner] inflation cost not respected on narrow corridor 


BriceRenaudeau:
## Bug report

The path computed by smac planner can go in "inscribed" areas and is not pushed toward middle of free space if the corridor is a bit narrow. When the corridor is a bit wider (2nd image), the path is correct:

![zoom_narrow_path_smoother_not_apply](https://github.com/user-attachments/assets/5afa032a-25b3-4b35-8958-184fd5946382)

![zoom_not_arrow_path_respect_costmap](https://github.com/user-attachments/assets/878bec64-c092-43f6-89b3-969c22f61d5b)

The purple cost is lethal cost and is created using forbidden areas.

- Full path with expansion
![narrow_path_disrespect_costmap](https://github.com/user-attachments/assets/97f5c1d8-43a9-4538-a9e7-bc91e3d376c4)

Another strange behavior is that the smoother is not applied when this bug happens. The blue line is the unsmoothed path, and the green is the smoothed one, In the first image, both paths are the same.

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- ROS2 Version:
  - Iron
- Version or commit hash:
  - iron
- DDS implementation:
  - cyclone

#### Steps to reproduce issue

Create a map with a narrow corridor and plan a path from one side to the other.
It can be reproduced with a smaller test environment and footprint:
![small_map_planner_wrong_path](https://github.com/user-attachments/assets/536e0059-4915-4ae5-9da8-687ad553de63)

#### Expected behavior
The path should respect the costmap and be more in the middle of the free space.

#### Actual behavior
The path doesn't respect the costmap and cut in the 'inscribed' cost.

#### Additional information
The planner config file

```yaml
GridBased:
      plugin: "nav2_smac_planner/SmacPlannerHybrid"
      debug_visualizations: true
      downsample_costmap: false # whether or not to downsample the map
      downsampling_factor: 1 # multiplier for the resolution of the costmap layer (e.g. 2 on a 5cm costmap would be 10cm)
      tolerance: 0.25 # dist-to-goal heuristic cost (distance) for valid tolerance endpoints if exact goal cannot be found.
      allow_unknown: false # allow traveling in unknown space
      max_iterations: -1 # maximum total iterations to search for before failing (in case unreachable), set to -1 to disable
      max_on_approach_iterations: 1000 # Maximum number of iterations after within tolerances to continue to try to find exact solution
[...]

PR #4848, opened 2025-01-14
[Smac Planner] expansion uses cell cost

BriceRenaudeau:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   |  #4808  |
| Primary OS tested on | ubuntu |
| Robotic platform tested on | our robot |

---

* Don't update `footprint_cost_` in `inCollision()` method to keep it with costmap_->getCost() value
* Don't use `adjustedFootprintCost(cost)` in heuristic

This allows `_cell_cost = collision_checker->getCost();` to keep the real cell cost and not the footprint cost.

- Maybe change the name `footprint_cost_` as its not the footprint cost anymore.

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/src/collision_checker.cpp
+++ b/nav2_smac_planner/src/collision_checker.cpp
@@ -151,14 +151,14 @@ bool GridCollisionChecker::inCollision(
       current_footprint.push_back(new_pt);
     }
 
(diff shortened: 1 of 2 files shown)


SteveMacenski:
Thanks, I'll take it from here now :-)

PR #4854, opened 2025-01-15
Fix goal pose stamp

redvinaa:
Follow-up on #4822

* There was a bug in the previous PR: the MPPI critics goal pose now uses stamp from global_path, which gets outdated by the TF buffer, and then fails to transform the goal pose.
* This doesn't come up if the plan is replanned frequently.
* As a solution, this PR uses the `robot_pose.stamp`, which gets updated frequently.

* None required

* In a custom simulation and on real robot, on the `humble` branch.

---

#### For Maintainers: 


SteveMacenski (approved):
I think this needs to be manually backported for Humble's version. I can do Jazzy/Rolling

PR #4864, opened 2025-01-17
Add acc limit consideration to avoid overshooting in RotationShimController

RBT22:
---

## Basic Info

| ------ | ----------- |

---

Currently we don't do the deceleration as a function of its target orientation with RotationShimController. This PR implements this behavior using the `max_angular_accel` param. 

The parameter description has to be updated.

---

#### For Maintainers: 

### DIFF
--- a/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
+++ b/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
@@ -300,6 +300,12 @@ RotationShimController::computeRotateToHeadingCommand(
   cmd_vel.twist.angular.z =
     std::clamp(angular_vel, min_feasible_angular_speed, max_feasible_angular_speed);
 
+  // Check if we need to slow down to avoid overshooting
+  double max_vel_to_stop = std::sqrt(2 * max_angular_accel_ * fabs(angular_distance_to_heading));
+  if (fabs(cmd_vel.twist.angular.z) > max_vel_to_stop) {
+    cmd_vel.twist.angular.z = sign * max_vel_to_stop;
+  }
+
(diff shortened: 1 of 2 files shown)


RBT22:
@SteveMacenski Let me know if you would introduce a separate deceleration limit parameter.

SteveMacenski (changes requested):
i'm curious - did you test this with a small value for `angular_disengage_threshold`? 

If this worked as expected, I would expect the behavior to be a:
* Rotate accelerating to full rotational speed 
* Rotating decelerating to the angle of the path, once in deceleration area 
* Stop relatively close to the angle, with a small disengagement threshold (20 deg or less?) 
* Then controller starts moving

If so, that would make a really nice behavior - very precise, very 'robotic'. A gif to put in the migration guide for this would be splendid to show it at work and the value of it. My hope was that this would be used to create very dynamic hand offs with the controllers, but I think experience has told me now from feedback that this isn't what people want (and most of the controllers don't take kindly to being started in a high-acceleration-turn when starting path tracking)

> Let me know if you would introduce a separate deceleration limit parameter or if a min_angular_speed param is needed here.

Nah, I think this can be symmetric. If someone has a problem with that, they can have a follow up PR. A bunch of controllers for rotating to heading do this, so this is consistent behavior

SteveMacenski:
> One thing to note however is that we moved the shim to open-loop to integrate it seamlessly with our setup. This approach worked well for our case, but I'm curious if this is something you'd consider supporting optionally.

Open in another PR and I can take a look. At face value, I think its a good idea as a parameterized option. 

---

I think the only thing left here is the migration guide entry + gif ideally and I can merge the pair, thanks for looking at this!

SteveMacenski (approved):

Issue #4754, opened 2024-11-18
[nav2_rviz_plugins/Selector] Unable to set goal pose if we run rviz2 later


evshary:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 24.04
- ROS2 Version:
  - Jazzy / Rolling
- Version or commit hash:
  - f426bdaadaa1a124bea0da5f32961cef6d679ae2 (Rolling latest commit)
  - 3140a6a65c2ecbcf975dac94e3f3c64c0d9efc84 (Jazzy latest commit)
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue

* Run the simulation first

```bash
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False use_rviz:=False
```

* Run the rviz2 in another terminal

```bash
ros2 launch nav2_bringup rviz_launch.py
```

#### Expected behavior

The result should do nothing with the execution order.

#### Actual behavior

The rviz is stuck for a while.

![image](https://github.com/user-attachments/assets/c1e26039-9684-4c21-83e0-6aff2acf70d0)

After waiting some time, we can set the initial pose, but the global costmap doesn't work properly.

![image](https://github.com/user-attachments/assets/52dfc2b8-658f-452f-b5ea-f025b5ec0068)

#### Additional information

There is no any issue if we run rviz and nav2 together.
It also works well if we run rviz first and then nav2.

Here is the log in nav2:

```
[component_container_isolated-7] [WARN] [1731909745.646348226] [amcl]: AMCL cannot publish a pose or update the transform. Please set the initial pose...
[component_container_isolated-7] [INFO] [1731909745.785165110] [global_costmap.global_costmap]: Timed out waiting for transform from base_link to map to become available, tf error: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
[component_container_isolated-7] [INFO] [1731909746.285144231] [global_costmap.global_costmap]: Timed out waiting for transform from base_link to map to become available, tf error: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
[component_container_isolated-7] [INFO] [1731909746.785142621] [global_costmap.global_costmap]: Timed out waiting for transform from base_link to map to become available, tf error: Invalid frame ID "map" passed to canTransform argument target_frame - frame does not exist
[...]

PR #4871, opened 2025-01-23
Fix rviz crash issue if run later

mini-1235:
---

## Basic Info

| Info                                        | Please fill out this column |
| ------------------------------------------- | --------------------------- |
| Ticket(s) this addresses                    | #4754                       |
| Primary OS tested on                        | Ubuntu 24.04                |
| Robotic platform tested on                  | turtlebot 3 simulation      |
| Does this PR contain AI generated software? | No                          |

---

On my machine, this not only stuck in getting parameter of planner, but also other plugins like controller, goal checker and so on. Thus, I did the following to solve this issue:
- Open a new thread to load plugins, to ensure that `get_parameter()` would not block the main thread
- Consequently, this new thread can also handle the case we start rviz first, and bringup later

None

Extensively tested on tb3 simulation

---

On the original implementation, the combo box will appear blank if it cannot get parameter from service, I now put it a "default" on those labels, I think this will be better (otherwise users will click on an empty label). 

I am not really familiar with QT, not sure if there is any problem in my implementation :)

Currently only tested on Jazzy, but I could build new images and test it on different distributions

#### For Maintainers: 

### DIFF
--- a/nav2_rviz_plugins/src/selector.cpp
+++ b/nav2_rviz_plugins/src/selector.cpp
@@ -69,8 +69,15 @@ Selector::Selector(QWidget * parent)
   main_layout_->addLayout(row_3_label_layout_);
   main_layout_->addLayout(row_3_layout_);
 
+  controller_->addItem("Default");
+  planner_->addItem("Default");
+  goal_checker_->addItem("Default");
+  smoother_->addItem("Default");
+  progress_checker_->addItem("Default");
+
+
   setLayout(main_layout_);
-  timer_.start(200, this);
+  loadPlugins();
 
   connect(
     controller_, QOverload<int>::of(&QComboBox::activated), this,
@@ -95,16 +102,15 @@ Selector::Selector(QWidget * parent)
 
 Selector::~Selector()
 {
+  if (load_plugins_thread_.joinable()) {
+    load_plugins_thread_.join();
+  }
 }
 
 // Publish the selected controller or planner
 void Selector::setSelection(
   QComboBox * combo_box, rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher)
 {
-  // If "default" option is selected, it gets removed and the next item is selected
-  if (combo_box->findText("Default") != -1) {
-    combo_box->removeItem(0);
-  }
 
   // if there are no plugins available, return
   if (combo_box->count() == 0) {
@@ -115,7 +121,6 @@ void Selector::setSelection(
   msg.data = combo_box->currentText().toStdString();
 
   publisher->publish(msg);
-  timer_.start(200, this);
 }
 
 // Call setSelection() for controller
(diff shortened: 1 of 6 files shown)


mini-1235:
Should I close this and re-open one targeting branch main

nav2_rviz_plugins/src/selector.cpp
```diff
@@ -148,37 +153,21 @@ void Selector::setProgressChecker()
...
+      client_node_, "controller_server", "controller_plugins", controller_);
+    nav2_rviz_plugins::pluginLoader(
+      client_node_, "planner_server", "planner_plugins", planner_);
+    nav2_rviz_plugins::pluginLoader(
+      client_node_, "controller_server", "goal_checker_plugins", goal_checker_);
+    nav2_rviz_plugins::pluginLoader(
+      client_node_, "smoother_server", "smoother_plugins", smoother_);
+    nav2_rviz_plugins::pluginLoader(
+      client_node_, "controller_server", "progress_checker_plugins",
+      progress_checker_);
```
SteveMacenski, line 169: It would be nice to track and log the failures here so that we can print something that it failed to connect. I think the `server_failed_` should be maintained

nav2_rviz_plugins/src/utils.cpp
```diff
@@ -21,35 +21,19 @@ namespace nav2_rviz_plugins
 {
 
 void pluginLoader(
-  rclcpp::Node::SharedPtr node, bool & server_failed, const std::string & server_name,
+  rclcpp::Node::SharedPtr node, const std::string & server_name,
   const std::string & plugin_type, QComboBox * combo_box)
 {
   auto parameter_client = std::make_shared<rclcpp::SyncParametersClient>(node, server_name);
 
-  // Do not load the plugins if the combo box is already populated
```
SteveMacenski, line 29: Why remove this?

nav2_rviz_plugins/src/utils.cpp
```diff
@@ -21,35 +21,19 @@ namespace nav2_rviz_plugins
...
-  }
-
   // Wait for the service to be available before calling it
-  bool server_unavailable = false;
   while (!parameter_client->wait_for_service(std::chrono::seconds(1))) {
     if (!rclcpp::ok()) {
       RCLCPP_ERROR(node->get_logger(), "Interrupted while waiting for the service. Exiting.");
       rclcpp::shutdown();
     }
-    RCLCPP_INFO(node->get_logger(), "%s service not available", server_name.c_str());
```
SteveMacenski, line 41: Why remove? If this failed, shouldn't we log and notify that the server is unavailable to not try to call the client below it? 

I tseems like the server_unavailable should be maintained.

SteveMacenski (changes requested):
Yes, please close this PR on Jazzy and apply to `main`, I'll backport the main PR to jazzy once complete and reviews are accepted :-)

Issue #4756, opened 2024-11-19
Feature Request:Make the isPathValid Service Configurable with Parameters


BCKSELFDRIVEWORLD:
#### Feature description
Currently, the isPathValid service considers a path valid as long as the costs are not lethal. However, in many scenarios, users might need more flexibility in determining what constitutes a valid path. For instance, there might be cases where paths with costs above a certain threshold (but below lethal cost) should also be considered invalid.

To address this, I propose introducing a parameter that allows users to specify the threshold cost for path validity. This would make the service more adaptable to different use cases and improve its overall utility.

#### Proposed Feature:

- Add a parameter to the isPathValid service that allows users to define the maximum allowable cost for a path to be considered valid.

SteveMacenski:
Seems reasonable to have a max cost parameter or in the service request itself. I think the service request having a field for the max, valid cost would be great & pretty easy contribution. @BCKSELFDRIVEWORLD can you submit a PR?

Youlun-Peng:
Hey, @SteveMacenski ! Is there any progress here? Can I contribute to address this issue? :D

BCKSELFDRIVEWORLD:
İ will sent pr this week

mini-1235:
> i will sent pr this week

Sorry, I just saw this reply after I made the PR 🥲

BCKSELFDRIVEWORLD:
If you haven’t seen it, no problem. Congratulations! I’m sorry for keeping you @SteveMacenski waiting this long; I was busy and forgot. :/

CesarBear:
@SteveMacenski I am open to contributing something else. Is there any issue suitable for my first trial help?

SteveMacenski:
@CesarBear please check out tickets with help wanted, enhancement, or good first issue!

* https://github.com/ros-navigation/navigation2/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22
* https://github.com/ros-navigation/navigation2/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22
* https://github.com/ros-navigation/navigation2/issues?q=is%3Aissue%20state%3Aopen%20label%3Aenhancement

In particular, one of these might be good!:
* https://github.com/ros-navigation/navigation2/issues/4403
* https://github.com/ros-navigation/navigation2/issues/4695
* https://github.com/ros-navigation/navigation2/issues/4794
* https://github.com/ros-navigation/navigation2/issues/4769
[...]

PR #4873, opened 2025-01-25
Add ispathvalid maxcost

mini-1235:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4756 |
| Robotic platform tested on | Custom simulation environment |

---

- Add max cost as described in the issue 
- Add corresponding test

Add new parameter description https://docs.nav2.org/configuration/packages/bt-plugins/conditions/IsPathValid.html

Wrote a new test to test this and also tested on custom simulation env

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
@@ -64,7 +64,7 @@ public:
     return {
       BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
       BT::InputPort<std::chrono::milliseconds>("server_timeout"),
-      BT::InputPort<uint8_t>("max_cost", 255, "Maximum cost of the path")
+      BT::InputPort<unsigned int>("max_cost", 255, "Maximum cost of the path")
     };
   }
 
@@ -74,7 +74,7 @@ private:
   // The timeout value while waiting for a response from the
   // is path valid service
   std::chrono::milliseconds server_timeout_;
-  uint8_t max_cost_;
+  unsigned int max_cost_;
 };
 
 }  // namespace nav2_behavior_tree
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
@@ -63,7 +63,8 @@ public:
   {
     return {
       BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
-      BT::InputPort<std::chrono::milliseconds>("server_timeout")
+      BT::InputPort<std::chrono::milliseconds>("server_timeout"),
+      BT::InputPort<uint8_t>("max_cost", 255, "Maximum cost of the path")
     };
   }
 
@@ -73,6 +74,7 @@ private:
   // The timeout value while waiting for a response from the
   // is path valid service
   std::chrono::milliseconds server_timeout_;
+  uint8_t max_cost_;
 };
 
 }  // namespace nav2_behavior_tree
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
@@ -64,7 +64,10 @@ public:
     return {
       BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
       BT::InputPort<std::chrono::milliseconds>("server_timeout"),
-      BT::InputPort<unsigned int>("max_cost", 255, "Maximum cost of the path")
+      BT::InputPort<unsigned int>("max_cost", 253, "Maximum cost of the path"),
+      BT::InputPort<bool>(
+          "consider_unknown_as_obstacle", false,
+          "Whether to consider unknown cost as obstacle")
     };
   }
 
@@ -75,6 +78,7 @@ private:
   // is path valid service
   std::chrono::milliseconds server_timeout_;
   unsigned int max_cost_;
+  bool consider_unknown_as_obstacle_;
 };
 
 }  // namespace nav2_behavior_tree
--- a/nav2_behavior_tree/nav2_tree_nodes.xml
+++ b/nav2_behavior_tree/nav2_tree_nodes.xml
@@ -283,6 +283,7 @@
     <Condition ID="IsPathValid">
       <input_port name="path"> Path to validate </input_port>
       <input_port name="server_timeout"> Server timeout </input_port>
+      <input_port name="max_cost"> Maximum cost of the path </input_port>
     </Condition>
 
     <Condition ID="WouldAControllerRecoveryHelp">
--- a/nav2_behavior_tree/nav2_tree_nodes.xml
+++ b/nav2_behavior_tree/nav2_tree_nodes.xml
@@ -107,6 +107,7 @@
       <input_port name="input_goals">A vector of goals to check if in collision</input_port>
       <input_port name="use_footprint">Whether to use the footprint cost or the point cost.</input_port>
       <input_port name="cost_threshold">The cost threshold above which a waypoint is considered in collision and should be removed.</input_port>
+      <input_port name="consider_unknown_as_obstacle"> If unknown cost is considered valid </input_port>
       <output_port name="output_goals">A vector of goals containing only those that are not in collision.</output_port>
     </Action>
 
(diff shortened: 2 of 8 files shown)


nav2_msgs/srv/IsPathValid.srv
```diff
@@ -1,6 +1,7 @@
 #Determine if the current path is still valid
 
 nav_msgs/Path path
+uint8 max_cost
```
SteveMacenski, line 4: You can set the default value to `254`, so that if not populated, the behavior doesn't change. 

```
fieldtype fieldname fielddefaultvalue
```

nav2_behavior_tree/nav2_tree_nodes.xml
```diff
@@ -283,6 +283,7 @@
     <Condition ID="IsPathValid">
       <input_port name="path"> Path to validate </input_port>
       <input_port name="server_timeout"> Server timeout </input_port>
+      <input_port name="max_cost"> Maximum cost of the path </input_port>
```
SteveMacenski, line 286: We need a docs PR to update the configuration guide to include this new port as well :-)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
```diff
@@ -63,7 +63,8 @@ class IsPathValidCondition : public BT::ConditionNode
   {
     return {
       BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
-      BT::InputPort<std::chrono::milliseconds>("server_timeout")
+      BT::InputPort<std::chrono::milliseconds>("server_timeout"),
+      BT::InputPort<uint8_t>("max_cost", 255, "Maximum cost of the path")
```
SteveMacenski, line 67:
```suggestion
      BT::InputPort<unsigned int>("max_cost", 255, "Maximum cost of the path")
```

Some of the different architectures don't like `uint8_t`, so using base types can be helpful for compiling on a variety of hardware. Please  change this here, in the variable definition in L77, and anywhere else used

nav2_planner/src/planner_server.cpp
```diff
@@ -705,7 +705,7 @@ void PlannerServer::isPathValid(
       {
         response->is_valid = false;
         break;
-      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE) {
+      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE || cost > request->max_cost) {
```
SteveMacenski, line 708: I think it should be in the `use_radius` condition as well, this seems related to both SE2 footprint and radius checks

SteveMacenski (changes requested):
Overall, looks very good to me, just small ~5 minute worth of nitpicks and we can merge this!

mini-1235:
Hi, thanks for reviewing @SteveMacenski, I have fixed the problem you have mentioned above, can you please take another look on it? Also the mppi test failed for some reason, but it passed on my machine, can you help on it 😄

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/is_path_valid_condition.hpp
```diff
@@ -63,7 +63,8 @@ class IsPathValidCondition : public BT::ConditionNode
   {
     return {
       BT::InputPort<nav_msgs::msg::Path>("path", "Path to Check"),
-      BT::InputPort<std::chrono::milliseconds>("server_timeout")
+      BT::InputPort<std::chrono::milliseconds>("server_timeout"),
+      BT::InputPort<unsigned int>("max_cost", 255, "Maximum cost of the path")
```
SteveMacenski, line 67:
```suggestion
      BT::InputPort<unsigned int>("max_cost", 253, "Maximum cost of the path")
```

nav2_behavior_tree/plugins/condition/is_path_valid_condition.cpp
```diff
@@ -23,7 +23,8 @@ namespace nav2_behavior_tree
 IsPathValidCondition::IsPathValidCondition(
   const std::string & condition_name,
   const BT::NodeConfiguration & conf)
-: BT::ConditionNode(condition_name, conf)
+: BT::ConditionNode(condition_name, conf),
+  max_cost_(255)
```
SteveMacenski, line 27:
```suggestion
  max_cost_(253)
```

nav2_planner/src/planner_server.cpp
```diff
@@ -700,12 +700,12 @@ void PlannerServer::isPathValid(
       }
 
       if (use_radius &&
-        (cost == nav2_costmap_2d::LETHAL_OBSTACLE ||
+        (cost > request->max_cost || cost == nav2_costmap_2d::LETHAL_OBSTACLE ||
```
SteveMacenski, line 703:
```suggestion
        (cost >= request->max_cost || cost == nav2_costmap_2d::LETHAL_OBSTACLE ||
```

nav2_planner/src/planner_server.cpp
```diff
@@ -700,12 +700,12 @@ void PlannerServer::isPathValid(
...
         cost == nav2_costmap_2d::INSCRIBED_INFLATED_OBSTACLE))
       {
         response->is_valid = false;
         break;
-      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE) {
+      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE || cost > request->max_cost) {
```
SteveMacenski, line 708:
```suggestion
      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE || cost >= request->max_cost) {
```

nav2_system_tests/src/planning/test_planner_is_path_valid.cpp
```diff
@@ -32,9 +32,10 @@ TEST(testIsPathValid, testIsPathValid)
   planner_tester->loadSimpleCostmap(TestCostmap::top_left_obstacle);
 
   nav_msgs::msg::Path path;
+  unsigned int max_cost = 255;
```
SteveMacenski, line 35:
```suggestion
  unsigned int max_cost = 253;
```

SteveMacenski (changes requested):
Small changes for setting the limit to be inscribed inflated by default, but otherwise LGTM!

nav2_planner/src/planner_server.cpp
mini-1235: I think we treat `NO_INFORMATION(255)` as `FREE_SPACE` here, so I should convert 255 to 0 here / do a simple `&& cost!=FREE_SAPCE` here, am I correct?

SteveMacenski: This is a good point - I think if we add in max_cost, we should also add in a bool service argument for `unknown_valid` and we check if the cost is larger than the max cost and if unknown should be considered valid or invalid (collision).

nav2_planner/src/planner_server.cpp
```diff
@@ -699,13 +699,19 @@ void PlannerServer::isPathValid(
             position.x, position.y, theta, footprint));
       }
 
+      if (cost == nav2_costmap_2d::NO_INFORMATION && request->consider_unknown_as_obstacle) {
+        cost = nav2_costmap_2d::LETHAL_OBSTACLE;
+      } else if(cost == nav2_costmap_2d::NO_INFORMATION) {
```
SteveMacenski, line 704:
```suggestion
      } else if (cost == nav2_costmap_2d::NO_INFORMATION) {
```

SteveMacenski (approved):

SteveMacenski:
One sec - lets let CI turn over again, a bunch of weird unusual tests failed and I want to make sure it wasn't just a fluke. Retriggering

nav2_planner/src/planner_server.cpp
```diff
@@ -699,13 +699,19 @@ void PlannerServer::isPathValid(
...
+
       if (use_radius &&
-        (cost == nav2_costmap_2d::LETHAL_OBSTACLE ||
+        (cost >= request->max_cost || cost == nav2_costmap_2d::LETHAL_OBSTACLE ||
         cost == nav2_costmap_2d::INSCRIBED_INFLATED_OBSTACLE))
       {
         response->is_valid = false;
         break;
-      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE) {
+      } else if (cost == nav2_costmap_2d::LETHAL_OBSTACLE || cost >= request->max_cost) {
```
tonynajjar, line 714: I think this introduced a bug or at the very least an unexpected behavior compared to how it was.

In this clause, use_radius is false so `cost` is a footprintCost. The default ` max_cost` that was set to 253. A footprintCost of 253 is acceptable (a vertex is intersecting a INSCRIBED_INFLATED_OBSTACLE cell) but according to this logic, this renders the path invalid.

@mini-1235 can you double-check my logic? Thank you

SteveMacenski: This is true, but that's why this is a configuration of the request. If you set `use_radius = false` then you should set your `max_cost` limit. 

Do you recommend changing the default to 254? That seems sensible enough to me so it doesn't trigger anything unless the user meaningfully overrides it with another value of their selection.

tonynajjar: > Do you recommend changing the default to 254?

Yes and that would not change anything for the case of where `use_radius` is true because we already check `cost == nav2_costmap_2d::INSCRIBED_INFLATED_OBSTACLE)`

> This is true, but that's why this is a configuration of the request

I just didn't expect a change in behavior in my IsPathValid BT node, at least not without notes.

PR #4877, opened 2025-01-29
Add acceleration limits to DriveOnHeading and BackUp behaviors (backport #4810)

RBT22:
Backport of #4810 to humble

---------

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


SteveMacenski (commented):
In order to backport this, we need to make some changes:

- The acceleration / deceleration need to be defaulted to `0` and a new "off" condition is checked for the `0` values. We don't want behavior to instantly change for users updating on Humble due to acceleration limits being introduced which may or may not be sensible for their vehicle 
- Remove from the parameter file so that the default current value persists 
- Update the parameter value check so that `0` is a valid condition

SteveMacenski (approved):

PR #4880, opened 2025-01-29
Add option to use open-loop control with Rotation Shim

RBT22:
---

## Basic Info

| ------ | ----------- |

---

New parameter description needed and migration guide.

---

#### For Maintainers: 

### DIFF
--- a/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
+++ b/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
@@ -415,6 +415,8 @@ RotationShimController::dynamicParametersCallback(std::vector<rclcpp::Parameter>
         rotate_to_goal_heading_ = parameter.as_bool();
       } else if (name == plugin_name_ + ".rotate_to_heading_once") {
         rotate_to_heading_once_ = parameter.as_bool();
+      } else if (name == plugin_name_ + ".closed_loop") {
+        closed_loop_ = parameter.as_bool();
       }
     }
   }
--- a/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
+++ b/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
@@ -122,6 +122,7 @@ void RotationShimController::activate()
 
   primary_controller_->activate();
   in_rotation_ = false;
+  last_angular_vel_ = std::numeric_limits<double>::max();
 
   auto node = node_.lock();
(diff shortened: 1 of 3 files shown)


nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -73,6 +73,8 @@ void RotationShimController::configure(
     node, plugin_name_ + ".rotate_to_goal_heading", rclcpp::ParameterValue(false));
   nav2_util::declare_parameter_if_not_declared(
     node, plugin_name_ + ".rotate_to_heading_once", rclcpp::ParameterValue(false));
+  nav2_util::declare_parameter_if_not_declared(
+    node, plugin_name_ + ".feedback", rclcpp::ParameterValue(std::string("CLOSED_LOOP")));
```
SteveMacenski, line 77: You could make this simply a bool for `closed_loop = true` default and avoid the string parsing :-) There's only 2 options

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -141,6 +158,8 @@ void RotationShimController::deactivate()
     node->remove_on_set_parameters_callback(dyn_params_handler_.get());
   }
   dyn_params_handler_.reset();
+
+  last_angular_vel_ = std::numeric_limits<double>::max();
```
SteveMacenski, line 162: I think putting into activate makes more sense so that the first initialization also has it reset (I know the header has it set on construction as this as well, but I don't like relying on that kind of behavior since we have alot of hands in the codebase)

SteveMacenski (changes requested):
There are some pretty reasonable blocks of code without coverage, so I think this needs some tests to cover those gaps by enabling the setting and checking for the expected outputs without odometry feedback

Otherwise, LTGM!

SteveMacenski (approved):
Small docs nitpick and I can merge the pair!

SteveMacenski (changes requested):
Whoops, missed one.

If you have a goal then stop navigating for some time, then start navigating again, the state will be stale. 

We should override `reset()` in the controller and reset the value to `max` there as well. The `reset()` was added to the controller API a little while ago for exactly these kinds of between-request state clearing operations 

https://github.com/ros-navigation/navigation2/blob/main/nav2_controller/src/controller_server.cpp#L766

SteveMacenski:
Thanks!

PR #4882, opened 2025-01-29
Fix rviz crash if launch later

mini-1235:
---

## Basic Info

| Info                                        | Please fill out this column |
| ------------------------------------------- | --------------------------- |
| Ticket(s) this addresses                    | #4754                       |
| Primary OS tested on                        | Ubuntu 24.04                |
| Robotic platform tested on                  | turtlebot 3 simulation      |
| Does this PR contain AI generated software? | No                          |

---

Revival of #4871 targeting the main branch with improved code 

- Open a new thread to load plugins, to ensure that `get_parameter()` would not block the main thread. The combo box will now have value if it successfully gets the parameter, otherwise empty

None

Extensively tested on tb3 simulation and all tests passed

---

#### For Maintainers: 

### DIFF
--- a/nav2_rviz_plugins/src/selector.cpp
+++ b/nav2_rviz_plugins/src/selector.cpp
@@ -150,7 +150,7 @@ void
 Selector::loadPlugins()
 {
   load_plugins_thread_ = std::thread([this]() {
-        rclcpp::Rate rate(0.1);
+        rclcpp::Rate rate(0.2);
         while (rclcpp::ok() && !plugins_loaded_) {
           RCLCPP_INFO(client_node_->get_logger(), "Trying to load plugins...");
           nav2_rviz_plugins::pluginLoader(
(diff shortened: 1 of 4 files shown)


mini-1235:
I now add the `combo_box->addItem("Default");` after we have loaded the parameter successfully, but I can move it before loading the plugin if you think it is better

nav2_rviz_plugins/src/selector.cpp
```diff
@@ -148,37 +147,35 @@ void Selector::setProgressChecker()
...
+            client_node_, server_failed_, "controller_server", "progress_checker_plugins",
+            progress_checker_);
+          if (controller_->count() > 0 &&
+          planner_->count() > 0 &&
+          goal_checker_->count() > 0 &&
+          smoother_->count() > 0 &&
+          progress_checker_->count() > 0)
+          {
+            plugins_loaded_ = true;
+          }
```
SteveMacenski, line 175: A log here in an else statement for the fact that it failed and we're going to try again would be nice 

I'd also change the rate to 0.2 to match the previous timer's behavior

SteveMacenski (changes requested):
Otherwise, small thing to update and should be good to go!

SteveMacenski (approved):

PR #4885, opened 2025-01-30
Fix for setting dynamically the footprint_clearing_enabled parameter in the StaticLayer costmap_2d plugin

georgflick:
In the StaticLayer costmap2d_plugin , the parameter `footprint_clearing_enabled` was never changed dynamically in the callback, because the `else if` condition has a duplicated condition (same as in [line 490](https://github.com/ros-navigation/navigation2/blob/dcc5db5eae4566c30dfa886920e43d2acc01af7b/nav2_costmap_2d/plugins/static_layer.cpp#L490) and [line 500](https://github.com/ros-navigation/navigation2/blob/dcc5db5eae4566c30dfa886920e43d2acc01af7b/nav2_costmap_2d/plugins/static_layer.cpp#L500)), so it never executed setting that parameter.

---

## Basic Info

| ------ | ----------- |

---

- Fixes the dynamical parameter change of the parameter `footprint_clearing_enabled` via a ros2 callback
- The `else if` condition was a duplicate of the one before, so it was never executed. This change is fixing it.

Not needed.

`ros2 param set /local_costmap/local_costmap obstacle_layer.footprint_clearing_enabled False` or True to disable/enable it

None

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/static_layer.cpp
+++ b/nav2_costmap_2d/plugins/static_layer.cpp
@@ -496,9 +496,7 @@ StaticLayer::dynamicParametersCallback(
         height_ = size_y_;
         has_updated_data_ = true;
         current_ = false;
-      }
-    } else if (param_type == ParameterType::PARAMETER_BOOL) {
-      if (param_name == name_ + "." + "footprint_clearing_enabled") {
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #4887, opened 2025-01-30
[BtActionNode] [BtServiceNode] clear between calls

doisyg:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   ||
| Primary OS tested on |Ubuntu 24.04|

---

In `BtActionNode` and `BtServiceNode`, the `goal_`, `result_` and `request_` objects were not cleared between 2 calls. Hence keeping an old value if any field of these was not explicitly set by a BT node inheriting `BtActionNode` or `BtServiceNode`.
I realized it while using `ComputePathToPose` and not understanding why the action was using a previous `start` pose when actually making a new call without a `start` parameter (where it is expected that the current pose of the robot will be used):
https://github.com/ros-navigation/navigation2/blob/447c8cec3b528ca271108200567905dc0ba9f005/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp#L33-L38

This PR reinitializes `goal_`, `result_` and `request_` between calls to solve this issue.

* Appropriate behavior of `ComputePathToPose` when calling it successively with and without a valid input `start` 

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp
@@ -194,6 +194,10 @@ public:
       // reset the flag to send the goal or not, allowing the user the option to set it in on_tick
       should_send_goal_ = true;
 
+      // Clear the input and output messages to make sure we have no leftover from previous calls
+      goal_ = typename ActionT::Goal();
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
Nice! I checked and see no problems with it. I thought maybe some node maintained state, but it doesn't appear so

PR #4890, opened 2025-01-31
Refactor/observation buffer

kfabian:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on |  Ubuntu 22.04 |

---

* Changed the Observation class to store the value of the point cloud instead of a ptr to it, which simplified the class because there is no need for user implemented destructor and copy constructor anymore.
* Changed the ObservationBuffer to store a shared_ptrs of the Observations so that no deep copies of the point clouds are made when querying the buffer

-

* Existing tests for costmap
* Added basic unit test for ObservationBuffer

---

* The ObservationBuffer has `lock` and `unlock` member functions. In my opinion this error prone, the ObservationBuffer should either be thread safe on its own (by not exposing the mutex, but locking it in the implementation when needed) or it should not have the mutex member at all and the calling code would be responsible.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/obstacle_layer.cpp
+++ b/nav2_costmap_2d/plugins/obstacle_layer.cpp
@@ -199,15 +199,13 @@ void ObstacleLayer::onInitialize()
 
     // create an observation buffer
     observation_buffers_.push_back(
-      std::shared_ptr<ObservationBuffer
-      >(
-        new ObservationBuffer(
-          node, topic, observation_keep_time, expected_update_rate,
+          std::make_shared<ObservationBuffer>(node, topic, observation_keep_time,
+        expected_update_rate,
           min_obstacle_height,
           max_obstacle_height, obstacle_max_range, obstacle_min_range, raytrace_max_range,
           raytrace_min_range, *tf_,
           global_frame_,
-          sensor_frame, tf2::durationFromSec(transform_tolerance))));
+          sensor_frame, tf2::durationFromSec(transform_tolerance)));
 
     // check if we'll add this buffer to our marking observation buffers
     if (marking) {
@@ -333,7 +331,7 @@ ObstacleLayer::dynamicParametersCallback(
 void
 ObstacleLayer::laserScanCallback(
   sensor_msgs::msg::LaserScan::ConstSharedPtr message,
-  const std::shared_ptr<nav2_costmap_2d::ObservationBuffer> & buffer)
+  const std::shared_ptr<ObservationBuffer> & buffer)
 {
   // project the laser into a point cloud
   sensor_msgs::msg::PointCloud2 cloud;
@@ -367,7 +365,7 @@ ObstacleLayer::laserScanCallback(
 void
 ObstacleLayer::laserScanValidInfCallback(
   sensor_msgs::msg::LaserScan::ConstSharedPtr raw_message,
-  const std::shared_ptr<nav2_costmap_2d::ObservationBuffer> & buffer)
+  const std::shared_ptr<ObservationBuffer> & buffer)
 {
   // Filter positive infinities ("Inf"s) to max_range.
   float epsilon = 0.0001;  // a tenth of a millimeter
@@ -433,7 +431,7 @@ ObstacleLayer::updateBounds(
   useExtraBounds(min_x, min_y, max_x, max_y);
 
   bool current = true;
-  std::vector<Observation> observations, clearing_observations;
+  std::vector<Observation::ConstSharedPtr> observations, clearing_observations;
 
   // get the marking observations
   current = current && getMarkingObservations(observations);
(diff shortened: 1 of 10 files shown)


nav2_costmap_2d/src/observation_buffer.cpp
```diff
@@ -171,40 +168,27 @@ void ObservationBuffer::bufferCloud(const sensor_msgs::msg::PointCloud2 & cloud)
...
-      if ((clock_->now() - obs.cloud_->header.stamp) >
-        observation_keep_time_)
-      {
-        observation_list_.erase(obs_it, observation_list_.end());
-        return;
-      }
+      observation_list_.erase(std::next(observation_list_.begin()), observation_list_.end());
+    } else {
+      // otherwise remove all observations that are older then the keep time
+      observation_list_.remove_if([now = clock_->now(), this](const auto observation){
```
kfabian, line 189: The previous implementation was assuming that the `observation_list_` was ordered, but I think that is not guaranteed.

SteveMacenski:
A couple of quick remarks before a full review:
* Please fix linting :-) 
* Can you highlight some of the run-time impacts of this? I imagine with some of the changes in the PC2 handling semantics this was motivated by performance and I'd like to look at this knowing how this changes things 

This code is 1:1 from ROS 1, so I don't doubt there's performance optimizations to be made and happy to review it. I'd just like to understand what the benefits are while reviewing it beyond shuffling code around

SteveMacenski:
@kfabian any update?

SteveMacenski:
@kfabian let me know if you want me to take a look again

kfabian:
Sorry for the very late reply. Regarding the performance impact, there was no specific performance issue I had. I just stumbled on this and thought it was a low hanging fruit to optimize. I will see if I can add a small benchmark.

SteveMacenski:
@kfabian I see a push which sent me a notification. Is this ready yet?

kfabian:
> @kfabian I see a push which sent me a notification. Is this ready yet?

Yes. I also added a small benchmark. In general I did this just for fun and there is no project where I would need this merged for.

nav2_costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -833,10 +832,8 @@ ObstacleLayer::reset()
 void
 ObstacleLayer::resetBuffersLastUpdated()
 {
-  for (unsigned int i = 0; i < observation_buffers_.size(); ++i) {
-    if (observation_buffers_[i]) {
-      observation_buffers_[i]->resetLastUpdated();
-    }
+  for (const auto & observation_buffer : observation_buffers_) {
+    observation_buffer->resetLastUpdated();
```
SteveMacenski: I think the `if` statement is necessary for some cases

SteveMacenski:
What were the results of that benchmark? Otherwise I don't see anything here that I really object to, this has been open for a long time and keen to help you merge this in before ROSCon :-)

nav2_costmap_2d/plugins/obstacle_layer.cpp
kfabian: Elements are only added to `observation_buffers_` in https://github.com/kfabian/navigation2/blob/6626d3f0c54d419e85c31b0f48091d288dc05a9c/nav2_costmap_2d/plugins/obstacle_layer.cpp#L208. So there could only by a nullptr added in a derived class. But in this case the check would be needed every time an element of one the buffers_ containers is accessed. It would probably make sense to make most of the members private instead.

SteveMacenski: The Voxel Layer is a derived class and I know for instance that there are a few examples of folks deriving from this for custom costmap layers.

kfabian: I added the check back and also added it for `marking_buffers_`and `clearing_buffers_`.

SteveMacenski:
Any notable performance changes? Otherwise LGTM

kfabian:
I change the benchmark to use  an actual obstacle layer so that it is more meaningful. On my machine the speed up is around 1.2x.

kfabian:
> @kfabian, your PR has failed to build. Please check CI outputs and resolve issues. You may need to rebase or pull in `main` due to API changes (or your contribution genuinely fails).

@SteveMacenski Looks like the build hit a timeout.

SteveMacenski:
@kfabian I retriggered. If it happens again, please rebase / pull in main. I recently did some work to make the balance of jobs better which it looks like may be missing from your branch. Sorry for the delay - ROSCon and all :-)

SteveMacenski:
Thanks for the followup here after all this time! Its great to get older PRs buttoned up :-) 

Let me know if you run into any other low-hanging fruit optimizations :wink:

Issue #4893, opened 2025-02-02
Unstable nav2_util test


mini-1235:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 24.04
- Computer:
  - 
- ROS2 Version:
  - Rolling
- Version or commit hash:
  - latest commit [78ea525](https://github.com/ros-navigation/navigation2/commit/78ea525b3b2eb1ce863937fb12f2cbb0abe7821d)
- DDS implementation:
  - rmw_fastrtps_cpp

#### Steps to reproduce issue

running 
```
colcon test 
```
for a multiple times, will produce the following result randomly

```
[ RUN      ] ActionTest.test_simple_action
unknown file: Failure
C++ exception with description "Goal handle is not known to this client." thrown in the test body.

 Teardown
 Teardown...
 Teardown complete
[  FAILED  ] ActionTest.test_simple_action (25 ms)
[ RUN      ] ActionTest.test_simple_action_with_feedback
unknown file: Failure
C++ exception with description "Goal handle is not known to this client." thrown in the test body.

 Teardown
 Teardown...
 Teardown complete
[  FAILED  ] ActionTest.test_simple_action_with_feedback (9 ms)
[ RUN      ] ActionTest.test_simple_action_activation_cycling
unknown file: Failure
C++ exception with description "Goal handle is not known to this client." thrown in the test body.

 Teardown
 Teardown...
 Teardown complete
[  FAILED  ] ActionTest.test_simple_action_activation_cycling (10 ms)
[ RUN      ] ActionTest.test_simple_action_preemption
Sent goal, spinning til complete...
[INFO] [1738492501.832007406] [fibonacci_server_node]: Deactivating
Sent goal, spinning til complete...
unknown file: Failure
C++ exception with description "Goal handle is not known to this client." thrown in the test body.

 Teardown
 Teardown...
 Teardown complete
[  FAILED  ] ActionTest.test_simple_action_preemption (20 ms)
[ RUN      ] ActionTest.test_simple_action_preemption_after_succeeded
unknown file: Failure
C++ exception with description "Goal handle is not known to this client." thrown in the test body.

 Teardown
 Teardown...
 Teardown complete
[  FAILED  ] ActionTest.test_simple_action_preemption_after_succeeded (10 ms)
[ RUN      ] ActionTest.test_handle_goal_deactivated
[INFO] [1738492501.862808930] [fibonacci_server_node]: Ignoring preemptions
[...]

PR #4894, opened 2025-02-02
Fix unstable test in nav2 util

mini-1235:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4893  |
| Primary OS tested on | Ubuntu24.04 |
| Robotic platform tested on | Custom env |

---

Fix unstable test as mentioned in the issue, this is due to the inactiveness of the server that causes the goal handle to be a nullptr, which is throwing the exception of `C++ exception with description "Goal handle is not known to this client.`. Thus, I am adding an `EXPECT_NE` for the goal handle, and add a short delay before we activate the server

None

Running the test for 1000 times independently

---

#### For Maintainers: 

### DIFF
--- a/nav2_util/test/test_actions.cpp
+++ b/nav2_util/test/test_actions.cpp
@@ -258,6 +258,7 @@ protected:
 
 TEST_F(ActionTest, test_simple_action)
 {
+  std::this_thread::sleep_for(std::chrono::milliseconds(5000));
   node_->activate_server();
 
   // The goal for this invocation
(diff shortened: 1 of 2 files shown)


nav2_util/test/test_actions.cpp
```diff
@@ -258,6 +258,7 @@ class ActionTest : public ::testing::Test
 
 TEST_F(ActionTest, test_simple_action)
 {
+  std::this_thread::sleep_for(std::chrono::milliseconds(5000));
```
SteveMacenski, line 261: Why 5s? That seems like a long time (and extends our CI runs by 5s each time as well). Would 1s be suitable?

SteveMacenski (changes requested):

nav2_util/test/test_actions.cpp
mini-1235: Yes, after testing, I agree that 1s would be enough. Sorry for that

SteveMacenski (approved):

PR #4899, opened 2025-02-03
Create footprint_collision_checker.py

ThomasHaley-neya:
Backport of footprint_collision_checker to Nav2 Humble

---

## Basic Info

| Info | A simple backport of the FootprintCollisionChecker API to Nav2 Humble |
| ------ | ----------- |
| Ticket(s) this addresses   | [(4898)](https://github.com/ros-navigation/navigation2/issues/4898) |
| Robotic platform tested on | Custom Gazebo Sim |

---

* Basically just copied FootPrintCollision checker to the Humble branch. One line change at the bottom of worldToMapValidated to make compatible with the Humble version of PyCostmap2D

* Unsure if any docs need changed.

* Tested in simulation, was able to use footprintCost and lineCost to detect a collision on my bot.

#### For Maintainers: 

### DIFF
--- a/nav2_simple_commander/nav2_simple_commander/footprint_collision_checker.py
+++ b/nav2_simple_commander/nav2_simple_commander/footprint_collision_checker.py
@@ -1,0 +1,220 @@
+#! /usr/bin/env python3
+# Copyright 2021 Samsung Research America
+# Copyright 2022 Afif Swaidan
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
@ThomasHaley-neya just sanity check me that you've tested this generally speaking and it works / doesn't crash from incorrect members.

ThomasHaley-neya:
> @ThomasHaley-neya just sanity check me that you've tested this generally speaking and it works / doesn't crash from incorrect members.

Yep, just went back and tested all member functions and everything seems kosher.
Screenshot from 2025-02-03 19-46-12
Screenshot from 2025-02-03 19-46-31

A crash can happen if the user calls setCostmap() and passes something other than an occupancy grid, but that is the same behavior that occurs on main. Maybe another ticket should be opened to add a check there?

SteveMacenski:
If the wrong thing is passed in, then I suppose that should be expected in Python without type checking

PR #4902, opened 2025-02-05
Jazzy sync 5: Feb 4, 2025

SteveMacenski:
Jazzy Sync 5: Feb 4, 2025 - 1.3.5


nav2_util/include/nav2_util/simple_action_server.hpp
```diff
@@ -150,6 +150,9 @@ class SimpleActionServer
     std::lock_guard<std::recursive_mutex> lock(update_mutex_);
 
     if (!server_active_) {
+      RCLCPP_INFO(
+        node_logging_interface_->get_logger(),
+        "Action server is inactive. Rejecting the goal.");
```
Timple, line 155: So actually this statement is not true then?

```
@brief handle the goal requested: accept or reject. This implementation always accepts.
```

SteveMacenski: https://github.com/ros-navigation/navigation2/pull/5150

PR #4903, opened 2025-02-05
cmake helper function and tool to auto generate TreeNodeModel xml files for groot

ewak:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | laptop |

---

* I added a generic tool and a cmake funciton to easily use it to auto generate TreeNodeModel xml files for groot.
* I applied it to nav_behaviour_tree and opennav_docker_bt.

---

#### For Maintainers: 


nav2_behavior_tree/CMakeLists.txt
```diff
@@ -263,6 +280,11 @@ install(DIRECTORY test/utils/
   DESTINATION include/${PROJECT_NAME}/nav2_behavior_tree/test/utils
 )
 
+install(DIRECTORY cmake
+  DESTINATION share/${PROJECT_NAME}
+)
+
+#NOTE: Still using manually managed nav2_tree_nodes.xml in source directory
```
SteveMacenski, line 287: Should we though? If we removed this, what would be the workflow for using groot?

nav2_docking/opennav_docking_bt/opennav_docking_tree_nodes.xml
```diff
@@ -0,0 +1,28 @@
+<root BTCPP_format="4">
```
SteveMacenski, line 1: Is the other Groot XML in the nav2_behavior_tree package not also needing an update?

SteveMacenski (commented):
Some other thoughts:

I remember chatting about this back when the plugins list for BT Navigator had the libraries stored in a similar way (as you've found here) that we weren't sure how to automatically update the groot XML in source as might be necessary for use. That way, we don't need to manually manage it ever again, it automatically updates on commit / merge. 

Or, a GitHub Action job runs to run this over and if it detects a problem it fails the job to make a human make the update for their violating PR. Verification is less ideal than automation, but verification is still very useful.

nav2_docking/opennav_docking_bt/opennav_docking_tree_nodes.xml
ewak: Yes.  I did not touch the manually managed file yet.  I will go all in.

ewak:
> Some other thoughts:
> 
> I remember chatting about this back when the plugins list for BT Navigator had the libraries stored in a similar way (as you've found here) that we weren't sure how to automatically update the groot XML in source as might be necessary for use. That way, we don't need to manually manage it ever again, it automatically updates on commit / merge.

I could can target the source directory location for the output of the generation.  And then install it from there.

My gut feeling is that the generated file should be the one that is installed and packaged and there should be a ctest to do a comparison between the generated file and one cached in the source directory.  It can use `cmake -E compare_files --ignore-eol` https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-compare_files

I will play around with an option, something like, UPDATE_SRC_CACHE_PATH, to have the generation step go ahead and update the source file if its different.   That way the developer gets a changed source file that they can choose when to commit.  If they don't the ctest will catch them in CI.

> Or, a GitHub Action job runs to run this over and if it detects a problem it fails the job to make a human make the update for their violating PR. Verification is less ideal than automation, but verification is still very useful.

Github Actions are a whole other thing to learn.  I'll try out the ctest route.

SteveMacenski:
Closing due to inactivity - though this would be nice to have!

Issue #4907, opened 2025-02-06
Redesign dynamic parameters patterns


SteveMacenski:
It is brought to my attention dynamic parameter handling matured in Humble that allows us to use the `add_post_set_parameters_callback` instead of `add_on_set_parameters_callback` for processing parameter updates into member functions. This was brought up from a conversation started by @pepisg and myself on MPPI dynamic parameter handling issues and explained in https://github.com/ros2/rclcpp/pull/2735 by Chris.

The tl;dr minimum of this ticket is to: Update globally all of the `add_on_set_parameters_callback` to use `add_post_set_parameters_callback` instead and update the dynamic parameter callback if the signature changes

This is a really nice first contribution that touches alot of code!

Secondarily, many nodes are not as well organized with larger sets of parameters like MPPI and RPP with parameter handler objects. Along with the migration, we should consider making that also a global pattern of having all objects have parameter handlers to manage the parameter's state and dynamic callbacks to simplify the application code and have consistency architecturally.

Also a really nice first-contribution that gets folks into the architecture a little!

Nils-ChristianIseke:
I will work on this.

Sahas-Ananth:
Can I also help with this?

Nils-ChristianIseke:
As suggested by @SteveMacenski in https://github.com/ros-navigation/navigation2/pull/4971#pullrequestreview-2676035908.

_Preliminary list of all nodes that need to be adjusted (searched repo for on_set_parameters_callback)
I will adapt this list_

1. - [ ] [nav2_mppi_controller/src/parameters_handler.cpp](https://github.com/ros-navigation/navigation2/blob/085d235db0ef0d189ca19315d7c69a359778fc93/nav2_mppi_controller/src/parameters_handler.cpp)
2. - [ ] [nav2_controller/plugins/pose_progress_checker.cpp](https://github.com/ros-navigation/navigation2/blob/085d235db0ef0d189ca19315d7c69a359778fc93/nav2_controller/plugins/pose_progress_checker.cpp)
3. - [ ] [nav2_costmap_2d/src/costmap_2d_ros.cpp](https://github.com/ros-navigation/navigation2/blob/085d235db0ef0d189ca19315d7c69a359778fc93/nav2_costmap_2d/src/costmap_2d_ros.cpp)
[...]

PR #4908, opened 2025-02-06
MPPI: Do not set params success in dynamic params callback to allow changing params in other plugins

pepisg:
---

## Basic Info

| Info | Theres a bug when setting params from other plugins when using mppi|
| ------ | ----------- |

---

Currently, the parameters handler library in the MPPI will return failure when trying to set a parameter in the controller server node that does not belong to the MPPI plugin. This is because rclcpp will stop processing dynamic params callbacks when one of them returns falure, see https://github.com/ros2/rclcpp/pull/2735 for details 

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/parameters_handler.cpp
+++ b/nav2_mppi_controller/src/parameters_handler.cpp
@@ -70,7 +70,6 @@ ParametersHandler::dynamicParamsCallback(
     {
       callback->second(param, result);
     } else {
-      result.successful = false;
       if (!result.reason.empty()) {
         result.reason += "\n";
       }
--- a/nav2_mppi_controller/src/parameters_handler.cpp
+++ b/nav2_mppi_controller/src/parameters_handler.cpp
@@ -69,11 +69,6 @@ ParametersHandler::dynamicParamsCallback(
       callback != get_param_callbacks_.end())
     {
       callback->second(param, result);
-    } else {
-      if (!result.reason.empty()) {
-        result.reason += "\n";
-      }
-      result.reason += "get_param_callback func for '" + param_name + "' not found.\n";
     }
   }
 
--- a/nav2_mppi_controller/test/parameter_handler_test.cpp
+++ b/nav2_mppi_controller/test/parameter_handler_test.cpp
@@ -267,13 +267,6 @@ TEST(ParameterHandlerTest, DynamicAndStaticParametersNotDeclaredTest)
   EXPECT_EQ(result.successful, true);
   EXPECT_TRUE(result.reason.empty());
 
-  // Try to access some parameters that have not been declared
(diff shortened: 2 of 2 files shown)


nav2_mppi_controller/src/parameters_handler.cpp
```diff
@@ -70,7 +70,6 @@ ParametersHandler::dynamicParamsCallback(
     {
       callback->second(param, result);
     } else {
-      result.successful = false;
       if (!result.reason.empty()) {
```
SteveMacenski, line 73: I think the reason also needs to be removed

@ewak BTW: the dynamic parameter callbacks are triggered for all objects in a node / process. This is actually causing issues because parameter updates are being rejected by MPPI that are meant for the goal checker. Its perfectly fine for parameters not in MPPI to pass through this function and we shouldn't be giving rejections based on it (apparently). 

#4907 was opened to also address some other related issues that came up from this process / conversation.

aosmw: > I think the `reason` also needs to be removed
> 
> @ewak BTW: the dynamic parameter callbacks are triggered for all objects in a node / process. This is actually causing issues because parameter updates are being rejected by MPPI that are meant for the goal checker. Its perfectly fine for parameters not in MPPI to pass through this function and we shouldn't be giving rejections based on it (apparently).

I saw the initial PR and thought hey that's there for a reason...  I subsequently clicked around and found the related tasks and intentions that I did not know about or account for.   I have no real need of this code, once I get the parameters right ;/, so revert or trim as you see fit.

aosmw: This is the commit where it was introduced.

https://github.com/ros-navigation/navigation2/commit/f45d05b809064f27ecd9d141043a01fd8b41f063

SteveMacenski (approved):

nav2_mppi_controller/src/parameters_handler.cpp
SteveMacenski: No worries, just letting you know for visibility!

SteveMacenski:
There are some test failures: https://app.circleci.com/pipelines/github/ros-navigation/navigation2/13610/workflows/4eea23eb-3231-4ea7-a2fe-ebf0d92506cb/jobs/40607/tests

- These can be deleted now: https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/test/parameter_handler_test.cpp#L270-L276
- Removing the undeclared variables that are deleted from the first bullet here and the expects: https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/test/parameter_handler_test.cpp#L278-L297

SteveMacenski (approved):

PR #4911, opened 2025-02-10
Fix Misplaced Footprint on Static Layer

HovorunBh:
---

## Basic Info

| ------ | ----------- |

---

I noticed that the robot footprint clearing is clearing the wrong spot on the local costmap. The offset is exactly odom->map TF
My static_layer on local costmap (rolling_window=True) subscribes to occupancy grid in 'map' frame
Screenshot from 2025-02-10 02-38-32

The robot pose received by the static layer is in the costmap's global frame, but it should be in the frame of the map it subscribes to, because otherwise the static layer is transformed by odom->map TF and the footprint ends up in the wrong place:
https://github.com/ros-navigation/navigation2/blob/088c423deb97a76f5a5f4ca133cb122338576fe1/nav2_costmap_2d/plugins/static_layer.cpp#L454

* Drive into static_layer's lethal space, see if the footprint is cleared in the correct place, and it is
 
Screenshot from 2025-02-10 02-35-13

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/static_layer.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/static_layer.hpp
@@ -49,6 +49,8 @@
 #include "nav_msgs/msg/occupancy_grid.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "nav2_costmap_2d/footprint.hpp"
+#include "nav2_util/robot_utils.hpp"
+#include "tf2/utils.hpp"
 
 namespace nav2_costmap_2d
 {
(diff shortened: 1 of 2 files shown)


SteveMacenski:
A few questions from a quick review to motivate discussion:

Should anything in updateBounds also be updated? 

If the costmap is rolling, we find the TF loopup from map to global frame in `updateCosts`. To keep this time synchronized, should we find the transforms at the same stamp for the footprint and have this all stored for use later? Or, should we be using the robot pose given (x,y,yaw) to transform that pose from the global_frame into the map_frame rather than getting a new raw robot pose itself with `getCurrentPose` -- it seems like we should use the inputs instead of overriding them? 

Finally, should the condition also be if the costmap is rolling or only when it is different?

SteveMacenski:
@HovorunBh can you respond to my questions from Feb?

SteveMacenski:
Closing due to inactivity and not answering necessary questions to move forward

tonynajjar:
superseeded by https://github.com/ros-navigation/navigation2/pull/6520. Hi @HovorunBh!

Issue #4769, opened 2024-11-28
Add precommit to Nav2


tonynajjar:
#### Feature description

Fixing linting issues after CI runs is always a pain. Giving the option for contributors to use pre-commit to fix all linting issues before pushing would lead to a smoother workflow, not to mention the saving of some CI time

SteveMacenski:
We have the CI linting-only jobs that should be accelerating that by checking the linters separate of the build / test stages.  https://github.com/ros-navigation/navigation2/blob/main/.github/workflows/lint.yml I believe this already does what you intend?

SteveMacenski:
You should be seeing those job results in the GitHub GUI in the same way as a precommit job https://github.com/ros-navigation/navigation2/pull/4761

tonynajjar:
Yes it's already great that the linting job runs before the long build and testing jobs but I meant to bring it one step closer to the developer (optionally). So what I'm proposing is a pre-commit config file (or similar) that will basically run the same thing as this linting job but locally before commiting (e.g. pre-commit setups a virtual environment). The user would need to run `pre-commit install` on their local Nav2 repo to "opt-in".
This would help fix linting issues before even committing which would save some time

SteveMacenski:
We could do that, yes. However I do wonder if it is necessary with good process. You should be running `colcon test` on the packages that you modify (at least) and as part of that, the linters are run. Running the tests are important before opening a PR to begin with to make sure you don't break existing behavior.

With all that said, if you added a precommit option, I'd merge it.

Nils-ChristianIseke:
I can work on this. I would suggest using what is done in ros2_control as a starting point: https://github.com/ros-controls/ros2_control/blob/master/.pre-commit-config.yaml.

SteveMacenski:
Adding in codespell might be nice too - figure that's another linting adjacent thing!

PR #4915, opened 2025-02-10
Pre-Commit

Nils-ChristianIseke:
#4769
Suggestion for a pre-commit configuration.
I've taken the one from ros2 control as a basis and added a few things I am using myself. I still need to put some work into it, but I wanted to provide this in advance as a basis for discussion.

For example, do we want to introduce ruff and maybe mypy?
What about other hooks?
How should we integrate that into the CI?


nav2_bringup/maps/tb3_sandbox.pgm
SteveMacenski, line 1: Revert these changes

nav2_docking/README.md
```diff
@@ -9,7 +9,7 @@ This is split into 4 packages
 - `opennav_docking`: Contains the main docking framework
 - `opennav_docking_core`: Contains the dock plugin header file to be implemented for each dock type
 - `opennav_docking_bt`: Contains behavior tree nodes and example XML files using the docking task server
-- `nova_carter_docking`: Contains an implementation using the Docking system with the Nvidia [Nova Carter](https://robotics.segway.com/nova-carter/) Robot platform and dock.
+- `nova_carter_docking`: Contains an implementation using the Docking system with the Nvidia [Nova Carter](https://robotics.segue.com/nova-carter/) Robot platform and dock.
```
SteveMacenski, line 12: Revert

Nils-ChristianIseke:
Thank you Steve. Sorry for not pointing out, that the changes are just a preview of what would change with the current preliminary config. Did not meant to create review noise for you.

> We have the linters already in our CI, but it might be value to add in codespell or some others that not already running as github actions.

I would vote for using pre-commit as drop in replacement for the [lint workflow](https://github.com/ros-navigation/navigation2/blob/main/.github/workflows/lint.yml). Running the linters as pre-commit hooks.

> What are the advantages of these tools? Keep in mind we have some python3 linters already with pep57 / flake8 that come from ament that we use (and have CI jobs that run)

I like [ruff](https://github.com/astral-sh/ruff)  mainly because of it's speed and support for multiple linter and formatters
Regarding linting this would need configuration of ruff to match ament_flake8 and ament_pep257, however maybe that is an overkill and we should wait for an ament_ruff package. Mypy introduces static type checking to python code, and would be a must if we wanted to enforce type hints via ruff.

SteveMacenski:
> I would vote for using pre-commit as drop in replacement for the [lint workflow](https://github.com/ros-navigation/navigation2/blob/main/.github/workflows/lint.yml). Running the linters as pre-commit hooks.

One nice thing about that though is that they each show up in the GitHub UX as a separate job, which makes it easier for a developer to see where the error is and break down a laundry list of tasks into more bite-sized results (and the jobs are named after the task for immediate review knowledge about the nature of the failure). Is there a way to do something like that with precommit? 

An option we could have is to have the precommit mirror what we check with CI jobs and developers can enable precommit locally to run the workflow, but the workflows in CI run with inidividual jobs that match the set of precommit jobs (so we'd need to add a codespell, etc in our linting workflow). 

Perhaps not ruff since we have pep57 and flake8. I try to keep this project fully aligned with the ROS 2 best practices as a reference for other applications. 

> Mypy introduces static type checking to python code, and would be a must if we wanted to enforce python type hints via ruff.

That sounds nice! Is there much we'd have to do to implement that? We honestly don't have much python anyway, so adding type hints on the returns of each method isn't actually a big deal (I think we can/should ignore `/tools` and launch files).

Nils-ChristianIseke:
> One nice thing about that though is that they each show up in the GitHub UX as a separate job, which makes it easier for a developer to see where the error is and break down a laundry list of tasks into more bite-sized results (and the jobs are named after the task for immediate review knowledge about the nature of the failure). Is there a way to do something like that with precommit?

The output of pre-commit is visible in the [summary of a job](https://github.com/ros-controls/ros2_control/actions/runs/13177407691/job/36779875876). I normally setup git hooks in my dev env, so that i am not allowed to commit if pre-commit fails (output of pre-commit looks the same). Not sure if there is a pretty way to run each pre-commit hook as seperate job.

> An option we could have is to have the precommit mirror what we check with CI jobs and developers can enable precommit locally to run the workflow, but the workflows in CI run with inidividual jobs that match the set of precommit jobs (so we'd need to add a codespell, etc in our linting workflow).

Wouldn't be maintaining parity between pre-commit and CI jobs an overhead? 

> That sounds nice! Is there much we'd have to do to implement that? We honestly don't have much python anyway, so adding type hints on the returns of each method isn't actually a big deal (I think we can/should ignore `/tools` and launch files).
I will open a separate issue regarding this.

SteveMacenski:
> The output of pre-commit is visible in the [summary of a job](https://github.com/ros-controls/ros2_control/actions/runs/13177407691/job/36779875876). I normally setup git hooks in my dev env, so that i am not allowed to commit if pre-commit fails (output of pre-commit looks the same). Not sure if there is a pretty way to run each pre-commit hook as seperate job.

That's not automatic, they manually call it from a github action workflow https://github.com/ros-controls/ros2_control/tree/5d4933d2945b72bf2625a2a28c98107433c31d9d/.github/workflows. I agree turning it on in the dev environment makes sense for those that want to use it! I think that's the intent behind the ticket. 

> Wouldn't be maintaining parity between pre-commit and CI jobs an overhead?

Sure, but if in exchange our CI has more clear failures to help developers fix those issues, I think that's worthwhile than piling everything into a single job that more junior engineers need to parse to find their issue(s). Keep in mind there's a wide array of developers that contribute to ROS 2 from the most junior writing the first bits of code in their career to Senior Staff Engineers. There are quite a few things in our process that are catered to the junior developers and hobbyists to lower the technical and process barriers of entry. 

Plus, its not like we modify linting often. We have [never actually made an update to our `lint.yml` job ](https://github.com/ros-navigation/navigation2/commits/main/.github/workflows/lint.yml)since we created it (beyond changing the image OS for the ros release). I don't think that's a major overhead :_)

Nils-ChristianIseke:
So I tested a few things, but apparently the logs from ament_flake8 and ament_pep257 are too noisy to really work well with pre-commit. Maybe that's one of the reasons why the ros2_control people don't use ament for these.
An alternative would be to use flake8 and pep257 in pre-commit, though I'm not sure how much maintenance this would require to match the ament counterparts in the lint workflow.

We could also decide to only run tools via pre-commit (e.g. codespell) that are not present in the lint workflow.

Apart from that, I have had very good experiences with pre-commit and junior developers as they get feedback earlier and there is less “push and pray”.

tonynajjar:
The motivation behind my ticket was simply the frustration of finding out about linting issues "a little bit too late". It's still not terrible since the linters run first, but still the sooner the better. Putting the discussion of which additional linters to have aside for a bit, I think just having the `lint.yaml` CI job run locally is already a win. Since I don't think running a CI job locally is common/feasible/popular, the next best thing is to use pre-commit to run the same linters. And that's exactly what this project does: https://github.com/botsandus/ament_lint_pre_commit/blob/main/.pre-commit-hooks.yaml. We could just use that.

Nils-ChristianIseke:
Thank you @SteveMacenski for charing your knowledge/ thougtht, i really appreciate this :).  I have posted a minimal config file. @tonynajjar have you tried running ament_flake8 / ament_pep257 via pre-commit? IMO it writes too much to stdout and disturbs the pre-commit output.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,13 @@
...
+  pull_request:
+
+jobs:
+  codespell:
+    name: Run codespell
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run codespell
+        uses: codespell-project/actions-codespell@v2
```
SteveMacenski, line 13: EOF line please (also the first line is blank, remove that) :-)

.github/workflows/lint.yml
```diff
@@ -11,11 +11,12 @@ jobs:
...
-          linter: [xmllint, cpplint, uncrustify, pep257, flake8]
+          linter: [xmllint, cpplint, uncrustify, pep257, flake8, mypy]
     steps:
     - uses: actions/checkout@v4
     - uses: ros-tooling/action-ros-lint@v0.1
       with:
         linter: ${{ matrix.linter }}
         distribution: rolling
         package-name: "*"
+        uses: codespell-project/actions-codespell@v2
```
SteveMacenski, line 22: Should this be deleted? Seems like the codespell.yml replaces this

.pre-commit-config.yaml
```diff
@@ -0,0 +1,71 @@
...
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.4.1
+    hooks:
+    -   id: codespell
+        additional_dependencies:
+        -   tomli
+-   repo: https://gitlab.com/vojko.pribudic.foss/pre-commit-update
+    rev: v0.6.1
+    hooks:
+    -   id: pre-commit-update
```
SteveMacenski, line 71: EOF line

.pre-commit-config.yaml
```diff
@@ -0,0 +1,71 @@
...
+      description: Check Python Type Annotations
+      language: system
+      types: [python]
+      entry: ament_mypy
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.4.1
+    hooks:
+    -   id: codespell
+        additional_dependencies:
+        -   tomli
```
SteveMacenski, line 67: What is tomli?

nav2_rviz_plugins/src/particle_cloud_display/flat_weighted_arrows_array.cpp
```diff
@@ -133,7 +133,7 @@ void FlatWeightedArrowsArray::setManualObjectVertices(
 
     for (const auto & vertex : vertices) {
       manual_object_->position(vertex);
-      manual_object_->colour(color);
+      manual_object_->colour(color);   // codespell:ignore colour
```
SteveMacenski, line 136: This doesn't seem necessary?

nav2_system_tests/src/error_codes/error_code_param.yaml
```diff
@@ -173,7 +173,7 @@ planner_server:
       plugin: "nav2_system_tests::NoValidPathErrorPlanner"
     no_viapoints_given:
       plugin: "nav2_system_tests::NoViapointsGivenErrorPlanner"
-    cancelled:
+    canceled:  # codespell:ignore
```
SteveMacenski, line 176: This was fixed, does it need codespell ignore?

nav2_navfn_planner/src/navfn.cpp
```diff
@@ -463,18 +463,18 @@ NavFn::updateCell(int n)
     if (pot < potarr[n]) {
       float le = INVSQRT2 * static_cast<float>(costarr[n - 1]);
       float re = INVSQRT2 * static_cast<float>(costarr[n + 1]);
-      float ue = INVSQRT2 * static_cast<float>(costarr[n - nx]);
+      float ue = INVSQRT2 * static_cast<float>(costarr[n - nx]);   // codespell:ignore ue
```
SteveMacenski, line 466: Can we create a dictionary of allowed words in `/tools` that contains ue, hist, nav2, ot, ws? I'd like to explicitly allow variables to not cause problems (this is more to me about spelling in comments and strings)

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[tool.codespell]
```
SteveMacenski, line 1: DOes this need to be in the repo's main directory, or can it be moved to `/tools` if the invocation is given the path to the file?

SteveMacenski (changes requested):
LGTM in general -- anything else you'd want to add here? 

@tonynajjar please review the precommit config file, I can do the rest

nav2_navfn_planner/src/navfn.cpp
SteveMacenski: I just say the yaml dictionary, we could add there too

nav2_bringup/maps/tb3_sandbox.pgm
SteveMacenski, line 1: Revert, we shouldn't mess with the binary files

SteveMacenski (commented):
There are a few PGM, SVG files that this modifies. We should make a rule to skip svg/pgm files since we'll have maps and svg design docs that shouldn't be modified

SteveMacenski:
The linters aren't running for some reason

Nils-ChristianIseke:
Hi Steve I will ping you as soon as it's ready to review :)

nav2_rviz_plugins/src/particle_cloud_display/flat_weighted_arrows_array.cpp
Nils-ChristianIseke: I enabled en-GB_to_en-US for codespell so colour would be wrong if we decide to keep this setting.

pyproject.toml
Nils-ChristianIseke: moved it to tools

nav2_navfn_planner/src/navfn.cpp
Nils-ChristianIseke: done

nav2_smac_planner/lattice_primitives/sample_primitives/5cm_resolution/1m_turning_radius/omni/output.json
```diff
@@ -6445,4 +6445,4 @@
 			]
 		}
 	]
-}
\ No newline at end of file
+}
```
SteveMacenski, line 6448: Can we skip all json and css files?  These are autogenerated

nav2_system_tests/src/error_codes/error_code_param.yaml
```diff
@@ -173,7 +173,7 @@ planner_server:
       plugin: "nav2_system_tests::NoValidPathErrorPlanner"
     no_viapoints_given:
       plugin: "nav2_system_tests::NoViapointsGivenErrorPlanner"
-    cancelled:
+    cancelled:  # codespell:ignore cancelled
```
SteveMacenski, line 176: his needs to be updated since you changed the value in `palnner_plugins`. This is a test, so easy enough. Just change here to `canceled` and remove the codespell ignore

.pre-commit-config.yaml
```diff
@@ -0,0 +1,90 @@
...
+# Or:
+#
+#     pre-commit install  # (runs every time you commit in git)
+#
+# To update this file:
+#
+#     pre-commit autoupdate
+#
+# See https://github.com/pre-commit/pre-commit
+exclude: ".pgm$|.svg$"
```
SteveMacenski, line 14: css / json

.pre-commit-config.yaml
```diff
@@ -0,0 +1,90 @@
+# To use:
```
SteveMacenski, line 1: @tonynajjar can you review just this file? I did the rest but I could use your eye on this to make sure you're also happy with it

SteveMacenski (changes requested):

SteveMacenski:
>     Using https://pre-commit.ci/ (autoupdates pre-commit hooks, autofixes prs), however that would need further modification to get ament stuff running in ci?
>    I'm still having a bit of trouble with the setup (ament in ci worflows and in pre-commit). I'm wondering if we should keep the ament stuff in pre-commit and only do additional stuff in pre-commit.

I think we decided that precommit would be for local users only, not for CI, no? Or maybe I'm missing what you're saying

>    Are you ok with adding mypy support later, as it is quite an effort to add the missing type hints?

Yes! I thought I was a PR that did some of this already, but I'm having a hard time finding it. Maybe it was a comment in the ticket

>    Should we enforce en-GB_to_en-US for codespell?

Up to you, I don't feel strongly about it

Nils-ChristianIseke:
> I think we decided that precommit would be for local users only, not for CI, no? Or maybe I'm missing what you're saying

Sorry, I had a typo: We could only execute things via pre-commit that are not available in the current ament workflows: Codespell and pre-commit-hooks. 
I would vote not to run the ament tools in pre-commit as they are currently quite noisy and prevent me from using pre-commit (Independent of running pre-commit in the CI or not).

If we do so not much additional setup would be needed for CI. This would also reopen the CI discussion, as we have mainly been discussing whether pre-commit should replace the current CI workflow. And with this setup pre-commit would be neither a replacement nor a duplication but would run as a supplement in the CI.

But if your earlier concern regarding pre-commit in CI was in general, then my bad for bringing it up again! Just wanted to make sure we’re on the same page.

.pre-commit-config.yaml
tonynajjar, line 1: - I love myself a python import sorter: https://github.com/PyCQA/isort (idk if there is one for C++ as well). What do you think of adding?
- I ran it locally and the `end-of-file-fixer` still fixes `nav2_costmap_2d/test/integration/CMakeLists.txt` and `.gitignore` for me. Doesn't it for you?
- `ament_flake8` also complains about files under the folder `tools` - I guess it is skipped in the CI so maybe also skip this folder with pre-commit?

tonynajjar:
> I would vote not to run the ament tools in pre-commit as they are currently quite noisy and prevent me from using pre-commit

hmm I can understand the verbose-ness being annoying but wasn't the original intent of the ticket/PR to make sure that happy pre-commit = happy lint CI? If so then we can't really remove them

Nils-ChristianIseke:
There was just one question that came up when I was working with Ament: Why don't we use ament_lint_auto in all packages? Isn't that the main advantage of ament_lint?

tonynajjar:
> > My proposal would be to replace `ros-tooling/action-ros-lint` with https://github.com/pre-commit/action or [pre-commit.ci](https://pre-commit.ci/). That way we don't have a duplicate way of running the `ament_x` tools and we introduce more linters in the CI (pre-commit hooks + codespell + ...)
> 
> I like that, but we'd have to convince Steve :D.

Sorry I just realized that this is already what you proposed in the beginning and Steve already presented his reasons for preferring `ros-tooling/action-ros-lint`. I think the argument mostly boils down to "having each linter in seperate jobs makes it easier for junior developers to know exactly what's wrong". Personally I find the output of precommit quite well divided even for junior engineers to understand, even if it's not in different jobs (multiple examples can be found [here](https://github.com/ros-controls/ros2_control/actions/workflows/humble-pre-commit.yml?query=is%3Afailure)). What is it that you don't find as clear @SteveMacenski? 
Also a (maybe small) downside of having separate jobs with `ros-tooling/action-ros-lint` is that each one takes 25 to 30 seconds of setup time - I suppose we run them in parallel so it's not time wasted for the user but still consumed CI time right?

> Personally, I actually run the tests before committing for the most part or manually execute ament_XYZ on the CLI to check. They really shouldn't need to push in order to get feedback. colcon test or ament_cpplint will tell them what they need to know

It's true that they don't need to push to get feedback but that's assuming that junior developers **know** what to test for, which is often not the case for first-time or rare contributors - pre-commit will allow them to run all the needed checks with one simple command (or even with no extra command if the hooks are installed)

Nils-ChristianIseke:
I think there are many more ament_lint tools we could use: https://github.com/ament/ament_lint. 
However, I would be happy if this PR is closed soon :). That said, I'm also happy with not running pre-commit in CI (even though I would prefer it).
But as a starting point we could use the current setup, and then reconsider the decision later on?

tools/pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+check-filenames = true
+check-hidden = true
+ignore-words = "tools/.codespell_ignore_words"
+interactive = 0
+quiet = 34
+skip="*.pgm,./build/*,./install/*,./log*,./.venv/*,./.git*,*.toml"
+uri-ignore-words-list = "segue"
+write-changes = false
+[tool.isort]
+profile = "google"
```
tonynajjar, line 12: any reason why you chose the `google` profile instead of `black`? I think we don't need to detach the imports from the same file, in other words I think this is fine and should stay as such:
```
from launch.actions import ExecuteProcess, IncludeLaunchDescription
```
I know it stays like this with `black` but I don't know if it can be achieved with additional configuration with `google`

Nils-ChristianIseke: Using profile black is in conflict with ament_flake8 (ordering of imports differs). But maybe there is another profile that isn't detaching imports and isn't in conflict with ament_flake8. I will check this.

Nils-ChristianIseke:
> I don't disagree! Its what our community currently uses and I want Nav2 to be aligned with ROS 2 best practices, even if there's improvements that could be made there 😉 I think that's an excuse to help improve the ROS 2 tools!

Fair point :)


> > Using profile black is in conflict with ament_flake8 (ordering of imports differs). But i still need to find the correct config that is matching ament_flake8.
> 
> This feels important to figure out before merging :-)
Done


> > My proposal would be to replace ros-tooling/action-ros-lint with https://github.com/pre-commit/action or [pre-commit.ci](https://pre-commit.ci/). That way we don't have a duplicate way of running the ament_x tools and we introduce more linters in the CI (pre-commit hooks + codespell + ...)
> 
> My main issue with that is that all the linter outputs are in the same job rather than separating by linting type. Can you re-enable the pre-commit and add a few errors in C++/Python/XML so I can see what the logs look like?
> 
> My goal is to not flood new users that have alot of errors with alot of Action output parsing to find the errors and what tool was triggering each. The job called `ament_cpplint` is very clear what tools caused it that they can test locally to see if they fixed it and the Action output of it is small enough that it doesn't require a huge amount of Action log parsing (for someone that this might be some of the first GitHub Actions logs they read).
> 
> > I'm also not the biggest fan of the ament tools, due to missing support for many args of the base tools.
> 
> I don't disagree! Its what our community currently uses and I want Nav2 to be aligned with ROS 2 best practices, even if there's improvements that could be made there 😉 I think that's an excuse to help improve the ROS 2 tools!
> 
> > I think there are many more ament_lint tools we could use: https://github.com/ament/ament_lint. However, I would be happy if this PR is closed soon :). That said, I'm also happy with not running pre-commit in CI (even though I would prefer it).
> 
> I agree, this should be in a follow up PR! Several, smaller PRs are better than one mega-one.
> 
> > Using profile black is in conflict with ament_flake8 (ordering of imports differs). But i still need to find the correct config that is matching ament_flake8.
> 
> This feels important to figure out before merging :-)



> > My proposal would be to replace ros-tooling/action-ros-lint with https://github.com/pre-commit/action or [pre-commit.ci](https://pre-commit.ci/). That way we don't have a duplicate way of running the ament_x tools and we introduce more linters in the CI (pre-commit hooks + codespell + ...)
> 
> My main issue with that is that all the linter outputs are in the same job rather than separating by linting type. Can you re-enable the pre-commit and add a few errors in C++/Python/XML so I can see what the logs look like?
> 
> My goal is to not flood new users that have alot of errors with alot of Action output parsing to find the errors and what tool was triggering each. The job called `ament_cpplint` is very clear what tools caused it that they can test locally to see if they fixed it and the Action output of it is small enough that it doesn't require a huge amount of Action log parsing (for someone that this might be some of the first GitHub Actions logs they read).
> 
> > I'm also not the biggest fan of the ament tools, due to missing support for many args of the base tools.
> 
> I don't disagree! Its what our community currently uses and I want Nav2 to be aligned with ROS 2 best practices, even if there's improvements that could be made there 😉 I think that's an excuse to help improve the ROS 2 tools!
> 
> > I think there are many more ament_lint tools we could use: https://github.com/ament/ament_lint. However, I would be happy if this PR is closed soon :). That said, I'm also happy with not running pre-commit in CI (even though I would prefer it).
> 
> I agree, this should be in a follow up PR! Several, smaller PRs are better than one mega-one.
> 
> > Using profile black is in conflict with ament_flake8 (ordering of imports differs). But i still need to find the correct config that is matching ament_flake8.
> 
> This feels important to figure out before merging :-)

SteveMacenski:
My thought after looking over the failure logs is that there are some of these linters that have a huge amount of output (look at `ament_cpplint` for example). If there were issues with flake8 that was concise, but a user had to scan through hundreds of lines of cpplint output to realize the issue wasn't there, I think that would slow down review-to-update development cycle.

So I think my preference would be to have CI setup as separate jobs for each linter to maximize the speed of identifying issues to resolve, rather than using a single precommit job for all of them. 

I ACK that this is largely subjective and creates a maintenance burden on myself to make sure these remain aligned, but I really value trying to lower the barrier to entry of new developers to Nav2 to encourage contributions from the community. I think the trade off of an occasional duplication given how infrequently we update linters and static analysis tools is worthwhile.  

So, I think we can merge this PR if we disable the precommit CI job and fix the errors introduced for demonstration! Then we can move on to adding more linters that you found :wink: 

---

An aside, I would appreciate if you didn't force push, it makes me look over all 176 files again to validate them versus just the files that changed on an update cycle of review :smile:

Nils-ChristianIseke:
I see the same problem: https://github.com/ament/ament_lint/pull/528

Maybe we can fully switch to pre-commit as soon as the ament_lint tooling is quiet.

Sry for force pushing I will avoid it in the future.

SteveMacenski:
> Maybe we can recosider using pre-commit as soon as the ament_lint tooling is supporting quiet mode for all the tools we use?

Sounds good!

Can you change the cache number to v32 from v31 in this file? https://github.com/ros-navigation/navigation2/blob/main/.circleci/config.yml It should be in there 3x that all need to be changed. That should fix CI and I can merge!

Nils-ChristianIseke:
@SteveMacenski i unfortunatly forgot to sign a commit and had to force push, sry for that!. Fortunately, I didn't have time to deal with the failing CI in detail. Otherwise I would have spent a lot of time wondering what I had done wrong :D. How were you able to track this down?

SteveMacenski:
> How were you able to track this down?

I've been doing this awhile and familiar with the types of things that cause the cache to have problems :wink: 

Also, I see that you added an `ament_lint` submodule, can you remove that? Otherwise, its ready to merge IMO

Nils-ChristianIseke:
That was just to emphasize that the hook - id: forbid-submodules was missing :D.

SteveMacenski (approved):

SteveMacenski:
Waiting on CI!

SteveMacenski:
Ah! Not so fast! Now we have a merge conflict. Can you resolve that? Once CI turns over, I'm ready to merge!

SteveMacenski:
@Nils-ChristianIseke thanks much for the iteration! Its in now!

Issue #4914, opened 2025-02-10
Perf Testing: Costmap2DPublisher::prepareCostmap


SteveMacenski:
I think we should instead do a memcpy or direct assignment instead of looping in the costmap publisher on this line: https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_publisher.cpp#L189

The others need loops due to using translation maps, but this is raw copy

gilesagnel:
@SteveMacenski, I can work on this.

SteveMacenski:
Please! 😄

glitchhopcore:
👀

PR #4919, opened 2025-02-13
removed the deep copy for the costmap_raw_ data using memcpy in prepareCostmap under costmap_2d_publisher.cpp

glitchhopcore:
---

## Basic Info

| Info |  |
| ------ | ----------- |
| Ticket(s) this addresses   | #4914  |

---

* I modified the loop that was being used to copy raw costmap by using a memcpy instead 

* None would be required as it is just an enhancement 

---

* Further profiling costmap related functionalities to identify which processes are costing us the most.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_publisher.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_publisher.cpp
@@ -185,9 +185,7 @@ void Costmap2DPublisher::prepareCostmap()
   costmap_raw_->data.resize(costmap_raw_->metadata.size_x * costmap_raw_->metadata.size_y);
 
   unsigned char * data = costmap_->getCharMap();
-  for (unsigned int i = 0; i < costmap_raw_->data.size(); i++) {
-    costmap_raw_->data[i] = data[i];
-  }
+  memcpy(costmap_raw_->data.data(), data, costmap_raw_->data.size());
 }
 
 std::unique_ptr<map_msgs::msg::OccupancyGridUpdate> Costmap2DPublisher::createGridUpdateMsg()


SteveMacenski (commented):
Per our email, please get some metrics. You can create a simple program that creates the Costmap2DPublisher object, gives it a costmap of realistic size and call prepareCostmap on it in a loop 10000x. Before and after the loop, use Chrono to take the current time and print to screen how long it took to do that operation.  Change to using memcpy and then run it again to see the performance change.

For the local costmap, a realistic size might be 10x10m at 0.05m resolution. For a global, maybe a 100x100m at 0.05m resolution.

glitchhopcore:
Thanks for the review @SteveMacenski, I will try to do so and share some results here before and after using memcpy!

glitchhopcore:
I have added some results from benchmarking the prepareCostmap() method over 10,000 iterations and its performance improvements before and after using memcpy.

For a local costmap of dimensions 10x10m:

Original method: ~287ms

Untitled design

After memcpy implementation: ~24ms

Untitled design(1)


For a global costmap of dimensions 100x100m:

Original method: ~33,888ms

Untitled design(2)

After memcpy implementation: ~5,607ms

Untitled design(3)

I used the following code for benchmarking these results:

```cpp
#include <chrono>
#include <iostream>
#include <random>
#include "rclcpp/rclcpp.hpp"
#include "rclcpp_lifecycle/lifecycle_node.hpp"
#include "nav2_costmap_2d/costmap_2d_publisher.hpp"
#include "nav2_costmap_2d/costmap_2d.hpp"

int main() {
    rclcpp::init(0, nullptr);
    auto node = std::make_shared<rclcpp_lifecycle::LifecycleNode>("costmap_bench");
    node->configure();
    node->activate();

    nav2_costmap_2d::Costmap2D cmap(2000, 2000, 0.05, 0.0, 0.0); // or 200, 200, 0.05 for local costmap
    
    std::mt19937 gen(std::random_device{}());
    std::uniform_int_distribution<> dist(0, 255);
    
    for(unsigned int i = 0; i < cmap.getSizeInCellsX(); i++) {
        for(unsigned int j = 0; j < cmap.getSizeInCellsY(); j++) {
            cmap.setCost(i, j, dist(gen));
        }
    }

    nav2_costmap_2d::Costmap2DPublisher pub(node, &cmap, "map", "costmap", true, 0.0);

    auto t1 = std::chrono::high_resolution_clock::now();
    for(int i = 0; i < 10000; i++) {
        pub.prepareCostmap();
    }
    auto t2 = std::chrono::high_resolution_clock::now();
    
    std::cout << "Time taken with memcpy: " << std::chrono::duration_cast<std::chrono::milliseconds>(t2 - t1).count() << "ms\n";

    node->deactivate();
    node->cleanup();
    node->shutdown();
    rclcpp::spin_some(node->get_node_base_interface());
    rclcpp::shutdown();
    return 0;
}

SteveMacenski:
Ah very nice and perfect, a good 5-10x speed up!

PR #4923, opened 2025-02-17
Optimize costmap2d publisher performance

mini-1235:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Related ticket is #4914  |

---

Motivated by the ticket, I did a quick profiling and tried to optimize the performance, for those loops using translation maps, I suggest to use std::transform, it is a 2-5x speed up, for the update ones, I suggest to get the costmap pointer instead of calling `getcost(x,y)`, also a `copy_n` is used in the inner loop, which is around 10x speed up

Run on the unit test and my simulation

---

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Great, this looks perfect, thanks!

PR #4924, opened 2025-02-17
Throttle collision monitor's polygon ingestion update log message

bkrodinger:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu 24 |
| Robotic platform tested on | gazebo sim and on-machine |

---

Throttle the RCLCPP_INFO log message in the `polygonCallback` of collision monitor. When a polygon update is sent to collision monitor the log message "Polygon shape update has been arrived". If you are providing polygon updates often this will fill logs. Would consider making debug message as well. It is nice to have in logs though. Also fixed the grammar of the log message while I was modifying the line.

+ In simulation and on-machine

---

#### For Maintainers: 

### DIFF
--- a/nav2_collision_monitor/src/polygon.cpp
+++ b/nav2_collision_monitor/src/polygon.cpp
@@ -588,9 +588,16 @@ Polygon::dynamicParametersCallback(
 
 void Polygon::polygonCallback(geometry_msgs::msg::PolygonStamped::ConstSharedPtr msg)
 {
-  RCLCPP_INFO(
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+  auto& clock = *node->get_clock();
+  RCLCPP_INFO_THROTTLE(
     logger_,
-    "[%s]: Polygon shape update has been arrived",
+    clock,
+    2000,
+    "[%s]: Polygon shape update has arrived",
     polygon_name_.c_str());
   updatePolygon(msg);
 }


nav2_collision_monitor/src/polygon.cpp
```diff
@@ -588,9 +588,16 @@ Polygon::dynamicParametersCallback(
 
 void Polygon::polygonCallback(geometry_msgs::msg::PolygonStamped::ConstSharedPtr msg)
 {
-  RCLCPP_INFO(
+  auto node = node_.lock();
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+  auto & clock = *node->get_clock();
```
SteveMacenski, line 595: I'd prefer if you stored the clock as a class member variable in the `configure` method of the polygon and then used `clock_` here instead of working with the `node` directly. We already get the node at that point and the clock doesn't change.

SteveMacenski (changes requested):
Nitpick for consistency, but this looks good to me. Sorry for the delay, I was on vacation.

SteveMacenski:
@bkrodinger any update?

nav2_collision_monitor/src/polygon.cpp
bkrodinger: Done

bkrodinger:
> Nitpick for consistency, but this looks good to me. Sorry for the delay, I was on vacation.

No worries! Hope you had a good vacation!

nav2_collision_monitor/src/polygon.cpp
bkrodinger: Done

SteveMacenski (approved):

SteveMacenski:
Thanks! Merged

PR #4925, opened 2025-02-17
[nav2_behavior_tree] Add force_use_current_pose to ComputePathToPoseAction

doisyg:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on |(Ubuntu|

---

Adds an optional input field `force_use_current_pose` to the `ComputePathToPoseAction` BT node.
When set to true, use the current robot pose, even if `start` is provided (effectively ignoring it).
This is useful when calling several time the `ComputePathToPoseAction` in a BT with different arguments and wanting to use the current pose of the robot after a call where `start` was used (and hence existing in the blackboard).

If accepted in this state, then update this: https://docs.nav2.org/configuration/packages/bt-plugins/actions/ComputePathToPose.html#computepathtopose

Extensively tested on my physical robot platform (but backported from Dexory jazzy branch)

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
@@ -80,6 +80,9 @@ public:
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
           "start", "Start pose of the path if overriding current robot pose"),
+        BT::InputPort<bool>(
+          "force_use_current_pose", "Use current robot pose, even if `start` is provided "
+          "(effectively ignoring it)"),
         BT::InputPort<std::string>(
           "planner_id", "",
           "Mapped name to the planner plugin type to use"),
--- a/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
+++ b/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
@@ -33,7 +33,13 @@ void ComputePathToPoseAction::on_tick()
   getInput("goal", goal_.goal);
   getInput("planner_id", goal_.planner_id);
   if (getInput("start", goal_.start)) {
-    goal_.use_start = true;
+    bool force_use_current_pose = false;
+    getInput("force_use_current_pose", force_use_current_pose);
+    if (force_use_current_pose) {
+      goal_.use_start = false;
+    } else {
+      goal_.use_start = true;
+    }
   }
 }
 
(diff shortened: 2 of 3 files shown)


nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/compute_path_to_pose_action.hpp
```diff
@@ -80,6 +80,9 @@ class ComputePathToPoseAction : public BtActionNode<nav2_msgs::action::ComputePa
         BT::InputPort<geometry_msgs::msg::PoseStamped>("goal", "Destination to plan to"),
         BT::InputPort<geometry_msgs::msg::PoseStamped>(
           "start", "Start pose of the path if overriding current robot pose"),
+        BT::InputPort<bool>(
+          "force_use_current_pose", "Use current robot pose, even if `start` is provided "
```
SteveMacenski, line 84: Can we change this to `override_start_as_current_pose` - its more descriptive.

Or, probably even better, just simple `use_start` and we check if `use_start` and set as `start` then (so invert the 2x `if` conditions in the cpp file)

Also, the complimentary docs PR is necessary to add in the new port

doisyg: I like the `use_start` name! But we have to maintain the legacy logic though (when start was used if it was set).
Wdyt of the following:

```
  // if "use_start" is provided try to enforce it (true or false), but we cannot enforce true if
  // start is not provided
  bool bb_use_start;
  if (getInput("use_start", bb_use_start)) {
    if (bb_use_start) {
      // check that we have a "start" pose
      geometry_msgs::msg::PoseStamped bb_start;
      if (getInput("start", bb_start)) {
        goal_.start = bb_start;
        goal_.use_start = true;
      } else {
        goal_.use_start = false;
        RCLCPP_ERROR(
          node_->get_logger(),
          "use_start is set to true but no start pose was provided, falling back to default behavior"
          ", i.e. using the current robot pose");
      }
    } else {
      goal_.use_start = false;
    }
  } else {
    // else if "use_start" is not provided, but "start" is, then use it in order to not change
    // the legacy behavior
    geometry_msgs::msg::PoseStamped bb_start;
    if (getInput("start", bb_start)) {
      goal_.start = bb_start;
      goal_.use_start = true;
    } else {
      goal_.use_start = false;
    }
  }

```

nav2_behavior_tree/plugins/action/compute_path_to_pose_action.cpp
```diff
@@ -32,8 +32,37 @@ void ComputePathToPoseAction::on_tick()
 {
   getInput("goal", goal_.goal);
   getInput("planner_id", goal_.planner_id);
-  if (getInput("start", goal_.start)) {
-    goal_.use_start = true;
+
+  // if "use_start" is provided try to enforce it (true or false), but we cannot enforce true if
+  // start is not provided
+  bool bb_use_start;
```
SteveMacenski, line 38:
```suggestion
  goal_.use_start = false;
  bool bb_use_start;
```

If you initialize it to `false`, then you can remove that line in the next 2x `else` statements (and remove the second else statement entirely).

SteveMacenski (commented):
What's `bb_`? Maybe just set as `start` / `use_start` without the prefix? :smile:

doisyg:
(and docs: https://github.com/ros-navigation/docs.nav2.org/pull/679)

SteveMacenski (approved):

Issue #4916, opened 2025-02-11
Introduction ament_mypy


Nils-ChristianIseke:
## Feature request
Introduce ament_mypy to the codebase, to enforce and check type hinting on python code.

#### Feature description
Increasing the Python code quality by introducing a static type checker.

#### Implementation considerations
See [how to use mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html).
To generate type hints automatically one could use tools like [pyannotate](https://github.com/pyannote/pyannote-audio) or [monkeytype](https://github.com/Instagram/MonkeyType).

SteveMacenski:
Should this be in nav2 or in the ament linting: https://github.com/ament/ament_lint

PS it looks like mypy is already there 🙃

SteveMacenski:
Smac Planner state lattice generator is completed!

leander-dsouza:
Hey Steve, I have tried integrating this over at #5022.

SteveMacenski:
Lovely, thanks!

SteveMacenski:
Reopening, I think that was just for smac planner, yeah?

leander-dsouza:
> Reopening, I think that was just for smac planner, yeah?

Yes, all the python packages need to be added separately.

Nils-ChristianIseke:
After adding type hints  to a package we should run:  
`ruff check --ignore Q000 --select ALL <package_name>` to ensure type hinting is according to the latest python standard.
We can ignore all non-type related issues.

leander-dsouza:
> After adding type hints to a package we should run: `ruff check --select UP <package_name>`

We can add the ruff linter to the pre-commit once #5029 is merged.

SteveMacenski:
I don't think (tentatively and ignorantly) that we should be ignoring any nav2 packages in the mypy system, else it seems to defeat some of the purpose of enabling it in the first place. 

```

[[tool.mypy.overrides]]
module = [
    "matplotlib.*",
    "rtree.*",
    "launch.*",
    "ament_index_python.*",
    "nav2_common.*",
    "nav2_msgs.*",
    "nav2_simple_commander.*",
    "launch_testing.*",
    "action_msgs.*",
    "geometry_msgs.*",
    "sensor_msgs.*",
    "tf2_ros.*",
    "nav_msgs.*",
    "rosgraph_msgs.*",
    "tf_transformations.*",
    "ament_pep257.*",
    "ament_flake8.*",
[...]

PR #4927, opened 2025-02-17
Update smac planner types

InvincibleRMC:
---

## Basic Info

| ------ | ----------- |

---

I saw #4916 to add ament_mypy to type check. I noticed there were some errors raised by ament_mypy so i resolved the ones in `nav2_smac_planner`.

I ran `ament_mypy nav2_smac_planner/` and got no errors.

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/lattice_primitives/lattice_generator.py
+++ b/nav2_smac_planner/lattice_primitives/lattice_generator.py
@@ -14,12 +14,23 @@
 
 from collections import defaultdict
 from enum import Enum
+from typing import Any, Dict, List, Tuple, TypedDict
 
 from helper import angle_difference, interpolate_yaws
 import numpy as np
 from rtree import index
-from trajectory import Path, Trajectory, TrajectoryParameters
+
+from trajectory import AnyFloat, FloatNDArray, Path, Trajectory, TrajectoryParameters
+
 from trajectory_generator import TrajectoryGenerator
+
+
+class ConfigDict(TypedDict):
+    grid_resolution: float
+    turning_radius: float
+    stopping_threshold: int
+    num_of_headings: int
+    motion_model: str
 
 
 class LatticeGenerator:
@@ -46,7 +57,7 @@ class LatticeGenerator:
         Y = 2
         BOTH = 3
 
-    def __init__(self, config: dict):
+    def __init__(self, config: ConfigDict):
         """Init the lattice generator from the user supplied config."""
         self.trajectory_generator = TrajectoryGenerator(config)
         self.grid_resolution = config['grid_resolution']
@@ -60,7 +71,7 @@ class LatticeGenerator:
         self.DISTANCE_THRESHOLD = 0.5 * self.grid_resolution
         self.ROTATION_THRESHOLD = 0.5 * (2 * np.pi / self.num_of_headings)
 
-    def _get_wave_front_points(self, pos: int) -> np.array:
+    def _get_wave_front_points(self, pos: int) -> FloatNDArray:
         """
         Calculate the end points that lie on the wave front.
 
@@ -97,7 +108,7 @@ class LatticeGenerator:
 
         return np.array(positions)
 
-    def _get_heading_discretization(self, number_of_headings: int) -> list:
+    def _get_heading_discretization(self, number_of_headings: int) -> List[int]:
         """
         Calculate the heading discretization based on the number of headings.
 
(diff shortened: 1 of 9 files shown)


Nils-ChristianIseke:
You should also enable ament_mypy in CI. We could configure ament_mypy so that is only checks the files we explicitly name by adding the following to tools/pyproject.toml.
[tool.mypy]
files = ["path/to/your/module/"]

InvincibleRMC:
> Thank you @InvincibleRMC for pushing this :). You should also enable ament_mypy in CI. We could configure ament_mypy so that is only checks the files we explicitly name by adding the following to tools/pyproject.toml. [tool.mypy] files = ["path/to/your/module/"] Then you need to run ament_mypy --config tools/pyproject.toml

Currently ament_mypy ignores the `files` argument see https://github.com/ament/ament_lint/issues/519 for more information. I can take a look into fixing in ament_lint but, for now I just limited the CI to run on the specific package.

Nils-ChristianIseke:
@InvincibleRMC I just tried running ament_mypy locally and it's still complaining about some errors in nav2_smac_planner. Are you able to reproduce?

nav2_smac_planner/lattice_primitives/py.typed
SteveMacenski, line 1: What is this file?

nav2_smac_planner/lattice_primitives/lattice_generator.py
```diff
@@ -517,6 +533,10 @@ def _create_complete_minimal_spanning_set(
                         )
                     )
 
+                    if (unflipped_trajectory is None or flipped_y_trajectory is None or
```
SteveMacenski, line 536: I believe this condition makes the other 2 eq in the `if/else` conditions redundant.

InvincibleRMC:
No I get the same output as CI. What do you see?

```
rmc@rmc-Legion-Pro-7-16IRX9H:~/navigation2/nav2_smac_planner$ ament_mypy .

9 files checked
0 errors
Success: no issues found in 9 source files


Checked files:

* lattice_primitives/constants.py
* lattice_primitives/generate_motion_primitives.py
* lattice_primitives/helper.py
* lattice_primitives/lattice_generator.py
* lattice_primitives/trajectory.py
* lattice_primitives/trajectory_generator.py
* lattice_primitives/tests/test_lattice_generator.py
* lattice_primitives/tests/test_trajectory_generator.py
* lattice_primitives/tests/trajectory_visualizer.py
```

nav2_smac_planner/lattice_primitives/lattice_generator.py
SteveMacenski: Also, is this strictly speaking a problem if there are none-typed trajectories while we're performing this search?

SteveMacenski (changes requested):
Thanks! Great first start to maturing our Python3 code!

Nils-ChristianIseke:
Apparently ament_mypy is not enforcing mypy strict. If we make the effort of introducing type hints we should consider adding checking type hints against my strict? https://mypy.readthedocs.io/en/stable/existing_code.html#getting-to-strict

SteveMacenski:
I generally go with the profiles setup by `ament_XYZ` linters to be consistent across the ROS 2 core repositories. With that said, if there are stricter methods you'd prefer, as long as it passes with the ament version, I'm cool with that :-) 

Or, you can open a PR with ament_mypy and see if you can get the strict one in place if the maintainers like?

InvincibleRMC:
We could definitely update this specific package to use strict but I know for sure upstream making ament_mypy strict would cause packages to fail.

nav2_smac_planner/lattice_primitives/py.typed
InvincibleRMC: To allow users who import this package to use the type hints. More info [here](https://peps.python.org/pep-0561/)

Nils-ChristianIseke:
Without strict, unhelpful types like Any are allowed. Introducing mypy --strict in navigation2 is not incompatible to non strict.

Nils-ChristianIseke:
> We could definitely update this specific package to use strict but I know for sure upstream making ament_mypy strict would cause packages to fail.

This is a breaking change. I'm pretty sure it's far too noisy to introduce that now to ament_mypy.

SteveMacenski:
ah ok - well I think we should stick with `ament_mypy` for now, but I'm happy to have it adjusted to strict if/when that's set to strict!

nav2_smac_planner/lattice_primitives/lattice_generator.py
SteveMacenski: @InvincibleRMC can you respond to these items?

InvincibleRMC: Sorry I missed this.

> I believe this condition makes the other 2 eq in the `if/else` conditions redundant.

For this do you mean mathematically a path has to exist? Or do you mean it  should be type narrowing somehow?

> Also, is this strictly speaking a problem if there are none-typed trajectories while we're performing this search?

If you prefer it could return an empty Dictionary rather than raising but, it seemed to the output shouldn't be empty.

SteveMacenski (approved):

PR #4928, opened 2025-02-18
MPPI Controller improvements(15-21% performance gain).

Ayush1285:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Optimizer Benchmark |
| Does this PR contain AI generated software? |No |

---

* Addressed a few comments from the previous PR for better code readability.
* Optimized functions which were major hotspots in the profiler data.
* Computation time improved by 15-21%. 21% for Omni-Drive, 15% for Diff-drive and 17% for Ackermann.
* Minor optimizations, such as reducing division operation and caching values.

* Testes it on Optimizer benchmark with multiple runs.

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -80,28 +80,24 @@ public:
     unsigned int n_cols = state.vx.cols();
 
     for (unsigned int i = 1; i < n_cols; i++) {
-      Eigen::ArrayXf vx_last = state.vx.col(i - 1);
       state.cvx.col(i - 1) = state.cvx.col(i - 1)
-        .cwiseMax(vx_last + min_delta_vx)
-        .cwiseMin(vx_last + max_delta_vx);
+        .cwiseMax(state.vx.col(i - 1) + min_delta_vx)
+        .cwiseMin(state.vx.col(i - 1) + max_delta_vx);
       state.vx.col(i) = state.cvx.col(i - 1);
 
-      Eigen::ArrayXf wz_last = state.wz.col(i - 1);
       state.cwz.col(i - 1) = state.cwz.col(i - 1)
-        .cwiseMax(wz_last - max_delta_wz)
-        .cwiseMin(wz_last + max_delta_wz);
+        .cwiseMax(state.wz.col(i - 1) - max_delta_wz)
+        .cwiseMin(state.wz.col(i - 1) + max_delta_wz);
       state.wz.col(i) = state.cwz.col(i - 1);
 
       if (is_holo) {
-        Eigen::ArrayXf vy_last = state.vy.col(i - 1);
         state.cvy.col(i - 1) = state.cvy.col(i - 1)
-          .cwiseMax(vy_last - max_delta_vy)
-          .cwiseMin(vy_last + max_delta_vy);
+          .cwiseMax(state.vy.col(i - 1) - max_delta_vy)
+          .cwiseMin(state.vy.col(i - 1) + max_delta_vy);
         state.vy.col(i) = state.cvy.col(i - 1);
       }
     }
   }
-
 
   /**
    * @brief Whether the motion model is holonomic, using Y axis
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -77,32 +77,31 @@ public:
     float max_delta_vy = model_dt_ * control_constraints_.ay_max;
     float max_delta_wz = model_dt_ * control_constraints_.az_max;
 
-    unsigned int n_rows = state.vx.rows();
     unsigned int n_cols = state.vx.cols();
 
-    // Default layout in eigen is column-major, hence accessing elements in
-    // column-major fashion to utilize L1 cache as much as possible
-    for (unsigned int i = 1; i != n_cols; i++) {
-      for (unsigned int j = 0; j != n_rows; j++) {
-        float vx_last = state.vx(j, i - 1);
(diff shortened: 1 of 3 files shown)


Ayush1285:
@SteveMacenski fixed the leftover comments from the previous PR. Also optimized a few functions with decent 10-17% perf gain.

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -174,16 +174,10 @@ void CostCritic::score(CriticData & data)
       // Let near-collision trajectory points be punished severely
       // Note that we collision check based on the footprint actual,
       // but score based on the center-point cost regardless
-      if (pose_cost >= 253.0f /*INSCRIBED_INFLATED_OBSTACLE in float*/) {
```
SteveMacenski, line 177: Is there any performance improvements to the change in this file? Otherwise, I'd prefer to keep as-is. I find these to be less clear to the reader on quick evaluation, even if expertly concise

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -77,33 +77,32 @@ class MotionModel
...
+        Eigen::ArrayXf vy_last = state.vy.col(i - 1);
+        state.cvy.col(i - 1) = state.cvy.col(i - 1)
+          .cwiseMax(vy_last - max_delta_vy)
+          .cwiseMin(vy_last + max_delta_vy);
+        state.vy.col(i) = state.cvy.col(i - 1);
       }
     }
   }
 
+
```
SteveMacenski, line 105: Empty line

nav2_mppi_controller/src/optimizer.cpp
```diff
@@ -315,18 +315,18 @@ void Optimizer::integrateStateVelocities(
...
+
+  float last_yaw = initial_yaw;
+  for(size_t i = 0; i != n_size; i++) {
+    last_yaw += wz(i) * settings_.model_dt;
+    traj_yaws(i) = last_yaw;
   }
 
-  Eigen::ArrayXf yaw_cos = traj_yaws.cos();
-  Eigen::ArrayXf yaw_sin = traj_yaws.sin();
+  Eigen::ArrayXf yaw_cos(traj_yaws.rows());
```
SteveMacenski, line 326: Does this block really make a meaningful performance change? 

```
  Eigen::ArrayXf yaw_cos(traj_yaws.rows());
  yaw_cos = traj_yaws.cos();
```

SteveMacenski (commented):
Amazing work, thanks alot! Its spectacular the improvements made to MPPI in the last year :1st_place_medal: This is a very high impact, high usage part of Nav2 and I'm very glad you've jumped in and make it even so much better!

Sorry for the late review, I was on a trip and getting back up to speed with everything now :-)

nav2_mppi_controller/src/critics/cost_critic.cpp
Ayush1285: Eigen::cwiseMax/Min are their vectorised operations for clamping. I saw some improvements after profiling. I'll give you exact profiled data. But I think yes, this ternary operation line can be removed. I did this just to reduce branch prediction misses.

nav2_mppi_controller/src/optimizer.cpp
Ayush1285: Not really, I'll remove them.

SteveMacenski: OK!

nav2_mppi_controller/src/critics/cost_critic.cpp
SteveMacenski: I'm not sure I understand. The change on this line is for the cost critic:

```
      traj_cost += (pose_cost >= 253.0f) ? critical_cost_ : (!near_goal ? pose_cost : 0.0f);
```

It is equivalent, but unless that actually creates a meaningful improvement in performance, I'd prefer to keep it as it was since its more immediately readable to future contributors.

Ayush1285: > It is equivalent, but unless that actually creates a meaningful improvement in performance, I'd prefer to keep it as it was since its more immediately readable to future contributors.

It is not adding meaningful improvements, I'll remove it.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -77,29 +77,24 @@ class MotionModel
...
-
-        if (is_holo) {
-          float vy_last = state.vy(j, i - 1);
-          float & cvy_curr = state.cvy(j, i - 1);
-          cvy_curr = utils::clamp(vy_last - max_delta_vy, vy_last + max_delta_vy, cvy_curr);
-          state.vy(j, i) = cvy_curr;
-        }
+    for (unsigned int i = 1; i < n_cols; i++) {
+      state.cvx.col(i - 1) = state.cvx.col(i - 1)
+        .cwiseMax(state.vx.col(i - 1) + min_delta_vx)
```
SteveMacenski, line 84: Last question: why get rid of `vx_last` and call `state.vx.col(i - 1)` multiple times - is that faster? Usually I'd think removing the lookup in a data structure is faster, though I suppose for Eigen I could see this not actually being a lookup and is a placeholder until executed, so let me know if this made any difference. 

As long as its not slower, I'm fine with the change :-)

Ayush1285: >  so let me know if this made any difference.

Yes, it improved the performance a little bit.
Screenshot from 2025-03-13 00-25-04
Screenshot from 2025-03-13 00-25-20

SteveMacenski (approved):
Great, love this, thanks!

PR #4931, opened 2025-02-20
Controller FollowPath feedback: Fix wrong frame in distance calculation, add more info to feedback

adivardi:
---

## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu 24.04, Jazzy branch |

---

- Fix a bug in controller's feedback computation: The robot pose was not transformed to the path pose, which gave wrong (and often changing) results
- Added `closest_path_pose_idx` and `distance_to_path` to the feedback
- very small fix in `nav2_behavior_tree/nav2_tree_nodes.xml`

none (?)

Tested on a Gazebo  simulation of our robot,  using Jazzy  (same commits, just based on jazzy)

---

Open question for discussion: Should the feedback from the `FollowPath` action be written into the Blackboard in the BT action node (`follow_path_action`)?  
  -  see commit `Write FollowPath Feedback to blackboard` (reverted for now)
  - Could also apply to any other action

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -663,25 +663,50 @@ void ControllerServer::computeAndPublishVelocity()
   feedback->speed = std::hypot(cmd_vel_2d.twist.linear.x, cmd_vel_2d.twist.linear.y);
 
   // Find the closest pose to current pose on global path
-  nav_msgs::msg::Path & current_path = current_path_;
-  auto find_closest_pose_idx =
-    [&pose, &current_path]() {
+  geometry_msgs::msg::PoseStamped robot_pose_in_path_frame;
+  rclcpp::Duration tolerance(rclcpp::Duration::from_seconds(costmap_ros_->getTransformTolerance()));
+  // TODO move to publishFeedback method
+  if (nav_2d_utils::transformPose(
+          costmap_ros_->getTfBuffer(), current_path_.header.frame_id, pose, robot_pose_in_path_frame, tolerance))
+  {
+    nav_msgs::msg::Path& current_path = current_path_;
+    auto find_closest_pose_idx = [&robot_pose_in_path_frame, &current_path]()
+    {
       size_t closest_pose_idx = 0;
       double curr_min_dist = std::numeric_limits<double>::max();
-      for (size_t curr_idx = 0; curr_idx < current_path.poses.size(); ++curr_idx) {
-        double curr_dist = nav2_util::geometry_utils::euclidean_distance(
-          pose, current_path.poses[curr_idx]);
-        if (curr_dist < curr_min_dist) {
+      for (size_t curr_idx = 0; curr_idx < current_path.poses.size(); ++curr_idx)
+      {
+        double curr_dist =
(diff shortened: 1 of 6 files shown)


nav2_msgs/action/FollowPath.action
```diff
@@ -24,4 +24,6 @@ string error_msg
 ---
 #feedback definition
 float32 distance_to_goal
+uint64 closest_path_pose_idx   # index of the pose on global path closest to current robot pose
```
SteveMacenski, line 27: I think these should be removed or at least adjusted. The `distance_to_path` is not very accurate since it is the nearest path point, not the actual distance from the path. When using very dense paths, this might be reasonably accurate, but when the path markers are further away, then these measures are pretty poor. 

How would these two values be used by client applications?

nav2_controller/src/controller_server.cpp
```diff
@@ -659,18 +659,32 @@ void ControllerServer::computeAndPublishVelocity()
...
+  publishVelocity(cmd_vel_2d);
+
+  // Find the closest pose to current pose on global path
+  geometry_msgs::msg::PoseStamped robot_pose_in_path_frame;
+  rclcpp::Duration tolerance(rclcpp::Duration::from_seconds(costmap_ros_->getTransformTolerance()));
+  if (!nav_2d_utils::transformPose(
+          costmap_ros_->getTfBuffer(), current_path_.header.frame_id, pose,
+          robot_pose_in_path_frame, tolerance))
+  {
+    RCLCPP_ERROR(get_logger(), "Robot pose is not available.");
```
SteveMacenski, line 672: I think an exception should be thrown, this is a true error condition if TF transformations are not available

SteveMacenski (changes requested):

nav2_msgs/action/FollowPath.action
adivardi: true, but I think that's the best estimation we have without linearly interpolating the path, which introduces a different assumption. Shouldn't most paths be quite dense if they are coming from a nav2 planner?

For my use case: I am using them to monitor whether the robot has deviated the too much from the path and switch to another behavior. For example: while following a coverage path re-plan a detour path, or requesting user input if the deviation is way too big.

nav2_controller/src/controller_server.cpp
adivardi: done

nav2_msgs/action/FollowPath.action
adivardi: removed the extra feedback.
I am not sure I have time to work on the extra util right now, should I open an issue so it is not forgotten?

SteveMacenski (approved):

SteveMacenski:
Sorry for the delayed review, I've been having more thought-required PRs filed this week than I had time for thoughts to review them. This looks perfect to me, thanks for finding this bug! 

Please do file file that ticket and please tag it with help wanted, and good first issue, its how alot of new contributors find projects and I think this is a good candidate for that!

adivardi:
Opened.
I am not able to change label though.

PR #4940, opened 2025-02-21
Edit bt nodes ports to stay consistent with the docs

mini-1235:
---

## Basic Info

| ------ | ----------- |

---

- Modified nav2_tree_nodes.xml to stay consistent with the code 

I have raised a PR for the docs https://github.com/ros-navigation/docs.nav2.org/pull/648

---

2 questions here:
- We have a condition called GlobalUpdatedGoal in some places and GloballyUpdatedGoal in others. Should we maintain consistency in the naming, or is this discrepancy intentional? https://docs.nav2.org/configuration/packages/bt-plugins/conditions/GloballyUpdatedGoal.html
- For [speed controller](https://docs.nav2.org/configuration/packages/bt-plugins/decorators/SpeedController.html), there used to be a filter_duration port parameter, which now appears to no longer exist.
https://github.com/ros-navigation/navigation2/blob/4694386872365851c840737515f89e3d409c1d9f/nav2_bt_navigator/src/bt_navigator.cpp#L92-L93
In bt_navigator, this parameter is set to 0.3 and it seems that users cannot define the filter_duration themselves.
Would it be better to move this parameter under bt_navigator in nav2_params.yaml to allow users to configure it independently?

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_passed_goals_action.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_passed_goals_action.hpp
@@ -48,7 +48,6 @@ public:
       BT::OutputPort<geometry_msgs::msg::PoseStampedArray>("output_goals",
           "Goals with passed viapoints removed"),
       BT::InputPort<double>("radius", 0.5, "radius to goal for it to be considered for removal"),
-      BT::InputPort<std::string>("global_frame", "Global frame"),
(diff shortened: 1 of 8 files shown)


SteveMacenski:
>  IMO, the title should be the same with the node name.

Agreed :+1: I'd appreciate it if you made that change!

I think the code should stay as-is. That is consistent with other nodes. 

Please resolve the merge conflict here and I can merge this change in today!

SteveMacenski (approved):

PR #4941, opened 2025-02-21
[AMCL] 50% initialisation (and reset) speed improvement 

doisyg:
---

## Basic Info

| ------ | ----------- |

---

On VERY big maps, the method `map_update_cspace` of AMCL takes ages. Which makes AMCL long to start, or reset (when for instance changing a dynamic parameter needing `reinit_laser = true`)
3 tricks by order of impact:
- using `std::queue` instead of `std::priority_queue`. But this is why this is a draft, I cannot understand why `std::priority_queue` was needed in the first place (and which criteria it used for ordering), so I am not sure.
- avoiding one `CellData` copy with `emplace` instead of push.
- computing MAP_INDEX(map, i, j) once instead of twice.

On a big map, AMCL activation times drops from 22s to 2.4s on my machine.

---

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/src/map/map_cspace.cpp
+++ b/nav2_amcl/src/map/map_cspace.cpp
@@ -110,11 +110,11 @@ get_distance_map(double scale, double max_dist)
  * @brief enqueue cell data for caching
  */
 void enqueue(
-    map_t * map, int i, int j,
-    int src_i, int src_j,
-    std::priority_queue<CellData> & Q,
-    CachedDistanceMap * cdm,
-    unsigned char * marked)
+  map_t * map, int i, int j,
+  int src_i, int src_j,
+  std::priority_queue<CellData> & Q,
+  CachedDistanceMap * cdm,
+  unsigned char * marked)
 {
   const int map_index = MAP_INDEX(map, i, j);
 
@@ -133,8 +133,8 @@ void enqueue(
   map->cells[map_index].occ_dist = distance * map->scale;
 
   Q.emplace(CellData{map, static_cast<unsigned int>(i), static_cast<unsigned int>(j),
-                     static_cast<unsigned int>(src_i), static_cast<unsigned int>(src_j),
-                     map->cells[map_index].occ_dist});
+      static_cast<unsigned int>(src_i), static_cast<unsigned int>(src_j),
+      map->cells[map_index].occ_dist});
 
   marked[map_index] = 1;
 }
--- a/nav2_amcl/src/map/map_cspace.cpp
+++ b/nav2_amcl/src/map/map_cspace.cpp
@@ -25,15 +25,15 @@
 #include "nav2_amcl/map/map.hpp"
 
 /*
- * @class CellData
+ * @struct CellData
  * @brief Data about map cells
  */
-class CellData
+struct CellData
 {
-public:
   map_t * map_;
   unsigned int i_, j_;
   unsigned int src_i_, src_j_;
+  float occ_dist;
 };
 
 /*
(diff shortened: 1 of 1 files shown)


nav2_amcl/src/map/map_cspace.cpp
```diff
@@ -130,18 +129,14 @@ void enqueue(
     return;
   }
 
-  map->cells[MAP_INDEX(map, i, j)].occ_dist = distance * map->scale;
+  const int map_index = MAP_INDEX(map, i, j);
```
SteveMacenski, line 132: This could be done before L120 to remove another MAPINDEX look up on L120.

SteveMacenski (commented):
I agree. I’m also not entirely clear on reading this how the priority queue ever worked considering no comparator for the class/struct was provided. I don’t see any reason except maybe someone thinking it would improve caching efficiency to do patches near each other consecutively. 

@mikeferguson do you see any reason to not merge this?

mikeferguson:
This code was added [15 years ago](https://github.com/ros-planning/navigation/commit/42e07c6d5b040afaa33e53cdef932d01431aa3eb) - by Gerkey (who generally knows what he is doing)

On line 82-87, there is a less-than comparator defined for CellData - it compares the occ_dist of each of the cells. The order in which cells are expanded is important - since we're basically doing wavefront expansion from occupied cells and on L120 we only ever set the occ_dist once. In theory, you will get nearly the same map - but possibly not identically the same map if you don't enforce expanding the closest cells first.

I see at least three possible options here:

1) Verify that the standard queue ends up the same (we know for sure that we enqueue all the occ_dist=0 cells first, outside of the queue processing loop. Then the occ_dist=0 cells would be expanded, enqueuing everything that is occ_dist=~1 from an obstacle. As those are expanded, we get the occ_dist=~2 cells. I think this tracks as being close to the same.... but I think it's not 100% the same since there is also the complication of the CachedDistanceMap - and so we're actually setting occ_dist using a bunch of floating point numbers).

2) Use the standard queue, but if we can't prove things never get unordered, you could change L120 so that we don't early exit, and instead verify that our new candidate source cell isn't closer. This is probably still faster than the priority queue, but I can't say for sure without testing.

3) Stick with the priority queue - but we could also probably speed it up a bit (at the cost of a bit higher memory usage) by inserting the occ_dist into the CellData structure and avoiding 2x MAP_INDEX lookup in every call to the less-than operator.

SteveMacenski:
Ah, I didn't see the comparator in the priority queue constructor nor after the class so I missed it.

I think (3) is the most sensible since it is both the least impactful and easiest to validate. @doisyg what do you think?

doisyg:
Sorry for the delay and thanks @mikeferguson for jumping in!
I am glad I asked, indeed I completely missed the comparator on lines line 82-87, that explains all.
I was originally optimistic as I did not see any obvious behavior changes when doing big map real robot testing without the priority queue, but looking at the structure generated, I can see and confirms that it is different, so better not to touch it.
As for the different index lookup optimizations, I tried without seeing any visible improvement of performance. All the speed improvement were coming for the queue type change. 

However, trying (3), I get a 50% speed improvement with only a small increase of memory!

(below peaks correspond to cspace map computation, I started AMCL normally and then triggered a second computation by changing the scan_topic parameter)

## Original (main)
```
[amcl-3] [INFO] [1744322808.679063828] [amcl]: Received a 10813 X 6743 map @ 0.050 m/pix
[amcl-3] [INFO] [1744322809.587338744] [amcl]: createLaserObject
[amcl-3] Updating map
[amcl-3] Map update took: 13.4365 seconds
```
Screenshot from 2025-04-10 23-04-13

## With comparator optimisation
```
[amcl-3] [INFO] [1744322960.928103210] [amcl]: Received a 10813 X 6743 map @ 0.050 m/pix
[amcl-3] [INFO] [1744322961.837226267] [amcl]: createLaserObject
[amcl-3] Updating map
[amcl-3] Map update took: 7.16248 seconds
```
Screenshot from 2025-04-10 23-01-39

And for the original problem that I was trying to address: switching scan_topic without having to wait a long time on big maps, I am going to submit another PR that just avoid recomputing the cspace map when not needed.

mikeferguson (approved):
Yeah, this looks like a pretty good memory <-> runtime trade off, and a nice cleanup of codebase

SteveMacenski (approved):

PR #4942, opened 2025-02-22
Remove unnecessary sqrt calculations

RasmusLar:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Isaac Sim v4.2.0 |
| Does this PR contain AI generated software? | Yes (GH Copilot) |

---

* I replaced calculations which used hypot with a comparison with the square values, to avoid sqrt calculations to improve performance
* EDIT: Unit tests moved over to #4988
* <strike>While writing unit tests, fixed discrepancy in goal checker orientation, which was checking for `< tolerance` instead of `<= tolerance`, as all the other limit checks are.</strike>
* <strike>Reduced tolerance time for the progress checker unit tests to 0.1 seconds, to reduce test runtime from ~17 to ~7 seconds.</strike>

* I think no documentation update is needed.

* I wrote unit tests and tested in Isaac sim for ~10 minutes

---

* Other performance improvements that I can see, reduced copying values from const references

#### For Maintainers: 

### DIFF
--- a/nav2_controller/include/nav2_controller/plugins/simple_progress_checker.hpp
+++ b/nav2_controller/include/nav2_controller/plugins/simple_progress_checker.hpp
@@ -52,6 +52,11 @@ protected:
    * @param pose Current pose of the robot
    */
   void resetBaselinePose(const geometry_msgs::msg::Pose2D & pose);
+  /**
+   * @brief Sets the radius of the required movement
+   * @param radius Radius of the required movement
+   */
+  void setRadius(const double radius);
 
   static double pose_distance_sqr(
     const geometry_msgs::msg::Pose2D &,
@@ -59,7 +64,7 @@ protected:
 
   rclcpp::Clock::SharedPtr clock_;
 
-  double radius_, radius_sqr_;
+  double radius_sqr_;
   rclcpp::Duration time_allowance_{0, 0};
 
   geometry_msgs::msg::Pose2D baseline_pose_;
--- a/nav2_controller/include/nav2_controller/plugins/simple_progress_checker.hpp
+++ b/nav2_controller/include/nav2_controller/plugins/simple_progress_checker.hpp
@@ -53,13 +53,13 @@ protected:
    */
   void resetBaselinePose(const geometry_msgs::msg::Pose2D & pose);
 
-  static double pose_distance(
+  static double pose_distance_sqr(
     const geometry_msgs::msg::Pose2D &,
     const geometry_msgs::msg::Pose2D &);
 
   rclcpp::Clock::SharedPtr clock_;
 
-  double radius_;
+  double radius_, radius_sqr_;
   rclcpp::Duration time_allowance_{0, 0};
 
   geometry_msgs::msg::Pose2D baseline_pose_;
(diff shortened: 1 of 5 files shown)


nav2_controller/plugins/simple_progress_checker.cpp
```diff
@@ -80,17 +81,17 @@ void SimpleProgressChecker::resetBaselinePose(const geometry_msgs::msg::Pose2D &
 
 bool SimpleProgressChecker::isRobotMovedEnough(const geometry_msgs::msg::Pose2D & pose)
 {
-  return pose_distance(pose, baseline_pose_) > radius_;
```
RasmusLar, line 83: I just thought of an edge case, if `radius_` was negative before, this would always return true.
This is no longer the case, should that be checked?

nav2_controller/include/nav2_controller/plugins/simple_progress_checker.hpp
```diff
@@ -53,13 +53,13 @@ class SimpleProgressChecker : public nav2_core::ProgressChecker
...
 
-  static double pose_distance(
+  static double pose_distance_sqr(
     const geometry_msgs::msg::Pose2D &,
     const geometry_msgs::msg::Pose2D &);
 
   rclcpp::Clock::SharedPtr clock_;
 
-  double radius_;
+  double radius_, radius_sqr_;
```
RasmusLar, line 62: Depending on how negative `radius_` values should be checked, this should also be removed?

nav2_controller/plugins/simple_progress_checker.cpp
RasmusLar: See new commit as a suggestion to solve this, resolve if suggestion is good.

nav2_controller/include/nav2_controller/plugins/simple_progress_checker.hpp
RasmusLar: See new commit as a suggestion to solve this, resolve if suggestion is good.

SteveMacenski:
> I replaced calculations which used hypot with a comparison with the square values, to avoid sqrt calculations to improve performance

I'm all for performance improvements, but can you highlight what the runtime metrics are for these improvements? Most of these operations are small and infrequent, so it kind of surprises me if this improves the performance in a measurable way. If there's not a measurable change, I'd prefer to not merge in additional complexity 

> Reduced tolerance time for the progress checker unit tests to 0.1 seconds, to reduce test runtime from ~17 to ~7 seconds.

I'm a huge fan of this though! Our Ci times are high and small improvements like this really add up quickly.

RasmusLar:
> I'm all for performance improvements, but can you highlight what the runtime metrics are for these improvements? Most of these operations are small and infrequent, so it kind of surprises me if this improves the performance in a measurable way. If there's not a measurable change, I'd prefer to not merge in additional complexity
I only ran them separately, and locally it was like a 20% improvement on the function itself.

I guess the improvements are minor in the whole context of Nav2, but also I thought the complexity increase was minor. I guess the biggest issue would be readability.
I can try to set some time aside to do some tests to see how big of a difference it does in the context of the whole.

> I'm a huge fan of this though! Our Ci times are high and small improvements like this really add up quickly.

I can separate the tests out into a different PR also when I get time

> Thanks for the contribution! Sorry for the delay, I was on a trip to India and then I fell sick due to said trip 😆

Ouch, the one downside about traveling, hope you're better! :smile:

SteveMacenski:
> I can separate the tests out into a different PR also when I get time

Please! That's easy for me to merge 

>  I can try to set some time aside to do some tests to see how big of a difference it does in the context of the whole.

OK!

RasmusLar:
> Please! That's easy for me to merge

Added that in separate PR: #4988 and updated this description

RasmusLar:
And I think you're right, the improvement is negligible. so I will close this :)

PR #4945, opened 2025-02-24
navfn : fix performance issue

Yancey2023:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4945 |
| Robotic platform tested on | reality |

---

`squared_distance(p, goal)` should be calculate when `potential < POT_HIGH`.

none.

test in reality.

---

none

#### For Maintainers: 

### DIFF
--- a/nav2_navfn_planner/src/navfn_planner.cpp
+++ b/nav2_navfn_planner/src/navfn_planner.cpp
@@ -293,11 +293,13 @@ NavfnPlanner::makePlan(
       p.position.x = goal.position.x - tolerance;
       while (p.position.x <= goal.position.x + tolerance) {
         potential = getPointPotential(p.position);
-        double sdist = squared_distance(p, goal);
-        if (potential < POT_HIGH && sdist < best_sdist) {
-          best_sdist = sdist;
-          best_pose = p;
-          found_legal = true;
+        if (potential < POT_HIGH) {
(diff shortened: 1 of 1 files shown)


Yancey2023:
I think this PR should be ok. I don't know how to handle the blocking of this merging.

Yancey2023:
I think this PR should be ok. I don't know how to handle the blocking of this merging.

SteveMacenski:
@Yancey2023 what performance change do you measure with this change? Moving one distance calculation I am surprised to hear has meaningful performance changes

Yancey2023:
@SteveMacenski I'm sorry for causing you confusion. Actually, this won't have a significant performance improvement, but it is indeed a small optimization. I need to implement the function of tracking objects, so when the destination is an obstacle, I need to navigate to the nearest area to the destination, which is implemented in Navfn. In the original method, for grids that are obstacles, the distance to the destination is also calculated, which is unnecessary. My improvement is to only calculate the distance when the destination is not an obstacle.

SteveMacenski:
Got it - what is the speed up that you actually see from this change in your application?

Yancey2023:
Because it's not the performance bottleneck, there's no real performance gain.

SteveMacenski (approved):
OK - I suppose it doesn't hurt!

PR #4948, opened 2025-02-25
Use main function instead of global variable to setup the test environment.

evshary:
---

## Basic Info

| ------ | ----------- |

---

Now the unit test relies on using global variables to setup the environment.
For example, https://github.com/ros-navigation/navigation2/blob/4694386872365851c840737515f89e3d409c1d9f/nav2_planner/test/test_dynamic_parameters.cpp#L60
I think it would be better to use the main function instead.

We discovered that while testing `rmw_zenoh`.
Due to [the issue](https://github.com/ros2/rmw_zenoh?tab=readme-ov-file#crash-when-program-terminates), rmw_zenoh couldn't shutdown in atexit stage. That means calling shutdown in the destructor of a global variable will cause panic.
We believe it would be better to avoid using global variables.

Here is the list of affected tests: https://github.com/ZettaScaleLabs/rmw_zenoh/issues/54

Not required

Run the CI test.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp
+++ b/nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp
@@ -63,10 +63,10 @@ public:
   }
 };
 
-class BTTestEnv
+class RemoveInCollisionGoalsTestFixture : public ::testing::Test
 {
 public:
-  BTTestEnv()
+  static void SetUpTestCase()
   {
     node_ = std::make_shared<rclcpp::Node>("in_collision_goals_test_fixture");
     factory_ = std::make_shared<BT::BehaviorTreeFactory>();
@@ -100,7 +100,7 @@ public:
       "RemoveInCollisionGoals", builder);
   }
 
-  ~BTTestEnv()
+  static void TearDownTestCase()
   {
     delete config_;
     config_ = nullptr;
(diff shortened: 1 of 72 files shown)


SteveMacenski (commented):
@evshary thanks for the PR! 

I broadly speaking agree and approve, though I'd like an explanation for the changes in `nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp` which don't match the same changes made in the other files. 

Otherwise, LGTM and happy to merge once I understand why only this one test is otherwise adjusted

evshary:
Hi @SteveMacenski 

Thank you for the review! Yes, a more detailed explanation for that should be better.
Originally, `RemoveInCollisionGoalsTestFixture` is used to initialize/clean the test environment by `static void SetUpTestCase()` and `static void TearDownTestCase()`. However, the fixture will touch the `rclcpp::Node` and destruct in `atexit` stage because it's a static variable.

In rmw_zenoh, there is a [limitation](https://github.com/ros2/rmw_zenoh?tab=readme-ov-file#crash-when-program-terminates) in that we can't run anything in the `atexit` stage because some context is no longer available, and it will panic. So to avoid initializing/cleaning the test environment with static, I refactored the class and only initialized/cleaned it inside the main function.

Please let me know if you want this in a separate PR or you have any concern on this.

YuanYuYuan:
Hi @evshary! Regarding the fix on _nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp_, I believe this patch could be a solution.

```diff
diff --git a/nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp b/nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp
index 307bd6e3..ea750fe4 100644
--- a/nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp
+++ b/nav2_behavior_tree/test/plugins/action/test_remove_in_collision_goals_action.cpp
@@ -106,6 +106,7 @@ public:
     config_ = nullptr;
     node_.reset();
     success_server_.reset();
+    failure_server_.reset();
     factory_.reset();
   }
```

SteveMacenski (approved):

SteveMacenski:
@evshary can you fix the merge conflict? Then happy to merge this in

Thanks for the help to make this work well with Zenoh!

evshary:
@SteveMacenski It's resolved now. Hope we can make rmw_zenoh pass all the tests in navigation2 in the near future :smile:

PR #4949, opened 2025-02-25
nav2_smac_planner: Fix incorrect conversion between costmap and worlrd coordinates

FrGrQuim:
The Smac planner computes paths using costmap coordinates (cell indexes).
Therefore, conversions between world coordinates and costmap coordinates are required.
The issue was that the world-to-costmap conversion used a floor approximation, while the costmap-to-world conversion behaved as if a ceil approximation had been used.
As a result, the planner's output was often shifted in the negative direction.
This commit corrects the inconsistency to ensure proper alignment between costmap and world coordinates.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/issues/4735 |

---

* I fix the bad conversion cotsmap_cell_index->Worlds in SmacPlanner

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/utils.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/utils.hpp
@@ -45,9 +45,9 @@ inline geometry_msgs::msg::Pose getWorldCoords(
 {
   geometry_msgs::msg::Pose msg;
   msg.position.x =
-    static_cast<float>(costmap->getOriginX()) + (mx - 0.5) * costmap->getResolution();
+    static_cast<float>(costmap->getOriginX()) + (mx + 0.5) * costmap->getResolution();
   msg.position.y =
-    static_cast<float>(costmap->getOriginY()) + (my - 0.5) * costmap->getResolution();
+    static_cast<float>(costmap->getOriginY()) + (my + 0.5) * costmap->getResolution();
   return msg;
 }
 


SteveMacenski (commented):
This was changed in July to supposedly fix the issue in reverse: https://github.com/ros-navigation/navigation2/commit/137b93b33077a0a0057d68711ef653c96049887e

What distribution did you test this against? Main / Rolling / Jazzy is `-`, but humble ? iron is `+` due to backport conflicts so it seems like humble the outlier. 

I'm therefore wondering if this change is actually correct. I think perhaps you need to invert that in Humble / Iron that you're using and the `main` change is incorrect. Or, do you think the previous PR to change it that's in Main/Rolling/Jazzy is wrong?

DylanDeCoeyer-Quimesis:
@SteveMacenski 

I notice that the [Iron version of the test function](https://github.com/ros-navigation/navigation2/blob/iron/nav2_smac_planner/test/test_utils.cpp#L157) does not contain any `assert`, while the [main/jazzy version](https://github.com/ros-navigation/navigation2/blob/main/nav2_smac_planner/test/test_utils.cpp#L161) does. So, it probably explains why the Iron/Humble versions pass the tests.

Regarding the change suggested by @FrGrQuim, I believe that it makes sense whatever the version **but only if `mx` and `my` are integers**. In this case, the first cell (0, 0) should have the coordinates `(0.5, 0.5) * costmap_resolution` (i.e. the center of the cell), and not `(-0.5, -0.5) * costmap_resolution` as it is the case currently.

However, `mx` and `my` are float (continuous map coords) in the `getWorldCoords` and the reciprocal `Costmap2D::worldToMapContinuous`. So, this -0.5/+0.5 added in those two functions does not make sense to me.

For example, if `mx = 0.1` (i.e. 10% of the first cell), origin is (0.0, 0.0), and resolution is 1, I would expect the world coordinate to be `x = origin_x + mx * resolution = 0.1`. No 0.5 should be involved in this calculation, since there is no conversion from "discrete world" to "continuous world".

SteveMacenski:
After analyzing the implementatinos of `worldToMap`, I'm in agreement that `worldToMapContinuous` should not have the `+ 0.5f`. That was introduced in https://github.com/ros-navigation/navigation2/pull/4255. @BriceRenaudeau can you review that again and let us know if you agree that this was added in error? I don't see that any other `worldToMap` implementation uses the `0.5` mechanic other than that one added, so that seems suspect.

---

On `mapToWorld`: the `worldToMap` in general will `floor` any value when converting to an `int` or `unsigned int`, so adding in the `+0.5` cells helps deal with that w.r.t. getting the world dimension later. If we have continuous coordinates that were changed in Brice's PR from April, 2024 then the conversion to world again after planning needs to also adjust in its implementation. 

I think that's what @HovorunBh was attempting to fix in https://github.com/ros-navigation/navigation2/pull/4506 about 2-3 months later, but changed the sign because that visually fixed the issue. @HovorunBh can you let us know about your thinking/testing for that? Was it just a visual comparison (and would just removing the `0.5` term also resolve)?

Assuming @BriceRenaudeau and @HovorunBh agree - I think that's a sensible move to remove from both terms.

BriceRenaudeau:
@SteveMacenski, You are right, this is clearly a mistake. I might have confused with mapToWorld...

SteveMacenski:
@FrGrQuim or @DylanDeCoeyer-Quimesis - mind opening a PR to adjust respectively?

HovorunBh:
My fix in https://github.com/ros-navigation/navigation2/pull/4506 was indeed based on visual testing and observed misalignments in planned paths. Additionally, I noticed an inconsistency in how +0.5 was handled between mapToWorld and worldToMap, which led to the argument in my PR:
>When we convert world to map coordinates we add 0.5
> Therefore when we convert map to world I think we should subtract 0.5. Otherwise we end up in the next costmap cell

I didn't go any deeper trying to verify whether simply removing the +0.5f from worldToMapContinuous would have had the same effect. But your explanation makes sense, I think it is a good idea and it will resolve both issues!

DylanDeCoeyer-Quimesis:
> @FrGrQuim or @DylanDeCoeyer-Quimesis - mind opening a PR to adjust respectively?
> 
> @BriceRenaudeau thanks for the fast feedback!

I'll push a PR within a few hours :-)

PR #4950, opened 2025-02-25
Nav2 route server start pose orientation scorer

alexanderjyuen:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Polymath Robotics' Ackermann agriculture client |

---

Presently there is no edge scorer that restricts or penalizes routes that start with an orientation that is very different from the robot start pose. A new edge scorer has been added to do that comparison and return false if a route pose is very different from the start pose

In order to do this, two major additions were made outside of including `start_pose_orientation_scorer.cpp` and start_pose_orientation_scorer.hpp` this includes:
- Adding a function to check if the node ID is the start node
- Passing the bool identifying an edge being the start edge all the way from the route server to the edge scorers
- Passing the tf buffer from route planner down to the edge scorers
- Modifying all edge scorers to take new arguments

Need to add configuration guide on full release of `route_server`

Unit tests were written to cover the major use case

- Documentation and more comprehensive testihng

#### For Maintainers: 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209670265483346


nav2_route/include/nav2_route/edge_scorer.hpp
```diff
@@ -63,7 +66,8 @@ class EdgeScorer
    * @return If edge is valid
    */
   bool score(
-    const EdgePtr edge, const geometry_msgs::msg::PoseStamped & goal_pose, bool final_edge,
+    const EdgePtr edge, const geometry_msgs::msg::PoseStamped & goal_pose, bool start_edge,
```
SteveMacenski, line 69: If we're going to add things like this, I think we should create an `enum` so more than be added without breaking API in the future. We can have an enum for `SpecialEdgeType` with start, final as current ones. If neither, then the default can be `NONE`.

nav2_route/include/nav2_route/interfaces/edge_cost_function.hpp
```diff
@@ -55,14 +56,17 @@ class EdgeCostFunction
...
+    const std::shared_ptr<tf2_ros::Buffer> tf_buffer,
     const std::string & name) = 0;
 
   /**
    * @brief Main scoring plugin API
    * @param edge The edge pointer to score, which has access to the
    * start/end nodes and their associated metadata and actions
    */
-  virtual bool score(const EdgePtr edge, const geometry_msgs::msg::PoseStamped & goal_pose, bool final_edge, float & cost) = 0;
+  virtual bool score(
```
SteveMacenski, line 67: Then propagated here as well

nav2_route/src/edge_scorer.cpp
```diff
@@ -58,11 +60,13 @@ EdgeScorer::EdgeScorer(nav2_util::LifecycleNode::SharedPtr node)
       throw ex;
     }
   }
+
+  tf_buffer_ = tf_buffer;
```
SteveMacenski, line 64: Does the edge scorer manager need the tf buffer? I think this can be removed (and in the hpp file)

nav2_route/include/nav2_route/plugins/edge_cost_functions/start_pose_orientation_scorer.hpp
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2024, Polymath Robotics Inc.
```
SteveMacenski, line 1: Update to 2025 (with cpp file too)

nav2_route/include/nav2_route/plugins/edge_cost_functions/start_pose_orientation_scorer.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
+#include "tf2/utils.h"
+#include "angles/angles.h"
+
+namespace nav2_route
+{
+
+/**
+ * @class StartPoseOrientationScorer
+ * @brief Scores initial edge edge by comparing the orientation of the robot's current pose and the orientation of the edge
```
SteveMacenski, line 38:
```suggestion
 * @brief Scores initial edge by comparing the orientation of the robot's current pose and the orientation of the edge
```

Please also provide additional information about how it scores. I don't think it actually scores as much as it rejects if > than a tolerance. (also the goal orientation one needs to be populated as well https://github.com/ros-navigation/navigation2/blob/d02bb473a743c3a44cb9291d541bb862d9232bca/nav2_route/include/nav2_route/plugins/edge_cost_functions/goal_orientation_scorer.hpp#L36)

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <memory>
+#include <string>
+
+#include "nav2_route/plugins/edge_cost_functions/start_pose_orientation_scorer.hpp"
+
+namespace nav2_route
+{
+
+void StartPoseOrientationScorer::configure(
```
SteveMacenski, line 23: I wanted to mention that this could cause you some real headaches with non-liberal values of the tolerance. If localization updates mess with your orientation, initial orientation is very wrong, initial orientation is not aligned with the graph when the robot starts up, or robot drives off the route for some reason, your orientation could not align well with any of the edges and cause them all to fail.

I wonder if you really want to apply a continuous cost function proportionate to the orientation instead of a straight-up rejection. It would then allow for much shorter paths to be able to use the reverse direction, which maybe isn't something you want. I think there should be modes (or separate plugins?) for both options. I could see both being useful and I think you would also have applications of this work that you would want to use either/or (such as for omni, differential, or small ackermann platforms)

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
```diff
@@ -0,0 +1,88 @@
...
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    getName() + ".robot_frame", rclcpp::ParameterValue("base_link"));
+
+  orientation_tolerance_ = node->get_parameter(getName() + ".orientation_tolerance").as_double();
+  route_frame_ = node->get_parameter(getName() + ".route_frame").as_string();
+  robot_frame_ = node->get_parameter(getName() + ".robot_frame").as_string();
+
+  tf_buffer_ = tf_buffer;
+
```
SteveMacenski, line 47: Remove empty line

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
```diff
@@ -0,0 +1,88 @@
...
+  const std::string & name)
+{
+  RCLCPP_INFO(node->get_logger(), "Configuring start pose orientation scorer.");
+  name_ = name;
+  logger_ = node->get_logger();
+
+  nav2_util::declare_parameter_if_not_declared(
+    node,
+    getName() + ".orientation_tolerance", rclcpp::ParameterValue(M_PI / 2.0));
+  nav2_util::declare_parameter_if_not_declared(
```
SteveMacenski, line 35: Shouldnt thins and the robot frame exist for the server itself (rather than the plugin)? If so, you don't need to duplicate parameters, you can simply get the server's values itself from `get_parameter` without declaring

SteveMacenski (changes requested):
Some medium sized items, but all should be technically easy (just touches a few files). Overall, looks good in concept

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
alexanderjyuen: fixed in https://github.com/ros-navigation/navigation2/pull/4950/commits/be1a59f4c9a777bb554100f7106ab6d7b73747ba

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
alexanderjyuen: That makes sense, should we fix that in this PR or open a new one for it. If its anything more sophisticated than a weight next to the deviation from the start pose, it probably deserves it own PR, otherwise I could just multiple the difference with some weight and have that weight be an input parameter to the plugin.

I've added a TODO for the time being

nav2_route/src/edge_scorer.cpp
alexanderjyuen: fixed in https://github.com/ros-navigation/navigation2/pull/4950/commits/b54c47b3b05273088f55b2dbfda275e3e034a059

nav2_route/include/nav2_route/plugins/edge_cost_functions/start_pose_orientation_scorer.hpp
alexanderjyuen: fixed in https://github.com/ros-navigation/navigation2/pull/4950/commits/be1a59f4c9a777bb554100f7106ab6d7b73747ba

nav2_route/include/nav2_route/interfaces/edge_cost_function.hpp
alexanderjyuen: enum added to types.hpp and propogated to all scorers and tests

nav2_route/include/nav2_route/plugins/edge_cost_functions/start_pose_orientation_scorer.hpp
SteveMacenski: ^ still open

nav2_route/include/nav2_route/edge_scorer.hpp
```diff
@@ -63,7 +66,7 @@ class EdgeScorer
    * @return If edge is valid
    */
   bool score(
-    const EdgePtr edge, const geometry_msgs::msg::PoseStamped & goal_pose, bool final_edge,
+    const EdgePtr edge, const geometry_msgs::msg::PoseStamped & goal_pose, EdgeType edge_type,
```
SteveMacenski, line 69: EdgeType a const reference?

nav2_route/include/nav2_route/interfaces/edge_cost_function.hpp
```diff
@@ -55,14 +56,17 @@ class EdgeCostFunction
...
+    const EdgePtr edge, const geometry_msgs::msg::PoseStamped & goal_pose,
+    EdgeType edge_type, float & cost) = 0;
```
SteveMacenski, line 69: EdgeType a const reference?

nav2_route/src/route_planner.cpp
```diff
@@ -192,4 +197,20 @@ bool RoutePlanner::isGoal(const NodePtr node)
...
+  return node->nodeid == start_id_;
+}
+
+nav2_route::EdgeType RoutePlanner::classifyEdge(const EdgePtr edge)
+{
+  if (isStart(edge->start)) {
+    return nav2_route::EdgeType::START;
+  } else if (isGoal(edge->end)) {
+    return nav2_route::EdgeType::END;
+  } else {
```
SteveMacenski, line 211: Remove the `else` and just have `return EdgeType::NONE;` so that we remove a branch - its the default return if not start/goal. 

Also, you can remove the `nav2_route` namespace in the other conditions as well, this node is within that namespace already

nav2_route/src/route_planner.cpp
```diff
@@ -161,7 +164,9 @@ bool RoutePlanner::getTraversalCost(
     return true;
   }
 
-  return edge_scorer_->score(edge, goal, isGoal(edge->end), score);
+  enum EdgeType edge_type = classifyEdge(edge);
+
+  return edge_scorer_->score(edge, goal, edge_type, score);
```
SteveMacenski, line 169:
```suggestion
  return edge_scorer_->score(edge, goal, classifyEdge(edge), score);
```

And remove the previous line, its more concise and I think still very clear

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
alexanderjyuen: added the described functionality and the corresponding unit tests in: https://github.com/ros-navigation/navigation2/pull/4950/commits/4bb3131c0c334066680071f832a43a8124065e97

nav2_route/include/nav2_route/plugins/edge_cost_functions/start_pose_orientation_scorer.hpp
alexanderjyuen: added docstring, better describing the edge scorers in https://github.com/ros-navigation/navigation2/pull/4950/commits/4124841db6d217e66195b101018c3459f41f621c

nav2_route/src/plugins/edge_cost_functions/costmap_scorer.cpp
```diff
@@ -72,10 +73,12 @@ void CostmapScorer::prepare()
   }
 }
 
+// TODO(sm) does this critic make efficiency sense at a
```
SteveMacenski, line 76: Remove TODO

nav2_route/include/nav2_route/plugins/edge_cost_functions/goal_orientation_scorer.hpp
```diff
@@ -33,7 +33,9 @@ namespace nav2_route
 
 /**
  * @class GoalOrientationScorer
- * @brief Scores final edge by comparing the
+ * @brief Rejects or accepts final edge depending on the edge's deviation
```
SteveMacenski, line 36: What if the request uses the goal_id (node ID) instead of the goal (pose)? `goal_pose` would be empty because `goal->goal` is not set. 

I think we should check if `goal_pose == geometry_msgs::msg::PoseStamped()` and throw an exception if so, since its not set and thus this is an invalid critic to use.

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
```diff
@@ -0,0 +1,95 @@
...
+        return false;
+      }
+    }
+
+    cost = orientation_weight_ * static_cast<float>(d_yaw);
+  }
+  return true;
+}
+
+geometry_msgs::msg::PoseStamped StartPoseOrientationScorer::getRobotPose()
```
SteveMacenski, line 77: Teh goal intent extractor has the start pose as either the action message's start or the current pose from TF depending on the request settings, in `findStartandGoal`. I think we should get the start pose that we should use in a new API `getStart` and pass that through the server to the scoring plugins with the goal pose. That way we're using the proper start pose that the user requested. Also, TF lookups like this each `score()` edge would be very computationally expensive, we would want to compute that once and cache it instead anyway. 

Similarly to the goal orientation scorer, if `== PoseStamped()` because its not used as we're working with the Node ID search not poses, then throw exception as an invalid plugin to use for the request type. 

Also, add in the docstrings for both of these plugins that its only valid when used with pose-request types (not node request types)

SteveMacenski: I think that would also remove the need to obtain the route and base frames

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    double start_orientation = tf2::getYaw(robot_pose.pose.orientation);
+    double d_yaw = std::abs(angles::shortest_angular_distance(edge_orientation, start_orientation));
+
+    if (use_orientation_threshold_) {
+      if (d_yaw > orientation_tolerance_) {
+        return false;
+      }
+    }
+
+    cost = orientation_weight_ * static_cast<float>(d_yaw);
```
SteveMacenski, line 72: Can we add this same weighted scoring to the goal pose orientation?

SteveMacenski (changes requested):
A couple of asks, but also a couple of bugs I found after refamiliarizing myself with this project this week w.r.t. request types using poses vs IDs.

Hate to make more work for you, but I think you'll thank yourself later when you use this on other projects

nav2_route/include/nav2_route/plugins/edge_cost_functions/goal_orientation_scorer.hpp
SteveMacenski: I think an new exception needs to be made in `nav2_core/route_exception.hpp`, `catch` in the server for it, and enum for it in the `.action` files to communicate it . Something like `InvalidCriticUse`

alexanderjyuen: I created a `RouteData` struct that has the start and goal node id, the start and goal pose (if relevant) and a bool flag that indicates whether we are using the poses or not, which is populated using the action goal's own `use_poses`. 

For both the goal pose orientation scorer and start pose orientation scorer, we throw the `InvalidCriticUse` exception (which I've also added), if `route_data.goal_poses` is false

Seem like a valid solution? Only wrinkle I see is if the user has set `use_poses` to true, but the start and goal `PoseStamped` in the action goals is not set

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
alexanderjyuen: modified all the edge scorers and this one in particular, to use the `RouteData` struct that contains all the necessary information. As you've recommended I've populated the `start_pose` field in the `RouteData` struct with the newly added `getStart()` goal intent extractor method. Deleted a few of the irrelevant parameters and the `getRobotPose()` method

nav2_route/src/plugins/edge_cost_functions/start_pose_orientation_scorer.cpp
alexanderjyuen: Added the same scoring and the corresponding tests

alexanderjyuen:
@SteveMacenski this is more or less wrapped up, take a peek when you get a chance

nav2_route/src/route_server.cpp
```diff
@@ -215,10 +215,17 @@ Route RouteServer::findRoute(
     route.route_cost = 0.0;
     route.start_node = &graph_.at(start_route);
   } else {
+    //populate struct with start id, goal id, start pose, goal pose, and whether we are using pose
```
SteveMacenski, line 218:
```suggestion
    // Populate request data (start & goal id, start & goal pose, if set) for routing
```

nav2_route/src/route_server.cpp
```diff
@@ -297,6 +304,10 @@ RouteServer::processRouteRequest(
     result->error_code = ActionT::Goal::INDETERMINANT_NODES_ON_GRAPH;
     result->error_msg = ex.what();
     action_server->terminate_current(result);
+  } catch (nav2_core::InvalidCriticUse & ex) {
+      exceptionWarning(goal, ex);
+      result->error_code = ActionT::Goal::INVALID_CRITIC_USE;
```
SteveMacenski, line 309: Add in `result->error-Msg = ex.what()`

nav2_route/include/nav2_route/types.hpp
```diff
@@ -203,6 +203,15 @@ struct Route
   float route_cost{0.0};
 };
 
+struct RouteData
```
SteveMacenski, line 206: Rename to `RouteRequest`, this is named a bit too generically. Make sure to update the variable uses from `route_data` appropriately.

Also, add doxygen for this struct

SteveMacenski (changes requested):
Generally looks near completion to me, just a couple of nits

nav2_route/src/route_server.cpp
alexanderjyuen: Added this to all the caught exceptions

nav2_core/include/nav2_core/route_exceptions.hpp
```diff
@@ -71,6 +71,13 @@ class IndeterminantNodesOnGraph : public RouteException
   : RouteException(description) {}
 };
 
+class InvalidCriticUse : public RouteException
```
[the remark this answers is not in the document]
alexanderjyuen: Changed all `critic` to `edge_scorer`

nav2_route/include/nav2_route/types.hpp
```diff
@@ -271,6 +286,13 @@ struct ReroutingState
   }
 };
 
+enum class EdgeType
```
SteveMacenski, line 289: Doxygen here too

SteveMacenski (approved):

nav2_route/include/nav2_route/types.hpp
SteveMacenski: Last thing and I'll merge

SteveMacenski (approved):

Issue #4601, opened 2024-08-07
Issue with Bidirectional Acceleration and Deceleration Settings in MPPI Controller


osama-z-salah:
I am using an MPPI controller to drive a robot capable of traveling bidirectionally with equal efficiency. I want to have different values for acceleration and deceleration. To achieve this, I set the acceleration by setting ax_max to 0.5 and the deceleration by setting ax_min to -1.5. These values are respected when the robot is traveling forward. However, when the robot is traveling backward, the acceleration value becomes the deceleration value, and vice versa.

SteveMacenski:
This is the block of code of interest https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp#L85-L113

What you mention here is very sensible and I think that should be enable-able. I'd be open to your suggestions, but I think a potential solution would be to have a new parameter option for bidirectionality - similar to `mode` in `PathAngleCritic` - which in this block can be checked and flip the signs as needed. The parameter should be stored in `control_constraints_`. 

Optionally, we could make that default behavior and just have the logic in that block for checking if we're slowing our total speed (apply deceleration) or increasing our total speed (apply acceleration), regardless of signed direction when applying limits.

I imagine that's what @doisyg would also want. @tonynajjar do you think that's reasonable to have it default have that behavior as a general case for your non-bidirectional case? I'd always prefer less parameters if it can be avoided, but it does seem like it should be applied in this way for all cases I think since its more about 'speed' than direction.

doisyg:
> I imagine that's what @doisyg would also want.

Yes!
Maybe some logic to reuse from this: https://github.com/ros-navigation/navigation2/pull/3529

SteveMacenski:
I think that is _precisely_ code to reuse here! @osama-z-salah can you create a PR for this? Happy to merge!

AarishShah22:
Hi, I would like to work on this issue! I have gone through the conversation above and will reuse the code from #3529. Thanks!

SteveMacenski:
Great, I'd love the contribution @AarishShah22 !

PR #4954, opened 2025-03-01
Fix mppi bidirectional settings

mini-1235:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4601  |

---

- Fix the acceleration and deceleration when going backward as mentioned in the issue, mostly copied from #3529 
- Regarding the performance issue mentioned in the issue, I ran benchmark and observed no significant difference from the original(tested with different parameters)

None

Tested on sim, wrote some new unit tests 

---

Maybe more unit tests? Looking for some suggestions before proceeding

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -81,7 +81,7 @@ public:
     unsigned int n_cols = state.vx.cols();
 
     for (unsigned int i = 1; i < n_cols; i++) {
-      Eigen::ArrayXf lower_bound_vx = (state.vx.col(i - 1) >
+      auto lower_bound_vx = (state.vx.col(i - 1) >
         0).select(state.vx.col(i - 1) + min_delta_vx,
         state.vx.col(i - 1) - max_delta_vx);
       Eigen::ArrayXf upper_bound_vx = (state.vx.col(i - 1) >
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -75,6 +75,7 @@ public:
     float max_delta_vx = model_dt_ * control_constraints_.ax_max;
     float min_delta_vx = model_dt_ * control_constraints_.ax_min;
     float max_delta_vy = model_dt_ * control_constraints_.ay_max;
+    float min_delta_vy = model_dt_ * control_constraints_.ay_min;
     float max_delta_wz = model_dt_ * control_constraints_.az_max;
 
     unsigned int n_cols = state.vx.cols();
@@ -92,7 +93,7 @@ public:
 
       if (is_holo) {
         state.cvy.col(i - 1) = state.cvy.col(i - 1)
-          .cwiseMax(state.vy.col(i - 1) - max_delta_vy)
+          .cwiseMax(state.vy.col(i - 1) + min_delta_vy)
           .cwiseMin(state.vy.col(i - 1) + max_delta_vy);
         state.vy.col(i) = state.cvy.col(i - 1);
       }
@@ -114,7 +115,7 @@ public:
 protected:
   float model_dt_{0.0};
   models::ControlConstraints control_constraints_{0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
-    0.0f};
+    0.0f, 0.0f};
 };
 
 /**
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -102,7 +102,7 @@ public:
         Eigen::ArrayXf lower_bound_vy = (state.vy.col(i - 1) >
           0).select(state.vy.col(i - 1) + min_delta_vy,
           state.vy.col(i - 1) - max_delta_vy);
-        Eigen::ArrayXf upper_bound_vy = (state.vy.col(i - 1) >
+        auto upper_bound_vy = (state.vy.col(i - 1) >
           0).select(state.vy.col(i - 1) + max_delta_vy,
           state.vy.col(i - 1) - min_delta_vy);
         state.cvy.col(i - 1) = state.cvy.col(i - 1)
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -84,7 +84,7 @@ public:
       auto lower_bound_vx = (state.vx.col(i - 1) >
         0).select(state.vx.col(i - 1) + min_delta_vx,
         state.vx.col(i - 1) - max_delta_vx);
-      Eigen::ArrayXf upper_bound_vx = (state.vx.col(i - 1) >
+      auto upper_bound_vx = (state.vx.col(i - 1) >
         0).select(state.vx.col(i - 1) + max_delta_vx,
         state.vx.col(i - 1) - min_delta_vx);
 
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -99,7 +99,7 @@ public:
       state.wz.col(i) = state.cwz.col(i - 1);
 
       if (is_holo) {
-        Eigen::ArrayXf lower_bound_vy = (state.vy.col(i - 1) >
+        auto lower_bound_vy = (state.vy.col(i - 1) >
           0).select(state.vy.col(i - 1) + min_delta_vy,
           state.vy.col(i - 1) - max_delta_vy);
         auto upper_bound_vy = (state.vy.col(i - 1) >
--- a/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
@@ -81,9 +81,16 @@ public:
     unsigned int n_cols = state.vx.cols();
 
     for (unsigned int i = 1; i < n_cols; i++) {
+      Eigen::ArrayXf lower_bound_vx = (state.vx.col(i - 1) >
+        0).select(state.vx.col(i - 1) + min_delta_vx,
+        state.vx.col(i - 1) - max_delta_vx);
+      Eigen::ArrayXf upper_bound_vx = (state.vx.col(i - 1) >
+        0).select(state.vx.col(i - 1) + max_delta_vx,
+        state.vx.col(i - 1) - min_delta_vx);
+
       state.cvx.col(i - 1) = state.cvx.col(i - 1)
-        .cwiseMax(state.vx.col(i - 1) + min_delta_vx)
-        .cwiseMin(state.vx.col(i - 1) + max_delta_vx);
+        .cwiseMax(lower_bound_vx)
+        .cwiseMin(upper_bound_vx);
       state.vx.col(i) = state.cvx.col(i - 1);
 
       state.cwz.col(i - 1) = state.cwz.col(i - 1)
@@ -92,9 +99,15 @@ public:
       state.wz.col(i) = state.cwz.col(i - 1);
 
       if (is_holo) {
+        Eigen::ArrayXf lower_bound_vy = (state.vy.col(i - 1) >
+          0).select(state.vy.col(i - 1) + min_delta_vy,
+          state.vy.col(i - 1) - max_delta_vy);
+        Eigen::ArrayXf upper_bound_vy = (state.vy.col(i - 1) >
+          0).select(state.vy.col(i - 1) + max_delta_vy,
+          state.vy.col(i - 1) - min_delta_vy);
         state.cvy.col(i - 1) = state.cvy.col(i - 1)
-          .cwiseMax(state.vy.col(i - 1) + min_delta_vy)
-          .cwiseMin(state.vy.col(i - 1) + max_delta_vy);
+          .cwiseMax(lower_bound_vy)
+          .cwiseMin(upper_bound_vy);
         state.vy.col(i) = state.cvy.col(i - 1);
       }
     }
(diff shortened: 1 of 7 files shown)


nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -72,8 +72,9 @@ class MotionModel
   virtual void predict(models::State & state)
   {
     const bool is_holo = isHolonomic();
-    float max_delta_vx = model_dt_ * control_constraints_.ax_max;
-    float min_delta_vx = model_dt_ * control_constraints_.ax_min;
+
+    float max_delta_vx = 0.0;
```
SteveMacenski, line 76: Why were these changed?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -86,6 +87,15 @@ class MotionModel
...
         float & cvx_curr = state.cvx(j, i - 1);
+        // Accelerating if magnitude of cvx is greater than vx_last and they have the same sign
+        // Decelerating otherwise
+        if (abs(cvx_curr) >= abs(vx_last) && cvx_curr * vx_last >= 0.0) {
+          max_delta_vx = control_constraints_.ax_max * model_dt_;
+          min_delta_vx = -control_constraints_.ax_max * model_dt_;
+        } else {
+          max_delta_vx = -control_constraints_.ax_min * model_dt_;
+          min_delta_vx = control_constraints_.ax_min * model_dt_;
+        }
```
SteveMacenski, line 98: I think this is also missing for `Y` direction as well

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -86,6 +87,15 @@ class MotionModel
       for (unsigned int j = 0; j != n_rows; j++) {
         float vx_last = state.vx(j, i - 1);
         float & cvx_curr = state.cvx(j, i - 1);
+        // Accelerating if magnitude of cvx is greater than vx_last and they have the same sign
+        // Decelerating otherwise
+        if (abs(cvx_curr) >= abs(vx_last) && cvx_curr * vx_last >= 0.0) {
+          max_delta_vx = control_constraints_.ax_max * model_dt_;
+          min_delta_vx = -control_constraints_.ax_max * model_dt_;
```
SteveMacenski, line 94: Shouldn't this be `ax_min` not max? We are setting the bounds of max acceleration, and max deceleration when we're currently accelerating.

I think this should just be 

```
min_delta_vx = control_constraints_.ax_min * model_dt_;
```

(the usual)

Ditto issue in the `else` condition. I think that should be the one inverting the signs and/or the value positions

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -86,6 +87,15 @@ class MotionModel
       for (unsigned int j = 0; j != n_rows; j++) {
         float vx_last = state.vx(j, i - 1);
         float & cvx_curr = state.cvx(j, i - 1);
+        // Accelerating if magnitude of cvx is greater than vx_last and they have the same sign
+        // Decelerating otherwise
```
SteveMacenski, line 91: I don't believe this is true. There are four potential states:
- Accelerating forward
- Decelerating forward
- Accelerating in reverse
- Decelerating in reverse

I don't think this properly handles all 4 cases. 

@Ayush1285 do you mind taking a look / reviewing this as well? Your recent changes touch this part of code as well

SteveMacenski (changes requested):

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
mini-1235: Assigning here is meaningless, we have moved it into the for loop, I can do it like this or `float max_delta_vx;` if you prefer

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
mini-1235: I didn't see the `ay_min` as well as `az_min` at first, so I only deal with the X direction, are we assuming that `ay_min = - ay_max` here?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
mini-1235: I am actually following the convention here 
https://github.com/ros-navigation/navigation2/blob/443915ca8e6622e2e953ee8ef8168fc7ee0ae716/nav2_velocity_smoother/src/velocity_smoother.cpp#L263-L268
shouldn't we stay consistent for that?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: Sure, I'll review it.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: Agreed its meaningless, but vy and wz didn't have this change, so we should make it consistent. Remove all or revert the vx ones please

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: It looks like ay_min is accidentally missing and should be added. If ax_max is max acceleration and ax_min is max deceleration, then the Y axis should have the same as another translational motion axis. 

The max acceleration should be applied for moving full-left and full-right (with appropriate sign changes to accomodate); while the deceleration is applied to slowing down movements both left and right.

Accelerating forward is ax_max, decelerating forward is ax_min. Accelerating in reverse is -ax_max and decelerating in reverse is -ax_min.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: @mini-1235 do you understand the 4 cases I'm laying out and how this currently doesn't implement 4 cases?

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski: Writing this out as part of the process. Accelerations min/max are signed (i.e. a_max is positive, a_min is negative). 

I see the following cases, centered around how we find the viable bounds to place on `cvx_curr` for what is dynamically feasible. We start with our current velocity from the last iteration, `vx_last`:

1. Moving forward, meaning  `vx_last` which is positive
* +ax_max * model_dt for maximum forward velocity possible while accelerating (numerically highest value)
* +ax_min * model_dt for minimum forward velocity possible while decelerating (numerically lowest value)
* Thus we clamp in the low-high range of (vx_last + ax_min * model_dt, vx_last  + ax_max * model_dt)

2. Moving in reverse, meaning `vx_last` which is negative
* -ax_max * model_dt for maximum reverse velocity possible while accelerating (numerically lowest value)
* -ax_min * model_dt for minimum reverse velocity possible while decelerating  (numerically highest value)
* Thus we clamp in the low-high range of (vx_last  - ax_max * model_dt, vx_last  - ax_min * model_dt)

If the robot is moving forward but the next velocity is in reverse due to random sampling, that should cover the scope as the bounds are already placed. 

So, I think that really this should only be looking at the sign of `vx_last` to know what to populate for `min_delta_vx` and `max_delta_vx`. I don't see why a comparison with `cvx_curr` or  ` cvx_curr * vx_last >= 0.0` is necessary here.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
Ayush1285: > So, I think that really this should only be looking at the sign of vx_last to know what to populate for min_delta_vx and max_delta_vx. I don't see why a comparison with cvx_curr or  cvx_curr * vx_last >= 0.0 is necessary here. Does that make sense?

Yes, the sign of vx_last should be enough to determine the delta values.

Ayush1285: > It looks like ay_min is accidentally missing and should be added.

When I was working on adding constraints, there was some discussion around it, as I remember. 
https://github.com/ros-navigation/navigation2/pull/4352#discussion_r1612237016
https://github.com/ros-navigation/navigation2/pull/4352#discussion_r1607142308

mini-1235: @SteveMacenski I have gone through the conversations in #4352, what are your current thoughts for `ay_min`? I think I should add it here(?)

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
mini-1235: Yes, this makes sense to me. Thanks for the detail explanation, I will fix this

SteveMacenski:
Sorry for my delayed review on this PR, I'm behind in reviews and wanted to get a few of your open PRs highly responsive than having them all be flaky :wink: I will look at this for sure tomorrow

@Ayush1285 can you give this a look since its highly correlated to your recent work.

SteveMacenski (approved):
Tentatively, I think this is correct, thanks! Though @Ayush1285 might have efficiency questions

Can you run this and check the performance change of the controller with this change? 

I see L88/L106 are marked as not having coverage. Can you check that you test the reverse case?

mini-1235:
> Sorry for my delayed review on this PR, I'm behind in reviews and wanted to get a few of your open PRs to be highly responsive than having them all be flaky 😉 Now these will be highly responsive

No worries, I am focusing on the other 2 PRs as well

> Can you run this and check the performance change of the controller with this change?
I see L88/L106 are marked as not having coverage. Can you check that you test the reverse case?

Yes, I am planning to do these as well as the documentation PR for `ay_min` within this week

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -75,14 +75,22 @@ class MotionModel
     float max_delta_vx = model_dt_ * control_constraints_.ax_max;
     float min_delta_vx = model_dt_ * control_constraints_.ax_min;
     float max_delta_vy = model_dt_ * control_constraints_.ay_max;
+    float min_delta_vy = model_dt_ * control_constraints_.ay_min;
     float max_delta_wz = model_dt_ * control_constraints_.az_max;
 
     unsigned int n_cols = state.vx.cols();
 
     for (unsigned int i = 1; i < n_cols; i++) {
+      Eigen::ArrayXf lower_bound_vx = (state.vx.col(i - 1) >
```
Ayush1285, line 84: Use auto instead of Eigen::ArrayXf. Assigning to Eigen::ArrayXf will immediately evaluate the template expression, resulting in a copy of the elements.

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -75,14 +75,22 @@ class MotionModel
...
+        0).select(state.vx.col(i - 1) + min_delta_vx,
+        state.vx.col(i - 1) - max_delta_vx);
+      Eigen::ArrayXf upper_bound_vx = (state.vx.col(i - 1) >
```
Ayush1285, line 87: Same here also. And evaluate the performance of unaryExpr vs select for these cases.

Ayush1285 (changes requested):

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski, line 84:
```suggestion
      auto lower_bound_vx = (state.vx.col(i - 1) >
```

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
SteveMacenski, line 87:
```suggestion
      auto upper_bound_vx = (state.vx.col(i - 1) >
```

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -91,9 +99,15 @@ class MotionModel
       state.wz.col(i) = state.cwz.col(i - 1);
 
       if (is_holo) {
+        Eigen::ArrayXf lower_bound_vy = (state.vy.col(i - 1) >
```
SteveMacenski, line 102:
```suggestion
        auto lower_bound_vy = (state.vy.col(i - 1) >
```

nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp
```diff
@@ -91,9 +99,15 @@ class MotionModel
...
+          0).select(state.vy.col(i - 1) + min_delta_vy,
+          state.vy.col(i - 1) - max_delta_vy);
+        Eigen::ArrayXf upper_bound_vy = (state.vy.col(i - 1) >
```
SteveMacenski, line 105:
```suggestion
        auto upper_bound_vy = (state.vy.col(i - 1) >
```

SteveMacenski:
Hey, I'm going to do the perf testing now, I updated the `auto`s and going to do some testing shortly in order to push this through!

SteveMacenski:
Testing is inconclusively different. Some tests it added 0.3ms, others 0.1ms. Either way, the worst case I see is a 5% performance drop from 7.7ms to 8.1ms and that feels inconsequential in comparison to the usefulness of this feature while still maintaining a runtime speed of 125Hz.

Merging!

mini-1235:
> Hey, I'm going to do the perf testing now, I updated the `auto`s and going to do some testing shortly in order to push this through! I'll take this one over 😄

Ok! Thanks for the help

Tacha-S:
Will this change be merged into the jazzy branch?

SteveMacenski:
None of this is API compatible with Jazzy since Jazzy uses xtensor and this uses Eigen. However, if you wanted this feature, I'd be happy to review and merge a manual backport doing this analogously with xtensor for Jazzy's branch

Issue #4837, opened 2025-01-09
Feature Request: Service Introspection


ottojo:
#### Feature description
It would be useful to be able to use [Service Introspection](https://docs.ros.org/en/jazzy/Tutorials/Demos/Service-Introspection.html) introduced in Jazzy with nav2 nodes (such as the costmap server). This requires some configuration for both the client and server, which is currently not enabled in nav2.

#### Implementation considerations
It has to be considered whether unconditionally enabling service introspection is a good idea. The tutorial in the ROS docs show how to introduce a parameter to enable or disable service introspection, it is unclear to me if it's expected to only turn this feature on during debugging, and if there are downsides to always enabling it.

SteveMacenski:
The client side is easy, I think most all uses of service clients use our `nav2_utils` service client wrapper, so one spot should hit most if not all of them (but worth doing a quick grep to make sure of that). 

The server side will be slightly more involved since I don't think we wrap anything there. I'm not sure if the right answer is to:
- create a `nav2_util` service server wrapper like the client
- create a `nav2_util` helper object that deals with these mechanics, getting the parameters, and setting it up that given a `node` and the server
- Better yet, contribute back to `rclcpp` by modifying `create_service` to handle this logic so that anyone creating a server can have this handled implicitly when a standardized parameter is set. I think you could do that with `create_client` too :wink:  
- Something else?

How do you think we should approach it? 

I'd say leave it default off, but let users flip it on with a parameter.

SteveMacenski:
@ottojo what do you think?

mini-1235:
I can work on this if no one is on it :)

SteveMacenski:
@mini-1235 please!

mini-1235:
> I'd say leave it default off, but let users flip it on with a parameter.
Where do you think we should put the parameter to flip this on, I haven't found a suitable place in a `nav2_params.yaml`, also, do you think it is necessary to support dynamic configuration of this parameter

PR #4955, opened 2025-03-02
Add service introspection

mini-1235:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4837 |

---

- Added service introspection as described in the issue (I created a service factory in `nav2_util` as not all service servers are derived from the LifecycleNode class)

Documents related to `nav2_params` would require a change, will raise a PR in the docs repo when this PR is ready 

I add a simple unit test, and the changes were mostly tested in my sim env (using cli to make sure service introspection works)

---

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -47,6 +47,7 @@
 #include "pluginlib/class_loader.hpp"
 #include "rclcpp/node_options.hpp"
 #include "sensor_msgs/msg/laser_scan.hpp"
+#include "nav2_util/service_server.hpp"
 #include "std_srvs/srv/empty.hpp"
 #include "tf2_ros/message_filter.h"
 #include "tf2_ros/transform_broadcaster.h"
@@ -206,7 +207,8 @@ protected:
    * @brief Initialize state services
    */
   void initServices();
-  rclcpp::Service<std_srvs::srv::Empty>::SharedPtr global_loc_srv_;
+  std::shared_ptr<nav2_util::ServiceServer<std_srvs::srv::Empty,
+    std::shared_ptr<nav2_util::LifecycleNode>>> global_loc_srv_;
   /*
    * @brief Service callback for a global relocalization request
    */
@@ -216,7 +218,8 @@ protected:
     std::shared_ptr<std_srvs::srv::Empty::Response> response);
 
   // service server for providing an initial pose guess
-  rclcpp::Service<nav2_msgs::srv::SetInitialPose>::SharedPtr initial_guess_srv_;
+  std::shared_ptr<nav2_util::ServiceServer<nav2_msgs::srv::SetInitialPose,
+    std::shared_ptr<nav2_util::LifecycleNode>>> initial_guess_srv_;
   /*
    * @brief Service callback for an initial pose guess request
    */
@@ -226,7 +229,8 @@ protected:
     std::shared_ptr<nav2_msgs::srv::SetInitialPose::Response> response);
 
   // Let amcl update samples without requiring motion
-  rclcpp::Service<std_srvs::srv::Empty>::SharedPtr nomotion_update_srv_;
+  std::shared_ptr<nav2_util::ServiceServer<std_srvs::srv::Empty,
+    std::shared_ptr<nav2_util::LifecycleNode>>> nomotion_update_srv_;
   /*
    * @brief Request an AMCL update even though the robot hasn't moved
    */
@@ -387,6 +391,7 @@ protected:
   double sigma_hit_;
   bool tf_broadcast_;
   tf2::Duration transform_tolerance_;
+  std::string service_introspection_mode_;
   double a_thresh_;
   double d_thresh_;
   double z_hit_;
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -391,7 +391,6 @@ protected:
   double sigma_hit_;
   bool tf_broadcast_;
   tf2::Duration transform_tolerance_;
-  std::string service_introspection_mode_;
   double a_thresh_;
   double d_thresh_;
   double z_hit_;
--- a/nav2_amcl/include/nav2_amcl/amcl_node.hpp
+++ b/nav2_amcl/include/nav2_amcl/amcl_node.hpp
@@ -207,8 +207,8 @@ protected:
    * @brief Initialize state services
    */
   void initServices();
-  std::shared_ptr<nav2_util::ServiceServer<std_srvs::srv::Empty,
-    std::shared_ptr<nav2_util::LifecycleNode>>> global_loc_srv_;
+  nav2_util::ServiceServer<std_srvs::srv::Empty,
+    std::shared_ptr<nav2_util::LifecycleNode>>::SharedPtr global_loc_srv_;
   /*
    * @brief Service callback for a global relocalization request
    */
@@ -218,8 +218,8 @@ protected:
     std::shared_ptr<std_srvs::srv::Empty::Response> response);
 
   // service server for providing an initial pose guess
-  std::shared_ptr<nav2_util::ServiceServer<nav2_msgs::srv::SetInitialPose,
-    std::shared_ptr<nav2_util::LifecycleNode>>> initial_guess_srv_;
+  nav2_util::ServiceServer<nav2_msgs::srv::SetInitialPose,
+    std::shared_ptr<nav2_util::LifecycleNode>>::SharedPtr initial_guess_srv_;
   /*
    * @brief Service callback for an initial pose guess request
    */
@@ -229,8 +229,8 @@ protected:
     std::shared_ptr<nav2_msgs::srv::SetInitialPose::Response> response);
 
   // Let amcl update samples without requiring motion
-  std::shared_ptr<nav2_util::ServiceServer<std_srvs::srv::Empty,
-    std::shared_ptr<nav2_util::LifecycleNode>>> nomotion_update_srv_;
+  nav2_util::ServiceServer<std_srvs::srv::Empty,
+    std::shared_ptr<nav2_util::LifecycleNode>>::SharedPtr nomotion_update_srv_;
   /*
    * @brief Request an AMCL update even though the robot hasn't moved
    */
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -23,6 +23,7 @@
 #include "nav2_util/node_utils.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "nav2_behavior_tree/bt_utils.hpp"
+#include "nav2_util/service_client.hpp"
 
 namespace nav2_behavior_tree
 {
@@ -173,7 +174,7 @@ public:
         return BT::NodeStatus::FAILURE;
       }
 
-      future_result_ = service_client_->async_send_request(request_).share();
+      future_result_ = service_client_->invoke_shared(request_);
       sent_time_ = node_->now();
       request_sent_ = true;
     }
@@ -265,8 +266,9 @@ protected:
   }
 
   std::string service_name_, service_node_name_;
-  typename std::shared_ptr<rclcpp::Client<ServiceT>> service_client_;
+  std::shared_ptr<nav2_util::ServiceClient<ServiceT>> service_client_;
   std::shared_ptr<typename ServiceT::Request> request_;
+  std::string service_introspection_mode_;
 
   // The node that will be used for any ROS operations
   rclcpp::Node::SharedPtr node_;
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -111,7 +111,7 @@ public:
       service_name_ = service_new;
       node_ = config().blackboard->template get<rclcpp::Node::SharedPtr>("node");
       service_client_ = std::make_shared<nav2_util::ServiceClient<ServiceT>>(
-      service_name_, node_, true);
+      service_name_, node_, true /*creates and spins an internal executor*/);
     }
   }
 
@@ -213,15 +213,15 @@ public:
     if (remaining > std::chrono::milliseconds(0)) {
       auto timeout = remaining > max_timeout_ ? max_timeout_ : remaining;
 
-      std::future_status future;
-      future = future_result_.wait_for(timeout);
-      if(future == std::future_status::ready) {
+      rclcpp::FutureReturnCode rc;
+      rc = service_client_->spin_until_complete(future_result_, timeout);
+      if (rc == rclcpp::FutureReturnCode::SUCCESS) {
         request_sent_ = false;
         BT::NodeStatus status = on_completion(future_result_.get());
         return status;
       }
 
-      if (future == std::future_status::timeout) {
+      if (rc == rclcpp::FutureReturnCode::TIMEOUT) {
         on_wait_for_result();
         elapsed = (node_->now() - sent_time_).template to_chrono<std::chrono::milliseconds>();
         if (elapsed < server_timeout_) {
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -174,7 +174,7 @@ public:
         return BT::NodeStatus::FAILURE;
       }
 
-      future_result_ = service_client_->invoke_shared(request_);
+      future_result_ = service_client_->async_call(request_);
       sent_time_ = node_->now();
       request_sent_ = true;
     }
@@ -268,7 +268,6 @@ protected:
   std::string service_name_, service_node_name_;
   std::shared_ptr<nav2_util::ServiceClient<ServiceT>> service_client_;
   std::shared_ptr<typename ServiceT::Request> request_;
-  std::string service_introspection_mode_;
 
   // The node that will be used for any ROS operations
   rclcpp::Node::SharedPtr node_;
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -110,14 +110,8 @@ public:
     if (service_new != service_name_ || !service_client_) {
       service_name_ = service_new;
       node_ = config().blackboard->template get<rclcpp::Node::SharedPtr>("node");
-      callback_group_ = node_->create_callback_group(
-        rclcpp::CallbackGroupType::MutuallyExclusive,
-        false);
-      callback_group_executor_.add_callback_group(callback_group_,
-          node_->get_node_base_interface());
-
       service_client_ = std::make_shared<nav2_util::ServiceClient<ServiceT>>(
-      service_name_, node_, callback_group_);
+      service_name_, node_, true);
     }
   }
 
@@ -219,15 +213,15 @@ public:
     if (remaining > std::chrono::milliseconds(0)) {
       auto timeout = remaining > max_timeout_ ? max_timeout_ : remaining;
 
-      rclcpp::FutureReturnCode rc;
-      rc = callback_group_executor_.spin_until_future_complete(future_result_, timeout);
-      if (rc == rclcpp::FutureReturnCode::SUCCESS) {
+      std::future_status future;
+      future = future_result_.wait_for(timeout);
+      if(future == std::future_status::ready) {
         request_sent_ = false;
         BT::NodeStatus status = on_completion(future_result_.get());
         return status;
       }
 
-      if (rc == rclcpp::FutureReturnCode::TIMEOUT) {
+      if (future == std::future_status::timeout) {
         on_wait_for_result();
         elapsed = (node_->now() - sent_time_).template to_chrono<std::chrono::milliseconds>();
         if (elapsed < server_timeout_) {
@@ -269,8 +263,6 @@ protected:
 
   // The node that will be used for any ROS operations
   rclcpp::Node::SharedPtr node_;
-  rclcpp::CallbackGroup::SharedPtr callback_group_;
-  rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
 
   // The timeout value while to use in the tick loop while waiting for
   // a result from the server
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -264,7 +264,7 @@ protected:
   }
 
   std::string service_name_, service_node_name_;
-  std::shared_ptr<nav2_util::ServiceClient<ServiceT>> service_client_;
+  typename nav2_util::ServiceClient<ServiceT>::SharedPtr service_client_;
   std::shared_ptr<typename ServiceT::Request> request_;
 
   // The node that will be used for any ROS operations
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
@@ -116,10 +116,8 @@ public:
       callback_group_executor_.add_callback_group(callback_group_,
           node_->get_node_base_interface());
 
-      service_client_ = node_->create_client<ServiceT>(
-        service_name_,
-        rclcpp::SystemDefaultsQoS(),
-        callback_group_);
+      service_client_ = std::make_shared<nav2_util::ServiceClient<ServiceT>>(
+      service_name_, node_, callback_group_);
     }
   }
 
--- a/nav2_costmap_2d/src/costmap_2d_publisher.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_publisher.cpp
@@ -55,7 +55,8 @@ Costmap2DPublisher::Costmap2DPublisher(
   std::string global_frame,
   std::string topic_name,
   bool always_send_full_costmap,
-  double map_vis_z)
+  double map_vis_z,
+  std::string service_introspection_mode)
 : costmap_(costmap),
   global_frame_(global_frame),
   topic_name_(topic_name),
@@ -82,11 +83,14 @@ Costmap2DPublisher::Costmap2DPublisher(
     topic_name + "_raw_updates", custom_qos);
 
   // Create a service that will use the callback function to handle requests.
-  costmap_service_ = node->create_service<nav2_msgs::srv::GetCostmap>(
-    "get_" + topic_name, std::bind(
-      &Costmap2DPublisher::costmap_service_callback,
-      this, std::placeholders::_1, std::placeholders::_2,
-      std::placeholders::_3));
+  costmap_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::GetCostmap,
+      std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
+    std::string("get_") + topic_name,
+    service_introspection_mode,
+    node->shared_from_this(),
+    std::bind(
+      &Costmap2DPublisher::costmap_service_callback, this,
+      std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
 
   if (cost_translation_table_ == NULL) {
     cost_translation_table_ = new char[256];
--- a/nav2_costmap_2d/src/costmap_2d_publisher.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_publisher.cpp
@@ -85,7 +85,7 @@ Costmap2DPublisher::Costmap2DPublisher(
   costmap_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::GetCostmap,
       std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
     std::string("get_") + topic_name,
-    node->shared_from_this(),
+    node,
     std::bind(
       &Costmap2DPublisher::costmap_service_callback, this,
       std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
--- a/nav2_costmap_2d/src/costmap_2d_publisher.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_publisher.cpp
@@ -55,8 +55,7 @@ Costmap2DPublisher::Costmap2DPublisher(
   std::string global_frame,
   std::string topic_name,
   bool always_send_full_costmap,
-  double map_vis_z,
-  std::string service_introspection_mode)
+  double map_vis_z)
 : costmap_(costmap),
   global_frame_(global_frame),
   topic_name_(topic_name),
@@ -86,7 +85,6 @@ Costmap2DPublisher::Costmap2DPublisher(
   costmap_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::GetCostmap,
       std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
     std::string("get_") + topic_name,
-    service_introspection_mode,
     node->shared_from_this(),
     std::bind(
       &Costmap2DPublisher::costmap_service_callback, this,
--- a/nav2_docking/opennav_docking/src/dock_database.cpp
+++ b/nav2_docking/opennav_docking/src/dock_database.cpp
@@ -55,16 +55,9 @@ bool DockDatabase::initialize(
     "Docking Server has %u dock types and %u dock instances available.",
     this->plugin_size(), this->instance_size());
 
-  if(!node->has_parameter("service_introspection_mode")) {
-    node->declare_parameter("service_introspection_mode", "disabled");
-  }
-  std::string service_introspection_mode_ =
-    node->get_parameter("service_introspection_mode").as_string();
-
   reload_db_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::ReloadDockDatabase,
       std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
     "~/reload_database",
-    service_introspection_mode_,
     node->shared_from_this(),
     std::bind(
       &DockDatabase::reloadDbCb, this,
--- a/nav2_docking/opennav_docking/src/dock_database.cpp
+++ b/nav2_docking/opennav_docking/src/dock_database.cpp
@@ -58,7 +58,7 @@ bool DockDatabase::initialize(
   reload_db_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::ReloadDockDatabase,
       std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
     "~/reload_database",
-    node->shared_from_this(),
+    node,
     std::bind(
       &DockDatabase::reloadDbCb, this,
       std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
--- a/nav2_docking/opennav_docking/src/dock_database.cpp
+++ b/nav2_docking/opennav_docking/src/dock_database.cpp
@@ -26,6 +26,7 @@ DockDatabase::~DockDatabase()
 {
   dock_instances_.clear();
   dock_plugins_.clear();
+  reload_db_service_.reset();
 }
 
 bool DockDatabase::initialize(
@@ -54,11 +55,19 @@ bool DockDatabase::initialize(
     "Docking Server has %u dock types and %u dock instances available.",
     this->plugin_size(), this->instance_size());
 
-  reload_db_service_ = node->create_service<nav2_msgs::srv::ReloadDockDatabase>(
+  // declare the service to reload the database
+  node->declare_parameter("service_introspection_mode", "disabled");
+  std::string service_introspection_mode_ =
+    node->get_parameter("service_introspection_mode").as_string();
+
+  reload_db_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::ReloadDockDatabase,
+      std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
     "~/reload_database",
+    service_introspection_mode_,
+    node->shared_from_this(),
     std::bind(
       &DockDatabase::reloadDbCb, this,
-      std::placeholders::_1, std::placeholders::_2));
+      std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
 
   return true;
 }
@@ -80,6 +89,7 @@ void DockDatabase::deactivate()
 }
 
 void DockDatabase::reloadDbCb(
+  const std::shared_ptr<rmw_request_id_t>/*request_header*/,
   const std::shared_ptr<nav2_msgs::srv::ReloadDockDatabase::Request> request,
   std::shared_ptr<nav2_msgs::srv::ReloadDockDatabase::Response> response)
 {
--- a/nav2_docking/opennav_docking/src/dock_database.cpp
+++ b/nav2_docking/opennav_docking/src/dock_database.cpp
@@ -55,8 +55,9 @@ bool DockDatabase::initialize(
     "Docking Server has %u dock types and %u dock instances available.",
     this->plugin_size(), this->instance_size());
 
-  // declare the service to reload the database
-  node->declare_parameter("service_introspection_mode", "disabled");
+  if(!node->has_parameter("service_introspection_mode")) {
+    node->declare_parameter("service_introspection_mode", "disabled");
+  }
   std::string service_introspection_mode_ =
     node->get_parameter("service_introspection_mode").as_string();
 
--- a/nav2_planner/src/planner_server.cpp
+++ b/nav2_planner/src/planner_server.cpp
@@ -61,6 +61,7 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
       declare_parameter(default_ids_[i] + ".plugin", default_types_[i]);
     }
   }
+
   // Setup the global costmap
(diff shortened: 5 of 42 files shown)


nav2_rviz_plugins/src/costmap_cost_tool.cpp
```diff
@@ -106,34 +116,11 @@ void CostmapCostTool::callCostService(float x, float y)
   request->use_footprint = false;
 
   // Call local costmap service
-  if (local_cost_client_->wait_for_service(std::chrono::seconds(1))) {
```
SteveMacenski, line 109: I think these should be reverted, the invoke shared would block the thread which would make all of rviz seize up until the service client got a response.

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -48,6 +49,18 @@ class ServiceClient
       service_name,
       rclcpp::SystemDefaultsQoS(),
       callback_group_);
+    rcl_service_introspection_state_t introspection_state = RCL_SERVICE_INTROSPECTION_OFF;
+
```
SteveMacenski, line 53: hHis is where you can declare and get the service introspection value internally to the class!

nav2_util/include/nav2_util/service_server.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include "rclcpp_lifecycle/lifecycle_node.hpp"
+
+namespace nav2_util
+{
+
+/**
+ * @class nav2_util::ServiceServer
+ * @brief A simple wrapper on ROS2 services for invoke() and block-style calling
+ */
+template<class ServiceT, typename NodeT = rclcpp::Node::SharedPtr>
```
SteveMacenski, line 30: If you look over the annotations, there's some pretty big testing gaps in this and the service client file from these changes. I think a few more unit tests in the various introspection modes are necessary

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -126,6 +139,41 @@ class ServiceClient
     return response.get();
   }
 
+  /**
+  * @brief Invoke the service and block until completed
+  * @param request The request object to call the service using
+  * @return std::shared_future<typename ResponseType::SharedPtr> The shared future of the service response
+  */
+  std::shared_future<typename ResponseType::SharedPtr> invoke_shared(
```
SteveMacenski, line 147: What's the value of this if we can return the response instead? Why return the shared future?

nav2_util/include/nav2_util/service_server.hpp
```diff
@@ -0,0 +1,81 @@
...
+{
+public:
+  using RequestType = typename ServiceT::Request;
+  using ResponseType = typename ServiceT::Response;
+  using CallbackType = std::function<void(const std::shared_ptr<rmw_request_id_t>,
+      const std::shared_ptr<RequestType>, std::shared_ptr<ResponseType>)>;
+
+  explicit ServiceServer(
+    const std::string & service_name,
+    std::string service_introspection_mode,
```
SteveMacenski, line 41: This could/should be removed and done in the constructor as well using the `node` input

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
```diff
@@ -148,7 +151,7 @@ class BtServiceNode : public BT::ActionNodeBase
         return BT::NodeStatus::FAILURE;
       }
 
-      future_result_ = service_client_->async_send_request(request_).share();
+      future_result_ = service_client_->invoke_shared(request_);
```
SteveMacenski, line 154: Is this the only necessary use of invoke_shared? If so, maybe we should have this remain a raw `create_client` and pass in the service introspection params into it

SteveMacenski (commented):
Great first stab!

nav2_util/include/nav2_util/service_server.hpp
mini-1235: Do you think I should wrap those clients and servers in the tests as well? Is it what codecov is looking for?

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
mini-1235: Also in bt_service_node, where the original code is 

https://github.com/ros-navigation/navigation2/blob/c7a5cc56f6b460a818cad27c5c47848f0ab9f0d0/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L118-L121

https://github.com/ros-navigation/navigation2/blob/c7a5cc56f6b460a818cad27c5c47848f0ab9f0d0/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L176-L178

Do you think it is better for a raw `create_client` for both of them?

nav2_util/include/nav2_util/service_server.hpp
SteveMacenski: I think you should create basically copies (or loops?) of the tests where you set the introspection mode to metadata, contents to exercise L60-L64 and make sure everything works properly (no crashes, seg faults on mode setting, etc)

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -96,6 +98,7 @@ WaypointFollower::on_configure(const rclcpp_lifecycle::State & state)
     nav2_util::ServiceClient<robot_localization::srv::FromLL,
```
SteveMacenski, line 98: Why is this one not in using the new method?

nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp
SteveMacenski: In `invoke_shared` you do a `spin_until_future_complete` which wouldn't be appropriate for the `tick` in `BtServiceNode` since that tick shouldn't block. The `check_future()` method does the spinning based on the BT tick rate so that it doesn't block for too long. That's I suppose why I was a little confused why an `invoke` method would return a future since it blocks until complete. The contexts when a future is useful necessarily means that we don't block (true of the costmap cost tool as well, the other place this method is used).

So, I think this should be changed to instead of `invoke_shared` we should come up with a new name that is clear that this is a non-blocking call and have this method return the future without spinning (so completely non-blocking). That would make it fit the context then used in the costmap rviz tool and in the `BtServiceNode` :smile: 

Does that make sense?

mini-1235: Ahh yes, I didn't notice this when I do a quick copy + paste from the `invoke` method, I will fix this

nav2_waypoint_follower/src/waypoint_follower.cpp
mini-1235: I think it is already using the new method(?) Am I missing anything?

nav2_amcl/include/nav2_amcl/amcl_node.hpp
```diff
@@ -206,7 +207,8 @@ class AmclNode : public nav2_util::LifecycleNode
    * @brief Initialize state services
    */
   void initServices();
-  rclcpp::Service<std_srvs::srv::Empty>::SharedPtr global_loc_srv_;
+  std::shared_ptr<nav2_util::ServiceServer<std_srvs::srv::Empty,
```
SteveMacenski, line 210: We could make this nicer if the `ServiceServer` had a typedef for `SharedPtr` so this could be changed to `nav2_util::ServiceServer<std_srvs::srv::Empty, std::shared_ptr<nav2_util::LifecycleNode>>>::SharedPtr`

That could then be used everywhere the type is declared

nav2_smac_planner/test/test_smac_2d.cpp
```diff
@@ -83,6 +83,7 @@ TEST(SmacTest, test_smac_2d) {
   costmap_ros->on_cleanup(rclcpp_lifecycle::State());
   node2D.reset();
   costmap_ros.reset();
+  std::cout << "SMAC 2D test complete" << std::endl;
```
SteveMacenski, line 86: Remove? :wink:

nav2_rviz_plugins/src/costmap_cost_tool.cpp
SteveMacenski: I think this is still true, while we don't block up the thread anymore since there's no `spin`, we immediately `.get()` the future, which is probably iinvalid since its not yet completed. I think the callback method where we call and triggers a callback when completed to log is still proper.

nav2_planner/src/planner_server.cpp
```diff
@@ -54,14 +54,15 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
   declare_parameter("expected_planner_frequency", 1.0);
   declare_parameter("action_server_result_timeout", 10.0);
   declare_parameter("costmap_update_timeout", 1.0);
+  declare_parameter("service_introspection_mode", "disabled");
```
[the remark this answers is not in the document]
mini-1235: I recently tried to move all the declare parameter + get parameter to the ServiceClient/ServiceServer, however, when I run the following test:

https://github.com/ros-navigation/navigation2/blob/9f09b064ef1145bbd82aa1e2eb211314ae20347b/nav2_smac_planner/test/test_smac_2d.cpp#L138-L143

it reports that `C++ exception with description "cannot set or declare a parameter, or change the callback from within set callback" thrown in the test body.`

After backtracing this, I found out that the error occurs in the following line

https://github.com/ros-navigation/navigation2/blob/9f09b064ef1145bbd82aa1e2eb211314ae20347b/nav2_smac_planner/src/costmap_downsampler.cpp#L36-L55

where `_downsampled_costmap_pub = std::make_unique<nav2_costmap_2d::Costmap2DPublisher>` create a service server that has not been created before, and calling the method `declare_parameter` in a dynamic callback will throw the exception above(?)

mini-1235:
I recently converted this to draft, I will ping you when this is ready:) @SteveMacenski

nav2_planner/src/planner_server.cpp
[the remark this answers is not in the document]
mini-1235: Can you help with the problem here @SteveMacenski :)

mini-1235:
Moving https://github.com/ros-navigation/navigation2/pull/4955#discussion_r2008292297

I recently tried to move all the declare parameter + get parameter to the ServiceClient/ServiceServer, however, when I run the following test:

https://github.com/ros-navigation/navigation2/blob/9f09b064ef1145bbd82aa1e2eb211314ae20347b/nav2_smac_planner/test/test_smac_2d.cpp#L138-L143

it reports that `C++ exception with description "cannot set or declare a parameter, or change the callback from within set callback" thrown in the test body.`

After backtracing this, I found out that the error occurs in the following line

https://github.com/ros-navigation/navigation2/blob/9f09b064ef1145bbd82aa1e2eb211314ae20347b/nav2_smac_planner/src/costmap_downsampler.cpp#L36-L55

where `_downsampled_costmap_pub = std::make_unique<nav2_costmap_2d::Costmap2DPublisher>` create a service server that has not been created before, and calling the method `declare_parameter` in a dynamic callback will throw the exception above(?)

Can you help me to take a look on this before I proceed @SteveMacenski

SteveMacenski:
Huh, neat. I looked into it in rclcpp and there's really no way around that. We could update the test to create the costmap downsampler object and `on_configure` it once so that the parameter is already declared once. Then when its set later to on, the parameter is already declared and will not throw. That is inline with how it would work in a deployed user application as well. I think that could be done by setting the downsample costmap parameter to true before these lines https://github.com/ros-navigation/navigation2/blob/9f09b064ef1145bbd82aa1e2eb211314ae20347b/nav2_smac_planner/test/test_smac_2d.cpp#L96-L98 and then updating the test to set https://github.com/ros-navigation/navigation2/blob/9f09b064ef1145bbd82aa1e2eb211314ae20347b/nav2_smac_planner/test/test_smac_2d.cpp#L138-L144 to false instead.

mini-1235:
> That is in line with how it would work in a deployed user application as well. I think that could be done by declaring the downsample costmap parameter to true before these lines

In the code, the CostmapDownSampler object is either created in the `configure` function or the dynamic callback, let's say the users did not specify the downsample costmap param(default false), and when they dynamically configure this param to true, it will throw the exception, isn't it? Shouldn't we do something else instead of just inverting the logic in unit test (?) as it doesn't handle the logic from `false` -> `true`

SteveMacenski:
Good point. How about we add in a `declare_parameter_if_not_declared` for that parameter in the Smac Planner plugin's `configure` methods with a brief comment that we need to declare it ahead of time for special reasons in the Smac Planners. That should resolve the issue without the need for adjusting the tests

mini-1235:
> Good point. How about we add in a `declare_parameter_if_not_declared` for that parameter in the Smac Planner plugin's `configure` methods with a brief comment that we need to declare it ahead of time for special reasons in the Smac Planners. That should resolve the issue without the need for adjusting the tests

Sounds good - I will add that. Thanks for the suggestions

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -126,6 +148,32 @@ class ServiceClient
...
+  }
+
+
+  /**
+  * @brief Asynchronously call the service with a callback
+  * @param request The request object to call the service using
+  * @param callback The callback to call when the service response is received
+  */
+  template<typename CallbackT>
+  void async_call(
```
SteveMacenski, line 170: This method needs test coverage

nav2_costmap_2d/src/costmap_2d_publisher.cpp
```diff
@@ -82,11 +82,13 @@ Costmap2DPublisher::Costmap2DPublisher(
...
   // Create a service that will use the callback function to handle requests.
-  costmap_service_ = node->create_service<nav2_msgs::srv::GetCostmap>(
-    "get_" + topic_name, std::bind(
-      &Costmap2DPublisher::costmap_service_callback,
-      this, std::placeholders::_1, std::placeholders::_2,
-      std::placeholders::_3));
+  costmap_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::GetCostmap,
+      std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
+    std::string("get_") + topic_name,
+    node->shared_from_this(),
```
SteveMacenski, line 88:
```suggestion
    node,
```

I think this should work?

nav2_docking/opennav_docking/src/dock_database.cpp
```diff
@@ -54,11 +55,13 @@ bool DockDatabase::initialize(
     "Docking Server has %u dock types and %u dock instances available.",
     this->plugin_size(), this->instance_size());
 
-  reload_db_service_ = node->create_service<nav2_msgs::srv::ReloadDockDatabase>(
+  reload_db_service_ = std::make_shared<nav2_util::ServiceServer<nav2_msgs::srv::ReloadDockDatabase,
+      std::shared_ptr<rclcpp_lifecycle::LifecycleNode>>>(
     "~/reload_database",
+    node->shared_from_this(),
```
SteveMacenski, line 61: Ditto

nav2_planner/src/planner_server.cpp
```diff
@@ -61,7 +61,6 @@ PlannerServer::PlannerServer(const rclcpp::NodeOptions & options)
       declare_parameter(default_ids_[i] + ".plugin", default_types_[i]);
     }
   }
-
```
SteveMacenski, line 64: Readd this separator line

nav2_rviz_plugins/include/nav2_rviz_plugins/costmap_cost_tool.hpp
```diff
@@ -48,8 +49,8 @@ class CostmapCostTool : public rviz_common::Tool
 private Q_SLOTS:
 
 private:
-  rclcpp::Client<nav2_msgs::srv::GetCosts>::SharedPtr local_cost_client_;
-  rclcpp::Client<nav2_msgs::srv::GetCosts>::SharedPtr global_cost_client_;
+  std::shared_ptr<nav2_util::ServiceClient<nav2_msgs::srv::GetCosts>> local_cost_client_;
```
SteveMacenski, line 52: Use `SharedPtr` you added? Make sure to look around the diffs here in github and change all to use those! :-)

nav2_util/include/nav2_util/service_server.hpp
```diff
@@ -0,0 +1,86 @@
...
+        this->callback_(request_header, request, response);
+      },
+      qos,
+      callback_group);
+
+    rcl_service_introspection_state_t introspection_state = RCL_SERVICE_INTROSPECTION_OFF;
+    if(!node->has_parameter("service_introspection_mode")) {
+      node->declare_parameter("service_introspection_mode", "disabled");
+    }
+    std::string service_introspection_mode_ =
```
SteveMacenski, line 61: This should not have an underscore since its not a class member

nav2_util/include/nav2_util/service_server.hpp
```diff
@@ -0,0 +1,86 @@
...
+      node->get_parameter("service_introspection_mode").as_string();
+    if (service_introspection_mode_ == "disabled") {
```
SteveMacenski, line 63: We default to OFF in L57 so you don't need to consider this case specifically :-)

nav2_util/include/nav2_util/service_server.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    this->server_->configure_introspection(
+    node->get_clock(), rclcpp::SystemDefaultsQoS(), introspection_state);
+  }
+
+protected:
+  std::string service_name_;
+  CallbackType callback_;
+  typename rclcpp::Service<ServiceT>::SharedPtr server_;
+  rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr
```
SteveMacenski, line 79: is `OnSetParametersCallbackHandle` used?

nav2_util/include/nav2_util/service_client.hpp
SteveMacenski: Also: why not remove this method and use the one that returns the future. Applications that don't need the future can just drop the return and not store it. I don't think this necessarily needs to be duplicated - so you could remove this entirely I think instead

nav2_smac_planner/test/test_smac_2d.cpp
SteveMacenski: ^ still relevant

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -37,17 +37,39 @@ class ServiceClient
...
       service_name,
       rclcpp::SystemDefaultsQoS(),
       callback_group_);
+    rcl_service_introspection_state_t introspection_state = RCL_SERVICE_INTROSPECTION_OFF;
+    if(!node_->has_parameter("service_introspection_mode")) {
+      node_->declare_parameter("service_introspection_mode", "disabled");
+    }
+    std::string service_introspection_mode_ =
+      node_->get_parameter("service_introspection_mode").as_string();
+    if (service_introspection_mode_ == "disabled") {
```
SteveMacenski, line 63: Ditto: can remove the disabled case since we default to off in L57

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -37,17 +37,39 @@ class ServiceClient
...
+
     client_ = node_->template create_client<ServiceT>(
       service_name,
       rclcpp::SystemDefaultsQoS(),
       callback_group_);
+    rcl_service_introspection_state_t introspection_state = RCL_SERVICE_INTROSPECTION_OFF;
+    if(!node_->has_parameter("service_introspection_mode")) {
+      node_->declare_parameter("service_introspection_mode", "disabled");
+    }
+    std::string service_introspection_mode_ =
```
SteveMacenski, line 61: Remove underscore, not a class member

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -37,17 +37,39 @@ class ServiceClient
...
     const std::string & service_name,
-    const NodeT & provided_node)
+    const NodeT & provided_node, rclcpp::CallbackGroup::SharedPtr callback_group = nullptr)
   : service_name_(service_name), node_(provided_node)
   {
-    callback_group_ = node_->create_callback_group(
-      rclcpp::CallbackGroupType::MutuallyExclusive,
-      false);
-    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
+    if(!callback_group) {
```
SteveMacenski, line 43: Can you explain this change? I think anything like this should be in a different PR since that's changign the functionality - unless there are usecases where this needed within this PR

SteveMacenski (changes requested):

SteveMacenski:
Small largely cleanup, even though it looks like quite a few items. Most will take 3 seconds

nav2_util/include/nav2_util/service_client.hpp
mini-1235: what about this one
https://github.com/ros-navigation/navigation2/blob/848f2b060725d60b390eb60e9a6b25a8003961c0/nav2_rviz_plugins/src/costmap_cost_tool.cpp#L109-L120
I think as you suggested earlier we still need the callback here?

mini-1235: Unfortunately when I tried using this wrapper, I found out that the callback is never entered, I am still debugging on this:)

SteveMacenski: Oh ok, yeah please!

nav2_util/include/nav2_util/service_client.hpp
mini-1235: In fact, the original line that probably needs this is 
https://github.com/ros-navigation/navigation2/blob/848f2b060725d60b390eb60e9a6b25a8003961c0/nav2_behavior_tree/include/nav2_behavior_tree/bt_service_node.hpp#L118-L121
Without this change, it returns RUNNING instead of SUCCESS in the following line
https://github.com/ros-navigation/navigation2/blob/848f2b060725d60b390eb60e9a6b25a8003961c0/nav2_behavior_tree/test/plugins/action/test_clear_costmap_service.cpp#L99-L112

SteveMacenski: Is that the only place that this used?

mini-1235: Yes

SteveMacenski: Got it, ok! This is all good

mini-1235:
As of now, I am passing a callback group for **CostmapCostTool** and **BtServiceNode** in the constructor of ServiceClient, preventing the node to create a new callback group and adding it to a new executor. 

I think an alternative to this would be to always create a callback group and a new executor, then the `AsyncCall` functions can be removed and we can call the original blocking `invoke` function directly (although invoke will call `spin_until_future_complete`, but it is using a different executor with the node, therefore should not be blocking the main thread(?)) Am I understanding this correctly and is this what you prefer? @SteveMacenski

mini-1235:
PR for configuration guide: https://github.com/ros-navigation/docs.nav2.org/pull/661

SteveMacenski (changes requested):
Did you fix the issues with the services not working properly that you mentioned? 

---

> I think an alternative to this would be to always create a callback group and a new executor, then the AsyncCall functions can be removed and we can call the original blocking invoke function directly (although invoke will call spin_until_future_complete, but it is using a different executor from the node, therefore should not be blocking the main thread(?))

 It looks like  we have service clients in the BTServiceNode, isPathValidCondition, LifecycleManagerClient, CostmapTool. LifecycleManagerClient already used the service client in nav2_util, so we can ignore that, the behavior is the same. 

BTServiceNode (`async_call(request_)`), isPathValidCondition (`invoke(request, server_timeout_)`), CostmapTool (`async_call(request, callback)`) remain. 

The async's don't use the `callback_group_executor_` to spin, so its not _technically_ an issue, but its a pretty fragile behavior that an executor still exists that someone could use that would be called in another thread of the main application (if changed to use `invoke`), causing subtle bugs.  Both of these cases `BTServiceNode` and `CostmapTool` we pass in a callback group, so maybe we should also pass in the executor to set as  `callback_group_executor_` in the `else` condition where we set `callback_group_ = callback_group;`? That would also maintain the analog behavior of what happened prior to this PR as well if used in that way. 

`isPathValidCondition` calling the service in another thread when there is a `Node` object on the blackboard with its own executor reserved for BT nodes to do work on is also a bit odd. It should ideally use the `node`s provided executor, although other subscriber nodes like the `ControllerSelectorNode` have their own internal executors to spin only its information, so I suppose its fine. Though, if we can pass in the callback group + executor like in the previous paragraph, we could use `node_` to do so here as well! 

Improved thought: We could instead have a `bool` for whether to create a new callback group, executor or use the provided `node` input argument's default callback group and default executor, rather than passing into it both the callback group and executor. We do this in the simple action server already called `spin_thread`. In this case, we wouldn't spin the thread, but we would use a new executor / callback group per-service. `use_internal_thread`! That would be the cleanest API and enable the behavior we want, while reducing the opportunity for mis-use by providing only the callback group or only the executor, but not both paired up. And then the input is still only just the `node`. I think that's the best direction that would be used by all 3x of the cases above.

https://github.com/ros-navigation/navigation2/blob/a904802fe222f961d4a211fa0c813349672c18eb/nav2_util/include/nav2_util/simple_action_server.hpp#L115-L138

mini-1235:
> Did you fix the issues with the services not working properly that you mentioned?

I did, passing the callback group in costmap tool when creating client solved the issue

> Improved thought: We could instead have a bool for whether to create a new callback group, executor or use the provided node input argument's default callback group and default executor, rather than passing into it both the callback group and executor. We do this in the simple action server already called spin_thread. In this case, we wouldn't spin the thread, but we would use a new executor / callback group per-service. use_internal_thread! That would be the cleanest API and enable the behavior we want, while reducing the opportunity for mis-use by providing only the callback group or only the executor, but not both paired up. And then the input is still only just the node. I think that's the best direction that would be used by all 3x of the cases above.

Thanks for the suggestions, I will take a look at that

SteveMacenski:
LInking docs PR https://github.com/ros-navigation/docs.nav2.org/pull/661

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -96,7 +96,8 @@ WaypointFollower::on_configure(const rclcpp_lifecycle::State & state)
...
     std::shared_ptr<nav2_util::LifecycleNode>>>(
     "/fromLL",
-    node);
+    node,
+    true);
```
SteveMacenski, line 100:
```suggestion
    true /*creates and spins an internal executor*/);
```

Add this comment anywhere true / false are set (where false is 'Does not create'). I think this is a subtle but key detail that folks shouldn't overlook. This is not a software requirement but something that will help me in the future as I'm debugging user issues :-)

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -34,24 +40,48 @@ class ServiceClient
...
+      introspection_state = RCL_SERVICE_INTROSPECTION_CONTENTS;
+    }
+
+    this->client_->configure_introspection(
+        node_->get_clock(), rclcpp::SystemDefaultsQoS(), introspection_state);
+    if(spin_thread) {
+      callback_group_executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+      callback_group_executor_->add_callback_group(callback_group_,
+          node_->get_node_base_interface());
+      executor_thread_ = std::make_unique<nav2_util::NodeThread>(callback_group_executor_);
```
SteveMacenski, line 78: Oh this can't be here, otherwise `spin_until_future_complete` will throw exceptions each time its called since the executor is already spinning.

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -148,8 +212,9 @@ class ServiceClient
 protected:
   std::string service_name_;
   NodeT node_;
-  rclcpp::CallbackGroup::SharedPtr callback_group_;
-  rclcpp::executors::SingleThreadedExecutor callback_group_executor_;
+  rclcpp::CallbackGroup::SharedPtr callback_group_{nullptr};
+  rclcpp::executors::SingleThreadedExecutor::SharedPtr callback_group_executor_;
+  std::unique_ptr<nav2_util::NodeThread> executor_thread_;
```
SteveMacenski, line 217: Remove

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -77,15 +107,27 @@ class ServiceClient
...
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
-
-    if (callback_group_executor_.spin_until_future_complete(future_result, timeout) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
-      // Pending request must be manually cleaned up if execution is interrupted or timed out
-      client_->remove_pending_request(future_result);
-      throw std::runtime_error(service_name_ + " service client: async_send_request failed");
+    if (callback_group_executor_) {
```
SteveMacenski, line 110: This if/else should be removed and instead set the `callback_group_executor_` to the node's default executor in the constructor so that its always populated (whether its the internal executor or the `node`'s). You also need to use `spin_until_future_complete` for the same reason I mention in the BT Service Node, manually handling the futures is not functional unfortunately. I think this needs to be reverted, I think what happened before was perfect, we just needed to set the `callback_group_executor_` appropriately

SteveMacenski (changes requested):
This introduced a few bugs that need to be fixed and I'm very surprised that our CI passed considering that I don't think anything involving the internal executor strictly speaking worked (or worked for wrong reasons because there was a background thread always spinning which has run-time overhead)

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -34,24 +40,48 @@ class ServiceClient
...
   {
-    callback_group_ = node_->create_callback_group(
-      rclcpp::CallbackGroupType::MutuallyExclusive,
-      false);
-    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
+    if(spin_thread) {
+      callback_group_ = node_->create_callback_group(
+        rclcpp::CallbackGroupType::MutuallyExclusive,
+        false);
+    }
```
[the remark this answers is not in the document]
SteveMacenski: The other `else` with the executor to use the `node`'s

mini-1235:
Ok, I think I have misunderstood this part
> We could instead have a bool for whether to create a new callback group, executor or use the provided node input argument's default callback group and default executor, rather than passing into it both the callback group and executor. We do this in the simple action server already called spin_thread. In this case, we wouldn't spin the thread, but we would use a new executor / callback group per-service. use_internal_thread

I will fix this now!

nav2_util/include/nav2_util/service_client.hpp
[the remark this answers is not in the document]
mini-1235: I know that we have `node_->get_node_base_interface()->get_default_callback_group();` to get the default callback group, but do we really have a method to get the node's default executor in rclcpp?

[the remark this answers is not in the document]
SteveMacenski: If you track down how `spin_until_future_complete` on a `node` is implemented, you'll find it I'm sure!

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -34,24 +35,51 @@ class ServiceClient
...
+        rclcpp::CallbackGroupType::MutuallyExclusive,
+        false);
+      callback_group_executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+      callback_group_executor_->add_callback_group(callback_group_,
+          node_->get_node_base_interface());
+    } else {
+      callback_group_ = node_->get_node_base_interface()->get_default_callback_group();
+      rclcpp::ExecutorOptions options;
+      options.context = node_->get_node_base_interface()->get_context();
+      callback_group_executor_ =
```
mini-1235, line 56: Some thoughts after I submitted the latest commit:
I don't think we can find the default executor here, based on my understanding, I can see `rclcpp::spin_until_future_complete` is doing these things: 
1. Creates a single threaded executor
```
options.context = node_ptr->get_context();
rclcpp::executors::SingleThreadedExecutor executor(options);
```
2. Add the node to the executor
```
executor.add_node(node_ptr);
```
3. Finally spin it as how `executor->spin_until_future_complete` is done

Now, back to our cases:
1. costmap cost tool: The node has probably been added to an executor, we need to find which executor belongs to the node, do something like `getExecutor` here, and add that callback group to prevent the callback group is added to different executors
2. is path valid: In the original code I can see it is calling `rclcpp::spin_until_future_complete`, I believe this means that the node has not been added to any executor, so we can't really get the "default executor" here
These are the two cases where we will use this `else` code segment, and I am not really sure if  I have implemented it correctly in the latest commit (I don't think get the executor for costmap tool)

mini-1235: Or even better, we could remove that else code segment when not using internal executor
- when a nullptr pass to `create_client` method, rclcpp will automatically find the default callback group for us, and that's the only place where it is using callback group
- we can then create an if & else for `spin_until_future_complete`, if using internal executor, we use `executor->spin_until_future_complete` otherwise we will call `rclcpp::spin_until_future_complete` and let `rclcpp` take care the complexity for us

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -34,24 +35,51 @@ class ServiceClient
...
-    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
+    if(use_internal_executor) {
+      callback_group_ = node_->create_callback_group(
+        rclcpp::CallbackGroupType::MutuallyExclusive,
+        false);
+      callback_group_executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+      callback_group_executor_->add_callback_group(callback_group_,
+          node_->get_node_base_interface());
+    } else {
+      callback_group_ = node_->get_node_base_interface()->get_default_callback_group();
```
SteveMacenski, line 53: I'd add this comment:

```
// This is how rclcpp::spin_until_future_complete obtains the default context executor to use
// We do so here to mimic that behavior for service calls that do not require a new thread to spin on
```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -113,19 +146,47 @@ class ServiceClient
       node_->get_logger(), "%s service client: send async request",
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
-
-    if (callback_group_executor_.spin_until_future_complete(future_result) !=
+    if(!use_internal_executor_) {
```
SteveMacenski, line 149:
```suggestion
    if (!use_internal_executor_) {
```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -77,14 +105,19 @@ class ServiceClient
       node_->get_logger(), "%s service client: send async request",
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
-
-    if (callback_group_executor_.spin_until_future_complete(future_result, timeout) !=
+    if(!use_internal_executor_) {
```
SteveMacenski, line 108:
```suggestion
    if (!use_internal_executor_) {
```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -34,24 +35,51 @@ class ServiceClient
...
+      callback_group_executor_ =
+        std::make_shared<rclcpp::executors::SingleThreadedExecutor>(options);
+    }
+
     client_ = node_->template create_client<ServiceT>(
       service_name,
       rclcpp::SystemDefaultsQoS(),
       callback_group_);
+    rcl_service_introspection_state_t introspection_state = RCL_SERVICE_INTROSPECTION_OFF;
+    if(!node_->has_parameter("service_introspection_mode")) {
```
SteveMacenski, line 65:
```suggestion
    if (!node_->has_parameter("service_introspection_mode")) {
```

nav2_util/include/nav2_util/service_client.hpp
SteveMacenski: > when a nullptr pass to create_client method, rclcpp will automatically find the default callback group for us, and that's the only place where it is using callback group

Oh nice! Just make sure to add in a comment stating that explicitly for someone later reading this to understand 

> we can then create an if & else for spin_until_future_complete, if using internal executor, we use executor->spin_until_future_complete otherwise we will call rclcpp::spin_until_future_complete and let rclcpp take care the complexity for us

:+1: Agreed!

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -34,24 +35,45 @@ class ServiceClient
...
-    const NodeT & provided_node)
-  : service_name_(service_name), node_(provided_node)
+    const NodeT & provided_node, bool use_internal_executor = false)
+  : service_name_(service_name), node_(provided_node), use_internal_executor_(use_internal_executor)
   {
-    callback_group_ = node_->create_callback_group(
-      rclcpp::CallbackGroupType::MutuallyExclusive,
-      false);
-    callback_group_executor_.add_callback_group(callback_group_, node_->get_node_base_interface());
+    if(use_internal_executor) {
```
SteveMacenski, line 45:
```suggestion
    if (use_internal_executor) {
```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -77,13 +99,22 @@ class ServiceClient
...
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
-
-    if (callback_group_executor_.spin_until_future_complete(future_result, timeout) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
-      // Pending request must be manually cleaned up if execution is interrupted or timed out
-      client_->remove_pending_request(future_result);
-      throw std::runtime_error(service_name_ + " service client: async_send_request failed");
+    if(use_internal_executor_) {
```
SteveMacenski, line 102:
```suggestion
    if (use_internal_executor_) {
```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -113,19 +144,51 @@ class ServiceClient
       node_->get_logger(), "%s service client: send async request",
       service_name_.c_str());
     auto future_result = client_->async_send_request(request);
-
-    if (callback_group_executor_.spin_until_future_complete(future_result) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+    if(use_internal_executor_) {
```
SteveMacenski, line 147:
```suggestion
    if (use_internal_executor_) {
```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -113,19 +144,51 @@ class ServiceClient
...
+      }
+    } else {
+      if (rclcpp::spin_until_future_complete(node_, future_result) !=
+        rclcpp::FutureReturnCode::SUCCESS)
+      {
+        // Pending request must be manually cleaned up if execution is interrupted or timed out
+        client_->remove_pending_request(future_result);
+        return false;
+      }
     }
```
SteveMacenski, line 163:
```suggestion
    }

```

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -77,13 +99,22 @@ class ServiceClient
...
     auto future_result = client_->async_send_request(request);
-
-    if (callback_group_executor_.spin_until_future_complete(future_result, timeout) !=
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
-      // Pending request must be manually cleaned up if execution is interrupted or timed out
-      client_->remove_pending_request(future_result);
-      throw std::runtime_error(service_name_ + " service client: async_send_request failed");
+    if(use_internal_executor_) {
+      if (callback_group_executor_->spin_until_future_complete(future_result, timeout) !=
```
SteveMacenski, line 103: Can the new `spin_until_complete` API added here be used instead? That handles the `if/else` on `use_internal_executor_`. This (and below) couldbe reduced to:

```
if (spin_until_complete(future_result, timeout) != rclcpp::FutureReturnCode::SUCCESS)
{
        // Pending request must be manually cleaned up if execution is interrupted or timed out
        client_->remove_pending_request(future_result);
        throw std::runtime_error(service_name_ + " service client: async_send_request failed");
}
```

This would also nicely remove the redundancy in the removing pending request and throwing code :smile:

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -113,19 +144,51 @@ class ServiceClient
...
-      rclcpp::FutureReturnCode::SUCCESS)
-    {
+    if(use_internal_executor_) {
+      if (callback_group_executor_->spin_until_future_complete(future_result) !=
+        rclcpp::FutureReturnCode::SUCCESS)
+      {
       // Pending request must be manually cleaned up if execution is interrupted or timed out
-      client_->remove_pending_request(future_result);
-      return false;
+        client_->remove_pending_request(future_result);
```
SteveMacenski, line 152: Ditto

SteveMacenski (changes requested):
OK - pretty much done. Have you tested each of the services to make sure it works after all of the adjustments we made?

mini-1235:
> OK - pretty much done. Have you tested each of the services to make sure it works after all of the adjustments we made?

Tested again after the latest commit, I think all services are working properly

nav2_util/include/nav2_util/service_client.hpp
```diff
@@ -136,6 +176,26 @@ class ServiceClient
...
+   * @param future The shared future to wait for completion.
+   * @param timeout The maximum time to wait for the future to complete. Default is -1 (no timeout).
+   * @return rclcpp::FutureReturnCode indicating the result of the spin operation.
+   */
+  template<typename FutureT>
+  rclcpp::FutureReturnCode spin_until_complete(
+    const FutureT & future,
+    const std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1))
+  {
+    // return callback_group_executor_->spin_until_future_complete(future, timeout);
```
SteveMacenski, line 191:
```suggestion
```

SteveMacenski (approved):

PR #4959, opened 2025-03-05
Fix the message header timestamp of the velocity_smoother, when the input timed out

georgflick:
---

## Basic Info

| ------ | ----------- |

---

* Populate the header stamp of the TwistStamped message with the actual time, in the case that the velocity command times out and sends zero velocities to stop the robot.

* Not needed

* Checked in simulation using `ros2 topic echo /cmd_vel_smoothed` , that the `header.stamp` is always populated by a timestamp and not zero at the end of a path or after cancelling a path

---

#### For Maintainers: 


SteveMacenski (approved):
Nice, I think this does it perfectly, thanks for the PR!

PR #4965, opened 2025-03-06
Added docking to BT xml

ajtudela:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Not needed |

---

-  Fix formatting.
- Added DockRobot and UndockRobot to nav2_tree_nodes.xml

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/nav2_tree_nodes.xml
+++ b/nav2_behavior_tree/nav2_tree_nodes.xml
@@ -256,6 +256,12 @@
       <output_port name="success">If the action was successful</output_port>
     </Action>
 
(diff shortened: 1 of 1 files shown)


mini-1235:
I think we’re also missing GetPoseFromPath in the BT XML. Can you add it to this PR as well? ❤️
https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/plugins/action/get_pose_from_path_action.cpp

SteveMacenski (approved):

SteveMacenski:
Thanks you both for the review and the contribution!

PR #4971, opened 2025-03-07
Dynamic param patterns

Nils-ChristianIseke:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4907  |

---

---

#### For Maintainers: 

@SteveMacenski I hope I understood the ticket correctly. I will continue to work on the rest of the codebase if these changes are the expected ones :).

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -189,28 +189,33 @@ ParameterHandler::ParameterHandler(
     params_.use_cost_regulated_linear_velocity_scaling = false;
   }
 
-  dyn_params_handler_ = node->add_on_set_parameters_callback(
+  post_set_params_handler_ = node->add_post_set_parameters_callback(
     std::bind(
-      &ParameterHandler::dynamicParametersCallback,
+      &ParameterHandler::updateParametersCallback,
+      this, std::placeholders::_1));
+  on_set_params_handler_ = node->add_on_set_parameters_callback(
+    std::bind(
+      &ParameterHandler::validateParameterUpdatesCallback,
       this, std::placeholders::_1));
 }
 
 ParameterHandler::~ParameterHandler()
 {
   auto node = node_.lock();
-  if (dyn_params_handler_ && node) {
-    node->remove_on_set_parameters_callback(dyn_params_handler_.get());
+  if (post_set_params_handler_ && node) {
+    node->remove_post_set_parameters_callback(post_set_params_handler_.get());
   }
-  dyn_params_handler_.reset();
+  post_set_params_handler_.reset();
+  if (on_set_params_handler_ && node) {
+    node->remove_on_set_parameters_callback(on_set_params_handler_.get());
+  }
+  on_set_params_handler_.reset();
 }
-
-rcl_interfaces::msg::SetParametersResult
-ParameterHandler::dynamicParametersCallback(
+rcl_interfaces::msg::SetParametersResult ParameterHandler::validateParameterUpdatesCallback(
   std::vector<rclcpp::Parameter> parameters)
 {
   rcl_interfaces::msg::SetParametersResult result;
-  std::lock_guard<std::mutex> lock_reinit(mutex_);
-
+  result.successful = true;
   for (auto parameter : parameters) {
     const auto & type = parameter.get_type();
     const auto & name = parameter.get_name();
@@ -221,8 +226,34 @@ ParameterHandler::dynamicParametersCallback(
           RCLCPP_WARN(
             logger_, "The value inflation_cost_scaling_factor is incorrectly set, "
             "it should be >0. Ignoring parameter update.");
-          continue;
+          result.successful = false;
         }
+      }
+    } else if (type == ParameterType::PARAMETER_BOOL) {
+      if (name == plugin_name_ + ".allow_reversing") {
+        if (params_.use_rotate_to_heading && parameter.as_bool()) {
+          RCLCPP_WARN(
+            logger_, "Both use_rotate_to_heading and allow_reversing "
+            "parameter cannot be set to true. Rejecting parameter update.");
+          result.successful = false;
+        }
+      }
+    }
+  }
+  return result;
+}
+void
+ParameterHandler::updateParametersCallback(
+  std::vector<rclcpp::Parameter> parameters)
+{
+  std::lock_guard<std::mutex> lock_reinit(mutex_);
+
+  for (const auto & parameter : parameters) {
+    const auto & type = parameter.get_type();
+    const auto & name = parameter.get_name();
+
+    if (type == ParameterType::PARAMETER_DOUBLE) {
+      if (name == plugin_name_ + ".inflation_cost_scaling_factor") {
         params_.inflation_cost_scaling_factor = parameter.as_double();
       } else if (name == plugin_name_ + ".desired_linear_vel") {
         params_.desired_linear_vel = parameter.as_double();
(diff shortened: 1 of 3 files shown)


nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -189,28 +189,34 @@ ParameterHandler::ParameterHandler(
     params_.use_cost_regulated_linear_velocity_scaling = false;
   }
 
-  dyn_params_handler_ = node->add_on_set_parameters_callback(
+  post_set_params_handler_ = node->add_post_set_parameters_callback(
     std::bind(
-      &ParameterHandler::dynamicParametersCallback,
+      &ParameterHandler::reactToUpdatedParametersCallback,
```
SteveMacenski, line 194: I'd update the names to `validateParameterUpdatesCallback` and `updateParametersCallback`

SteveMacenski (changes requested):
Otherwise, looks good to me!

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
Nils-ChristianIseke: Will change it.

I used the naming from the examples: https://github.com/ros2/demos/blob/rolling/demo_nodes_cpp/src/parameters/set_parameters_callback.cpp.

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -189,28 +189,34 @@ ParameterHandler::ParameterHandler(
...
+  if (on_set_params_handler_ && node) {
+    node->remove_on_set_parameters_callback(on_set_params_handler_.get());
   }
-  dyn_params_handler_.reset();
+  on_set_params_handler_.reset();
 }
-
-rcl_interfaces::msg::SetParametersResult
-ParameterHandler::dynamicParametersCallback(
+// Follow up on this example: https://github.com/ros2/demos/blob/rolling/demo_nodes_cpp/src/parameters/set_parameters_callback.cpp
```
SteveMacenski, line 214: Remove comment

SteveMacenski (changes requested):
Otherwise, this is a good template to use for other future ports!

Nils-ChristianIseke:
@SteveMacenski  i am not sure if i like the approach of [589474c](https://github.com/ros-navigation/navigation2/pull/4971/commits/589474c0c4dfadfa062beb1819772193ff3f6d9b). What is your opinion on it?

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -162,56 +162,80 @@ ParameterHandler::ParameterHandler(
...
-  dyn_params_handler_ = node->add_on_set_parameters_callback(
+  on_set_params_handler_ = node->add_on_set_parameters_callback(
+    std::bind(
+      &ParameterHandler::validateParameterUpdatesCallback,
+      this, std::placeholders::_1));
+  post_set_params_handler_ = node->add_post_set_parameters_callback(
+    std::bind(
+      &ParameterHandler::updateParametersCallback,
+      this, std::placeholders::_1));
+  pre_set_params_handler_ = node->add_pre_set_parameters_callback(
```
SteveMacenski, line 182: Is the pre-set actually triggered on the initialization of the node to remove the checks above? If so, wouldn't we be able to remove all the parameter gets because the pre/post/on-set would be triggered on initialization?

SteveMacenski:
> @SteveMacenski i am not sure if i like the approach of [589474c](https://github.com/ros-navigation/navigation2/pull/4971/commits/589474c0c4dfadfa062beb1819772193ff3f6d9b). What is your opinion on it?

I'm in the same boat as you; I don't know that I like it either. The declares I like better than the gets, considering the param vectors are separated by types. I do like the simplicity of the below though:

```
      auto it = double_parameters.find(name);
      if (it != double_parameters.end()) {
        it->second(parameter.as_double());
      }
```

However, I think if we did something like this, I think we should have an internal struct within the param handler quick includes everything: the parameter name, its default, its value, the setter lambda, and update validation lambda. That way, we define that once and its used for declare/get/dynamic parameter update/dynamic parameter validation. The repeated vector lists I think make it less readable to a casual developer compared to what's already there with declare/get calls that are pretty standard across ROS 2.

FYI you might be interested in https://github.com/PickNikRobotics/generate_parameter_library. I'm not sure I love this method (what if we want more custom parts of our parameter handler, like wanting to mutex lock parameter access for multi-threaded situations where the dynamic parameter executor is different from the running executor) -- but maybe those are handled and/or you see value/inspiration here

Nils-ChristianIseke:
I'm not sure I'm happy with it yet (we discussed about abstracting some parts to get rid of boilerplate, however i can think about that during another PR). 

You also mentioned that further validation should be done:

> Otherwise, looks good to me! It might be valuable as part of this work to also add in other parameter validations while we're making these changes and auditing dynamic parameters. For example:

Should I take another closer look? So far I don't have enough context knowledge to know what range checks would be meaningful, for example.
But we can move on, and when I'm more comfortable with these changes, I can think about this stuff.

SteveMacenski:
For this one, you can check if any parameter value is a `double` then it must be positive. All the floating point values in this particular algorithm must be `>= 0.0`. I think that's the only validation that can be done in this algorithm since the others are application dependent (i.e. lookaheads could be 10cm or 10m).

SteveMacenski (approved):

SteveMacenski:
I think we have the recipe down now! I think if you open the PR in this format, I can help on the parameter ranges and knock these out one at a time.

Next on your list is MPPI - lets skip that one and circle back since that parameter design is very different and disrupts our flow. 

Pose Progress Checker is next up. Lets do all the nav2_controller progress checker / goal checker plugins next

Pose Progress Checker
- required_movement_angle should be > 0 and < 2PI

Simple Goal Checker:
- XY goal tolerance should be >0
- Yaw should be 0 <= tol <= 2PI

Simple progress checker 
- required_movement_angle should be > 0 and < 2PI
- movement_time_allowance >= 0

Stopped goal checker 
- rot_stopped_velocity >= 0
- trans_stopped_velocity >= 0

Nianzu:
We just updated our nav2 after a while on the same version, and it looks like this PR is causing some issues for us. Is there some reason that a check for double's not being negative was added? We have had negative values in our param files for a long time, was there some change to how we are supposed to handle the yaml files that we missed?

SteveMacenski:
@Nianzu what value? I thought we checked and made sure to call out any negatives, but if we missed one or more, please let us know

Nianzu:
Sorry for the delay. I'm still trying to understand whats happening. It seems that this change is somehow applying outside of where it should. Our nav2_params.yaml file is affected, and we're seeing an error and then a crash anytime there is a negative value assigned to a double parameter. I'm wondering if the issue isn't in this PR, but in another later PR that causes this to apply to a wider scope than it should? Or is the parameter_handler.cpp here used elsewhere?

Nianzu:
We checked out [https://github.com/ros-navigation/navigation2/releases/tag/1.3.6](https://github.com/ros-navigation/navigation2/releases/tag/1.3.6) and that seemed to resolve the problem for us, but I haven't gone through the process of checking each release since to see if it was in fact a later PR that brought up this issue.

Nianzu:
Maybe this is something I should open a separate ticket for?

SteveMacenski:
To make sure I understand, you're saying a newer release fixed your issue? Then I think I know what's happening here and that was fully resolved - just use the newer release. I could dig through and find the specific patch for you, but we did find a couple of bugs with this initial PR that we resolved in the following weeks of its merge

Nianzu:
Sorry for the lack of clarity, that might be the case, but so far I just tested the release right before this PR and that worked. Best I can tell, whatever is in the package manager right now is still causing this problem. I will try the latest release tag and report back whether we're still seeing the problem.

PR #4974, opened 2025-03-09
nav2_rviz_plugins: Remove slots without implementation

wentasah:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | none so far |
| Primary OS tested on | Nix |

---

Without this change, when compiled for Jazzy with Nix using gcc14 and qt-5.15, loading the plugins to rviz fails with the following error:

    [rviz2]: PluginlibFactory: The plugin for class
    'nav2_rviz_plugins/Navigation 2' failed to load. Error: Failed to
    load library /nix/store/...-ros-env/lib/libnav2_rviz_plugins.so.
    Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in
    the library code, and that names are consistent between this macro
    and your XML. Error string: Could not load library dlopen error:
    /nix/store/...-ros-env/lib/libnav2_rviz_plugins.so: undefined
    symbol: _ZN17nav2_rviz_plugins12DockingPanel9onStartupEv, at
    /build/rcutils-release-release-jazzy-rcutils-6.7.2-1/src/shared_library.c:99

The dynamic linker complains about undefined symbol, which corresponds to a Qt slot, which is declared, but has no implementation. I'm not sure what is the root cause of this behavior, but removing the slot definitions fixes the problem and the plugins can be loaded.

This PR is related to https://github.com/lopsided98/nix-ros-overlay/pull/591#issuecomment-2708725318.

None

Running `rviz2` and manually adding Navigation 2 panel.
---

Testing with non-Nix builds. Hopefully, will be done by CI.

#### For Maintainers: 

### DIFF
--- a/nav2_rviz_plugins/include/nav2_rviz_plugins/costmap_cost_tool.hpp
+++ b/nav2_rviz_plugins/include/nav2_rviz_plugins/costmap_cost_tool.hpp
@@ -46,7 +46,6 @@ public:
   void handleGlobalCostResponse(rclcpp::Client<nav2_msgs::srv::GetCosts>::SharedFuture);
 
 private Q_SLOTS:
-  void updateAutoDeactivate();
 
 private:
   rclcpp::Client<nav2_msgs::srv::GetCosts>::SharedPtr local_cost_client_;
--- a/nav2_rviz_plugins/include/nav2_rviz_plugins/docking_panel.hpp
+++ b/nav2_rviz_plugins/include/nav2_rviz_plugins/docking_panel.hpp
@@ -57,7 +57,6 @@ public:
 
 private Q_SLOTS:
   void startThread();
-  void onStartup();
   void onDockingButtonPressed();
   void onUndockingButtonPressed();
   void onCancelDocking();


SteveMacenski (approved):

SteveMacenski:
Thanks, looks good. Looks like a simple copy+paste error that someone started with the nav2 panel and didn't remove old methods for the docking panel

SteveMacenski:
Please open a PR to readd!

wentasah:
No.

It seems that the `SLOT(updateAutoDeactivate())` should be replaced with `nullptr`. The parameter [expects the slot as string](https://github.com/ros2/rviz/blob/bc21eb832d02b23501a25ed6bdf51c1cc63020e3/rviz_common/src/rviz_common/properties/bool_property.cpp#L45) and tries to connect the "property" to that slot [here](https://github.com/ros2/rviz/blob/bc21eb832d02b23501a25ed6bdf51c1cc63020e3/rviz_common/src/rviz_common/properties/property.cpp#L89). Since the slot doesn't exist, it complains, but even if the slot exists, it will not execute anything (I guess), because the slot didn't have any implementation.

I can send a PR to remove it from here iv @SteveMacenski  agrees.

SteveMacenski:
That makes sense - sure! Can you tag @mini-1235 in your PR to give a review to make sure it resolves the report?

PR #4983, opened 2025-03-13
Switch from geometry_msgs/PoseStampedArray to nav_msgs/Goals

tfoote:
This is a follow on to https://github.com/ros2/common_interfaces/pull/269 

And is blocking https://github.com/ros2/common_interfaces/pull/270

I've done my best to replicate the changes in: #4791  

As this is a code base that I'm not familiar with I suspect that it won't be perfect as I haven't had a chance to iterate or test at all. 

I think that there's one tweak to resort includes with the new name in places. I suspect that the linter will error on that. 

Unfortunately I'm going to have to submit this and won't have time for significant followups as I'm going on leave for the next two weeks and I think that we don't want to wait that long to resolve this. @ahcorde maybe you can take a look to help get this across the line and unblock 270 above.

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_utils.hpp
@@ -136,18 +136,18 @@ inline std::vector<geometry_msgs::msg::PoseStamped> convertFromString(const Stri
 }
 
 /**
- * @brief Parse XML string to geometry_msgs::msg::PoseStampedArray
+ * @brief Parse XML string to nav_msgs::msg::Goals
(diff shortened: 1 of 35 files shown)


SteveMacenski:
@tfoote did you see https://github.com/ros-navigation/navigation2/pull/4980? I think a contributor beat you to it in response to the ticket I filed

tfoote:
I hadn't, this was a backlog thing I'm trying to get out before I go off the grid and I'm already mostly disconnected. I'm happy to defer to that implementation.

SteveMacenski:
Great, I appreciate the follow up on this! Have a good time off the grid!

Issue #4985, opened 2025-03-14
Input port parsing issue in AreErrorCodesPresent BT node


DylanDeCoeyer-Quimesis:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Computer:
  - 
- ROS2 Version:
  - All
- Version or commit hash:
  - 
- DDS implementation:
  - 

#### Steps to reproduce issue

Add the node to a BT, set an error code in a BT variable, and try to use this node to compare the variable against the expected value.

#### Expected behavior
It returns BT::NodeStatus::SUCCESS.
#### Actual behavior
It returns BT::NodeStatus::FAILURE.
#### Additional information
The `AreErrorCodesPresent` BT node has an input port called `error_codes_to_check`. This [port](https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/are_error_codes_present_condition.hpp#L58) is expecting a `std::set<uint16_t>`. However, BehaviorTree-CPP's `getInput` function is unable to parse and convert a `string` to `std::set<uint16_t>`. So, this [line](https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/are_error_codes_present_condition.hpp#L37) has no effect, meaning that the node is useless.

To make `AreErrorCodesPresent` work, I suggest to define the input port as a `std::vector<int>` which is supported by the `getInput` function. Once loaded, the vector can be converted to a `std::set<uint16_t>`.

In addition, the example provided in the [documentation](https://docs.nav2.org/configuration/packages/bt-plugins/conditions/AreErrorCodesPresent.html#example) suggests that the provided error codes to check should be comma-separated. In BehaviorTree-CPP's convertFromString implementation, I can see no case where the comma (,) is used. Only the semicolon (;) is used as separator. So, the documentation is likely incorrect.

DylanDeCoeyer-Quimesis:
I made a PR for the documentation as well: https://github.com/ros-navigation/docs.nav2.org/pull/655

PR #4986, opened 2025-03-14
nav2_behavior_tree: fix input port parsing error in AreErrorCodesPresent

DylanDeCoeyer-Quimesis:
The getInput method does not support std::set<uint16_t> parsing. So, let's replace the type of the input port by std::vector<int> which is supported, and convert the list to a std::set<uint16_t>.

This commit fixes issue #4985.

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/are_error_codes_present_condition.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/are_error_codes_present_condition.hpp
@@ -34,7 +34,12 @@ public:
     const BT::NodeConfiguration & conf)
   : BT::ConditionNode(condition_name, conf)
   {
-    getInput<std::set<uint16_t>>("error_codes_to_check", error_codes_to_check_); //NOLINT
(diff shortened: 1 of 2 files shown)


DylanDeCoeyer-Quimesis:
I made a PR for the documentation as well: https://github.com/ros-navigation/docs.nav2.org/pull/655

SteveMacenski (approved):
I'm pretty irritated that this wasn't tested thoroughly by the original developer, but I see looking at the unit tests for this node how this got based us. This was the base class of the derived Controller/Planner/Smoother variants that set the values in their constructor based on contextual knowledge, so I see how this may not have been appropriately tested as a stand alone node. That's my bad in quality control, sorry!

SteveMacenski:
I think the unit test just needs the input type updated to vector:

```
unknown file
C++ exception with description "The creation of the tree failed because the port [error_codes_to_check] was initially created with type [std::set<unsigned short, std::less<unsigned short>, std::allocator<unsigned short> >] and, later type [std::vector<int, std::allocator<int> >] was used somewhere else." thrown in SetUp().
```

PR #4994, opened 2025-03-19
Make NavigateThroughPoses navigator report waypoint statuses information. (fix #4846)

### LINKED ISSUE #4846: More info about the state of poses in NavigateThroughPoses
#### Feature description

I have my application calling NavigateThroughPoses and using a BT similar to the demo [one](https://github.com/ros-navigation/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_through_poses_w_replanning_and_recovery.xml). 
I would like to pass back to my application more information on the state of the poses as the action progresses; for example the poses removed by RemovePassedGoals would be "completed". The poses removed by "RemoveInCollisionGoals" would be "skipped/aborted".

This would be useful for the application node for various reasons, some examples:

- going back to the unsuccessful poses in a coverage application
- maintain a list of completed poses for pausing/resuming purposes
- visualizing the poses by color

The BT nodes would communicate this feedback with the navigator through the blackboard and the navigator would pass it back through the action's feedback channel. I'm not sure in what form though, it could be through separate values e.g. int[] skipped_poses or a supercharged wrapper around geometry_msgs/PoseStampedArray poses.

Let me know your general thoughts

zz990099:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4846 |

---

* This PR addresses #4846.
* Make NavigateThroughPoses navigator report waypoint statuses information. 
* More information could be exposed by extending the `WaypointState` message type.

* A parameter called `waypoint_statuses_blackboard_id` in `NavigateThroughPoses` should be added in documentation.
* I will submit a PR to [docs.nav2.org](https://github.com/ros-navigation/docs.nav2.org) if this PR could been merged.

* I test these codes in gazebo simulation with some modifications in `nav2_rviz_plugins`(which is not included in this PR). 
* Test results validate proper code execution.

---

#### For Maintainers: 


nav2_msgs/action/NavigateThroughPoses.action
```diff
@@ -22,3 +22,4 @@ builtin_interfaces/Duration estimated_time_remaining
 int16 number_of_recoveries
 float32 distance_remaining
 int16 number_of_poses_remaining
+WaypointStatus[] waypoint_statuses
```
SteveMacenski, line 25: I think this should be in the `result` too so that its both on feedback for live updates but as well in the final result of the task

nav2_msgs/msg/WaypointStatus.msg
```diff
@@ -0,0 +1,5 @@
+# Waypoint state for robot in NavigateThroughPoses or WaypointFollower
+uint8 PENDING=0 # Waypoint is not processed
+uint8 COMPLETED=1 # Waypoint is completed
+uint8 SKIPPED=2 # Waypoint is skipped
+uint8 waypoint_status
```
SteveMacenski, line 5: Make sure to add an end of file line

nav2_msgs/msg/WaypointStatus.msg
```diff
@@ -0,0 +1,5 @@
+# Waypoint state for robot in NavigateThroughPoses or WaypointFollower
+uint8 PENDING=0 # Waypoint is not processed
```
SteveMacenski, line 2:
```suggestion
uint8 PENDING=0 # Waypoint is not processed or processing
```

nav2_msgs/msg/WaypointStatus.msg
SteveMacenski, line 5: It might be useful to add in the posestamped here as well of the goal for context

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -63,6 +64,10 @@ class RemoveInCollisionGoals : public BtServiceNode<nav2_msgs::srv::GetCosts>
           "Whether to consider unknown cost as obstacle"),
         BT::OutputPort<nav_msgs::msg::Goals>("output_goals",
           "Goals with in-collision goals removed"),
+        BT::InputPort<std::vector<nav2_msgs::msg::WaypointStatus>>("input_waypoint_statuses",
```
SteveMacenski, line 67: Instead: you could have a `string` input port that is `waypoint_statuses_id` that is the string to the waypoint status blackboard entry that you can `get` and then `set` as needed. As well for the other BT node

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -63,12 +63,18 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
     return BT::NodeStatus::FAILURE;
   }
 
+  // get the `waypoint_statuses` vector if provided
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
```
SteveMacenski, line 67: Then here you get the input as a string, and then find that in the blackboard directly to get/set

SteveMacenski (changes requested):
I'm not sure this work for the RemoveInCollision considering it overrides the `goals` variable with a pruned list rather than maintaining it. When it removes from `goals` the side of `waypoint_statuses` and `goals` no longer match. If this is called repeatedly and different ones are removed, the `i` that you're setting to `SKIPPED` would no longer map to the actual waypoint.

I think what we might want to do is actually to update the way that the `goal` and `goals` are stored in the blackboard from `geometry_msgs/PoseStamped` and `nav_msgs/Goals` (respectively) to being a struct containing those values, the current goal, and the waypoint statuses. Then the BT nodes would modify metadata on these to skip or mark as completed.

This is a bit more complicated than we initially discussed, however! 

---

Some food for thought: We also have this https://github.com/ros-navigation/navigation2/blob/f2755f66ba550332aa603485ccba267c8a33236a/nav2_msgs/msg/MissedWaypoint.msg message as well. I'm curious if we can combine this message and the one you made or make a new one that combines the use cases for the waypoint follower and this navigate through poses application!

zz990099:
Thank you for your patient guidance and thorough review! My apologies—I overlooked the possibility of `RemoveInCollision` being called multiple times in the initial implementation. After carefully reviewing your feedback and revisiting the code, I’d like to discuss a few considerations:

1. ​Impact of `goal`/`goals` Type Refactoring:
Taking `NavigateThroughPoses` as an example, changing the type of `goals` in the blackboard would indeed affect the port implementations of multiple BTAction nodes (`ComputePathThroughPoses`/`GoalUpdated`...) and potentially the logic of `NavigateToPose`. 

2. `PoseStamped` Matching Approach:
Your proposed solution—matching indices based on the `PoseStamped` value of the current goal with `find_first_of` method—is practical! I will prioritize implementing this and test it thoroughly to ensure indexing stability across repeated `RemoveInCollision` calls.

3. ​Combining `MissedWaypoint` and `WaypointStatus`:
Perhaps we could integrate the `MissedWaypoint` concept into the `WaypointStatus` message by adding fields like:
`waypoint_status (e.g., MISSED, SKIPPED, ACTIVE, COMPLETED)`
`goal_index`
`error_code`
`error_msg`
`goal`
This would allow `WaypointFollower` to directly use `WaypointStatus` messages to mark the failed points.

Thank you again for your thoughtful review and suggestions! Please let me know if these directions align with your expectations.

SteveMacenski:
No worries, that's what code reviews are for! I didn't realize that was going to be the case either before looking at the code itself :wink: 

1. Agreed, I think maybe we should not do this, even though it might be the 'cleanest' solution, due to the massive  breaking change it would cause people. Also (2) is not actually that bad of a solution overall. It involves some similar elements of looking over the vector to find the goals, but this one would also impact how the ComputePathThroughPoses looks as the goals parameter which would be problematic for other users with highly customized use-cases of it. 

2. Sounds good! I think having a couple of utility functions to do the lookups would be nice so that they can be reused in both BT nodes and other user custom nodes (and easily unit testable).

3. Agreed!

zz990099:
Hello @SteveMacenski !,

I’ve implemented most of the changes as discussed, but I’m encountering an edge case issue with how `RemovePassedGoals` handles empty goals. Currently, `RemovePassedGoals` skips returning empty goals to avoid triggering exceptions in `ComputePathThroughPoses`(as per [link](https://github.com/ros-navigation/navigation2/blob/d9cc4ba320e327c32ffaa52aa1fdca3fc4af1d2f/nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp#L71)). However, this prevents the last waypoint's status in `waypoint_statuses` from being updated properly.

Here are potential fixes I’ve considered:

1. Allow empty goals in `ComputePathThroughPoses`: Return success instead of throwing an error. While simple, this might introduce unintended side effects for other modules and test cases.

2. Add edge-case handling within `RemovePassedGoals`: This could clutter the existing logic and introduce maintenance overhead.

3. ​Update `goalCompleted` in `ComputeThroughPoses`: Handle the last waypoint’s status here. Though straightforward, this may decentralize the `waypoint_statuses` management logic.

I’d appreciate your guidance on which approach aligns best with the project’s design goals. Happy to refine any of these paths ​or any suggestions you might have! :)

SteveMacenski:
> Add edge-case handling within RemovePassedGoals: This could clutter the existing logic and introduce maintenance overhead.

What do you have in mind?

> ​Update goalCompleted in ComputeThroughPoses: Handle the last waypoint’s status here. Though straightforward, this may decentralize the waypoint_statuses management logic.

I think this is a pretty good option. In the migration / docs updates we'll do for this PR, lets make sure to highlight where this is managed (Navigator to initialize; goal complete for goal. removed / passed goals for state updates on completion or failure) & that users can add more BT nodes that modify it for their applications at their set blackboard ID.

zz990099:
> Add edge-case handling within RemovePassedGoals: This could clutter the existing logic and introduce maintenance overhead.

To prevent empty goals output while ensuring all waypoint statuses are properly recorded, ​perhaps we could modify the code in `RemovePassedGoals` as follows:
```cpp
  // get the `waypoint_statuses` vector
  std::string waypoint_statuses_id;
  getInput("waypoint_statuses_id", waypoint_statuses_id);
  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
  waypoint_statuses = config().blackboard->get<decltype(waypoint_statuses)>(waypoint_statuses_id);

  double dist_to_goal;
  while (goal_poses.goals.size() > 0) {
    dist_to_goal = euclidean_distance(goal_poses.goals[0].pose, current_pose.pose);

    if (dist_to_goal > viapoint_achieved_radius_) {
      break;
    }

    // mark waypoint statuses before the goal is erased from goals
    auto cur_waypoint_index =
      find_goal_in_waypoint_statuses(waypoint_statuses, goal_poses.goals[0]);
    if (cur_waypoint_index != -1) {
      waypoint_statuses[cur_waypoint_index].waypoint_status =
        nav2_msgs::msg::WaypointStatus::COMPLETED;
    }
    
    // prevent from removing the last goal
    if (goal_poses.goals.size() == 1) break;

    goal_poses.goals.erase(goal_poses.goals.begin());
  }

  setOutput("output_goals", goal_poses);
  // set `waypoint_statuses` output
  config().blackboard->set<decltype(waypoint_statuses)>(waypoint_statuses_id, waypoint_statuses);
```
This should ensure goals are never empty and consistently track all waypoint statuses. 
However, if we need to throw an exception when `find_goal_in_waypoint_statuses` fails to find a suitable result, additional logic may need to be added to this part.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,37 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
 
+  // get the `waypoint_statuses` vector
+  std::string waypoint_statuses_id;
+  getInput("waypoint_statuses_id", waypoint_statuses_id);
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  waypoint_statuses = config().blackboard->get<decltype(waypoint_statuses)>(waypoint_statuses_id);
+
   double dist_to_goal;
-  while (goal_poses.goals.size() > 1) {
+  while (goal_poses.goals.size() > 0) {
```
zz990099, line 77: Enable handling the last waypoint to ensure status consistency.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,37 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
+    // mark waypoint statuses before the goal is erased from goals
+    auto cur_waypoint_index =
+      find_goal_in_waypoint_statuses(waypoint_statuses, goal_poses.goals[0]);
+    if (cur_waypoint_index != -1) {
+      waypoint_statuses[cur_waypoint_index].waypoint_status =
+        nav2_msgs::msg::WaypointStatus::COMPLETED;
+    }
+
+    // prevent from removing the last goal
+    if (goal_poses.goals.size() == 1) {break;}
```
zz990099, line 93: prevent from removing the last goal

zz990099:
Hi! @SteveMacenski 

I've completed the code revisions. Could you please review the changes when you have a moment?

Regarding the WaypointFollower component (currently uncommitted), I'm considering :

1. Keep returning missed_waypoints in the result via FollowWaypoint.action, or
2. Align with NavigateThroughPoses by returning a full status collection of all waypoints?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.hpp
```diff
@@ -63,6 +64,8 @@ class RemoveInCollisionGoals : public BtServiceNode<nav2_msgs::srv::GetCosts>
           "Whether to consider unknown cost as obstacle"),
         BT::OutputPort<nav_msgs::msg::Goals>("output_goals",
           "Goals with in-collision goals removed"),
+        BT::InputPort<std::string>("waypoint_statuses_id",
+          "The id of input waypoint_statuses maintained in blackboard")
```
SteveMacenski, line 68:
```suggestion
          "The id of the waypoint_statuses maintained in blackboard")
```

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_passed_goals_action.hpp
```diff
@@ -49,6 +50,8 @@ class RemovePassedGoals : public BT::ActionNodeBase
           "Goals with passed viapoints removed"),
       BT::InputPort<double>("radius", 0.5, "radius to goal for it to be considered for removal"),
       BT::InputPort<std::string>("robot_base_frame", "Robot base frame"),
+      BT::InputPort<std::string>("waypoint_statuses_id",
+          "The id of input waypoint_statuses maintained in blackboard")
```
SteveMacenski, line 54:
```suggestion
          "The id of the waypoint_statuses maintained in blackboard")
```

nav2_msgs/msg/WaypointStatus.msg
```diff
@@ -0,0 +1,10 @@
+# Waypoint state for robot in NavigateThroughPoses or WaypointFollower
```
SteveMacenski, line 1: You can remove the missed waypoints msg if its not used anymore

SteveMacenski, line 1:
```suggestion
# Waypoint state for multi-goal navigation or waypoint following
```

nav2_msgs/msg/WaypointStatus.msg
```diff
@@ -0,0 +1,10 @@
...
+uint8 PENDING=0 # Waypoint is not processed or processing
+uint8 COMPLETED=1 # Waypoint is completed
+uint8 SKIPPED=2 # Waypoint is skipped
+uint8 MISSED=3 # Waypoint is missed
+uint8 waypoint_status
+uint32 index
```
SteveMacenski, line 7: `waypoint_index` and `waypoint_pose` I think would be good

nav2_waypoint_follower/src/waypoint_follower.cpp
```diff
@@ -305,7 +305,7 @@ void WaypointFollower::followWaypointsHandler(
       current_goal_status_.status == ActionStatus::FAILED ||
       current_goal_status_.status == ActionStatus::UNKNOWN)
     {
-      nav2_msgs::msg::MissedWaypoint missedWaypoint;
+      nav2_msgs::msg::WaypointStatus missedWaypoint;
```
SteveMacenski, line 308: For both in this file: make sure to populate the status!

nav2_msgs/msg/WaypointStatus.msg
```diff
@@ -0,0 +1,10 @@
+# Waypoint state for robot in NavigateThroughPoses or WaypointFollower
+uint8 PENDING=0 # Waypoint is not processed or processing
+uint8 COMPLETED=1 # Waypoint is completed
+uint8 SKIPPED=2 # Waypoint is skipped
+uint8 MISSED=3 # Waypoint is missed
```
SteveMacenski, line 5: I think MISSED is redundant with SKIPPED, no? 

I think we could keep MISSED and remove SKIPPED. What do you think?

nav2_util/include/nav2_util/geometry_utils.hpp
```diff
@@ -179,6 +181,30 @@ inline double calculate_path_length(const nav_msgs::msg::Path & path, size_t sta
...
 }
 
+/**
+ * @brief Find the index of the first goal in `PENDING` status matching the
+ * given target pose.
+ * @param waypoint_statuses List of waypoint statuses to search through.
+ * @param goal Target pose to match against waypoint goals.
+ * @return Index of the first matching goal in PENDING status, -1 if not found.
+ */
+inline int find_goal_in_waypoint_statuses(
```
SteveMacenski, line 191: This needs unit testing :-)

SteveMacenski: I think this also needs to be renamed, since its not just finding the goal, its finding the next pending goal. How about 'find_next_matching_goal_in_waypoint_statuses'?

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -262,6 +275,16 @@ NavigateThroughPosesNavigator::initializeGoalPoses(ActionT::Goal::ConstSharedPtr
   blackboard->set<nav_msgs::msg::Goals>(goals_blackboard_id_,
       std::move(goals_array));
 
+  // Reset the waypoint states vector in the blackboard
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses(goals_array.goals.size());
+  for (size_t waypoint_index = 0 ; waypoint_index < goals_array.goals.size() ; ++waypoint_index) {
+    waypoint_statuses[waypoint_index].index = waypoint_index;
```
SteveMacenski, line 281: Question: do we need the index if we store the pose and we know they should be ordered, even if remove passed goals edits the 'goals' variable?I think the index is implicit now by the waypoint status vector dimensions

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -262,6 +275,16 @@ NavigateThroughPosesNavigator::initializeGoalPoses(ActionT::Goal::ConstSharedPtr
...
+    waypoint_statuses[waypoint_index].goal = goals_array.goals[waypoint_index];
+  }
+  blackboard->set<decltype(waypoint_statuses)>(waypoint_statuses_blackboard_id_,
+      std::move(waypoint_statuses));
+  blackboard->set<std::string>("waypoint_statuses_id", waypoint_statuses_blackboard_id_);
```
SteveMacenski, line 286: Remove, I think

nav2_bt_navigator/behavior_trees/navigate_through_poses_w_replanning_and_recovery.xml
```diff
@@ -14,7 +14,7 @@
         <RateController hz="0.333">
           <RecoveryNode number_of_retries="1" name="ComputePathThroughPoses">
             <ReactiveSequence>
-              <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.7"/>
+              <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.7" waypoint_statuses_id="{waypoint_statuses_id}" />
```
SteveMacenski, line 17: I think this should be 'waypoint_statuses', analog to the path and goal/goals handling.

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -63,12 +64,31 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
...
   }
 
+  // get the `waypoint_statuses` vector
+  std::string waypoint_statuses_id;
+  auto waypoint_statuses_id_get_res = getInput("waypoint_statuses_id", waypoint_statuses_id);
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  if (waypoint_statuses_id_get_res) {
+    waypoint_statuses =
+      config().blackboard->get<decltype(waypoint_statuses)>(waypoint_statuses_id);
+  }
```
SteveMacenski, line 74: Maybe log warn once if its not found / available?

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -63,12 +64,31 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
...
     if ((response->costs[i] == 255 && !consider_unknown_as_obstacle_) ||
       response->costs[i] < cost_threshold_)
     {
       valid_goal_poses.goals.push_back(input_goals_.goals[i]);
+    } else if (waypoint_statuses_id_get_res) {
+      using namespace nav2_util::geometry_utils;  // NOLINT
+      auto cur_waypoint_index =
+        find_goal_in_waypoint_statuses(waypoint_statuses, input_goals_.goals[i]);
+      if (cur_waypoint_index == -1) {
+        throw std::runtime_error(
```
SteveMacenski, line 87: If all uses of -1 return type are to throw, we can throw in the function itself!

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,44 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
   }
 
+  // get the `waypoint_statuses` vector
+  std::string waypoint_statuses_id;
+  auto waypoint_statuses_id_get_res = getInput("waypoint_statuses_id", waypoint_statuses_id);
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  if (waypoint_statuses_id_get_res) {
+    waypoint_statuses =
+      config().blackboard->get<decltype(waypoint_statuses)>(waypoint_statuses_id);
+  }
```
SteveMacenski, line 77: Maybe log warn once if its not found / available?

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,44 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
 
     if (dist_to_goal > viapoint_achieved_radius_) {
       break;
     }
 
+    // mark waypoint statuses before the goal is erased from goals
+    if (waypoint_statuses_id_get_res) {
+      auto cur_waypoint_index =
+        find_goal_in_waypoint_statuses(waypoint_statuses, goal_poses.goals[0]);
+      if (cur_waypoint_index != -1) {
```
SteveMacenski, line 91: If it doesn't throw (see comment above) then we know its not -1

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
SteveMacenski: Nice, just expand this so its not all on one line, like the rest of the object's styling :-)

SteveMacenski (commented):
Nice work! A few things to resolve and I think this is close to merge!

SteveMacenski:
> However, if we need to throw an exception when find_goal_in_waypoint_statuses fails to find a suitable result, additional logic may need to be added to this part.

If we fail, we should use another one of the callbacks (like the on completion callback) that knows that it failed to populate the remaining waypoints as SKIPPED / MISSED, I think. We catch exceptions and return a BT Status of Failure that you can check and if so, any pending goals should be set to skipped / missed. 

This is a subtle point and may be worth some automated and/or manual test coverage to make sure the feedback/returns are properly populated in success, nominal failure conditions (not related to this), and failure conditions related to this. 

> While modifying the code, I noticed that this approach ensures the waypoint_statuses in the result meet expectations, ​but it does not update the waypoint_statuses in the feedback, leading to inconsistencies between the two. This mismatch might introduce potential hidden issues in user's application.

Can you expand on these?

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
SteveMacenski: Oh I see, this is needed for the WPF, I think we should keep

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
zz990099: I think returning `-1` when no match is found might be preferable for this function. Here’s my consideration:

1. Throwing exceptions could lead to repetitive handling in `RemovePassedGoals`, as this function would fail to find a match when the robot is nearing the last waypoint and the last waypoint is already marked before.

2. Since this function’s responsibility is to find matches, determining whether a missing result is exceptional should be delegated to its caller.

These are my initial thoughts. Please let me know if this doesn’t align with your expectations—I’d be happy to adjust!

tonynajjar:
Nice to see this PR and thank you for the good work @zz990099 . I implemented something similar in my fork, I'll try to find the time (maybe weekend) to go over the differences and give my 2 cents

zz990099:
Thank you so much for your kind words, @tonynajjar ! 🙌 I truly appreciate you taking the time to review this PR and share your perspective. Looking forward to your thoughts whenever it’s convenient for you.

tonynajjar:
> ​Impact of goal/goals Type Refactoring:
Taking NavigateThroughPoses as an example, changing the type of goals in the blackboard would indeed affect the port implementations of multiple BTAction nodes (ComputePathThroughPoses/GoalUpdated...) and potentially the logic of NavigateToPose.

This is actually what I went for in my fork; I created a new Waypoint msg that contains a PoseStamped and a status. But yes like mentioned, that's a breaking change and needed to change multiple things. I think on the long-run this would probably be the "cleaner" way but I'm not against, even probably prefer, having a proof of concept that is not breaking, like implemented in this PR.


> uint8 PENDING = 0
uint8 VISITED = 1
uint8 SKIPPED = 2
uint8 FAILED = 3


These are the statuses I came up with. We match on PENDING. I have COMPLETED instead of VISITED; I don't have a strong preference or really any argument for or against either.
I have SKIPPED and FAILED which are combined in this PR into MISSED: I think it would be useful to make a differentiation between waypoints we missed "on purpose" (SKIPPED) and those we missed because of some kind of error. E.g. I have RemoveGoalsInCollision mark waypoints as SKIPPED whereas the navigator could mark a waypoint as FAILED if the BT fails or ComputePathThroughPoses could mark it as FAILED if it fails to generate a plan


Otherwise design looks good!

zz990099:
Hi @tonynajjar !

Thank you for the suggestions! This is really helpful to me.

> But yes like mentioned, that's a breaking change and needed to change multiple things. I think on the long-run this would probably be the "cleaner" way.

Agreed! However, I’m not entirely confident about handling such a breaking change in a large-scale open-source project like this 🙏.

> I think it would be useful to make a differentiation between waypoints we missed "on purpose" (SKIPPED) and those we missed because of some kind of error.

Based on your reasoning, I also think splitting `MISSED` into `SKIPPED` and `FAILED` makes sense.

zz990099:
Hi, @SteveMacenski !

> While modifying the code, I noticed that this approach ensures the waypoint_statuses in the result meet expectations, but it does not update the waypoint_statuses in the feedback, leading to inconsistencies between the two. This mismatch might introduce potential hidden issues in user's application.

Apologies for the confusion earlier! I wanted to clarify: modifying `waypoint_statuses` in the `goalCompleted` function based on the behavior tree's `final_bt_status` updates the action's **result** section, but the **feedback** section's `waypoint_statuses` may remain unupdated.

I’ve addressed most of your review feedback in the updated code, with ​two points may remaining for discussion:

1. Exception handling in `find_next_matching_goal_in_waypoint_statuses`:
I’ve shared my thoughts in the review thread about whether to throw an exception when no matches are found. [link](https://github.com/ros-navigation/navigation2/pull/4994#discussion_r2018283584)

2. Waypoint status design:
@tonynajjar and I discussed whether to merge `SKIPPED` and `FAILED` statuses. We think keeping them separate would help ​clearly distinguish between intentional skips (e.g., collision avoidance) and unexpected failures (e.g., planning errors).

Looking forward to your thoughts on these!

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,44 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
 
     if (dist_to_goal > viapoint_achieved_radius_) {
       break;
     }
 
+    // mark waypoint statuses before the goal is erased from goals
+    if (waypoint_statuses_get_res) {
+      auto cur_waypoint_index =
+        find_next_matching_goal_in_waypoint_statuses(waypoint_statuses, goal_poses.goals[0]);
+      if (cur_waypoint_index != -1) {
```
SteveMacenski, line 89: If `== -1` we need to error handle here, no?

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,44 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
+  // get the `waypoint_statuses` vector
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  auto waypoint_statuses_get_res = getInput("input_waypoint_statuses", waypoint_statuses);
+  if (!waypoint_statuses_get_res) {
+    RCLCPP_WARN(node_->get_logger(), "Missing [input_waypoint_statuses] port input!");
+  }
+
   double dist_to_goal;
-  while (goal_poses.goals.size() > 1) {
+  while (goal_poses.goals.size() > 0) {
```
SteveMacenski, line 78: This could be changed to a `do while` rather than a `while` to check the condition at the end to keep the `size() > 1` and not need the addition `if` check on it to `break`

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
SteveMacenski: Isn't a lack of matching goal a serious issue? The goal poses is the set of goals, if we are not able to find one, that's a critical error that something is not right. We should at the very least be logging a loud error while returning `FAILURE`. If not, throwing an exception to kill the navigation task.

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -63,12 +64,29 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
     return BT::NodeStatus::FAILURE;
   }
 
+  // get the `waypoint_statuses` vector
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  auto waypoint_statuses_get_res = getInput("input_waypoint_statuses", waypoint_statuses);
+  if (!waypoint_statuses_get_res) {
+    RCLCPP_WARN(node_->get_logger(), "Missing [input_waypoint_statuses] port input!");
```
SteveMacenski, line 71:
```suggestion
    RCLCPP_ERROR(node_->get_logger(), "Missing [input_waypoint_statuses] port input!");
```

nav2_behavior_tree/plugins/action/remove_in_collision_goals_action.cpp
```diff
@@ -78,6 +96,11 @@ BT::NodeStatus RemoveInCollisionGoals::on_completion(
       "All goals are in collision and have been removed from the list");
   }
   setOutput("output_goals", valid_goal_poses);
+  // set `waypoint_statuses` output
+  if (waypoint_statuses_get_res) {
```
SteveMacenski, line 100: You don't need the `if` here, it would just set it to an empty struct since `waypoint_statuses` was never populated.

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,44 @@ inline BT::NodeStatus RemovePassedGoals::tick()
     return BT::NodeStatus::FAILURE;
   }
 
+  // get the `waypoint_statuses` vector
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  auto waypoint_statuses_get_res = getInput("input_waypoint_statuses", waypoint_statuses);
+  if (!waypoint_statuses_get_res) {
+    RCLCPP_WARN(node_->get_logger(), "Missing [input_waypoint_statuses] port input!");
```
SteveMacenski, line 74:
```suggestion
    RCLCPP_ERROR(node_->get_logger(), "Missing [input_waypoint_statuses] port input!");
```

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,18 +67,44 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
+    if (goal_poses.goals.size() == 1) {
+      break;
+    }
+
     goal_poses.goals.erase(goal_poses.goals.begin());
   }
 
   setOutput("output_goals", goal_poses);
+  // set `waypoint_statuses` output
+  if (waypoint_statuses_get_res) {
```
SteveMacenski, line 105: Ditto on `if`

SteveMacenski (changes requested):

nav2_bt_navigator/behavior_trees/navigate_through_poses_w_replanning_and_recovery.xml
```diff
@@ -14,7 +14,7 @@
         <RateController hz="0.333">
           <RecoveryNode number_of_retries="1" name="ComputePathThroughPoses">
             <ReactiveSequence>
-              <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.7"/>
+              <RemovePassedGoals input_goals="{goals}" output_goals="{goals}" radius="0.7" input_waypoint_statuses="{waypoint_statuses}" output_waypoint_statuses="{waypoint_statuses}"/>
```
SteveMacenski, line 17: In our Nav2 docs, please update all BT uses of RemovePassedGoals / RemoveInCollision to have these set. I think our migration guide also needs to mention that NavigateThroughPoses navigator now has this waypoint statuses struct in the feedback/result. Also mention how it is managed and what its role is (notifying users about the status of each goal set in the request)

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
```diff
@@ -67,6 +67,13 @@ inline BT::NodeStatus RemovePassedGoals::tick()
...
 
+  // get the `waypoint_statuses` vector
+  std::vector<nav2_msgs::msg::WaypointStatus> waypoint_statuses;
+  auto waypoint_statuses_get_res = getInput("input_waypoint_statuses", waypoint_statuses);
+  if (!waypoint_statuses_get_res) {
+    RCLCPP_ERROR(node_->get_logger(), "Missing [input_waypoint_statuses] port input!");
+  }
+
   double dist_to_goal;
   while (goal_poses.goals.size() > 1) {
```
SteveMacenski, line 78: DIdn't you need this to be a `do {} (while)` loop?

SteveMacenski (changes requested):
Otherwise, LGTM!

nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp
zz990099: Sorry for not mentioning this change earlier! Since the edge cases are now handled in `NavigateThroughPoses`'s `goalCompleted` logic, we no longer need to address them in `RemovePassedGoals`. To align with its original behavior, I’ve reverted the condition from `goal_poses.goals.size() > 0` back to `goal_poses.goals.size() > 1`. Testing confirms the results remain consistent with the previous version.

zz990099:
Hi!

I’ve submitted the PR for the documentation updates. https://github.com/ros-navigation/docs.nav2.org/pull/669

> I think our migration guide also needs to mention that NavigateThroughPoses navigator now has this waypoint statuses struct in the feedback/result. Also mention how it is managed and what its role is (notifying users about the status of each goal set in the request)

I’m not entirely clear on where to make these changes. Could you kindly point me to the specific sections of the documentation that need adjustments? Thank you for your guidance!

SteveMacenski:
I followed up on that point in the docs pR!

SteveMacenski (approved):
Just a few small changes for the docs PR and happy to merge the pair!

SteveMacenski:
Thanks for this! This is a really wonderful new feature!

tonynajjar:
I created the issue: https://github.com/ros-navigation/navigation2/issues/5213

>  (and possibly helping bring to fruition?).

I'm not sure I can get the time for it right now since we have a working version.

icpmoles:
I noticed this was backported to Humble. @SteveMacenski can you do a mergify to the jazzy branch too?

mini-1235:
This PR modify the message definition, and I don't think we can backport this to Jazzy. I don't see where it is backported to humble, are you referring to humble_main?

icpmoles:
Oh yes, my bad. I didn't know the difference.

mini-1235:
The humble_main branch is a backport from main that introduces a lot of API/ABI changes. It is not released to the buildfarm, but it can be built from source. You can find more context here: https://github.com/ros-navigation/navigation2/issues/5212.

Unfortunately, this PR also breaks the API/ABI, so we couldn't backport it to Jazzy. However, you can still get this functionality by building the main branch from source on Jazzy as well

Issue #4361, opened 2024-05-22
Full footprint collsion distance in MPPI obstacle critic and use of collision_margin_distance.


mangoschorle:
The distance to an obstacle computation (if using full footprint checking) is lower bound by the inscribed radius.

The inflation layer therefore doesn't support distance computations for a full footprint in situations where it would be interesting (e.g. driving along walls).

For example, irrespective of the orientation of the red footprint we would get the same distance result (=inscribed radius, in this case many times larger than the actual distance) and even though we might get arbitrarily close to an obstacle, `collision_margin_distance`would never be reached and the assumed distance would be the inscribed radius (in this case 0.6m).

My point being: I fail to see the use of the footprint distance checking. Imho would be of great help, but only given that we computed the actual distances. 

Do I have a knot in my head, or does it make sense what I am saying here. Does it make sense to try to calculate the actual distances efficiently?

![image](https://github.com/ros-navigation/navigation2/assets/94619337/99357a0b-1553-4c8d-b734-672a3bbb8d51)

SteveMacenski:
> My point being: I fail to see the use of the footprint distance checking. Imho would be of great help, but only given that we computed the actual distances.

Its about checking for collisions and less to do with finding the distance. The ObstacleCritic has 2 components: the "critical" collision rejection of samples that collide and the "behavioral" element which scores higher cost, but valid, areas higher to generally incentivize actions that aren't getting into unnecessary high cost areas where there are other options. The collision check is to find the highest cost on the footprint to checking if its valid - then we can use that same information to find then the closest distance to score. 

We actually have a different critic now the `CostCritic` which is now default that doesn't use distances but rather fully based on costs. It falls into the same trap as the `ObstacleCritic` since the `253` cost is still set for all inscribed values, but it does have some nice behavioral advantages for some - if not most - situations (notably: can tune obstacle behavior relative to the inflation exponential function rather than linear on distance).  

> Does it make sense to try to calculate the actual distances efficiently?

Perhaps, yes. I suppose for the local costmap given its smaller that creating a TSDF or similar wouldn't be too bad. It was just something I was trying very hard to avoid for computational reasons, but maybe I was overblowing it in my mind. Its definitely worth an experiment. Moreover, if we had a generic TDSF utility for costmaps in Nav2, I'm sure it would find utility for planning, control, collision checking, etc even if I can't put my finger on exactly how I'd use it off the top of my head.

mangoschorle:
[...]

PR #4996, opened 2025-03-19
Add near collision cost and warnings for misaligned parameter settings in MPPI critics

mini-1235:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4361  |
| Robotic platform tested on | tb3 |

---

Fix dead code as discussed in the issue

None

Tested on sim, cost of 253 works

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -26,6 +26,7 @@ void CostCritic::initialize()
   getParam(power_, "cost_power", 1);
   getParam(weight_, "cost_weight", 3.81f);
   getParam(critical_cost_, "critical_cost", 300.0f);
+  getParam(near_collision_cost_, "near_collision_cost", 253.0f);
   getParam(collision_cost_, "collision_cost", 1000000.0f);
   getParam(near_goal_distance_, "near_goal_distance", 0.5f);
   getParam(inflation_layer_name_, "inflation_layer_name", std::string(""));
@@ -52,6 +53,14 @@ void CostCritic::initialize()
       " the inflation radius to be at MINIMUM half of the robot's largest cross-section. See "
       "github.com/ros-planning/navigation2/tree/main/nav2_smac_planner#potential-fields"
       " for full instructions. This will substantially impact run-time performance.");
+  }
+
+  if(costmap_ros_->getUseRadius()==consider_footprint_){
+    RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
+  }
+
+  if(near_collision_cost_ > 253.0f){
+    RCLCPP_WARN(logger_, "Near collision cost is set higher than INSCRIBED_INFLATED_OBSTACLE");
   }
 
   RCLCPP_INFO(
@@ -153,28 +162,24 @@ void CostCritic::score(CriticData & data)
       // The footprintCostAtPose will always return "INSCRIBED" if footprint is over it
       // So the center point has more information than the footprint
       if (!worldToMapFloat(Tx, Ty, x_i, y_i)) {
-        if (!is_tracking_unknown_) {
-          traj_cost = collision_cost_;
-          trajectory_collide = true;
-          break;
-        }
         pose_cost = 255.0f;  // NO_INFORMATION in float
       } else {
         pose_cost = static_cast<float>(costmap->getCost(getIndex(x_i, y_i)));
         if (pose_cost < 1.0f) {
           continue;  // In free space
         }
-        if (inCollision(pose_cost, Tx, Ty, traj_yaw(i, j))) {
-          traj_cost = collision_cost_;
-          trajectory_collide = true;
-          break;
-        }
+      }
+
+      if (inCollision(pose_cost, Tx, Ty, traj_yaw(i, j))) {
+        traj_cost = collision_cost_;
+        trajectory_collide = true;
+        break;
       }
 
       // Let near-collision trajectory points be punished severely
       // Note that we collision check based on the footprint actual,
       // but score based on the center-point cost regardless
-      if (pose_cost >= 253.0f /*INSCRIBED_INFLATED_OBSTACLE in float*/) {
+      if (pose_cost >= near_collision_cost_) {
         traj_cost += critical_cost_;
       } else if (!near_goal) {  // Generally prefer trajectories further from obstacles
         traj_cost += pose_cost;
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -15,6 +15,7 @@
 
 #include <cmath>
 #include "nav2_mppi_controller/critics/cost_critic.hpp"
+#include "nav2_core/controller_exceptions.hpp"
 
 namespace mppi::critics
 {
@@ -61,7 +62,7 @@ void CostCritic::initialize()
     "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
     "in both the costmap and the cost critic.");
     if (costmap_ros_->getUseRadius()) {
-      throw std::invalid_argument(
+      throw nav2_core::ControllerException(
       "Considering footprint in collision checking but no robot footprint provided in the "
       "costmap.");
     }
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -55,11 +55,11 @@ void CostCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
-  if(costmap_ros_->getUseRadius()==consider_footprint_){
+  if(costmap_ros_->getUseRadius() == consider_footprint_) {
     RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
   }
 
-  if(near_collision_cost_ > 253.0f){
+  if(near_collision_cost_ > 253.0f) {
     RCLCPP_WARN(logger_, "Near collision cost is set higher than INSCRIBED_INFLATED_OBSTACLE");
   }
 
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -26,7 +26,7 @@ void CostCritic::initialize()
   getParam(power_, "cost_power", 1);
   getParam(weight_, "cost_weight", 3.81f);
   getParam(critical_cost_, "critical_cost", 300.0f);
-  getParam(near_collision_cost_, "near_collision_cost", 253.0f);
+  getParam(near_collision_cost_, "near_collision_cost", 253);
   getParam(collision_cost_, "collision_cost", 1000000.0f);
   getParam(near_goal_distance_, "near_goal_distance", 0.5f);
   getParam(inflation_layer_name_, "inflation_layer_name", std::string(""));
@@ -55,11 +55,19 @@ void CostCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
-  if(costmap_ros_->getUseRadius() == consider_footprint_) {
-    RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
+  if (costmap_ros_->getUseRadius() == consider_footprint_) {
+    RCLCPP_WARN(
+    logger_,
+    "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
+    "in both the costmap and the cost critic.");
+    if (costmap_ros_->getUseRadius()) {
+      throw std::invalid_argument(
+      "Considering footprint in collision checking but no robot footprint provided in the "
+      "costmap.");
+    }
   }
 
-  if(near_collision_cost_ > 253.0f) {
+  if(near_collision_cost_ > 253) {
     RCLCPP_WARN(logger_, "Near collision cost is set higher than INSCRIBED_INFLATED_OBSTACLE");
   }
 
@@ -179,7 +187,7 @@ void CostCritic::score(CriticData & data)
       // Let near-collision trajectory points be punished severely
       // Note that we collision check based on the footprint actual,
       // but score based on the center-point cost regardless
-      if (pose_cost >= near_collision_cost_) {
+      if (pose_cost >= static_cast<float>(near_collision_cost_)) {
         traj_cost += critical_cost_;
       } else if (!near_goal) {  // Generally prefer trajectories further from obstacles
         traj_cost += pose_cost;
--- a/nav2_mppi_controller/src/critics/obstacles_critic.cpp
+++ b/nav2_mppi_controller/src/critics/obstacles_critic.cpp
@@ -44,8 +44,16 @@ void ObstaclesCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
-  if(costmap_ros_->getUseRadius() == consider_footprint_) {
-    RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
+  if (costmap_ros_->getUseRadius() == consider_footprint_) {
+    RCLCPP_WARN(
+    logger_,
+    "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
+    "in both the costmap and the obstacle critic.");
+    if (costmap_ros_->getUseRadius()) {
+      throw std::invalid_argument(
+      "Considering footprint in collision checking but no robot footprint provided in the "
+      "costmap.");
+    }
   }
 
   RCLCPP_INFO(
--- a/nav2_mppi_controller/src/critics/obstacles_critic.cpp
+++ b/nav2_mppi_controller/src/critics/obstacles_critic.cpp
@@ -16,6 +16,8 @@
 #include <cmath>
 #include "nav2_mppi_controller/critics/obstacles_critic.hpp"
 #include "nav2_costmap_2d/inflation_layer.hpp"
+#include "nav2_core/controller_exceptions.hpp"
+
 namespace mppi::critics
 {
 
@@ -50,7 +52,7 @@ void ObstaclesCritic::initialize()
     "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
     "in both the costmap and the obstacle critic.");
     if (costmap_ros_->getUseRadius()) {
-      throw std::invalid_argument(
+      throw nav2_core::ControllerException(
       "Considering footprint in collision checking but no robot footprint provided in the "
       "costmap.");
     }
--- a/nav2_mppi_controller/src/critics/obstacles_critic.cpp
+++ b/nav2_mppi_controller/src/critics/obstacles_critic.cpp
@@ -44,7 +44,7 @@ void ObstaclesCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
-  if(costmap_ros_->getUseRadius()==consider_footprint_){
+  if(costmap_ros_->getUseRadius() == consider_footprint_) {
     RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
   }
 
--- a/nav2_mppi_controller/src/critics/obstacles_critic.cpp
+++ b/nav2_mppi_controller/src/critics/obstacles_critic.cpp
@@ -44,6 +44,10 @@ void ObstaclesCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
+  if(costmap_ros_->getUseRadius()==consider_footprint_){
+    RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
+  }
+
   RCLCPP_INFO(
     logger_,
     "ObstaclesCritic instantiated with %d power and %f / %f weights. "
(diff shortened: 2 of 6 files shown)


nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -95,6 +95,24 @@ float CostCritic::findCircumscribedCost(
   return circumscribed_cost_;
 }
 
+bool CostCritic::inCollision(float cost) const
```
SteveMacenski, line 98: This is not eq, I think the old `inCollision` method needs to be reverted. We need to check a full SE2 footprint if set. This is not eq:

```
    case (INSCRIBED_INFLATED_OBSTACLE):
      return consider_footprint_ ? false : true;
```

Because we never check the footprint itself. Also, that method was implemented in the .hpp file for a reason, the inlining there made the code perform faster

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -153,27 +171,36 @@ void CostCritic::score(CriticData & data)
       // The footprintCostAtPose will always return "INSCRIBED" if footprint is over it
       // So the center point has more information than the footprint
       if (!worldToMapFloat(Tx, Ty, x_i, y_i)) {
-        if (!is_tracking_unknown_) {
-          traj_cost = collision_cost_;
-          trajectory_collide = true;
-          break;
-        }
-        pose_cost = 255.0f;  // NO_INFORMATION in float
+        pose_cost = nav2_costmap_2d::NO_INFORMATION;
```
SteveMacenski, line 174: Revert; if we're not tracking unknown space, this is a collision case and we can skip scoring the rest of this trajectory

SteveMacenski (changes requested):
I'm not sure that this is equivalent or as performant as before, unfortunately!

nav2_mppi_controller/src/critics/cost_critic.cpp
mini-1235: In my new code, I evaluate the footprint cost here
https://github.com/mini-1235/navigation2/blob/ad4f1ecb994bf17243e7469003af2b91de706675/nav2_mppi_controller/src/critics/cost_critic.cpp#L183-L191
instead of doing it inside the function
This is because we will need this footprint cost when scoring(using point cost only will be dead code)
https://github.com/mini-1235/navigation2/blob/ad4f1ecb994bf17243e7469003af2b91de706675/nav2_mppi_controller/src/critics/cost_critic.cpp#L204-L207
This is how it is done in the obstacle critic as well (check for the point/footprint cost and then passed into the incollision function to check if it is in collision)
I could have return the footprint cost in the incollision function, but I think the current one would be better in terms of readability(?)

SteveMacenski: Ok - I think I need to look at this more carefully

SteveMacenski: Interestingly the [code coverage metrics show coverage in the supposedly dead code segment](https://app.codecov.io/gh/ros-navigation/navigation2/blob/main/nav2_mppi_controller%2Fsrc%2Fcritics%2Fcost_critic.cpp#L178). I suspect this happens when the[ `pose_cost` is `UNKNOWN` because its off the costmap](https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/src/critics/cost_critic.cpp#L155-L161) which is greater than `253`. I think this logic is on the face sensible since we're prefer moving in FREE or known-costed space over unknown, even when unknown / off costmap is technically valid for consideration.

However if that's the intent here (which to be clear: it wasn't, this entire block is a vestigial bit from a previous method for collision checking), then this should be written more clearly that this is for penalizing unknown space, not "near collision trajectory points" as the comment suggests. 

This shouldn't trigger in the case of `253` or `254`, since those should be invalid - but I'd like to have some validation that when the code we're talking about is triggered, its only in that `255` case (that we're not all mistaking something that really happens from a static analysis). 

I'm seeing though that in that `255` case, we skip the `inCollision` check, so it seems sensible to me that you move that out of the `else` block in this PR.

---

So I'm trying to think about what needs to change here w.r.t. the original code:
- The `inCollision` should be moved out of the `else` statement so that both cases of `worldToMapFloat` result in checking on the pose cost.
- Then, I agree that the `if (!is_tracking_unknown_)` can be removed from the `if` statement since the `inCollision` that's later run should handle that
- After validating that the `if (pose_cost >= 253.0f /*INSCRIBED_INFLATED_OBSTACLE in float*/) ` is only happening when pose_cost is `255`, that condition can be changed to checking if its `255` rather than `>= 253` to be clear. Maybe even remove this `critical_cost` thing altogether since it doesn't have the desired effect?

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -153,27 +171,36 @@ void CostCritic::score(CriticData & data)
...
+        // Skip footprint cost if point cost has already exceeded inscribed obstacle cost
+        if (pose_cost >= nav2_costmap_2d::INSCRIBED_INFLATED_OBSTACLE) {
           trajectory_collide = true;
           break;
         }
+        // Get footprint cost
+        if (consider_footprint_ &&
+          (pose_cost >= possible_collision_cost_ || possible_collision_cost_ < 1.0f))
+        {
+          pose_cost = static_cast<float>(collision_checker_.footprintCostAtPose(
```
SteveMacenski, line 187: We dont want to use the footprint cost for the cost to apply, only the collision checking process itself

nav2_mppi_controller/src/critics/cost_critic.cpp
mini-1235: @SteveMacenski  Thanks for looking into this PR! 

>  but I'd like to have some validation that when the code we're talking about is triggered, its only in that 255 case (that we're not all mistaking something that really happens from a static analysis).

Before submitting the PR, and again after your comment, I ran several experiments to better understand when this code segment is actually triggered. Let me share a little bit for my findings and my ideas behind this PR:)


> Interestingly the [code coverage metrics show coverage in the supposedly dead code segment](https://app.codecov.io/gh/ros-navigation/navigation2/blob/main/nav2_mppi_controller%2Fsrc%2Fcritics%2Fcost_critic.cpp#L178). I suspect this happens when the[ pose_cost is UNKNOWN because its off the costmap](https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/src/critics/cost_critic.cpp#L155-L161) which is greater than 253.

I agree that this is not a 100% dead code, and maybe it is more suitable to call it as a logic error here, sorry for the confusion. 
Based on my observations, the code segment would be triggered in three cases:
1. As you mentioned: "when the pose_cost is UNKNOWN because its off the costmap", then the pose cost is 255, >= 253
2. When we set consider footprint as true in CostCritic, but using robot radius in the costmap instead of footprint(default nav2 params), then 
https://github.com/ros-navigation/navigation2/blob/a904802fe222f961d4a211fa0c813349672c18eb/nav2_mppi_controller/include/nav2_mppi_controller/critics/cost_critic.hpp#L71-L78
point cose of 253 would not be considered as in collision, therefore triggering the code segment
3. PointCost returns a 253, and footprint cost also returns a 253, this is very rare, and I didn't notice this when submitting the PR(sorry for that)

Before proceeding further with this PR, there are two points I would really appreciate your insight on:

1.  > We dont want to use the footprint cost for the cost to apply, only the collision checking process itself
Can you elaborate more on this? From the documentation side, I see that critical cost is described as "Cost to apply to a pose with any point in inflated space to prefer distance from obstacles.", That description made it seem reasonable to me that we should consider footprint cost during scoring.. Also, using the footprint cost of `253` helps penalizes the near collision trajectory points, as the comment suggested.
2. Does point cost of `253` result in footprint cost of `254`? 
My setup:
```
footprint: "[ [0.21, 0.195], [0.21, -0.195], [-0.21, -0.195], [-0.21, 0.195] ]"
```
I ran this in a tb3 sim environment, and from the log, I saw when pointcost returns a `253`, the footprint cost(aka linecost) returns:
[component_container_isolated-8] line cost 158
[component_container_isolated-8] line cost 220
[component_container_isolated-8] line cost 253
[component_container_isolated-8] line cost 251
I am a bit confused by this, and this is the third condition I mentioned above(I have always thought that if pointcost returns `253` then footprint cost is certainly `254`)

If those questions are just my misunderstandings, I agree to close this PR:)

mini-1235:
Moving https://github.com/ros-navigation/navigation2/pull/4996#discussion_r2012831752

@SteveMacenski  Thanks for looking into this PR! 

>  but I'd like to have some validation that when the code we're talking about is triggered, its only in that 255 case (that we're not all mistaking something that really happens from a static analysis).

Before submitting the PR, and again after your comment, I ran several experiments to better understand when this code segment is actually triggered. Let me share a little bit for my findings and my ideas behind this PR:)


> Interestingly the [code coverage metrics show coverage in the supposedly dead code segment](https://app.codecov.io/gh/ros-navigation/navigation2/blob/main/nav2_mppi_controller%2Fsrc%2Fcritics%2Fcost_critic.cpp#L178). I suspect this happens when the[ pose_cost is UNKNOWN because its off the costmap](https://github.com/ros-navigation/navigation2/blob/main/nav2_mppi_controller/src/critics/cost_critic.cpp#L155-L161) which is greater than 253.

I agree that this is not a 100% dead code, and maybe it is more suitable to call it as a logic error here, sorry for the confusion. 
Based on my observations, the code segment would be triggered in three cases:
1. As you mentioned: "when the pose_cost is UNKNOWN because its off the costmap", then the pose cost is 255, >= 253
2. When we set consider footprint as true in CostCritic, but using robot radius in the costmap instead of footprint(default nav2 params), then 
https://github.com/ros-navigation/navigation2/blob/a904802fe222f961d4a211fa0c813349672c18eb/nav2_mppi_controller/include/nav2_mppi_controller/critics/cost_critic.hpp#L71-L78
253 would not be considered as in collision, therefore triggering the code segment
3. Point cost returns a 253, and footprint cost also returns a 253, this is very rare, and I didn't notice this when submitting the PR(sorry for that)

Before proceeding further with this PR, there are two points I would really appreciate your insight on:

1.  > We dont want to use the footprint cost for the cost to apply, only the collision checking process itself

Can you elaborate more on this? From the documentation side, I see that critical cost is described as "Cost to apply to a pose with any point in inflated space to prefer distance from obstacles.", That description made it seem reasonable to me that we should consider footprint cost during scoring. Also, using the footprint cost of `253` helps penalizes the near collision trajectory points, as the comment suggested.

2. Does point cost of `253` always result in footprint cost of `254`? 
My setup:
```
footprint: "[ [0.21, 0.195], [0.21, -0.195], [-0.21, -0.195], [-0.21, 0.195] ]"
```
I ran this in a tb3 sim environment, and from the log, I saw when pointcost returns a `253`, the footprint cost(aka max linecost) returns:
```
[component_container_isolated-8] line cost 158
[component_container_isolated-8] line cost 220
[component_container_isolated-8] line cost 253
[component_container_isolated-8] line cost 251
```
I am a bit confused by this, and this is the third condition I mentioned above(I have always thought that if point cost returns `253` then footprint cost is certainly `254`, resulting in colision)

If those questions are just my misunderstandings, I agree to close this PR:)

SteveMacenski:
> When we set consider footprint as true in CostCritic, but using robot radius in the costmap instead of footprint(default nav2 params),  253 would not be considered as in collision, therefore triggering the code segment

Oh interesting. I think we should probably log a warning in this case, this should be loudly mentioned to make sure folks know that is the case that the robot information in the local costmap and settings in `consider_footprint` are not aligned (in both cases `if (costmap_ros_->getUseRadius() == consider_footprint) {// Problem!!}`. Probably for the Obstacle critic too. 

>   We dont want to use the footprint cost for the cost to apply, only the collision checking process itself

Basically if we include the footprint cost, anywhere that the robot base has any point in the inflated 253 space, its always scored as 253 (max valid when using footprints). So if just one point on the robot is in that space, its the same as the robot driving up against a wall. Instead, if we score based on the center point-costs, then there's a penalization of proximity of the robot against the wall all the way until its basically touching. Also, that penalty is then directly scaling by the inflation layer's exponential function users tune - which could raise dramatically near obstacles or more gradually over space (so its tunable within that close-to-obstacles regime). 

This was a big reason that the Cost Critic was added and replaces the Obstacles Critic -- that used the distance which succomes to that exact problem always since the distance calculated from cost will always have the distance as the same when the footprint is touching 253 space. 

Maybe the docs need updating / clarification. Does that make sense? 


> Does point cost of 253 always result in footprint cost of 254?

Generally, yes, except for discretization situations where you're just barely touching the corner of a costmap cell that is 253 for the center cost and the ray-marching on the footprint's resolution is coarse enough to miss the bit that might in 254. Generally speaking, yes, if its "in" the inflated 253 space, the footprint should be 254. The exception to that is if the center point is just barely skirting the edge of the 253 inflated space, it can sometimes also return 253. 

if you run that again and have the robot squarely in the 253 inflated space, you should always see 254.

mini-1235:
> The exception to that is if the center point is just barely skirting the edge of the 253 inflated space, it can sometimes also return 253.

Should we describe this kind of situation as near collision or in collision?

SteveMacenski:
Basically the 2 edge situations to think about are:
- The robot's pose is chipping the corner of the inflation radius past actual collision by some fractional amount of the cell resolution -- so it says there's a collision, but technically its not in collision by some very small amount. A center-check would result in a collision but if you actually checked the polygon it may not be (but **barely**) 
- When you're near an obstacle  and do a polygon check, it could conceivable for an irregular shaped obstacle, or an obstacle misaligned with the costmap's major X and Y axes, there are part of the polygon that collide with just a tiny amount of an occupied cell and says its not in collision (but still 253). This is because we use an iterator to march over the line segment at a resolution and that resolution might miss that tiny amount. We could reduce the resolution to be infinitesimally small, but that comes then at an infinite computational cost. So a trade off has to be made for reason, which I believe it checking every 0.5 costmap cell resolution (so 2x checks per cell, on average). 

Then the obvious situations where the robot is in the collision by center cost and actually is; and robot's footprint collides with obstacle and we detect it. 

The edge cases are created by the fact that we're representing continuous space by a binning quantization (occupancy grid square cells), which both (a) obstacles can be at an arbitrary angle with respect to and (b) the footprint of the robot can be at any angle with respect to. There's some aliasing as an artifact of that quantization resolution choice.  In actuality for the first case, the  center point cost is the one that is wrong, not the polygonal check, since the far corner of the square cell is outside of the actual inflation radius (the radius is computed by cell center distances). In the second case, its up in the air about whether or not that could be considered slightly not a collision or slightly in collision, but either way they would be scored as highly costed or invalid, respectively. Its hard to say whether (a) the obstacle is misrepresented by the cell's aliasing or bininng of sensor data or (b) the iterator marching resolution barely misses detecting the colliding cell. 

Sorry, I know that was alot. But tl;dr is I think we should treat it as a near collision, since its likely just an artifact of square-cell aliasing angles and/or binning quantization

mini-1235:
> Sorry, I know that was alot. But tl;dr is I think we should treat it as a near collision, since its likely just an artifact of square-cell aliasing angles and/or binning quantization

No worries at all, I actually love hearing this kind of breakdown. Super insightful, and that really clarifies my questions. I think there are two things to be done based on #4361 and your explanations in this PR.

1. Calculate the actual distance (this is not discussed in this PR and we might need to implement other costmap layers)

2. Regarding this code segment
https://github.com/ros-navigation/navigation2/blob/b3ca86ed9dde20f6c699259c710c570fb58388f9/nav2_mppi_controller/src/critics/cost_critic.cpp#L177-L179
This is actually not dead code, it will trigger on 1. 253(point)+253(footprint)+consider footprint 2. 255 + is tracking unknown case, so we can actually keep this, unless you prefer to make it as a tunable option as you suggested earlier 

> have that check be changed to be useful by parameterizing the cost, above which the critical_cost is applied (which defaults to something like 254 to be disabled) as a tunable option for users to set a max proximity before which it should avoid even if non-collision inducing

However, we could/should do these things to cost critic:

- Warn users when settings are not aligned
> I think we should probably log a warning in this case, this should be loudly mentioned to make sure folks know that is the case that the robot information in the local costmap and settings in consider_footprint are not aligned (in both cases if (costmap_ros_->getUseRadius() == consider_footprint) {// Problem!!}. Probably for the Obstacle critic too.
- Move the `inCollision` outside the `else` block for better readability
>The inCollision should be moved out of the else statement so that both cases of worldToMapFloat result in checking on the pose cost.

What are your thoughts on these? I think I can close this PR and open a new one for the second point?

SteveMacenski:
>  This is actually not dead code, it will trigger on 1. 253(point)+253(footprint)+consider footprint 2. 255 + is tracking unknown case, so we can actually keep this, unless you prefer to make it as a tunable option as you suggested earlier

I think keep it and make it tunable with 253 as the default would be good! I don't think that would cause any issues with your assessment of the (3) cases for 253 being triggered to possibly lower it. I would add in a warning log if its set `> 253`

> Warn users when settings are not aligned

Agreed! :+1: 

>     Move the inCollision outside the else block for better readability

Agreed! :+1: 

A new PR probably makes sense!

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -26,6 +26,7 @@ void CostCritic::initialize()
   getParam(power_, "cost_power", 1);
   getParam(weight_, "cost_weight", 3.81f);
   getParam(critical_cost_, "critical_cost", 300.0f);
+  getParam(near_collision_cost_, "near_collision_cost", 253.0f);
```
SteveMacenski, line 29: Docs update with this would be good (+ migration guide for the new parameter entry)

SteveMacenski: Shouldn't this be an `int` though, not a double (as well in the param files)? This is a costmap cost

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -54,6 +55,14 @@ void CostCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
+  if(costmap_ros_->getUseRadius() == consider_footprint_) {
+    RCLCPP_WARN(logger_, "Considering footprint but robot radius set in costmap");
```
SteveMacenski, line 59: The inverted could be true (not using footprint but footprint set in costmap). I'd reword this to be that there is a misalignment of parameterization of the collision checking. 

If this condition is true **and** `!costmap_ros_->getUseRadius()` I think the log here was an `throw` is good since that is a true error case that you're asking for footprint but no footprint is given. 

Ditto for the obstacles critic

SteveMacenski (changes requested):
LGTM! Just sanity check performance for me? If you add some print logs here when the critic is the cost critic, does the performance meaningfully change? 

https://github.com/ros-navigation/navigation2/blob/be05299bcc9a4b40c8418f65b75ab33c1878cd3f/nav2_mppi_controller/src/critic_manager.cpp#L74

nav2_mppi_controller/src/critics/cost_critic.cpp
mini-1235: > If this condition is true and !costmap_ros_->getUseRadius() I think the log here was an throw is good since that is a true error case that you're asking for footprint but no footprint is given.

I think this should be `costmap_ros_->getUseRadius()`(?)
`costmap_ros_->getUseRadius()`: true, `consider_footprint:`true -> asking for footprint but no footprint is given
`costmap_ros_->getUseRadius()`: false, `consider_footprint:`false -> footprint set for costmap, but not asking for footprint so this is ok

mini-1235:
> If you add some print logs here when the critic is the cost critic, does the performance meaningfully change?

I have tested with default params, no performance change

Link to docs: https://github.com/ros-navigation/docs.nav2.org/pull/665

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -54,6 +55,22 @@ void CostCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
+  if (costmap_ros_->getUseRadius() == consider_footprint_) {
+    RCLCPP_WARN(
+    logger_,
+    "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
+    "in both the costmap and the cost critic.");
+    if (costmap_ros_->getUseRadius()) {
+      throw std::invalid_argument(
```
SteveMacenski, line 64: `nav2_core::ControllerException`

nav2_mppi_controller/src/critics/obstacles_critic.cpp
```diff
@@ -44,6 +44,18 @@ void ObstaclesCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
+  if (costmap_ros_->getUseRadius() == consider_footprint_) {
+    RCLCPP_WARN(
+    logger_,
+    "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
+    "in both the costmap and the obstacle critic.");
+    if (costmap_ros_->getUseRadius()) {
+      throw std::invalid_argument(
```
SteveMacenski, line 53: Ditto

SteveMacenski (changes requested):
Minor change and LGTM!

mini-1235:
Should we backport this PR (except for the near collision cost change) to jazzy/humble?

SteveMacenski (approved):

SteveMacenski:
Yeah its not currently setup for non-rolling builds at the moment

PR #4997, opened 2025-03-19
Allow main (rolling) branch to build with the Jazzy binaries distribution

SteveMacenski:


tonynajjar:
So will `main` be permanently maintained from now on to successfully build on all previous (non-EOL) distros or is this a one-time effort? I'm asking because I'm considering switching to the main branch for our Jazzy production robots (I still have my fork and therefore still have to manually bring new features in so I'm not worried about new semi-tested features creeping in)

SteveMacenski:
Good question - I recently worked on adding in a new workflow for building ROS Distros against Main to test compatibility in PRs (https://github.com/ros-navigation/navigation2/blob/main/.github/workflows/build_main_against_distros.yml). My intention is to support Jazzy until a point that it is 'unreasonable' to continue due to API breaking changes in Rolling that force such a major update that it is no longer possible and/or makes the code so unreadable due to the number of preprocessor conditions. 

I know that's a bit non-specific, but I'm open to a more formal policy or trip wire for when that would be dropped if you wanted something concrete. My intention is with the new CI job is that every PR will tell us if it breaks compatibility. If we do, have it fixed at that time to always be compatible. That would continue up until the point that the fixes required to keep it humming are so intrusive or requiring architectural updates to make possible. 

I will plan to add a Kilted job once that's released at the end of the week as well, but that will be only active for the next year or so. 

Let me know what you think. If we think a policy is important, I'd be happy to open up a google doc with you and try to find the balance and have that published on the Nav2 websits.

tonynajjar:
> My intention is to support Jazzy until a point that it is 'unreasonable' to continue due to API breaking changes in Rolling that force such a major update that it is no longer possible and/or makes the code so unreadable due to the number of preprocessor conditions.

I think that's reasonable as a first attempt. From what I've seen and by pure gut feeling, I think it's quite likely for the non-LTS ROS version to reach EOL before we get to that stage where it's too clumsy too maintain; I don't see so many breaking changes.

I would personally leave off writing a policy to when we hit some issues that would better inform what that policy should be.

Issue #4865, opened 2025-01-18
[Graceful Controller] Additional Controller Techniques - Experimentation


SteveMacenski:
Right now the Graceful Controller computes a smooth spiral curve from the control law from the robot's current state to a pose on the path to track. The path pose is based on a lookahead distance that is shortened up if there is a collision in the trajectory's spiral curve on the way. Getting out some ideas / thoughts so we can collaborate on this as a community instead of staying in some notes document 

There are quite a few other ways, I think that this control law could be utilized that are worth experimenting with for improvements or different behaviors:
- [ ] Interpolate the path point for the distance requested by the lookahead, to smooth out the trajectories from frame to frame. We do this in the RPP controller, so this should be a relatively straight forward improvement that improves the controller as-is
- [ ] Consider committing to a given trajectory for a period of time before computing a new spiral. Each iteration, we compute a new spiral and perhaps its worth committing to the current spiral for a couple of seconds (except if collision or large change in cost) so we commit to actions more fully, before then computing a new curve
- [ ] Consider sampling a set of spiral trajectories with different spiral parameters instead of them as static parameters. Sample within some range of values and score them for the best trajectory
- [ ] Consider the above, with relaxation on the exact goal pose so we compute (collision valid) spirals towards or along the path, but no so exact. Further, we could sample based on some time duration into the future from the robot's starting pose + spiral parameters. Part of the scoring could be how close / near the path or it is (so not even considering the path's poses in the spiral constraints) 
- [ ] Perhaps rather than randomly sampling (like RRT) or sampling in fixed intervals (like DWA) of the the spiral parameters, attempt to use convex optimization on some cost functions to select the best trajectory and goal [Unconstrained Model Predictive Control for Robot Navigation under
Uncertainty](https://assets.amazon.science/61/6a/49cbec1749e9a41410593433295d/unconstrained-model-predictive-control-for-robot-navigation-under-uncertainty.pdf) [Robot Navigation with MPEPC in Dynamic and Uncertain
[...]

PR #5003, opened 2025-03-22
Calculate next path point using lookahead for Graceful Controller

SakshayMahna:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4865  |

---

- Added default parameter files for Regulated Pure Pursuit Controller and Graceful Controller
- Added 2 new parameters for Graceful Controller (`lookahead_resolution` and `interpolate_after_goal`)
- Added lookahead logic to calculate next path point for Graceful Controller

**Additional Information**

- Graceful Controller now calculates the next point on path using lookahead logic, it starts from a maximum lookahead value and searches for a path using lookahead distance in intervals of `lookahead_resolution`.
- The controller also calculates interpolated point after goal. This can be switched on and off using the parameter `interpolate_after_goal`.

- Added new parameter, so need to add that to default configs and documentation page.

- Same unit tests and tested in simulation.

---

- The interpolation after goal logic seems a bit problematic and may need refinement.

#### For Maintainers: 


nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -218,9 +216,6 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
         double yaw = std::atan2(target_pose.pose.position.y, target_pose.pose.position.x);
         target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(yaw);
       }
-    } else if (dist_to_target < params_->min_lookahead) {
```
SteveMacenski, line 221: Why removed?

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,21 +195,19 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
     //  * Be as far away as possible from the robot (for smoothness)
     //  * But no further than the max_lookahed_ distance
     //  * Be feasible to reach in a collision free manner
-    geometry_msgs::msg::PoseStamped target_pose = transformed_plan.poses[i];
-    double dist_to_target = target_distances[i];
+    geometry_msgs::msg::PoseStamped target_pose = getLookAheadPoint(
+      lookahead, transformed_plan, params_->interpolate_after_goal);
 
-    // Continue if target_pose is too far away from robot
-    if (dist_to_target > params_->max_lookahead) {continue;}
```
SteveMacenski, line 212: Why remove?

SteveMacenski (commented):
Thinking aloud a bit:

This bit of code looks like we iterate through the path, try to find the `max_lookahead` distance away and back up closer and closer to the robot until we find a solution that is collision free.

Is that case, we should attempt to find the interpolated lookahead point for the `max_lookahead` away to start off with. After that point, its a bit arbitrary, isn't it? Selecting a `lookahead_resolution` versus marching through the path points I don't think improves the smoothness, does it?

For example, the resolution you set was 0.1, but a normal path from a planning plugin probably does it closer to 0.05 for the costmap resolution when that resolution is 5cm. So wouldn't this case actually be more coarse?

nav2_graceful_controller/include/nav2_graceful_controller/graceful_controller.hpp
```diff
@@ -107,6 +107,32 @@ class GracefulController : public nav2_core::Controller
...
+  /**
+   * @brief Find the intersection a circle and a line segment.
+   * This assumes the circle is centered at the origin.
+   * If no intersection is found, a floating point error will occur.
+   * @param p1 first endpoint of line segment
+   * @param p2 second endpoint of line segment
+   * @param r radius of circle
+   * @return point of intersection
+   */
+  static geometry_msgs::msg::Point circleSegmentIntersection(
```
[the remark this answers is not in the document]
SakshayMahna: There are 2 functions that I took directly from RPP, `circleSegmentIntersection` and `getLookaheadPoint`. Moving to a shared directory would be good, (along with their respective tests).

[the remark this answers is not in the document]
SteveMacenski: nav2_utils seems fine. Make a `controller_utils` or `interpolation_utils` file, or something similar if you have a better idea

SteveMacenski:
> That's why thought of putting lookahead_resolution to adjust what may work best for the user.

Do you see an improvement using the `lookahead_resolution` instead of marching through the path as originally done in the back up procedure? 

By default you have it set to 0.1, but the path resolution is ~0.05. If you lower that to 0.05 or even lower like 0.02, is it smoother than before?

If so, I think that's a neat feature and we can keep it, users can tune to their liking and we can set it default to the costmap resolution.

If not, I think maybe we should revert that bit and only use the interpolation when using the full lookahead distance. Else, just use the path points since there's a jump anyway.

SakshayMahna:
No, I don't see any improvement while using a smaller lookahead resolution. 

The plots for the 3 cases (smaller resolution, larger resolution, using points) all look similar.
I'll just use the path points then.

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,21 +195,19 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
     // Else, fall through and see if we should follow control law longer
   }
 
-  // Precompute distance to candidate poses
-  std::vector<double> target_distances;
-  computeDistanceAlongPath(transformed_plan.poses, target_distances);
```
mikeferguson, line 200: I believe this will cause a minor regression - the getLookaheadPoint() function simply looks at the distance between the target point and the robot - not the distance ALONG the path. In some instances, this can lead to the controller massively shortcutting the desired path.

I would suggest making getLookaheadPoint use distance along path instead of hypot().

SakshayMahna: Right!
 I can make this change in this PR. 
Or another issue/PR would be appropriate for this?

nav2_util/include/nav2_util/interpolation_utils.hpp
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2024 Sakshay Mahna
```
SteveMacenski, line 1: Ditto

nav2_util/include/nav2_util/interpolation_utils.hpp
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2020 Shrijit Singh
```
SteveMacenski, line 1: Maybe `controller_utils.hpp` and `cpp`?

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: @SakshayMahna I believe this item is resolved, correct? @mikeferguson can you take another look?

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,56 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
+      // Underlying control law needs a single target pose, which should:
+      //  * Be as far away as possible from the robot (for smoothness)
+      //  * But no further than the max_lookahed_ distance
+      //  * Be feasible to reach in a collision free manner
+      target_pose = transformed_plan.poses[i];
+      dist_to_target = target_distances[i];
+
+      if(validateTargetPose(target_pose, dist_to_target, dist_to_goal,
+                            local_plan, costmap_transform, cmd_vel))
+      {
```
SteveMacenski, line 227:
```suggestion
      if (validateTargetPose(target_pose, dist_to_target, dist_to_goal,
        local_plan, costmap_transform, cmd_vel))
      {
```

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,56 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
+  bool valid_target_pose_found = false;
+  nav_msgs::msg::Path local_plan;
+
+  // Calculate target pose through lookahead interpolation
+  double dist_to_target = params_->max_lookahead;
+  geometry_msgs::msg::PoseStamped target_pose = nav2_util::getLookAheadPoint(
+    dist_to_target, transformed_plan, params_->interpolate_after_goal);
+
+  if (!validateTargetPose(target_pose, dist_to_target, dist_to_goal,
+                          local_plan, costmap_transform, cmd_vel))
```
SteveMacenski, line 209:
```suggestion
  if (!validateTargetPose(target_pose, dist_to_target, dist_to_goal,
    local_plan, costmap_transform, cmd_vel))
```

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,56 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
-      if (params_->prefer_final_rotation) {
-        // Avoid instability and big sweeping turns at the end of paths by
-        // ignoring final heading
-        double yaw = std::atan2(target_pose.pose.position.y, target_pose.pose.position.x);
-        target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(yaw);
+  // Find a valid target pose and its trajectory
+  bool valid_target_pose_found = false;
+  nav_msgs::msg::Path local_plan;
+
+  // Calculate target pose through lookahead interpolation
```
SteveMacenski, line 203:
```suggestion
  // Calculate target pose through lookahead interpolation to get most accurate lookahead point, if possible
```

SteveMacenski (changes requested):
There's one superficial refactor that would in my opinion make this easier to read / understand:

```
  // Calculate target pose through lookahead interpolation
  double dist_to_target = params_->max_lookahead;
  geometry_msgs::msg::PoseStamped target_pose = nav2_util::getLookAheadPoint(
    dist_to_target, transformed_plan, params_->interpolate_after_goal);
  bool valid_target_pose_found = validateTargetPose(target_pose, dist_to_target, dist_to_goal, local_plan, costmap_transform, cmd_vel);

  if (!valid_target_pose_found) {
    ...
    ...

    valid_target_pose_found = validateTargetPose(target_pose, dist_to_target, dist_to_goal, local_plan, costmap_transform, cmd_vel);
    if (valid_target_pose_found) {
      break;
    }
  }
```

This would remove the `else` statement and explicitly setting the valid_target_pose_found instead of just capturing the validate method's return.

SteveMacenski:
@SakshayMahna can you work on these last couple of updates?

SakshayMahna:
Hello Steve,
Yes, missed the last few notifications and emails. Working on it.

SakshayMahna:
Hi @SteveMacenski 
I did make the requested changes. However, I am not able to think of a way to get the code coverage to target (90.72%), I am stuck at 87.64%.

What I understand it is because of the fallback code that was added, that in case Lookahead Point is not found, fallback to default behaviour of checking the poses one by one.

The only way I think this can be tested is by adding an obstacle in the cost map. That way, the Lookahead Point will become invalid because of collision check by `simulate_trajectory`, and the execution would go to fallback code. But still I have doubts on how to make this work, because there is a set number of max iterations `simulate_trajectory` works for.

What do you think about this? Any ideas to improve the code coverage?

nav2_graceful_controller/src/parameter_handler.cpp
```diff
@@ -41,6 +41,8 @@ ParameterHandler::ParameterHandler(
     node, plugin_name_ + ".min_lookahead", rclcpp::ParameterValue(0.25));
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".max_lookahead", rclcpp::ParameterValue(1.0));
+  declare_parameter_if_not_declared(
```
SteveMacenski: I believe you said that this made it worse than the original - shouldn't we just remove this parameter and its logic if its not useful?

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    if (!pose_found) {
+      goal_pose_it = poses.end();
+    }
+  }
+
+  // If the no pose is not far enough, take the last pose
+  if (goal_pose_it == transformed_plan.poses.end()) {
+    if (interpolate_after_goal) {
```
SteveMacenski, line 87: If we set `interpolate_after_goal` always to false if we remove the parameter for graceful, then we can remove the `interpolated_orientation` bits.

Where we call this function in Graceful, just make sure to add a comment that `interpolate_after_goal` must be false since we don't implement the orientation bits in the interpolation

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,161 @@
...
+  const nav_msgs::msg::Path & transformed_plan,
+  const bool interpolate_after_goal)
+{
+  // Find the first pose which is at a distance greater than the lookahead distance
+  // Using distance along the path
+  const auto & poses = transformed_plan.poses;
+  auto goal_pose_it = poses.begin();
+  double d = 0.0;
+
+  if (!poses.empty()) {
```
SteveMacenski: Is this check necessary? If its empty, then `i < poses.size()` would fail and we wouldn't execute the first loop iteration

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,161 @@
...
+    bool pose_found = false;
+    for (size_t i = 1; i < poses.size(); i++) {
+      const auto & prev_pose = poses[i - 1].pose.position;
+      const auto & curr_pose = poses[i].pose.position;
+
+      d += std::hypot(curr_pose.x - prev_pose.x, curr_pose.y - prev_pose.y);
+      if (d >= lookahead_dist) {
+        goal_pose_it = poses.begin() + i;
```
SteveMacenski: Simplify: I think you can just set `goal_pose_it = poses.begin() + i;` each iteration and when `d >= lookahead_dist` only do the `break`. That would let you remove the `pose_found` logic and the later ` goal_pose_it = poses.end();`.

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,57 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
-
-    // Flip the orientation of the motion target if the robot is moving backwards
-    bool reversing = false;
-    if (params_->allow_backward && target_pose.pose.position.x < 0.0) {
-      reversing = true;
-      target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
-        tf2::getYaw(target_pose.pose.orientation) + M_PI);
     }
+  } else {
+    valid_target_pose_found = true;
```
SteveMacenski: Is this `else` actually needed? I think not.

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,57 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
+      // Underlying control law needs a single target pose, which should:
+      //  * Be as far away as possible from the robot (for smoothness)
+      //  * But no further than the max_lookahed_ distance
+      //  * Be feasible to reach in a collision free manner
+      target_pose = transformed_plan.poses[i];
+      dist_to_target = target_distances[i];
+
+      valid_target_pose_found = validateTargetPose(
+        target_pose, dist_to_target, dist_to_goal, local_plan, costmap_transform, cmd_vel);
+      if(valid_target_pose_found) {
```
SteveMacenski:
```suggestion
      if (valid_target_pose_found) {
```

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,57 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
-        target_pose, params_->slowdown_radius);
-      slowdown_pub_->publish(slowdown_marker);
-      // Publish the local plan
-      local_plan.header = transformed_plan.header;
-      local_plan_pub_->publish(local_plan);
-      // Successfully found velocity command
-      return cmd_vel;
-    }
+  // Compute velocity at this moment if valid target pose is found
+  if (valid_target_pose_found) {
```
SteveMacenski: I think this can be simplified.  If right after the `for (int i = transformed_plan.poses.size() - 1; i >= 0; --i)` loop exits we move the `throw nav2_core::NoValidControl("Collision detected in trajectory");` there (because it would have `break`ed if we found something in the loop), we wouldn't hit the publish conditions if the original check didn't pass or something in the loop didn't pass. 

I think that may also allow you to remove `valid_target_pose_found` variable as well since you can just check in the `if` statements the value of `validateTargetPose` without later needing to recall it.

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,58 +196,57 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
+  bool valid_target_pose_found = false;
+  nav_msgs::msg::Path local_plan;
+
+  // Calculate target pose through lookahead interpolation to get most accurate
+  // lookahead point, if possible
+  double dist_to_target = params_->max_lookahead;
+  geometry_msgs::msg::PoseStamped target_pose = nav2_util::getLookAheadPoint(
+    dist_to_target, transformed_plan, params_->interpolate_after_goal);
+
+  valid_target_pose_found = validateTargetPose(
```
SteveMacenski: What if `getLookAheadPoint` returns `transformed_plan.poses.back()` because `!pose_found`? Then `dist_to_target` may not be `params_->max_lookahead` for the input

SakshayMahna: The loop in `getLookAheadPoint` iterates from each pose calculating distance along path. The loop for it only breaks when distance found is greater than `max_lookahead`. The only case, we get `!pose_found` is when the distance is not able to add up to `max_lookahead`. 

This is fine, because the logic inside `validateTargetPose` only rejects the point if it is at a distance greater than `max_lookahead` or less than `min_lookahead`. Otherwise, the code logic continues to run the same.

SakshayMahna:
Hi @SteveMacenski
I made a mistake while syncing the main branch. And managed to get all my commits deleted.
However, I had a backup, and was able to recover the relevant changes.
And I tested the changes running on simulation as well.

Sorry for the mess.

nav2_graceful_controller/src/parameter_handler.cpp
SteveMacenski: @SakshayMahna

nav2_util/src/controller_utils.cpp
SteveMacenski: @SakshayMahna

nav2_util/src/controller_utils.cpp
SteveMacenski: @SakshayMahna

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,46 +196,36 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
     }
 
-    // Flip the orientation of the motion target if the robot is moving backwards
-    bool reversing = false;
-    if (params_->allow_backward && target_pose.pose.position.x < 0.0) {
-      reversing = true;
-      target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
-        tf2::getYaw(target_pose.pose.orientation) + M_PI);
-    }
+    valid_target_pose_found = validateTargetPose(
```
SteveMacenski: `valid_target_pose_found` variable is not needed, you can simply call `validateTargetPose` in the `if()` condition.

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,46 +196,36 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
-      if (params_->prefer_final_rotation) {
-        // Avoid instability and big sweeping turns at the end of paths by
-        // ignoring final heading
-        double yaw = std::atan2(target_pose.pose.position.y, target_pose.pose.position.x);
-        target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(yaw);
-      }
-    } else if (dist_to_target < params_->min_lookahead) {
-      // Make sure target is far enough away to avoid instability
-      break;
+  for (int i = transformed_plan.poses.size(); i >= 0; --i) {
```
SteveMacenski: If you make this an `unsigned int` instead of an `int`, then you don't need the `static_cast` below and nothing else needs to change

SteveMacenski (changes requested):
You have a few open comments above, but overall doesn't this look simpler and cleaner! Great work :smile:

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: I'm saying that you ahve this block of code:

```

      dist_to_target = params_->max_lookahead;
      target_pose = nav2_util::getLookAheadPoint(
        dist_to_target, transformed_plan, params_->interpolate_after_goal);
```

If the `target_pose` is not exactly `max_lookahead` away because `!pose_found`, then the inputs to `validateTargetPose` aren't consistent. The `target_pose` is not actually `dist_to_target` away. `dist_to_target` is not updated based on the actual return of `getLookAheadPoint`. `dist_to_target == params_->max_lookahead` always in this case even when `target_pose` is the last pose instead. 

I think something like the following is necessary: 

```
if (target_pose == transformed_plan.poses.back() {
  // update dist_to_target
}
```

SakshayMahna: Yes, understood now. It makes sense to change the lookahead distance if the path is clipped to end at the last pose.

nav2_graceful_controller/src/graceful_controller.cpp
SakshayMahna: On making it unsigned int, `i>=0` condition will always remain valid.
And I am getting a compiler error for the same.

nav2_util/src/controller_utils.cpp
SakshayMahna: I gave the suggested implementation quite some tries. But for the suggested one, the robot gets assigned a goal pose beyond the end point automatically (even with `interpolate_after_goal=false`). And the behaviour just doesn't feel right.

However, with the current implementation everything seems perfectly fine.

I don't understand the reason, but I'm sure I maybe missing something.

SteveMacenski:
@SakshayMahna please rebase on main to make CI turn over. There was a message filter API change that we resolved in `main`

nav2_graceful_controller/include/nav2_graceful_controller/parameter_handler.hpp
```diff
@@ -35,6 +35,7 @@ struct Parameters
   double transform_tolerance;
   double min_lookahead;
   double max_lookahead;
+  bool interpolate_after_goal;
```
SteveMacenski: Remove since now unused :-)

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: @SakshayMahna `unsigned int i = transformed_plan.poses.size(); i >= 0; --i` then `if (i == transformed_plan.poses.size())`

nav2_util/src/controller_utils.cpp
SteveMacenski: Can you make a branch off this one and send a link here? I'll take a look

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: @SakshayMahna OK! Let me know when its implemented :-)

SteveMacenski (changes requested):
Super close now! Take a look at the remaining open items from previous reviews :-) Thanks for the iterations - this is much improved!

nav2_graceful_controller/src/graceful_controller.cpp
SakshayMahna: I get this error:
```
error: comparison of unsigned expression in ‘>= 0’ is always true [-Werror=type-limits]
  207 |   for (unsigned int i = transformed_plan.poses.size(); i >= 0; --i) {
      |                                                        ~~^~~~
```

nav2_graceful_controller/src/graceful_controller.cpp
SakshayMahna: I have implemented this:
In the function `getLookAheadPoint` in `controller_utils.cpp`. The parameter `lookahead_dist` is not constant. Whatever the calculated lookahead distance comes out to be, it is updated to the variable `lookahead_dist`.

SakshayMahna:
Link to branch of the current work:
https://github.com/SakshayMahna/navigation2/tree/lookahead-optimization

nav2_graceful_controller/src/graceful_controller.cpp
SteveMacenski: @SakshayMahna there are many solutions

```
for (auto it = vec.rbegin(); it != vec.rend(); ++it)
for (size_t i = vec.size(); i-- > 0;)
```

nav2_util/src/controller_utils.cpp
SteveMacenski: @SakshayMahna ?

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,147 @@
...
+      const auto interpolated_position = circleSegmentIntersection(
+        last_pose_it->pose.position, projected_position, lookahead_dist);
+
+      geometry_msgs::msg::PoseStamped interpolated_pose;
+      interpolated_pose.header = last_pose_it->header;
+      interpolated_pose.pose.position = interpolated_position;
+
+      return interpolated_pose;
+    } else {
+      lookahead_dist = d;
```
SteveMacenski:
```suggestion
      lookahead_dist = d;  // Updating lookahead distance since using the final point
```

nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -195,46 +196,35 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
...
-      reversing = true;
-      target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
-        tf2::getYaw(target_pose.pose.orientation) + M_PI);
+  for (int i = transformed_plan.poses.size(); i >= 0; --i) {
+    if (i == static_cast<int>(transformed_plan.poses.size())) {
+      // Calculate target pose through lookahead interpolation to get most accurate
+      // lookahead point, if possible
+      dist_to_target = params_->max_lookahead;
+      // Interpolate after goal false for graceful controller
+      // Requires interpolating the orientation which is not yet implemented
```
SteveMacenski, line 214:
```suggestion
      // Requires interpolating the orientation which is not yet implemented
      // Updates dist_to_target for target_pose returned if using the point on the path
```

SteveMacenski:
@SakshayMahna sorry for the delay, that's my bad. I've been a bit underwater the last couple of weeks. I did another review and updated comments on open action items. Beyond those, I added 2 inline comments to be added while will help (mostly me) future developers looking at this code to understand important details. 

Please fix the conflicts and this should be good to go! The other open actions are pretty small too

nav2_util/src/controller_utils.cpp
SakshayMahna: https://github.com/SakshayMahna/navigation2/tree/lookahead-optimization

SteveMacenski:
@SakshayMahna any update? :smile:

SakshayMahna:
Hi Steve,
I tried to do the rebase. But I guess, again ended up with the issue I had previously. All the commits that are to this repository are showing up now, otherwise I made all the changes that you suggested.

I looked into a few ways to resolve this but couldn't find much success. Any idea how this can be resolved?

SteveMacenski:
Since I don't know entirely how you got into this state, my suggestions are rather limited. Push comes to shove, we've all been there were we checked out the updated main branch and either manually or via cherry-picking your previous commits (and dealing with the merge conflicts then and there) to get the changes. Then, open a new PR. Its not probably the best way, but given the number of conflicts this is showing and commit history, that would probably be faster and less frustrating than the "correct" technical solution.

SakshayMahna:
Fresh update the PR.

nav2_graceful_controller/src/graceful_controller.cpp
SakshayMahna: Went with this:
```cpp
bool is_first_iteration = true;
for (int i = transformed_plan.poses.size() - 1; i >= 0; --i) {
    if (is_first_iteration) ...
    else ...
}
```

SakshayMahna:
Hi @SteveMacenski 
I have made all the necessary changes. All the tests are passing as well (except `mypy` as discussed). How do we merge this now?

SteveMacenski:
On it right now :-)

This is again ... my bad. I really don't like this being the contributor experience to my projects, this has just had updates at wonky times in my schedule with either overload of Nav2 PRs to review / issues to respond to or in the middle of a project I needed dedicated days of work to make progress on. My apologies and I promise your PRs won't be like this in the future. I'm actually looking to bring on some part-time help with maintainer responsibilities to help fill gaps when I'm MIA. 

Edit: This looks good to me. 2 small changes and we should be able to merge this once pushed & CI is green

nav2_graceful_controller/src/parameter_handler.cpp
```diff
@@ -41,6 +41,8 @@ ParameterHandler::ParameterHandler(
...
+    node, plugin_name_ + ".interpolate_after_goal", rclcpp::ParameterValue(false));
```
SteveMacenski: @SakshayMahna didn't we talk about removing this for graceful bc it made it worse and just passing in false to the get lookahead point function? Did this get lost in the rebase issues? As well in the dynamic parameters and the test. It looks like the get lookahead point function is passed simply `false`, but we just need to remove the parameter

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,147 @@
...
+      prev_pose_it->pose.position,
+      goal_pose_it->pose.position, lookahead_dist);
+
+    // Calculate orientation towards interpolated position
+    // Convert yaw to quaternion
+    double yaw = atan2(
+      point.y - prev_pose_it->pose.position.y,
+      point.x - prev_pose_it->pose.position.x);
+
+    geometry_msgs::msg::Quaternion orientation;
```
SteveMacenski: We have a Nav2 util to do this for you: https://github.com/ros-navigation/navigation2/blob/main/nav2_util/include/nav2_util/geometry_utils.hpp#L39-L44 `orientationAroundZAxis`

SakshayMahna:
No problem, I'll try to close this by today or tomorrow.

nav2_graceful_controller/src/parameter_handler.cpp
SakshayMahna: Right, we removed it. Somehow got missed.

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,144 @@
...
+      point.y - prev_pose_it->pose.position.y,
+      point.x - prev_pose_it->pose.position.x);
+
+    geometry_msgs::msg::Quaternion orientation = geometry_utils::orientationAroundZAxis(yaw);
+
+    geometry_msgs::msg::PoseStamped pose;
+    pose.header.frame_id = prev_pose_it->header.frame_id;
+    pose.header.stamp = goal_pose_it->header.stamp;
+    pose.pose.position = point;
+    pose.pose.orientation = orientation;
```
SteveMacenski:
```suggestion
    pose.pose.orientation = geometry_utils::orientationAroundZAxis(yaw);;
```

Then you can remove L132:

```
    geometry_msgs::msg::Quaternion orientation = geometry_utils::orientationAroundZAxis(yaw);
```

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,144 @@
...
+      goal_pose_it->pose.position, lookahead_dist);
+
+    // Calculate orientation towards interpolated position
+    // Convert yaw to quaternion
+    double yaw = atan2(
+      point.y - prev_pose_it->pose.position.y,
+      point.x - prev_pose_it->pose.position.x);
+
+    geometry_msgs::msg::Quaternion orientation = geometry_utils::orientationAroundZAxis(yaw);
+
```
SteveMacenski:
```suggestion
```

SteveMacenski:
@SakshayMahna please pull in `main` so that CI can turn over. We had some API breaking changes in Rolling that need to be reflected here for it to build properly.

SteveMacenski (approved):

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // Convert yaw to quaternion
+    double yaw = atan2(
+      point.y - prev_pose_it->pose.position.y,
+      point.x - prev_pose_it->pose.position.x);
+
+    geometry_msgs::msg::PoseStamped pose;
+    pose.header.frame_id = prev_pose_it->header.frame_id;
+    pose.header.stamp = goal_pose_it->header.stamp;
+    pose.pose.position = point;
+    pose.pose.orientation = geometry_utils::orientationAroundZAxis(yaw);;
```
SteveMacenski:
```suggestion
    pose.pose.orientation = geometry_utils::orientationAroundZAxis(yaw);
```

SteveMacenski:
@SakshayMahna this ended up in a bad state again - I'm not exactly sure what commands you're running in rebase or pull in changes, but this needs to be reverted and try again. This changes 250+ files and creates conflicts. 

I just needed CI to turn over :smiling_face_with_tear:

SakshayMahna:
Reverted the extra commits and pulled from main.

SteveMacenski:
Some of the tests are failing from these changes:

```
/opt/overlay_ws/src/navigation2/nav2_graceful_controller/test/test_graceful_controller.cpp:295
  node->get_parameter("test.transform_tolerance").as_double()
    Which is: 0.1
  1.0
    Which is: 1

/opt/overlay_ws/src/navigation2/nav2_graceful_controller/test/test_graceful_controller.cpp:295
  node->get_parameter("test.transform_tolerance").as_double()
    Which is: 0.1
  1.0
    Which is: 1/opt/overlay_ws/src/navigation2/nav2_graceful_controller/test/test_graceful_controller.cpp:296
  node->get_parameter("test.min_lookahead").as_double()
    Which is: 0.25
  1.0
    Which is: 1

/opt/overlay_ws/src/navigation2/nav2_graceful_controller/test/test_graceful_controller.cpp:296
  node->get_parameter("test.min_lookahead").as_double()
    Which is: 0.25
  1.0
    Which is: 1/opt/overlay_ws/src/navigation2/nav2_graceful_controller/test/test_graceful_controller.cpp:297
  node->get_parameter("test.max_lookahead").as_double()
    Which is: 1
  2.0
    Which is: 2

/opt/overlay_ws/src/navigation2/nav2_graceful_controller/test/test_graceful_controller.cpp:297
  node->get_parameter("test.max_lookahead").as_double()
    Which is: 1
  2.0
    Which is: 2unknown file
C++ exception with description "test.interpolate_after_goal" thrown in the test body.

unknown file
C++ exception with description "test.interpolate_after_goal" thrown in the test body.
```

SteveMacenski:
> C++ exception with description "test.interpolate_after_goal" thrown in the test body.

This seems to be the issue, it doesn't update since that doesn't exist as a valid dynamic parameter so the other tests fail due to those not being updated from the failure

SteveMacenski (approved):

SteveMacenski:
@SakshayMahna thank you for the diligence! I know that took awhile! This is a really great update though to the controller and in architecture so we don't have that duplication - thanks so much!

SakshayMahna:
Thank you for pushing through as well Steve!
Happy to see it finally merged.

mini-1235:
After this PR, I often observe that the motion target heading is opposite to the robot's movement direction, is this expected? 
image

SteveMacenski:
Can you expand on that? I'm not sure @SakshayMahna and I fully grokk the picture's meaning.

nav2_util/src/controller_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    auto prev_pose_it = std::prev(goal_pose_it);
+    auto point = circleSegmentIntersection(
+      prev_pose_it->pose.position,
+      goal_pose_it->pose.position, lookahead_dist);
+
+    // Calculate orientation towards interpolated position
+    // Convert yaw to quaternion
+    double yaw = atan2(
+      point.y - prev_pose_it->pose.position.y,
+      point.x - prev_pose_it->pose.position.x);
```
SteveMacenski, line 130: @SakshayMahna this block is new compared to the version in RPP. Maybe this shuld be scrutinized?

mini-1235:
> Can you expand on that? I'm not sure @SakshayMahna and I fully grokk the picture's meaning. 

The red arrow is the motion target published by the graceful controller, while the blue path is the transformed plan. I thought the head of the red arrow should be the same direction with the transformed plan

If that's unclear, I will record a video tomorrow

SteveMacenski:
I agree it probably should be. Did you do an A-B test that this is definitely the PR that changes that behavior?

mini-1235:
> I agree it probably should be. Did you do an A-B test that this is definitely the PR that changes that behavior? 

I did, after reverting this commit, I saw the heading is always the same as the transformed plan

SteveMacenski:
@SakshayMahna does that make sense? Can you take a look :-) 

@mini-1235 do you have a reproduction example on the Tb3/4 sims that he can use as a reference to test?

mini-1235:
> @mini-1235 do you have a reproduction example on the Tb3/4 sims that he can use as a reference to test? 

It is very late in my time zone, I will write it down here tomorrow:)

mini-1235:
> @mini-1235 do you have a reproduction example on the Tb3/4 sims that he can use as a reference to test?

I replace the controller plugin in nav2_params and left everything else as default:
```
FollowPath:
      plugin: "nav2_graceful_controller::GracefulController"
```

Then launch the tb3 demo
```
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=0
```

I will show my result before and after this PR:

Before

https://github.com/user-attachments/assets/44510574-9a7d-41b9-b81d-4cb953509a20

After

https://github.com/user-attachments/assets/bf1b4db3-21a5-4791-a571-2525937442d3

SakshayMahna:
Hi @mini-1235 ,
Thank you for raising this. I understand the problem. I'll have a look.

nav2_util/src/controller_utils.cpp
SakshayMahna: Yes, this is the problem. There is a discontinuity taking place because of which the angle comes out to be reversed (added / subtracted by pi).

SakshayMahna:
As pointed out by Steve above, the problem is in the yaw interpolation. There is a discontinuity taking place because of which the angle comes out to be reversed (added / subtracted by pi).

I think I have a fix for it. Keeping the yaw value in a way to keep it in the direction of motion of robot.

mini-1235:
> As pointed out by Steve above, the problem is in the yaw interpolation. There is a discontinuity taking place because of which the angle comes out to be reversed (added / subtracted by pi).
> 
> I think I have a fix for it. Keeping the yaw value in a way to keep it in the direction of motion of robot.

@SakshayMahna Feel free to ping me when you open the PR, I can test it :)

SakshayMahna:
Thank you!
@mini-1235 Could you please have a look? I have created this PR: #5530

SteveMacenski:
Love it guys - please take a look and I will too!

Issue #5005, opened 2025-03-23
Parametrize collision checking in the nav2_graceful_controller


suchetanrs:
#### Feature description
Hello!
In certain cases of my robot's navigation, I want pure path-following behaviour without any collision detection. However, I want to do this with the graceful_controller.
Can we provide optional collision checking in nav2_graceful_controller like how it is done in the docking_server.

SteveMacenski:
Got it - sounds good!

PR #5006, opened 2025-03-23
* Parametrize collision checking in nav2_graceful_controller

suchetanrs:
## Basic Info
| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #5005 ) |
| Primary OS tested on | (Ubuntu 24.04 running Jazzy) |
| Robotic platform tested on | (Simulation of a custom diff-drive robot) |

---

* I added a parameter to nav2_graceful_controller to enable/disable collision checking

* Added new parameter, so need to add that to default configs and documentation page.

* I ran the test with a huge robot radius in the local_costmap to always have collision in the controller's path. The parametrization worked to enable/disable it.

#### For Maintainers: 

### DIFF
--- a/nav2_graceful_controller/src/graceful_controller.cpp
+++ b/nav2_graceful_controller/src/graceful_controller.cpp
@@ -56,10 +56,9 @@ void GracefulController::configure(
     params_->v_linear_min, params_->v_linear_max, params_->v_angular_max);
 
   // Initialize footprint collision checker
-  if(params_->use_collision_detection)
-  {
+  if(params_->use_collision_detection) {
     collision_checker_ = std::make_unique<nav2_costmap_2d::
-    FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>(costmap_ros_->getCostmap());
+        FootprintCollisionChecker<nav2_costmap_2d::Costmap2D *>>(costmap_ros_->getCostmap());
   }
 
   // Publishers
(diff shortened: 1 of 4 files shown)


nav2_graceful_controller/src/graceful_controller.cpp
```diff
@@ -177,7 +180,7 @@ geometry_msgs::msg::TwistStamped GracefulController::computeVelocityCommands(
       next_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(yaw);
       geometry_msgs::msg::PoseStamped costmap_pose;
       tf2::doTransform(next_pose, costmap_pose, costmap_transform);
-      if (inCollision(
+      if (use_collision_detection && inCollision(
```
SteveMacenski, line 183:
```suggestion
      if (params_->use_collision_detection && inCollision(
```

And remove L174 where `use_collision_detection` is set, that's not needed

SteveMacenski (commented):
LGTM, please update the docs https://docs.nav2.org/configuration/packages/configuring-graceful-motion-controller.html to include this new parameter and happy to merge the pair with a small adjustment

suchetanrs:
Thanks a lot for your response!
I have addressed your comment and also added the parameter to dynamic reconfigure.

A PR for the docs is raised here: https://github.com/ros-navigation/docs.nav2.org/pull/660

SteveMacenski:
@suchetanrs thanks and merged!

Issue #4995, opened 2025-03-19
clearAroundRobot and clearExceptRobot are limited to costmap bounds - STVL voxel area clearing mismatch


tonynajjar:
> With clearExceptRobot and clearAroundRobot, [we enforce the start and end point to be in the costmap](https://github.com/ros-navigation/navigation2/blob/c7a5cc56f6b460a818cad27c5c47848f0ab9f0d0/nav2_costmap_2d/src/clear_costmap_service.cpp#L138) and we pass these to STVL's [clearArea](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/7673fac40bb41ebf7419166045442a9d0dcd8af5/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp#L968). Because of that we can't reliably control the clearing (or lack thereof) of STVL's voxels e.g. with clearAroundRobot we can't clear voxels outside the costmap or with clearExceptRobot, we can't exclude voxels outside the costmap from clearing.
With clearEntire, it is not a problem because we call resetLayers() which resets the whole grid [here](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/7673fac40bb41ebf7419166045442a9d0dcd8af5/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp#L603)
Not really sure what the generic solution would be here; is there any other layer to your knowledge that stores data outside the costmap bounds?

See https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/320 for complete discussion

SteveMacenski:
This is a good first issue for users; essentially a user could request a clearing operation around the robot that exceeds the side of the local costmap. In that case, we should just clear the entire costmap rather than reducing the size to the costmap's size.

This is the main function which needs to be modified https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/src/clear_costmap_service.cpp#L126-L146 and the call for totally clearing the costmap is `clearEntirely()`. 

More context:

> Oh.. your radius is set generally larger than your local costmap size?  ... We could check in Costmap2D if that's the case and clear the local costmap completely since that should be an eq. operation in the case that radius >= costmap_size_x That should handle your comments in STVL implicitly

PR #322, opened 2018-11-14
Voxel Layer Port

bpwilcox:
---

## Basic Info

| Info | Please fill out this column |
| ------ | ----------- |
| Ticket(s) this addresses   | #193  |
| Primary OS tested on | Ubuntu 16.04 |
| Primary platform tested on | NUC |

--- 

## Description of contribution
- Adding port of voxel layer
- renamed namespaces in nav2_voxel_grid
- added voxel grid msg in nav2_msgs
- depends on obstacle layer, so only to be merged after PR #318 is merged
--- 

## Future work
- Everything compiles but I haven't run-time tested the voxel_layer port
- there are no unit/component tests written for voxel layer in the original costmap_2d test code

---

PR #5010, opened 2025-03-24
Clear costmap if reset distance exceeds costmap bounds.

leander-dsouza:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4995  |
| Robotic platform tested on | Turtlebot3 simulation in Gazebo |

---

* Shifted bound checking in the clearArea function, so that the STVL layer can safely override and implement voxel clearing for `reset_distance` greater than costmap width.
* Added `mapToWorldNoBounds` function for STVL overrides.

* N/A

* Add an STVL layer to global costmap, and use a depthimage to pointcloud converter node for the turtlebot4 example provided.

* Launch the turtlebot3 simulation with nav2 and gazebo:

   ```bash
   ros2 launch nav2_bringup tb4_simulation_launch.py headless:=False
   ```
* View the costmap behaviour in RViz after setting the `reset_distance` to 1000:

   ```bash
  ros2 service call /local_costmap/clear_around_local_costmap nav2_msgs/srv/ClearCostmapAroundRobot "{reset_distance: 1000.0}"
   ```
   This would still clear the STVL voxels displayed on RViz.
   
---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp
+++ b/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp
@@ -174,6 +174,15 @@ public:
   void mapToWorld(unsigned int mx, unsigned int my, double & wx, double & wy) const;
 
   /**
+  * @brief  Convert from map coordinates to world coordinates with no bounds checking
+  * @param  wx The x world coordinate
+  * @param  wy The y world coordinate
+  * @param  mx Will be set to the associated map x coordinate
+  * @param  my Will be set to the associated map y coordinate
+  */
+  void mapToWorldNoBounds(int mx, int my, double & wx, double & wy) const;
+
+  /**
    * @brief  Convert from world coordinates to map coordinates
    * @param  wx The x world coordinate
    * @param  wy The y world coordinate
--- a/nav2_costmap_2d/src/clear_costmap_service.cpp
+++ b/nav2_costmap_2d/src/clear_costmap_service.cpp
@@ -135,8 +135,8 @@ void ClearCostmapService::clearLayerRegion(
   double end_point_y = start_point_y + reset_distance;
 
   int start_x, start_y, end_x, end_y;
-  costmap->worldToMapEnforceBounds(start_point_x, start_point_y, start_x, start_y);
-  costmap->worldToMapEnforceBounds(end_point_x, end_point_y, end_x, end_y);
+  costmap->worldToMapNoBounds(start_point_x, start_point_y, start_x, start_y);
+  costmap->worldToMapNoBounds(end_point_x, end_point_y, end_x, end_y);
 
   costmap->clearArea(start_x, start_y, end_x, end_y, invert);
 
--- a/nav2_costmap_2d/src/costmap_2d.cpp
+++ b/nav2_costmap_2d/src/costmap_2d.cpp
@@ -282,6 +282,12 @@ void Costmap2D::mapToWorld(unsigned int mx, unsigned int my, double & wx, double
   wy = origin_y_ + (my + 0.5) * resolution_;
 }
 
+void Costmap2D::mapToWorldNoBounds(int mx, int my, double & wx, double & wy) const
+{
+  wx = origin_x_ + (mx + 0.5) * resolution_;
+  wy = origin_y_ + (my + 0.5) * resolution_;
+}
+
 bool Costmap2D::worldToMap(double wx, double wy, unsigned int & mx, unsigned int & my) const
 {
   if (wx < origin_x_ || wy < origin_y_) {
--- a/nav2_costmap_2d/src/costmap_layer.cpp
+++ b/nav2_costmap_2d/src/costmap_layer.cpp
@@ -66,10 +66,19 @@ void CostmapLayer::clearArea(int start_x, int start_y, int end_x, int end_y, boo
 {
   current_ = false;
   unsigned char * grid = getCharMap();
-  for (int x = 0; x < static_cast<int>(getSizeInCellsX()); x++) {
+
+  int size_x = getSizeInCellsX();
+  int size_y = getSizeInCellsY();
+
+  start_x = std::clamp(start_x, 0, size_x);
+  start_y = std::clamp(start_y, 0, size_y);
+  end_x = std::clamp(end_x, 0, size_x);
+  end_y = std::clamp(end_y, 0, size_y);
+
+  for (int x = 0; x < size_x; x++) {
     bool xrange = x > start_x && x < end_x;
 
-    for (int y = 0; y < static_cast<int>(getSizeInCellsY()); y++) {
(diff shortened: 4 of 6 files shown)


SteveMacenski:
@tonynajjar can you review since this solves your issue?

nav2_costmap_2d/src/clear_costmap_service.cpp
```diff
@@ -129,6 +129,18 @@ void ClearCostmapService::clearLayerRegion(
...
 
+  // Clear the entire costmap if the reset distance exceeds the costmap bounds
+  // for the case of clear around robot.
+  if (!invert) {
+    if (reset_distance >= costmap->getSizeInMetersX() &&
+      reset_distance >= costmap->getSizeInMetersY())
+    {
+      RCLCPP_INFO(logger_, "Reset area exceeds local costmap size. Clearing entire costmap.");
+      clearEntirely();
+      return;
```
SteveMacenski, line 140: I think we also want this block:

```

  double ox = costmap->getOriginX(), oy = costmap->getOriginY();
  double width = costmap->getSizeInMetersX(), height = costmap->getSizeInMetersY();
  costmap->addExtraBounds(ox, oy, ox + width, oy + height);
```

SteveMacenski (approved):

SteveMacenski:
Approved, just waiting on Tony's review!

nav2_costmap_2d/src/clear_costmap_service.cpp
```diff
@@ -129,6 +129,22 @@ void ClearCostmapService::clearLayerRegion(
 {
   std::unique_lock<Costmap2D::mutex_t> lock(*(costmap->getMutex()));
 
+  // Clear the entire costmap if the reset distance exceeds the costmap bounds
+  // for the case of clear around robot.
+  if (!invert) {
+    double width = costmap->getSizeInMetersX(), height = costmap->getSizeInMetersY();
+
+    if (reset_distance >= width && reset_distance >= height) {
+      RCLCPP_INFO(logger_, "Reset area exceeds local costmap size. Clearing entire costmap.");
```
tonynajjar, line 138: This could also be for the global costmap right? You could use `costmap_.getName()`

tonynajjar (commented):
I don't think this fully solves my problem. To my understanding what this does assuming non-inverted region: if the reset_distance is bigger than the costmap size, then the full costmap is cleared including **all the STVL voxels** (thourgh `resetLayers`)
I want only the STVL voxels in within the reset_distance to be cleared.
Similar logic for the inverted region.

Does it make sense?

nav2_costmap_2d/src/clear_costmap_service.cpp
SteveMacenski: Good point!

SteveMacenski:
> I want only the STVL voxels in within the reset_distance to be cleared. Similar logic for the inverted region.

I don't understand. I thought this is what we discussed in STVL and in the ticket body. 

The voxels are cleared within the reset_distance or except for the reset_distance depending on `invert`. When reset_distance is larger than the costmap we should:
- Clear the costmap entirely, when we're clearing inside of reset_distance -- as all of the costmap is within bounds for clearing
- Clear nothing when we're clearing outside of reset_distance -- as all of the costmap is out of bounds for clearing

nav2_costmap_2d/src/clear_costmap_service.cpp
leander-dsouza: Resolved :)

SteveMacenski:
OK, so I talked with tony and I think this is what is better to do: 

>  We could change the logic to have the boundary checking done by the layers themselves instead. All the layers but STVL would do that same bound checking, but STVL could bound check to clear the occ grid but use the fuller bounds to clear the internal data structure

That sounds like it solves the problem fully and more precisely. Does that make sense @leander-dsouza ?

leander-dsouza:
Hey @SteveMacenski , I am not sure why CircleCI cannot checkout my code.

leander-dsouza:
> Thanks and sorry for the wild goose chase!

Thank you for the clarification, Steve :)

* The issue with [Tony's PR](https://github.com/angsa-robotics/navigation2/compare/jazzy...angsa-robotics:navigation2:clear-area-fix?expand=1#diff-faa8e1875bb5cb72864f5a6b15b0b19e9d0317fb95d845493c0fad1d8fcbd759R68) was that the bound checking implemented was happening on the world coordinates rather than map.

`worldToMapEnforceBounds` is not an equivalent operation to apply bounds coming from `worldToMapNoBounds`, as it expects world coordinates.

* In order to combat this, I have clamped the coordinates to match the bounds set in [worldToMapEnforceBounds](https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d.cpp#L321).

* However, due to moving the bounds checking to `checkArea`,  `worldToMapNoBounds` output negative values for map if `reset_distance` is greater than costmap width. This causes an issue over at the STVL repository as [mapToWorld](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/ros2/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp#L964) expects positive values.

* Hence, [#322](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/322) must be merged to avoid underflow issues.

SteveMacenski (approved):

SteveMacenski:
@tonynajjar I think this is the intent of the branch you made - I think the actual diff you did wasn't quite right but what @leander-dsouza fixes that

SteveMacenski:
It looks like there's an issue on circle CI's side https://status.circleci.com/, I'd give it a day before we need to look into it ourselves since its likely related to them

tonynajjar (approved):
I tested it locally and it works well! one small nitpick from my side about function brief but will already pre-approve

nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp
```diff
@@ -181,6 +181,18 @@ class Costmap2D
...
+
+  /**
+  * @brief  Convert from map coordinates to world coordinates with no bounds checking
+  * @param  wx The x world coordinate
+  * @param  wy The y world coordinate
+  * @param  mx Will be set to the associated map x coordinate
+  * @param  my Will be set to the associated map y coordinate
+  * @return True if the conversion was successful (legal bounds) false otherwise
+  */
+
```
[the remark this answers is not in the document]
leander-dsouza: Resolved :)

leander-dsouza:
I think this is ready to merge, @SteveMacenski :)

SteveMacenski (approved):

SteveMacenski:
Thanks @leander-dsouza for the help and reimplementing based on the new understanding of the issue! Your help is very appreciated -- let me know if you want to take on something more meaty :cut_of_meat:

PR #5016, opened 2025-03-24
nav2_rviz_plugins: Don't use non-existent slot

wentasah:
The definition of the slot was removed in commit e6f500e5 ("nav2_rviz_plugins: Remove slots without implementation (#4974)", 2025-03-10), because it had no implementation. But we forgot to remove the reference to this slot, because the compiler cannot detect it.

We remove the reference now. Without this, rviz shows warnings like:

    QObject::connect: No such slot nav2_rviz_plugins::CostmapCostTool::updateAutoDeactivate()
    QObject::connect:  (sender name:   'Single click')

@mini-1235 Are you able to check that this removes the warnings for you?

---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_rviz_plugins/src/costmap_cost_tool.cpp
+++ b/nav2_rviz_plugins/src/costmap_cost_tool.cpp
@@ -30,7 +30,7 @@ CostmapCostTool::CostmapCostTool()
   auto_deactivate_property_ = new rviz_common::properties::BoolProperty(
     "Single click", true,
     "Switch away from this tool after one click.",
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

SteveMacenski:
@wentasah can you fix the DCO sign off? 

@mini-1235 if this works for you, I can merge!

mini-1235:
Yes, thanks for the fix, it removes the warning

PR #5022, opened 2025-03-26
Configured mypy strict for nav2_smac_planner

leander-dsouza:
---

## Basic Info

| ------ | ----------- |

---

* Added a mypy configuration file to ignore missing libraries and to treat each package independently, due to conflicting files.
* Restructured `nav2_smac_planner` for strict mypy.

* N/A

* Ran `colcon build` and `colcon test` to test the changes.

---

* N/A

#### For Maintainers: 

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,6 +31,10 @@ jobs:
           - "nav2_smac_planner"
     steps:
     - uses: actions/checkout@v4
+
+    - name: Install typeshed
+      run: sudo apt update && sudo apt install -y python3-typeshed
+
     - uses: ros-tooling/action-ros-lint@v0.1
       with:
         linter: mypy
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,10 +25,15 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: rostooling/setup-ros-docker:ubuntu-noble-ros-rolling-ros-base-latest
+    strategy:
+      matrix:
+        mypy_packages:
+          - "nav2_smac_planner"
     steps:
     - uses: actions/checkout@v4
     - uses: ros-tooling/action-ros-lint@v0.1
       with:
         linter: mypy
         distribution: rolling
-        package-name: "nav2_smac_planner"
+        package-name: "${{ join(matrix.mypy_packages, ' ') }}"
+        arguments: --config tools/pyproject.toml
--- a/tools/pyproject.toml
+++ b/tools/pyproject.toml
@@ -12,3 +12,11 @@ write-changes = false
 profile = "google"
 force_single_line = false
 line_length = 99
+
+[tool.mypy]
+explicit_package_bases = true
+strict = true
+
+[[tool.mypy.overrides]]
+module = ["matplotlib.*", "rtree.*"]
+ignore_missing_imports = true
--- a/nav2_common/package.xml
+++ b/nav2_common/package.xml
@@ -12,6 +12,7 @@
   <depend>osrf_pycommon</depend>
   <depend>rclpy</depend>
   <depend>python3-yaml</depend>
+  <depend>python3-types-pyyaml</depend>
 
   <buildtool_depend>ament_cmake_core</buildtool_depend>
 
--- a/nav2_smac_planner/lattice_primitives/generate_motion_primitives.py
+++ b/nav2_smac_planner/lattice_primitives/generate_motion_primitives.py
@@ -20,11 +20,11 @@ from pathlib import Path
 import time
 from typing import Any, cast, Dict, List, TypedDict
 
-import constants
-from lattice_generator import ConfigDict, LatticeGenerator
 import matplotlib.pyplot as plt
+from nav2_smac_planner.lattice_primitives import constants
+from nav2_smac_planner.lattice_primitives.lattice_generator import ConfigDict, LatticeGenerator
+from nav2_smac_planner.lattice_primitives.trajectory import Trajectory
 import numpy as np
-from trajectory import Trajectory
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
--- a/nav2_smac_planner/lattice_primitives/lattice_generator.py
+++ b/nav2_smac_planner/lattice_primitives/lattice_generator.py
@@ -16,13 +16,12 @@ from collections import defaultdict
 from enum import Enum
 from typing import Any, Dict, List, Tuple, TypedDict
 
-from helper import angle_difference, interpolate_yaws
+from nav2_smac_planner.lattice_primitives.helper import angle_difference, interpolate_yaws
+from nav2_smac_planner.lattice_primitives.trajectory import (AnyFloat, FloatNDArray, Path,
+                                                             Trajectory, TrajectoryParameters)
+from nav2_smac_planner.lattice_primitives.trajectory_generator import TrajectoryGenerator
 import numpy as np
 from rtree import index
-
-from trajectory import AnyFloat, FloatNDArray, Path, Trajectory, TrajectoryParameters
-
-from trajectory_generator import TrajectoryGenerator
 
 
 class ConfigDict(TypedDict):
--- a/nav2_smac_planner/lattice_primitives/tests/test_lattice_generator.py
+++ b/nav2_smac_planner/lattice_primitives/tests/test_lattice_generator.py
@@ -14,7 +14,7 @@
 
 import unittest
 
-from lattice_generator import ConfigDict, LatticeGenerator
+from nav2_smac_planner.lattice_primitives.lattice_generator import ConfigDict, LatticeGenerator
 import numpy as np
 
 MOTION_MODEL = 'ackermann'
--- a/nav2_smac_planner/lattice_primitives/tests/test_trajectory_generator.py
+++ b/nav2_smac_planner/lattice_primitives/tests/test_trajectory_generator.py
@@ -15,9 +15,10 @@
 from typing import cast
 import unittest
 
+from nav2_smac_planner.lattice_primitives.trajectory import Trajectory
+from nav2_smac_planner.lattice_primitives.trajectory_generator import (
+    TrajectoryGenerator, TrajectoryGeneratorConfigDict)
 import numpy as np
-from trajectory import Trajectory
-from trajectory_generator import TrajectoryGenerator, TrajectoryGeneratorConfigDict
 
 TURNING_RADIUS = 1
 STEP_DISTANCE = 0.1
--- a/nav2_smac_planner/lattice_primitives/trajectory.py
+++ b/nav2_smac_planner/lattice_primitives/trajectory.py
@@ -15,7 +15,7 @@
 from dataclasses import dataclass
 from typing import Any, Union
 
-from helper import angle_difference, normalize_angle
+from nav2_smac_planner.lattice_primitives.helper import angle_difference, normalize_angle
 import numpy as np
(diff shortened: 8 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -95,3 +95,10 @@ repos:
         language: system
         types: [python]
         entry: ament_pep257
+    -   id: ament_mypy
+        name: ament_mypy
+        description: Check Python code using mypy.
+        language: system
+        types: [python]
+        args: ["--config", "./tools/pyproject.toml"]
+        entry: ament_mypy
```
Nils-ChristianIseke, line 104: Here you use a custom configuration for ament_mypy. In the CI run you don't. What is the reasoning behind this?

.github/workflows/lint.yml
```diff
@@ -11,24 +11,29 @@ jobs:
...
-        package-name: "nav2_smac_planner"
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+      - uses: pre-commit/action@v3.0.1
+        env:
+          SKIP: >-
+            ament_lint_cmake,
+            ament_cpplint,
+            ament_uncrustify,
+            ament_xmllint,
```
Nils-ChristianIseke, line 36: I like running precommit in the CI (skipping the ament_stuff). Nice idea!. Just a note: I would do this in a separate PR, as it has nothing to do with mypy in terms of content. Then you'll get the PR completed more quickly with fewer discussions

.pre-commit-config.yaml
```diff
@@ -95,3 +95,10 @@ repos:
         language: system
         types: [python]
         entry: ament_pep257
+    -   id: ament_mypy
+        name: ament_mypy
+        description: Check Python code using mypy.
+        language: system
+        types: [python]
+        args: ["--config", "./tools/pyproject.toml"]
```
Nils-ChristianIseke, line 103: Running ament_mypy via pre-commit you use a custom configuration, running ament_mypy via CI you dont't?
What is the reasoning behind this?

tools/pyproject.toml
```diff
@@ -12,3 +12,6 @@ write-changes = false
 profile = "google"
 force_single_line = false
 line_length = 99
+[tool.mypy]
+disable_error_code = ["import-untyped", "import-not-found", "attr-defined"]
+explicit_package_bases = true
```
Nils-ChristianIseke, line 17: Why is explicit_package_bases = true necessary? I would assume that ament_mypy is handling this already.

tools/pyproject.toml
```diff
@@ -12,3 +12,6 @@ write-changes = false
 profile = "google"
 force_single_line = false
 line_length = 99
+[tool.mypy]
+disable_error_code = ["import-untyped", "import-not-found", "attr-defined"]
```
Nils-ChristianIseke, line 16: I generally like the idea of using a custom configuration for ament_mypy. E.g. to enable ament_mypy --strict. However i think @SteveMacenski wants to use the default configuration of ament_mypy. Furthermore i don't think disabling all kinds of mypy errors via this config is the way to go here.

Nils-ChristianIseke (commented):
Thanks @leander-dsouza! I have a few questions regarding the configuration of ament_mypy. Furthermore, i think the cleaner solution would be to introduce type hints along with ament_mypy (see #4916). Accordingly, a ament_mypy support package by package in seperate PRs.

tools/pyproject.toml
Nils-ChristianIseke, line 17: I am a bit surprised that this is not part of the [ament_mypy configuration](https://github.com/ament/ament_lint/blob/rolling/ament_mypy/ament_mypy/configuration/ament_mypy.ini). Maybe we should open a PR regarding this configuration setting in the ament_lint repo.

.github/workflows/lint.yml
SteveMacenski: Agreed, please just setup mypy without only the smac planner and we can merge this.

A follow up PR can be used for that :-)

tools/pyproject.toml
SteveMacenski: Agreed, unless there are a few exceptional errors that we don't want to deal with for tangible reasons (which may actually exist)

tools/pyproject.toml
SteveMacenski: I agree if you think its good to include!

SteveMacenski (commented):
I approve of all code changes, the CI setup and the config file have some open questions though! :100:

tools/pyproject.toml
SteveMacenski: I don't have a strong feeling one way or another on this one, should we keep it on?

tools/pyproject.toml
leander-dsouza: I have removed the section containing the disabling of certain error codes.

tools/pyproject.toml
leander-dsouza: The default [ament_mypy_configuration](https://github.com/ament/ament_lint/blob/rolling/ament_mypy/ament_mypy/configuration/ament_mypy.ini) has `ignore_missing_imports = true`. I have additionally added `explicit_package_bases = true` to treat each package independently.

I think this can be added as a PR for the [ament_lint](https://github.com/ament/ament_lint/tree/rolling) repository.

leander-dsouza: I think we can keep this file, to include `strict = true` for strict type checking.

leander-dsouza:
> Thanks @leander-dsouza! I have a few questions regarding the configuration of ament_mypy. Furthermore, I think the cleaner solution would be to introduce type hints along with ament_mypy (see #4916). Accordingly, a ament_mypy support package by package in seperate PRs. But this is only my personal preference :)

You're welcome, @Nils-ChristianIseke :)
I think that we can introduce type hints with the configuration file that I have attached (`strict = true`)
I feel that this PR can focus on just getting `ament_mypy` enabled for all packages without strict type checking, and for future development, individual PRs for each package can be made for `strict=true`.

tools/pyproject.toml
SteveMacenski: I think these 2 additions sound good to me. Just waiting for @Nils-ChristianIseke's OK

.github/workflows/lint.yml
```diff
@@ -11,24 +11,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          linter: [xmllint, cpplint, uncrustify, pep257, flake8]
+          linter: [xmllint, cpplint, uncrustify, pep257, flake8, mypy]
     steps:
     - uses: actions/checkout@v4
+
+    - name: Install typeshed for mypy
```
SteveMacenski, line 18: Why?

SteveMacenski: Can we make this a more explicit dependency of nav2_common or something? Adding this here would cause problems for users trying to run it locally then

leander-dsouza: Yes, you are right. I have added `python3-types-pyyaml` as a dependency for local development as it installs `typeshed` through [rosdep](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml#L10447).

SteveMacenski (approved):

SteveMacenski:
Just waiting on @Nils-ChristianIseke approval as a second check but I'm happy with this!

tools/pyproject.toml
Nils-ChristianIseke: Yes let use strict = True. However that will result in many mypy errors, so you just have to configure mypy to only check the packages that have type hints (currently only nav2_smac_planner). An you will most certanly have to tweak the type hints of nav2_smac_planner.

tools/pyproject.toml
```diff
@@ -12,3 +12,6 @@ write-changes = false
 profile = "google"
 force_single_line = false
 line_length = 99
+[tool.mypy]
+explicit_package_bases = true
+ignore_missing_imports = true
```
Nils-ChristianIseke, line 17: Why is this needed? If that is set globally we weaking the static type checking very much (see [mypy doc](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker). If there are imports that do not support typing, they should be ignored by adding the following to pyproject.toml: 

[[tool.mypy.overrides]]
module = ["foobar.*"]
ignore_missing_imports = true

Nils-ChristianIseke (changes requested):
see latest comments

leander-dsouza:
> My approach would be:
> 
> 1. Enable --strict now to ensure all new code meets high typing standards.
> 2. Roll out type hints incrementally (package by package) to manage workload.

Thank you for the suggestions, @Nils-ChristianIseke :)
I have now made the PR focused on applying a strict mypy convention for `nav2_smac_planner`. I will make PRs for each package, henceforth including strict mypy standards.

tools/pyproject.toml
leander-dsouza: Thank you for the tip!
I have applied this to the package accordingly :)

Nils-ChristianIseke:
I've tested locally but can't reproduce an error-free run with either mypy or ament_mypy. Could you share how you invoked an error free run?

Some more issues i ran into:
- We should move all configs to pyproject.toml (currently the arg package is used in CI, but should be moved to pyproject.toml) 
- CI uses bare mypy (@SteveMacenski I think that slipped through the review of #4927), however ament_mypy ignores the  package arg (needed for incremental mypy adoption, to only run mypy checks on specified packages)

@SteveMacenski, apologies for circling back on this - given ament_mypy's limitations, would you be okay with sticking with bare mypy for now?
 I'll try to submit a PR to support the package arg for ament_mypy.

.github/workflows/lint.yml
```diff
@@ -27,8 +27,13 @@ jobs:
...
     - uses: actions/checkout@v4
+
+    - name: Install typeshed
+      run: sudo apt update && sudo apt install -y python3-typeshed
+
     - uses: ros-tooling/action-ros-lint@v0.1
       with:
         linter: mypy
         distribution: rolling
         package-name: "nav2_smac_planner"
```
Nils-ChristianIseke, line 38: should be moved to pyproject.toml

leander-dsouza:
> I've tested locally but can't reproduce an error-free run with either mypy or ament_mypy. Could you share how you invoked an error free run?

You can test the `nav2_smac_planner` package as follows:

```bash
ament_mypy --config tools/pyproject.toml nav2_smac_planner/
```

Nils-ChristianIseke:
> I was thinking that we can append to the CI for each package henceforth, and finally, when all packages are integrated, just replace it with `*`.

> I do not think we can append packages to the pyproject.toml yet, as it is not implemented.

@leander-dsouza Sorry for the noise, I made a mistake in the review, you're right

Nils-ChristianIseke (approved):
@leander-dsouza Thank you, for your effort and patience :).

Nils-ChristianIseke:
> You can test the `nav2_smac_planner` package as follows:
> 
> ```shell
> ament_mypy --config tools/pyproject.toml nav2_smac_planner/
> ```

Just for your information, I had a mypy version that I had previously installed via pip (which was different (the latest) from the mypy version installed via apt), which resulted in errors.

SteveMacenski (approved):

SteveMacenski:
Is the next step to enable the packages one by one?

It might be good to do the rest in one-go so the packages being selected are `'*'`. Using the 'all' wildcard would be good so that I / contributors don't forget to add new python packages to the list if/when new ones are added in the future

Nils-ChristianIseke:
Supporting type hints in for all packages in one go would be a quite big PR:
_Found 1529 errors in 100 files (checked 115 source files)_

PR #5023, opened 2025-03-26
Declare_parameter_if_not_declared in docking navigator

ajtudela:
---

## Basic Info

| ------ | ----------- |

---

* Replaced "declare_parameter("navigator_bt_xml")" with declare_parameter_if_not_declared in the docking_server navigator. This change allows the docking server to be deactivated and activated again.
* Added a new test to navigator.

---

#### For Maintainers: 

### DIFF
--- a/nav2_docking/opennav_docking/src/navigator.cpp
+++ b/nav2_docking/opennav_docking/src/navigator.cpp
@@ -24,7 +24,8 @@ Navigator::Navigator(const rclcpp_lifecycle::LifecycleNode::WeakPtr & parent)
 : node_(parent)
 {
   auto node = node_.lock();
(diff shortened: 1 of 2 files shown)


SteveMacenski (commented):
I'm a little unclear about this - we have plenty of nodes that have declares in the constructor (example: https://github.com/ros-navigation/navigation2/blob/a904802fe222f961d4a211fa0c813349672c18eb/nav2_planner/src/planner_server.cpp#L53) 

This should be fine when you reset the smart pointer since that deallocates the memory and its later reallocated so the node is a different instance.

SteveMacenski (approved):
Oh, but this is given the node that it declares, I see. 

Approved!

PR #5027, opened 2025-03-27
Halt recovery action when main action succeeds to reset status

### LINKED ISSUE #4737: RoundRobin in recovery is not reset when main branch success
## Bug report

**Required Info:**

- Operating System:
  - Custom Linux
- ROS2 Version:
  - Iron source
- Version or commit hash:
  - 1.2.9
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
In our Behavior Tree, we have a RecoveryNode with a RoundRobin in the recovery branch, structured as follows:

revovery_round_robin drawio

#### Expected behavior
We expect that each time the main action succeeds, the round_robin is reset. Therefore, the next time the main action fails, the first recovery action is triggered.

This behavior is illustrated by the following sequence:
1 - Main action: Fail
2 - Recovery A: Success
3 - Main action: Fail
4 - Recovery B: Success
5 - Main action: Success
... (later)
6 - Main action: Fail
7 - **Recovery A is call**
#### Actual behavior
Currently, we observe that the roundRobin is not reset when the main branch succeeds. Therefore, the next time the main action fails, the roundRobin continues incrementing the child index.

This behavior is illustrated by the following sequence:
1 - Main action: Fail
2 - Recovery A: Success
3 - Main action: Fail
4 - Recovery B: Success
5 - Main action: Success
... (later)
6 - Main action: Fail
7 - **Recovery C is call**

#### Additional information
After checking the code, we noticed that the only way to reset the roundRobin is by calling the halt function. However, in the BehaviorTreeCPP library, the halt function is only called when the target node is in the RUNNING state (which seems consistent with the function name). The issue is that the Recovery node does not call the roundRobin's halt function when it is in the RUNNING state.

My question, then, is: Is this the intended behavior? And if so, do you have any suggestions on how to implement the expected behavior sequence?

SteveMacenski:
Resolves #4737

@tonynajjar  @FrGrQuim can you review?

### DIFF
--- a/nav2_behavior_tree/plugins/control/recovery_node.cpp
+++ b/nav2_behavior_tree/plugins/control/recovery_node.cpp
@@ -52,6 +52,7 @@ BT::NodeStatus RecoveryNode::tick()
 
         case BT::NodeStatus::SUCCESS:
           // reset node and return success when first child returns success
+          // also halt the recovery action as the main action is successful, reset its state
           ControlNode::haltChild(1);
           halt();
           return BT::NodeStatus::SUCCESS;
--- a/nav2_behavior_tree/plugins/control/recovery_node.cpp
+++ b/nav2_behavior_tree/plugins/control/recovery_node.cpp
@@ -52,6 +52,7 @@ BT::NodeStatus RecoveryNode::tick()
 
         case BT::NodeStatus::SUCCESS:
           // reset node and return success when first child returns success
+          ControlNode::haltChild(1);
           halt();
           return BT::NodeStatus::SUCCESS;
 


nav2_behavior_tree/plugins/control/recovery_node.cpp
```diff
@@ -52,6 +52,7 @@ BT::NodeStatus RecoveryNode::tick()
 
         case BT::NodeStatus::SUCCESS:
           // reset node and return success when first child returns success
+          ControlNode::haltChild(1);
```
SteveMacenski, line 55:
```suggestion
          // also halt the recovery action as the main action is successful, reset its state
          ControlNode::haltChild(1);
```

tonynajjar (approved):

PR #5034, opened 2025-04-01
feat(nav2_rotation_shim_controller): add use_path_orientations

bektaskemal:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | None|

---

- With this PR, rotation shim controller optionally use orientations from the path so the robot rotates to pose orientation. This allows it to be usable with reverse driving.

- Tested in simulation.

---

- I would like to backport this to jazzy.

#### For Maintainers: 

### DIFF
--- a/nav2_rotation_shim_controller/README.md
+++ b/nav2_rotation_shim_controller/README.md
@@ -38,6 +38,7 @@ See its [Configuration Guide Page](https://docs.nav2.org/configuration/packages/
 | `max_angular_accel` | Maximum angular acceleration for rotation to heading |
 | `simulate_ahead_time` | Time in seconds to forward simulate a rotation command to check for collisions. If a collision is found, forwards control back to the primary controller plugin. |
 | `rotate_to_goal_heading` | If true, the rotationShimController will take back control of the robot when in XY tolerance of the goal and start rotating to the goal heading |
+| `use_path_orientations` | If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. Use for for feasible planners like the Smac Planner which generate feasible paths with orientations for forward and reverse motion. |
 
 Example fully-described XML with default parameter values:
 
@@ -70,6 +71,7 @@ controller_server:
       max_angular_accel: 3.2
       simulate_ahead_time: 1.0
       rotate_to_goal_heading: false
+      use_path_orientations: false
 
       # DWB parameters
       ...
--- a/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
+++ b/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
@@ -75,6 +75,8 @@ void RotationShimController::configure(
     node, plugin_name_ + ".rotate_to_heading_once", rclcpp::ParameterValue(false));
   nav2_util::declare_parameter_if_not_declared(
     node, plugin_name_ + ".closed_loop", rclcpp::ParameterValue(true));
+  nav2_util::declare_parameter_if_not_declared(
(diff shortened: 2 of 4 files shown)


nav2_rotation_shim_controller/README.md
```diff
@@ -38,6 +38,7 @@ See its [Configuration Guide Page](https://docs.nav2.org/configuration/packages/
 | `max_angular_accel` | Maximum angular acceleration for rotation to heading |
 | `simulate_ahead_time` | Time in seconds to forward simulate a rotation command to check for collisions. If a collision is found, forwards control back to the primary controller plugin. |
 | `rotate_to_goal_heading` | If true, the rotationShimController will take back control of the robot when in XY tolerance of the goal and start rotating to the goal heading |
+| `use_path_orientations` | If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. |
```
SteveMacenski, line 41:
```suggestion
| `use_path_orientations` | If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. Use for for feasible planners like the Smac Planner which generate feasible paths with orientations for forward and reverse motion. |
```

I assume this is what you mean to use this with? 

Make sure to open a docs PR in the docs.nav2.org repository to also add this parameter. Add a migration guide entry too since this is a new feature of note

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -207,10 +210,17 @@ geometry_msgs::msg::TwistStamped RotationShimController::computeVelocityCommands
 
     std::lock_guard<std::mutex> lock_reinit(mutex_);
     try {
-      geometry_msgs::msg::Pose sampled_pt_base = transformPoseToBaseFrame(getSampledPathPt());
-
-      double angular_distance_to_heading =
-        std::atan2(sampled_pt_base.position.y, sampled_pt_base.position.x);
+      auto sampled_pt = getSampledPathPt();
+      auto angular_distance_to_heading = [&](){
```
SteveMacenski, line 214: This is a little odd as a lambda, can you just make this an if/else case to populate the `angular_distance_to_heading` value?

SteveMacenski (changes requested):
Great idea, minor updates needed but can merge after that :-)

nav2_rotation_shim_controller/README.md
bektaskemal: Docs PR: https://github.com/ros-navigation/docs.nav2.org/pull/666

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -207,10 +210,17 @@ geometry_msgs::msg::TwistStamped RotationShimController::computeVelocityCommands
...
-      geometry_msgs::msg::Pose sampled_pt_base = transformPoseToBaseFrame(getSampledPathPt());
-
-      double angular_distance_to_heading =
-        std::atan2(sampled_pt_base.position.y, sampled_pt_base.position.x);
+      auto sampled_pt = getSampledPathPt();
+      double angular_distance_to_heading;
+      if (use_path_orientations_) {
+        angular_distance_to_heading = angles::shortest_angular_distance(
+                  tf2::getYaw(pose.pose.orientation),
+                  tf2::getYaw(sampled_pt.pose.orientation));
```
SteveMacenski, line 218:
```suggestion
          tf2::getYaw(pose.pose.orientation),
          tf2::getYaw(sampled_pt.pose.orientation));
```

SteveMacenski (approved):

mini-1235:
I noticed that one of the conflict in #5176 actually comes from here, should be backport this to Jazzy as well? @SteveMacenski

SteveMacenski:
Sure :-)

Issue #4842, opened 2025-01-10
condition distance traveled always returns failure


guidout:
Since when the behavior tree is first loaded, all nodes are in the IDLE state before they are ticked for the first time, doesn't this if statement always return IDLE and therefore the tick method always return FAILURE?

https://github.com/ros-navigation/navigation2/blob/507b3659e6c622d9339b61b53810320b600526a2/nav2_behavior_tree/plugins/condition/distance_traveled_condition.cpp#L46

I also asked the question here but didn't get answers:
https://robotics.stackexchange.com/questions/114059/nav2-distance-traveled-condition

SteveMacenski:
Please do not cross post.

Have you actually tried to use it and see that it doesn't work or just speculating looking at the software? We have a unit test that seems to point to it working [1]

[1] https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/test/plugins/condition/test_distance_traveled.cpp

gritelli:
Yes, I did test it and the condition always returns IDLE.
This is the BT I used for the test:
```
<root main_tree_to_execute="MainTree">
  <BehaviorTree ID="MainTree">

    <RateController hz="1.0">
      <ReactiveFallback name="CheckDistanceTravelled">
        <DistanceTraveled distance="2.0" global_frame="map" robot_base_frame="base_link"
        />

        <PipelineSequence>
          <GoalUpdatedController>
            <RecoveryNode number_of_retries="1" name="RecoveryComputePathToPose">
              <ComputePathToPose goal="{goal}" path="{path}"
                planner_id="GridBased_SmacHybrid_production_line" />
              <ClearEntireCostmap name="ClearGlobalCostmap-Context"
                service_name="global_costmap/clear_entirely_global_costmap" />
            </RecoveryNode>
          </GoalUpdatedController>
          <RecoveryNode number_of_retries="1" name="RecoveryFollowPath">
            <FollowPath path="{path}" controller_id="FollowPath_TEB_prod_line"
              goal_checker_id="general_goal_checker" />
            <ClearEntireCostmap name="ClearLocalCostmap-Context"
              service_name="local_costmap/clear_entirely_local_costmap" />
          </RecoveryNode>
        </PipelineSequence>
      </ReactiveFallback>
    </RateController>
  </BehaviorTree>
</root>
```
I added a print out in the `distance_traveled_condition.cpp`:
```
BT::NodeStatus DistanceTraveledCondition::tick()
{
[...]

PR #5035, opened 2025-04-01
Add note to Asynchronous nodes

adivardi:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Discussion from #4842 |

---

Add a warning to nodes using `BT::isStatusActive()` to initialize. Such nodes should be considred Asynchronous and should be used cautiously with Reactive control nodes (only one Asynchronous action may be used). This is because, when a node reuturn RUNNING inside a Reactive control node, all other nodes are halted. In the next tick, halted nodes will not be active and will reinitialize again.

Some of the nodes may also return RUNNING, and the warning also points that.

---

See discussion in #4842 (in particular [this message and the one below it](https://github.com/ros-navigation/navigation2/issues/4842#issuecomment-2725525872)

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
@@ -34,7 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what controller must be used. It is usually used before of
  * the FollowPath. The selected_controller output port is passed to controller_id
  * input port of the FollowPath
- * @note This is an Asynchronous node. It will reset when halted.
+ * @note This is an Asynchronous node. It will re-initialize when halted.
  */
 class ControllerSelector : public BT::SyncActionNode
 {
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
@@ -34,6 +34,8 @@ namespace nav2_behavior_tree
  * to get the decision about what controller must be used. It is usually used before of
  * the FollowPath. The selected_controller output port is passed to controller_id
  * input port of the FollowPath
+ * @warning This is an Asynchronous Action, which will reset when halted
+ *          Use cautiously with Reactive control nodes (only one Asynchronous action may be used)
  */
 class ControllerSelector : public BT::SyncActionNode
 {
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
@@ -34,8 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what controller must be used. It is usually used before of
  * the FollowPath. The selected_controller output port is passed to controller_id
  * input port of the FollowPath
- * @warning This is an Asynchronous Action, which will reset when halted
- *          Use cautiously with Reactive control nodes (only one Asynchronous action may be used)
+ * @note This is an Asynchronous node. It will reset when halted.
  */
 class ControllerSelector : public BT::SyncActionNode
 {
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/goal_checker_selector_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/goal_checker_selector_node.hpp
@@ -34,7 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what goal_checker must be used. It is usually used before of
  * the FollowPath. The selected_goal_checker output port is passed to goal_checker_id
  * input port of the FollowPath
- * @note This is an Asynchronous node. It will reset when halted.
+ * @note This is an Asynchronous node. It will re-initialize when halted.
  */
 class GoalCheckerSelector : public BT::SyncActionNode
 {
--- a/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/goal_checker_selector_node.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/goal_checker_selector_node.hpp
@@ -34,6 +34,8 @@ namespace nav2_behavior_tree
  * to get the decision about what goal_checker must be used. It is usually used before of
  * the FollowPath. The selected_goal_checker output port is passed to goal_checker_id
  * input port of the FollowPath
+ * @warning This is an Asynchronous Action, which will reset when halted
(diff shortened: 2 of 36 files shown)


SteveMacenski (commented):
I don't think this is a 'warning', but rather just a note. Something like the following:

This is a long-running node which may return a RUNNING state while executing.

Also, in following PRs, please don't check box the 'For Maintainers' section, that is for maintainers to go over :-)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/remove_passed_goals_action.hpp
```diff
@@ -28,6 +28,10 @@
 namespace nav2_behavior_tree
 {
 
+/**
+ * @brief A BT::ActionNodeBase that removes goals that the robot passed near to
+ * @note This is an Asynchronous node. It will reset when halted.
```
SteveMacenski, line 33: I don't think this is able to return running

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/smoother_selector_node.hpp
```diff
@@ -34,6 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what smoother must be used. It is usually used before of
  * the FollowPath. The selected_smoother output port is passed to smoother_id
  * input port of the FollowPath
+ * @note This is an Asynchronous node. It will reset when halted.
```
SteveMacenski, line 37: I don't think this is able to return running

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/progress_checker_selector_node.hpp
```diff
@@ -33,6 +33,7 @@ namespace nav2_behavior_tree
  * to get the decision about what progress_checker must be used. It is usually used before of
  * the FollowPath. The selected_progress_checker output port is passed to progress_checker_id
  * input port of the FollowPath
+ * @note This is an Asynchronous node. It will reset when halted.
```
SteveMacenski, line 36: I don't think this is able to return running

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/planner_selector_node.hpp
```diff
@@ -34,6 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what planner must be used. It is usually used before of
  * the ComputePathToPoseAction. The selected_planner output port is passed to planner_id
  * input port of the ComputePathToPoseAction
+ * @note This is an Asynchronous node. It will reset when halted.
```
SteveMacenski, line 37: I don't think this is able to return running

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/goal_checker_selector_node.hpp
```diff
@@ -34,6 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what goal_checker must be used. It is usually used before of
  * the FollowPath. The selected_goal_checker output port is passed to goal_checker_id
  * input port of the FollowPath
+ * @note This is an Asynchronous node. It will reset when halted.
```
SteveMacenski, line 37: I don't think this is able to return running

nav2_behavior_tree/include/nav2_behavior_tree/plugins/action/controller_selector_node.hpp
```diff
@@ -34,6 +34,7 @@ namespace nav2_behavior_tree
  * to get the decision about what controller must be used. It is usually used before of
  * the FollowPath. The selected_controller output port is passed to controller_id
  * input port of the FollowPath
+ * @note This is an Asynchronous node. It will reset when halted.
```
SteveMacenski, line 37: I don't think this is able to return running

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/transform_available_condition.hpp
```diff
@@ -29,6 +29,8 @@ namespace nav2_behavior_tree
 /**
  * @brief A BT::ConditionNode that returns SUCCESS if there is a valid transform
  * between two specified frames and FAILURE otherwise
+ * @note This is an Asynchronous (long-running) node which may return a RUNNING state while executing.
```
SteveMacenski, line 32: I don't think this is able to return running

SteveMacenski (changes requested):
I marked a few, but didn't get through all of them - I don't think some of these nodes actually do return a `RUNNING` state as a valid condition.

adivardi:
> I marked a few, but didn't get through all of them - I don't think some of these nodes actually do return a `RUNNING` state in their implementations. In some cases, I don't think that BT node type is able to do return RUNNING as a matter of the BT.CPP library.

Yes, but the issue is not the nodes returning `RUNNING`, but the use of 
```
 if (!BT::isStatusActive(status())) {
    initialize();
  }
```
Every node which has this code, will re-initialize itself after it was halted. If it is under a `reactiveSequence` with another node which may return RUNNING, it will re-initialize every tick.

For example, in this subtree (taken from https://github.com/adivardi/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_through_poses_w_replanning_and_recovery.xml#L16), `RemovePassedGoals::initialize` is called at every tick while `ComputePathThroughPoses` is `RUNNING`

image

Which is why I added the note `It will reset when halted.`  (maybe `reinitialize` is clearer, I will change that)

SteveMacenski (approved):

nav2_behavior_tree/include/nav2_behavior_tree/plugins/condition/goal_reached_condition.hpp
```diff
@@ -29,6 +29,8 @@ namespace nav2_behavior_tree
 /**
  * @brief A BT::ConditionNode that returns SUCCESS when a specified goal
  * is reached and FAILURE otherwise
+ * @note This is an Asynchronous (long-running) node which may return a RUNNING state while executing.
```
mini-1235, line 32: > Every node which has this code, will re-initialize itself after it was halted. If it is under a reactiveSequence with another node which may return RUNNING, it will re-initialize every tick.

I think I understand this

But I don't understand the notes in condition nodes, will they return RUNNING while executing?

adivardi: The note was shortened than the original, maybe it doesn't apply to all of these nodes. The main motivation was to highlight that the use of
```
  if (!BT::isStatusActive(status())) {
    initialize();
  }
```
causes these nodes to reinitialize whenever they are halted (which happens every loop in a Reactive control sequence with another `Asynchronous` node).

PR #5039, opened 2025-04-01
Backport #4996 (Humble)

mini-1235:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/critics/cost_critic.cpp
+++ b/nav2_mppi_controller/src/critics/cost_critic.cpp
@@ -15,6 +15,7 @@
 
 #include <cmath>
 #include "nav2_mppi_controller/critics/cost_critic.hpp"
+#include "nav2_core/exceptions.hpp"
 
 namespace mppi::critics
 {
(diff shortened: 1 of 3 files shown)


nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -52,6 +53,18 @@ void CostCritic::initialize()
       " for full instructions. This will substantially impact run-time performance.");
   }
 
+  if (costmap_ros_->getUseRadius() == consider_footprint_) {
+    RCLCPP_WARN(
+      logger_,
+      "Inconsistent configuration in collision checking. Please verify the robot's shape settings "
+      "in both the costmap and the cost critic.");
+    if (costmap_ros_->getUseRadius()) {
+      throw nav2_core::PlannerException(
```
SteveMacenski, line 62: I don't think this is proper - there is no planner exception in Humble, I do not believe: https://github.com/ros-navigation/navigation2/blob/humble/nav2_planner/src/planner_server.cpp#L451

I think we should just warn for Humble, not to break users currently deployed

SteveMacenski (changes requested):

nav2_mppi_controller/src/critics/cost_critic.cpp
mini-1235: I think planner exception is actually used in other controllers
https://github.com/ros-navigation/navigation2/blob/7ee3d2f65fcf42e18c41e0801199a33aa0b4c76e/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L359

that's why I am switching from ControllerException to PlannerException

SteveMacenski: Ah ok - that works

SteveMacenski (approved):

Issue #4735, opened 2024-10-29
SmacPlanner hybrid can generate path that pass inside lethal area


FrGrQuim:
## Bug report

**Required Info:**

- Operating System:
  - Linux (custom)
- ROS2 Version:
  - Iron sources
- Version or commit hash:
  - 1.2.9
- DDS implementation:
  - CycloneDDS

#### Steps to reproduce issue
- Get a costmap with obsacles (static or dynamic)
- Place the robot like in the following picture:
 
![path_inside_lethal](https://github.com/user-attachments/assets/1245eb03-9e7f-42e9-ac27-e6beaddb30be)

Ask to SmacPlanner to compute a path to an accessible goal.

#### Expected behavior
SmacPlanner fails to compute a path or give a path that don't pass by the lethal inflation.

#### Actual behavior
![Screenshot from 2024-10-29 14-22-01](https://github.com/user-attachments/assets/0f2c23c9-3a53-4f48-bf85-d56398fa8754)

#### Additional information
Planner parameters:

```
planner_server:
  ros__parameters:
    expected_planner_frequency: 0.1 # 10s
    planner_plugins: ["SmacPlannerHybrid"]
    SmacPlannerHybrid:
      plugin: "nav2_smac_planner/SmacPlannerHybrid"
      max_iterations: -1
      max_planning_time: 30.0              # max time in s for planner to plan, smooth
      minimum_turning_radius: 1.8         # minimum turning radius in m of path / vehicle
      analytic_expansion_max_length: 2.0
      cost_penalty: 100.0
      use_quadratic_cost_penalty: True
      cache_obstacle_heuristic: True
```

padhupradheep:
I think to debug it better, we might also need your costmap params. 

Any by the way, are you showing the global costmap in the image above?

FrGrQuim:
Thanks for your response.
Yes, this is the global costmap. Here are the costmap parameters:
```
global_costmap:
  global_costmap:
    ros__parameters:
      publish_frequency: 1.0
      update_frequency: 5.0
      global_frame: map
      robot_base_frame: base_link
      robot_radius: 0.9 # 0.7 (real radius) + 0.2 (noise margin)
      resolution: 0.1
      track_unknown_space: False
      use_maximum: true

      plugins: ["low_obstacle_layer", "high_obstacle_layer", "obstacle_filter_layer", "fov_filter_layer", "obstacles_avoidance_inflation_layer", "static_layer", "static_inflation_layer"]
[...]

PR #5049, opened 2025-04-03
Fix incorrect world <-> map coordinates conversions

DylanDeCoeyer-Quimesis:
The conversion between world and map continuous (!) coordinates do not require a +/-0.5. This offset is only required when converting discrete map cell indexes to the coordinates of its center.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4735 |
| Robotic platform tested on | Custom robot |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d.cpp
+++ b/nav2_costmap_2d/src/costmap_2d.cpp
@@ -309,8 +309,8 @@ bool Costmap2D::worldToMapContinuous(double wx, double wy, float & mx, float & m
     return false;
   }
 
-  mx = static_cast<float>((wx - origin_x_) / resolution_) + 0.5f;
-  my = static_cast<float>((wy - origin_y_) / resolution_) + 0.5f;
+  mx = static_cast<float>((wx - origin_x_) / resolution_);
+  my = static_cast<float>((wy - origin_y_) / resolution_);
 
   if (mx < size_x_ && my < size_y_) {
     return true;
--- a/nav2_smac_planner/include/nav2_smac_planner/utils.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/utils.hpp
@@ -45,9 +45,9 @@ inline geometry_msgs::msg::Pose getWorldCoords(
 {
   geometry_msgs::msg::Pose msg;
   msg.position.x =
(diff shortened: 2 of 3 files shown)


SteveMacenski (approved):

SteveMacenski:
CI failed on:

```

/opt/overlay_ws/src/navigation2/nav2_smac_planner/test/test_smoother.cpp:177
Value of: smoother->smooth(plan, costmap, maxtime)
  Actual: true
Expected: false
```

I retriggered in case it was a flaky test, but I don't think we have anything flaky in the smac planner, so I think this may need to be looked at in conjunction with this change

DylanDeCoeyer-Quimesis:
> CI failed on:
> 
> ```
> 
> /opt/overlay_ws/src/navigation2/nav2_smac_planner/test/test_smoother.cpp:177
> Value of: smoother->smooth(plan, costmap, maxtime)
>   Actual: true
> Expected: false
> ```
> 
> I retriggered in case it was a flaky test, but I don't think we have anything flaky in the smac planner, so I think this may need to be looked at in conjunction with this change

@SteveMacenski I investigated the error and ended up thinking that this test makes no sense. From what I understand, the smoother assumes that the path provided as parameter is valid, and **only checks validity on the new smoothed segments that it generates**.

**Details**
An invalid pose is added to the path to be smoothed, but this pose is not coherent (close and approx. aligned) with the end of the path generated by the A* planner.
Figure_1


So, it is ignored by `Smoother::findDirectionalPathSegments` whose purpose is to detect cusp. This way, it is not considered in `Smoother::smoothImpl` where the smoothed path segments validity is evaluated. Finally, this pose is re-added at the end of `Smoother::smooth` whatever if it's valid or invalid.

**Conclusion**
I'm not sure to understand why/how it was working before this PR, but I think that the test should be adapted to
- force the smoother to generate a new segment that crosses a lethal area,
**OR**
- make the area around the path generated by the A* planner lethal to ensure that the smoothed path will be invalid.

Alternatively, we could also make the smoother return a failure if the input path is invalid. In this case, the test could be left unchanged. This is my preferred approach.

SteveMacenski:
Sure thing - I'm happy with the direction you think is best!

SteveMacenski (approved):

SteveMacenski:
Backport to Jazzy in progress

PR #5052, opened 2025-04-04
Mypy nav2 loopback sim

leander-dsouza:
---

## Basic Info

| ------ | ----------- |

---

* Validated linting using pre-commit.
* Tested package using colcon.
* Corrected unsafe fixes using `ruff` - `ruff check --select UP --fix --unsafe-fixes nav2_loopback_sim`

* Restructured `nav2_loopback_sim` for strict mypy.

* N/A

* Tested using pre-commit, colcon testing, and ruff.
---

* N/A

#### For Maintainers: 

### DIFF
--- a/tools/pyproject.toml
+++ b/tools/pyproject.toml
@@ -25,11 +25,18 @@ module = [
     "ament_index_python.*",
     "nav2_common.*",
     "nav2_msgs.*",
+    "nav2_simple_commander.*",
     "launch_testing.*",
     "action_msgs.*",
     "geometry_msgs.*",
     "sensor_msgs.*",
     "tf2_ros.*",
+    "nav_msgs.*",
+    "rosgraph_msgs.*",
+    "tf_transformations.*",
+    "ament_pep257.*",
+    "ament_flake8.*",
+    "ament_copyright.*",
 ]
 
 ignore_missing_imports = true
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,17 +25,6 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: rostooling/setup-ros-docker:ubuntu-noble-ros-rolling-ros-base-latest
-    strategy:
-      matrix:
-        mypy_packages:
-          - "nav2_smac_planner"
-          - "nav2_common"
-          - "nav2_bringup"
-          - "nav2_collision_monitor"
-          - "nav2_costmap_2d"
-          - "opennav_docking"
-          - "nav2_lifecycle_manager"
-          - "nav2_loopback_sim"
     steps:
     - uses: actions/checkout@v4
 
@@ -46,7 +35,15 @@ jobs:
       with:
         linter: mypy
         distribution: rolling
-        package-name: "${{ join(matrix.mypy_packages, ' ') }}"
+        package-name: >-
+          nav2_smac_planner
+          nav2_common
+          nav2_bringup
+          nav2_collision_monitor
+          nav2_costmap_2d
+          opennav_docking
+          nav2_lifecycle_manager
+          nav2_loopback_sim
         arguments: --config tools/pyproject.toml
 
   pre-commit:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,6 +35,7 @@ jobs:
           - "nav2_costmap_2d"
           - "opennav_docking"
           - "nav2_lifecycle_manager"
+          - "nav2_loopback_sim"
     steps:
     - uses: actions/checkout@v4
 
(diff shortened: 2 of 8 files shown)


tools/pyproject.toml
```diff
@@ -25,11 +25,18 @@ module = [
...
     "nav2_common.*",
     "nav2_msgs.*",
+    "nav2_simple_commander.*",
     "launch_testing.*",
     "action_msgs.*",
     "geometry_msgs.*",
     "sensor_msgs.*",
     "tf2_ros.*",
+    "nav_msgs.*",
+    "rosgraph_msgs.*",
```
SteveMacenski, line 35: I'm not sure I understand this list, can you read me in? If this is what we're checking, I Don't think we should be checking / validating all these non-Nav2 packages which could have their own violations separate of Nav2 causing our CI to fail

leander-dsouza: The list ignores the errors relating to the [missing import](https://github.com/ros-navigation/navigation2/blob/main/tools/pyproject.toml#L35) of these packages. We are not validating these non-nav2 packages; instead, we are ignoring them so that the linter can move ahead with no errors.

We are ignoring the following errors:

```bash
nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py:32:1: error: Skipping analyzing "tf2_ros": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

This is what happens if I do not exclude checking for imports regarding the `tf2_ros` package.

SteveMacenski: Why are we skipping any Nav2 packages then? 

```

    "nav2_common.*",
    "nav2_msgs.*",
    "nav2_simple_commander.*",
```

leander-dsouza: As far as I know, it is because the `ament_mypy` linter runs on the unbuilt packages themselves. Hence, it cannot find the installed library of the nav2 packages.

SteveMacenski: Why is that a problem? (forgive the ignorance, I'm learning here) Shouldn't we have the linters passing on the source version?

SteveMacenski:
Woah - we have Nx of these `ament_mypy` jobs instead of one in our CI if you take a look at the expanded set of checks. Why isn't this just 1 mypy check?

leander-dsouza:
> Woah - we have Nx of these `ament_mypy` jobs instead of one in our CI if you take a look at the expanded set of checks. Why isn't this just 1 mypy check?

We can remove the matrix strategy, and have the packages mentioned explicitly.

```yaml
      with:
        linter: mypy
        distribution: rolling
        package-name: "nav2_smac_planner nav2_common nav2_bringup nav2_collision_monitor nav2_costmap_2d opennav_docking nav2_lifecycle_manager nav2_loopback_sim"
        arguments: --config tools/pyproject.toml
 ```

**Comment by SteveMacenski** [comment]
I'd much prefer that! :smile:

**Comment by leander-dsouza** [comment]
> I'd much prefer that! 😄

No worries, do you want me to include the `mypy_packages` in this PR, or should I open a fresh one for it?

File: tools/pyproject.toml (hunk 1 above)
Comment (leander-dsouza) [reply tools/pyproject.toml:35]: I am a novice myself regarding the ament packages :sweat_smile: 

The main issue I believe is that these packages are not built first, in order for `ament_mypy` to find it as a library.

For instance, if you run `ament_mypy` on `nav2_bringup` without building the  `nav2_common` first:

```bash
Regular report continues:
nav2_bringup/launch/tb4_loopback_simulation_launch.py:27:1: error: Cannot find implementation or library stub for module named "nav2_common.launch"  [import-not-found]
```
After building the workspace, you get the resulting message:

```bash
Regular report continues:
Success: no issues found in 16 source files
```

SteveMacenski: Is that because there's not init.py in the workspace root?

leander-dsouza: For instance, in this PR, `nav2_simple_commander` was introduced as a package to ignore when `ament_mypy` fails to find it.

However, there is already an [__init__.py](https://github.com/ros-navigation/navigation2/blob/main/nav2_simple_commander/nav2_simple_commander/__init__.py) present at the python package of `nav2_simple_commander`, but you still get the below error:

```bash
Regular report continues:
nav2_loopback_sim/nav2_loopback_sim/loopback_simulator.py:21:1: error: Cannot find implementation or library stub for module named "nav2_simple_commander.line_iterator"  [import-not-found]
```

leander-dsouza: I might not be quite well-versed with this, maybe @Nils-ChristianIseke can have a look :eyes:

Nils-ChristianIseke: Isn't the problem that there are no type hints for the nav2 packages yet, which are ignored here? We should only ignore imports with missing type hints, not because the modules cannot be found (if this is the case we should setup mypy so that it's able to find the imports). I won't have a chance to have a detailed look, as I am on vacation until 15.04.

SteveMacenski:
We can leave this for now not to block your other PRs for the mypy compliance in other packages, but I'll post on the mypy ticket about this to continue later

Issue #5041, opened 2025-04-02
Callback not executed in IsBatteryCharging BT


ajtudela:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 24.04
- ROS2 Version:
  - Jazzy / rolling
- Version or commit hash:
  - Source
- DDS implementation:
  - Cyclone

##
I have an problem since we moved to Jazzy, but its also happening in Rolling. Both installed from source.

I have a behavior tree where the battery level is checked before trying to undock the robot, similar to this [one](https://github.com/ros-navigation/navigation2/blob/main/nav2_docking/opennav_docking_bt/behavior_trees/application_example.xml): 

But I have to call the NavigateToPose action twice before the robot starts to move. 
- On the first call, IsBatteryCharging returns FALSE, so the other branch of the BT is executed. As the robot is docked, the global result is a failure.
- On the second call, IsBatteryCharging returns SUCCESS, performing the undocking action and moving as usual. 

After checking the code [here](https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp), I noticed that the **batteryCallback** is never executed in the first call, but it is in the second one. I think it's because this **spin_some** doesn't spin the first time.

https://github.com/ros-navigation/navigation2/blob/08268f9922f92511f12ad9b951148d1c1d30096b/nav2_behavior_tree/plugins/condition/is_battery_charging_condition.cpp#L66-L72

After reading other code, I found a similar problem fixed with a double call:
https://github.com/ros-navigation/navigation2/blob/08268f9922f92511f12ad9b951148d1c1d30096b/nav2_behavior_tree/plugins/decorator/goal_updater_node.cpp#L95-L97

I'm not sure if this is fixed or not because putting another **spin_some** in the **IsBatteryCharging** BT fixed the problem.

SteveMacenski:
Linking: https://github.com/ros-navigation/navigation2/issues/4567 we should NOT revert this yet

ajtudela:
> If you still see this, that's a problem. Is your Jazzy/Rolling up to date? 

Yes, I updated this morning.

> If so, can you follow up with a new rclcpp ticket on this and link those PRs that this is still an issue?

Sure!


> I think that should just be applied here as well. Call spin_some twice, the first time with a duration of 1ns.
> - is_battery_charging_condition
> - is_battery_low_condition
> - XYZ_selector_node's

[...]

PR #5055, opened 2025-04-04
Add double spin_some in some BT nodes

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #5041 |

---

* Added a double spin_some in is_battery_charging_condition , is_battery_low_condition and XYZ_selector_node to temporary fix a rclcpp regression.

---

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/plugins/action/controller_selector_node.cpp
+++ b/nav2_behavior_tree/plugins/action/controller_selector_node.cpp
@@ -33,6 +33,9 @@ ControllerSelector::ControllerSelector(
 : BT::SyncActionNode(name, conf)
 {
   initialize();
+
+  // Spin multiple times due to rclcpp regression in Jazzy requiring a 'warm up' spin
+  callback_group_executor_.spin_some(std::chrono::nanoseconds(1));
 }
 
 void ControllerSelector::initialize()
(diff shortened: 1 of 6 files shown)


ajtudela:
This seems to be fixed in [source](https://github.com/ros2/rclcpp/issues/2794#issuecomment-2778487998) but there is no release yet.

On Monday I'll build rclcpp from source and see if it's fixed, but I think this PR should be merged until a new rclcpp binary is released.

SteveMacenski:
Did you validate this fixes your issue?

I was also just thinking: For this and for the existing double-spin node, can we instead have the spin for a short time in the node's constructor instead? That way it doesn't impact normal run-time behavior of the tree's ticking?  :smile:

SteveMacenski (approved):
I approve, but please update me on the 2x topics above!

ajtudela:
> Did you validate this fixes your issue?

Yees, I only tested the is_battery_charging_condition as it is the one I use in my navigator and the other two work the same...

> I was also just thinking: For this and for the existing double-spin node, can we instead have the spin for a short time in the node's constructor instead? That way it doesn't impact normal run-time behavior of the tree's ticking? 😄

Good thinking. I'll move the first spin_some to the constructor and see if it works again.

SteveMacenski (approved):

ajtudela:
@SteveMacenski Could you backport it to jazzy?

SteveMacenski:
Mergify - do yo thang!

SteveMacenski:
@ajtudela it wasn't able to be autobackported due to changes in the node initialization in Rolling. Can you manually backport or handle the merge conflicts? https://github.com/ros-navigation/navigation2/pull/5067

(frankly for how small it is; just manually backporting will probably be faster given how small it is)

ajtudela:
Yeah, no problem. I can backport it manually.

PR #5057, opened 2025-04-04
Publishing a full time-series MPPI trajectory if desirable

SteveMacenski:


tonynajjar:
Out of curiosity, did you ever end up using this topic for visualization or  in some downstream component?

SteveMacenski:
Yes, in a project that someone needed a way to get the future trajectory samples to make decisions based off of due to some rather unique dynamics

PR #5059, opened 2025-04-05
Mypy nav2 simple commander

leander-dsouza:
---

## Basic Info

| ------ | ----------- |

---

* Validated linting using pre-commit.
* Tested package using colcon.
* Corrected unsafe fixes using `ruff` - `ruff check --select UP --fix --unsafe-fixes nav2_simple_commander`

* Restructured `nav2_simple_commander` for strict mypy.

* N/A

* Tested using pre-commit, colcon testing, and ruff.
---

* N/A

#### For Maintainers: 

### DIFF
--- a/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
+++ b/nav2_simple_commander/nav2_simple_commander/robot_navigator.py
@@ -16,7 +16,7 @@
 
 from enum import Enum
 import time
-from typing import Optional
+from typing import Any, Optional
 
 from action_msgs.msg import GoalStatus
 from builtin_interfaces.msg import Duration
@@ -53,10 +53,9 @@ class BasicNavigator(Node):
         super().__init__(node_name=node_name, namespace=namespace)
         self.initial_pose = PoseStamped()
         self.initial_pose.header.frame_id = 'map'
-        self.goal_handle: Optional[ClientGoalHandle[NavigateToPose.Goal, NavigateToPose.Result,
-                                                    NavigateToPose.Feedback]] = None
+        self.goal_handle: Optional[ClientGoalHandle[Any, Any, Any]] = None
         self.result_future: \
-            Optional[Future[GetResultServiceResponse[NavigateToPose.Result]]] = None
+            Optional[Future[GetResultServiceResponse[Any]]] = None
         self.feedback: str = ''
         self.status: Optional[int] = None
         self.last_action_error_code = 0
@@ -500,6 +499,8 @@ class BasicNavigator(Node):
             if self.goal_handle is not None:
                 future = self.goal_handle.cancel_goal_async()
                 rclpy.spin_until_future_complete(self, future)
+            else:
+                self.error('No goal handle to cancel!')
         self.clearTaskError()
         return
 
@@ -523,6 +524,10 @@ class BasicNavigator(Node):
                         f' error code:{result.error_code}'
                         f' error msg:{result.error_msg}'
                     )
+                    return True
+                else:
+                    self.setTaskError(0, 'No result received')
+                    self.debug('Task failed with no result received')
                     return True
         else:
             # Timed out, still processing, not complete yet
@@ -602,20 +607,9 @@ class BasicNavigator(Node):
 
         self.result_future = self.goal_handle.get_result_async()
         rclpy.spin_until_future_complete(self, self.result_future)
+        self.status = self.result_future.result().status  # type: ignore[union-attr]
 
-        result_response = self.result_future.result()
-
-        if result_response is None:
-            self.error('Get path failed!')
-            self.status = GoalStatus.UNKNOWN
-            result = ComputePathToPose.Result()
-            result.error_code = ComputePathToPose.UNKNOWN
-            result.error_msg = 'Get path failed'
-            return result
-
-        self.status = result_response.status
-
-        return result_response.result
+        return self.result_future.result().result  # type: ignore[union-attr]
 
     def getPath(
         self, start: PoseStamped, goal: PoseStamped,
@@ -675,19 +669,9 @@ class BasicNavigator(Node):
 
         self.result_future = self.goal_handle.get_result_async()
         rclpy.spin_until_future_complete(self, self.result_future)
+        self.status = self.result_future.result().status  # type: ignore[union-attr]
 
-        result_response = self.result_future.result()
-
-        if result_response is None:
-            self.error('Get path failed!')
-            result = ComputePathThroughPoses.Result()
-            result.error_code = ComputePathThroughPoses.UNKNOWN
-            result.error_msg = 'Get path failed!'
-            return result
-
-        self.status = result_response.status
-
-        return result_response.result
+        return self.result_future.result().result  # type: ignore[union-attr]
 
     def getPathThroughPoses(
         self, start: PoseStamped, goals: Goals,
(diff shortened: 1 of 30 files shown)


nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -43,14 +49,16 @@ class TaskResult(Enum):
...
-    def __init__(self, node_name='basic_navigator', namespace=''):
+    def __init__(self, node_name: str = 'basic_navigator', namespace: str = '') -> None:
         super().__init__(node_name=node_name, namespace=namespace)
         self.initial_pose = PoseStamped()
         self.initial_pose.header.frame_id = 'map'
-        self.goal_handle = None
-        self.result_future = None
-        self.feedback = None
-        self.status = None
+        self.goal_handle: Optional[ClientGoalHandle[NavigateToPose.Goal, NavigateToPose.Result,
```
SteveMacenski, line 56: These aren't limited to navigate to pose types, its also path following, behavior execution, etc

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -479,12 +493,13 @@ def undockRobot(self, dock_type=''):
...
         return True
 
-    def cancelTask(self):
+    def cancelTask(self) -> None:
         """Cancel pending task request of any type."""
         self.info('Canceling current task.')
         if self.result_future:
-            future = self.goal_handle.cancel_goal_async()
-            rclpy.spin_until_future_complete(self, future)
+            if self.goal_handle is not None:
```
SteveMacenski, line 500: Log error if `else`

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -494,28 +509,33 @@ def isTaskComplete(self) -> bool:
...
             if self.status != GoalStatus.STATUS_SUCCEEDED:
-                result = self.result_future.result().result
-                self.setTaskError(result.error_code, result.error_msg)
-                self.debug('Task with failed with'
-                           f' status code:{self.status}'
-                           f' error code:{result.error_code}'
-                           f' error msg:{result.error_msg}')
-                return True
+                result = result_response.result
+                if result is not None:
```
SteveMacenski, line 518: Else log errro

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -580,11 +602,24 @@ def _getPathImpl(
 
         self.result_future = self.goal_handle.get_result_async()
         rclpy.spin_until_future_complete(self, self.result_future)
-        self.status = self.result_future.result().status
 
-        return self.result_future.result().result
+        result_response = self.result_future.result()
+
+        if result_response is None:
```
SteveMacenski, line 608: This spins without timeout, so we know we have a response if this completes. I think these changes can be reverted

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -639,11 +675,23 @@ def _getPathThroughPosesImpl(
...
```
SteveMacenski, line 681: This spins without timeout, so we know we have a response if this completes. I think these changes can be reverted

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -690,13 +739,23 @@ def _smoothPathImpl(
...
```
SteveMacenski, line 745: This spins without timeout, so we know we have a response if this completes. I think these changes can be reverted

SteveMacenski (changes requested):
Thanks for this, I know this one is probably the most effort. There's a few instances where the server is short and not going to be spinning multiple times (get paths, smooth paths) where the changes would create dead code that I think can be reverted. I know it mimicks the style of other long-running tasks (follow path, navigate to pose) but these don't long-blocking tasks with timeouts on the `spin`

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
leander-dsouza: I reverted the change and added a skip in type hint checking with `ignore[union-attr]`.

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
leander-dsouza: I reverted the change and added a skip in type hint checking with `ignore[union-attr]`.

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
leander-dsouza: I reverted the change and added a skip in type hint checking with `ignore[union-attr]`.

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
leander-dsouza: I have declared them as Any instead.

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
leander-dsouza: Resolved

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
leander-dsouza: Resolved

leander-dsouza:
> Thanks for this, I know this one is probably the most effort. 

No worries, Steve. I am glad that I could help.

nav2_simple_commander/nav2_simple_commander/robot_navigator.py
```diff
@@ -43,14 +49,15 @@ class TaskResult(Enum):
...
         super().__init__(node_name=node_name, namespace=namespace)
         self.initial_pose = PoseStamped()
         self.initial_pose.header.frame_id = 'map'
-        self.goal_handle = None
-        self.result_future = None
-        self.feedback = None
-        self.status = None
+        self.goal_handle: Optional[ClientGoalHandle[Any, Any, Any]] = None
+        self.result_future: \
+            Optional[Future[GetResultServiceResponse[NavigateToPose.Result]]] = None
```
SteveMacenski, line 58: This can also not be just for `NavigateToPose`

SteveMacenski (commented):
I also have this PR open that messes with the simple commander API (https://github.com/ros-navigation/navigation2/pull/5056). I change a few things like None / `TaskEnum` for actions so we know what action we should be getting task completion or feedback for. 

I also have `getRoute(self, start, goal, use_start=False)` where start/goal can be either be `int` or `PoseStamped`. 

Is there a good way around these?  

For reference: https://github.com/ros-navigation/navigation2/blob/bc79b6b07f3a8ce89579682265c88cdd8174fcb9/nav2_simple_commander/nav2_simple_commander/robot_navigator.py

leander-dsouza:
> I also have this PR open that messes with the simple commander API (#5056). I change a few things like None / `TaskEnum` for actions so we know what action we should be getting task completion or feedback for.
> 
> I also have `getRoute(self, start, goal, use_start=False)` where start/goal can be either be `int` or `PoseStamped`.
> 
> Is there a good way around these?

* I think after rebasing this PR, you can first check for conflicting types by:

   ```bash
   ament_mypy --config tools/pyproject.toml nav2_simple_commander/
   ```
* For declaring the type of start/goal, you can use `Union`:

   ```python
   from typing import Optional, Union
   from nav_msgs.msg import Path
   from nav2_msgs.msg import Route

   def getRoute(self, start: Union[int, PoseStamped],
                goal: Union[int, PoseStamped], use_start: bool = False) \
           -> Optional[list[Union[Path, Route]]]:
       """Send a `ComputeRoute` action request."""
       self.clearTaskError()
       rtn = self._getRouteImpl(start, goal, use_start=False)

       if self.status != GoalStatus.STATUS_SUCCEEDED:
           self.setTaskError(rtn.error_code, rtn.error_msg)
           self.warn('Getting route failed with'
                       f' status code:{self.status}'
                       f' error code:{rtn.error_code}'
                       f' error msg:{rtn.error_msg}')
           return None

       return [rtn.path, rtn.route]
   ```

SteveMacenski (approved):

SteveMacenski:
Sweet, thanks!

PR #5071, opened 2025-04-09
Update map_io library to use Eigen method for faster map loading

vickzk:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | -- |

---

* Replaced the Magick-based image loading and processing in the Map IO library with OpenCV.
* Replaced per-pixel loops with OpenCV matrix operations for faster processing

* Tested extensively with various maps
* Tested with all the 3 modes of maps (Trinary, Scale and Raw)
* Tested for PNG and PGM format maps
* Tested with larger maps that used to take 10 seconds to load previously, loaded within 2 seconds now

---

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/package.xml
+++ b/nav2_map_server/package.xml
@@ -26,6 +26,7 @@
   <depend>nav2_util</depend>
   <depend>graphicsmagick</depend>
   <depend>lifecycle_msgs</depend>
+   <depend>libopencv-dev</depend>
 
   <test_depend>ament_lint_common</test_depend>
   <test_depend>ament_lint_auto</test_depend>
--- a/nav2_map_server/package.xml
+++ b/nav2_map_server/package.xml
@@ -26,7 +26,6 @@
   <depend>nav2_util</depend>
   <depend>graphicsmagick</depend>
   <depend>lifecycle_msgs</depend>
-   <depend>libopencv-dev</depend>
 
   <test_depend>ament_lint_common</test_depend>
   <test_depend>ament_lint_auto</test_depend>
--- a/nav2_map_server/src/map_io.cpp
+++ b/nav2_map_server/src/map_io.cpp
@@ -243,18 +243,7 @@ void loadMapFromFile(
   Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
   gray_matrix(buffer.data(), height, width);
 
-  // Prepare alpha (transparency) matrix, if present
   bool has_alpha = img.matte();
-  Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> alpha_matrix;
-
-  if (has_alpha) {
-    std::vector<uint8_t> alpha_buf(width * height);
-    img.write(0, 0, width, height, "A", Magick::CharPixel, alpha_buf.data());
-
-    alpha_matrix = Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic,
-        Eigen::RowMajor>>(
-      alpha_buf.data(), height, width);
-  }
 
   // Handle different map modes with if else condition
   // Trinary and Scale modes are handled together
@@ -306,8 +295,16 @@ void loadMapFromFile(
 
     // Apply alpha transparency mask: mark transparent cells as UNKNOWN
     if (has_alpha) {
-      auto transparent_mask = (alpha_matrix.array() < 255);
-      result = transparent_mask.select(nav2_util::OCC_GRID_UNKNOWN, result);
+      // Allocate buffer only once and map directly to Eigen without extra copy
+      std::vector<uint8_t> alpha_buf(width * height);
+      img.write(0, 0, width, height, "A", Magick::CharPixel, alpha_buf.data());
+    
+      // Map alpha buffer as Eigen::Array for efficient elementwise ops
+      Eigen::Map<Eigen::Array<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>> alpha_array(
+        alpha_buf.data(), height, width);
+    
+      // Apply mask directly with Eigen::select
+      result = (alpha_array < 255).select(nav2_util::OCC_GRID_UNKNOWN, result);
     }
 
     // Flip image vertically (as ROS expects origin at bottom-left)
--- a/nav2_map_server/src/map_io.cpp
+++ b/nav2_map_server/src/map_io.cpp
@@ -238,81 +238,76 @@ void loadMapFromFile(
   std::vector<uint8_t> buffer(width * height);
   gray.write(0, 0, width, height, "I", Magick::CharPixel, buffer.data());
 
-  Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>> 
-      gray_matrix(buffer.data(), height, width);
-  
-  bool has_alpha = img.matte(); 
+  Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+  gray_matrix(buffer.data(), height, width);
+
+  bool has_alpha = img.matte();
   Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> alpha_matrix;
 
   if (has_alpha) {
     std::vector<uint8_t> alpha_buf(width * height);
     img.write(0, 0, width, height, "A", Magick::CharPixel, alpha_buf.data());
 
-    Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>> 
-        alpha(alpha_buf.data(), height, width);
+    Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+    alpha(alpha_buf.data(), height, width);
 
     alpha_matrix = alpha;
   }
 
-  if (load_parameters.mode == MapMode::Trinary || load_parameters.mode == MapMode::Scale)
-  {
-    Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> normalized = gray_matrix.cast<float>() / 255.0f;
+  if (load_parameters.mode == MapMode::Trinary || load_parameters.mode == MapMode::Scale) {
+    Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic,
+      Eigen::RowMajor> normalized = gray_matrix.cast<float>() / 255.0f;
     if (!load_parameters.negate) {
       normalized = (1.0f - normalized.array()).matrix();
     }
-    
+
     Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> occupied =
       (normalized.array() >= load_parameters.occupied_thresh).cast<uint8_t>();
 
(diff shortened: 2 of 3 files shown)


nav2_map_server/package.xml
```diff
@@ -26,6 +26,7 @@
   <depend>nav2_util</depend>
   <depend>graphicsmagick</depend>
   <depend>lifecycle_msgs</depend>
+   <depend>libopencv-dev</depend>
```
SteveMacenski, line 29:
```suggestion
  <depend>libopencv-dev</depend>
```

SteveMacenski:
OpenCV is a very heavy dependency to introduce, I'm not sure I'd ideally like to do that if we can improve the existing implementation instead. Can you identify where the compute time is spent with the current implementation and if that can be improved instead?

vickzk:
mapIO_magick
I’ve shared the image showing the Magick image load and highlighted how the nested for loop is the most time-consuming part. I’ve tested it with the largest map so far. As you suggested, I’ll now look into optimizing this section without using OpenCV and will keep you updated.

Thanks for re kicking the CI :heart:

nav2_map_server/src/map_io.cpp
```diff
@@ -225,70 +228,88 @@ void loadMapFromFile(
...
-            } else {
-              map_cell = nav2_util::OCC_GRID_UNKNOWN;
-            }
-            break;
-          }
-        default:
-          throw std::runtime_error("Invalid map mode");
+  // Convert the image to grayscale
+  Magick::Image gray = img;
+  gray.type(Magick::GrayscaleType);
```
SteveMacenski, line 233: What does this operation specifically do? Is it exactly analog to https://github.com/ros-navigation/navigation2/blob/main/nav2_map_server/src/map_io.cpp#L233-L245 or are there changes that are important to acknowledge?

nav2_map_server/src/map_io.cpp
```diff
@@ -225,70 +228,88 @@ void loadMapFromFile(
...
+  Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> alpha_matrix;
+
+  if (has_alpha) {
+    std::vector<uint8_t> alpha_buf(width * height);
+    img.write(0, 0, width, height, "A", Magick::CharPixel, alpha_buf.data());
+
+    Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+    alpha(alpha_buf.data(), height, width);
+
+    alpha_matrix = alpha;
```
SteveMacenski, line 255: Can `alpha_matrix` just be populated with `alpha_buf` without going through the intermediary?

Ex something like `alpha_matrix(alpha_buf.data(), height, width))`?

nav2_map_server/src/map_io.cpp
```diff
@@ -225,70 +228,88 @@ void loadMapFromFile(
...
+      gray_matrix.cast<int8_t>();
+
+    auto out_of_bounds = (result.array() < nav2_util::OCC_GRID_FREE) ||
+      (result.array() > nav2_util::OCC_GRID_OCCUPIED);
+
+    result = out_of_bounds.select(nav2_util::OCC_GRID_UNKNOWN, result);
+
+    Eigen::Matrix<int8_t, Eigen::Dynamic, Eigen::Dynamic,
+      Eigen::RowMajor> flipped = result.colwise().reverse();
+    std::memcpy(msg.data.data(), flipped.data(), width * height);
```
SteveMacenski, line 310: Is this just due to the difference in column vs row major storage?

nav2_map_server/src/map_io.cpp
```diff
@@ -225,70 +228,88 @@ void loadMapFromFile(
...
+    if (has_alpha) {
+      auto transparent_mask = (alpha_matrix.array() < 255);
+      result = transparent_mask.select(nav2_util::OCC_GRID_UNKNOWN, result);
+    }
+
+    Eigen::Matrix<int8_t, Eigen::Dynamic, Eigen::Dynamic,
+      Eigen::RowMajor> flipped = result.colwise().reverse();
+    std::memcpy(msg.data.data(), flipped.data(), width * height);
+  } else if (load_parameters.mode == MapMode::Raw) {
+    Eigen::Matrix<int8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> result =
```
SteveMacenski, line 300: Don't these need to be `uint8` for `0-255` versus `-128-128`

nav2_map_server/src/map_io.cpp
vickzk: The `gray.type(Magick::GrayscaleType)` operation is not exactly the same as a manual RGB average — it uses a **weighted luminance formula** that is generally superior for preserving features and is more performant. This method is ideal for non-color maps, such as typical occupancy grids. However, if the map encodes information using specific colors (e.g., RGB layers with semantic meaning), it's important to evaluate whether grayscale conversion is appropriate for the intended use case.

nav2_map_server/src/map_io.cpp
vickzk: Yes, functionally it's doing the same vertical flip — it's just done manually with index math during assignment in the old approach: https://github.com/ros-navigation/navigation2/blob/1a365e66444b7b45f85707c0fcbc1b2acb40367a/nav2_map_server/src/map_io.cpp#L290C7-L290C75
Now it's handled in one step using `colwise().reverse()` on the entire matrix.

nav2_map_server/src/map_io.cpp
vickzk: Good point, the source grayscale image data is indeed in the `uint8_t` range (0–255). But since `nav_msgs/OccupancyGrid` expects `int8[]` data, we cast to `int8_t`. The values wrap correctly (e.g., 255 → -1) to follow the ROS convention (`-1` = unknown). So while `uint8_t` is fine for buffers, the final matrix must be `int8_t` for compatibility.

nav2_map_server/src/map_io.cpp
vickzk: Yeah, this one got updated!

nav2_map_server/src/map_io.cpp
SteveMacenski: Keep in mind that the `OccupancyGrid` message is from 0-100 with -1 being unknown, it is not the same as the 0-255 range for images or costmaps. `int8` in the map server computation while handling image data will threshold values to `127` when we want the full range until we start having results populate the costmaps

nav2_map_server/src/map_io.cpp
SteveMacenski: OK - how did you test this to make sure the parsing is 1:1 for each of the settings? Is the existing unit test coverage sufficient to prove that or do you need to add some additional unit testing to make sure these changes don't break existing behaviors?

vickzk: I've created the **maps** using `map_saver` in the following modes:  
- Raw  
- Scale  
- Trinary  

For **each of the two file formats**:  
- PGM  
- PNG  

This results in a total of 6 maps. I used both the `old_map_server` and `new_map_server` to load these maps and visually verified them across all modes and file formats.  

After that, I wrote a script to save the output occupancy grid data to a file, followed by another script to compare the files. The result from that comparison is shown below.  

I haven't checked the unit test case yet — will do that next.


| New Map Server                              | Old Map Server                                | Match   |
|-------------------------------------------------|-------------------------------------------------|---------|
| new_map_server_pgm_raw.json             | old_map_server_pgm_raw.json             | True    |
| new_map_server_pgm_scale.json         | old_map_server_pgm_scale.json         | True    |
| new_map_server_pgm_trinary.json     | old_map_server_pgm_trinary.json     | True    |
| new_map_server_png_raw.json         | old_map_server_png_raw.json         | True    |
| new_map_server_png_trinary.json | old_map_server_png_trinary.json | True    |
| new_map_server_png_scale.json     | old_map_server_png_scale.json     | True    |

nav2_map_server/src/map_io.cpp
vickzk: The **raw map** saved by `map_saver` already contains occupancy values:
- 0 to 100 for free and occupied areas.
- 255 for unknown areas.

No other values are present in the map, and it is **directly storing these values as the occupancy map** as per the `map saving` part. 
https://github.com/ros-navigation/navigation2/blob/5423eba732ed16b4f97a7353027c1f444b6edc60/nav2_map_server/src/map_io.cpp#L521-L529

So only the out-of-bounds values (in this case, the 255 values) will be replaced with -1 (UNKNOWN).
All other valid occupancy values (between 0 and 100) will remain as they are, directly taken from the file.

nav2_map_server/src/map_io.cpp
SteveMacenski: That is really helpful, thank you!

nav2_map_server/src/map_io.cpp
```diff
@@ -225,70 +228,110 @@ void loadMapFromFile(
...
+  // Map the grayscale buffer to an Eigen matrix (row-major layout)
+  Eigen::Map<Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+  gray_matrix(buffer.data(), height, width);
+
+  // Prepare alpha (transparency) matrix, if present
+  bool has_alpha = img.matte();
+  Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> alpha_matrix;
+
+  if (has_alpha) {
+    std::vector<uint8_t> alpha_buf(width * height);
```
SteveMacenski, line 251: The alpha_matrix is only used in 1 place. Can we create the alpha matrix, populate it, and use it in that one place only?

SteveMacenski (changes requested):
Otherwise LGTM!

nav2_map_server/src/map_io.cpp
vickzk: Done

nav2_map_server/src/map_io.cpp
```diff
@@ -225,70 +228,108 @@ void loadMapFromFile(
...
+    Eigen::Matrix<int8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> result =
+      gray_matrix.cast<int8_t>();
+
+      // Clamp out-of-bound values (outside [-1, 100]) to UNKNOWN (-1)
+    auto out_of_bounds = (result.array() < nav2_util::OCC_GRID_FREE) ||
+      (result.array() > nav2_util::OCC_GRID_OCCUPIED);
+
+    result = out_of_bounds.select(nav2_util::OCC_GRID_UNKNOWN, result);
+
+      // Flip image vertically (as ROS expects origin at bottom-left)
```
josephduchesne, line 326: Minor nitpick:

Rather than 
```
if (Trinary || Scale) {
   // ternary, scale stuff
   // Flip image vertically (as ROS expects origin at bottom-left)
} else if (Raw) {
  // raw stuff
   // Flip image vertically (as ROS expects origin at bottom-left)
} else {
  // throw
}
```

Can we do:

```
if (Trinary || Scale) {
   // ternary, scale stuff
   // Flip image vertically (as ROS expects origin at bottom-left)
} else if (Raw) {
  // raw stuff
} else {
  // throw
}
// Flip image vertically (as ROS expects origin at bottom-left)
```

This wins at code golf and doesn't sacrifice readibility much. You'd have to move the result allocation out of the if statement, which does mean invalid map modes would have an extra allocation, but I don't think that's a performance critical code path.

Alternatively, the two lines two flip could be an inline function, or things are fine as they are? 

The code seems pretty tight to me overall.

josephduchesne (approved):
I'm more familiar with opencv's cv::Mat operations (which are extremely performant) and was skeptical about Eigen as a replacement but I stand corrected. Nice work!

nav2_map_server/src/map_io.cpp
vickzk: Good catch on the duplication, I’ve updated the code to move the **flipping** and `memcpy` outside the `if-else` block. It’s cleaner now without repeating that part. Thanks for the suggestion!

PR #5075, opened 2025-04-10
Lock costmap in getCostsCallback

tonynajjar:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Custom simulation|

---

* Lock costmap before getting cost

```
ros2 service call /global_costmap/get_cost_global_costmap nav2_msgs/srv/GetCosts "{
  use_footprint: true,
  poses: [
    {
      header: {
        frame_id: 'map'
      },
      pose: {
        position: {x: 40.62, y: 26.08, z: 0.0},
        orientation: {x: 0.0, y: 0.0, z: 0.0, w: 1.0}
      }
    }
  ]
}"
```

I ran this command multiple times on a spot with a cost of 254. Before the fix, I would occasionally get 0, after the fix, I always got 254 

---

#### For Maintainers: 


nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -831,6 +831,7 @@ void Costmap2DROS::getCostsCallback(
   unsigned int mx, my;
 
   Costmap2D * costmap = layered_costmap_->getCostmap();
+  std::unique_lock<Costmap2D::mutex_t> lock(*(costmap->getMutex()));
```
tonynajjar, line 834: One option would also be to lock and unlock closer to where the costmap is queried to reduce the time it's locked but I don't know if that time difference is really significant

SteveMacenski: i think this is fine. If we wanted to optimize for that here, what we could do is restructure the loop to do all the TF calls first in batch, then lock and loop over the footprint cost getting. TF is the big operation that would be wasted with respect to the costmap locking

nav2_costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -875,6 +876,7 @@ void Costmap2DROS::getCostsCallback(
       response->costs.push_back(static_cast<float>(costmap->getCost(mx, my)));
     }
   }
+  lock.unlock();
```
SteveMacenski, line 879: I don't think this is explicitly necessary. When the lock goes out of scope it unlocks

SteveMacenski (changes requested):

nav2_costmap_2d/src/costmap_2d_ros.cpp
tonynajjar: Okay, for lack of time right now, will skip that

SteveMacenski (approved):

PR #5076, opened 2025-04-10
[AMCL] Rebuild cspace map only when needed

doisyg:
---

## Basic Info

| ------ | ----------- |

---

Recompute cspace map only when needed

Ran on Dexory robot in Dexory warehouse for several days

---

#### For Maintainers: 

### DIFF
--- a/nav2_amcl/src/sensors/laser/likelihood_field_model.cpp
+++ b/nav2_amcl/src/sensors/laser/likelihood_field_model.cpp
@@ -35,7 +35,14 @@ LikelihoodFieldModel::LikelihoodFieldModel(
   z_hit_ = z_hit;
   z_rand_ = z_rand;
   sigma_hit_ = sigma_hit;
-  map_update_cspace(map, max_occ_dist);
+
+  // recompute cpace only when nescessary, i.e. if:
+  // - max_occ_dist changed
+  // OR
+  // - cspace was not computed yet, i.e. when map->max_occ_dist == 0.0
+  if (map->max_occ_dist != max_occ_dist || map->max_occ_dist == 0.0) {
+    map_update_cspace(map, max_occ_dist);
+  }
 }
 
 double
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
@mikeferguson any objections?

SteveMacenski:
The new precommit job failed for both:

```

-  // recompute cpace only when nescessary, i.e. if:
+  // recompute cpace only when necessary, i.e. if:
```

nav2_amcl/src/sensors/laser/likelihood_field_model.cpp
```diff
@@ -35,7 +35,14 @@ LikelihoodFieldModel::LikelihoodFieldModel(
   z_hit_ = z_hit;
   z_rand_ = z_rand;
   sigma_hit_ = sigma_hit;
-  map_update_cspace(map, max_occ_dist);
+
+  // recompute cpace only when nescessary, i.e. if:
```
mikeferguson, line 39: typo - cspace is missing the "s"

nav2_amcl/src/sensors/laser/likelihood_field_model.cpp
```diff
@@ -35,7 +35,14 @@ LikelihoodFieldModel::LikelihoodFieldModel(
...
+  // - max_occ_dist changed
+  // OR
+  // - cspace was not computed yet, i.e. when map->max_occ_dist == 0.0
+  if (map->max_occ_dist != max_occ_dist || map->max_occ_dist == 0.0) {
```
mikeferguson, line 43: if we're going to rely on "== 0.0" - we should probably update map_alloc() (in map.c) to explicitly set max_occ_dist to 0.0 (it doesn't look like that currently happens).

Also, is 0.0 even in a valid max_occ_dist to be passed in here? It seems like not, in which case the first condition will cover us as long as we make sure an uninitialized map has 0.0 (which would be covered by updating map_alloc() in map.c)

mikeferguson (commented):
General idea seems ok - but I'm concerned about uninitialized memory possibly leading to wonky outputs - see my second comment for suggested improvement to map_alloc() and possibly also the removal of the second condition (since, the first one would suffice)

mikeferguson (approved):

Issue #4695, opened 2024-09-27
Docking backwards as plugin param


cgasconb:
## Feature request
The dock_backwards parameter as a dock_plugin parameter

#### Feature description
The dock_backwards param is a general param of the node, so it is fixed once the node starts. This way, the dock_plugins and dock instances are always forward or backward. In the case of needing a dock_plugin as forward and another one as backward, it is necessary to start 2 nodes changing this parameter.
As this param is called each time the velocity is calculated, it would be great if this param is part of the dock_plugin params, so we can have 2 plugins for forward and backward in one node.

SteveMacenski:
This makes sense and you are not the first to bring this up, so I think this is worthwhile.

I took a look and it appears that most of the uses of `dock_backwards_` are trivial to replace with a `dock->plugin->dockDirection()` or so forth. Then in the simple docking plugins, I can make the `dock_direction` a parameter so you can specify for each plugin instance which to use. Thus, you could have a forward and reverse plugin instance within a single server instance. 

A couple of APIs would need to change (`resetApproach` to pass in the direction or the current plugin; move the parameter off the server) but nothing technically complicated.

I would propose the following:
- The dock plugin base class has a `bool dockForward()` API. Optionally we could make the return type an enum with forward and reverse as current implementations, but leave room for more in the future (should there be lateral docking uses for omni robots)
- At the start of `dockRobot` and `undockRobot`, obtain this in a local `dock_direction` parameter which can replace all uses of `dock_backwards_`. Pass this into `resetApproach` as well.

I think that should work. Then the open question I have is: 
- How should we / should we support existing uses of `dock_backwards_`? 

We could still get the parameter and log a warning that this is depreciated and to use the new dock plugin API. However in that case if we get it, we need to deconflict it when we use it in  `dockRobot` and `undockRobot`. 

Alternatively, we could have the simple plugins have their own `<dock name>.dock_direction` parameter, and if that's not set, look for the global `dock_backwards` parameter instead to use. Then also log that its depreciated and use `<dock name>.dock_direction` instead. 

I think the latter is cleaner, but works primarily for users of the simple charging / non-charging plugins provided. I suppose that is 'fine' as long as we don't backport these changes and only available in Rolling and newer. 

What do you think? Interest in contributing this or something I should work on on the way to ROSCon?

PR #5079, opened 2025-04-11
Docking backwards as plugin param

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4695 |

---

* Added a new enum type ```DockDirection``` to indicate the direction of the docking.
* Added ```getDockDirection```to ```ChargingDock``` base class that return the direction.
* Added new ```dock_direction``` parameter to the ```SimpleChargingDock``` and ```SimpleNonChargingDock``` plugins.
* Set old ```dock_backwards```parameter as an optional, log warning and deconfliction.

---

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/params/nav2_params.yaml
+++ b/nav2_bringup/params/nav2_params.yaml
@@ -439,7 +439,6 @@ docking_server:
     max_retries: 3
     base_frame: "base_link"
     fixed_frame: "odom"
-    dock_backwards: false
     dock_prestaging_tolerance: 0.5
     service_introspection_mode: "disabled"
 
@@ -452,6 +451,7 @@ docking_server:
       use_external_detection_pose: true
       use_battery_status: false # true
       use_stall_detection: false # true
+      dock_direction: "forward"
 
       external_detection_timeout: 1.0
       external_detection_translation_x: -0.18
--- a/nav2_docking/README.md
+++ b/nav2_docking/README.md
@@ -207,7 +207,6 @@ For debugging purposes, there are several publishers which can be used with RVIZ
 | max_retries        | Maximum number of retries to attempt    | int |  3      |
 | base_frame        | Robot's base frame for control law   | string |  "base_link"      |
 | fixed_frame        | Fixed frame to use, recommended to be a smooth odometry frame **not** map   | string |  "odom"      |
-| dock_backwards        | Whether the robot is docking with the dock forward or backward in motion | bool |  false      |
 | dock_prestaging_tolerance  |  L2 distance in X,Y,Theta from the staging pose to bypass navigation | double |  0.5      |
 | dock_plugins  | A set of dock plugins to load | vector<string> |  N/A      |
 | dock_database  |  The filepath to the dock database to use for this environment | string |  N/A  |
@@ -251,6 +250,7 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
 | docking_threshold        | If not using stall detection, the pose threshold to the docking pose where `isDocked() = true`    | double |  0.05     |
 | staging_x_offset        | Staging pose offset forward (negative) of dock pose (m)    | double |  -0.7    |
 | staging_yaw_offset        | Staging pose angle relative to dock pose (rad)    | double |  0.0    |
+| dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"      |
 
 Note: The external detection rotation angles are setup to work out of the box with Apriltags detectors in `image_proc` and `isaac_ros`.
 
--- a/nav2_docking/README.md
+++ b/nav2_docking/README.md
@@ -250,7 +250,7 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
 | docking_threshold        | If not using stall detection, the pose threshold to the docking pose where `isDocked() = true`    | double |  0.05     |
 | staging_x_offset        | Staging pose offset forward (negative) of dock pose (m)    | double |  -0.7    |
 | staging_yaw_offset        | Staging pose angle relative to dock pose (rad)    | double |  0.0    |
-| dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"      |
+| dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"  or "backward"    |
 
 Note: The external detection rotation angles are setup to work out of the box with Apriltags detectors in `image_proc` and `isaac_ros`.
 
--- a/nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
+++ b/nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
@@ -187,9 +187,13 @@ inline double l2Norm(const geometry_msgs::msg::Pose & a, const geometry_msgs::ms
 
 inline DockDirection getDockDirectionFromString(const std::string & direction)
 {
-  if (direction == "forward") {
+  auto upper_direction = direction;
+  std::transform(
+    upper_direction.begin(), upper_direction.end(), upper_direction.begin(), ::toupper);
+
+  if (upper_direction == "FORWARD") {
     return DockDirection::FORWARD;
-  } else if (direction == "backward") {
+  } else if (upper_direction == "BACKWARD") {
     return DockDirection::BACKWARD;
   } else {
     return DockDirection::UNKNOWN;
--- a/nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
+++ b/nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
@@ -24,6 +24,7 @@
 #include "nav2_util/geometry_utils.hpp"
 #include "angles/angles.h"
 #include "opennav_docking/types.hpp"
+#include "opennav_docking_core/charging_dock.hpp"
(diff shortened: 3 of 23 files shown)


nav2_bringup/params/nav2_params.yaml
```diff
@@ -452,6 +451,7 @@ docking_server:
       use_external_detection_pose: true
       use_battery_status: false # true
       use_stall_detection: false # true
+      dock_direction: "forward"
```
SteveMacenski, line 454: I think we need to make sure to update the configuration guide that this is now a plugin param, not server.

Also, I think the migration guide needs this mentioned as well since its subtle. 

Can we also check for the `dock_backwards` parameter on the server to see if its declared? If so, log a warning at this use is depreciated and use it as part of the plugins

nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
```diff
@@ -183,6 +185,17 @@ inline double l2Norm(const geometry_msgs::msg::Pose & a, const geometry_msgs::ms
     delta_angle * delta_angle);
 }
 
+inline DockDirection getDockDirectionFromString(const std::string & direction)
```
SteveMacenski, line 188: Why not just have this be a boolean so there doesn't need to be a parser?

nav2_docking/opennav_docking_core/include/opennav_docking_core/charging_dock.hpp
```diff
@@ -123,6 +129,13 @@ class ChargingDock
    */
   virtual bool isCharger() {return true;}
 
+  /**
+   * @brief Indicates the direction of the dock. This is used to determine if the
+   * robot should drive forwards or backwards onto the dock.
+   * @return DockDirection The direction of the dock
+   */
+  virtual DockDirection getDockDirection() {return DockDirection::FORWARD;}
```
SteveMacenski, line 137: This could simply be a boolean for forward/back. Do you see other cases we'd want to add?

SteveMacenski (changes requested):
Generally looks good to me! I appreciate the deprecation warning message and handling, I think that's important for this case since its subtle

nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
ajtudela: Yes, with a ```bool``` it would be much simpler, but I thought that since we are going to modify it, we can leave the door open for future community plugins. For example, a robot that has omnidirectional wheels and wants to dock laterally.

ajtudela:
Doc here: https://github.com/ros-navigation/docs.nav2.org/pull/678

nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
SteveMacenski: Eh ok. I typed out a whole thing but we can go with this :-)

nav2_docking/README.md
```diff
@@ -251,6 +250,7 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
 | docking_threshold        | If not using stall detection, the pose threshold to the docking pose where `isDocked() = true`    | double |  0.05     |
 | staging_x_offset        | Staging pose offset forward (negative) of dock pose (m)    | double |  -0.7    |
 | staging_yaw_offset        | Staging pose angle relative to dock pose (rad)    | double |  0.0    |
+| dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"      |
```
SteveMacenski, line 253:
```suggestion
| dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"  or "backward"    |
```

nav2_docking/opennav_docking_core/include/opennav_docking_core/non_charging_dock.hpp
```diff
@@ -129,6 +129,13 @@ class NonChargingDock : public ChargingDock
    * @brief Gets if this is a charging-typed dock
    */
   bool isCharger() final {return false;}
+
+  /**
+   * @brief Indicates the direction of the dock. This is used to determine if the
+   * robot should drive forwards or backwards onto the dock.
+   * @return DockDirection The direction of the dock
+   */
+  virtual DockDirection getDockDirection() = 0;
```
SteveMacenski, line 138: Why not default to forward as well, as ChargingDock does? I would think we should at least make them consistent, one way or another

nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
```diff
@@ -268,6 +274,11 @@ void SimpleNonChargingDock::jointStateCallback(const sensor_msgs::msg::JointStat
   is_stalled_ = (velocity < stall_velocity_threshold_) && (effort > stall_effort_threshold_);
 }
 
+opennav_docking_core::DockDirection SimpleNonChargingDock::getDockDirection()
+{
+  return utils::getDockDirectionFromString(dock_direction_);
```
SteveMacenski, line 279: Why not set `dock_direction_` as a `DockDirection` type and return it here instead? That way we don't repeatedly need to call `getDockDirectionFromString` and can check the validity of the string on initialization in `configure` so that if its invalid, we can throw an exception or fail configuring?

Ditto for the charging plugin

nav2_docking/opennav_docking/include/opennav_docking/utils.hpp
```diff
@@ -183,6 +185,17 @@ inline double l2Norm(const geometry_msgs::msg::Pose & a, const geometry_msgs::ms
...
+{
+  if (direction == "forward") {
```
SteveMacenski, line 190: You can use `toupper` in order to convert a string to all upper case. Can you convert the input to all upper and then compare against an all upper version of forward/backward? That way if users define it in any medium of caps, it'll evaluate properly?

SteveMacenski (changes requested):

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -98,6 +103,13 @@ void SimpleChargingDock::configure(
   node_->get_parameter(name + ".staging_x_offset", staging_x_offset_);
   node_->get_parameter(name + ".staging_yaw_offset", staging_yaw_offset_);
 
+  std::string dock_direction;
+  node_->get_parameter(name + ".dock_direction", dock_direction);
+  dock_direction_ = utils::getDockDirectionFromString(dock_direction);
+  if (dock_direction_ == opennav_docking_core::DockDirection::UNKNOWN) {
+    RCLCPP_ERROR(node_->get_logger(), "Dock direction is not valid");
```
SteveMacenski, line 110: I'd throw here, these need to be valid for this plugin

nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
```diff
@@ -87,6 +92,13 @@ void SimpleNonChargingDock::configure(
   node_->get_parameter(name + ".staging_x_offset", staging_x_offset_);
   node_->get_parameter(name + ".staging_yaw_offset", staging_yaw_offset_);
 
+  std::string dock_direction;
+  node_->get_parameter(name + ".dock_direction", dock_direction);
+  dock_direction_ = utils::getDockDirectionFromString(dock_direction);
+  if (dock_direction_ == opennav_docking_core::DockDirection::UNKNOWN) {
+    RCLCPP_ERROR(node_->get_logger(), "Dock direction is not valid");
```
SteveMacenski, line 99: I'd throw here, these need to be valid for this plugin

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -24,22 +25,32 @@ using std::placeholders::_1;
 namespace opennav_docking
 {
 
-DockingServer::DockingServer(const rclcpp::NodeOptions & options)
-: nav2_util::LifecycleNode("docking_server", "", options)
+DockingServer::DockingServer(rclcpp::NodeOptions options)
+: nav2_util::LifecycleNode("docking_server", "",
+    options.automatically_declare_parameters_from_overrides(true))
```
SteveMacenski, line 30: Why the overrides? That's a big change - I think that should be reverted

SteveMacenski:
Otherwise, LGTM!

nav2_docking/opennav_docking/src/docking_server.cpp
ajtudela: With the override, you are allowed to read the ````dock_backwards```` parameter from the yaml file without having previously declared it. This way, if the node has the parameter, it is because it came from an old yaml configuration.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
ajtudela: Standard throw or new ```DockingException``` (DockPluginNotConfigured, for example?

SteveMacenski: It actually doesn't matter, it gets caught as a `std::exception` of any form to return `false` and fail the server initialization

SteveMacenski (approved):

PR #5097, opened 2025-04-21
Increase docking coverage && fix backward docking

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Only testing|

---

- Added more test to increase coverage of dock_database, utils, pose_filter, simple_charging_dock and simple_non_charging_dock.
- Improve docking_server smoke test with environment variables in order to reuse tests.
- Added test for SimpleNonChargingDock plugin and backward docking.
- Fix dock pose and staging pose calculations while backward docking.
- Reduce period in smoke test to reduce testing time. This was check using pytest --force-flaky up to 10 times. From ~54seg to ~38 seg.

---

#### For Maintainers: 


nav2_docking/opennav_docking/src/dock_database.cpp
```diff
@@ -203,7 +203,7 @@ bool DockDatabase::getDockInstances(const rclcpp_lifecycle::LifecycleNode::Share
       node->get_logger(), "Loading dock from database file  %s.", dock_filepath.c_str());
     try {
       return utils::parseDockFile(dock_filepath, node, dock_instances_);
-    } catch (YAML::ParserException & e) {
+    } catch (YAML::BadConversion & e) {
```
SteveMacenski, line 206: Can you explain this one? `BadConversion`'s name implies that it is more specific than `ParserException` which I figure is better to cover all of the `YAML ` failure cases

nav2_docking/opennav_docking/test/test_docking_server.py
```diff
@@ -172,7 +190,7 @@ def test_docking_server(self) -> None:
         self.tf_broadcaster = TransformBroadcaster(self.node)
 
         # Create a timer to run "control loop" at 20hz
-        self.timer = self.node.create_timer(0.05, self.timer_callback)
+        self.timer = self.node.create_timer(0.033, self.timer_callback)
```
SteveMacenski, line 193: Why change the timer and the `self.period`? Just curious.

SteveMacenski (changes requested):

nav2_docking/opennav_docking/src/dock_database.cpp
ajtudela: I had to change the ```ParserException``` to ```BadConversion``` because the ```utils::parseDockFile``` has its own try-catch when loading a file to catch the exceptions, here:

https://github.com/ros-navigation/navigation2/blob/5423eba732ed16b4f97a7353027c1f444b6edc60/nav2_docking/opennav_docking/include/opennav_docking/utils.hpp#L50-L55

The ```utils::parseDockFile``` function only throws a BadConversion if  the value is malformed, in this part:
https://github.com/ros-navigation/navigation2/blob/5423eba732ed16b4f97a7353027c1f444b6edc60/nav2_docking/opennav_docking/include/opennav_docking/utils.hpp#L67-L69

The ```ParserException``` is never thrown.

nav2_docking/opennav_docking/test/test_docking_server.py
ajtudela: Adding two more tests was computationally expensive, so I tried to reduce the duration of the tests.

nav2_docking/opennav_docking/test/test_docking_server.py
```diff
@@ -96,10 +109,14 @@ def tearDownClass(cls) -> None:
 
     def setUp(self) -> None:
         # Create a ROS node for tests
+        self.period = 0.15  # 30Hz
```
mikeferguson, line 112: The comment here doesn't make sense? If period = 0.15, then that's ~6.6hz, not 30hz? Also, how does this work correctly when the timer below is then setup to run at 30hz, but the period doesn't match that?

nav2_docking/opennav_docking/test/test_docking_server.py
```diff
@@ -155,8 +173,8 @@ def nav_execute_callback(
                                       NavigateToPose.Result, NavigateToPose.Feedback]
     ) -> NavigateToPose.Result:
         goal = goal_handle.request
-        self.x = goal.pose.pose.position.x - 0.05
-        self.y = goal.pose.pose.position.y + 0.05
+        self.x = goal.pose.pose.position.x - self.period
+        self.y = goal.pose.pose.position.y + self.period
```
mikeferguson, line 177: I'm don't think the 0.05 here is actually related to the period - I think it's just intended to be a bit of random noise, right?

nav2_docking/opennav_docking/test/test_docking_server.py
mikeferguson: If we keep this, the comment on the line above should change from 20hz->30hz. Even better, we should probably replace 0.05 or 0.033 with the new self.period (and maybe just remove the "at 20hz" part from this comment since the actual rate is defined elsewhere when creating self.period)

SteveMacenski (approved):

ajtudela:
Thanks Steve!

icpmoles:
Hi, I think the bug described in open-navigation/opennav_docking#68 is still present in Jazzy.

https://github.com/ros-navigation/navigation2/blob/e70bedbf115a89d1ebbb648154aad6877462c016/nav2_docking/opennav_docking/src/docking_server.cpp#L434-L449

vs 

https://github.com/ros-navigation/navigation2/blob/8ae7103c89b842d01fcbb59a4112b5fdb0059202/nav2_docking/opennav_docking/src/docking_server.cpp#L509-L524


Is this ever going to be backported to Jazzy?

mini-1235:
https://github.com/Mergifyio backport jazzy

SteveMacenski:
https://github.com/Mergifyio backport jazzy

(I'm surprised you adding the tag didn't work; I understand calling the bot, but the automation I would have thought to run)

SteveMacenski:
Cannot be autobackported https://github.com/ros-navigation/navigation2/pull/5970 probably why it wasn't to date. @icpmoles can you open a PR?

Issue #5088, opened 2025-04-15
MPPI GoalCritic and Path Inversion Bug


alex-roba:
## Bug report

**Required Info:**

- Operating System:
  - ubuntu 24.04
- Computer:
  - 13th Gen Intel NUC
- ROS2 Version:
  - Jazzy
- Version or commit hash:
  - main repository
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
The MPPI controller's GoalCritic behavior, modified in PR #4822, causes the robot to ignore local path inversions and focus solely on the global goal.

1. Set up a Nav2 environment with the MPPI controller.
2. Create a path with an inversion point (e.g., in a narrow aisle requiring a 180-degree turn).
3. Send a goal pose with an orientation opposite to the robot's current direction.
4. Observe the robot bypassing the inversion point and moving directly to the final goal.

#### Expected behavior
The robot should respect path inversions, navigating to intermediary goals (e.g., inversion points) when required, before approaching the final goal.

#### Actual behavior
The robot to ignore local path inversions and focus solely on the global goal.

#### Additional information
Proposed Solution:
Default to using the "real" (global) goal for GoalCritic scoring and other critics.

If a path inversion is detected and inversions are considered (based on a parameter), use the intermediary goal (end of the pruned path or inversion point).

SteveMacenski:
I agree that this is an issue.  See more context in this thread: https://github.com/ros-navigation/navigation2/pull/4822#issuecomment-2795578162 for the issue and possible solutions. 

Anyone in the community open to helping with this? It involves adding the parameter for enforcing path inversions and using the global goal pose vs the path's end pose considering path inversions when MPPI is used with a feasible planner like Smac Planner Hybrid-A*.

BriceRenaudeau:
@alex-roba , Thanks for creating the issue.

BriceRenaudeau:
The quickest solution is to revert the faulty MR #4822

redvinaa:
@BriceRenaudeau reverting #4822 would reintroduce the bug that it solved :) Instead, you could open a PR with the solution that @SteveMacenski suggested.

BriceRenaudeau:
Yes, I understand, for me it was a feature, not a bug. 👍

SteveMacenski:
[...]

PR #5105, opened 2025-04-24
fix MPPI goal critic inversion (#5088)

brayanpa:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (#5088) |
| Primary OS tested on | (Ubuntu 24.04.2 LTS) |
| Robotic platform tested on | (gazebo simulation, Gary Robot Hardware) |

---

colcon testing

*  Added logic to the GoalCritic in the MPPI controller to check the enforce_path_inversion parameter.
*  When enforce_path_inversion is set to true, the end of the path is used instead of the goal pose to compute costs.
*  This resolves cases where the robot would get stuck near the goal in an incorrect orientation due to undetected path inversion, especially when using planners that generate feasible paths.

Tested in simulation and on a real robot across multiple maps and scenarios where feasible planners produced paths that passed near the goal before performing an inversion to reach it with the correct orientation.

Previously, the GoalCritic would interpret this proximity as having reached the goal, preventing further progress and causing the robot to get stuck.

With the new logic using the path endpoint instead of the goal pose when enforce_path_inversion is true, the robot was able to continue following the path and reach the goal correctly.

All test cases confirmed that the robot no longer gets stuck near the goal due to orientation mismatches caused by undetected path inversions.

---

#### For Maintainers: 


nav2_mppi_controller/src/critics/goal_critic.cpp
```diff
@@ -21,6 +21,9 @@ namespace mppi::critics
 
 void GoalCritic::initialize()
 {
+  auto getParentParam = parameters_handler_->getParamGetter(parent_name_);
+  getParentParam(enforce_path_inversion_, "enforce_path_inversion", false);
```
SteveMacenski, line 25: I think this should be using the `name_` getter instead. This is a goal critic plugin specific parameterization

nav2_mppi_controller/src/critics/goal_critic.cpp
```diff
@@ -40,8 +43,14 @@ void GoalCritic::score(CriticData & data)
     return;
   }
 
-  const auto & goal_x = data.goal.position.x;
-  const auto & goal_y = data.goal.position.y;
+  auto goal_x = data.goal.position.x;
```
SteveMacenski, line 46: Should `withinPositionGoalTolerance` use the back of the path as well in this case @BriceRenaudeau? Possibly need to be considered for other critic plugins as well

brayanpa: I think it would be good to add it to the `goal_angle_critic`. Let me know if you think it should be added to any others as well.

Regarding adding it to `withinPositionGoalTolerance`, I think the best approach would be to create a parameter in each critic plugin to choose whether to use the existing `withinPositionGoalTolerance` logic or a new one that considers path inversions. What do you think?

SteveMacenski: @brayanpa I think you're probably right that we should have `enforce_path_inversion` be a plugin-level parameter that is used in the individual plugins as-needed.

I think Goal and GoalAngle are obvious ones that need to look at that and use the goal vs local goal for both is computation, but also its within tolerance check. 

The others are Cost, Obstacles, Twirl, Path Align, Path Follow, Path Angle, and Prefer Forward.

The Path Align/Follow/Angle have natural hand-offs from the Goal and GoalAngle critics, so anything that the Path critics sees the Goal critics also need to see. If intermediary goals for path inversions are near obstacles like the goal, I think the Cost and Obstacles critics probably should have it too. At that point, lets just make it universal for consistency. 

So, I think the first thing we should do in each critics' `score` method is to check on the inversion parameter and store the `data.goal` or `data.path(-1)` for the goal to check. Then that should be used for `withinPositionGoalTolerance` and for the computation of the critic value where necessary.

brayanpa: I have added the `enforce_path_inversion` parameter to all critics and modified the goal between the end of the path and the original goal accordingly in the necessary critics. In the tests I performed, it is working very well. The robot is able to handle path inversion correctly, and when `enforce_path_inversion` is set to false, it behaves just like before.

I'll be on the lookout in case any further changes are needed, or if you think any additional clarification is required in the README or documentation.

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -119,6 +122,13 @@ void CostCritic::score(CriticData & data)
     return;
   }
 
+  geometry_msgs::msg::Pose active_goal;
```
SteveMacenski, line 125: How about just `goal` (here and all other cases)?

nav2_mppi_controller/src/critics/goal_angle_critic.cpp
```diff
@@ -35,14 +36,26 @@ void GoalAngleCritic::initialize()
...
+      threshold_to_consider_, data.state.pose.pose, active_goal))
   {
     return;
   }
 
-  const auto goal_idx = data.path.x.size() - 1;
-  const float goal_yaw = data.path.yaws(goal_idx);
+  tf2::Quaternion goal_orientation_q;
+  tf2::fromMsg(active_goal.orientation, goal_orientation_q);
+  double goal_yaw = tf2::getYaw(goal_orientation_q);
```
SteveMacenski, line 58: I believe this works with just a geometry_msgs/Quaternion too https://github.com/ros-navigation/navigation2/blob/6e58fbe2888cf88d3040edb4b083b8dc1dfd0cd6/nav2_behaviors/plugins/spin.cpp#L86

nav2_mppi_controller/src/critics/cost_critic.cpp
```diff
@@ -119,6 +122,13 @@ void CostCritic::score(CriticData & data)
...
+  if (enforce_path_inversion_) {
```
SteveMacenski, line 126: I think we can abstract this even more to a util:

```
geometry_msgs::msg::Pose getCriticGoal(data, enforce_path_inversion)
{
  if (enable_path_inversion) {
    return getLastPathPose(data.path);
  } else {
    return data.goal;
  }
}
```

that would make that function testable as a unit and we don't have all the missing coverage warnings in the individual critics for this block since we'd just call that one util in each to return the goal extracted

SteveMacenski (changes requested):
Pedantic stuff for code coverage and readability, this is good to merge after this.

thanks so much!

For docs: Updating the docs.nav2.org configuration guide page for MPPI with what you added to the readme would be fabulous

SteveMacenski (approved):
For docs: Updating the docs.nav2.org configuration guide page for MPPI with what you added to the readme would be fabulous

@brayanpa CI failed, one of the tests is crashing related to the critic unit tests - please take a look :-) 

@BriceRenaudeau  please review again to approve :-)

brayanpa:
@SteveMacenski You were right, there was an error in the tests. I'll make some time as soon as possible to review how to update the docs.nav2.org configuration guide page.

SteveMacenski (approved):

SteveMacenski:
@brayanpa update this file: https://github.com/ros-navigation/docs.nav2.org/blob/master/configuration/packages/configuring-mppic.rst You just need to add some context to that parameter so the changes should be easy

brayanpa:
@SteveMacenski Thank you for the explanation. I've created this PR https://github.com/ros-navigation/docs.nav2.org/pull/691.

SteveMacenski:
Fabulous, thank you!

nav2_mppi_controller/src/critics/obstacles_critic.cpp
```diff
@@ -23,6 +23,9 @@ namespace mppi::critics
 
 void ObstaclesCritic::initialize()
 {
+  auto getParentParam = parameters_handler_->getParamGetter(parent_name_);
+  getParentParam(enforce_path_inversion_, "enforce_path_inversion1", false);
```
tonynajjar, line 27: Is "enforce_path_inversion1" a typo?

brayanpa: You're right, that was a typo — it should be "enforce_path_inversion" instead of "enforce_path_inversion1". Thanks for catching it!

Issue #4400, opened 2024-06-05
Add in optional keepout / speed zones for example in new warehouse maps


SteveMacenski:
Show the features in use on the TB4 in warehouse

Might be good to have them turn-out-able (launch CLI parameter) so that we can have easy toggle with them off for normal navigation.

sterlingm:
Is this already implemented? The following two tutorial pages discuss keepout and speed limit zones: https://docs.nav2.org/tutorials/docs/navigation2_with_keepout_filter.html
https://docs.nav2.org/tutorials/docs/navigation2_with_speed_filter.html

SteveMacenski:
Not yet, this is to have an in-repo example of each of these using the TB4 launch file added in `main` and `jazzy` of a warehouse as technology demonstration and for testing and evaluation. This would be a great first-time contribution to add in some keepout zones, speed restriction zones, and configurations so folks have a working out-of-the-box example! :smile:

stevedanomodolor:
Assign this to me @SteveMacenski

SteveMacenski:
As an extension of this work - it would be great to have a checker for if we're currently in a keepout zone to enable the robot to get out of it by setting the costs to very-high but still navigable. I know this is a common feature among keepout users that we don't yet have addressed in our designs to-date

SteveMacenski:
@stevedanomodolor how's this going? 😄

stevedanomodolor:
> @stevedanomodolor how's this going? 😄 

Was working on this, had an annoying bug that was still figuring out what the issue was. I will continue this and post more info on this, this week. 
Sorry for the lateness, it's been a busy month.

SteveMacenski:
Totally understand, just checking in since we had a few new contributors reach out this month, so I wanted to know if this was still on your radar. Thanks!

rohinraj:
Hi, @SteveMacenski 

I have created a PR (https://github.com/ros-navigation/navigation2/pull/5128)

This contains both keepout filter and speed filters added in seperate launches for the new tb4 warehouse map.

I am not that sure weather this is what exactly required or not. Kindly take a look in to it and let me know if any additional changes are required.

SteveMacenski:
Will do

SteveMacenski:
Keepout zone complete

SteveMacenski:
Great, thanks @leander-dsouza !

PR #5125, opened 2025-05-04
Added keepout filter for depot and warehouse maps

leander-dsouza:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4400 |
| Robotic platform tested on | Turtlebot4 Simulation|

---

* Validated lintiing using pre-commit
* Tested the package using colcon test.

### Depot

* Added a keepout region at the centre of the depot map in turtlebot4.

   tb4_keepout_depot

   https://github.com/user-attachments/assets/215278d9-74c5-4d45-bd28-802dc6a9864a

### Warehouse

* Added a keepout region at the centre of the warehouse map in turtlebot4.

   tb4_keepout

   https://github.com/user-attachments/assets/cd2ce5e6-b623-4313-857a-4902edf31be6

* Added a `keepout_map`  and `use_keepout_filter` as launch arguments.
* Added a constant dictionary for easily switching between `depot` and `warehouse` maps for turtlebot4.

* A small tutorial can be added once the speed zones are added to both depot and warehouse worlds.

* A goal was given, and the robot avoided the keepout region.

* Add speed zones in warehouse and depot maps.

#### For Maintainers: 


nav2_bringup/maps/warehouse_keepout.pgm
SteveMacenski, line 1: What map is this? I don't recognize it. I don't think this is the depot map that the TB4 brings up with :-)

nav2_bringup/launch/tb4_simulation_launch.py
```diff
@@ -79,10 +93,17 @@ def generate_launch_description() -> LaunchDescription:
...
         'map',
-        default_value=os.path.join(bringup_dir, 'maps', 'depot.yaml'),  # Try warehouse.yaml!
+        default_value=os.path.join(bringup_dir, 'maps', f'{MAP_TYPE}.yaml'),
         description='Full path to map file to load',
     )
 
+    declare_keepout_map_yaml_cmd = DeclareLaunchArgument(
+        'keepout_map',
+        default_value='',
+        # default_value=os.path.join(bringup_dir, 'maps', 'warehouse_keepout.yaml'),
```
SteveMacenski, line 103: Setting a default is reasonable! Just key this to the MAP_TYPE like you did the others

nav2_bringup/launch/tb4_simulation_launch.py
```diff
@@ -41,6 +54,7 @@ def generate_launch_description() -> LaunchDescription:
     slam = LaunchConfiguration('slam')
     namespace = LaunchConfiguration('namespace')
     map_yaml_file = LaunchConfiguration('map')
+    keepout_map_yaml_file = LaunchConfiguration('keepout_map')
```
SteveMacenski, line 57: Having a configuration to disable / enable the keepout map would be nice (default on)

nav2_bringup/launch/localization_launch.py
```diff
@@ -31,6 +31,7 @@ def generate_launch_description() -> LaunchDescription:
 
     namespace = LaunchConfiguration('namespace')
     map_yaml_file = LaunchConfiguration('map')
+    keepout_map_yaml_file = LaunchConfiguration('keepout_map')
```
SteveMacenski, line 34: Move the keepout/speed stuff into its own launch file that is run in `bringup_launch.py` alongside of navigation, localization as another launch file entirely. That way it can be easily disabled / enabled using the IfCondition based on if we want to use the filters or not

SteveMacenski (changes requested):
Pretty good!

nav2_bringup/maps/warehouse_keepout.pgm
SteveMacenski: I see now reading your description. Lets do the depot first since that's the default bringup, but happy to include both!

leander-dsouza: I have added the depot map to have a keepout filter region as well.

nav2_bringup/launch/tb4_simulation_launch.py
leander-dsouza: I have added `_keepout.yaml` as the default argument.

nav2_bringup/launch/tb4_simulation_launch.py
leander-dsouza: I have added `use_keepout_filter` as a boolean to toggle this behaviour.

nav2_bringup/launch/localization_launch.py
leander-dsouza: I have created a `map_modifier.launch.py` to keep the keepout filter behaviour. The next steps would be to add the speed limits to both maps in a future commit to this file.

nav2_bringup/launch/bringup_launch.py
```diff
@@ -43,6 +44,7 @@ def generate_launch_description() -> LaunchDescription:
     use_respawn = LaunchConfiguration('use_respawn')
     log_level = LaunchConfiguration('log_level')
     use_localization = LaunchConfiguration('use_localization')
+    use_keepout_filter = LaunchConfiguration('use_keepout_filter')
```
SteveMacenski, line 47: Maybe `use_keepout_zones` to abstract the implementation detail a bit that its a costmap filter. I think 'keepout zones' is more common language that'll be more clear

nav2_bringup/launch/bringup_launch.py
```diff
@@ -35,6 +35,7 @@ def generate_launch_description() -> LaunchDescription:
     namespace = LaunchConfiguration('namespace')
     slam = LaunchConfiguration('slam')
     map_yaml_file = LaunchConfiguration('map')
+    keepout_map_yaml_file = LaunchConfiguration('keepout_map')
```
SteveMacenski, line 38: `keepout_mask`

nav2_bringup/launch/map_modifier.launch.py
```diff
@@ -0,0 +1,219 @@
+# Copyright (c) 2025 Leander Stephen Desouza
```
SteveMacenski, line 1: I like the naming intent, but I want to make this more immediately clear to end users. From trying to come up with suggestions I can ACK the difficulty here to name a file to include both zone info nodes :-) 

How about: `keepout_zone_launch.py` and we have a separate `speed_zone_launch.py` as well, since I think we're going to have 2 `use_XYZ` launch args for keepout and speed zone use. 

Also, use `_launch.py` not `.launch.py` to be aligned with the other file extensions

nav2_bringup/params/nav2_params.yaml
```diff
@@ -224,6 +224,11 @@ local_costmap:
       resolution: 0.05
       robot_radius: 0.22
       plugins: ["voxel_layer", "inflation_layer"]
+      filters: ["keepout_filter"]
+      keepout_filter:
```
SteveMacenski, line 228: Curious: if this is included but no keepout zone is published because `use_keepout_zones` isn't enabled, does this work? Obviously logs some warnings about nothing published, I assume, but navigation works?

nav2_bringup/params/nav2_params.yaml
```diff
@@ -224,6 +224,11 @@ local_costmap:
...
+        plugin: "nav2_costmap_2d::KeepoutFilter"
+        enabled: True
+        filter_info_topic: "/costmap_filter_info"
```
SteveMacenski, line 231: For all: remove the `/` so that this works with namespacing

nav2_bringup/maps/depot_keepout.pgm
SteveMacenski, line 1: I hate being picky about this kind of thing, but given this is the first demo most users will experience this feature with, I think its important to make sure these zones are actually useful and represent a realistic need.

For this can we instead have the keepouts:
- Surrounding the area with the 8 pallets, since that's an area of high movement where you typically might not want a robot to enter if not doing work in it
- Directly in front of the stairwell so that it doesn't block the stairs
- Under the shelving units so the robot doesn't go under

nav2_bringup/maps/warehouse_keepout.pgm
SteveMacenski, line 1: For this, I actually dont know what this gazebo model looks like well off of the top of my head. But, picking something reasonable as well for demo purposes :-)

SteveMacenski (changes requested):
LGTM! Lets get this in before doing the speed zones in a follow up PR. You can open a PR against this branch if you have that done already so we can review, but it would be good to keep these 2 PRs decoupled to make it easier to review

nav2_bringup/launch/bringup_launch.py
leander-dsouza: I have renamed the variable and the launch argument to be `use_keepout_zones`.

nav2_bringup/launch/bringup_launch.py
leander-dsouza: I have renamed the variable and the launch argument to use `keepout_mask` instead.

nav2_bringup/launch/map_modifier.launch.py
leander-dsouza: Thank you for the detailed explanation, Steve.
I have renamed the launch file to `keepout_zone_launch.py`.

nav2_bringup/params/nav2_params.yaml
leander-dsouza: * You get the following warning if `use_keepout_zones` is set to False:
   ```bash
   [local_costmap.local_costmap]: KeepoutFilter: Filter mask was not received
   ```
* The navigation works as expected:

   https://github.com/user-attachments/assets/2ed2b8dd-76bf-4276-93f4-0f08a0423ead

nav2_bringup/params/nav2_params.yaml
leander-dsouza: * I do not think this works because if I remove the `/`, the global costmap and local costmap expect the filter info topic to be under its respective namespace instead.

   ```bash
   $ ros2 topic list | grep costmap_filter_info

   /costmap_filter_info
   /costmap_filter_info_server/transition_event
   /global_costmap/costmap_filter_info
   /local_costmap/costmap_filter_info
   ```

nav2_bringup/maps/depot_keepout.pgm
leander-dsouza: I have attached the keepout maps as specified.

* Gazebo world for reference

   image

* Keepout regions in RViz

   image

nav2_bringup/maps/warehouse_keepout.pgm
leander-dsouza: For the warehouse map, I have added keepout regions on the table and chairs.

* Gazebo world for reference

   image

* Keepout regions in RViz

   image

SteveMacenski: I think setting that center area with the tape as a keep out would be sensible in the warehouse.

nav2_bringup/maps/depot_keepout.pgm
SteveMacenski: For depot: looks good! Can we just reduce the shapes on the shelfs to be the posts without the extra padding? The zones are **really** close to the route graph nodes.

nav2_bringup/params/nav2_params.yaml
SteveMacenski: But wouldn't there be a collision then if doing both speed and keepout zones?

nav2_bringup/maps/warehouse_keepout.pgm
leander-dsouza: I have modified the keepout to match the square drawn in the center.

image

nav2_bringup/maps/depot_keepout.pgm
leander-dsouza: I have reduced the padding of the shelves and the eight pallet area as follows:

image

nav2_bringup/params/nav2_params.yaml
leander-dsouza: I have included a [utility script](https://github.com/ros-navigation/navigation2/pull/5125/files#diff-8cdb8b3aba974946a53f53b45b24249e58d50c6919b88b3859d416c276b62e26R44) that has `joinWithParentNamespace`. This was to prevent redefining it under the costmap filters.

SteveMacenski: @leander-dsouza you could just move the function to the `Layer` object from the `CostmapLayer` object instead, since the filters use the `Layer` as well

SteveMacenski:
After that `Layer` update, I think this is good to merge!

SteveMacenski (approved):

SteveMacenski:
FYI @leander-dsouza now that this is default behavior, I filed a ticket about a known quirk that we may want to address. Its a bit more substantive that you can sink your teeth into if you're interested: https://github.com/ros-navigation/navigation2/issues/5139

nav2_bringup/params/nav2_params.yaml
leander-dsouza: Thank you for the suggestion, Steve.
I have migrated the `joinWithParentNamespace` function over to the Layer script.

SteveMacenski:
Done! I look forward to the speed filter next! 

I think you should make a nice gif of the roobt doing its thing with keepout and speed zones. That would be good to include in the Migration Guide and we can jointly announce that on LinkedIn :-)

leander-dsouza:
> Done! I look forward to the speed filter next!
> 
> I think you should make a nice gif of the robot doing its thing with keepout and speed zones. That would be good to include in the Migration Guide and we can jointly announce that on LinkedIn :-)

That sounds great :)
I will make the PR for the speed filter shortly.

PR #5137, opened 2025-05-09
removing the start navigation msg in the paused state from rviz button

padhupradheep:
---

## Basic Info

| ------ | ----------- |

---

When we enter the paused mode in the nav2_rviz_plugin, there are currently three buttons appearing, resume, reset and start navigation. Apparently start navigation button does not have any use and looks to be bit misleading. Therefore removing that and also have disabled the button. 

---

#### For Maintainers: 


SteveMacenski (approved):
Nice!

PR #5138, opened 2025-05-09
[nav2_map_server] add launcher for map_server

adivardi:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Simulation with rviz |

---

This PR fixes a bug when loading .pgm maps in the trinary mode. In this mode, the `occupied_thresh` and `free_thresh` specified in the `.yaml` file are meaningless, since the values of the image pixels are constants determined when the map was saved. Thus the threshold used when loading should correspond to these pixels values.
Without this fix, a map whose `free_thresh` is above 0.196 will show all `UNKNOWN` pixels as `FREE`. 

Example using `nav2_bringup/.../depot.yaml`:

Without the fix:
Screenshot from 2025-05-09 15-11-02

With the fix:
Screenshot from 2025-05-09 15-12-37

negated:
Screenshot from 2025-05-09 15-13-01

Also added a standalone launcher for the map_server so it can be tested

None?

* Visual test using the launcher
* Performed linting validation using colcon test

---

#### For Maintainers: 

### DIFF
--- a/nav2_map_server/launch/map_server.launch.py
+++ b/nav2_map_server/launch/map_server.launch.py
@@ -23,38 +23,38 @@ import launch_ros.actions
 
 def generate_launch_description() -> LaunchDescription:
     # Parameters
-    bringup_dir = get_package_share_directory("nav2_bringup")
-    map_yaml_file = os.path.join(bringup_dir, "maps", "depot.yaml")
+    bringup_dir = get_package_share_directory('nav2_bringup')
+    map_yaml_file = os.path.join(bringup_dir, 'maps', 'depot.yaml')
 
-    lifecycle_nodes = ["map_server"]
+    lifecycle_nodes = ['map_server']
     use_sim_time = True
     autostart = True
-    topic_name = "map"
-    frame_id = "map"
-    service_introspection_mode = "disabled"
+    topic_name = 'map'
+    frame_id = 'map'
+    service_introspection_mode = 'disabled'
 
     start_map_server_cmd = launch_ros.actions.Node(
-        package="nav2_map_server",
-        executable="map_server",
-        name="map_server",
-        output="screen",
+        package='nav2_map_server',
+        executable='map_server',
+        name='map_server',
+        output='screen',
         parameters=[
-            {"yaml_filename": map_yaml_file},
-            {"topic_name": topic_name},
-            {"frame_id": frame_id},
-            {"service_introspection_mode": service_introspection_mode},
+            {'yaml_filename': map_yaml_file},
+            {'topic_name': topic_name},
+            {'frame_id': frame_id},
+            {'service_introspection_mode': service_introspection_mode},
         ],
     )
 
     start_lifecycle_manager_cmd = launch_ros.actions.Node(
-        package="nav2_lifecycle_manager",
-        executable="lifecycle_manager",
-        name="lifecycle_manager",
-        output="screen",
+        package='nav2_lifecycle_manager',
+        executable='lifecycle_manager',
+        name='lifecycle_manager',
+        output='screen',
         parameters=[
-            {"use_sim_time": use_sim_time},
-            {"autostart": autostart},
-            {"node_names": lifecycle_nodes},
+            {'use_sim_time': use_sim_time},
+            {'autostart': autostart},
+            {'node_names': lifecycle_nodes},
         ],
     )
 
--- a/nav2_map_server/launch/map_server.launch.py
+++ b/nav2_map_server/launch/map_server.launch.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2020 Samsung Research Russia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from ament_index_python.packages import get_package_share_directory
+from launch import LaunchDescription
+import launch_ros.actions
+
+
+def generate_launch_description() -> LaunchDescription:
+    # Parameters
+    bringup_dir = get_package_share_directory("nav2_bringup")
+    map_yaml_file = os.path.join(bringup_dir, "maps", "depot.yaml")
+
+    lifecycle_nodes = ["map_server"]
+    use_sim_time = True
+    autostart = True
+    topic_name = "map"
+    frame_id = "map"
+    service_introspection_mode = "disabled"
+
+    start_map_server_cmd = launch_ros.actions.Node(
+        package="nav2_map_server",
+        executable="map_server",
+        name="map_server",
+        output="screen",
+        parameters=[
+            {"yaml_filename": map_yaml_file},
+            {"topic_name": topic_name},
+            {"frame_id": frame_id},
+            {"service_introspection_mode": service_introspection_mode},
+        ],
+    )
+
+    start_lifecycle_manager_cmd = launch_ros.actions.Node(
+        package="nav2_lifecycle_manager",
+        executable="lifecycle_manager",
+        name="lifecycle_manager",
+        output="screen",
+        parameters=[
+            {"use_sim_time": use_sim_time},
+            {"autostart": autostart},
+            {"node_names": lifecycle_nodes},
+        ],
+    )
+
+    ld = LaunchDescription()
+
+    ld.add_action(start_map_server_cmd)
+    ld.add_action(start_lifecycle_manager_cmd)
+
+    return ld
--- a/nav2_map_server/src/map_io.cpp
+++ b/nav2_map_server/src/map_io.cpp
@@ -261,12 +261,26 @@ void loadMapFromFile(
       normalized = (1.0f - normalized.array()).matrix();
     }
 
+    double occupied_thresh = load_parameters.occupied_thresh;
+    double free_thresh = load_parameters.free_thresh;
+
+    // For Trinary mode, the thresholds depend on the constants used when saving the map,
+    // not on the thresholds from the yaml file
+    // occupied is always 1.0, free is always 0.00392157 ( = 1.0 - 254 / 255.0),
+    // and unknown is always 0.196078431 ( = 1.0 - 205 / 255.0)
+    // For Scale mode, the thresholds are set by the user in the yaml file
+    if (load_parameters.mode == MapMode::Trinary)
+    {
+      occupied_thresh = 254 / 255.0;
+      free_thresh = 2 / 255.0;
+    }
+
     // Compute binary occupancy masks
     Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> occupied =
-      (normalized.array() >= load_parameters.occupied_thresh).cast<uint8_t>();
+      (normalized.array() >= occupied_thresh).cast<uint8_t>();
 
     Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> free =
-      (normalized.array() <= load_parameters.free_thresh).cast<uint8_t>();
+      (normalized.array() <= free_thresh).cast<uint8_t>();
 
     // Initialize occupancy grid with UNKNOWN values (-1)
     result.setConstant(nav2_util::OCC_GRID_UNKNOWN);
--- a/nav2_map_server/src/map_io.cpp
+++ b/nav2_map_server/src/map_io.cpp
@@ -269,8 +269,7 @@ void loadMapFromFile(
     // occupied is always 1.0, free is always 0.00392157 ( = 1.0 - 254 / 255.0),
     // and unknown is always 0.196078431 ( = 1.0 - 205 / 255.0)
     // For Scale mode, the thresholds are set by the user in the yaml file
-    if (load_parameters.mode == MapMode::Trinary)
-    {
+    if (load_parameters.mode == MapMode::Trinary) {
       occupied_thresh = 254 / 255.0;
       free_thresh = 2 / 255.0;
     }


nav2_map_server/src/map_io.cpp
```diff
@@ -261,12 +261,25 @@ void loadMapFromFile(
       normalized = (1.0f - normalized.array()).matrix();
     }
 
+    double occupied_thresh = load_parameters.occupied_thresh;
```
SteveMacenski, line 264: Remove and use `load_parameters.occupied_thresh` directly please - we do so lots of places in the code below so changing just the one instance below this isn't consistent

nav2_map_server/src/map_io.cpp
```diff
@@ -261,12 +261,25 @@ void loadMapFromFile(
...
+    double free_thresh = load_parameters.free_thresh;
+
+    // For Trinary mode, the thresholds depend on the constants used when saving the map,
+    // not on the thresholds from the yaml file
+    // occupied is always 1.0, free is always 0.00392157 ( = 1.0 - 254 / 255.0),
+    // and unknown is always 0.196078431 ( = 1.0 - 205 / 255.0)
+    // For Scale mode, the thresholds are set by the user in the yaml file
+    if (load_parameters.mode == MapMode::Trinary) {
```
SteveMacenski, line 272: I dont' see this as consistent with the previous behavior https://github.com/ros-navigation/navigation2/blob/jazzy/nav2_map_server/src/map_io.cpp#L254-L260

nav2_map_server/src/map_io.cpp
```diff
@@ -261,12 +261,25 @@ void loadMapFromFile(
       normalized = (1.0f - normalized.array()).matrix();
     }
 
+    double occupied_thresh = load_parameters.occupied_thresh;
+    double free_thresh = load_parameters.free_thresh;
+
+    // For Trinary mode, the thresholds depend on the constants used when saving the map,
```
SteveMacenski, line 267: But isn't that supposed to be represented in the thresholds in the yaml file? That is kind of the point of the yaml file. 

I don't know about occupied is always 1.0 and free is always 0.039. Where is that coming from?

SteveMacenski (changes requested):

nav2_map_server/src/map_io.cpp
adivardi: The behavior in Jazzy has the same issue. I was planning to fix it for Jazzy, even before the new Eigen PR

nav2_map_server/src/map_io.cpp
adivardi: I am assuming the images used were saved by the `map_saver`, not by hand. If so, the image created can only have 3 pixel values (0 205 254),   from [here](https://github.com/ros-navigation/navigation2/blob/main/nav2_map_server/src/map_io.cpp#L541-L548)
When loading, the values are normalized and complemented to 1 (I think by Magick). So you can only have the 3 values in the comment.

Regarding thresholds: the yaml file now records the thresholds used when saving the file. Those still make sense if the map being saved has graduation and are used to threshold it. Only after the shareholding to Trinary (so when loading), they become meaningless.
Another option could be to overwrite the th thresholds when saving a trinary map. I like this idea less because if someone modifies the yaml file by hand it will break again. Also, specifying thresholds values when saving a map and the discovering different values in the yaml file would look like a bug.

nav2_map_server/src/map_io.cpp
SteveMacenski: Can you point to where this is in ROS 1 then?

nav2_map_server/src/map_io.cpp
SteveMacenski: I don't think the assumption that it is always the map saver is sound, unfortunately. Others have their own map saving utilities so we should look at the yaml values stored (which should correspond properly to those 3 values, right?)

adivardi: Right now the map_saver will save to yaml the thresholds used to threshold the probability map being saved.
These are not correct when loading Trinary map.
For example, if you use the default map_saver thresholds of free=0.25, occupied=0.65, it will save a nice map properly. But when loading later on, it won't work since the UNKNOWN value is 0.19 and doesn't lie between the 2 thresholds.

If that assumption is wrong, so maybe the other option is better? When saving with the map_saver, save the thresholds for loading instead of the thresholds for saving. Like I mentioned, it will break if someone change the file by hand and may also be a bit confusing, but imo still better than nothing.

nav2_map_server/src/map_io.cpp
adivardi: Loading: https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/image_loader.cpp
Saving pixels: https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/map_saver.cpp#L77-L82
Saving yaml file: https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/map_saver.cpp#L96-L116

In ROS1 they use the 2nd option I proposed, i.e. always saving `occupied_thresh: 0.65, free_thresh: 0.196` regardless of the thresholds used to threshold the map (given by the user with `--occ` and `--free`

vickzk:
> @vickzk can you review? This seems like it might fix a gap in your previous PR and I want to make sure you agree with it

Sure. I'll check.

nav2_map_server/launch/map_server.launch.py
```diff
@@ -0,0 +1,66 @@
...
+import launch_ros.actions
+
+
+def generate_launch_description() -> LaunchDescription:
+    # Parameters
+    bringup_dir = get_package_share_directory('nav2_bringup')
+    map_yaml_file = os.path.join(bringup_dir, 'maps', 'depot.yaml')
+
+    lifecycle_nodes = ['map_server']
+    use_sim_time = True
```
vickzk, line 30: Suggest passing `use_sim_time` as a launch argument

nav2_map_server/src/map_io.cpp
SteveMacenski: In ROS 2 we use the thresholds the user set though https://github.com/ros-navigation/navigation2/blob/main/nav2_map_server/src/map_saver/map_saver.cpp

nav2_map_server/src/map_io.cpp
SteveMacenski: Why is it not correct? Aren't these all user defined settings they can set to their needs in Nav2? 

I'm not trying to be obtuse here, but I'm also very cautious about any changes to map server or AMCL since these are mega-legacy packages that have alot of annoying subtle implications so I prefer to not touch them or their behavior unless we have extremely good reason. I don't really understand the issue.

vickzk:
@adivardi  @SteveMacenski 

Right now, when users don’t provide threshold values while saving a map with `map_saver`, it defaults to a free threshold of `0.25`. For `trinary` maps, it would make more sense to use `0.196` instead, since that’s a more appropriate default for this mode. So, we can update this to use `0.196` when saving in `trinary` mode. 
**Note:** Users can also achieve this manually by setting the `free_thresh_default` [parameter](https://github.com/ros-navigation/navigation2/blob/711817f995d5a33dba29e7330874bd93b50fe860/nav2_map_server/src/map_saver/map_saver.cpp#L70C41-L70C60) 

If users still want to customize the thresholds for their own needs or when saving from different sources, they absolutely can. The `map_server` will continue to read the thresholds from the `YAML` file and use those when publishing the map.

This small change doesn’t affect existing behavior but makes the defaults more aligned with how trinary maps are typically used.

nav2_map_server/src/map_io.cpp
adivardi: Yes, but those are used to threshold the probability map. They are also used in ROS 1 [here](https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/map_saver.cpp#L77-L82)
But ROS1 doesn't save these to file, instead always saving 0.196 & 0.65.

adivardi:
> @adivardi @SteveMacenski
> 
> Right now, when users don’t provide threshold values while saving a map with `map_saver`, it defaults to a free threshold of `0.25`. For `trinary` maps, it would make more sense to use `0.196` instead, since that’s a more appropriate default for this mode. So, we can update this to use `0.196` when saving in `trinary` mode. **Note:** Users can also achieve this manually by setting the `free_thresh_default` [parameter](https://github.com/ros-navigation/navigation2/blob/711817f995d5a33dba29e7330874bd93b50fe860/nav2_map_server/src/map_saver/map_saver.cpp#L70C41-L70C60)
> 
> If users still want to customize the thresholds for their own needs or when saving from different sources, they absolutely can. The `map_server` will continue to read the thresholds from the `YAML` file and use those when publishing the map.
> 
> This small change doesn’t affect existing behavior but makes the defaults more aligned with how trinary maps are typically used.

I think this hides the problem more than solves it.
The threshold values used when saving should be appropriate to the input probability map. If it is very noisy, users may need to increase the threshold. Once they change it, the yaml file will contain their new threshold, and the map won't load correctly anymore. They would need to manually change the yaml file value back to 0.196

SteveMacenski:
Looking over the code in the PR, I don't know what necessarily the right answer is from this thread, but I know just hardcoding the values as this PR is currently implemented is not something I would be open to merging. It should be configurable, respecting YAML configurations, parameter configurations, or at least something. I would not assume that every map saved in Nav2 is from the map saver as some SLAM algorithms export occupancy grid maps themselves. 

Can you adjust the PR to be more configurable while meeting the need / issue you present ?

adivardi:
How about the ROS1 solution where the yaml   values are hard coded when saving a map?(together with the hard coded pixels values)

SteveMacenski:
I went back to the base premise of this PR

> This PR fixes a bug when loading .pgm maps in the trinary mode. In this mode, the occupied_thresh and free_thresh specified in the .yaml file are meaningless, since the values of the image pixels are constants determined when the map was saved. Thus the threshold used when loading should correspond to these pixels values.

I don't understand how these values are 'meaningless' in Trinary mode. These are set as configurations anyone can set to align with their needs when they save a map from their SLAM vendor. If the normalized value is > 0.65 is occupied or < 0.25 its free, [for example](https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/maps/depot.yaml). The values in the image are constants, but they can be interpreted differently. A map saved in a continuous probability format like from Cartographer can later be served as trinary where we set thresholds for what we consider on the 0-99% scale to be occupied vs free in this situation. 

The 'unknown' in the middle from 0.25-0.65 is a bit odd but also not what that map was saved to do (and makes alot of sense in scale mode). -1 and 255 are intepreted as unknown which is typically what is used. 

In short, I'm really not seeing an issue, but if there is an issue, its due to the need for more configurability not less. I wouldn't accept a PR that would hardcode values.

adivardi:
The issue right now, is that if you use both map_saver and map loader from nav2, you have to manually edit the yaml file for the map to load properly.

As I mentioned before, I originally assumed that all maps loaded are created by the map_saver. With this assumption, the thresholds are meaningless because pixels can only have 3 values. 
When we allow other map sources, this is not true anymore and I agree that the solution implemented in this PR is not good.

Instead, I suggested changing the thresholds in the yaml file when saving using the map_saver. Since the map saver always save pixels with only 3 constant values, the threshold saved could be constants.
This is what was done in ROS1 and would only affect maps saved by the map_saver. 

What are your thoughts on this solution? (replacing the current PR change, not in addition)


If that's also not acceptable, I'm open to other ideas. Otherwise, I will just write a wrapper for myself to overwrite the yaml files, so I can get automatic mapping to work without human in the loop.
If you are open to it, I can open a PR in the documentation to add a note that if you saved a map using the map_saver, you must manually edit the yaml file for it to load properly by the map_server.

SteveMacenski:
> Instead, I suggest changing the thresholds in the yaml file when saving using the map_saver. Since the map saver always save pixels with only 3 constant values, the threshold saved could be constants. This is what was done in [ROS1](https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/map_saver.cpp#L96-L116) and would only affect maps saved by the map_saver.

But what about other folks that are saving them as non-Trinary? There are already [Trinary values set for saving](https://github.com/ros-navigation/navigation2/blob/main/nav2_map_server/src/map_io.cpp#L540-L549) based on the configured values in the saving service (and cases for each of the others as well).  The Map Saver, even if a user does use it, is not static in its values. They are configurations set by the user in the [save request](https://github.com/ros-navigation/navigation2/blob/fda475c8009d3f2522292a62b721a0585b786867/nav2_map_server/src/map_saver/map_saver.cpp#L129-L133). You can adjust these in calling the saving service for whatever is appropriate for your needs.

I don't see any situation in which removing configurability and hardcoding values in IO is the right solution. Maybe you're asking for an update to default values used in the Request fields? But keep in mind that things aren't just Trinary, all settings need to be considered when adjusting defaults.

adivardi:
I am not sure I follow. The configured values you linked are for the thresholds. I am not suggesting we change these thresholds at all, nor removing their configurability.
The part which causes the issue, is that the pixel values saved are always hardcoded (lines 542, 544, 546 & 548 in [here](https://github.com/ros-navigation/navigation2/blob/main/nav2_map_server/src/map_io.cpp#L540-L549)). They cannot be adjusted right now.
The way this was [handled in ROS 1](https://github.com/ros-planning/navigation/blob/noetic-devel/map_server/src/map_saver.cpp#L96-L116), is that the threshold values saved to yaml were hard-coded, **after** the thresholding, so the thresholding is not affected and can be configured by the user as now. We can limit this change to saving Trinary map.
This means it should affect only Trinary yaml files saved by the map_saver after the new code is introduced. It also doesn't limit configurability, it simply write to file other values than the values used for thresholding (which are not applicable anymore as the hard-coded pixels values' are not guaranteed to lie between them)

SteveMacenski:
>  it simply write to file other values than the values used for thresholding 

It looks like ROS 1 deals with this from a [parameter for loading to override](https://github.com/ros-planning/navigation/blob/9ad644198e132d0e950579a3bc72c29da46e60b0/map_server/src/main.cpp#L181-L182) when desired. I think that makes sense here to do. 

I would not accept a PR that adds in hardcoded values generally speaking in the codebase, but configurable values to override the yaml settings when the parameters are set (otherwise use yaml values) seems reasonable and in line with previous behavior in ROS 1. 

---

Also, what if `negate` is set?

adivardi:
> > it simply write to file other values than the values used for thresholding
> 
> It looks like ROS 1 deals with this from a [parameter for loading to override](https://github.com/ros-planning/navigation/blob/9ad644198e132d0e950579a3bc72c29da46e60b0/map_server/src/main.cpp#L181-L182) when desired. I think that makes sense here to do.

This is the map loading function. While my PR originally modified this code, I already accepted that modifying the map loading is not a good idea, and suggested that we modify the map saving.

I think that because we are discussing in a PR and not in an issue, we keep referring back to what this PR does. I would like us to forget about it, and concentrate on solving the underlying issue.
So I will first close this PR. Would you prefer moving this discussion to a new issue?

-----

> Also, what if `negate` is set?

`negate` only exist in map loading.

adivardi:
Now that we forgot about this PR's bad solution :sweat_smile: :

I am proposing a similar solution to that of ROS 1, only modifying the saved yaml file, from [here](https://github.com/ros-planning/navigation/blob/9ad644198e132d0e950579a3bc72c29da46e60b0/map_server/src/map_saver.cpp#L95-L116), which hard-codes the saved yaml file. However,  instead of hard-coding it fully, we will only hardcode `occupied_thresh` and `free_thresh` and only if the map_mode is Trinary. [negate is already hard-coded in nav2](https://github.com/ros-planning/navigation2/blob/8d69483309e0c5564ed2a692dfb02f0a44128f0f/nav2_map_server/src/map_io.cpp#L604)

I know you are against hard-coded values, and I totally get it. However, the saved pixel values are already [hard-coded](https://github.com/ros-planning/navigation2/blob/8d69483309e0c5564ed2a692dfb02f0a44128f0f/nav2_map_server/src/map_io.cpp#L542) in Trinary mode, so I really do not see why the thresholds should not be as well.

We could of course make the saved thresholds values into prams of the `map_saver`. That would work. But there would be no reason to modify them (e.g. change the `free threshold` between 0 and 0.196 won't have any effect, setting it higher than 0.196 will just cause UNKNOWN pixels to become FREE when the yaml+map are loaded)

Do let me know if you prefer moving the discussion to a new issue

SteveMacenski:
OK - please open a PR for the differed values for the map saver trinary mode. I'm OK with that.  Sorry we went the long way around the block on this one :sweat_smile: 

> [negate is already hard-coded in nav2](https://github.com/ros-planning/navigation2/blob/8d69483309e0c5564ed2a692dfb02f0a44128f0f/nav2_map_server/src/map_io.cpp#L604)

That actually is legit to hardcode because the map saver doesn't invert the image pixel values. Its a literal non-feature of the map saver to support.

Issue #5094, opened 2025-04-18
Use of stateful goal checker in Rotation Shim Controller


Ayush1285:
#### Feature description
Problem:
The rotation shim controller uses normal tolerance-based goal checking for the x-y goal. Once the tolerance is matched, it starts the rotation cycle.
Code: https://github.com/ros-navigation/navigation2/blob/5423eba732ed16b4f97a7353027c1f444b6edc60/nav2_rotation_shim_controller/include/nav2_rotation_shim_controller/tools/utils.hpp#L33C1-L33C41
The problem we are facing is that due to localisation shifts, it goes back into primary controller mode if the x-y shift is outside the tolerance, causing back-and-forth Rotation and translation switches. 

#### Implementation considerations

Solution:
Since goal_checker already implements the stateful feature, and goal_checker pointer is available in the `withinPositionGoalTolerance`,  we can use `isGoalReached` function of the goal checker instead of comparing ourselves. `isGoalReached` will take care of the stateful feature. Ref:   https://github.com/ros-navigation/navigation2/blob/5423eba732ed16b4f97a7353027c1f444b6edc60/nav2_controller/plugins/simple_goal_checker.cpp#L109

SteveMacenski:
Please submit a PR and I can take a look! It may be that this was implemented in RPP before it was available in the Goal Checker API so it was strapped in place and we missed a potential usecase spot. 

Note that `isGoalReached` looks at the orientation as well in that implementation, not just the translation. The RPP use of `withinPositionGoalTolerance` checks just the XY to know if we want to enact a rotation, so we need to only look at XY (and also have a stateful setting to fix the problem you mention, either through the GoalChecker use or our own implementation within RPP if need be).

SteveMacenski:
@Ayush1285 any update?

PR #5140, opened 2025-05-09
Fix - #5094 rotation shim stateful goal checking

PrabhavSaxena:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #5094 ) |
| Robotic platform tested on | (gazebo simulation of Tally) |

---

* Added a new plugin called position_goal_checker, which checks only the xy tolerance and ignores the yaw tolerance. Used the same plugin internally in Rotation Shim Controller.

* Should I add position_goal_checker in the documentation, as it is a new plugin?

* Validated RotationShimController in turtlebot simulation.

---

#### For Maintainers: 


SteveMacenski:
I don't think this works from the discussion in https://github.com/ros-navigation/navigation2/issues/5094#issuecomment-2816288519 can you comment on that?

PrabhavSaxena:
> I don't think this works from the discussion in [#5094 (comment)](https://github.com/ros-navigation/navigation2/issues/5094#issuecomment-2816288519) can you comment on that?

The issue occurs because the rotation shim controller uses position-only tolerance checking to decide when to switch to rotation mode. However, due to localization shifts, if the robot briefly moves outside the XY tolerance during rotation, it switches back to the primary controller, causing oscillation.

The change I've implemented uses the existing stateful behavior of goal_checker->isGoalReached().

1. First check if the robot is within both XY and angular tolerances
2. Remember when the robot has entered the goal region
3. Continue to return true for subsequent calls, even with minor position shifts

This "memory" is exactly what's needed to solve the oscillation problem - once the controller enters rotation mode, it will stay there despite minor localization shifts.

One potential concern might be that isGoalReached() also checks orientation. For a more complete solution, we might need to:

1. Either implement a truly position-only stateful check
2. Or modify the withinPositionGoalTolerance to maintain its own state once position tolerance is met

I'd appreciate your guidance on which approach would better align with the architecture of Nav2.

SteveMacenski:
> Or modify the withinPositionGoalTolerance to maintain its own state once position tolerance is met

I think this is what we should do ultimately. The goal checker that is given to the controller from the server can be any plugin considering any number of criteria. We can't rely on its implementation for internal controller behavior. We could essentially create a "stateful position goal checker" style object (or even that _literal_ object that we directly instantiate) to use in the controller to meet this need.

fivemserverhostings:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PrabhavSaxena:
> > Or modify the withinPositionGoalTolerance to maintain its own state once position tolerance is met
> 
> I think this is what we should do ultimately. The goal checker that is given to the controller from the server can be any plugin considering any number of criteria. We can't rely on its implementation for internal controller behavior. We could essentially create a "stateful position goal checker" style object (or even that _literal_ new plugin that we directly instantiate and use internally) to use in the controller to meet this need.

I have added the plugin to check the position only. I am unsure about what do you mean by using the plugin internally in the rotation shim controller. I was thinking of adding.

bool position_goal_reached = false;
inline void resetPositionGoalState(){
    position_goal_reached = false;
  }

in tools/utils.hpp, then updating
void RotationShimController::setPlan(const nav_msgs::msg::Path & path)
{
  path_updated_ = true;
  current_path_ = path;
  primary_controller_->setPlan(path);
  utils::resetPositionGoalState();
}

nav2_controller/include/nav2_controller/plugins/position_goal_checker.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include "nav2_costmap_2d/costmap_2d_ros.hpp"
+
+namespace nav2_controller
+{
+
+/**
+ * @class PositionGoalChecker
+ * @brief Goal Checker plugin that only checks XY position, ignoring orientation
+ */
+class PositionGoalChecker : public nav2_core::GoalChecker
```
SteveMacenski, line 33: Needs to be added in docs to this list with its configuration page: https://docs.nav2.org/configuration/packages/configuring-controller-server.html#provided-plugins. Here as well: https://docs.nav2.org/plugins/index.html.

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -69,6 +71,10 @@ void RotationShimController::configure(
     node, plugin_name_ + ".rotate_to_goal_heading", rclcpp::ParameterValue(false));
   nav2_util::declare_parameter_if_not_declared(
     node, plugin_name_ + ".closed_loop", rclcpp::ParameterValue(true));
+  nav2_util::declare_parameter_if_not_declared(
+    node, plugin_name_ + ".position_checker.xy_goal_tolerance", rclcpp::ParameterValue(0.25));
```
SteveMacenski, line 75: These are already declared within the plugin, this is not required (?)

SteveMacenski: How do we set these values to match the values from the Controller-provided XY goal tolerance?

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -168,11 +178,7 @@ geometry_msgs::msg::TwistStamped RotationShimController::computeVelocityCommands
         throw std::runtime_error("Failed to transform pose to base frame!");
       }
 
-      if (utils::withinPositionGoalTolerance(
```
SteveMacenski, line 171: Should this be removed from the package if not used?

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -168,11 +178,7 @@ geometry_msgs::msg::TwistStamped RotationShimController::computeVelocityCommands
...
-          goal_checker,
-          pose.pose,
-          sampled_pt_goal.pose))
-      {
+      if (position_goal_checker_->isGoalReached(pose.pose, sampled_pt_goal.pose, velocity)) {
```
SteveMacenski, line 181: We need to use the provided goal checker's XY tolerance, not the one from this plugin.

SteveMacenski (changes requested):
Last thing is that we need this PR to also target `main` so that Rolling, Jazzy, etc can have these changes also reflected. That way when you eventually upgrade, this is also included. All new features must be in `main`.

PrabhavSaxena:
Hi, I will resolve the comments right now, I had few questions about the same.
1. Should I direct this PR to main branch (which requires some effort because I made this PR out of humble branch) or should I merge this PR (with humble) and create a new one which targets main branch. Or if there is some easier way to make my changes target the main branch.
2. Should I remove the tools/utils.hpp file in nav2_rotation_shim_controller as we are no longer using the withinPositionGoalTolerance function?

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: I have added a function in position_goal_checker which sets the xy_tolerances. So this will not be needed and I can set the XY_tolerance of the position_goal_checker from the controller provided XY_tolerance

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: Added the same

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -353,6 +364,9 @@ void RotationShimController::setPlan(const nav_msgs::msg::Path & path)
   path_updated_ = true;
   current_path_ = path;
   primary_controller_->setPlan(path);
+  if (position_goal_checker_) {
+    position_goal_checker_->reset();
+  }
```
SteveMacenski, line 369: For main we should also have the goal checker reset when we reset the controller (which happens at the end of a planning request https://github.com/ros-navigation/navigation2/blob/main/nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp#L401-L405)

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -122,6 +124,9 @@ void RotationShimController::activate()
     std::bind(
       &RotationShimController::dynamicParametersCallback,
       this, std::placeholders::_1));
+  if (position_goal_checker_) {
```
SteveMacenski, line 127: If configuration has happened, isn't this always a non-nullptr?

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -168,11 +174,16 @@ geometry_msgs::msg::TwistStamped RotationShimController::computeVelocityCommands
         throw std::runtime_error("Failed to transform pose to base frame!");
       }
 
-      if (utils::withinPositionGoalTolerance(
-          goal_checker,
-          pose.pose,
-          sampled_pt_goal.pose))
-      {
+      if (goal_checker) {
```
SteveMacenski, line 177: If configuration has happened, isn't this always a non-nullptr? I don't think this can be called otherwise

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -168,11 +174,16 @@ geometry_msgs::msg::TwistStamped RotationShimController::computeVelocityCommands
...
+        geometry_msgs::msg::Pose pose_tolerance;
+        geometry_msgs::msg::Twist vel_tolerance;
+        goal_checker->getTolerances(pose_tolerance, vel_tolerance);
+        if (position_goal_checker_) {
```
SteveMacenski, line 181: Ditto

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -353,6 +364,9 @@ void RotationShimController::setPlan(const nav_msgs::msg::Path & path)
   path_updated_ = true;
   current_path_ = path;
   primary_controller_->setPlan(path);
+  if (position_goal_checker_) {
```
SteveMacenski, line 367: Ditto

SteveMacenski:
1. Leave this PR and lets get this approved first. Then once its done, you should apply this to `main` (with a tweak I mention in a comment above) and we can merge all of them together
2. Yes please unless you think it has reusability elsewhere :-)

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: truee, got it. Removing the redundant check

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: done

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: done

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: done

PrabhavSaxena:
> 1. Leave this PR and lets get this approved first. Then once its done, you should apply this to `main` (with a tweak I mention in a comment above) and we can merge all of them together
> 
>     2. Yes please unless you think it has reusability elsewhere :-)

cool, got it :)

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
```diff
@@ -122,6 +123,7 @@ void RotationShimController::activate()
     std::bind(
       &RotationShimController::dynamicParametersCallback,
       this, std::placeholders::_1));
+      position_goal_checker_->reset();
```
SteveMacenski, line 126:
```suggestion
    position_goal_checker_->reset();
```

SteveMacenski (changes requested):

SteveMacenski (approved):
I look forward to the `main` PR :-)

PrabhavSaxena:
> I look forward to the `main` PR :-)

I created a PR for the same, here is the link to it [](https://github.com/ros-navigation/navigation2/pull/5162)

nav2_rotation_shim_controller/src/nav2_rotation_shim_controller.cpp
PrabhavSaxena: done

PR #5149, opened 2025-05-12
[nav2_mppi_controller] Reset motion model with optimizer (for param change)

adivardi:
---

## Basic Info

| ------ | ----------- |

---

- Fix a bug where the motion_model is not reset when the optimizer is reset. This causes parameters changes to not be updated in the motion model, even though they are updated in the optimizer.
- I noticed the MPPI behavior getting worse when dynamically changing control constraints (in my case `az_max`)

None

The old fashion way - using prints:
Added this print after the line changed in this PR.
```
  std::cout << "Optimizer::reset: " << "\n settings_.constraints.az_max: "
            << settings_.constraints.az_max
            << "\n settings_.base_constraints.az_max: " << settings_.base_constraints.az_max
            << "\n motion_model_.az_max " << motion_model_->control_constraints_.az_max << std::endl;
```
The call `ros2 param set /controller_server FollowPath.az_max 1.0`
Without the change, the last line value is not updated to the new value requested.
You can also add prints in places such as [optimizer](https://github.com/enwaytech/navigation2/blob/main/nav2_mppi_controller/src/optimizer.cpp#L248) and [motion_model](https://github.com/enwaytech/navigation2/blob/main/nav2_mppi_controller/include/nav2_mppi_controller/motion_models.hpp#L78)

I am happy to hear suggestions on how to test this more properly.

---

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/src/optimizer.cpp
+++ b/nav2_mppi_controller/src/optimizer.cpp
@@ -147,6 +147,8 @@ void Optimizer::reset()
   generated_trajectories_.reset(settings_.batch_size, settings_.time_steps);
 
   noise_generator_.reset(settings_, isHolonomic());
+  motion_model_->initialize(settings_.constraints, settings_.model_dt);
+
   RCLCPP_INFO(logger_, "Optimizer reset");
 }
 


SteveMacenski (approved):

adivardi:
Could you please sync this to Jazzy? It has the same issue

SteveMacenski:
In progress

Issue #4403, opened 2024-06-08
Nav2 Docking: Add Option for Reversing into Dock


SteveMacenski:
Currently, we detect the dock and use the vision control loop to drive into the dock. If you have sensors in the back, you can set the staging pose to reversed & `dock_backwards` to dock. If you do **not** have sensors in back, there should be an option to `dock_backwards` but stage forward, thus needs:
- Detect dock
- Rotate to 180 deg heading
- Reverse into the dock (without updates to detected pose)

We'd just need to add the option for that 180 maneuver

SantoshGovindaraj:
Hello @SteveMacenski, I would like to contribute to this issue as my first step in open-source. 
I see that #4406 would be good starting point for me to understand the premise. 

Let me know some additional reference that would be helpful to better understand docking.

SantoshGovindaraj:
Hi @SteveMacenski, I wanted to kindly follow-up. I am really interested in contributing and would love to help move this forward if it's still open or if there are any updates. 

Please let me know if there's anything I can assist with or if the issue is now closed. Looking forward to hearing from you.
Thank you in advance. 🙌

SteveMacenski:
Hi,

We'd love the contribution! 

This issue is still open and describes the creation of a behavior that allows the robot to detect the dock while pointed forward towards the dock (where a sensor may exist), rotate 180 to be in reverse (where a sensor may not exist) and then back into the dock based on that single-point detection. 

This is likely less accurate than having a sensor, but if you're hardware constrained or do not need very fine accuracies, this is a valuable feature addition. Essentially, we want a parameter that enacts this behavior. After we detect the dock, we check if we should rotate before reversing into the dock. We should also check when we grab the parameters if its properly configured (i.e. this parameter is only valid when `dock_backward = true`) 

Does that make sense? I think that should be enough to get started!  You can use the Nav2 Rviz docking panel as well for testing of this feature to make it really easy :-)

CesarBear:
@SteveMacenski I will take this one :D

SteveMacenski:
Ok great! Its probably worth reviewing the current thread in #4749 where this was started but didn't get completed. You can probably either start from that and read the review comments or take some inspiration from it!

padhupradheep:
@YoulunPeng do you have any updates here?

ajtudela:
@padhupradheep I'm finishing this up. I'll submit the PR in a couple of days. I'm just writing the final tests.

[...]

Issue #68, opened 2018-09-07
Add -Werror flag to CMakeLists.txt files 


mkhansen-intel:
To ensure we are warning free, we should add the -Werror flag to our CMakeLists.txt files.

mkhansen-intel:
This is merged in #96

PR #5153, opened 2025-05-13
Backward docking without sensors 

ajtudela:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4403, [#68](https://github.com/open-navigation/opennav_docking/issues/68) and [PR4749](https://github.com/ros-navigation/navigation2/pull/4749)
| Primary OS tested on | Ubuntu 24.04|

---

* Enhanced the docking server to allow backward docking without continuous pose refinement, using only the initially detected docking pose.
* Added a parameter ```backward_blind``` to the SimpleChargingDock and SimpleNonChargingDock.

---

#### For Maintainers: 

### DIFF
--- a/nav2_docking/README.md
+++ b/nav2_docking/README.md
@@ -201,7 +201,8 @@ For debugging purposes, there are several publishers which can be used with RVIZ
 | controller_frequency         | Control frequency (Hz) for vision-control loop          | double | 50.0      |
 | initial_perception_timeout   | Timeout (s) to wait to obtain initial perception of the dock | double | 5.0   |
 | wait_charge_timeout          | Timeout (s) to wait to see if charging starts after docking  | double | 5.0  |
-| dock_approach_timeout        | timeout (s) to attempt vision-control approach loop    | double |  30.0      |
+| dock_approach_timeout        | Timeout (s) to attempt vision-control approach loop    | double |  30.0      |
+| rotate_to_dock_timeout       | Timeout (s) to attempt rotate-to-dock loop             | double |  10.0      |
 | undock_linear_tolerance      | Tolerance (m) to exit the undocking control loop at staging pose    | double |  0.05      |
 | undock_angular_tolerance     | Angular tolerance (rad) to exit undocking loop at staging pose    | double |  0.05      |
 | max_retries        | Maximum number of retries to attempt    | int |  3      |
--- a/nav2_docking/README.md
+++ b/nav2_docking/README.md
@@ -207,7 +207,6 @@ For debugging purposes, there are several publishers which can be used with RVIZ
 | max_retries        | Maximum number of retries to attempt    | int |  3      |
 | base_frame        | Robot's base frame for control law   | string |  "base_link"      |
 | fixed_frame        | Fixed frame to use, recommended to be a smooth odometry frame **not** map   | string |  "odom"      |
-| backward_blind        | Initial forward detection, then dock in reverse | bool |  false      |
 | odom_topic        | The topic to use for the odometry data | string |  "odom"      |
 | rotation_angular_tolerance  | Angular tolerance (rad) to exit the rotation loop when backward_blind is enable | double | 0.05      |
 | dock_prestaging_tolerance  |  L2 distance in X,Y,Theta from the staging pose to bypass navigation | double |  0.5      |
@@ -256,6 +255,7 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
 | staging_x_offset        | Staging pose offset forward (negative) of dock pose (m)    | double |  -0.7    |
 | staging_yaw_offset        | Staging pose angle relative to dock pose (rad)    | double |  0.0    |
 | dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"  or "backward"    |
+| backward_blind        | Initial forward detection, then dock in reverse | bool |  false      |
 
 Note: The external detection rotation angles are setup to work out of the box with Apriltags detectors in `image_proc` and `isaac_ros`.
 
--- a/nav2_docking/README.md
+++ b/nav2_docking/README.md
@@ -203,11 +203,13 @@ For debugging purposes, there are several publishers which can be used with RVIZ
 | wait_charge_timeout          | Timeout (s) to wait to see if charging starts after docking  | double | 5.0  |
 | dock_approach_timeout        | timeout (s) to attempt vision-control approach loop    | double |  30.0      |
 | undock_linear_tolerance      | Tolerance (m) to exit the undocking control loop at staging pose    | double |  0.05      |
-| undock_angular_tolerance     | Angular Tolerance (rad) to exist undocking loop at staging pose    | double |  0.05      |
+| undock_angular_tolerance     | Angular tolerance (rad) to exit undocking loop at staging pose    | double |  0.05      |
 | max_retries        | Maximum number of retries to attempt    | int |  3      |
 | base_frame        | Robot's base frame for control law   | string |  "base_link"      |
 | fixed_frame        | Fixed frame to use, recommended to be a smooth odometry frame **not** map   | string |  "odom"      |
 | backward_blind        | Initial forward detection, then dock in reverse | bool |  false      |
+| odom_topic        | The topic to use for the odometry data | string |  "odom"      |
+| rotation_angular_tolerance  | Angular tolerance (rad) to exit the rotation loop when backward_blind is enable | double | 0.05      |
 | dock_prestaging_tolerance  |  L2 distance in X,Y,Theta from the staging pose to bypass navigation | double |  0.5      |
 | dock_plugins  | A set of dock plugins to load | vector<string> |  N/A      |
 | dock_database  |  The filepath to the dock database to use for this environment | string |  N/A  |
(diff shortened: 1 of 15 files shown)


nav2_docking/README.md
```diff
@@ -251,6 +255,7 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
 | staging_x_offset        | Staging pose offset forward (negative) of dock pose (m)    | double |  -0.7    |
 | staging_yaw_offset        | Staging pose angle relative to dock pose (rad)    | double |  0.0    |
 | dock_direction        | Whether the robot is docking with the dock forward or backward in motion | string |  "forward"  or "backward"    |
+| backward_blind        | Initial forward detection, then dock in reverse | bool |  false      |
```
SteveMacenski, line 258: Not necessarily asking for a change, but can we brainstorm on names here? I like this as a short hand you and I talking about this can understand, but I'm not sure I love this for clarity of a random user looking for the first time.

We're trying to communicate that we do initial detection forward, rotate to the heading of the detection and use that for deadreckoning in reverse (when dock_direction is backwards). Maybe `deadreckon` should be part of it? Both for this param name and for the internal APIs / bool instance.  

Also in the docs we should be clear what staging pose orientation direction should be used in this case (forward, I assume, even though docking backward).

nav2_docking/opennav_docking/src/controller.cpp
```diff
@@ -117,6 +125,23 @@ bool Controller::computeVelocityCommand(
   return isTrajectoryCollisionFree(pose, is_docking, backward);
 }
 
+geometry_msgs::msg::Twist Controller::computeRotateToHeadingCommand(
+  const double & angular_distance_to_heading,
+  const geometry_msgs::msg::Twist & current_velocity,
+  const double & dt)
+{
+  geometry_msgs::msg::Twist cmd_vel;
```
SteveMacenski, line 133: Can we use the angular acceleration and target orientation to accelerate / decelerate towards the goal? https://github.com/ros-navigation/navigation2/blob/main/nav2_behaviors/plugins/spin.cpp#L140

That way we slow down towards the goal instead of a sudden jerk when we stop and reach the appropriate angle. That'll introduce less odometry slippage

nav2_docking/opennav_docking/include/opennav_docking/docking_server.hpp
```diff
@@ -248,11 +255,16 @@ class DockingServer : public nav2_util::LifecycleNode
   std::optional<bool> dock_backwards_;
   // The tolerance to the dock's staging pose not requiring navigation
   double dock_prestaging_tolerance_;
+  // Enable docking with initial detection only, without any update
+  bool backward_blind_;
```
SteveMacenski, line 259: Unused?

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -284,15 +295,20 @@ void DockingServer::dockRobot()
...
       dock_backwards_.value() :
       (dock->plugin->getDockDirection() == opennav_docking_core::DockDirection::BACKWARD);
+    bool backward_blind = dock->plugin->isBackwardBlind();
 
     // Docking control loop: while not docked, run controller
     rclcpp::Time dock_contact_time;
     while (rclcpp::ok()) {
       try {
+        // Perform pure rotation to dock orientation
+        if (backward_blind) {
```
SteveMacenski, line 309:
```suggestion
        if (dock->plugin->isBackwardBlind()) {
```

This is just a bool, we don't need `bool backward_blind = dock->plugin->isBackwardBlind();` stored beforehand

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -447,7 +492,8 @@ bool DockingServer::approachDock(
     }
 
     // Update perception
-    if (!dock->plugin->getRefinedPose(dock_pose, dock->id)) {
+    bool backward_blind = dock->plugin->isBackwardBlind();
+    if (!backward_blind && !dock->plugin->getRefinedPose(dock_pose, dock->id)) {
```
SteveMacenski, line 496: Ditto, just use `dock->plugin->isBackwardBlind()` here

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -424,6 +440,35 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
   }
 }
 
+void DockingServer::rotateToDock(const geometry_msgs::msg::PoseStamped & dock_pose)
```
SteveMacenski, line 443: Design question: Can we add this to the controller? This feels like a controller action perhaps (but open to keep here if you think it should be)

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -424,6 +440,35 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
...
+{
+  rclcpp::Rate loop_rate(controller_frequency_);
+  const double dt = 1.0 / controller_frequency_;
+  auto target_pose = dock_pose;
+  target_pose.pose.orientation = nav2_util::geometry_utils::orientationAroundZAxis(
+    tf2::getYaw(target_pose.pose.orientation) + M_PI);
+
+  while (rclcpp::ok()) {
```
SteveMacenski, line 451: Introduce a `rotate_to_dock_timeout` parameter to compliment the other timeouts for while loops. While loops scare me bc they're infinite so I want a fallback to get out of them after a generous time period in case a problem occurs

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -424,6 +440,35 @@ void DockingServer::doInitialPerception(Dock * dock, geometry_msgs::msg::PoseSta
...
+    auto robot_pose = getRobotPoseInFrame(dock_pose.header.frame_id);
+    auto angular_distance_to_heading = angles::shortest_angular_distance(
+      tf2::getYaw(robot_pose.pose.orientation), tf2::getYaw(target_pose.pose.orientation));
+    if (fabs(angular_distance_to_heading) < rotation_angular_tolerance_) {
+      break;
+    }
+
+    auto current_vel = std::make_unique<geometry_msgs::msg::TwistStamped>();
```
SteveMacenski, line 459: Design question: I think because we're getting updates on the dock pose orientation in the robot frame from TF, we can make the rotation closed loop if we do my suggestion in the controller of looking at hte acceleration / deceleration as a function of the distance to the goal orientation instead. That would make this open-loop and not rely on the odometry data (but closed loop in the sense that we know where we are from TF to accelerate / decelerate based off of). 

That would simplify this loop and remove the odom subscriber

SteveMacenski (changes requested):

nav2_docking/README.md
SteveMacenski: `deadreckon_backward` perhaps? I think `_docking` is a bit redundant after some thought given the server. 

Or `rotate_to_dock`? Is there anything that **requires** the 180 rotation to happen and not continue to detect? It could be that they prefer to stage forward and then want the docking server to rotate to reverse and backup (both blind or non-blind). I don't _think_ anything in your PR prevents future detections from being used after that point **or** no further detections are given. The behavior could be more generalized to not just be for single-shot detection but any case where we might want to rotate to dock (though document in the parameters that intention for allowing backward docking without sensor coverage as a main rationale for it).

P.S. Should we also rotate back to the initial staging pose after we `undock`? I think perhaps we should so we return to the same pose as initially staged at.

ajtudela: I agree that ```rotate_to_dock``` seems better, and it could be used in future for more generalized behaviour.

P.D.: Yeap, rotating back to the initial staging pose after undocking would improve consistency.

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @ajtudela can you respond to this one? We have other rotate to heading behaviors in other places that do so in this way without the odom smoother

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -652,6 +714,13 @@ void DockingServer::undockRobot()
     geometry_msgs::msg::PoseStamped staging_pose =
       dock->getStagingPose(dock_pose.pose, dock_pose.header.frame_id);
 
+    // If we performed a rotation before docking backward, we must rotate the staging pose
```
SteveMacenski, line 717: Sanity check for me: can you test docking / undocking 2x in a row and this all works out ok :smile:

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -447,7 +509,7 @@ bool DockingServer::approachDock(
     }
 
     // Update perception
-    if (!dock->plugin->getRefinedPose(dock_pose, dock->id)) {
+    if (!dock->plugin->shouldRotateToDock() && !dock->plugin->getRefinedPose(dock_pose, dock->id)) {
```
SteveMacenski, line 512: Can we invert this to be `!dock->plugin->getRefinedPose(dock_pose, dock->id) && !dock->plugin->shouldRotateToDock()`? 

I think that would enable us to get new refined poses when backing up when they are available (if not blind). If we do the opposite, then we check if rotating and if so, we so skip the second check which does the update.

SteveMacenski (commented):
Doc update in the readme to mention we can do this + in the migration guide + in the configuration guide PR is important to make it clear this is now possible

nav2_docking/opennav_docking/src/docking_server.cpp
ajtudela: Of course!
I've recorded a quick video. This shows the smoking test twice in a row: docking, preemption, rotating, approaching, waiting for the charge, undocking, rotating, and so on.
[Uploading Grabación de pantalla desde 2025-05-16 09-31-54.webm…]()

nav2_docking/opennav_docking/src/docking_server.cpp
ajtudela: Sorry! I forgot.

I have mixed feelings about this.

On the one hand, I like the simplicity and independence of the odometry for the docking/undocking process.

The other rotate to heading behaviors are in the regulated_pure_pursuit and rotation_shim_controller; both of these get the current angular velocity from the controller and the odom subscriber. If we remove the odom subscriber, how can we obtain the angular velocity?

On the other hand, without odometry, we would only be relying on the quality of the TF. However, angular errors could be corrected by the smooth control law.

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -685,6 +754,12 @@ void DockingServer::undockRobot()
           !dock_backward))
       {
         RCLCPP_INFO(get_logger(), "Robot has reached staging pose");
+
+        // Perform a 180º to the original staging pose
```
SteveMacenski, line 758: Whoops, this should be done before the log above

>         RCLCPP_INFO(get_logger(), "Robot has reached staging pose");

SteveMacenski (approved):
Last minor change and we can merge!

SteveMacenski (approved):

PR #5154, opened 2025-05-13
Adding parameter util to node utils

MarcoMatteoBassa:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Independent |

---

The added utility tackles 2 issues:
- Reduces the amount of boilerplate code required to declare and get parameters, and consequently makes it less likely to introduce typos: it combines declare_parameter_if_not_declared and get_parameter in a single instruction.

Example:

```
    node,
    "max_rotational_vel", rclcpp::ParameterValue(1.0));
  node->get_parameter("max_rotational_vel", max_rotational_vel_);
```

Can be rewritten as

`max_rotational_vel_ = nav2_util::declare_or_get_param(node, "max_rotational_vel", 1.0);`

- Helps tracking typos / mismatching namespaces in the parameters overrides: if a parameter is not found in the overrides, resulting in the default value being used, the utility can optionally print a WARN message. This can often help the developers catch configuration mistakes, and is particularly useful when relying on the default value likely results in degraded performance.

The utility can be used either through a generic node pointer, or (useful when using derived nodes) using the internal node interfaces.

- Added a unit test.
- As I didn't add the utilty to any node yet, it won't affect any existing behavior.

---

- The utility can be used for refactoring purposes or for the development of new modules.

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/node_utils.hpp
+++ b/nav2_util/include/nav2_util/node_utils.hpp
@@ -142,9 +142,14 @@ rclcpp::ParameterValue declare_or_get_parameter(
   const rclcpp::ParameterType & param_type,
   const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
 {
-  return node->has_parameter(parameter_name) ?
-         node->get_parameter(parameter_name).get_parameter_value() :
-         node->declare_parameter(parameter_name, param_type, parameter_descriptor);
+  if (node->has_parameter(parameter_name)) {
+    try {
+      return node->get_parameter(parameter_name).get_parameter_value();
+    } catch (const std::exception & exception) {
+      return rclcpp::ParameterValue();
+    }
+  }
+  return node->declare_parameter(parameter_name, param_type, parameter_descriptor);
 }
 
 using NodeParamInterfacePtr = rclcpp::node_interfaces::NodeParametersInterface::SharedPtr;
@@ -210,8 +215,7 @@ ParamType declare_or_get_parameter(
   const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
 {
   return declare_or_get_parameter(node->get_logger(), node->get_node_parameters_interface(),
-      parameter_name,
-      default_value, warn_if_no_override, parameter_descriptor);
+      parameter_name, default_value, warn_if_no_override, parameter_descriptor);
 }
 
 /// Gets the type of plugin for the selected node and its plugin
--- a/nav2_util/include/nav2_util/node_utils.hpp
+++ b/nav2_util/include/nav2_util/node_utils.hpp
@@ -127,29 +127,31 @@ void declare_parameter_if_not_declared(
 /// Declares a parameter with the specified type if it was not already declared
 /// and returns its value if available.
 /* Declares a parameter with the specified type if it was not already declared.
- * If the parameter was overridden, its value is returned, otherwise a
- * parameter not set type is returned.
+ * If the parameter was overridden, its value is returned, otherwise an
+ * rclcpp::exceptions::InvalidParameterValueException is thrown
  *
  * \param[in] node A node in which given parameter to be declared
  * \param[in] parameter_name Name of the parameter
  * \param[in] param_type The type of parameter
  * \param[in] parameter_descriptor Parameter descriptor (optional)
+ * \return The value of the parameter or an exception
  */
-template<typename NodeT>
-rclcpp::ParameterValue declare_or_get_parameter(
+template<typename ParameterT, typename NodeT>
+ParameterT declare_or_get_parameter(
   NodeT node,
   const std::string & parameter_name,
   const rclcpp::ParameterType & param_type,
   const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
 {
   if (node->has_parameter(parameter_name)) {
-    try {
-      return node->get_parameter(parameter_name).get_parameter_value();
-    } catch (const std::exception & exception) {
-      return rclcpp::ParameterValue();
-    }
+    return node->get_parameter(parameter_name).template get_value<ParameterT>();
   }
-  return node->declare_parameter(parameter_name, param_type, parameter_descriptor);
+  auto parameter = node->declare_parameter(parameter_name, param_type, parameter_descriptor);
+  if (parameter.get_type() == rclcpp::ParameterType::PARAMETER_NOT_SET) {
+    std::string description = "Parameter " + parameter_name + " not in overrides";
+    throw rclcpp::exceptions::InvalidParameterValueException(description.c_str());
+  }
+  return parameter.template get<ParameterT>();
 }
 
 using NodeParamInterfacePtr = rclcpp::node_interfaces::NodeParametersInterface::SharedPtr;
@@ -215,7 +217,7 @@ ParamType declare_or_get_parameter(
   const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
 {
   return declare_or_get_parameter(node->get_logger(), node->get_node_parameters_interface(),
-      parameter_name, default_value, warn_if_no_override, parameter_descriptor);
+    parameter_name, default_value, warn_if_no_override, parameter_descriptor);
 }
 
 /// Gets the type of plugin for the selected node and its plugin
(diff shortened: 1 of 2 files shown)


SteveMacenski (commented):
As a potential suggestion, this is `declare_parameter_if_not_declared`

```
template<typename NodeT>
void declare_parameter_if_not_declared(
  NodeT node,
  const std::string & param_name,
  const rclcpp::ParameterValue & param_value,
  const rcl_interfaces::msg::ParameterDescriptor & parameter_descriptor =
  rcl_interfaces::msg::ParameterDescriptor())
{
  if (!node->has_parameter(param_name)) {
    node->declare_parameter(param_name, param_type, parameter_descriptor);
  }
}
```

The `declare_parameter` bit can/will return a `ParameterValue` that we can work with if its not been declared, and otherwise we can add a `get_parameter` as an `else` condition to return the value. 

We could actually just adapt the `declare_parameter_if_not_declared` to have a return type instead and use that as a single-call. Possibly even rename at that point to `declare_or_get_parameter` and use that globally, except the one situation below which is only in a handful of places. 

Note: We also have a version for [parameter types](https://github.com/ros-navigation/navigation2/blob/54ba78f6257480ce6b1001ed18ce8d86cfc9a9f9/nav2_util/include/nav2_util/node_utils.hpp#L114-L125) rather than values for the case in which we want to declare a parameter without a default value but only declare its type. We'd have to keep that in place with the current workflow of declare & get I think. 

What do you think?

MarcoMatteoBassa:
Hi, yes the util is basically doing that, but in addition it also removes the need to explicitally transform the parameters with rclcpp::ParameterValue() when calling it by using the type template.
The WARN part is also quite handy, but if you feel that would be too spammy, we could default it to false.

I didn't want to modify declare_parameter_if_not_declared to avoid intrusion in the code base and possibly generate unused-return-value warnings, but using an unified strategy later would for sure be beneficial :)

Yes, the Type-Only version of declare_parameter_if_not_declared should stay there. As in this case the parameter might not have a value, I would leave it as void; alternatively we could add a version of declare_or_get returning an optional rclcpp::ParameterValue if an rclcpp::ParameterType is used.

Let me know what you prefer :)

nav2_util/include/nav2_util/node_utils.hpp
```diff
@@ -124,6 +124,70 @@ void declare_parameter_if_not_declared(
...
+ * \param[in] logger Node logging interface
+ * \param[in] param_interface Node parameter interface
+ * \param[in] param_name Name of the parameter
+ * \param[in] default_value Default value of the parameter
+ * \param[in] warn_if_no_override If true, prints a warning whenever the parameter override is missing
+ * \param[in] param_descriptor Optional parameter descriptor
+ * \return The value of the param from the override if existent, otherwise the default value
+ */
+template<typename ParamType>
+ParamType declare_or_get_param(
```
SteveMacenski, line 146:
```suggestion
ParamType declare_or_get_parameter(
```

Maybe just rename to not be shortened

SteveMacenski:
I suppose I'm less tied to a particular implementation and more interested in just removing redundancy as much as possible.  Maybe we don't need to do everything at once :-) It would be nice as you say to have a version that doesn't produce a unused return value warning --  

> Yes, the Type-Only version of declare_parameter_if_not_declared should stay there. As in this case the parameter might not have a value, I would leave it as void; alternatively we could add a version of declare_or_get returning an optional rclcpp::ParameterValue if an rclcpp::ParameterType is used.

We could add in the version but allow `get_parameter` throw in the case it was not declared (as it does). As long as the doxygen mentions this behavior, I think that's OK. So it would return a value if one is given, else throws. That should be a known behavior in that case and I believe all uses in Nav2 wraps those calls in try/catch anyway to capture it and allow for a 1:1 replacement. 

>  The WARN part is also quite handy, but if you feel that would be too spammy, we could default it to false.

I do like default false personally. I'm curious to know why its a warning to use a default value - is your expectation that all nodes are full configured in YAML files? I suppose that is good practice, but would spam the terminal quite alot for existing users that are less disciplined (including `nav2_bringup` currently :smiling_face_with_tear: )

SteveMacenski:
Honestly I totally agree. If I was building applications I would be fully describing my configuration files. However, I don't want to overwhelm people reading the config file for the first few times (its already _alot_)

SteveMacenski (approved):
LGTM, ready to merge on your front?

nav2_util/test/test_node_utils.cpp
```diff
@@ -78,6 +79,28 @@ TEST(DeclareParameterIfNotDeclared, DeclareParameterIfNotDeclared)
   ASSERT_EQ(param, "fred");
 }
 
+TEST(DeclareOrGetParam, DeclareOrGetParam)
+{
+  auto node = std::make_shared<rclcpp::Node>("test_node");
+
+  // test declared parameter
+  node->declare_parameter("foobar", "foo");
+  std::string param = declare_or_get_parameter(node, "foobar", std::string{"bar"});
```
SteveMacenski, line 88:
```suggestion
  std::string param = declare_or_get_parameter(node, "foobar", std::string{"bar"}, true);
```

SteveMacenski (commented):
Just to trigger that warning message to complete coverage metrics

SteveMacenski:
Mhm, there are some tests that are failing, can you rebase? We did make some changes in the default bringup recently related to the failing tests and I'm wondering if there's a caching issue.

MarcoMatteoBassa:
I realized that subsequent calls of declare_or_get by type would result in an exception instead of an uninitialized parameter, so I changed it to consistently return an uninitialized parameter.

PS, a possible way to allow the user to globally choose for a node if he wants warnings for missing overrides, could be to enable it through an optional dedicated boolean parameter, ex by changing it to:
```
template<typename ParamType, typename NodeT>
ParamType declare_or_get_parameter(
  NodeT node, const std::string & parameter_name,
  const ParamType & default_value, const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
{
  const auto ParamBool = rclcpp::ParameterType::PARAMETER_BOOL;
  auto warn_value = declare_or_get_parameter(node, "warn_missing_parameter_overrides", ParamBool);
  const bool warn = warn_value.get_type() == ParamBool && warn_value.template get<bool>();
  return declare_or_get_parameter(node->get_logger(), node->get_node_parameters_interface(),
      parameter_name, default_value, warn, parameter_descriptor);
}
```
In this way there wouldn't be a need to hard-code a choice in the function call, and by default there would be no warning.

nav2_util/include/nav2_util/node_utils.hpp
```diff
@@ -81,47 +81,141 @@ rclcpp::Node::SharedPtr generate_internal_node(const std::string & prefix = "");
...
+  NodeT node,
+  const std::string & parameter_name,
+  const rclcpp::ParameterType & param_type,
+  const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
+{
+  if (node->has_parameter(parameter_name)) {
+    try {
+      return node->get_parameter(parameter_name).get_parameter_value();
+    } catch (const std::exception & exception) {
+      return rclcpp::ParameterValue();
```
SteveMacenski, line 149: A point of discussion: 

Right now the ROS 2 API is documented and expected to throw if you get a type-declared parameter but that parameter wasn't provided. In the places in Nav2 where we do type declarations we rely on the throw behavior to tell us a user didn't give a compulsory parameter to fail bringup. 

Personally, I think this is better to be like your other implementation below where we template by the parameter type to return rather than returning a `ParameterValue` or uninitialized one for a later throw to happen. If we're trying to wrap the details of parameter handling from application code, I think we should fully hide it and making the application code deal with a `rclcpp::ParameterValue` seems clunky.

In short, I think we should also template by the parameter value and have the returns return that direct value like the other implementation. Also, we should have this method throw when the parameter isn't provided rather than returning a default value.

MarcoMatteoBassa: I understand your concern, this format was used because the ROS declare parameter function doesn't throw when it's called with a type the first time, even if the parameter is not overridden, but returns a param not set.
As this function may also call get_parameter however, it is not possible to keep a 1-1 mapping in behavior; so if you prefer the exception, I can always raise an exception if the override isn't there.
I can also use a template and return the C++ type, however, in contrast with the latter case, the caller will have to specify the template type explicitly.

SteveMacenski (approved):
Love it, anything else or can I merge?

MarcoMatteoBassa:
All good from my side :) Maybe I'll do a follow-up proposal for enabling the warning through a parameter, but that might not be very "elegant".

PR #5155, opened 2025-05-13
Publish planned footprints after smoothing

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


SteveMacenski (changes requested):
I think this PR is fatally flawed. The smoother does not consider the full SE2 footprints and the planned footprints is to debug visualize the planner itself, not including the smoother stage. I don't disagree this is useful to move for analyzing your current issue, but I do not believe this is the correct action in general for merging this PR.

Please let me know what you think :-)

tonynajjar:
I get what you mean but I would argue that both could be useful to debug depending on the scenario.
Actually most of the time as a user, I don't really care about the inner workings of SMAC, I rather care to see what my planner is feeding my controller, and that is the smoothed path.
So do you think we should add another '/planned_footprints_smoothed'?

SteveMacenski:
Sure :)

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -574,6 +582,27 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
...
+  if (_debug_visualizations) {
+    // plot footprint path planned for debug
+    if (_planned_footprints_smoothed_publisher->get_subscription_count() > 0) {
+      auto marker_array = std::make_unique<visualization_msgs::msg::MarkerArray>();
+
+      visualization_msgs::msg::Marker clear_all_marker;
+      clear_all_marker.action = visualization_msgs::msg::Marker::DELETEALL;
+      marker_array->markers.push_back(clear_all_marker);
+
+      _planned_footprints_smoothed_publisher->publish(std::move(marker_array));
```
SteveMacenski, line 594: This doesn't need to be a standalone message, it can just be the first marker in the markerarray I believe.

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -574,6 +582,27 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
...
+
+      marker_array = std::make_unique<visualization_msgs::msg::MarkerArray>();
+      for (size_t i = 0; i < plan.poses.size(); i++) {
+        const std::vector<geometry_msgs::msg::Point> edge =
+          transformFootprintToEdges(plan.poses[i].pose, _costmap_ros->getRobotFootprint());
+        marker_array->markers.push_back(createMarker(edge, i, _global_frame, _clock->now()));
```
SteveMacenski, line 600: Store the clock outside of the for loop and use here (so that all poses have the exact same timestamp)

SteveMacenski (commented):
All comments repeated for the Lattice case :-)

nav2_smac_planner/src/smac_planner_hybrid.cpp
tonynajjar: Not sure I get it, the msg has to be a visualization_msgs::msg::MarkerArray, can't be a visualization_msgs::msg::Marker (because that's what the publisher is initialized with)

nav2_smac_planner/src/smac_planner_hybrid.cpp
```diff
@@ -534,19 +543,21 @@ nav_msgs::msg::Path SmacPlannerHybrid::createPlan(
...
     _expansions_publisher->publish(msg);
 
-    // plot footprint path planned for debug
     if (_planned_footprints_publisher->get_subscription_count() > 0) {
+      // Clear all markers first
       auto marker_array = std::make_unique<visualization_msgs::msg::MarkerArray>();
+      visualization_msgs::msg::Marker clear_all_marker;
+      clear_all_marker.action = visualization_msgs::msg::Marker::DELETEALL;
+      marker_array->markers.push_back(clear_all_marker);
+      _planned_footprints_publisher->publish(std::move(marker_array));
```
tonynajjar, line 552: I added this to clear the previous footprints because as it was before, the footprints would just superimpose

SteveMacenski (approved):

PR #5158, opened 2025-05-14
Add DynamicParameterPatterns to pose_pogress_checker plugin

Nils-ChristianIseke:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4907 |

No additional test added.

Apply dynamic parameter patters to the rest of nav2_controller.

#### For Maintainers: 


Nils-ChristianIseke:
@SteveMacenski I'm not sure how you want to structure this for the plugins of nav2_controller.
Before applying changes to the rest of the nav2_controller plugins some feedback would be great (naming, namespaces, general structure).

nav2_controller/include/nav2_controller/plugins/parameter_handler_pose_progress_checker.hpp
```diff
@@ -0,0 +1,65 @@
...
+#define NAV2_CONTROLLER__PLUGINS__PARAMETER_HANDLER_POSE_PROGRESS_CHECKER_HPP_
+
+#include <string>
+#include <vector>
+#include "rclcpp/rclcpp.hpp"
+#include "nav2_controller/plugins/simple_progress_checker.hpp"
+#include "rclcpp_lifecycle/lifecycle_node.hpp"
+#include "nav2_util/node_utils.hpp"
+
+namespace nav2_controller
```
Nils-ChristianIseke, line 25: Naming preferences for namespace (unique for each plugin in nav2_controller).

nav2_controller/plugins/pose_progress_checker.cpp
```diff
@@ -46,7 +46,6 @@ void PoseProgressChecker::initialize(
 
 
   plugin_name_ = plugin_name;
-  logger_ = node->get_logger();
```
Nils-ChristianIseke: Not sure which logger we want here. Same problem exist in nav2_regulated_pure_pursuit_controller where the logger_ is initialize in the nav2_regulated_pure_pursuit_controller
[header](https://github.com/Nils-ChristianIseke/navigation2/blob/33bebad3c1314b0b41fc123d61047116614709bb/nav2_regulated_pure_pursuit_controller/include/nav2_regulated_pure_pursuit_controller/regulated_pure_pursuit_controller.hpp#L215) and reinited in the [source](https://github.com/Nils-ChristianIseke/navigation2/blob/33bebad3c1314b0b41fc123d61047116614709bb/nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp#L54)

SteveMacenski:
For trivial nodes / objects like this, I think it would be fine to define their parameter handler in the scope of its header file (or even inside of the class itself to be a specialization for the class). I don't think we can do that for the main planner/controller/smoother/etc that have alot of parameters and validations going on, but for small things like this it seems overkill.

nav2_controller/include/nav2_controller/plugins/parameter_handler_pose_progress_checker.hpp
SteveMacenski: If we move this into the Progress Checker's header file, in the class itself, you can keep it as simply `ParameterHandler` since its unique to the class.

nav2_controller/include/nav2_controller/plugins/parameter_handler_pose_progress_checker.hpp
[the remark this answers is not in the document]
SteveMacenski: See below :-)

nav2_controller/include/nav2_controller/plugins/pose_progress_checker.hpp
```diff
@@ -31,6 +32,41 @@ namespace nav2_controller
...
+    double required_movement_angle;
+  };
+
+/**
+ * @class nav2_controller::PoseProgressChecker::ParameterHandler
+ * @brief This class handls parameters and dynamic parameter updates for the nav2_controller.
+ */
+  class ParameterHandler
+  {
+public:
```
SteveMacenski, line 46:
```suggestion
  public:
```

nav2_controller/include/nav2_controller/plugins/pose_progress_checker.hpp
```diff
@@ -31,6 +32,41 @@ namespace nav2_controller
...
+    ParameterHandler(
+      rclcpp_lifecycle::LifecycleNode::SharedPtr node,
+      std::string & plugin_name, rclcpp::Logger & logger);
+    ~ParameterHandler();
+    Parameters * getParams() {return &params_;}
+
+protected:
```
SteveMacenski, line 53:
```suggestion
  protected:
```

nav2_controller/include/nav2_controller/plugins/pose_progress_checker.hpp
```diff
@@ -50,18 +86,15 @@ class PoseProgressChecker : public SimpleProgressChecker
...
-   * @brief Callback executed when a parameter change is detected
-   * @param parameters list of changed parameters
-   */
-  rcl_interfaces::msg::SetParametersResult
-  dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters);
+  rclcpp::Logger logger_ {rclcpp::get_logger("PoseProgressChecker")};
+  rclcpp_lifecycle::LifecycleNode::WeakPtr node_;
+  std::unique_ptr<nav2_controller::PoseProgressChecker::ParameterHandler> param_handler_;
 };
+
```
SteveMacenski, line 96:
```suggestion
```

SteveMacenski (changes requested):
I think this works (minus some linting items) for the smaller objects, but for the larger ones I would keep in mind:
* Maybe better to return a shared pointer to the parameters rather than a raw pointer? 
* There are going to be situations where dynamic parameters are on another executor from the main task server thread so we need to lock the changes / access. 
* Maybe atomic for the Parameters object would be sensible in that case, have a lock for accessing values on the value object, or do similar to what we do now and lock the dynamic callbacks and the action server callbacks (param_handler->getLock()) to generally lock up the resources during the entire scope of execution to not keep locking/unlocking

nav2_controller/include/nav2_controller/plugins/pose_progress_checker.hpp
Nils-ChristianIseke: This is enforced by uncrustify at the moment. I will have a look into the uncrustify configurration, also for your other linting comments. This should be ideally covered by the linting tools itself.

SteveMacenski:
Let me know when you want me to take a look again, you can request a new review in the github UX (or just say so in a comment)

nav2_controller/include/nav2_controller/plugins/pose_progress_checker.hpp
Nils-ChristianIseke: uncrustify enforces this.

nav2_controller/include/nav2_controller/plugins/pose_progress_checker.hpp
Nils-ChristianIseke: Seems to be an issue with the uncrustify configuration, as this format is enforced by uncrustify.

SteveMacenski: So this fails uncrustify if you make the change? OK

Nils-ChristianIseke: yes

Nils-ChristianIseke:
>     * Maybe better to return a shared pointer to the parameters rather than a raw pointer?

Changed that, a raw pointer for this is also present in some other parameter_handlers.


>     * There are going to be situations where dynamic parameters are on another executor from the main task server thread so we need to lock the changes / access.
> 
>     * Maybe atomic for the Parameters object would be sensible in that case, have a lock for accessing values on the value object, or do similar to what we do now and lock the dynamic callbacks and the action server callbacks (param_handler->getLock()) to generally lock up the resources during the entire scope of execution to not keep locking/unlocking

I have too look into this in more detail. After a first glimps i don't get the implementation of the mutex locks that are present in the other param handlers.

mini-1235:
Hi @Nils-ChristianIseke, I'll take over this since there hasn't been any recent activity. Thanks a lot for your effort here!

PR #5162, opened 2025-05-15
Addition of position goal checker, fix of rotation shim controller

PrabhavSaxena:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #5094 ) |
| Robotic platform tested on | ( gazebo simulation of Tally) |

---

* Added a new plugin called position_goal_checker, which checks only the xy tolerance and ignores the yaw tolerance. Used the same plugin internally in Rotation Shim Controller.

* Added to documentation [Link](https://github.com/ros-navigation/docs.nav2.org/pull/689)

* Validated RotationShimController in turtlebot simulation

---

#### For Maintainers: 


nav2_rotation_shim_controller/CMakeLists.txt
```diff
@@ -29,6 +30,7 @@ target_include_directories(${library_name}
 )
 target_link_libraries(${library_name} PUBLIC
   ${geometry_msgs_TARGETS}
+  nav2_controller::nav2_controller
```
SteveMacenski, line 33: Don't we want `nav2_controller::position_goal_checker`?

nav2_controller/plugins/position_goal_checker.cpp
```diff
@@ -0,0 +1,147 @@
...
+  xy_goal_tolerance_ = tolerance;
+  xy_goal_tolerance_sq_ = tolerance * tolerance;
+}
+
+rcl_interfaces::msg::SetParametersResult
+PositionGoalChecker::dynamicParametersCallback(std::vector<rclcpp::Parameter> parameters)
+{
+  rcl_interfaces::msg::SetParametersResult result;
+  for (auto & parameter : parameters) {
+    const auto & type = parameter.get_type();
```
SteveMacenski, line 127: Modifty to do:

```
   const auto & param_type = parameter.get_type();
    const auto & param_name = parameter.get_name();
    if (param_name.find(plugin_name_ + ".") != 0) {
      continue;
    }
```

There's some work in https://github.com/ros-navigation/navigation2/pull/5106/files to add these checks to all the plugins so if we're adding this, we should make sure to have this reflected as well

nav2_rotation_shim_controller/CMakeLists.txt
SteveMacenski: Oh, CI doesn't even build with this :laughing: 
```

CMake Error at CMakeLists.txt:31 (target_link_libraries):
  Target "nav2_rotation_shim_controller" links to:

    nav2_controller::nav2_controller

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

```

SteveMacenski (changes requested):

nav2_controller/plugins/position_goal_checker.cpp
PrabhavSaxena: Got it

nav2_rotation_shim_controller/CMakeLists.txt
PrabhavSaxena: Ohhh, I will fix this

SteveMacenski (approved):

mini-1235:
Can we backport this to Jazzy? :)

SteveMacenski:
I don't see why not

mini-1235:
> I don't see why not

Thanks @SteveMacenski !

mini-1235:
> @mini-1235 it would not able to be autobackported #5176. Can you do a backport PR and I can merge that in?

Sure

PR #5163, opened 2025-05-16
suppress rviz logging to warnings and above

SteveMacenski:


nav2_bringup/launch/rviz_launch.py
```diff
@@ -58,8 +58,8 @@ def generate_launch_description() -> LaunchDescription:
         package='rviz2',
         executable='rviz2',
         namespace=namespace,
-        arguments=['-d', rviz_config_file],
-        output='screen',
+        arguments=['-d', rviz_config_file, '--ros-args', '--log-level', 'warn'],
+        output='log',
```
SteveMacenski, line 62:
```suggestion
        output='screen',
```

mini-1235:
Are there any specific reasons we are doing this? It actually breaks the costmap cost tool logging https://github.com/ros-navigation/navigation2/blob/6a8560c09639900b3d47734991fd97cbf552672d/nav2_rviz_plugins/src/costmap_cost_tool.cpp#L128-L142

SteveMacenski:
The logging from rviz was getting super annoying and errors from rviz were distracting from Nav2 logging. Its fine if the visualization tool has some droppage at the start, but folks were starting to diagnose that as a Nav2 error and giving reports. Also it was also just annoying me personally as I was bringing up the system repeatedly looking for other errors.

I agree that this is problematic. Anything you suggest to get around this?

mini-1235:
I reverted it back to INFO, and I can indeed saw a lot of log from rviz related to `tf2_ros::MessageFilter`, I think this is also related to your ticket in 2020 https://github.com/ros2/rviz/issues/513. 

I think changing it to `INFO_THROTTLE` for this section
https://github.com/ros2/geometry2/blob/5ab41706082765bf3f91ddfe861d88ef0e124697/tf2_ros/include/tf2_ros/message_filter.hpp#L770-L774
makes most sense to me. A similar PR is merged for `canTransform` https://github.com/ros2/geometry2/pull/663, what do you think?

SteveMacenski:
Yeah I think that would be great! Tag me in a PR so I can make sure to follow up with maintainers if they don't respond (we have a separate thread)

PR #5165, opened 2025-05-16
Prevent MPPI controller from resetting speed limits upon goal execution.

leander-dsouza:
---

## Basic Info

| ------ | ----------- |

---

* Added a default parameter to the MPPI controller that prevents it from resetting speed limits whenever a goal is reached/terminated.

* N/A

* The turtlebot now begins the movement in a speed limit zone with its speed reduced to the speed limit of the zone.

---

* N/A

#### For Maintainers: 

### DIFF
--- a/nav2_mppi_controller/include/nav2_mppi_controller/optimizer.hpp
+++ b/nav2_mppi_controller/include/nav2_mppi_controller/optimizer.hpp
@@ -120,8 +120,9 @@ public:
 
   /**
    * @brief Reset the optimization problem to initial conditions
+   * @param Whether to reset the constraints to its base values
    */
-  void reset();
+  void reset(bool reset_dynamic_speed_limits = true);
 
   /**
    * @brief Get the motion model time step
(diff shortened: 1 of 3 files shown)


nav2_mppi_controller/include/nav2_mppi_controller/optimizer.hpp
```diff
@@ -120,8 +120,9 @@ class Optimizer
 
   /**
    * @brief Reset the optimization problem to initial conditions
+   * @param Whether to reset the constraints to its base values
    */
-  void reset();
+  void reset(bool reset_constraints = true);
```
SteveMacenski, line 125:
```suggestion
  void reset(bool reset_dynamic_speed_limits = true);
```

Ditto elsewhere

nav2_mppi_controller/src/controller.cpp
```diff
@@ -86,7 +86,7 @@ void MPPIController::deactivate()
 
 void MPPIController::reset()
 {
-  optimizer_.reset();
+  optimizer_.reset(false);
```
SteveMacenski, line 89:
```suggestion
  optimizer_.reset(false /*Don't reset zone-based speed limits between requests*/);
```

SteveMacenski (changes requested):

SteveMacenski:
Good find here on MPPI

SteveMacenski (approved):
FYI @pepisg

PR #5167, opened 2025-05-16
Add stateful to regulated pure pursuit controller

jadhm:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (N/A) |
| Robotic platform tested on | (gazebo simulation and my own robot) |

---

*  Added  stateful  parameter, to the RPP controller

* Updated Doc [Here](https://github.com/ros-navigation/docs.nav2.org/pull/690)

* Tested this feature on a real robot

---

#### For Maintainers: 

### DIFF
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -102,7 +102,7 @@ ParameterHandler::ParameterHandler(
     node, plugin_name_ + ".use_collision_detection",
     rclcpp::ParameterValue(true));
   declare_parameter_if_not_declared(
-      node, plugin_name_ + ".stateful", rclcpp::ParameterValue(false));
+      node, plugin_name_ + ".stateful", rclcpp::ParameterValue(true));
 
   node->get_parameter(plugin_name_ + ".desired_linear_vel", params_.desired_linear_vel);
   params_.base_desired_linear_vel = params_.desired_linear_vel;
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -101,6 +101,8 @@ ParameterHandler::ParameterHandler(
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".use_collision_detection",
     rclcpp::ParameterValue(true));
+  declare_parameter_if_not_declared(
+      node, plugin_name_ + ".stateful", rclcpp::ParameterValue(false));
 
   node->get_parameter(plugin_name_ + ".desired_linear_vel", params_.desired_linear_vel);
   params_.base_desired_linear_vel = params_.desired_linear_vel;
@@ -181,6 +183,7 @@ ParameterHandler::ParameterHandler(
   node->get_parameter(
     plugin_name_ + ".use_collision_detection",
     params_.use_collision_detection);
+  node->get_parameter(plugin_name_ + ".stateful", params_.stateful_);
 
   if (params_.inflation_cost_scaling_factor <= 0.0) {
     RCLCPP_WARN(
@@ -308,6 +311,8 @@ ParameterHandler::updateParametersCallback(
         params_.use_cost_regulated_linear_velocity_scaling = parameter.as_bool();
       } else if (name == plugin_name_ + ".use_collision_detection") {
         params_.use_collision_detection = parameter.as_bool();
+      } else if (name == plugin_name_ + ".stateful") {
+        params_.stateful = parameter.as_bool();
       } else if (name == plugin_name_ + ".use_rotate_to_heading") {
         params_.use_rotate_to_heading = parameter.as_bool();
       } else if (name == plugin_name_ + ".use_cancel_deceleration") {
--- a/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
+++ b/nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
@@ -183,7 +183,7 @@ ParameterHandler::ParameterHandler(
   node->get_parameter(
     plugin_name_ + ".use_collision_detection",
     params_.use_collision_detection);
-  node->get_parameter(plugin_name_ + ".stateful", params_.stateful_);
+  node->get_parameter(plugin_name_ + ".stateful", params_.stateful);
 
   if (params_.inflation_cost_scaling_factor <= 0.0) {
     RCLCPP_WARN(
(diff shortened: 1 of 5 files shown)


nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -101,6 +101,8 @@ ParameterHandler::ParameterHandler(
   declare_parameter_if_not_declared(
     node, plugin_name_ + ".use_collision_detection",
     rclcpp::ParameterValue(true));
+  declare_parameter_if_not_declared(
+      node, plugin_name_ + ".stateful", rclcpp::ParameterValue(false));
```
SteveMacenski, line 105: Make sure to update docs.nav2.org's configuration guide for RPP for this new parameter.

Also, should we potentially make this default true? Seems reasonable.

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
```diff
@@ -308,6 +311,8 @@ ParameterHandler::updateParametersCallback(
         params_.use_cost_regulated_linear_velocity_scaling = parameter.as_bool();
       } else if (name == plugin_name_ + ".use_collision_detection") {
         params_.use_collision_detection = parameter.as_bool();
+      } else if (name == plugin_name_ + ".stateful") {
+        params_.stateful = parameter.as_bool();
```
SteveMacenski, line 315: make sure to add this to the dynamic parameters unit test please

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -493,6 +502,7 @@ void RegulatedPurePursuitController::reset()
 {
   cancelling_ = false;
   finished_cancelling_ = false;
+  has_reached_xy_tolerance_ = false;
```
SteveMacenski, line 505: I think it also needs to be reset when a new plan is given in `setPlan`

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
```diff
@@ -320,8 +320,17 @@ bool RegulatedPurePursuitController::shouldRotateToGoalHeading(
   const geometry_msgs::msg::PoseStamped & carrot_pose)
 {
   // Whether we should rotate robot to goal heading
-  double dist_to_goal = std::hypot(carrot_pose.pose.position.x, carrot_pose.pose.position.y);
-  return params_->use_rotate_to_heading && dist_to_goal < goal_dist_tol_;
+  double dist_to_goal =
+    std::hypot(carrot_pose.pose.position.x, carrot_pose.pose.position.y);
+
+  if (params_->stateful) {
```
SteveMacenski, line 326: I tihnk this can all be simplified a bit.

At the start of the function 

```
if (!params_->use_rotate_to_heading) {
  return false;
}
```

That removes the need for that in any of the other areas below. We can then simply the rest I think more easily

SteveMacenski (changes requested):
FYI: A complimentary PR introduces a new positional goal checker that is used for Rotation Shim. I'm not saying you have to use it here, but would it be useful here? https://github.com/ros-navigation/navigation2/pull/5162

SteveMacenski:
Please also look at the DCO failure to fix please :-)

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
jadhm: I add your suggestion

nav2_regulated_pure_pursuit_controller/src/regulated_pure_pursuit_controller.cpp
jadhm: Done

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
jadhm: I added it to the unit test

nav2_regulated_pure_pursuit_controller/src/parameter_handler.cpp
jadhm: https://github.com/ros-navigation/docs.nav2.org/pull/690

jadhm:
@SteveMacenski  can you take a look at it again. I resolved your comments

jadhm:
> FYI: A complimentary PR introduces a new positional goal checker that is used for Rotation Shim. I'm not saying you have to use it here, but would it be useful here? #5162

Btw I just feel that this problem will just happen in every "new" controller once the robot reaches the tolerance region maybe as future work it would be maybe handy to add stateful to the interface of the goal checker `nav2_core::GoalChecker` this way you can implement the stateful logic in the controllers using the trigger from the goal checker and get rid of all the individual parameters in the controllers. (just a suggestion for the future)

SteveMacenski (approved):

SteveMacenski:
> Btw I just feel that this problem will just happen in every "new" controller once the robot reaches the tolerance region maybe as future work it would be maybe handy to add stateful to the interface of the goal checker nav2_core::GoalChecker this way you can implement the stateful logic in the controllers using the trigger from the goal checker and get rid of all the individual parameters in the controllers.

I'm not sure I understand, there are stateful features in the goal checkers

jadhm:
> > Btw I just feel that this problem will just happen in every "new" controller once the robot reaches the tolerance region maybe as future work it would be maybe handy to add stateful to the interface of the goal checker nav2_core::GoalChecker this way you can implement the stateful logic in the controllers using the trigger from the goal checker and get rid of all the individual parameters in the controllers.
> 
> I'm not sure I understand, there are stateful features in the goal checkers

What I wanted to say that if the goal checker is configured as stateful, I am expecting that the controller behavior is also stateful accordingly. Now you have to set the goal checker as well as the controller to be stateful separately. What I suggested is to add the stateful as a  variable in the nav2_core::GoalChecker so that if you set it to true in the goal checker this will set the controller to be stateful since that each controller has a reference to the goal checker.

SteveMacenski:
Ah ok, I do agree that would be a nice update! There's only 2 controllers that actually do an internal rotate to heading (RPP you did here and Rotation Shim the other PR I linked to implements) so we should now have the full coverage of statefulness at least now. An improvement could be to have a new API for the goal checkers `getStateful` that defaults to false but implementations can override it to `true` when that setting is active. In that case, RPP / Shim can do that behavior

jadhm:
@SteveMacenski  is it possible to backport this to jazzy?

SteveMacenski:
@jadhm yes, it should be done in the PR linked above

PR #5171, opened 2025-05-16
Fixing docking server when already docked at the requested dock

SteveMacenski:
Resolves https://github.com/ros-navigation/navigation2/issues/5144


josephduchesne:
Thanks for the fix Steve! I'd normally have submitted the tiny PR myself but we're deploying robots to corn fields right now and I've been 100% busy for weeks dealing with all of the new corner cases that always crop up with new deployments in real world environments.

PR #5184, opened 2025-05-20
action timeout in BT client edits error code and string

SteveMacenski:
Addresses https://github.com/ros-navigation/navigation2/issues/4946

For all actions that contain timeout error codes (or are sensible to have them added), this PR enables the `TIMEOUT` error code ID to be returned if the action client in the behavior tree is not able to communicate properly with the action server and returns a timeout.


Ekemini-Udofia (approved):
.

PR #5187, opened 2025-05-22
Option to Reduce Lethal to High-Cost Navigable To Get Out of Keepout Zones if Wandered In

SteveMacenski:
This resolves https://github.com/ros-navigation/navigation2/issues/5139

In it, we check if we're in a lethal keepout zone. If optionally enabled (default off, but on in `nav2_params.yaml`), we reduce the lethal cost to a very high but navigable cost (252) in order to allow the robot to leave on its own but in a very efficient way. 

This cost can be tuned by users to weight the incentives to continue towards the goal (lower cost) with incentives to leave the keepout zone immediately (252 max). 

We update the costmap in the full window size during this mode in order to make sure we set all cells as lethal again once exiting the keepout zone (to keep them in update scope).

### DIFF
--- a/nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
+++ b/nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
@@ -254,7 +254,7 @@ void KeepoutFilter::process(
   unsigned int mg_max_x_u = static_cast<unsigned int>(mg_max_x);
   unsigned int mg_max_y_u = static_cast<unsigned int>(mg_max_y);
 
-  // Lets find the pose's cost if we are allowed to override the lethal cost
+  // Let's find the pose's cost if we are allowed to override the lethal cost
   bool is_pose_lethal = false;
   if (override_lethal_cost_) {
     geometry_msgs::msg::Pose2D mask_pose;
--- a/nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
+++ b/nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
@@ -261,8 +261,8 @@ void KeepoutFilter::process(
     if (transformPose(global_frame_, pose, filter_mask_->header.frame_id, mask_pose)) {
       unsigned int mask_robot_i, mask_robot_j;
       if (worldToMask(filter_mask_, mask_pose.x, mask_pose.y, mask_robot_i, mask_robot_j)) {
-        auto keepout_mask_data = getMaskCost(filter_mask_, mask_robot_i, mask_robot_j);
-        is_pose_lethal = (keepout_mask_data == 253 || keepout_mask_data == 254);
+        auto data = getMaskCost(filter_mask_, mask_robot_i, mask_robot_j);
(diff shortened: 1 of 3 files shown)


nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
```diff
@@ -74,6 +74,11 @@ void KeepoutFilter::initializeFilter(
     std::bind(&KeepoutFilter::filterInfoCallback, this, std::placeholders::_1));
 
   global_frame_ = layered_costmap_->getGlobalFrameID();
+
+  declareParameter("override_lethal_cost", rclcpp::ParameterValue(false));
+  node->get_parameter(name_ + "." + "override_lethal_cost", override_lethal_cost_);
+  declareParameter("lethal_override_cost", rclcpp::ParameterValue(252));
+  node->get_parameter(name_ + "." + "lethal_override_cost", lethal_override_cost_);
```
leander-dsouza, line 81: We can probably add a clamping function so that values larger than `MAX_NON_OBSTACLE` can be handled:

```suggestion
  node->get_parameter(name_ + "." + "lethal_override_cost", lethal_override_cost_);
  
  // clamp lethal_override_cost_ in case if higher than MAX_NON_OBSTACLE is given
  lethal_override_cost_ = \
    std::clamp<unsigned int>(lethal_override_cost_, FREE_SPACE, MAX_NON_OBSTACLE);
```

nav2_costmap_2d/plugins/costmap_filters/keepout_filter.cpp
```diff
@@ -74,6 +74,11 @@ void KeepoutFilter::initializeFilter(
     std::bind(&KeepoutFilter::filterInfoCallback, this, std::placeholders::_1));
 
   global_frame_ = layered_costmap_->getGlobalFrameID();
+
+  declareParameter("override_lethal_cost", rclcpp::ParameterValue(false));
+  node->get_parameter(name_ + "." + "override_lethal_cost", override_lethal_cost_);
+  declareParameter("lethal_override_cost", rclcpp::ParameterValue(252));
```
leander-dsouza, line 80: I think adding a constant would make it clearer.

```suggestion
  declareParameter("lethal_override_cost", rclcpp::ParameterValue(MAX_NON_OBSTACLE));
```

leander-dsouza (commented):
I did some local testing on the branch, and it seems that it works as expected!

### Test 1: Change the pose of the robot to the keepout zone

The idea is to see whether the overall zone is reduced to a traversable zone (dark red - 252 - `MAX_NON_OBSTACLE`)

https://github.com/user-attachments/assets/32f53dcf-2d1d-492c-81fa-fd6836388c23

### Test 2: Test for escape in a keepout zone
Drive manually to a keepout zone, and see if the robot can escape after a goal is given. This works as expected.

https://github.com/user-attachments/assets/90781129-437f-4b6c-a767-ed9e961162c5

### Test 3: Check if applied keepout value is applied as an override.

I have set the overridden value to 10, to check if the color of the map should appear somewhere bluish, and it does as expected.


https://github.com/user-attachments/assets/479f9ef2-ca05-4d34-950f-b6245bf6aa30

PR #5189, opened 2025-05-23
Adding parameter warn_when_defaulting_parameters to control default parameter warnings

MarcoMatteoBassa:
## Basic Info

| ------ | ----------- |
| Primary OS tested on | Ubuntu,) |

---

As previously discussed in https://github.com/ros-navigation/navigation2/pull/5154, as most of the parameters in the stack are associated to default values, it is currently difficult for code maintainers to spot parameters that are not loaded because of typos, wrong namespaces, or because of code changes.

The problem can be solved associating a full definition of the parameters in the yaml configuration files with a mechanism that warns when an association fails.
As this would likely result in a lot of spam for users without a full yaml definition, the current change proposes optionally enabling the feature for a node through a parameter.
This would leave the default policy to not-warning while allowing the user to enable it without modifying source-code.

Unit tests
---

- Applying the utility functions in the code base to simplify it and enable the functionality

#### For Maintainers: 

### DIFF
--- a/nav2_util/include/nav2_util/node_utils.hpp
+++ b/nav2_util/include/nav2_util/node_utils.hpp
@@ -206,16 +206,21 @@ ParamType declare_or_get_parameter(
  * \param[in] node Pointer to a node object
  * \param[in] parameter_name Name of the parameter
  * \param[in] default_value Default value of the parameter
- * \param[in] warn_if_no_override If true, prints a warning whenever the parameter override is missing
  * \param[in] parameter_descriptor Optional parameter descriptor
  * \return The value of the param from the override if existent, otherwise the default value
  */
 template<typename ParamType, typename NodeT>
 ParamType declare_or_get_parameter(
   NodeT node, const std::string & parameter_name,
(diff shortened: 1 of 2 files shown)


nav2_util/include/nav2_util/node_utils.hpp
```diff
@@ -161,63 +161,88 @@ using NodeParamInterfacePtr = rclcpp::node_interfaces::NodeParametersInterface::
...
  */
 template<typename ParamType, typename NodeT>
 ParamType declare_or_get_parameter(
   NodeT node, const std::string & parameter_name,
-  const ParamType & default_value, bool warn_if_no_override = false,
+  const ParamType & default_value,
   const ParameterDescriptor & parameter_descriptor = ParameterDescriptor())
 {
+  bool warn_if_no_override{false};
+  try {
```
SteveMacenski, line 233: I don't see any bugs, but why not have this be set as default value of `false` rather than the try/catch based on declarations based on type for `warn_if_no_override` and `strict_param_loading`?

PR #5191, opened 2025-05-25
Add value rewrites to RewrittenYaml

leander-dsouza:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #5160 |

---

* Added a `value_rewrites` parameter to RewrittenYaml.
* This takes in a dictionary with a string as key, and a launch argument as a parameter.
   For example, if you want to replace a `placeholder_value` in your YAML, with the contents of `replacing_arg` launch argument in runtime, you can use the following:

   ```python
   value_rewrites={'placeholder_value': LaunchConfiguration('replacing_arg')}
   ```
* Added a test module verifying this functionality for bool, int, float, list and dictionary launch argument values.
* Prevent logging of costmap filters by using suitable placeholders through the yaml file.
   
* N/A

* Verified behaviour with a module attached for testing this functionality.
* Verified if the costmap filter zones were suitably toggled by changing the launch arguments.
---

* N/A

#### For Maintainers: 

### DIFF
--- a/nav2_bringup/launch/bringup_launch.py
+++ b/nav2_bringup/launch/bringup_launch.py
@@ -51,15 +51,17 @@ def generate_launch_description() -> LaunchDescription:
     # Map fully qualified names to relative ones so the node's namespace can be prepended.
     remappings = [('/tf', 'tf'), ('/tf_static', 'tf_static')]
 
+    yaml_substitutions = {
+        'KEEPOUT_ZONE_ENABLED': use_keepout_zones,
+        'SPEED_ZONE_ENABLED': use_speed_zones,
+    }
+
     configured_params = ParameterFile(
         RewrittenYaml(
             source_file=params_file,
             root_key=namespace,
             param_rewrites={},
-            value_rewrites={
-                'KEEPOUT_ZONE_ENABLED': use_keepout_zones,
-                'SPEED_ZONE_ENABLED': use_speed_zones,
-            },
+            value_rewrites=yaml_substitutions,
             convert_types=True,
         ),
         allow_substs=True,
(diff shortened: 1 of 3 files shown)


nav2_bringup/launch/bringup_launch.py
```diff
@@ -56,6 +56,10 @@ def generate_launch_description() -> LaunchDescription:
             source_file=params_file,
             root_key=namespace,
             param_rewrites={},
+            value_rewrites={
```
SteveMacenski, line 59: Create a map above like `remappings` rather than defined inline the `RewrittenYaml` call

SteveMacenski (approved):

nav2_bringup/launch/bringup_launch.py
SteveMacenski: Ditto in the other 2 files. Otherwise, ready to merge!

leander-dsouza: I have replaced it with `costmap_filter_remappings` in the bringup, keepout and speed launch files.

nav2_bringup/launch/bringup_launch.py
```diff
@@ -51,11 +51,17 @@ def generate_launch_description() -> LaunchDescription:
     # Map fully qualified names to relative ones so the node's namespace can be prepended.
     remappings = [('/tf', 'tf'), ('/tf_static', 'tf_static')]
 
+    costmap_filter_remappings = {
```
SteveMacenski, line 54: I'd call this `yaml_substitutions`

leander-dsouza: Done :)

PR #5194, opened 2025-05-27
fixing static layer update bounds with different fram

2linkthefire:
---

## Basic Info

| ------ | ----------- |

---

Fix incomplete update bounds issue for static layers if global frame has rotation with map frame.

Fixed and tested in ROS1 navigation package on my physical robot platform in production for 1 week, migrated to ROS2.

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/plugins/static_layer.cpp
+++ b/nav2_costmap_2d/plugins/static_layer.cpp
@@ -363,15 +363,53 @@ StaticLayer::updateBounds(
 
   useExtraBounds(min_x, min_y, max_x, max_y);
 
-  double wx, wy;
+  // Might even be in a different frame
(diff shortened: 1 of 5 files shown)


2linkthefire:
> Does this try to fix the same thing as #4911?

Not exactly. This is a basic map range error that existed in ros1. The footprint clearing function mentioned in this question seems to be a new addition to ros2.

SteveMacenski (commented):
I think this is correct, but how did you test / validate this? 

There's also alot of untested code if you look at the codecov annotations - please fill these gaps with an integration test

Generally this looks good to me and I have no notes

SteveMacenski:
@2linkthefire any update? :-)

SteveMacenski:
Closing due to lack of response

PR #5196, opened 2025-05-27
Adding options to override the footprint subscription qos

MarcoMatteoBassa:
---

## Basic Info

| ------ | ----------- |

---

It is common for robots to publish their footprint on a latched topic.
This change allows the final user to configure the desired qos for the footprint subscription in the costmap objects.
Ex to set the topic as latched in the local costmap:

```
local_costmap:
  local_costmap:
    ros__parameters:
      qos_overrides:
        /footprint_topic_name:
          subscription:
            reliability: "reliable"
            depth: 1
            durability: "transient_local"
```

New parameters available

Validated in simulation that the qos changes using the cli.

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/costmap_2d_ros.cpp
+++ b/nav2_costmap_2d/src/costmap_2d_ros.cpp
@@ -44,6 +44,8 @@
 #include <vector>
 #include <utility>
 
+#include "rclcpp/qos.hpp"
+#include "rclcpp/subscription_options.hpp"
 #include "nav2_costmap_2d/layered_costmap.hpp"
 #include "nav2_util/execution_timer.hpp"
 #include "nav2_util/node_utils.hpp"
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I'm going to let this sit a little while because I'm working this week on migrating Nav2 to our own QoS profiles and utilities to have better control over the stack (and easier to change in one place). Adding in a Nav2 SubOptions seems totally in the scope of that work and I've noted it down to plan to incorporate. I'm leaving it open until that's ready but expect something in the next few days that I'll tag you in to review :-)

SteveMacenski:
This is turning out to be quite the rabbit hole as this sparked a few new refactoring action items attached to all these changes. I've made good progress so far and added a new `nav2_ros_common` package that contains ROS-based wrappers (Service Client, Service Server, Simple Action, Node Thread, Node Utils, Lifecycle Node) and new interface factories that wrap all the `create_*` to output Nav2 wrappers instead of the ROS 2 base tools to make it easy to use / ergonomic. 

So for example, now a `create_service` will output a `nav2::ServiceServer` instead of an `rclcpp::Service` coming with all the tools and conveniences that they entail.  As part of that, I enforce consistent QoS settings for things like Actions and Services so that they cannot be incompatible. Also, I have new Nav2 Publisher and Subscription options factories that include things like QoS overrides, callback groups, and lost message callbacks as exposed to the user and default can be set via ROS 2 param files using the param `allow_parameter_qos_overrides`. 

tl;dr: I'm working on it in https://github.com/ros-navigation/navigation2/tree/4888 and cleaning up some architectural bits along the way to make QoS easier to deal with and make sure things based on a Nav2 stack are consistent.

SteveMacenski:
@MarcoMatteoBassa please review https://github.com/ros-navigation/navigation2/pull/5288 - this should supersede and generalize to more topics and interfaces

Issue #5192, opened 2025-05-26
nav2_smac_planner,  path is not continuous


wang-fire:
**Required Info:**

    Operating System: Ubuntu22.04

    Computer: X86_64

    ROS2 Version:ros-humble（**humble_main branch**）

    Version or commit hash: humble_main:latest

**Steps to reproduce issue**

Simulation Environment: Turtlebot3 + gazebo  +  nav2_smac_planner (Hybrid a star)

**Question**

I am using humble_main branch, nav2_smac_planner, I believe this route is divided into two segments: one segment is the forward extension, and the second segment is generated by the analytic_expansion. My questions are: 
         1)  The endpoint of the forward extension and the starting point of the analytic_expansion should be coincide, rather than overlap? 
         2) If question1) is right, I want to avoid any backward motion caused by 1）. Can this be achieved through a smoothing plugin?

![Image](https://github.com/user-attachments/assets/754d3a47-6bc3-466c-834b-fc28e781f05b)
![Image](https://github.com/user-attachments/assets/88c0b170-2fa5-4bcd-8103-e374e1f46f37)

```
#### Tuning / Configuration Goal
<!-- Description in a few sentences what the aim is that you're trying to tune / configure -->

local_costmap:
  local_costmap:
    ros__parameters:
      update_frequency: 4.0
      publish_frequency: 2.0
      global_frame: odom
      robot_base_frame: base_link
      rolling_window: true
      width: 3
      height: 3
      resolution: 0.05
      # footprint: "[[-0.27, -0.32], [-0.27, 0.32], [0.27, 0.32], [0.27, -0.32]]"
      footprint: "[[-0.125, -0.125], [-0.125, 0.125], [0.125, 0.125], [0.125, -0.125]]"
      #robot_radius: 0.22
      plugins: ["voxel_layer", "inflation_layer"]
      # filters: ["keepout_filter"]
      # keepout_filter:
      #   plugin: "nav2_costmap_2d::KeepoutFilter"
      #   enabled: True
      #   filter_info_topic: "costmap_filter_info"
      inflation_layer:
        plugin: "nav2_costmap_2d::InflationLayer"
        cost_scaling_factor: 3.0
        inflation_radius: 0.40
      voxel_layer:
        plugin: "nav2_costmap_2d::VoxelLayer"
        enabled: True
        publish_voxel_map: True
        origin_z: 0.0
        z_resolution: 0.05
        z_voxels: 16
        max_obstacle_height: 2.0
[...]

PR #5198, opened 2025-05-27
fixed bug

stevedanomodolor:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
@@ -119,7 +119,7 @@ public:
    * @return The score of the refined path
    */
   float refineAnalyticPath(
-    const NodePtr & current_node,
+    NodePtr & node,
     const NodePtr & goal_node,
     const NodeGetter & getter,
     AnalyticExpansionNodes & analytic_nodes);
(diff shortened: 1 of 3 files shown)


nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
```diff
@@ -119,7 +119,7 @@ class AnalyticExpansion
    * @return The score of the refined path
    */
   float refineAnalyticPath(
-    const NodePtr & current_node,
+    NodePtr & node,
```
SteveMacenski, line 122: Add doxygen that this node head may change as a result of refinement

SteveMacenski (approved):

nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
stevedanomodolor: done in https://github.com/ros-navigation/navigation2/pull/5198/commits/de6b5b24897f06ba27a78817352db3f2d64d7f0a

stevedanomodolor:
> Are you able to show a before-and-after to resolve the issue this introduces (Issue 1 in my recent comment here #5192) or do we need to wait on @wang-fire to test and review

I will do it, but would appreciate if @wentasah  can also confirm this also.

SteveMacenski:
@stevedanomodolor that change pulled in a bunch of other commits that shouldn't be here

stevedanomodolor:
> @stevedanomodolor that change pulled in a bunch of other commits that shouldn't be here

fixed

SteveMacenski (approved):

PR #5202, opened 2025-05-28
Pass node options to costmap node

MarcoMatteoBassa:
## Basic Info

| ------ | ----------- |

---

Currently it is impossible to remap the publishers and subscribers of the costmap objects, because the node options of the controller server are not passed to the child node.
The change allows the child costmap node to inherit the node options of the parent, while still using customized options for the name, namespace and use_sim_time.

- verified that the costmap nodes retain the expected name and namespace
- verified that the costmap topics retain the expected name and namespace
- verified that the costmap services retain the expected name and namespace
- verified that it's possible to remap topic names for the costmap subscribers.

---

The free functions might be moved to utils and used where other child nodes are created if the use-case arises.

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -68,7 +68,7 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
     "local_costmap", std::string{get_namespace()},
-    get_parameter("use_sim_time").as_bool());
+    get_parameter("use_sim_time").as_bool(), options);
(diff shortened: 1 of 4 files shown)


SteveMacenski (approved):

SteveMacenski:
I fixed the linting but I can't do DCO for you

MarcoMatteoBassa:
Thanks! It would look like the namespace is correctly passed to the child nodes if added, yes Ex:
```
/hello_namespace/local_costmap/costmap
/hello_namespace/local_costmap/costmap_raw
/hello_namespace/local_costmap/costmap_raw_updates
/hello_namespace/local_costmap/costmap_updates
/hello_namespace/local_costmap/footprint
...
```
I passed the options to the global costmap from the planner server too for completeness.

PR #5203, opened 2025-05-28
adding the fixes for navigate through poses terminating pre-maturely

padhupradheep:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | #4304  |
| Does this PR contain AI generated software? | (no) |

---

Thanks to @anath93 for the contribution. 

This still needs to be tested. 

---

#### For Maintainers: 

### DIFF
--- a/nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
+++ b/nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
@@ -35,6 +35,7 @@ NavigateThroughPosesNavigator::configure(
   }
 
   goals_blackboard_id_ = node->get_parameter("goals_blackboard_id").as_string();
+  orient_angle_rad_ = node->get_parameter("orient_angle_rad", orient_angle_rad_);
 
   if (!node->has_parameter("path_blackboard_id")) {
     node->declare_parameter("path_blackboard_id", std::string("path"));
@@ -51,6 +52,12 @@ NavigateThroughPosesNavigator::configure(
 
   // Odometry smoother object for getting current speed
   odom_smoother_ = odom_smoother;
+
+  if (!node->has_parameter("orient_angle_rad")) {
+    node->declare_parameter("orient_angle_rad", rclcpp::ParameterValue(1.57));
+  }
+
+  node->get_parameter("orient_angle_rad", orient_angle_rad_);
 
   if (!node->has_parameter(getName() + ".enable_groot_monitoring")) {
     node->declare_parameter(getName() + ".enable_groot_monitoring", false);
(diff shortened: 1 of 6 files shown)


nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -137,6 +144,15 @@ NavigateThroughPosesNavigator::goalCompleted(
   result->waypoint_statuses = std::move(waypoint_statuses);
 }
 
+inline double createYawFromQuat(const geometry_msgs::msg::Quaternion & orientation)
```
SteveMacenski, line 147: tf2 Utils has a function that does this `getYaw`

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -169,19 +185,24 @@ NavigateThroughPosesNavigator::onLoop()
...
         double curr_min_dist = std::numeric_limits<double>::max();
         for (size_t curr_idx = 0; curr_idx < current_path.poses.size(); ++curr_idx) {
           double curr_dist = nav2_util::geometry_utils::euclidean_distance(
             current_pose, current_path.poses[curr_idx]);
-          if (curr_dist < curr_min_dist) {
+          double orient_diff = createYawFromQuat(current_pose.pose.orientation) -
+            createYawFromQuat(current_path.poses[curr_idx].pose.orientation);
+          if (curr_dist < curr_min_dist &&
+            abs(orient_diff) < orient_angle_rad)
+          {
```
SteveMacenski, line 205:
```suggestion
          if (curr_dist < curr_min_dist && abs(orient_diff) < orient_angle_rad) {
```

nav2_bt_navigator/src/navigators/navigate_through_poses.cpp
```diff
@@ -169,19 +185,24 @@ NavigateThroughPosesNavigator::onLoop()
...
   res = blackboard->get(path_blackboard_id_, current_path);
   if (res && current_path.poses.size() > 0u) {
     // Find the closest pose to current pose on global path
     auto find_closest_pose_idx =
-      [&current_pose, &current_path]() {
+      [&current_pose, &current_path, &orient_angle_rad]() {
         size_t closest_pose_idx = 0;
         double curr_min_dist = std::numeric_limits<double>::max();
         for (size_t curr_idx = 0; curr_idx < current_path.poses.size(); ++curr_idx) {
           double curr_dist = nav2_util::geometry_utils::euclidean_distance(
```
SteveMacenski, line 199: Instead we could check for the distance not just on XY but XY,Yaw so it would be 3D.

SteveMacenski (changes requested):
I actually don't think this really works. What happens if the final goal is literally the same pose as an intermediary pose? There would be not difference in the angle. 

I think something like what was done in https://github.com/ros-navigation/navigation2/pull/4789 is the better approach by not just looking at the pose, but also the path related to that pose.

PR #5217, opened 2025-06-01
Fixing type conversion issues in Smac Planner changes

SteveMacenski:
May resolve Issue 3 in #5192 @slazarev8 please test!

CC @stevedanomodolor

### DIFF
--- a/nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
+++ b/nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
@@ -182,7 +182,6 @@ protected:
   NodeSet _goals_set;
   GoalStateVector _goals_state;
   CoordinateVector _goals_coordinate;
-  NodePtr primary_goal;
 };
 
 }  // namespace nav2_smac_planner
--- a/nav2_smac_planner/src/a_star.cpp
+++ b/nav2_smac_planner/src/a_star.cpp
@@ -230,9 +230,12 @@ void AStarAlgorithm<NodeT>::setGoal(
   switch (goal_heading_mode) {
     case GoalHeadingMode::DEFAULT: {
         // add a single goal node with single heading
-        auto goal = addToGraph(NodeT::getIndex(mx, my, dim_3));
-        goal->setPose(typename NodeT::Coordinates(
-          static_cast<int>(mx), static_cast<int>(my), static_cast<int>(dim_3)));
+        auto goal = addToGraph(
+          NodeT::getIndex(
+            static_cast<unsigned int>(mx),
+            static_cast<unsigned int>(my),
+            dim_3));
+        goal->setPose(typename NodeT::Coordinates(mx, my, static_cast<float>(dim_3)));
         goals_state.push_back({goal, true});
         break;
       }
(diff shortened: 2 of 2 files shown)


SteveMacenski:
See https://github.com/ros-navigation/navigation2/pull/5221 and https://github.com/ros-navigation/navigation2/pull/5222 -- closing and to be included in a regression fix PR

slazarev8:
> May resolve Issue 3 in #5192 @slazarev8 please test!
> 
> CC @stevedanomodolor

Sorry for the late reply, but it did not resolve issue 3. However, I see that you have already reverted the entire PR.

I've been trying to debug the issue for some time, and here is what I found:

1. At the finish point, obstacle_heuristic is 0.00001 (which looks like an uninitialized value) — this is strange. Hence, NodeLattice::getDistanceHeuristic returns 0.0.
2. As a result, A* finishes at this point:
 ```
if (approach_iterations >= getOnApproachMaxIterations()) {
        return _graph.at(_best_heuristic_node.second).backtracePath(path);
      }
```

But I still have no idea how this is connected to the PR, and why I can't reproduce the issue after reverting it.

**Comment by stevedanomodolor** [comment]
@slazarev8 Really appreciate you testing this. We were able to find the issue. It look like I was managing incorrectly pointer. It happened in an edge case when cache_obstacle_heuristic was set to true. 

Can you test this branch https://github.com/stevedanomodolor/navigation2/tree/fix/smac_planner_orientation_goals. This should include the fix for the three problem. I am still testing it to make sure I am not missing something.

The problem was bas pointer handling. 

Again thanks you for bug you found.

**Comment by slazarev8** [comment]
> Can you test this branch https://github.com/stevedanomodolor/navigation2/tree/fix/smac_planner_orientation_goals. This should include the fix for the three problem. I am still testing it to make sure I am not missing something.

Yes, this one fixes my issue 

>  It look like I was managing incorrectly pointer

@stevedanomodolor Could you, please, provide more details which pointer was the issue?

**Comment by stevedanomodolor** [comment]
> > Can you test this branch https://github.com/stevedanomodolor/navigation2/tree/fix/smac_planner_orientation_goals. This should include the fix for the three problem. I am still testing it to make sure I am not missing something.
> 
> Yes, this one fixes my issue 
> 
> >  It look like I was managing incorrectly pointer
> 
> @stevedanomodolor Could you, please, provide more details which pointer was the issue? 
> 



The long explanation is 

The issue occurs when cache_obstacle_heuristic is set to true. By default, it's false, which is why I wasn't able to reproduce the problem initially.
When caching is enabled, the code enters the "goal has changed" logic. The idea of the code which i missed here is that addtograph reuses the same pointers in between goals to be computationally efficient.
The logic step i use when i assign the previous goal to the current goal is wrong because since the goal in the goal manager and the goal generated from the addtograph function uses the same pointer, when i create a new goal with a new value, i change the goal values stored in goal manager.
As a result, I'm effectively assigning the previous goal to be the current goal again. This means that on the second attempt, we never detect a goal change, and therefore never reset the cache. In addition the start goal and the end😬.

```
 auto goalHasChanged = [&]() -> bool {
   if (previous_goals.empty()) {
     std::cout << "No previous goals, assuming goal has changed." << std::endl;
     return true;
   }
 
   const auto & prev = previous_goals.front().goal;
   const auto & curr = goals_state.front().goal;
   return (prev->pose.x != curr->pose.x) || (prev->pose.y != curr->pose.y);
 };
 
 // Goal has changed — reset the obstacle heuristic if needed
 std::cout << "Checking if goal has changed..." << std::endl;
 std::cout << goalHasChanged() << std::endl;
 
 if (!_search_info.cache_obstacle_heuristic || goalHasChanged()) {
   std::cout << "Resetting obstacle heuristic..." << std::endl;
   if (!_start) {
     throw std::runtime_error("Start must be set before goal.");
   }
 }
```

Which makes sense why ithe first iteration works and then it stops working.

Issue #5015, opened 2025-03-24
[Docking] Create API to enable / disable detector


SteveMacenski:
This would be useful so that the detector running for docking (which may be very computationally expensive, like AI) doesn't need to be always running when not in a docking situation. You can enable it when starting to dock and disable it once done - then the detector can be a task server instead of an always-on node (optionally)

ayushkumar8340:
Hi,
I am interested in working on this problem statement. Could you please assign this issue to me and provide further details?

SteveMacenski:
@ayushkumar8340 thanks for showing interest!

I think what we want to do here is to:
* Update the Dock Plugin API to include a new "startDetection()" "stopDetection()" methods 
* Update documentation in the readme and nav2 website about them
* Implement these 2 functions in the 2 plugins: Set a service name parameter to enable and disable the task server doing dock detections, with the dock type name as a request field (in case multiple dock types, to enable detections on the the current right one). If none provided (make the default empty string), then we skip and don't do anything, meaning the detector is always on

That would allow the explicit start/end of detections when they're needed for docking in case a high computation cost detector is being used (like AI) 

Does that make sense? 😄

ayushkumar8340:
Yup got it. I'll keep you posted about the updates.

ayushkumar8340:
I had a few doubts regarding the implementation of this problem statement :

1. As you mentioned, there should be  a mechanism to turn on/off the task server doing docking detection but when i went through the code i found out that both the plugins: simple charging dock and simple non-charging dock subscribes to a topic to get the dock pose and there no such mechanism to trigger the detection node as it is a separate package. 

![Image](https://github.com/user-attachments/assets/9c421481-c7cc-4b1c-b93c-ffed3fffc2e8)
Source : https://github.com/open-navigation/opennav_docking/tree/ubr1_testing

So do you want me to add some logic inside the detection node but according to me that would not be a good and scalable solution.
Or do I need to use the std::system() function to start a process (the detector node) and then kill it using the same.

I don't think both the solutions are good enough. What do you recommend ?

2. Does the service which i have to create will have 1 field (Just confirmation): 
a. type -> dock plugin type to enable

SteveMacenski:
1. This is true for the current detectors, but not all detectors are that way. Many are based on AI solutions which are heavy to run continuously and should only be enabled when required. 

[...]

PR #5218, opened 2025-06-02
feat(opennav_docking): add on-demand detector client & unit tests for dock plugins (#5015)

bkoensgen:
* Add detector client initialization in **SimpleNonChargingDock**
* Mirror the same logic in **SimpleChargingDock**
* Expand unit-test coverage for both dock plugins

---

## Basic Info

| Info | Value |
|------|-------|
| Ticket(s) this addresses | #5015 |
| Primary OS tested on | Official Nav2 **Rolling** Docker image |
| Robotic platform tested on | Unit tests only (`colcon test`) |
| Does this PR contain AI-generated software? | No |

---

* Introduce **on-demand external-detector control** for both charging and non-charging dock plugins  
  * New parameters: `detector_service_name`, `detector_service_timeout`, `subscribe_toggle`
  * Detector started/stopped via `std_srvs/Trigger` only when first needed
* **SimpleNonChargingDock** now creates the detector client (symmetry with charging dock)
* **Unit tests** added to validate service call logic and TF handling

---

## Description of documentation updates required

* Added the three new parameters to `nav2_docking/README.md`  

---

* `colcon test --packages-select opennav_docking` → **all tests pass**
* Manual sanity check: plugins load without warnings inside the Rolling dev-container

---

#### For Maintainers 
- [ ] Check that any new parameters added are updated in **docs.nav2.org**
- [ ] Check that any new plugins are added to the plugins page
- [ ] If BT Node, additionally: add to BT's XML index of nodes for Groot, BT package's readme table, and BT library lists

### DIFF
--- a/nav2_docking/opennav_docking/src/simple_charging_dock.cpp
+++ b/nav2_docking/opennav_docking/src/simple_charging_dock.cpp
@@ -465,6 +465,8 @@ void SimpleChargingDock::cleanup()
 {
   detector_client_.reset();
   dock_pose_sub_.reset();
+  detection_active_ = false;
+  initial_pose_received_ = false;
   RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock cleaned up");
 }
 
--- a/nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
+++ b/nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
@@ -431,6 +431,8 @@ void SimpleNonChargingDock::cleanup()
 {
   detector_client_.reset();
   dock_pose_sub_.reset();
+  detection_active_ = false;
+  initial_pose_received_ = false;
   RCLCPP_DEBUG(node_->get_logger(), "SimpleNonChargingDock cleaned up");
 }
 


nav2_docking/opennav_docking/CMakeLists.txt
```diff
@@ -152,6 +153,9 @@ target_link_libraries(simple_non_charging_dock PRIVATE
   tf2::tf2
 )
 
+ament_target_dependencies(simple_charging_dock PUBLIC std_srvs)
```
SteveMacenski: This is deprecitaed, add to the `target_link_libraries` of the plugins

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -51,19 +52,23 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
     const std::string & name, std::shared_ptr<tf2_ros::Buffer> tf);
 
   /**
-   * @brief Method to cleanup resources used on shutdown.
+   * @brief Final cleanup, in case deactivate() was skipped.
```
SteveMacenski: This comment is incorrect. Cleanup is its own lifecycle state and is not due to skipping deactivate.

SteveMacenski: In general, remove all comment changes to these headers, they add unnecessary and sometimes incorrect information.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,7 +342,85 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+{
+  stopDetection();
+  detector_client_.reset();
+  detected_pose_sub_.reset();
+  RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock cleaned up");
+}
+
 }  // namespace opennav_docking
 
+
```
SteveMacenski: Remove

nav2_docking/opennav_docking/include/opennav_docking/simple_non_charging_dock.hpp
```diff
@@ -19,6 +19,7 @@
 #include <memory>
 #include <vector>
 
+#include <std_srvs/srv/trigger.hpp>
```
SteveMacenski: Use quotes not `<>` (ditto other file)

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -123,6 +139,11 @@ void SimpleChargingDock::configure(
   node_->get_parameter(name + ".filter_coef", filter_coef);
   filter_ = std::make_unique<PoseFilter>(filter_coef, external_detection_timeout_);
 
+  if (!detector_service_name_.empty()) {
+    detector_client_ = node_->create_client<std_srvs::srv::Trigger>(
```
SteveMacenski, line 158: Use the Nav2 utils service client, we don't use raw service clients in Nav2 for various reasons

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -196,6 +212,10 @@ bool SimpleChargingDock::getRefinedPose(geometry_msgs::msg::PoseStamped & pose,
     return true;
   }
 
+  if (detector_state_ == DetectorState::OFF) {
+    startDetection();
```
SteveMacenski: Question: should this be explicitly enabled / disabled by the server application rather than implicitly in the dock plugins? The hot-start here could work but you never actually disable the detector after docking is complete again. If we did this as part of the Docking Server, then its something that can be enabled when we hit the staging pose & disabled once completed.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,7 +342,85 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+
+void SimpleChargingDock::deactivate()
+{
+  stopDetection();
+  RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock deactivated");
+}
+
+void SimpleChargingDock::cleanup()
+{
+  stopDetection();
```
SteveMacenski: Remove, this should be already stopped by `deactivate`.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,7 +342,85 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+}
+
+void SimpleChargingDock::stopDetection()
+{
+  // Skip if already OFF
+  if (detector_state_ == DetectorState::OFF) {return;}
+
+  // 1. Service STOP request
+  //    Fire-and-forget; detector should teardown itself
+  if (detector_client_ && detector_client_->service_is_ready()) {
```
SteveMacenski: With the Nav2 Utils Service Client, you can just do a `detector_client_ && detector_client_->invoke()`. Make sure to use the API with the timeout, we cannot have indefinite spins. There has to be error handling somewhere on failures.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,7 +342,85 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+    auto req = std::make_shared<std_srvs::srv::Trigger::Request>();
+    detector_client_->async_send_request(req);
```
SteveMacenski: For example: this is never spun to actually know if this was completed.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -133,11 +154,6 @@ void SimpleChargingDock::configure(
 
   if (use_external_detection_pose_) {
     dock_pose_.header.stamp = rclcpp::Time(0);
-    dock_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
```
SteveMacenski, line 136: If `!subscribe_toggle_`, I think we should create this, no?

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,7 +342,85 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+    if (!detector_client_->wait_for_service(
+          std::chrono::duration<double>(detector_service_timeout_)))
+    {
+      RCLCPP_WARN(node_->get_logger(),
+                  "Detector service '%s' unavailable (timeout %.1f s), "
+                  "continuing with subscription only",
+                  detector_service_name_.c_str(), detector_service_timeout_);
+    } else {
+      auto req = std::make_shared<std_srvs::srv::Trigger::Request>();
+      auto future = detector_client_->async_send_request(req);
```
SteveMacenski: `invoke` which does the `spin` this is missing

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,7 +342,85 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+  if (subscribe_toggle_ && !detected_pose_sub_) {
+    detected_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
+      "detected_dock_pose", rclcpp::SensorDataQoS(),
+      [this](const geometry_msgs::msg::PoseStamped::SharedPtr pose) {
+        detected_dock_pose_ = *pose;
+        detector_state_ = DetectorState::ON;
+      });
+  }
+
+  detector_state_ = DetectorState::WAITING_START;
```
SteveMacenski: This state is never queried, so is it useful?

SteveMacenski (changes requested):
Good first stab! I have some questions and suggestions, but largely its a good starting point for a couple of hopefully quick iterations. The changes requested in this plugin are the same in both, but I only reviewed the one just to avoid duplicating every comment. Take each comment for both plugin files

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: https://github.com/ros-navigation/navigation2/blob/main/nav2_util/include/nav2_util/service_client.hpp

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
bkoensgen: hi @SteveMacenski,

Thanks for the question,

You're right that having the DockingServer explicitly control when detection starts/stops would be more optimal than the current approach where plugins manage it internally via getRefinedPose().

The current PR implements the detector control within plugins (started by getRefinedPose(), stopped by deactivate()). To implement your suggested architecture, we'd need to:

1. Add public API methods (e.g., `startDetectionProcess()`/`stopDetectionProcess()`) to the dock plugin interface
2. Modify DockingServer to call these at appropriate stages (after staging pose, after docking complete)
3. Update getRefinedPose() to assume detection is already active

?: Should I expand this PR to include these DockingServer-level changes, or would you prefer to merge the current plugin-internal implementation (after fixing the ServiceClient issues) and tackle server-side control in a follow-up PR?

Happy to go either direction ;just want to make sure we align on scope

SteveMacenski: > started by getRefinedPose(), stopped by deactivate()

Just an FYI, that wouldn't meet the intent of the ticket. getRefinedPose occurs while docking but deactivate only happens when the server is deactivated during shutdown. That would mean after the first docking attempt the robot would continue to run the detector - which is kind of the point to shut down between dock requests. 

With respect to your action items - I agree! I think it makes sense here as a single PR :smile:

SteveMacenski:
@bkoensgen re-request a review from me in the github UX when you're ready for me to take a look again (and CI is building :wink: )

SteveMacenski:
@bkoensgen you may want to compile locally and test rather than using CI as your builder. you'd probably have faster iteration cycles.

bkoensgen:
@SteveMacenski

Thanks for the detailed review and guidance! I've pushed up the changes addressing your feedback.

The biggest change is that I've moved the detector lifecycle management (`start/stopDetectionProcess`) into the plugin API itself. The `DockingServer` now just calls these methods, which gives individual plugins full control over how and when they manage their perception resources (like starting/stopping a GPU-intensive node). This felt a lot cleaner and more robust.

Here’s a quick rundown of the other changes:

*   **Service Client & Dependencies:** Switched over to `nav2_util::ServiceClient` as you suggested and cleaned up the CMakeLists to properly link `std_srvs`.
*   **Test Coverage:** Beefed up the unit tests for both `SimpleChargingDock` and `SimpleNonChargingDock`. They now cover the new `subscribe_toggle` logic and the service failure edge cases we discussed.
*   **Test Reliability:** That also gave me a chance to fix a race condition in the Python smoke tests related to TF propagation. They should be much more stable in CI now.
*   **Git History:** Squashed everything down to a single commit for a clean merge.

Should be in good shape for another look now. Let me know what you think!

nav2_docking/opennav_docking_core/include/opennav_docking_core/non_charging_dock.hpp
```diff
@@ -125,6 +125,16 @@ class NonChargingDock : public ChargingDock
     throw std::runtime_error("This dock is not a charging dock!");
   }
 
+  /**
+   * @brief Start any detection pipelines required for pose refinement.
+   */
+  virtual void startDetectionProcess() = 0;
```
SteveMacenski, line 131: I think both of these can be initialized as empty methods, ie `{}` so that they are not required to be implemented if there is no enable/disable logic. I'd add into the doxygen that this can be implemented if the detector is able to be toggled on and off. Otherwise, do not implement.

nav2_docking/opennav_docking_core/include/opennav_docking_core/charging_dock.hpp
```diff
@@ -124,6 +124,16 @@ class ChargingDock
    */
   virtual bool hasStoppedCharging() = 0;
 
+  /**
+   * @brief Start any detection pipelines required for pose refinement.
+   */
+  virtual void startDetectionProcess() = 0;
```
SteveMacenski, line 130: Ditto

nav2_docking/opennav_docking/test/test_docking_server.py
```diff
@@ -127,8 +129,27 @@ def setUp(self) -> None:
         self.command = Twist()
         self.node = rclpy.create_node('test_docking_server')
         # Publish odometry
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer, self.node)
         self.odom_pub = self.node.create_publisher(Odometry, 'odom', 10)
 
+    def wait_for_node_to_be_active(self, node_name, timeout_sec=10.0):
+        """Attend qu'un nœud managé passe à l'état actif."""
```
SteveMacenski: Can we adjust this to English please :-)

nav2_docking/opennav_docking/test/test_docking_server.py
```diff
@@ -127,8 +129,27 @@ def setUp(self) -> None:
...
+        start_time = time.time()
+        while time.time() - start_time < timeout_sec:
+            req = GetState.Request()
+            future = client.call_async(req)
+            rclpy.spin_until_future_complete(self.node, future, timeout_sec=1.0)
+            if future.result() and future.result().current_state.id == 3:  # 3 = ACTIVE
+                self.node.get_logger().info(f'Node {node_name} is active.')
+                return
+            time.sleep(0.5)
+        self.fail(f'Node {node_name} did not become active within {timeout_sec} seconds.')
```
SteveMacenski, line 152: This is a log, not actually returning a failure or throwing an exception. I think this should be propagated to fail the test, else we wouldn't know

nav2_docking/README.md
```diff
@@ -36,12 +36,14 @@ The `ChargingDock` and `NonChargingDock` plugins are the heart of the customizab
...
-3. Use the dock's plugin to initially detect the dock and return the docking pose
-4. Enter a vision-control loop where the robot attempts to reach the docking pose while its actively being refined by the vision system
-5. Exit the vision-control loop once contact has been detected or charging has started
-6. Wait until charging starts (if applicable) and return success.
+3. Call the dock's plugin `startDetectionProcess()` method to activate any external detection mechanisms.
+4. Use the dock's plugin to initially detect the dock (`getRefinedPose`) and return the docking pose.
+5. Enter a vision-control loop where the robot attempts to reach the docking pose while it's actively being refined by the vision system.
+6. Exit the vision-control loop once contact has been detected or charging has started (if applicable).
+7. Wait until charging starts (if applicable) and return success.
+8. Call the dock's plugin `stopDetectionProcess()` method to deactivate any external detection mechanisms (this is also called on failure/cancellation).
```
SteveMacenski:
```suggestion
8. Call the dock's plugin `stopDetectionProcess()` method to deactivate any external detection mechanisms.
```

nav2_docking/README.md
```diff
@@ -36,12 +36,14 @@ The `ChargingDock` and `NonChargingDock` plugins are the heart of the customizab
...
 
-If anywhere this procedure is unsuccessful, `N` retries may be made by driving back to the dock's staging pose and trying again. If still unsuccessful, it will return a failure code to indicate what kind of failure occurred to the client.
+If anywhere this procedure is unsuccessful (before step 8), `N` retries may be made. This involves calling `stopDetectionProcess()`, driving back to the dock's staging pose, and then restarting the process from step 3. If still unsuccessful after retries, it will return a failure code to indicate what kind of failure occurred to the client.
```
SteveMacenski:
```suggestion
If anywhere this procedure is unsuccessful (before step 8), `N` retries may be made, driving back to the dock's staging pose, and then restarting the process from step 3. If still unsuccessful after retries, it will return a failure code to indicate what kind of failure occurred to the client.
```

nav2_docking/README.md
```diff
@@ -151,6 +153,10 @@ There are two functions used during dock approach:
    for charging to start by calling `isCharging`.
    * This may be implemented using the `sensor_msgs/BatteryState` message to check the power status or for charging current.
    * Used for charging-typed plugins.
+   + * `startDetectionProcess`: Start any detection pipelines required for pose refinement.
```
SteveMacenski: These bullets are needing 1 less set of indentations, currently they're nested under `isCharging`. Also, then the `two functions` should be 4. 

Personally I Think these can be removed, these are just describing the 'main' functions, there are more than just these and I don't think these count as 'main' functions.

nav2_docking/README.md
```diff
@@ -175,8 +181,9 @@ some robots.
 `getStagingPose` applies a parameterized translational and rotational offset to the dock pose to obtain the staging pose.
 
 `getRefinedPose` can be used in two ways.
-1. A blind approach where the returned dock pose will simply be equal to whatever was passed in from the dock database. This may work with a reduced success rate on a real robot (due to global localization error), but is useful for initial testing and simulation.
+1. A blind approach where the returned dock pose will simply be equal to whatever was passed in from the dock database. This may work with a reduced success rate on a real robot (due to global localization error), but is useful for initial testing and simulation. The `start/stopDetectionProcess` methods would typically do nothing in this case.
```
SteveMacenski, line 180: Remove, TMI

nav2_docking/README.md
```diff
@@ -175,8 +181,9 @@ some robots.
...
 2. The more realistic use case is to use an AR marker, dock pose detection algorithm, etc. The plugin will subscribe to a `geometry_msgs/PoseStamped` topic `detected_dock_pose`. This can be used with the `image_proc/TrackMarkerNode` for Apriltags or other custom detectors for your dock. It is unlikely the detected pose is actually the pose you want to dock with, so several parameters are supplied to represent your docked pose relative to the detected feature's pose.
+The subscription to `detected_dock_pose` can be managed dynamically (see `subscribe_toggle` parameter). Additionally, an external detector process can be triggered via a service call (see `detector_service_name`). The `DockingServer` calls `startDetectionProcess()` before `getRefinedPose` (which is called in a loop) and `stopDetectionProcess()` after the docking action concludes (successfully or not). This allows plugins to manage resources like GPU usage by only running detection when needed.
```
SteveMacenski, line 182:
```suggestion
The subscription to `detected_dock_pose` can be managed dynamically (see `subscribe_toggle` parameter). Additionally, an external detector process can be triggered via a service call (see `detector_service_name`). The `DockingServer` calls `startDetectionProcess()` before `getRefinedPose` and `stopDetectionProcess()` after the docking action concludes (successfully or not). This allows plugins to manage resources like GPU usage by only running detection when needed.
```

nav2_docking/README.md
```diff
@@ -257,6 +264,9 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
...
 | rotate_to_dock        | Enables backward docking without requiring a sensor for detection during the final approach. When enabled, the robot approaches the staging pose facing forward with sensor coverage for dock detection; after detection, it rotates and backs into the dock using only the initially detected pose for dead reckoning. | bool |  false      |
+| detector_service_name      | Trigger service name to start/stop the detector (e.g., a camera node or an AprilTag detector node). Leave empty to disable service calls. | string | ""    |
+| detector_service_timeout   | Timeout (seconds) to wait for the detector service (if `detector_service_name` is set) to become available or respond. | double | 5.0   |
+| subscribe_toggle           | If true, dynamically subscribe/unsubscribe to `detected_dock_pose` topic when `start/stopDetectionProcess` are called. If false, the subscription is kept alive throughout the plugin's lifecycle if `use_external_detection_pose` is true. This can be useful if the detector node is always running and publishing. | bool   | true  |
```
SteveMacenski:
```suggestion
| subscribe_toggle           | If true, dynamically subscribe/unsubscribe to `detected_dock_pose` topic when `start/stopDetectionProcess` are called. If false, the subscription is kept alive throughout the plugin's lifecycle if `use_external_detection_pose` is true. This can be useful if the detector node only computes values when it has a subscription. | bool   | true  |
```

SteveMacenski: Update this to be default false in the docs and the code

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -103,6 +105,16 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
    */
   virtual bool hasStoppedCharging();
 
+  /**
+   * @brief Start external detection process (service call + subscribe).
+   */
+  void startDetectionProcess() override {startDetection();}
```
SteveMacenski: Just implement startDetectionProcess with startDetection's content, there's not a need for a new method.

Ditto with the Stop functions

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -149,6 +161,29 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
...
+  // Client used to call the Trigger service
+  nav2_util::ServiceClient<
+    std_srvs::srv::Trigger,
+    rclcpp_lifecycle::LifecycleNode::SharedPtr>::SharedPtr detector_client_;
+
+  // Dynamic subscription pointer (can be reset to release CPU/GPU)
+  rclcpp::Subscription<geometry_msgs::msg::PoseStamped>::SharedPtr detected_pose_sub_;
+
+  // Simple finite-state machine for detector status
+  enum class DetectorState { OFF, ON };
```
SteveMacenski: If its just 2 states, this can simply be a bool `detector_enabled_`

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -149,6 +161,29 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
...
+  DetectorState detector_state_{DetectorState::OFF};
+
+  // Internally enable detector (service + subscribe)
+  void startDetection();
```
SteveMacenski: Remove these 2 methods

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -164,7 +164,10 @@ DockingServer::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
   tf2_buffer_.reset();
   docking_action_server_.reset();
   undocking_action_server_.reset();
-  dock_db_.reset();
+  if (dock_db_) {
```
SteveMacenski: Cleanup is the opposite of configure. This would never be called if configure isn't called, so we know `dock_db_` must exist at this point

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -260,6 +263,9 @@ void DockingServer::dockRobot()
         get_logger(), "Robot is already docked and/or charging (if applicable), no need to dock");
       result->success = true;
       docking_action_server_->succeeded_current(result);
+      if (dock) {
```
SteveMacenski: We know dock exists, we call it in conditions above

SteveMacenski: This is also called before `doInitialPerception` so this wouldn't be actived yet to need to stop it

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -329,13 +335,15 @@ void DockingServer::dockRobot()
             stashDockData(goal->use_dock_id, dock, true);
             publishZeroVelocity();
             docking_action_server_->succeeded_current(result);
+            dock->plugin->stopDetectionProcess();
```
SteveMacenski, line 334: Move this before `succeeded_current` line. More about consistency across the codebase than a technical issue. We tend to do the very last thing as succeeding / terminating actions so that timers that can be inserted for metrics

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -351,6 +359,7 @@ void DockingServer::dockRobot()
         // Cancelled, preempted, or shutting down
         stashDockData(goal->use_dock_id, dock, false);
         publishZeroVelocity();
+        dock->plugin->stopDetectionProcess();
```
SteveMacenski, line 359: Should we reset here? We're going to go back around to try again, I don't think there's any reason to stop detection, this is not a terminal state

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -395,6 +404,9 @@ void DockingServer::dockRobot()
   }
 
   // Store dock state for later undocking and delete temp dock, if applicable
+  if (dock) {
```
SteveMacenski: we know `dock` exists, we don't need to  check it

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -37,6 +40,14 @@ void SimpleChargingDock::configure(
   nav2_util::declare_parameter_if_not_declared(
     node_, name + ".use_battery_status", rclcpp::ParameterValue(true));
 
+  // Parameters for optional detector control
+  nav2_util::declare_parameter_if_not_declared(
+    node_, name + ".detector_service_name", rclcpp::ParameterValue(""));
+  nav2_util::declare_parameter_if_not_declared(
+    node_, name + ".detector_service_timeout", rclcpp::ParameterValue(5.0));
+  nav2_util::declare_parameter_if_not_declared(
+    node_, name + ".subscribe_toggle", rclcpp::ParameterValue(true));
```
SteveMacenski: default to false to keep current behavior

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -133,11 +155,6 @@ void SimpleChargingDock::configure(
 
   if (use_external_detection_pose_) {
```
SteveMacenski:
```suggestion
  if (use_external_detection_pose_ && !subscribe_toggle_) {
```

No?

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: I think the code should move back up here. You also don't need to check `!detected_pose_sub_`

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -157,6 +174,17 @@ void SimpleChargingDock::configure(
   filtered_dock_pose_pub_ = node_->create_publisher<geometry_msgs::msg::PoseStamped>(
     "filtered_dock_pose", 1);
   staging_pose_pub_ = node_->create_publisher<geometry_msgs::msg::PoseStamped>("staging_pose", 1);
+
+  // If not toggling subscription and using external detections, keep
+  // subscription alive by creating it here during configuration
+  if (!subscribe_toggle_ && use_external_detection_pose_ && !detected_pose_sub_) {
+    detected_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
+      "detected_dock_pose", rclcpp::SensorDataQoS(),
```
SteveMacenski: Revert the QoS when you move it back

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+}
+
+void SimpleChargingDock::activate()
+{
+  RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock activated");
+}
+
+void SimpleChargingDock::deactivate()
+{
+  RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock deactivated");
```
SteveMacenski, line 461: If you make deactivated call `stopDetection`, you can remove the explicit `stopDetection` call in the dock database. Its probably good that these reset its own resources internally.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+    detected_pose_sub_.reset();
+  }
+
+  detector_state_ = DetectorState::OFF;
+  RCLCPP_INFO(node_->get_logger(), "Detector STOP requested");
+}
+
+void SimpleChargingDock::activate()
+{
+  RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock activated");
```
SteveMacenski: This is unused. It can be reverted into a non-override / `{}`

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
   is_stalled_ = (velocity < stall_velocity_threshold_) && (effort > stall_effort_threshold_);
 }
 
+void SimpleChargingDock::startDetection()
+{
+  // Skip if already starting or ON
+  if (detector_state_ != DetectorState::OFF) {return;}
```
SteveMacenski:
```suggestion
  if (detector_state_ == DetectorState::ON) {return;}
```

This seems more rational (though I think you're going to change this to a bool anyway)

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+
+  // 1. Service START request
+  if (detector_client_) {
+    auto req = std::make_shared<std_srvs::srv::Trigger::Request>();
+    try {
+      detector_client_->invoke(
```
SteveMacenski: You should capture the return and make sure the success is `true`

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+        req,
+        std::chrono::duration_cast<std::chrono::nanoseconds>(
+          std::chrono::duration<double>(detector_service_timeout_)));
+    } catch (const std::exception & e) {
+      RCLCPP_WARN(node_->get_logger(),
+        "Detector service '%s' failed: %s",
+        detector_service_name_.c_str(), e.what());
```
SteveMacenski, line 384: This should be an error message

Also, I think the signature of the start/stop should be `bool` so that if this happens, we can return false and the application can handle that error in the docking server if it cannot be enabled by throwing an exception `opennav_docking_core::FailedToDetectDock`

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+    }
+  }
+
+  // 2. Subscription toggle
+  //    Only subscribe once; will set state to ON on first message
+  if (subscribe_toggle_ && !detected_pose_sub_) {
+    detected_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
+      "detected_dock_pose", rclcpp::SensorDataQoS(),
```
SteveMacenski: QoS

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+      [this](const geometry_msgs::msg::PoseStamped::SharedPtr pose) {
+        detected_dock_pose_ = *pose;
```
SteveMacenski, line 395: Should uses of `detected_dock_pose_` make sure that `detector_state_` is on? Throw otherwise

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +350,88 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+        detector_state_ = DetectorState::ON;
```
SteveMacenski: Remove. This is always set anyway on L384 below when we exist the function. It has no impact in the callback (ditto for the subscriber above).

nav2_docking/README.md
SteveMacenski: TMI

nav2_docking/README.md
SteveMacenski: TMI

nav2_docking/opennav_docking/test/test_docking_server.py
bkoensgen: Thanks for raising this

Quick confirmation before I call it a night: `self.fail()` raises `AssertionError` unconditionally (see std-lib docs), so the test is already marked **FAILED** when the node stays inactive. I’ll add a small `# raises AssertionError` comment in my next push to make that explicit.

¹ https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail

I still have a couple of changes to finish; I’ll batch-reply to the remaining threads and push the updated diff tomorrow.

Let me know if I missed anything in the meantime

SteveMacenski: Got it - this is right then. We have other wrappers were things like `self.fail` and `self.info` map to logging. I didn't check that this had a unittesting base class.

nav2_docking/README.md
```diff
@@ -257,6 +260,9 @@ Note: `dock_plugins` and either `docks` or `dock_database` are required.
...
```
SteveMacenski: Update this to be default false in the docs and the code

SteveMacenski:
Mark this as ready to review again once you'd like me to take a look!

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -149,6 +161,29 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
 
   rclcpp_lifecycle::LifecycleNode::SharedPtr node_;
   std::shared_ptr<tf2_ros::Buffer> tf2_buffer_;
+
+  // Detector control parameters
+  std::string detector_service_name_;
+  double detector_service_timeout_{5.0};
+  bool subscribe_toggle_{true};
```
SteveMacenski:
```suggestion
  bool subscribe_toggle_{false};
```

nav2_docking/opennav_docking/include/opennav_docking/simple_non_charging_dock.hpp
```diff
@@ -127,6 +139,29 @@ class SimpleNonChargingDock : public opennav_docking_core::NonChargingDock
 
   rclcpp_lifecycle::LifecycleNode::SharedPtr node_;
   std::shared_ptr<tf2_ros::Buffer> tf2_buffer_;
+
+  // Detector control parameters
+  std::string detector_service_name_;
+  double detector_service_timeout_{5.0};
+  bool subscribe_toggle_{true};
```
SteveMacenski:
```suggestion
  bool subscribe_toggle_{false};
```

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen this was not completed :-)

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: This was not completed either

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: This was not completed either

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: This was not completed either

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: This was not completed either

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: This was not completed either

SteveMacenski:
@bkoensgen the majority of my review comments were not completed, please see above

nav2_docking/opennav_docking/src/docking_server.cpp
bkoensgen: Ah, good catch. Yeah, we definitely don't want to stop detection if we're just going to retry.

I double-checked, and this specific `if` block is for when the whole action gets cancelled or preempted. It's a hard exit, so it's our last chance to call `stopDetectionProcess()` to prevent a leak.

The actual retry logic in the `catch` block correctly leaves detection running. So I think we're good here. Let me know if you see it differently!

nav2_docking/opennav_docking/src/docking_server.cpp
bkoensgen: Yeah, you're right, that makes sense. Normally, `on_cleanup` would only run after a successful `on_configure`.

I had a quick question about an edge case, though. I might be missing something, but what happens if `on_configure` throws an exception halfway through, before `dock_db_` gets initialized?
My understanding is that the lifecycle manager would still call on_cleanup to be safe. If so, `dock_db_` would be null, and the if (`dock_db_`) check would prevent an issue.

Does that sound right? Just want to make sure I'm not being overly defensive with the code.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
bkoensgen: good point. I checked getRefinedPose and it already has a guard clause at the top (`if (!detector_enabled_)`) that prevents `detected_dock_pose_` from being used improperly.

Returning false there seems like the right behavior for the control loop, so I think we're all set on this one.

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
bkoensgen: You're right, I did remove this initially based on your feedback! But it turned out to cause a subtle race condition that made the tests fail.

What happened was that `startDetection(`) would set the detector to "enabled" immediately, but `getRefinedPose()` would then be called before the first message from the topic actually arrived. This caused the timestamp check to fail.

The only way I found to make it robustly pass the tests was to put the state update back here, inside the callback. This guarantees the plugin only considers itself 'active' once it has actually received data.

It's a bit counter-intuitive, but it seems necessary to handle that timing issue. Let me know if you have other ideas on how to tackle this!

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
SteveMacenski: @bkoensgen this was still not completed

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
SteveMacenski: @bkoensgen this is related to avoid

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: This is already handled 

```
  if (!dock_db_->initialize(node, tf2_buffer_)) {
    on_cleanup(state);
    return nav2_util::CallbackReturn::FAILURE;
  }
```

Anything that can throw is already protected

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -326,6 +329,7 @@ void DockingServer::dockRobot()
             }
             result->success = true;
             result->num_retries = num_retries_;
+            dock->plugin->stopDetectionProcess();
```
SteveMacenski: Move after publishZeroVelocity() so the order is the same in all terminal conditions

nav2_docking/opennav_docking/src/docking_server.cpp
SteveMacenski: @bkoensgen

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: @bkoensgen

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -149,6 +161,28 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
...
   rclcpp_lifecycle::LifecycleNode::SharedPtr node_;
   std::shared_ptr<tf2_ros::Buffer> tf2_buffer_;
+
+  // Detector control parameters
+  std::string detector_service_name_;
+  double detector_service_timeout_{5.0};
+  bool subscribe_toggle_{false};
+
+  // Client used to call the Trigger service
+  nav2_util::ServiceClient<
```
SteveMacenski: Rebase  / merge in changes. This class is now different (nav2::ServiceClient) and should be gotten from the `nav2::LifecycleNode`'s `create_client`. See https://docs.nav2.org/migration/Kilted.html#new-nav2-ros-common-nav2-lifecycle-node

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -149,6 +161,28 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
...
+    std_srvs::srv::Trigger,
+    rclcpp_lifecycle::LifecycleNode::SharedPtr>::SharedPtr detector_client_;
+
+  // Dynamic subscription pointer (can be reset to release CPU/GPU)
+  rclcpp::Subscription<geometry_msgs::msg::PoseStamped>::SharedPtr detected_pose_sub_;
```
SteveMacenski: Ditto, see https://docs.nav2.org/migration/Kilted.html#new-nav2-ros-common-nav2-lifecycle-node needs to be a nav2::Subscription

nav2_docking/opennav_docking/include/opennav_docking/simple_non_charging_dock.hpp
```diff
@@ -127,6 +139,28 @@ class SimpleNonChargingDock : public opennav_docking_core::NonChargingDock
...
   rclcpp_lifecycle::LifecycleNode::SharedPtr node_;
   std::shared_ptr<tf2_ros::Buffer> tf2_buffer_;
+
+  // Detector control parameters
+  std::string detector_service_name_;
+  double detector_service_timeout_{5.0};
+  bool subscribe_toggle_{false};
+
+  // Client used to call the Trigger service
+  nav2_util::ServiceClient<
```
SteveMacenski: Ditto

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -196,6 +220,11 @@ bool SimpleChargingDock::getRefinedPose(geometry_msgs::msg::PoseStamped & pose,
     return true;
   }
 
+  // Guard against using pose data before the first detection has arrived.
+  if (!detector_enabled_) {
```
SteveMacenski: Should log a warning a message

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +351,107 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+  if (subscribe_toggle_ && detected_pose_sub_) {
+    detected_pose_sub_.reset();
+  }
+
+  detector_enabled_ = false;
+  RCLCPP_INFO(node_->get_logger(), "Detector STOP requested");
+  return true;
+}
+
+void SimpleChargingDock::activate() {}
```
SteveMacenski: Remove override if this is not going to do anything

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +351,107 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+  if (subscribe_toggle_ && !detected_pose_sub_) {
+    detected_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
+      "detected_dock_pose", 1,
+      [this](const geometry_msgs::msg::PoseStamped::SharedPtr pose) {
+        detected_dock_pose_ = *pose;
+        detector_enabled_ = true;
+      });
+  }
+
+  RCLCPP_INFO(node_->get_logger(), "Detector START requested.");
```
SteveMacenski: Make this log a bit more formal :-)

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +351,107 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+  }
+
+  // 2. Unsubscribe to release resources
+  //    reset() will tear down the topic subscription immediately
+  if (subscribe_toggle_ && detected_pose_sub_) {
+    detected_pose_sub_.reset();
+  }
+
+  detector_enabled_ = false;
+  RCLCPP_INFO(node_->get_logger(), "Detector STOP requested");
```
SteveMacenski: Ditto

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -322,6 +351,107 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+  }
+
+  // 2. Subscription toggle
+  //    Only subscribe once; will set state to ON on first message
+  if (subscribe_toggle_ && !detected_pose_sub_) {
+    detected_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
+      "detected_dock_pose", 1,
+      [this](const geometry_msgs::msg::PoseStamped::SharedPtr pose) {
+        detected_dock_pose_ = *pose;
+        detector_enabled_ = true;
```
SteveMacenski: Remove. This is the detector state not the topic state

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -105,6 +116,22 @@ void SimpleChargingDock::configure(
...
+  node_->get_parameter(name + ".subscribe_toggle", subscribe_toggle_);
+
+  // Create persistent subscription if toggling is disabled.
+  if (use_external_detection_pose_ && !subscribe_toggle_) {
+    dock_pose_.header.stamp = rclcpp::Time(0);
+    detected_pose_sub_ = node_->create_subscription<geometry_msgs::msg::PoseStamped>(
+      "detected_dock_pose", 1,
+      [this](const geometry_msgs::msg::PoseStamped::SharedPtr pose) {
+        detected_dock_pose_ = *pose;
+        detector_enabled_ = true;
```
SteveMacenski: Ditto

SteveMacenski (commented):
Overall, much improved! I still see a number of previous comments that were not fixed yet. Make sure to un-collapse the "see X more items" in case that's why they were missed. 

All feedback from the simple charging dock should also be applied to the non charging dock plugin as well since they're 1:1

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: This is still a problem. detector_state_ is the state of the detector running, not whether or not you've received data. 

You need to update this so `detector_state_` is `true` on `startDetection` so that the state is properly managed from that perspective. Then, you can have another variable for `detected_pose_received` that can be set to `true` when we receive a message (set to false in whatever initialization / reset conditions required) that can be used in the `getRefinedPose` function.

However, I would argue that you could **probably** remove the `detector_state_` variable in that case. Right now, that's only protecting irraneous calls to `stopDetection` when stopped already and `startDetection` when already started. Given the logic of the server is otherwise sound, I don't think that is actually necessary.

You could just change the name of the variable to `detected_pose_received_`, have it set to `true` in the subscription callback, and have it initialized to `false` on `configure`, `stopDetection`, etc.

bkoensgen:
Thanks for the detailed feedback, @SteveMacenski! It all makes sense. I'll get this rebased on main and updated. Will let you know when it's ready for another look

bkoensgen:
hi @SteveMacenski, ready for a fresh pass.

- Server control: `DockingServer` calls `startDetectionProcess()` before the `getRefinedPose()` loop, and `stopDetectionProcess()` on every terminal
path; order standardized after `publishZeroVelocity()` and before `succeeded_current()` terminate.
- APIs/migration: moved to `nav2::LifecycleNode`, `nav2::ServiceClient`, and `nav2::Subscription`; using `nav2::qos::StandardTopicQoS`.
- Subscriptions: `subscribe_toggle` defaults to false (code + docs). When toggling is disabled, the persistent subscription is created in configure().
- State/race: replaced the mini state machine with two flags `detection_started_` and `initial_pose_received_`; `getRefinedPose()` guards until the first pose arrives and checks freshness.
- Errors: start/stop return bool; start failure logs at ERROR and is propagated to the server.
- Docs: I’ll open the docs.nav2 PR (config + migration notes for `DockingServer` and new params) once you confirm the API/defaults here. If you prefer it upfront, I can send it now.

Open to confirm: `stopDetectionProcess()` placement (after `publishZeroVelocity(),` before `succeeded_current()`).

nav2_docking/opennav_docking/include/opennav_docking/simple_non_charging_dock.hpp
```diff
@@ -89,14 +90,24 @@ class SimpleNonChargingDock : public opennav_docking_core::NonChargingDock
    */
   virtual bool isDocked();
 
+  /**
+   * @brief Start external detection process (service call + subscribe).
+   */
+  bool startDetectionProcess() override {return startDetection();}
```
SteveMacenski: Remove `startDetection` and `stopDetection`. Implement those functions inside of `startDetectionProcess` and `stopDetectionProcess`

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -104,14 +105,24 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
    */
   virtual bool hasStoppedCharging();
 
+  /**
+   * @brief Start external detection process (service call + subscribe).
+   */
+  bool startDetectionProcess() override {return startDetection();}
```
SteveMacenski: Remove startDetection and stopDetection. Implement those functions inside of startDetectionProcess and stopDetectionProcess

nav2_docking/opennav_docking/include/opennav_docking/simple_charging_dock.hpp
```diff
@@ -150,6 +161,27 @@ class SimpleChargingDock : public opennav_docking_core::ChargingDock
...
+  // Detector control parameters
+  std::string detector_service_name_;
+  double detector_service_timeout_{5.0};
+  bool subscribe_toggle_{false};
+
+  // Client used to call the Trigger service
+  nav2::ServiceClient<std_srvs::srv::Trigger>::SharedPtr detector_client_;
+
+  // Dynamic subscription pointer (can be reset to release CPU/GPU)
+  nav2::Subscription<geometry_msgs::msg::PoseStamped>::SharedPtr detected_pose_sub_;
```
SteveMacenski: Why this subscription? We already have `dock_pose_sub_`

nav2_docking/opennav_docking/include/opennav_docking/simple_non_charging_dock.hpp
```diff
@@ -128,6 +139,27 @@ class SimpleNonChargingDock : public opennav_docking_core::NonChargingDock
...
+  // Detector control parameters
+  std::string detector_service_name_;
+  double detector_service_timeout_{5.0};
+  bool subscribe_toggle_{false};
+
+  // Client used to call the Trigger service
+  nav2::ServiceClient<std_srvs::srv::Trigger>::SharedPtr detector_client_;
+
+  // Dynamic subscription pointer (can be reset to release CPU/GPU)
+  nav2::Subscription<geometry_msgs::msg::PoseStamped>::SharedPtr detected_pose_sub_;
```
SteveMacenski: Why this subscription? We already have dock_pose_sub_

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -105,6 +116,27 @@ void SimpleChargingDock::configure(
   node_->get_parameter(name + ".staging_x_offset", staging_x_offset_);
   node_->get_parameter(name + ".staging_yaw_offset", staging_yaw_offset_);
 
+  // Reader parameters to control external detector service and subscription
```
SteveMacenski:
```suggestion
```

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: ^

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: ^

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: ^

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
SteveMacenski: ^

nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
```diff
@@ -289,6 +323,110 @@ void SimpleNonChargingDock::jointStateCallback(const sensor_msgs::msg::JointStat
...
+      "detected_dock_pose",
+      [this](const geometry_msgs::msg::PoseStamped::SharedPtr pose) {
+        detected_dock_pose_ = *pose;
+        initial_pose_received_ = true;
+      },
+      nav2::qos::StandardTopicQoS());
+  }
+
+  detection_started_ = true;
+  RCLCPP_INFO(node_->get_logger(), "Detector START requested.");
```
SteveMacenski: Ditto

nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
```diff
@@ -289,6 +323,110 @@ void SimpleNonChargingDock::jointStateCallback(const sensor_msgs::msg::JointStat
...
+
+  // 2. Unsubscribe to release resources
+  //    reset() will tear down the topic subscription immediately
+  if (subscribe_toggle_ && detected_pose_sub_) {
+    detected_pose_sub_.reset();
+  }
+
+  detection_started_ = false;
+  initial_pose_received_ = false;
+  RCLCPP_INFO(node_->get_logger(), "Detector STOP requested");
```
SteveMacenski: Ditto

nav2_docking/opennav_docking/src/simple_non_charging_dock.cpp
```diff
@@ -289,6 +323,110 @@ void SimpleNonChargingDock::jointStateCallback(const sensor_msgs::msg::JointStat
...
+  return true;
+}
+
+void SimpleNonChargingDock::activate() {}
+
+void SimpleNonChargingDock::deactivate()
+{
+  stopDetectionProcess();
```
SteveMacenski, line 423: Ditto

SteveMacenski (changes requested):
Closer! There are some still open items from my July 2 review that I added a "^" comment to for still relevant. Make sure to expand the comments if the GitHub UX collapses them due to the size of the thread in the PR.

But by in large, this looks very close to ready!

bkoensgen:
Hi Steve,

Thanks again for the review. I revisted the detector subscription
 and simplified things: both plugins now reuse the existing
`dock_pose_sub_` handle instead of introducing a second pointer. When
subscribe_toggle_ is enabled we create the same subscription on demand in
`startDetectionProcess()` and reset it in `stopDetectionProcess(),` and for the
“always on” case it’s created once during `configure()`.

With that the extra `detected_pose_sub_` member is gone from
the headers, and the cleanup paths just reset `dock_pose_sub_`. Everything
still lines up with the lifecycle hooks we added earlier (activate()/
deactivate() manage the publishers, and deactivate() calls
`stopDetectionProcess()` so each plugin handles its own tear down).

let me know if you’d like any other tweaks!

SteveMacenski:
@bkoensgen Checkout the test failure - its minor linting :-) https://app.circleci.com/pipelines/github/ros-navigation/navigation2/16074/workflows/42e6fc1f-9658-4d8b-b46d-0ba90539e3f2/jobs/46964/tests

nav2_docking/opennav_docking/src/docking_server.cpp
```diff
@@ -167,6 +167,7 @@ DockingServer::on_cleanup(const rclcpp_lifecycle::State & /*state*/)
   tf2_buffer_.reset();
   docking_action_server_.reset();
   undocking_action_server_.reset();
+  dock_db_->deactivate();
```
SteveMacenski: Remove, this should already be deactivated if we're in the cleanup transition

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -105,6 +116,26 @@ void SimpleChargingDock::configure(
   node_->get_parameter(name + ".staging_x_offset", staging_x_offset_);
   node_->get_parameter(name + ".staging_yaw_offset", staging_yaw_offset_);
 
+  node_->get_parameter(name + ".detector_service_name", detector_service_name_);
+  node_->get_parameter(name + ".detector_service_timeout", detector_service_timeout_);
+  node_->get_parameter(name + ".subscribe_toggle", subscribe_toggle_);
+
+  // Initialize detection state
+  detection_started_ = false;
```
SteveMacenski: Rename here and in the noncharging plugin to be `detection_active_` to be more clear

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -323,6 +356,118 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+  dock_pose_pub_->on_deactivate();
+  filtered_dock_pose_pub_->on_deactivate();
+  staging_pose_pub_->on_deactivate();
+  RCLCPP_DEBUG(node_->get_logger(), "SimpleChargingDock deactivated");
+}
+
+void SimpleChargingDock::cleanup()
+{
+  detector_client_.reset();
+  dock_pose_sub_.reset();
```
SteveMacenski, line 467: Set for both plugins:
```

  detection_started_ = false;
  initial_pose_received_ = false;
```

nav2_docking/opennav_docking/src/simple_charging_dock.cpp
```diff
@@ -323,6 +356,118 @@ void SimpleChargingDock::jointStateCallback(const sensor_msgs::msg::JointState::
...
+      RCLCPP_ERROR(
+        node_->get_logger(), "Calling detector service '%s' failed: %s",
+        detector_service_name_.c_str(), e.what());
+      return false;
+    }
+  }
+
+  // 2. Subscription toggle
+  //    Only subscribe once; will set state to ON on first message
+  if (subscribe_toggle_ && !dock_pose_sub_) {
```
SteveMacenski, line 391: I'm not directly asking you to do this, just a point of discussion:

So in the startup of the sensor, we could, in concept, have a while loop to spin/block for the perception timeout, then return true/false if we obtain data or not. That would remove the `initial_pose_received_` variable since we wouldn't continue if we didn't obtain it.

SteveMacenski:
@bkoensgen can you handle the conflict? Please open a PR on docs.nav2.org migration guide to add the new feature + Configuration Guide for the new parameters added

Otherwise LGTM. I'll do a quick once over before merging but seems good to me if CI passes now.

bkoensgen:
So MergyBot should be happy now, and I’ve created the PR. Here’s the link: https://github.com/ros-navigation/docs.nav2.org/pull/773

SteveMacenski (approved):

nav2_docking/opennav_docking/test/test_simple_non_charging_dock.cpp
```diff
@@ -98,7 +110,7 @@ TEST(SimpleNonChargingDockTests, StallDetection)
   pub->publish(msg);
   rclcpp::Rate r(2);
   r.sleep();
-  executor.spin_some();
+  rclcpp::spin_some(node->get_node_base_interface());
```
mini-1235, line 113: Hi, `rclcpp::spin_some` is going to be deprecated in the next release of rclcpp, this should not be changed, see more information in https://github.com/ros-navigation/navigation2/pull/5521

nav2_docking/opennav_docking/test/test_simple_charging_dock.cpp
```diff
@@ -140,7 +149,7 @@ TEST(SimpleChargingDockTests, StallDetection)
   pub->publish(msg);
   rclcpp::Rate r(2);
   r.sleep();
-  executor.spin_some();
+  rclcpp::spin_some(node->get_node_base_interface());
```
mini-1235, line 152: Same here

mini-1235:
This PR actually reverts some of the changes in https://github.com/ros-navigation/navigation2/pull/5521, I think we need to revert the changes related to `rclcpp::spin_some`

nav2_docking/opennav_docking/test/test_simple_non_charging_dock.cpp
SteveMacenski: @bkoensgen can you address this in a follow up PR?

nav2_docking/opennav_docking/test/test_simple_charging_dock.cpp
SteveMacenski: @bkoensgen can you address this in a follow up PR?

PR #5226, opened 2025-06-02
Route server corner smoothing

alexanderjyuen:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Gazebo simulation of ag vehicle |

---

* Added a corner arc class that stitches two successive edges with a circle
* Modified `densify` in `path_converter.cpp` to enable corner smoothing

* No documentation yet

* Changes were tested via launch testing to give smoothed path like in the image shown: 
image

---

* Unit testing needs to be written for `corner_arc` class
* Documentation for new parameters required

#### For Maintainers: 

### DIFF
--- a/nav2_route/CMakeLists.txt
+++ b/nav2_route/CMakeLists.txt
@@ -34,6 +34,7 @@ add_library(${library_name} SHARED
   src/route_server.cpp
   src/route_planner.cpp
   src/route_tracker.cpp
+  src/corner_arc.cpp
   src/edge_scorer.cpp
   src/operations_manager.cpp
   src/node_spatial_tree.cpp
--- a/nav2_route/include/nav2_route/corner_arc.hpp
+++ b/nav2_route/include/nav2_route/corner_arc.hpp
@@ -41,6 +41,12 @@ public:
 
   double getSignedAngleBetweenEdges(const EdgePtr start_edge, const EdgePtr end_edge);
 
+  bool isCornerValid(){ return valid_corner_; }
+
+  Coordinates getCornerStart(){ return start_coordinate_; }
+
+  Coordinates getCornerEnd(){ return end_coordinate_; }
+
 private:
   EdgePtr start_edge_;
   EdgePtr end_edge_;
@@ -48,9 +54,10 @@ private:
   double end_edge_length_;
   double minimum_radius_;
   double signed_angle_;
-  std::vector<double> start_point_;
-  std::vector<double> end_point_;
-  std::vector<double> circle_center_;
+  bool valid_corner_{false};
+  Coordinates start_coordinate_;
+  Coordinates end_coordinate_;
+  Coordinates circle_center_coordinate_;
 };
 
 }  // namespace nav2_route
--- a/nav2_route/include/nav2_route/corner_arc.hpp
+++ b/nav2_route/include/nav2_route/corner_arc.hpp
@@ -33,7 +33,7 @@ public:
 
   ~CornerArc() = default;
 
-  void interpolateArc(std::vector<geometry_msgs::msg::PoseStamped> & poses);
+  void interpolateArc(const double & max_angle_resolution, std::vector<geometry_msgs::msg::PoseStamped> & poses);
 
   double getEdgeLength(const EdgePtr edge);
 
--- a/nav2_route/include/nav2_route/corner_arc.hpp
+++ b/nav2_route/include/nav2_route/corner_arc.hpp
@@ -1,0 +1,58 @@
+// Copyright (c) 2025, Polymath Robotics
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NAV2_ROUTE__CORNER_ARC_HPP_
+#define NAV2_ROUTE__CORNER_ARC_HPP_
+
+#include <vector>
+#include <cmath>
+
+#include "nav2_route/types.hpp"
+#include "nav2_route/utils.hpp"
+
+namespace nav2_route
+{
+class CornerArc
+{
+public:
+  /**
+   * @brief Constructor
+   */
+  CornerArc(EdgePtr start_edge, EdgePtr end_edge, double minimum_radius);
+
+  ~CornerArc() = default;
+
+  void interpolateArc(std::vector<geometry_msgs::msg::PoseStamped> & poses);
+
+  double getEdgeLength(const EdgePtr edge);
+
+  double getAngleBetweenEdges(const EdgePtr start_edge, const EdgePtr end_edge);
+
+  double getSignedAngleBetweenEdges(const EdgePtr start_edge, const EdgePtr end_edge);
+
+private:
+  EdgePtr start_edge_;
+  EdgePtr end_edge_;
+  double start_edge_length_;
+  double end_edge_length_;
+  double minimum_radius_;
+  double signed_angle_;
+  std::vector<double> start_point_;
+  std::vector<double> end_point_;
+  std::vector<double> circle_center_;
+};
+
+}  // namespace nav2_route
+
+#endif  // NAV2_ROUTE__CORNER_ARC_HPP_
--- a/nav2_route/include/nav2_route/corner_arc.hpp
+++ b/nav2_route/include/nav2_route/corner_arc.hpp
@@ -41,11 +41,11 @@ public:
 
   double getSignedAngleBetweenEdges(const EdgePtr start_edge, const EdgePtr end_edge);
 
-  bool isCornerValid(){ return valid_corner_; }
+  bool isCornerValid() const { return valid_corner_; }
 
-  Coordinates getCornerStart(){ return start_coordinate_; }
+  Coordinates getCornerStart() const { return start_coordinate_; }
 
-  Coordinates getCornerEnd(){ return end_coordinate_; }
+  Coordinates getCornerEnd() const { return end_coordinate_; }
 
 private:
   EdgePtr start_edge_;
--- a/nav2_route/src/corner_arc.cpp
+++ b/nav2_route/src/corner_arc.cpp
@@ -1,0 +1,133 @@
+// Copyright (c) 2025, Polymath Robotics
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "nav2_route/corner_arc.hpp"
+#include <iostream>
+
+namespace nav2_route
+{
+
+CornerArc::CornerArc(EdgePtr start_edge, EdgePtr end_edge, double minimum_radius)
+{
+  start_edge_ = start_edge;
+  end_edge_ = end_edge;
+
+  double angle = getAngleBetweenEdges(start_edge, end_edge);
+
+  double tangent_length = minimum_radius/(std::tan(angle/2));
+
+  start_edge_length_ = getEdgeLength(start_edge);
+  end_edge_length_ = getEdgeLength(end_edge);
+
+  if(tangent_length < start_edge_length_ && tangent_length < end_edge_length_){
+    std::vector<double> start_edge_unit_tangent;
+    std::vector<double> end_edge_unit_tangent;
+    std::vector<double> unit_bisector;
+
+    start_edge_unit_tangent.push_back((start_edge->end->coords.x -start_edge->start->coords.x)/start_edge_length_);
+    start_edge_unit_tangent.push_back((start_edge->end->coords.y -start_edge->start->coords.y)/start_edge_length_);
+    end_edge_unit_tangent.push_back((end_edge->end->coords.x - end_edge->start->coords.x)/end_edge_length_);
+    end_edge_unit_tangent.push_back((end_edge->end->coords.y - end_edge->start->coords.y)/end_edge_length_);
+
+    double bisector_x = start_edge_unit_tangent[0]+end_edge_unit_tangent[0];
+    double bisector_y = start_edge_unit_tangent[1]+end_edge_unit_tangent[1];
+    double bisector_magnitude = std::sqrt(bisector_x*bisector_x + bisector_y*bisector_y);
+
+    unit_bisector.push_back(bisector_x/bisector_magnitude);
+    unit_bisector.push_back(bisector_y/bisector_magnitude);
+
+    start_point_.push_back(start_edge->end->coords.x - start_edge_unit_tangent[0]*tangent_length);
+    start_point_.push_back(start_edge->end->coords.y - start_edge_unit_tangent[1]*tangent_length);
+
+    end_point_.push_back(end_edge->start->coords.x + end_edge_unit_tangent[0]*tangent_length);
+    end_point_.push_back(end_edge->start->coords.y + end_edge_unit_tangent[1]*tangent_length);
+
+    double signed_angle = getSignedAngleBetweenEdges(start_edge, end_edge);
+
+    double bisector_length = minimum_radius/std::sin(signed_angle/2);
+
+    circle_center_.push_back(end_edge->start->coords.x + unit_bisector[0]*bisector_length);
+    circle_center_.push_back(end_edge->start->coords.y + unit_bisector[1]*bisector_length);
+
+    minimum_radius_ = minimum_radius;
+
+  }
+}
+
+void CornerArc::interpolateArc(std::vector<geometry_msgs::msg::PoseStamped> & poses)
+{
+
+  std::vector<double> r_start{ start_point_[0]-circle_center_[0], start_point_[1]-circle_center_[1] };
+  std::vector<double> r_end{ end_point_[0]-circle_center_[0],  end_point_[1]-circle_center_[1] };
+  double cross = r_start[0]*r_end[1] - r_start[1]*r_end[0];
+  double dot = r_start[0]*r_end[0] + r_start[1]*r_end[1];
+  signed_angle_ = std::atan2(cross, dot);
+  double max_angle_resolution = 0.1; //[rads]
+  int N = std::ceil(std::abs(signed_angle_)/max_angle_resolution);
+  double angle_resolution = signed_angle_/N;
+
+  float x, y;
+
+  for(int i = 0; i <= N; i++){
+    double angle = i*angle_resolution;
+    x = circle_center_[0] + (r_start[0]*std::cos(angle) - r_start[1]*std::sin(angle));
+    y = circle_center_[1] + (r_start[0]*std::sin(angle) + r_start[1]*std::cos(angle));
+    poses.push_back(utils::toMsg(x, y));
+  }
+}
+
+double CornerArc::getEdgeLength(const EdgePtr edge){
+    const Coordinates & start = edge->start->coords;
+    const Coordinates & end = edge->end->coords;
+
+    float dx = end.x - start.x;
+    float dy = end.y - start.y;
+
+    return double(sqrt(dx*dx + dy*dy));
+  }
+
+double CornerArc::getAngleBetweenEdges(const EdgePtr start_edge, const EdgePtr end_edge){
+
+    double start_dx = start_edge->start->coords.x - start_edge->end->coords.x;
+    double start_dy = start_edge->start->coords.y - start_edge->end->coords.y;
+
+    double end_dx = end_edge->end->coords.x - end_edge->start->coords.x;
+    double end_dy = end_edge->end->coords.y - end_edge->start->coords.y;
+
+    double angle = acos((start_dx*end_dx + start_dy*end_dy)/(getEdgeLength(start_edge)*getEdgeLength(end_edge)));
+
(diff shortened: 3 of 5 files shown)


nav2_route/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ add_library(${library_name} SHARED
   src/route_server.cpp
   src/route_planner.cpp
   src/route_tracker.cpp
+  src/corner_arc.cpp
```
SteveMacenski: I'd love this as a header-only file as a utility.

nav2_route/src/path_converter.cpp
```diff
@@ -121,4 +177,14 @@ void PathConverter::interpolateEdge(
   }
 }
 
+double PathConverter::getEdgeLength(const EdgePtr edge){
```
SteveMacenski: I think this could be made a member of `DirectionalEdge` instead. In that, you should compute this using `hypotf` instead and return a `float` instead of a `double` since the coordinates themselves area only `float`.

nav2_route/src/path_converter.cpp
```diff
@@ -54,19 +63,66 @@ nav_msgs::msg::Path PathConverter::densify(
...
+    end = edge->end->coords;
+
+    CornerArc corner_arc(edge, next_edge, smoothing_radius_);
+
+    if(corner_arc.isCornerValid() && smooth_corners_){
+      //if an arc exists, end of the first edge is the start of the arc
+      end = corner_arc.getCornerStart();
+      interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+
+      corner_arc.interpolateArc(angle_of_interpolation_, path.poses);
```
SteveMacenski: `static_cast` the parameter down to a float and pass in. Update the CornerArc API appropriately.

nav2_route/src/path_converter.cpp
```diff
@@ -54,19 +63,66 @@ nav_msgs::msg::Path PathConverter::densify(
...
-    interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+
+  for (unsigned int i = 0; i < edges.size() - 1; i++) {
+
+    const EdgePtr edge = edges[i];
+    const EdgePtr next_edge = edges[i+1];
+
+    end = edge->end->coords;
+
+    CornerArc corner_arc(edge, next_edge, smoothing_radius_);
```
SteveMacenski: `static_cast` the parameter down to a float and pass in. Update the CornerArc API appropriately.

nav2_route/include/nav2_route/corner_arc.hpp
```diff
@@ -0,0 +1,65 @@
...
+   */
+  CornerArc(EdgePtr start_edge, EdgePtr end_edge, double minimum_radius);
+
+  ~CornerArc() = default;
+
+  void interpolateArc(const double & max_angle_resolution, std::vector<geometry_msgs::msg::PoseStamped> & poses);
+
+  double getEdgeLength(const EdgePtr edge);
+
+  double getAngleBetweenEdges(const EdgePtr start_edge, const EdgePtr end_edge);
```
SteveMacenski: A few of these I think can be protected methods since they really shouldn't need to be used by applications

nav2_route/src/corner_arc.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void CornerArc::interpolateArc(const double & max_angle_resolution, std::vector<geometry_msgs::msg::PoseStamped> & poses)
+{
+
+  std::vector<double> r_start{ start_coordinate_.x-circle_center_coordinate_.x, start_coordinate_.y-circle_center_coordinate_.y };
+  std::vector<double> r_end{ end_coordinate_.x-circle_center_coordinate_.x,  end_coordinate_.y-circle_center_coordinate_.y };
+  double cross = r_start[0]*r_end[1] - r_start[1]*r_end[0];
+  double dot = r_start[0]*r_end[0] + r_start[1]*r_end[1];
+  signed_angle_ = std::atan2(cross, dot);
```
SteveMacenski: I only see this being used in this function. I don't think it needs to be a class member

nav2_route/src/path_converter.cpp
```diff
@@ -20,7 +20,6 @@
 #include <algorithm>
 
 #include "nav2_route/path_converter.hpp"
-
```
SteveMacenski: Readd

nav2_route/src/path_converter.cpp
```diff
@@ -30,9 +29,19 @@ void PathConverter::configure(nav2_util::LifecycleNode::SharedPtr node)
   nav2_util::declare_parameter_if_not_declared(
     node, "path_density", rclcpp::ParameterValue(0.05));
   density_ = static_cast<float>(node->get_parameter("path_density").as_double());
+  nav2_util::declare_parameter_if_not_declared(
+    node, "smoothing_radius", rclcpp::ParameterValue(1.0));
+  smoothing_radius_ = node->get_parameter("smoothing_radius").as_double();
+  nav2_util::declare_parameter_if_not_declared(
+    node, "angle_of_interpolation", rclcpp::ParameterValue(0.1));
```
SteveMacenski: Would it be more natural for this to be a distance instead of an angle for parameterization? Or do you need an angle? I would think density of points would make more sense here - and we have a parameter `path_density` that could be used directly

nav2_route/src/path_converter.cpp
```diff
@@ -54,19 +63,66 @@ nav_msgs::msg::Path PathConverter::densify(
...
+      corner_arc.interpolateArc(angle_of_interpolation_, path.poses);
+
+      //if an arc exists, start of the subsequent edge is end of the arc
+      auto new_start_node = std::make_unique<Node>(*next_edge->start);
+      new_start_node->coords = corner_arc.getCornerEnd();
+      auto new_next_edge   = std::make_unique<DirectionalEdge>(*next_edge);
+      new_next_edge->start = new_start_node.get();
+
+      //reassign the next edge to have the modified start coordinate
+      edges[i+1] = new_next_edge.get();
```
SteveMacenski: This... seems really convoluted (perhaps AI assisted? :wink:)

We have an vector `edges`. Just logically speaking, if we have a route of edges, those edges should never change. Those are our route. So, why are we copying the `route.edges`?  

`edges[i+1] = new_next_edge.get();` --> `new_next_edge   = std::make_unique<DirectionalEdge>(*next_edge);` --> `EdgePtr next_edge = edges[i+1];`. These are the exact same thing. 

I think I understand why you're doing this and I think it can be simplified. When we iterate 0->N-1, we have the last edge we need to populate which could have the corner applied. So, we just need to store outside of the loop the `edge_start` (where the edge starts after the last circle ends) so that we can use it after the loop terminates. Bonus, that can store the state between loop iterations as well for the same reason. We don't need to modify the underlying edge at all. Since, as far as I can tell, that is the only thing that the `new_nodes` and `new_edges` actually modify 

```
new_start_node->coords = corner_arc.getCornerEnd();
```

nav2_route/include/nav2_route/corner_arc.hpp
```diff
@@ -0,0 +1,63 @@
...
+  bool isCornerValid() const { return valid_corner_; }
+
+  Coordinates getCornerStart() const { return start_coordinate_; }
+
+  Coordinates getCornerEnd() const { return end_coordinate_; }
+
+private:
+  EdgePtr start_edge_;
+  EdgePtr end_edge_;
+  float start_edge_length_;
```
SteveMacenski: These 2 are only used in 1 function, similarly this could be removed from class membership

nav2_route/include/nav2_route/corner_arc.hpp
alexanderjyuen: moved getAnglbetweenEdges and getSignedAngleBetweenEdges to protected, removed getEdgeLength

nav2_route/CMakeLists.txt
alexanderjyuen: would you want this as another class defined inside `utils.hpp` or `types.hpp` or would you just want the entire class defined in `corner_arc.hpp`

SteveMacenski: I'd say `corner_arc.hpp`. It seems too big for just the tiny utils bin. 

Maybe rename `corner_smoothing.hpp`?

nav2_route/src/path_converter.cpp
alexanderjyuen: removed `angle_of_interpolation` and infering it with the following:

```
corner_arc.interpolateArc(density_/smoothing_radius_, path.poses);
```

nav2_route/CMakeLists.txt
alexanderjyuen: removed `corner_arc.cpp`, everything is in `corner_smoothing.hpp` now

nav2_route/src/path_converter.cpp
alexanderjyuen: Changed the API for corner arc, above code is no longer relevant

nav2_route/include/nav2_route/types.hpp
```diff
@@ -194,6 +196,13 @@ struct Node
   }
 };
 
+inline float DirectionalEdge::getEdgeLength()
```
SteveMacenski, line 199: Is there a reason this isn't defined in the struct itself?

nav2_route/src/path_converter.cpp
```diff
@@ -30,9 +30,16 @@ void PathConverter::configure(nav2_util::LifecycleNode::SharedPtr node)
   nav2_util::declare_parameter_if_not_declared(
     node, "path_density", rclcpp::ParameterValue(0.05));
   density_ = static_cast<float>(node->get_parameter("path_density").as_double());
+  nav2_util::declare_parameter_if_not_declared(
+    node, "smoothing_radius", rclcpp::ParameterValue(1.0));
```
SteveMacenski, line 34: Add to Readme + Docs configuration guide (ditto on smooth_corners)

nav2_route/src/path_converter.cpp
```diff
@@ -30,9 +30,16 @@ void PathConverter::configure(nav2_util::LifecycleNode::SharedPtr node)
...
+  smoothing_radius_ = static_cast<float>(node->get_parameter("smoothing_radius").as_double());
+  nav2_util::declare_parameter_if_not_declared(
+    node, "smooth_corners", rclcpp::ParameterValue(false));
```
SteveMacenski, line 37: I support the false default in the code, but I'd like it to be true in the Nav2 params file: https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/params/nav2_params.yaml#L442

nav2_route/src/path_converter.cpp
```diff
@@ -54,17 +61,44 @@ nav_msgs::msg::Path PathConverter::densify(
     interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
   }
 
+  Coordinates start = route.edges[0]->start->coords;
+  Coordinates end;
+
   // Fill in path via route edges
-  for (unsigned int i = 0; i != route.edges.size(); i++) {
+
```
SteveMacenski:
```suggestion
```

nav2_route/src/path_converter.cpp
```diff
@@ -54,17 +61,44 @@ nav_msgs::msg::Path PathConverter::densify(
...
     const EdgePtr edge = route.edges[i];
-    const Coordinates & start = edge->start->coords;
-    const Coordinates & end = edge->end->coords;
-    interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+    const EdgePtr next_edge = route.edges[i + 1];
+
+    end = edge->end->coords;
+
+    CornerArc corner_arc(start, end, next_edge->end->coords, smoothing_radius_);
+
```
SteveMacenski:
```suggestion
```

nav2_route/src/path_converter.cpp
```diff
@@ -54,17 +61,44 @@ nav_msgs::msg::Path PathConverter::densify(
...
+      // interpolate to start of arc
+      interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+
+      // interpolate arc
+      corner_arc.interpolateArc(density_ / smoothing_radius_, path.poses);
+
+      // new start of next edge is end of smoothing arc
+      start = corner_arc.getCornerEnd();
+    } else {
+      interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
```
SteveMacenski: Do an `if (smooth_corners_)` log warning that the edges (print IDs) couldn't make a curve between them

nav2_route/src/path_converter.cpp
```diff
@@ -54,17 +61,44 @@ nav_msgs::msg::Path PathConverter::densify(
...
   // Fill in path via route edges
-  for (unsigned int i = 0; i != route.edges.size(); i++) {
+
+  for (unsigned int i = 0; i < route.edges.size() - 1; i++) {
     const EdgePtr edge = route.edges[i];
-    const Coordinates & start = edge->start->coords;
-    const Coordinates & end = edge->end->coords;
-    interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+    const EdgePtr next_edge = route.edges[i + 1];
+
```
SteveMacenski:
```suggestion
```

nav2_route/src/path_converter.cpp
```diff
@@ -54,17 +61,44 @@ nav_msgs::msg::Path PathConverter::densify(
...
   // Fill in path via route edges
-  for (unsigned int i = 0; i != route.edges.size(); i++) {
+
+  for (unsigned int i = 0; i < route.edges.size() - 1; i++) {
     const EdgePtr edge = route.edges[i];
-    const Coordinates & start = edge->start->coords;
-    const Coordinates & end = edge->end->coords;
-    interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+    const EdgePtr next_edge = route.edges[i + 1];
```
SteveMacenski:
```suggestion
    const EdgePtr & next_edge = route.edges[i + 1];
```

nav2_route/src/path_converter.cpp
```diff
@@ -54,17 +61,44 @@ nav_msgs::msg::Path PathConverter::densify(
...
+      interpolateEdge(start.x, start.y, end.x, end.y, path.poses);
+      start = end;
+    }
   }
 
   if (route.edges.empty()) {
     path.poses.push_back(utils::toMsg(route.start_node->coords.x, route.start_node->coords.y));
   } else {
+    interpolateEdge(
+      route.edges.back()->start->coords.x, route.edges.back()->start->coords.y,
```
SteveMacenski: Shouldn't this actually be `start` so that it uses the last interpolated curve?

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,134 @@
...
+    std::vector<double> r_end{end_coordinate_.x - circle_center_coordinate_.x,
+      end_coordinate_.y - circle_center_coordinate_.y};
+    double cross = r_start[0] * r_end[1] - r_start[1] * r_end[0];
+    double dot = r_start[0] * r_end[0] + r_start[1] * r_end[1];
+    double signed_angle = std::atan2(cross, dot);
+    int N = std::ceil(std::abs(signed_angle) / max_angle_resolution);
+    double angle_resolution = signed_angle / N;
+
+    float x, y;
+
```
SteveMacenski:
```suggestion
```

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,134 @@
...
+#include <vector>
+#include <cmath>
+
+#include "nav2_route/types.hpp"
+#include "nav2_route/utils.hpp"
+
+namespace nav2_route
+{
+class CornerArc
```
SteveMacenski, line 32: In this class: convert all doubles to floats, there's mixing going on which has burned me in the past in planners. I don't think anything here needs double percision, so you could just search + replace for this file and be OK

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,134 @@
...
+      {(start.x - corner.x) / start_edge_length_, (start.y - corner.y) / start_edge_length_};
+      std::vector<double> end_edge_unit_tangent =
+      {(end.x - corner.x) / end_edge_length_, (end.y - corner.y) / end_edge_length_};
+
+      double bisector_x = start_edge_unit_tangent[0] + end_edge_unit_tangent[0];
+      double bisector_y = start_edge_unit_tangent[1] + end_edge_unit_tangent[1];
+      double bisector_magnitude = std::sqrt(bisector_x * bisector_x + bisector_y * bisector_y);
+
+      std::vector<double> unit_bisector =
+      {bisector_x / bisector_magnitude, bisector_y / bisector_magnitude};
```
SteveMacenski, line 70: Just validating: is there any chance that the bisectors magnitudes could be  0? If so, we should check for NaN / Inf and set the variables appropriately for a failed corner attempt

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,134 @@
...
+    const Coordinates & end)
+  {
+    double start_dx = start.x - corner.x;
+    double start_dy = start.y - corner.y;
+
+    double end_dx = end.x - corner.x;
+    double end_dy = end.y - corner.y;
+
+    double angle =
+      acos((start_dx * end_dx + start_dy * end_dy) / (start_edge_length_ * end_edge_length_));
```
SteveMacenski, line 160: Ditto on divide by zero

nav2_route/include/nav2_route/types.hpp
alexanderjyuen: `colcon` was giving compilation errors as `Node` was defined after `DirectionalEdge` and it seemed `getEdgeLength()` requires `Node` to be defined first to compile successfully, I tried to move  `DirectionalEdge` after `Node` but then I got compilation errors for `Node` as `addEdge` seemed to require `DirectionalEdge` be defined. Is there a nicer way to stream line this?

nav2_route/include/nav2_route/corner_smoothing.hpp
alexanderjyuen: would this: https://github.com/alexanderjyuen/navigation2/blob/route_server_corner_smoothing/nav2_route/include/nav2_route/corner_smoothing.hpp#L41

be sufficient to avoid the divide by zero since it will never hit this method if either lengths are 0.0? Or is it better practice to put that check into the method?

nav2_route/include/nav2_route/types.hpp
SteveMacenski: Good point - I think this is the right answer. I don't think a forward declare is enough in this case since we access members of Node.

nav2_route/include/nav2_route/corner_smoothing.hpp
SteveMacenski: More divide by zero potential cases:

- https://github.com/alexanderjyuen/navigation2/blob/route_server_corner_smoothing/nav2_route/include/nav2_route/corner_smoothing.hpp#L66
- https://github.com/alexanderjyuen/navigation2/blob/route_server_corner_smoothing/nav2_route/include/nav2_route/corner_smoothing.hpp#L45
- https://github.com/alexanderjyuen/navigation2/blob/route_server_corner_smoothing/nav2_route/include/nav2_route/corner_smoothing.hpp#L48-L51 (for the numerators, if start/center or end/center are zero in both XY to make the sum in `bisector_x` and `bisector_y` zero)

SteveMacenski: you have a few different divide by zero cases, several of which I think you'd run into if you applied this work on the uses I know you have (the ag field, representing rows) since you do a tan and sin function. Those denominators can be 0 when the angle is either 0 (for the pairs being A->B->A, or A->B->C where C is A) or when the angle is 180 which would just be a straight line. I think you need to check the angles for validity or `return` without setting the validity parameter.

0 or PI are invalid divide by zero cases -- it would surprise me if you didn't run into these yourself if you tried to use it

SteveMacenski:
@alexanderjyuen there's a few open comments still - make sure to expand the collapsed "See 3 more comments" (or whatever number it is) to see a few there that are still open. Its all very small now.

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,141 @@
...
+  {
+    start_edge_length_ = hypotf(corner.x - start.x, corner.y - start.y);
+    end_edge_length_ = hypotf(end.x - corner.x, end.y - corner.y);
+
+    // invalid scenario would cause equations to blow up
+    if (start_edge_length_ == 0.0 || end_edge_length_ == 0.0) {return;}
+
+    float angle = getAngleBetweenEdges(start, corner, end);
+
+    // no need to smooth an angle that is very shallow
```
SteveMacenski:
```suggestion
    // cannot smooth a 0 (back and forth) or 180 (straight line) angle
```

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include "nav2_route/types.hpp"
+#include "nav2_route/utils.hpp"
+
+namespace nav2_route
+{
+class CornerArc
+{
+public:
+  /**
+   * @brief Constructor
```
SteveMacenski: Add doxygen for each method and for the class in whole

nav2_route/src/path_converter.cpp
alexanderjyuen: README updated, docs getting updated here: https://github.com/ros-navigation/docs.nav2.org/pull/708

SteveMacenski (approved):

SteveMacenski:
@alexanderjyuen just missing the migration guide for docs and we're good to go!

nav2_route/include/nav2_route/corner_smoothing.hpp
```diff
@@ -0,0 +1,171 @@
...
+      valid_corner_ = true;
+    }
+  }
+
+  /**
+   * @brief A destructor for nav2_route::CornerArc
+   */
+  ~CornerArc() = default;
+
+  void interpolateArc(
```
SteveMacenski, line 97:
```suggestion
  /**
   * @brief interpolates the arc for a path of certain density
   * @param max_angle_resolution Resolution to interpolate path to
   * @param poses Pose output
   */
  void interpolateArc(
```

PR #5235, opened 2025-06-04
Fix/smac planner orientation goals

stevedanomodolor:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 


nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * @brief Sets the internal goals' state list to the provided goals.
+   * @param goals Vector of goals state to set,
+   */
+  void setGoalStates(
+    GoalStateVector & goals_state)
+  {
+    clear();
+    _goals_state = goals_state;
```
SteveMacenski, line 83: Copy: does this make any problem?

SteveMacenski:
@wang-fire can you test? This should fix the issue theoretically

nav2_smac_planner/include/nav2_smac_planner/types.hpp
```diff
@@ -188,6 +188,18 @@ struct MotionPrimitive
...
 
+  /**
+   * @struct nav2_smac_planner::GoalState
+   * @brief A struct to store the goal state
+   */
+
+template<typename NodeT>
+struct GoalState
+{
+  NodeT * goal = nullptr;
```
SteveMacenski, line 198: So we do `NodePtr &` in alot of places and while I don't see an issue with this implementation, it does trigger my defensive programming to not want to change things as they were occurring before. All the other parts of the code related to expanding nodes (getting neighbors, queuing them in the search graph, etc) use the `&()` pointer returned from `addToGraph` and wrapped into the `NeighorGetter`. This would break that full pipeline of everything using the raw memory received from the data structure. I don't see in the current implementation where that would break anything, but I wouldn't like to find out since that would create very subtle bugs. 

 `_goal` was set as the output of `addToGraph` which returns a reference to the direct memory inserted in the data structure. `goal` now being coped into `GoalState` changes that behavior.

if we made this into a `NodeT * &` that would resolve it and the constructor would turn into `&goal` instead. 

---

Honestly though for this comment and others, this is mostly a precautionary item. I looked and _I don't think_ I see any issues, but it is inconsistent styling with the rest of the planner so it gives me pause

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
SteveMacenski: This also copies

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
```diff
@@ -0,0 +1,215 @@
...
+    _goals_coordinate.clear();
+  }
+
+  /**
+   * @brief Populates coarse and fine goal lists for analytic expansion.
+   * @param coarse_check_goals Output list of goals for coarse search expansion.
+   * @param fine_check_goals Output list of goals for fine search refinement.
+   * @param coarse_search_resolution Number of fine goals per coarse goal.
+   */
+  void prepareGoalsForAnalyticExpansion(
```
SteveMacenski, line 100: prepareGoalsForAnalyticExpansion and removeInvalidGoals seem to copy the goal pointers as well for the check goals and goals set.

I don't think the analytic expansion is problemaic since the raw memory is never used in that process, only accessing data from its stored memory.

SteveMacenski: Goals set concerns me the most since that's used for 

```
  inline bool isGoal(NodePtr & node)
  {
    return _goals_set.find(node) != _goals_set.end();
  }
```

Which I think should work properly - but since we compare the pointers directly its not necessarily clear that this is the case. I think there's no comparisons for the analytic expansion, so that's OK to copy since we don't look at the memory. 

NodeSet could be `std::unordered_set<NodePtr &>` and `_goals_set.insert(&_goals_state[i].goal);`

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
stevedanomodolor: @SteveMacenski Do you suggest I do a move instead?
_goal_manager.setGoalStates(std::move(goals_state));
_goals_state = std::move(goals_state);

Out of curiosity, would the copy be a problem?, given now that I store a ref to the Node*?.

SteveMacenski: Yeah `std::move` would work. The other option is to change the API to be `_goal_manager.addGoal(NodePtr &)` and have the `goal` based by reference `&goal`. Then it could use that to populate the `_goal_states` internally with the `true` status. 

That actually might be a cleaner API in general. At the start of the setGoals in A*, you could just have a `_goal_manager.clear()`  and then `_goal_manager.addGoal()` for each goal. That feels pretty ergonomic from the A* application side

SteveMacenski: > Out of curiosity, would the copy be a problem?, 

When the object is constructed, the object address pointer is still correct but the instance of the pointer is different, so its address is different.  Everywhere else we store the actual memory given by `addToGraph` so if there's any behavior that is relying on the goal pointers to be exactly equal, then this would fail that check. 

Frankly, I do not think this is necessary. However the rest of the planner does this and there was behavior (which may be gone now?) that relied on that fact. I'm being probably overly cautious. 

Looking at the underlying behavior I want to keep in tact (maybe unnecessarily):

```
int x = 5;
int* a = &x;
int* b = &x;

a == b; // true, because both point to the same address (&x)
```

```
int x = 5;
int* a = &x;
int* b = a;

a == b; // <-- Also true, since the comparison is off of the address of the object, not the address of the pointer instance
&a == &b;     // false: different pointer **variables** stored at different addresses
```

The `&node == &my_goal` is what I think before this PR would be true and want to maintain in case there's something subtle still that relies on it. We could spend a bunch of time to prove to ourselves its not necessary anywhere, but I think its more expedient just to make sure we maintain this address in the ~3x places or so its needed (`_goal_states` and anyone that accesses `_goal_states[i]->goal` to take the address `&`) 

> given now that I store a ref to the Node*?.

I guess I don't see that yet, I don't think that's been pushed :smile:

stevedanomodolor: All my changes are local sorry,
I like the addgoal and clear idea. I will modify that. 
In the first example(actually in both cases), 
&a == &b  //false because &a or &b are different pointers to the same pointer address. 
The reason why i dont think it is a problem is because we pass it through functions as references. Actually the comparision is the pointer to the goal itself.

I actually cant do this in cpp, i think

```
template<typename NodeT>
struct GoalState
{
  NodeT * &goal;
  bool is_valid = true;
};
```

The other alternative here is to mantain as is, and when i call the add goal

```

template<typename NodeT>
struct GoalState
{
  NodeT * goal;
  bool is_valid = true;
};

  void addGoal(NodePtr goal)
  {
    GoalState<NodeT> state;
    state.goal = goal;
    state.is_valid = true;  


```

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
stevedanomodolor: I am not able to do this with an unordered set(Nodeptr &)
NodeSet could be std::unordered_set<NodePtr &> and _goals_set.insert(&_goals_state[i].goal);

SteveMacenski: I'm not sure I understand, isn't `unordered set(Nodeptr &)` and `std::unordered_set<NodePtr &> ` describing the same thing?

stevedanomodolor: My bad, sorry, bad explanation from my part. What I meant was I can't do an unordered_set<NodePtr &> in cpp i guess because the unorder set only allows hashable and copiable types.

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
stevedanomodolor: Sorry I have the changes locally  jajaj. 

I can change the Goal state to use the NodePtr * & but vectors and unorder set cant store that variable type. 
As of the state of the code as I understood it(maybe I am wrong here),  the memory pointer has to be the same, independent of if variable is an alias or not. Since I never create a new instance of the ointer like 

goal_new = new NodeT, this should not be a problem, basically I dont have two pointer to the goal address.

stevedanomodolor: Also an example is this

```
template<typename NodeT>
typename AStarAlgorithm<NodeT>::NodePtr & AStarAlgorithm<NodeT>::getGoal()
{
  return _goal;
}
```
The reason why this works is because we return the same pointer and not new instance Nodeptr because of the typename AStarAlgorithm<NodeT>::NodePtr &. 

maybe there is a way to enforce this so it somehow to prevent someone from easily changing this behavior. But if somehow  someone changes this behavior, the planner should fail spectacularly, I am an example of this :laughing:

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
SteveMacenski: Curse you `unordered_set`! 

We could change `GoalState` to use a `Node * &` so we store that up until the `_goals_set.insert(_goals_state[i].goal);`, but that seems moot except if `coarse_check_goals` or `fine_check_goals` required it, which I'm confident it doesn't. If you're saying that `isGoal` works then, that's good enough and consistent with my understanding of `find()` on pointers. 

I won't block over it, but my thoughts: I don't love that this chain is broken, it makes me want to break it down more to remove even the inkling that this is possible to compare `&node == &goal` (i.e. remove all the `NodePtr &` to `NodePtr`s. At this point, it looks like I did some crazy things for no point (**cough** and perhaps I did **cough**) and it could be simplified. 

I do like the idea of add the `add()` API the goal manager though (and `clear()` on a new goal to the A*) regardless, it seems more ergonomic than `setGoalStates` since we don't to expose the goal states concept or the boolean to the A* scope. It would hide more implementation details about the goal manager and just have the `manager.add(goal)` seems nice and understandable.

nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
```diff
@@ -154,12 +163,6 @@ class AStarAlgorithm
    */
   NodePtr & getStart();
```
SteveMacenski, line 164: Remove `&`?

SteveMacenski: - addNode
- getHeuristicCost
- onVisitationCheckNode
- populateExpansionsLog
- getStart

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+  }
+
+  /**
+   * @brief Check if a given node is part of the goal set.
+   * @param node Node pointer to check.
+   * @return if node matches any goal in the goal set.
+   */
+  inline bool isGoal(const NodePtr & node)
```
SteveMacenski, line 152: `&`?

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
```diff
@@ -0,0 +1,218 @@
...
+  bool goalsIsEmpty()
+  {
+    return _goals_state.empty();
+  }
+
+  /**
+   * @brief Adds goal to the goal vector
+   *@param goal Reference to the NodePtr
+   */
+  void addGoal(NodePtr & goal)
```
SteveMacenski, line 79: `&`?

nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
stevedanomodolor: Now that i think of it, I would maintain all const &, it is more efficient to prevent copying, since its a constant, there shouldnt be the fear of modifying the data. In the end I am moving aroung pointers, so I would leave it as is.

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
stevedanomodolor: same as before.

nav2_smac_planner/include/nav2_smac_planner/a_star.hpp
stevedanomodolor: There is the argument that since it is a pointer, it is small, but still I woul leave it as is.

nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
```diff
@@ -59,7 +65,33 @@ class AnalyticExpansion
...
 
-  typedef std::vector<AnalyticExpansionNode> AnalyticExpansionNodes;
+  /**
+   * @struct AnalyticExpansionNodes
+   * @brief Analytic expansion nodes and associated metadata
+   *
+   * This structure holds a collection of analytic expansion nodes and the number of direction
+   * changes encountered during the expansion.
+   */
+  struct AnalyticExpansionNodes
```
SteveMacenski, line 75: Wait - why isn't this being used? It should be being called

SteveMacenski: Huh I just checked - it doesn't seem (?) like analytic expansions are specifically tested anywhere. 

Can you add a high level smoke test that calls the smac planner to plan using the reeds-shepp motion model to exercise all of this

SteveMacenski:
@MarcoMatteoBassa this is ready for testing again!

nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
stevedanomodolor: it is i think.
https://github.com/ros-navigation/navigation2/pull/5216/files. It is from this pr.

nav2_smac_planner/include/nav2_smac_planner/goal_manager.hpp
stevedanomodolor: I am not sure on this one though.

MarcoMatteoBassa:
Looks good now :)

nav2_smac_planner/include/nav2_smac_planner/analytic_expansion.hpp
stevedanomodolor: There is a test_a_star_analytic_expansion that tests the anlyticalñ expansion with reeds shep, we should be good.

PR #5245, opened 2025-06-08
Add parameter publish_zero_velocity from branch main->humble

haider8645:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | ([Arter](https://robotik.dfki-bremen.de/de/forschung/robotersysteme/arter)) |

---

* We need to have a smooth transition in between segments of a path without the linear and angular velocities resetted to zero. I copied the change from the main to the humble branch.

AS-IS:
The humble branch currently resets the cmd_vel to zero upon goal exit. This results in unwanted /cmd_vel jumps to 0 for robots which required a piece-wise path following approach over a segmented global path.
image

TO-BE:
After the parameter is added and set to false the cmd's have smooth transitions in between segments.
image

Added new parameter, so need to add that to default configs and documentation page. The humble version of nav2 will have a new parameter called ```publish_zero_velocity: true```

I tested it on a piece-wise global path of a mobile robot and the resulting cmd_vel has smooth transitions in between path segment transitions.

#### For Maintainers: 

### DIFF
--- a/nav2_controller/src/controller_server.cpp
+++ b/nav2_controller/src/controller_server.cpp
@@ -60,6 +60,7 @@ ControllerServer::ControllerServer(const rclcpp::NodeOptions & options)
   declare_parameter("speed_limit_topic", rclcpp::ParameterValue("speed_limit"));
 
   declare_parameter("failure_tolerance", rclcpp::ParameterValue(0.0));
+  declare_parameter("publish_zero_velocity", rclcpp::ParameterValue(true));
 
   // The costmap node is used in the implementation of the controller
   costmap_ros_ = std::make_shared<nav2_costmap_2d::Costmap2DROS>(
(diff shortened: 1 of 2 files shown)


nav2_controller/src/controller_server.cpp
```diff
@@ -569,16 +570,19 @@ void ControllerServer::publishVelocity(const geometry_msgs::msg::TwistStamped &
...
-  velocity.twist.angular.x = 0;
-  velocity.twist.angular.y = 0;
-  velocity.twist.angular.z = 0;
-  velocity.twist.linear.x = 0;
-  velocity.twist.linear.y = 0;
-  velocity.twist.linear.z = 0;
-  velocity.header.frame_id = costmap_ros_->getBaseFrameID();
-  velocity.header.stamp = now();
-  publishVelocity(velocity);
+  if (get_parameter("publish_zero_velocity").as_bool()) {
```
SteveMacenski: Store as a class member, I don't like getting parameters in the application

Also, please implement analog to the main branch implementation: the `if` check should be where `publishZeroVelocity` is called, not internal to the function.

SteveMacenski:
No issue in concept, just minor nits. Also, make sure to fix the DCO job

nav2_controller/src/controller_server.cpp
haider8645: - [x] Add publish_zero_velocity as class member,
- [x] Add IF where the function  publishZeroVelocity is called

haider8645:
@SteveMacenski I have added the `if` condition with `publish_zero_velocity` only in the case of a successful goal. In contrast to the implementation in the `main` branch, for all other cases except goal success, it still makes sense to directly publish zero velocity. Do you agree? Please see the table below.

| Exit Case                  | Code Publishes Zero Velocity? | Comments                                                                                  |
|----------------------------|--------------------------------------|-------------------------------------------------------------------------------------------|
| Goal Reached               | If `publish_zero_velocity_` is true  | Optional stop; often desired to ensure robot halts exactly at goal                        |
| Canceled                   | Yes                                  | Always stops immediately to ensure safe halt on user request                              |
| Exception/Failure          | Yes                                  | Ensures robot halts safely if an error interrupts control                                 |
| Action Server Inactive     | No (just returns)                    | Could be risky if missed; may want to ensure zero velocity in this rare exit case         |

I am of the opinion that the `publish_zero_velocity` parameter should apply only to normal working conditions—specifically, when a goal is successfully reached and executed. In any exceptional cases, such as exceptions or failures, zero velocity should always be published to bring the robot to a safe stop, regardless of the parameter setting.

SteveMacenski:
I agree with that!

PR #5246, opened 2025-06-08
[BtActionServer] default enable_groot_monitoring_ to false to solve `Original error: Address already in use` migration error

doisyg:
---

## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Dexory simulation|

---

Since https://github.com/ros-navigation/navigation2/pull/5065, groot monitoring is enabled by default for all `BtActionServer`s : https://github.com/ros-navigation/navigation2/blob/7f561b0f5aa8e229334bb6f71a7a99e6074efd75/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp#L288

This causes issue for users of simple `BtActionServer`s that by default are not calling `setGrootMonitoring`: https://github.com/ros-navigation/navigation2/blob/7f561b0f5aa8e229334bb6f71a7a99e6074efd75/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server_impl.hpp#L239.
And hence, if they have more than one such action, they are faced with a `Original error: Address already in use` error.

Full launch of Dexory stack

---

For non `BehaviorTreeNavigator` actions, it would have been simpler to have `enable_groot_monitoring` and `groot_server_port` as parameters of `BtActionServer` (like for instance `bt_loop_duration`, `default_server_timeout`, `wait_for_service_timeout` and `always_reload_bt_xml`) but I understand the need for being able to set different values for 2 different pluggins of 1 `bt_navigator`.
Or maybe a future work could combine both ? 

#### For Maintainers: 

### DIFF
--- a/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
+++ b/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_server.hpp
@@ -285,7 +285,7 @@ protected:
   bool always_reload_bt_xml_ = false;
 
   // Parameters for Groot2 monitoring
-  bool enable_groot_monitoring_ = true;
+  bool enable_groot_monitoring_ = false;
   int groot_server_port_ = 1667;
 
   // User-provided callbacks


ajtudela:
Hi @doisyg, I implemented the feature. Could you provide more details about your implementation so I can troubleshoot the error?

If I understand correctly, you have two ```BtActionServer``` running at the same time and, because ```enable_groot_monitoring_``` is set in the class, the ports collide and throw an error.

I apologize for that; I didn't consider a use case for ```BtActionServer``` outside of Navigators.

I'll look into combining the parameters of the ```BtActionServer``` with the navigator plugins.

doisyg:
> If I understand correctly, you have two `BtActionServer` running at the same time and, because `enable_groot_monitoring_` is set in the class, the ports collide and throw an error.

Exactly, and more than 2!

> I apologize for that; I didn't consider a use case for `BtActionServer` outside of Navigators.

No worries, as there are no such use cases in nav2, it was not necessarily straightforward to foresee.

> I'll look into combining the parameters of the `BtActionServer` with the navigator plugins.

I am not even 100% sure it is worth the trouble. If a non Navigators user of a `BtActionServer` needs Groot monitoring, he could for now call `setGrootMonitoring` in its code. I will let @SteveMacenski recommend the best solution.

SteveMacenski (approved):
This is very reasonable, we disable it anyway in the nav2 params defaults https://github.com/ros-navigation/navigation2/blob/7f561b0f5aa8e229334bb6f71a7a99e6074efd75/nav2_bringup/params/nav2_params.yaml#L55 I think this is the right answer

Issue #5213, opened 2025-05-30
Enable Pause/Resume in NavigateThroughPoses & NavigateToPose


tonynajjar:
#### Feature description

https://github.com/ros-navigation/navigation2/pull/4994 introduced the concept of WaypointStatus to keep track of the PENDING vs COMPLETED waypoints and that works well for one-shot call to NavigateThroughPoses. 
Typically these missions can run for longer and it would be desirable to pause and resume. 
This feature request is therefore to implement a pause/resume capability.

#### Implementation considerations

We started some implementation discussions already [here](https://github.com/ros-navigation/navigation2/pull/4994#issuecomment-2916001729)

My custom approach:

```
Here is roughly how I have it implemented:

- The navigator client (some kind of mission manager) is responsible for keeping track of the status of all the waypoints by subscribing to the feedback of the navigator.
- My NavigateThroughPoses goal sends `WaypointStatus[]` instead of `nav_msgs/Goals` so that we can pass the statuses
- The navigator's internal `waypoint_statuses` is initialized with the ones provided from the client and only the `goals` that are PENDING are set on the blackboard 
```

Steve's suggestions:

```
Some other thoughts for a hybrid approach:
- We could have a "ResumeNavigateThroughPoses" and "ResumeNavigateToPose" (with complimentary "Pause" ones) so that we have different semantically meaningful APIs for Start/Cancel/Pause/Resume that each have their contextual information. No need to change the main Action APIs, just have new ones that have that information. I think sharper messages are good to have rather than trying to cram everything into one.  
- The Pause can return the state and the Resume can send it back. 
- Or, the Pause can store that in the navigator itself and the Resume can load it back up in a new tree, so that the application doesn't have to care (but can still return it anyway in case they care or want to persist between reboots. If Resume is given values, it uses those instead of its internally stored one).
- Or, the Pause can simply stop ticking the tree to cold store it and the Resume can continue the ticking loop (TBD: some nodes may care about timing or have progress checking, that would need to be considered. Maybe we just `halt` the tree but keep its blackboard active to resume in a fresh new-tick state? That's essentially what your current implementation and the other ideas above do anyway by resume being a re-goal-submission)
```

SteveMacenski:
My response from the other thread: https://github.com/ros-navigation/navigation2/pull/4994#issuecomment-2917353049

[...]

PR #5247, opened 2025-06-09
Add Pause and SequenceWithBlackboardMemory BT nodes

redvinaa:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (add tickets here #5213) |
| Robotic platform tested on | custom robot |

---

* Implementation of BT nodes needed for pause / resume functionality, as discussed in #5213.

* Headers are well-documented
* Added 2 new control nodes, they need to be described in the docs with examples (create mwe from the code in the discussion)

* Have been using for a long time on humble, now on jazzy
* Haven't tested on rolling, because I don't have a rolling stack
---

* Tests still need to be written for both BT nodes!

#### For Maintainers: 


nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause.hpp
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2019 Intel Corporation
```
SteveMacenski: I think all of these should be updated :-)

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause.hpp
```diff
@@ -0,0 +1,98 @@
...
+enum state_t {UNPAUSED, PAUSED, PAUSE_REQUESTED, ON_PAUSE, RESUME_REQUESTED, ON_RESUME};
+
+/* @brief Controlled through service calls to pause and resume the execution of the tree
+ * It has one mandatory child for the UNPAUSED, and three optional for the PAUSED state,
+ * the ON_PAUSE event and the ON_RESUME event.
+ * It has two input ports:
+ * - pause_service_name: name of the service to pause
+ * - resume_service_name: name of the service to resume
+ */
+class Pause : public BT::ControlNode
```
SteveMacenski: More context in the name / plugin registered name would be nice. `PauseResumeController` `PauseResumeBehaviorTree`, etc.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause.hpp
```diff
@@ -0,0 +1,98 @@
...
+  //! @brief Service callback to resume
+  void resume_service_callback(
+    const std::shared_ptr<Trigger::Request> request,
+    std::shared_ptr<Trigger::Response> response);
+
+  rclcpp::Node::SharedPtr node_;
+  rclcpp::CallbackGroup::SharedPtr cb_group_;
+  rclcpp::executors::SingleThreadedExecutor::SharedPtr executor_;
+  std::unique_ptr<std::thread> spinner_thread_;
+  rclcpp::Service<Trigger>::SharedPtr pause_srv_;
```
SteveMacenski: Please use the Nav2 Utils Service Server - it includes important QoS and introspection parameters

nav2_behavior_tree/nav2_tree_nodes.xml
```diff
@@ -390,6 +390,15 @@
 
     <Control ID="RoundRobin"/>
 
+    <Control ID="Pause">
```
SteveMacenski: Note: Add to the docs.nav2.org for new pages for these 2 nodes + migration guide update for the nodes and how to use them for pause / resume navigation behavior within the BT

nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace nav2_behavior_tree
+{
+
+SequenceWithBlackboardMemoryNode::SequenceWithBlackboardMemoryNode(
+  const std::string & name,
+  const BT::NodeConfiguration & conf)
+: BT::ControlNode::ControlNode(name, conf)
+{
+  setRegistrationID("SequenceWithBlackboardMemory");
```
SteveMacenski: Question: What's this about? We don't currently do this in our BT nodes... should we be for newer versions of BT.CPP?

nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void SequenceWithBlackboardMemoryNode::halt()
```
SteveMacenski: Does this need to be overridden if just calling the base class?

nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.cpp
```diff
@@ -0,0 +1,100 @@
...
+          setOutput("current_child_idx", 0);
+          return child_status;
+        }
+      case BT::NodeStatus::SUCCESS: {
+          current_child_idx++;
+          setOutput("current_child_idx", current_child_idx);
+        }
+        break;
+
+      case BT::NodeStatus::SKIPPED: {
```
SteveMacenski: I think we mostly treat skipped as a failure in other control nodes: https://github.com/ros-navigation/navigation2/blob/d438a6e186e449a2977a9e6188cdbd0c99a1d307/nav2_behavior_tree/plugins/control/recovery_node.cpp#L83-L92

nav2_behavior_tree/plugins/control/pause.cpp
```diff
@@ -0,0 +1,279 @@
...
+    rclcpp::ServicesQoS(), cb_group_);
+
+  std::string resume_service_name;
+  getInput("resume_service_name", resume_service_name);
+  resume_srv_ = node_->create_service<Trigger>(
+    resume_service_name,
+    std::bind(&Pause::resume_service_callback, this, _1, _2),
+    rclcpp::ServicesQoS(), cb_group_);
+
+  spinner_thread_ = std::make_unique<std::thread>(
```
SteveMacenski: Please use the nav2 util node thread

nav2_behavior_tree/plugins/control/pause.cpp
```diff
@@ -0,0 +1,279 @@
...
+Pause::~Pause()
+{
+  RCLCPP_DEBUG(node_->get_logger(), "Shutting down Pause BT node");
+  executor_->cancel();
+}
+
+BT::NodeStatus Pause::tick()
+{
+  unsigned int children_count = children_nodes_.size();
+  if (children_count < 1 || children_count > 4) {
```
SteveMacenski: Docs on the `< 1 || > 4` is important. I'd add it both in the header doxygen, the BT walkthrough or tutorial, and the configuration guide page.

nav2_behavior_tree/plugins/control/pause.cpp
```diff
@@ -0,0 +1,279 @@
...
+    throw BT::LogicError(
+            "PauseNode must have at least one and at most four children "
+            "(currently has " + std::to_string(children_count) + ")");
+  }
+
+  std::lock_guard<std::mutex> lock(state_mutex_);
```
SteveMacenski: Instead of the node thread / thread, you can instead spin_some https://github.com/ros-navigation/navigation2/blob/35148527deec33db1bcf3c90bd78e8f44d6a63c8/nav2_behavior_tree/plugins/action/planner_selector_node.cpp#L77 so that we know that there's no parallelization issues. You can then remove the mutex and the thread and make this all single threaded. Adding a thread here for these 2 services seems extreme

nav2_behavior_tree/plugins/control/pause.cpp
```diff
@@ -0,0 +1,279 @@
...
+  if (status() == BT::NodeStatus::IDLE) {
```
SteveMacenski: This could be a switch statement

nav2_behavior_tree/plugins/control/pause.cpp
SteveMacenski: This may also simplify the state logic since when we spin, we can easily check then with a function if something was changed to request a pause/unpause handling

nav2_behavior_tree/plugins/control/pause.cpp
```diff
@@ -0,0 +1,279 @@
...
+          RCLCPP_INFO(node_->get_logger(), "Switched to state: PAUSED");
+          state_ = PAUSED;
+          break;
+        case BT::NodeStatus::FAILURE:
+          RCLCPP_ERROR(node_->get_logger(), "ON_PAUSE child returned FAILURE");
+          setStatus(BT::NodeStatus::FAILURE);
+          break;
+        default:
+          throw BT::LogicError("A child node must never return IDLE");
+      }
```
SteveMacenski: Alot of this is boilerplat between conditions, minus the log strings and the child IDX. Could this be abstracted into a function that is passed in the child to check on and the string of the state (`"PAUSED"`) to use for logging?

SteveMacenski (changes requested):
So this is missing tests coverage - but you know that :-) Also make sure to fix DCO

nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.cpp
redvinaa: No idea, we also don't use it in the other nodes :D

nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.cpp
redvinaa: I really don't think the whole sequence should fail when one of the children is skipped. In the [original implementation](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/src/controls/sequence_with_memory_node.cpp#L68) from BehaviorTree.CPP, the also don't return failure. (They do return skipped if all children were skipped, but I think this node should be stateless, so I don't think we should store skipped_count. Let me know what you think of this.)

nav2_behavior_tree/plugins/control/pause.cpp
redvinaa: You mean the ifs regarding the `state_`, right?

redvinaa:
Woah, this suddenly became more serious than I expected :D
I still have to write the tests and make this work with my jazzy stack, so it might take some time.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause.hpp
SteveMacenski: @redvinaa I meant updated with your company since you wrote these :-)

nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.cpp
SteveMacenski: Hah. It looks like I cherry picked randomly the only Nav2 BT node where we halt on a skipped child. Thanks for pushing back.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause_resume_controller.hpp
```diff
@@ -0,0 +1,129 @@
...
+  void resumeServiceCallback(
+    const std::shared_ptr<rmw_request_id_t>/*request_header*/,
+    const std::shared_ptr<Trigger::Request> request,
+    std::shared_ptr<Trigger::Response> response);
+
+  BT::NodeStatus tickChildAndTransition();
+
+  void switchState(const state_t new_state);
+
+  rclcpp::Node::SharedPtr node_;
```
SteveMacenski: Store the logger not the node please

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
```diff
@@ -0,0 +1,212 @@
...
+  if (status() == BT::NodeStatus::IDLE) {
+    std::string warn_msg = "PauseResumeController BT node has not been ticked yet";
+    response->success = false;
+    response->message = warn_msg;
+    RCLCPP_ERROR(node_->get_logger(), "%s", warn_msg.c_str());
+    return;
+  }
+
+  if (state_ != PAUSED) {
+    RCLCPP_INFO(node_->get_logger(), "PAUSE_REQUESTED");
```
SteveMacenski: I think this log can be a little more than this :wink:

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
```diff
@@ -0,0 +1,212 @@
...
+  if (status() == BT::NodeStatus::IDLE) {
+    std::string warn_msg = "PauseResumeController BT node has not been ticked yet";
+    response->success = false;
+    response->message = warn_msg;
+    RCLCPP_ERROR(node_->get_logger(), "%s", warn_msg.c_str());
+    return;
+  }
+
+  if (state_ == PAUSED) {
+    RCLCPP_INFO(node_->get_logger(), "RESUME_REQUESTED");
```
SteveMacenski: I think this log can be a little more than this :wink:

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause_resume_controller.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+/* @brief Controlled through service calls to pause and resume the execution of the tree
+ * It has one mandatory child for the UNPAUSED, and three optional for the PAUSED state,
+ * the ON_PAUSE event and the ON_RESUME event.
+ * It has two input ports:
+ * - pause_service_name: name of the service to pause
+ * - resume_service_name: name of the service to resume
+ *
+ * Usage:
+ * <Pause pause_service_name="/pause" resume_service_name="/resume">
```
SteveMacenski: Can we work a little on the naming here: Unpaused --> Active? Resume? Processing? 

Unpaused gave me a bit of back and forth to understand & the `ON_RESUME` branch is using the word "resume" on "unpaused". 

---

Also, `Pause` should now be `PauseResumeController`

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  state_ = new_state;
+  RCLCPP_INFO(node_->get_logger(), "Switched to state: %s", state_names.at(state_).c_str());
+}
+
+BT::NodeStatus PauseResumeController::tickChildAndTransition()
+{
+  // Return RUNNING and do nothing if specific child is not used
+  const uint child_idx = child_indices.at(state_);
```
SteveMacenski: Type out (here and anywhere else) `unsigned int`. This helps with windows compatibility

SteveMacenski: Can we do this logic in the `switchState` function? Switch to the correct state if not implemented so that when `tickChildAndTransition` is called we do useful work.

SteveMacenski (commented):
> Woah, this suddenly became more serious than I expected :D

I think this is really great work! Definitely happy to have it contributed! :1st_place_medal:

redvinaa:
@SteveMacenski I started writing the tests and I have a few questions.

The dummy BT node used for tests now returns `IDLE` after reset. Shouldn't it always return the status that it was set to?
The `SequenceWithBlackboardMemoryNode` logic has a `resetChildren` call, and all the dummy nodes return `IDLE` after it, making the tests fail.

Also, could you give me some pointers on how to write tests for the `SequenceWithBlackboardMemoryNode`? It needs changing the value of input ports between tests. I only found tests where the input port values are set in the `SetUp` part, but that cannot be parameterized. I also tried creating a new BT node instance for every single test, but I haven't managed to make that work yet, and even if I did, it wouldn't be a nice solution.

redvinaa:
I found the bug which didn't let me set `current_child_idx` input_port: #5295

SteveMacenski:
Sorry for the delay - I've been underwater the last couple of weeks and now committing today to get through my email review backlog. I apologize for the delay and thank you for your help!

For tests, you can do something more in the theme of the action nodes: https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/test/plugins/action/test_append_goal_pose_to_goals_action.cpp. Here, we make an [example BT XML that is loaded and used](https://github.com/ros-navigation/navigation2/blob/main/nav2_behavior_tree/test/plugins/action/test_append_goal_pose_to_goals_action.cpp#L98-L104) instead. This is actually a better way to write the tests IMO and the dummy node manual tree population done in the control test you linked to is an outdated way. We know those can have some problems since the tree isn't being constructed in the same way it would be in an application. You could still potentially use a dummy node within the tree, but the tree itself is constructed fully, including the ports and blackboard. 

That way, the blackboard of the BT is maintained between ticks since it acts like a proper behavior tree (because it is one). I think that answers your question!

redvinaa:
I've implemented the tests using xml strings as in the action nodes, and DummyNodes. I've had to be creative, because none of the existing tests needed this level of complexity. Tell me what you think of this solution @SteveMacenski. I've added a new header file to access nodes from the tree and cast them to the correct type. I also had to modify the DummyNode implementation a bit, so it returns IDLE after resetting, but on tick it again returns the state that it was set to. I don't think it breaks existing tests.

redvinaa:
@SteveMacenski can you take a look at this?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/pause_resume_controller.hpp
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) 2025 Intel Corporation
```
SteveMacenski: Updating copyrights for all files for your company

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * - If all the children return SUCCESS, this node returns SUCCESS.
+ *
+ * - If a child returns RUNNING, this node returns RUNNING.
+ *   Loop is NOT restarted, the same running child will be ticked again.
+ *
+ * - If a child returns FAILURE, stop the loop and return FAILURE.
+ *   Restart the loop only if (reset_on_failure == true)
+ *
+ */
+class SequenceWithBlackboardMemoryNode : public BT::ControlNode
```
SteveMacenski: If this is otherwise the same as the Sequence node, can we inherit from it, call the base class within the `tick` function and return the value? We can set the port if needed on the terminal states. That would make it easier to maintain unless there's a reason we cannot. 

Also, do we want this to be a normal sequence or a reactive sequence? Typically I think of users with reactive sequences or pipeline sequences with Nav2

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+public:
+  SequenceWithBlackboardMemoryNode(const std::string & name, const BT::NodeConfiguration & conf);
```
SteveMacenski: This name is also very similar to a BT.CPP provided node: https://www.behaviortree.dev/docs/nodes-library/sequencenode/#sequencewithmemory which means a different thing. Maybe we can rename here like `SequenceWithState`?

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
```diff
@@ -0,0 +1,190 @@
...
+        // Resumed child returned SUCCESS, we're done
+        return BT::NodeStatus::SUCCESS;
+      }
+      switchToNextState();
+      // If any branch other than RESUMED returned SUCCESS, keep ticking
+      return BT::NodeStatus::RUNNING;
+    case BT::NodeStatus::FAILURE:
+      RCLCPP_ERROR(
+        logger_, "%s child returned FAILURE", state_names.at(state_).c_str());
+      return BT::NodeStatus::FAILURE;
```
SteveMacenski, line 111: Question: What's the logical recovery if an On Paused or On Resume branches return a failure?

SteveMacenski (commented):
Last bit are docs:
- Add to the Nav2 Specific BT Nodes page for how to use
- Migration guide
- Configuration guide for param descriptions on the nodes + examples
- Add to the Navigation Plugins page

SteveMacenski:
Do with the AI generated code review as you see fit. If you think its pedantic or unnecessary, ignore. If it brings up a good point, please fix

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
redvinaa: Maybe it's because I've seen it so much lately, but I think this name captures perfectly what the node does. I think they also don't clash with the (non-blackboard) memory node, both are memory but one stores it on the blackboard. Let me know what you think.

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
redvinaa: Right now in that case we halt and set the internal state to RESUMED.
I think the transition branches shouldn't fail, unless there was an unexpected error that should fail this node too. At least that was my intuition, but I haven't put much thought into this edge case.

The reasoning is also valid, that if a transition failed, we should stay in the state before the transition. Though I don't know how we would handle that, because we either add recovery branches to this node (which would make it practically unusable), or return failure (but that halts the node, and maybe risks accidentally staying in PAUSED state).

Let me know what you think.

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
redvinaa: You're right, it is confusing. I think the name should definitely include blackboard because it's what mostly distinguishes this node's functionality. State I think is also confusing because all the nodes have states already (like RUNNING and SUCCESS). What about SequenceWithBlackboardIndex?

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
redvinaa: That would make it a lot easier, but as I see the current_child_idx is private in the member class. I opened a PR to change that, but it's gonna take time to get accepted (or rejected).

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
SteveMacenski: Maybe have a retry tag for the number of times to retry a state transition? We have that in other BT nodes so that wouldn't be out of place here and give the opportunity for small networking errors to be recoverable. If it ultimately fails, maybe its best to set the entire BT node to failure and halt it. Let the upstream nodes of this one figure out how to handle that. 

I don't think it would be good to just resume, if the ON_RESUME state didn't finish, for example. There could be some initializations or activations that didn't happen that wouldn't make navigating possible. I think if those transition branches fail, that's a terminal failure state (after a couple of retries).

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
SteveMacenski: I think of blackboard being an implementation detail if you said something like 'output', but I don't object. 

This is actually a good ChatGPT question:
- SequenceWithProgressOutput
- SequenceWithIndex
- IndexedSequence
- SequenceWithRunningIndex


How about something like `IndexedSequence`? I think just `WithBlackboardIndex` or `WithIndex` implies that the port is only an output, not a possible input as well. But I don't want to spend too much time on naming, so whatever you feel is best

redvinaa: You're right, blackboard is an implementation detail, but the name definitely has to include the info that the index is read and written to (and also that it's a sequence, of course). I think calling it something with "blackboard" is close enough to the truth, while being far more descriptive than the chatgpt generated ones. Do you have something better than "blackboard" that still captures this info?

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
redvinaa: > Maybe have a retry tag for the number of times to retry a state transition?

Why would this be the responsibility of the `PauseResumeController`? There is a `Retry` BT node, I think that should be used to utilize the modularity of the BT.

> I don't think it would be good to just resume, if the ON_RESUME state didn't finish

Definitely agree with this one.

SteveMacenski: If we return failure from the `PauseResumeController`, do we maintain the current errored state for a retry to come in and re-tick to attempt? I don't think so from 

```

  if (status() == BT::NodeStatus::IDLE) {
    state_ = RESUMED;
  }
```

Which I think is actually the correct behavior. Its easier to retry within this node knowing its current state since a failure caused by the RESUME or PAUSED branch don't need retries, only the ON_RESUME and ON_PAUSED branches need retries generated by a failed transition. I don't think it makes sense to do that externally since that's tied to internal implementation information. It would be very difficult for another node to know if a failure from this one should trigger a retry or not without leaking alot of its state onto the blackboard. That starts to get a little complex and I think retry within a Controller BT node is pretty common across our Nav2 and BT.CPP nodes.

redvinaa: Sorry, I meant retry as part of the ON_PAUSE and ON_RESUME branches, like this
```xml
<PauseResumeController pause_service_name="/pause" resume_service_name="/resume">
    <!-- RESUMED branch -->

    <!-- PAUSED branch (optional) -->

    <RetryUntilSuccessful num_attempts="3">
        <!-- ON_PAUSE branch (optional) -->
    </RetryUntilSuccessful>

    <RetryUntilSuccessful num_attempts="3">
        <!-- ON_RESUME branch (optional) -->
    </RetryUntilSuccessful>
</Pause>
```
This would be transparent and wouldn't complicate the node logic unnecessarily.

What's the advantage of doing it as part of the PauseResumeController?

SteveMacenski: Oh, uh, that makes sense too :-) 

Can you actually add that as part of your docs examples to have the ON_ transition branches use retries? That would meet my request!

nav2_behavior_tree/include/nav2_behavior_tree/plugins/control/sequence_with_blackboard_memory.hpp
SteveMacenski: `StatefulSequence`, `PersistentSequence`, `ResumableSequence`

I'm thinking Stateful or Resumable?

redvinaa: I think stateful is not a lot better than state, for the same reasons. Resumable also means something different, like stopping and then continuing again. This sequence doesn't stop, it just does the processing in chunks (node by node). So nothing to resume in my opinion.

If it's out of these three, I'd go with persistent. But it's still hard to decipher from the name what exactly persists.
Then I'd rather go back to `IndexedSequence`, the word index I think also takes us closer to the truth.

SteveMacenski: > Resumable also means something different, like stopping and then continuing again. 

Isn't this how it works? If you give it a child idx of 3, wouldn't it start at child idx 3? 

> If it's out of these three, I'd go with persistent. But it's still hard to decipher from the name what exactly persists.

Works for me :+1: I think that's what the doc guides are to some degree. I don't think anyone could know what Truncate Path vs Truncate Path Local means without looking at some docs without knowing anything about them

SteveMacenski:
@redvinaa I also noticed the BT.CPP change was merged in - but no rush

redvinaa:
**WARNING: don't accept this yet, still waiting for submodule changes**

I've finally implemented the requested changes, but it turns out to inherit from the `SequenceNode`, more changes are required (namely the tick method should also be protected, and the config should be passed to the parent `ControlNode`, which for some reason in the `SequenceNode` isn't). I'm gonna open another PR agains BT.CPP and see what happens.

Also, how to "upgrade" the BT.CPP version used in this repo? I added a commit hash from the BT.CPP repo HEAD in the `tools/underlay.repos` but I'm not sure this is the way.

P.S. I also wrote the docs

SteveMacenski:
Pull in / rebase once https://github.com/ros-navigation/navigation2/pull/5409 is merged - we're having an issue with circleci time limits and I'm load balancing the builds. 

For a temp basis, the BT custom build is OK but not ideal. I'd just make sure we poke Davide to release a new rolling version once those are merged so that we can have that be as temporary as possible. Or, we have the main version you have now and we can have a follow up PR once new binaries are available so Nav2 `main` users don't need to build BT.CPP from source in the interim. I'm OK with a follow up to strip out the redundant code as well :-) 

Let me know!

redvinaa:
Okay! Then let's wait for the BT.CPP release, and then I'm gonna open a follow up.

Until then, this PR should be ready to merge :D I removed the parts that need the new BT version.

SteveMacenski (approved):
Just the small docs update and I can merge the pair!

SteveMacenski:
@redvinaa thanks so much!

By the way... if you had a nice video / gif of this, this would make a really nice linkedin post to promote the changes and contribution :wink:

nav2_behavior_tree/plugins/control/pause_resume_controller.cpp
```diff
@@ -0,0 +1,190 @@
...
+{
+
+using namespace std::placeholders;
+
+PauseResumeController::PauseResumeController(
+  const std::string & xml_tag_name,
+  const BT::NodeConfiguration & conf)
+: BT::ControlNode(xml_tag_name, conf)
+{
+  auto node = this->config().blackboard->get<rclcpp::Node::SharedPtr>("node");
```
tonynajjar, line 34:
```suggestion
  auto node = config().blackboard->get<nav2::LifecycleNode::SharedPtr>("node");
```

As per https://github.com/ros-navigation/navigation2/pull/5288

redvinaa: Nice catch, opened a PR to fix it

PR #5250, opened 2025-06-10
Add footprintAreaCost

tonynajjar:
---

## Basic Info

| ------ | ----------- |

---

---

#### For Maintainers: 

### DIFF
--- a/nav2_costmap_2d/src/footprint_collision_checker.cpp
+++ b/nav2_costmap_2d/src/footprint_collision_checker.cpp
@@ -85,6 +85,106 @@ double FootprintCollisionChecker<CostmapT>::footprintCost(const Footprint & foot
 }
 
 template<typename CostmapT>
+bool FootprintCollisionChecker<CostmapT>::isRectangularFootprint(
+  const Footprint & footprint, double & min_x, double & max_x, 
+  double & min_y, double & max_y) const
+{
+  // Only 4-point footprints can be rectangles
+  if (footprint.size() != 4) {
+    return false;
+  }
+
+  // Find bounding box
+  min_x = std::numeric_limits<double>::max();
+  max_x = std::numeric_limits<double>::lowest();
+  min_y = std::numeric_limits<double>::max();
+  max_y = std::numeric_limits<double>::lowest();
+
+  for (const auto & point : footprint) {
+    min_x = std::min(min_x, point.x);
+    max_x = std::max(max_x, point.x);
+    min_y = std::min(min_y, point.y);
+    max_y = std::max(max_y, point.y);
+  }
+
+  // Check if all points are at corners of bounding box (axis-aligned rectangle)
+  const double tolerance = 1e-6;
+  int corner_count = 0;
+  
+  for (const auto & point : footprint) {
+    bool at_corner = 
+      (std::abs(point.x - min_x) < tolerance || std::abs(point.x - max_x) < tolerance) &&
+      (std::abs(point.y - min_y) < tolerance || std::abs(point.y - max_y) < tolerance);
+    
+    if (at_corner) {
+      corner_count++;
+    }
+  }
+
+  return corner_count == 4;
+}
+
+template<typename CostmapT>
+double FootprintCollisionChecker<CostmapT>::footprintAreaCost(const Footprint & footprint)
+{
+  // Find bounding box of footprint
+  double min_x = std::numeric_limits<double>::max();
+  double max_x = std::numeric_limits<double>::lowest();
+  double min_y = std::numeric_limits<double>::max();
+  double max_y = std::numeric_limits<double>::lowest();
+
+  for (const auto & point : footprint) {
+    min_x = std::min(min_x, point.x);
+    max_x = std::max(max_x, point.x);
+    min_y = std::min(min_y, point.y);
+    max_y = std::max(max_y, point.y);
+  }
+
+  // Convert to grid coordinates
+  unsigned int min_mx, min_my, max_mx, max_my;
+  if (!worldToMap(min_x, min_y, min_mx, min_my) ||
+      !worldToMap(max_x, max_y, max_mx, max_my)) {
+    return static_cast<double>(LETHAL_OBSTACLE);
+  }
+
+  double max_cost = 0.0;
+
+  // Check if footprint is rectangular for optimization
+  bool is_rectangular = isRectangularFootprint(footprint, min_x, max_x, min_y, max_y);
+
+  // Check all cells within bounding box
+  for (unsigned int mx = min_mx; mx <= max_mx; ++mx) {
+    for (unsigned int my = min_my; my <= max_my; ++my) {
+      bool cell_in_footprint;
+      
+      if (is_rectangular) {
+        // For rectangular footprints, all cells in bounding box are inside
+        cell_in_footprint = true;
+      } else {
+        // For non-rectangular footprints, check point-in-polygon
+        double wx, wy;
+        costmap_->mapToWorld(mx, my, wx, wy);
+        cell_in_footprint = isPointInFootprint(wx, wy, footprint);
+      }
+
+      if (cell_in_footprint) {
+        double cost = pointCost(mx, my);
+        
+        if (cost == static_cast<double>(LETHAL_OBSTACLE)) {
+          return cost;
+        } else if (cost == static_cast<double>(NO_INFORMATION)) {
+          // Return NO_INFORMATION for unknown cells, let caller decide how to handle
+          return cost;
+        }
+        max_cost = std::max(max_cost, cost);
+      }
+    }
+  }
+
+  return max_cost;
+}
+
+template<typename CostmapT>
 double FootprintCollisionChecker<CostmapT>::lineCost(int x0, int x1, int y0, int y1) const
 {
   double line_cost = 0.0;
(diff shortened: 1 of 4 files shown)


nav2_costmap_2d/src/footprint_collision_checker.cpp
```diff
@@ -84,6 +84,106 @@ double FootprintCollisionChecker<CostmapT>::footprintCost(const Footprint & foot
...
+  unsigned int min_mx, min_my, max_mx, max_my;
+  if (!worldToMap(min_x, min_y, min_mx, min_my) ||
+      !worldToMap(max_x, max_y, max_mx, max_my)) {
+    return static_cast<double>(LETHAL_OBSTACLE);
+  }
+
+  double max_cost = 0.0;
+
+  // Check if footprint is rectangular for optimization
+  bool is_rectangular = isRectangularFootprint(footprint, min_x, max_x, min_y, max_y);
```
tonynajjar: optimization: doesn't need to be checked every time

nav2_costmap_2d/src/footprint_collision_checker.cpp
```diff
@@ -143,6 +243,41 @@ double FootprintCollisionChecker<CostmapT>::footprintCostAtPose(
   return footprintCost(oriented_footprint);
 }
 
+template<typename CostmapT>
+bool FootprintCollisionChecker<CostmapT>::isPointInFootprint(
```
tonynajjar, line 196: taken from collision_monitor. Libraries also exist for this e.g boost

nav2_smac_planner/src/collision_checker.cpp
```diff
@@ -150,7 +148,8 @@ bool GridCollisionChecker::inCollision(
       current_footprint.push_back(new_pt);
     }
 
-    float footprint_cost = static_cast<float>(footprintCost(current_footprint));
+    // Check full area covered by footprint
+    float footprint_cost = static_cast<float>(footprintAreaCost(current_footprint));
```
tonynajjar, line 152: can be parametrized to allow user to tune for speed vs accuracy

tonynajjar:
I asked my LLM to benchmark the 2 functions and this is the result. Take it with a grain of salt, I didn't sanity-check it but the results look reasonable.

```
=== Collision Method Benchmark ===
Comparing footprintCost (edge-only) vs footprintAreaCost (full area)

Running 10000 iterations per test...

Testing small_square... Done
  Vertices: 4
  Area (cells): 16.00
  Edge method: 0.038 μs/call
  Area method: 0.048 μs/call
  Speedup ratio: 1.28x (Area method SLOWER)
  Costs match: Yes

Testing medium_rect... Done
  Vertices: 4
  Area (cells): 72.00
  Edge method: 0.060 μs/call
  Area method: 0.152 μs/call
  Speedup ratio: 2.52x (Area method SLOWER)
  Costs match: Yes

Testing large_rect... Done
  Vertices: 4
  Area (cells): 200.00
  Edge method: 0.110 μs/call
  Area method: 0.303 μs/call
  Speedup ratio: 2.75x (Area method SLOWER)
  Costs match: Yes

Testing xlarge_rect... Done
  Vertices: 4
  Area (cells): 800.00
  Edge method: 0.351 μs/call
  Area method: 1.086 μs/call
  Speedup ratio: 3.10x (Area method SLOWER)
  Costs match: No
    Edge cost: 253.00, Area cost: 254.00

Testing hexagon... Done
  Vertices: 6
  Area (cells): 41.57
  Edge method: 0.013 μs/call
  Area method: 0.391 μs/call
  Speedup ratio: 30.62x (Area method SLOWER)
  Costs match: Yes

Testing octagon... Done
  Vertices: 8
  Area (cells): 101.82
  Edge method: 0.013 μs/call
  Area method: 0.793 μs/call
  Speedup ratio: 61.10x (Area method SLOWER)
  Costs match: Yes

Testing complex_12_sided... Done
  Vertices: 12
  Area (cells): 75.00
  Edge method: 0.013 μs/call
  Area method: 0.733 μs/call
  Speedup ratio: 54.49x (Area method SLOWER)
  Costs match: Yes

Testing complex_16_sided... Done
  Vertices: 16
  Area (cells): 76.54
  Edge method: 0.013 μs/call
  Area method: 0.895 μs/call
  Speedup ratio: 69.63x (Area method SLOWER)
  Costs match: Yes

Testing narrow_robot... Done
  Vertices: 4
  Area (cells): 64.00
  Edge method: 0.131 μs/call
  Area method: 0.109 μs/call
  Speedup ratio: 0.83x (Area method FASTER)
  Costs match: Yes

=== SUMMARY ===
Average edge method time: 0.082 μs
Average area method time: 0.501 μs
Overall speedup ratio: 6.08x
Area method faster in: 1/9 cases
Edge method faster in: 8/9 cases
Costs match in: 8/9 cases

WARNING: Some costs don't match! This indicates different behavior between methods.
```

tonynajjar:
@SteveMacenski do you have time for this soon? I have about an hour still where I could finish this

SteveMacenski:
@tonynajjar I just commented in the issue - looking at this now

nav2_costmap_2d/src/footprint_collision_checker.cpp
```diff
@@ -84,6 +84,106 @@ double FootprintCollisionChecker<CostmapT>::footprintCost(const Footprint & foot
...
+  }
+
+  double max_cost = 0.0;
+
+  // Check if footprint is rectangular for optimization
+  bool is_rectangular = isRectangularFootprint(footprint, min_x, max_x, min_y, max_y);
+
+  // Check all cells within bounding box
+  for (unsigned int mx = min_mx; mx <= max_mx; ++mx) {
+    for (unsigned int my = min_my; my <= max_my; ++my) {
```
SteveMacenski: I think `for y; for x` is actually faster, no?

SteveMacenski (commented):
Other than your comments and 1-2 nitpicks I would make, this seems sane for this method. I'm curious what LLM / notebook / tools did you do for this?

tonynajjar:
I use Github Copilot in VScode with agent mode with the latest Claude Sonnet or GPT models. I still iterated quite a bit to get to this state but it's a huge help for sure...

Anyway, you advocated for the other strategy (swept area), so what would be your plan for this PR? Complete it and merge it until someone  can implement the other strategy? Or discard it?
You're right that this method has the risk of jumping over a lethal space but practically from my tests that didn't happen yet so I'll use this solution until we implement better.

tonynajjar:
image

I just got a case where this approach fails (and it doesn't even involve single cells) because the angle change between 2 planned footprints is too big. The red footprint is TEB's feasibility checker kicking in showing where the collision would be if we try to follow SMAC's plan.

tonynajjar:
I increased the turning radius to 0.4 to see if that was a problem only with small turning radii but even in this case I'm not confident that the interpolation between the planned footprint is collision-free but I guess it depends on the interpolation model

image

tonynajjar:
With those changes I'm noticing a weird behavior but I don't see how it relates: For goals moderatly close to lethal space (i don't know what the distance limit is), SMAC doesn't reach the desired orientation.

Example: in the following image, pose 432 is sent as the goal from SMAC but the final pose does not match the orientation. (blue rectangle is the last planned footprint)

I'm sure that it is because of this PR as I tested without and it reaches any orientation

image

Do you have a guess why this PR could cause that?


Update1: I noticed that when that happens, there are many more arrows in the `expansions` visualizations. What could that mean?

image


Update2: Okay I think I'm slowly understanding better how SMAC works - I think the fact that there are expansion arrows up until the the goal means that SMAC never successfully used analytical expansion, which I assume would be responsible of reaching the right orientation accurately? If yes then next question: what causes the the analytical expansion to fail?

SteveMacenski:
Alot here to process, so I'll go in order

> Anyway, you advocated for the other strategy (swept area), so what would be your plan for this PR? Complete it and merge it until someone can implement the other strategy? Or discard it?

I think from your later findings you might be on Team Swept Area now? For right now, I think we can leave this open for discussion until we find the final solution and/or use this in the interim or permanently. 

> I just got a case where this approach fails (and it doesn't even involve single cells) because the angle change between 2 planned footprints is too big

:arrow_up: So swept area might be better? 

> I increased the turning radius to 0.4

Its not that 0.4m is a magic number (though it did come from somewhere; I selected it based on the turning radius of common RC car toys that are able to get through tight spaces in houses, I figured if it can do it, then that's a good baseline for us. It also makes good turns in and out of aisles as a baseline that isn't just a point-turn but also isn't so arced out that its highly dependent on the width of aisles to successfully compute). It just interpolates more than the jumps for basically zero-radius turns. It doesn't surprise me that you could reproduce it, but I would expect it to be harder. 

---

You're right on the analytic expansion. Are you using a tolerance > 0? If so, then it could return a solution nearby if it was unable to compute an exact match. 

> I'm sure that it is because of this PR as I tested without and it reaches any orientation ... what causes the the analytical expansion to fail?

I ... have no idea. Perhaps there's a bug in your Area check that is checking some wrong cells, so when you're near an obstacle, its checking part of the window it shouldn't. When analytic expansion tries to compute the exact path, then it is unable to do so because they're being reported in collision. The only relationship with the collision checking and analytic expansion is that the expansions are checked for collision using the collision checker. 

So, I think that would indicate to me that something about this is buggy. I don't see an immediate problem though, so I'd need to add in some prints to log in the analytic expansion to see what it thinks is going on.

tonynajjar:
> Perhaps there's a bug in your Area check that is checking some wrong cells

Yes that was it and it was because of of the optimization for rectangular footprints - it wasn't checking the right cells (in short, it was checking the cells in the origin-oriented circumscribed rectangle to the footprint). Anyway I removed the optimization for now just so it works.
As I don't have the resources right now to work on the more-involved "swept area" method and we suspect that most of our collisions would be solved with this PR, our pragmatic plan is to test this PR in the field first and see if we need to implement the "swept area" method. I'd of course be happy if someone else finds the time to implement it in the meantime - maybe I should create a standalone issue for more visibility?

SteveMacenski:
Sounds good - can you close the original ticket you reopened if you open a dedicated ticket to this issue?

SteveMacenski:
@tonynajjar should we close this now with your new PR #5416?

tonynajjar:
Ah completely forgot about that one, yes can be closed

PR #5261, opened 2025-06-12
Parametrizing obstacle layer tf filter tolerance

MarcoMatteoBassa:
## Basic Info

| ------ | ----------- |
| Robotic platform tested on | Simulation and Hardware |

---

Setting a tf filter tolerance != 0 might be undesired as it increases the chance of the messages being delayed or even discarded if the tf_timeout tolerance associated with the buffer is not big enough. Hence the option to set it to a different value through a parameter.

Also note that this tolerance is not set for pointcloud sources.

Added new parameter

Simulation and hardware -> it was verified that setting the parameter to 0.0 with a transform tolerance of 100ms prevents the filter from dropping messages, while a bigger tolerance is required with the default value of 0.05.

---

#### For Maintainers: 


SteveMacenski (approved):

SteveMacenski:
Docs PR: https://github.com/ros-navigation/docs.nav2.org/pull/709

